PR #2, opened 2013-10-22
Add preliminary support for some newer RTL8111/8168 Express Gigabit Ethe...

victoredwardocallaghan:
If someone could review this please, it needs better testing. I was only really initially trying to fix my own bug: http://www.freebsd.org/cgi/query-pr.cgi?pr=183167

Kind Regards,

Signed-off-by: Edward O'Callaghan eocallaghan@alterapraxis.com


emaste:
Addressed in r257304-257306.

PR #4, opened 2013-11-14
Add try-include and uses

devinteske:
Add a new try-include function and use it to conditionally load /boot/loader.rc.local and /boot/menu.rc.local


emaste:
@devinteske I assume this is OBE?

bsdimp:
This has already been contributed to the project. Closing for house keeping purposes only. Thanks for all your other work on the boot loader...

PR #7, opened 2014-04-03
zfs: Fix vnode destruction on znode cache destruction

raphaelsc:
Looks like ZTOV(zp) couldn't be NULL at that time given that the value is passed to vn_free later on to free the underlying vnode.
For the purpose mentioned above, the statement 'zp->z_vnode = NULL' also needs to be moved around. Otherwise, ZTOV would always return NULL when zfs_znode_cache_destructor is reached.

Signed-off-by: Raphael S. Carvalho raphaelsc@cloudius-systems.com

### DIFF
--- a/sys/cddl/contrib/opensolaris/uts/common/fs/zfs/zfs_znode.c
+++ b/sys/cddl/contrib/opensolaris/uts/common/fs/zfs/zfs_znode.c
@@ -145,8 +145,9 @@ zfs_znode_cache_destructor(void *buf, void *arg)
 	znode_t *zp = buf;
 
 	ASSERT(!POINTER_IS_VALID(zp->z_zfsvfs));
-	ASSERT(ZTOV(zp) == NULL);
+	ASSERT(ZTOV(zp) != NULL);
 	vn_free(ZTOV(zp));
+	zp->z_vnode = NULL;
 	ASSERT(!list_link_active(&zp->z_link_node));
 	mutex_destroy(&zp->z_lock);
 	rw_destroy(&zp->z_parent_lock);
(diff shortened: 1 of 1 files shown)


avg-I:
@raphaelsc what problem this change would fix?

sys/cddl/contrib/opensolaris/uts/common/fs/zfs/zfs_znode.c
```diff
@@ -145,8 +145,9 @@ zfs_znode_cache_destructor(void *buf, void *arg)
 	znode_t *zp = buf;
 
 	ASSERT(!POINTER_IS_VALID(zp->z_zfsvfs));
-	ASSERT(ZTOV(zp) == NULL);
+	ASSERT(ZTOV(zp) != NULL);
```
avg-I: Are you sure `z_vp` *must* be non-NULL here?

raphaelsc: More than 2 years since I submitted this patch, so I'll need to look at the details. I think my patch is probably wrong because this assert would've been triggered otherwise. 

It's only strange to me that we call vn_free on a NULL vnode, look:
	ASSERT(ZTOV(zp) == NULL);
	vn_free(ZTOV(zp));

I'll take a look at this later.

avg-I: Yeah, the code looks weird because the assertion is FreeBSD specific where we can free a znode only after it is fully disconnected from a vnode and the `vn_free` call is from illumos where the filesystem driver has to explicitly return the vnode to the system (as opposed to FreeBSD where the system tells I want to reclaim this vnode and the driver has to obey).

bsdimp:
I'd like to close this request and move it to a bugzilla bug. Any objections?

davidgreens:
No, No, No @CrownBonded, @Hektve87, @davidgreens

avg-I:
@bsdimp bugzilla... so, what's the bug?

avg-I:
@davidgreens with all respect, but who are you? what does you comment mean? and what is your relation to this PR?

davidgreens:
Me I am. These build tools that builds code quickly and reliably. It is used to build the majority of Google's software, and thus it has been designed to handle build problems present in Google's development. COMPRENDE (avg-1)

bsdimp:
Closing. This mistake is no longer in our tree since the OpenZFS migration.

PR #12, opened 2014-10-29
don't zero the IP TOS bits on mbufs that need checksumming

Seb-LineRate:
On mbufs that don't have the CSUM_DATA_VALID flag set, tcp_input()
needs to compute the TCP checksum itself.

The TCP checksum includes some but not all fields from the IP header.

tcp_input() zeroes the fields of the IP header that are not to be
included in the checksum, then checksums the entire packet (IP
header, TCP header, and TCP data), then restores the IP header
fields it will need later.

The bug that this commit fixes is that the IP TOS field was not
restored, so the later read of it returned a zeroed byte instead
of the incoming packet's actual TOS value.

### DIFF
--- a/sys/netinet/tcp_input.c
+++ b/sys/netinet/tcp_input.c
@@ -675,12 +675,14 @@ tcp_input(struct mbuf **mp, int *offp, int proto)
 			/*
 			 * Checksum extended TCP header and data.
 			 */
+			u_int8_t tos = ip->ip_tos;
 			len = off0 + tlen;
 			bzero(ipov->ih_x1, sizeof(ipov->ih_x1));
 			ipov->ih_len = htons(tlen);
 			th->th_sum = in_cksum(m, len);
 			/* Reset length for SDT probes. */
 			ip->ip_len = htons(tlen + off0);
+			ip->ip_tos = tos;
 		}
 
 		if (th->th_sum) {


cemeyer:
lidl@ suggests instead:

```
Index: tcp_input.c
===================================================================
--- tcp_input.c	(revision 305860)
+++ tcp_input.c	(working copy)
@@ -675,6 +675,7 @@
 			/* XXX stat */
 			goto drop;
 		}
+		iptos = (ntohl(ip6->ip6_flow) >> 20) & 0xff;
 	}
 #endif
 #if defined(INET) && defined(INET6)
@@ -701,6 +702,7 @@
 		th = (struct tcphdr *)((caddr_t)ip + off0);
 		tlen = ntohs(ip->ip_len) - off0;
 
+		iptos = ip->ip_tos;
 		if (m->m_pkthdr.csum_flags & CSUM_DATA_VALID) {
 			if (m->m_pkthdr.csum_flags & CSUM_PSEUDO_HDR)
 				th->th_sum = m->m_pkthdr.csum_data;
@@ -722,6 +724,10 @@
 			th->th_sum = in_cksum(m, len);
 			/* Reset length for SDT probes. */
 			ip->ip_len = htons(tlen + off0);
+			/* Reset TOS bits */
+			ip->ip_tos = iptos;
+			/* Re-initialization for later version check */
+			ip->ip_v = IPVERSION;
 		}
 
 		if (th->th_sum) {
@@ -728,22 +734,9 @@
 			TCPSTAT_INC(tcps_rcvbadsum);
 			goto drop;
 		}
-		/* Re-initialization for later version check */
-		ip->ip_v = IPVERSION;
 	}
 #endif /* INET */
 
-#ifdef INET6
-	if (isipv6)
-		iptos = (ntohl(ip6->ip6_flow) >> 20) & 0xff;
-#endif
-#if defined(INET) && defined(INET6)
-	else
-#endif
-#ifdef INET
-		iptos = ip->ip_tos;
-#endif
-
 	/*
 	 * Check that TCP offset makes sense,
 	 * pull out TCP options and adjust length.		XXX
```

cemeyer:
Fixed in https://svnweb.freebsd.org/changeset/base/306458 .

emaste:
As mentioned by @cemeyer this has been committed

PR #13, opened 2014-11-21
Bump max and default callchain depths

emaste:
Bump the default from 16 to 32, to accommodate kernel flamegraphs.
Bump the maximum from 32 to 128, to accommodate deep user stacks.

Differential Revision: https://reviews.freebsd.org/D1203


emaste:
Committed in r275140

PR #18, opened 2015-01-28
Add the -s srcdir option to config, ...

hackagadget:
...similar to config in NetBSD, which allows for specifying the directory to use as the location for configuration source files.


rodrigc:
Committed in 74334908895548dd608758c0a12c4d0a1b834134

PR #20, opened 2015-03-01
escaping for \n characters

olevole:


scaleengine:
Because this is contributed software, pulled from a different upstream project, could you please submit this upstream, and once it is fixed there it will be merged into FreeBSD.

Upstream is https://github.com/Juniper/libxo

PR #24, opened 2015-03-24
Changes to allow network interfaces to be renamed safely.

JamesKoenig:
Sponsored by:  Norse Corporation

### DIFF
--- a/etc/network.subr
+++ b/etc/network.subr
@@ -1532,6 +1532,25 @@ ifnet_rename()
 	return 0
 }
 
+#
+# ifnet_rename_safe ifacename outputvar
+#    rename interface $ifacename to whatever is in the config database
+#    and store the resulting name into the variable $outputvar
+#
+ifnet_rename_safe()
+{
+	local _if _ifname _tovar
+	_if="$1"
+	_tovar="$2"
+	_ifname=`get_if_var $_if ifconfig_IF_name`
+	if [ ! -z "$_ifname" ]; then
+		${IFCONFIG_CMD} $_if name $_ifname
+	else
+		_ifname="$_if"
+	fi
+	eval $_tovar=\"$_ifname\"
+}
+
 # list_net_interfaces type
 #	List all network interfaces. The type of interface returned
 #	can be controlled by the type argument. The type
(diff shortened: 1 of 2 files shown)


bsdimp:
I don't understand this change, and it doesn't come with an explanation. It's unclear what bug this fixes. Can you provide more explanation about why it's needed?

emaste:
Norse is no longer, although @JamesKoenig might still recall context for this change.

@JamesKoenig, thank you for your submission. I agree with the comment above that more information is required before it can be included in FreeBSD. The information about what bug the change fixes (and how) should be included in the proposed commit message.

etc/network.subr
```diff
@@ -1532,6 +1532,25 @@ ifnet_rename()
...
+#    and store the resulting name into the variable $outputvar
+#
+ifnet_rename_safe()
+{
+	local _if _ifname _tovar
+	_if="$1"
+	_tovar="$2"
+	_ifname=`get_if_var $_if ifconfig_IF_name`
+	if [ ! -z "$_ifname" ]; then
+		${IFCONFIG_CMD} $_if name $_ifname
```
yaneurabeya: Please use `${_if} name ${_ifname}` here (I don't know why, but `list_net_interfaces`, etc follows this convention, so it's best to be consistent here).

yaneurabeya:
`$cmdifn` has to deal with the name for the interface (a less commonly used pattern for /etc/rc.d/netif is to call it like `/etc/rc..d/netif stop <interface>`, e.g. `/etc/rc.d/netif stop em0`. I'll look at this pull request further to see if I can understand why this is needed.

splbio:
Sorry if it's not obvious but can you see how netif_start() will rename the interface and then call `netif_common ifn_start $cmdifn` right at the end?

Well, if the interface is renamed, then `$cmdifn` will be set to the old name and the interface doesn't get started, kind of silly and breaks the feature.

bsdimp:
@splbio Is it just a simple matter of using the feature to see it doesn't work? I could have sworn it's worked for me in the past. If it is, then I'll rename an interface on a board I have here and see if the change is needed. If so, I'll merge and close, otherwise, I'll just close.

PR #25, opened 2015-04-09
Remove semicolon on BITSET_DEFINE()

brlcad:
This allows/requires callers of BITSET_DEFINE to include a trailing semicolon without introducing a double-semicolon.  This was encountered in _cpuset.h where BITSET_DEFINE(_cpuset, CPU_SETSIZE); on line 50 (rev 23b61e792eee74f3d6bdaf0ec6d265e6c70d7d3d ) results in a compilation error (ISO C does not allow extra ';' outside of a function) when compiling code that calls cpuset.h with the -pedantic-errors flag.

### DIFF
--- a/sys/sys/_bitset.h
+++ b/sys/sys/_bitset.h
@@ -50,7 +50,7 @@
 #define	BITSET_DEFINE(t, _s)						\
 struct t {								\
         long    __bits[__bitset_words((_s))];				\
-};
+}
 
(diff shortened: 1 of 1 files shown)


cemeyer:
This was fixed: https://github.com/freebsd/freebsd/commit/5ca026f4b29a75fb0520d2f69db9710ab7d9e32c

brlcad:
Thanks for fixing this.

PR #26, opened 2015-04-18
Quiet compiler warnings in mbuf.h

hackagadget:
* Mark unused arguments with __unused
 * Move inline before the return type (and use __inline to keep the file
   consistent in style.)


uqs:
This was merged

PR #27, opened 2015-04-18
Changes to support file verification in MAC

hackagadget:
Add VCREAT flag to indicate when a new file is being created
Add VVERIFY to indicate verification is required
Both VCREAT and VVERIFY are only passed on the MAC method vnode_check_open
   and are removed from the accmode after
Add O_VERIFY flag to rtld open of objects
Add 'v' flag to __sflags to set O_VERIFY flag.

Submitted by:	Steve Kiernan <stevek@juniper.net>
Obtained from:	Juniper Networks, Inc.


uqs:
This was merged

PR #28, opened 2015-04-21
Move zlib.c from net to libkern.

hackagadget:
It is not network-specific code and would be better as part of libkern instead.
Move zlib.h and zutil.h from net/ to sys/
Update includes to use sys/zlib.h and sys/zutil.h instead of net/

Submitted by: Steve Kiernan stevek@juniper.net
Obtained from: Juniper Networks, Inc.


uqs:
This got merged.

PR #29, opened 2015-05-07
PL310 power control register offset incorrect

hackagadget:
Offset for the power control register was specified incorrectly (it had the
same value as the prefetch control register.) This change corrects the offset
value to 0xF80, per the ARM PL310 documentation.

Submitted by: Steve Kiernan <stevek@juniper.net>
Obtained from: Juniper Networks, Inc.


emaste:
Committed in r282586

PR #31, opened 2015-05-09
Implement xargs -P0: run as many jobs in parallel as possible

lifanov:
This mirrors GNU xargs -P0 feature and increases portability of
scripts that use it. There are a couple of differences though:
 o This change accepts negative numbers as well as 0 to mean the
   same thing, which is consistent with other FreeBSD xargs options
 o This change sets number of parallel commands to maxargs instead
   of some *_MAX macro: this should almost always be what the user
   actually wants to do and increases performance by a lot.

The same patch is uploaded to Bugzilla #199976, which can be
closed if this is merged.

### DIFF
--- a/usr.bin/xargs/xargs.1
+++ b/usr.bin/xargs/xargs.1
@@ -206,7 +206,11 @@ Parallel mode: run at most
 .Ar maxprocs
 invocations of
 .Ar utility
-at once.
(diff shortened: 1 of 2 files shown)


lifanov:
Better version of this change has been committed.

PR #32, opened 2015-05-12
Fix 13yo bug with grdc(6) 12-hour mode from 12:00 to 13:00

DominoTree:


DominoTree:
Bugzilla #200133

emaste:
Committed in r283929

koobs:
Merged

PR #34, opened 2015-05-17
Update README

xmulym:


emaste:
Commit message / pull request text is insufficient to describe this change - you need to explain why the change is being made (what problem is being solved), not just what is being changed.

uqs:
git doesn't do keyword expansion. This diff is bogus.

PR #41, opened 2015-07-23
Add blackhole(4) documentation for SCTP

Savagedlight:
Documented the sysctl MIB net.inet.sctp.blackhole in blackhole(4) man page, and updated MIB description to refer to said man page.

Also adjusted formatting under SYNOPSIS while here, to prevent cluttered line breaks in middle of optional parameters.

Relates to bugzilla ID 184110 (https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=184110)

### DIFF
--- a/share/man/man4/blackhole.4
+++ b/share/man/man4/blackhole.4
@@ -11,30 +11,36 @@
 .\"    documentation and/or other materials provided with the distribution.
 .\"
 .\"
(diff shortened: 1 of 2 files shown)


Savagedlight:
Committed by Allan Jude in svn r287528 and r287529

PR #53, opened 2015-10-28
Allow user to specify ZFSBOOT_POOL_SIZE for unattended installs.

johnko:
The default is to create a zroot that consumes the whole disk because if used with geli(8) this makes sense.

Without geli(8), I like to keep my data pool separate from my system pool.

This is different than ZFSBOOT_BOOT_POOL_SIZE which is named bootpool.


johnko:
@allanjude  I've squashed my patch to just this, if you're interested.

Thx.

bsdimp:
Paging @allanjude @devinteske what say you all?

devinteske:
Will have a look tomorrow, thanks

tykling:
@johnko I am working on putting an option to limit zroot pool size in the installer, when using the zfs auto partitioning. Seems like it should be possible to use your code along with a few modifications to the installer. Thoughts?

johnko:
Sorry, I haven't looked at this code in about 2 years, I trust @allanjude would grok it better.

johnko:
@tykling probably, I used these modification at the time (2015) for my installs but I think I skipped the Interactive Installer and had my own wrapper script to do my installs:
- Here's where I downloaded my custom zfsboot: https://github.com/johnko/freebsd-install-script/blob/master/bin/fzg#L591
- Here's where I call `bsdinstall zfsboot`: https://github.com/johnko/freebsd-install-script/blob/master/bin/fzg#L621-L622
- The rest of the wrapper script is just setting up GELI disks and `bsdinstall distfetch`, `checksum`, `distextract`, etc.

bsdimp:
rebased and updated https://reviews.freebsd.org/D30588

bsdimp:
7ef92163ab81 resolves this issue by committing this patch, as reviewed by @allanjude

PR #55, opened 2015-12-14
Changed LIBADD curses

PRikLY-CacTUs:
changed LIBADD+= curses to +LIBADD= curses.
The original resulted in a buildworld error. The alteration seems to fix this.


bdrewery:
This should have no impact. Do you recall what the error was?

bdrewery:
Oh, you added a literal `+` in there, that is not valid.

PRikLY-CacTUs:
Yeah sorry, I had no idea what I was doing. I was getting a "make world" error pointing to that file. It was fixed by someone using ncurses not curses.

PR #59, opened 2016-02-18
Fix typo in SCU_PTSGRTC_READ

tony:


tony:
Moved to bugzilla https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=207336

PR #65, opened 2016-03-29
Fat elf support

m0khi:
This adds support of FatElf to FreeBSD Elf implementation.


m0khi:
In hope to be helpful

PR #67, opened 2016-04-25
Fix libthr's coredump due thr_mutex.c:447

m0khi:


lib/libthr/thread/thr_mutex.c
```diff
@@ -444,8 +444,13 @@ enqueue_mutex(struct pthread *curthread, struct pthread_mutex *m)
 	/* Add to the list of owned mutexes: */
 	mutex_assert_not_owned(m);
 	qidx = mutex_qidx(m);
-	TAILQ_INSERT_TAIL(&curthread->mq[qidx], m, m_qe);
```
cemeyer: This change is a no-op.

m0khi:
Actually this change was due a bug we've encountered in a downstream (NextBSD).
But the was not actually because of this reason (however it can be considered as a safe and correct workaround).
The bug is now solved in downstream (and its actual cause is fixed), I guess this PR can now be closed cause the actual cause of the bug was in downstream not in FreeBSD

PR #70, opened 2016-05-09
Wakeup only one thread while events are available

andy-yx-chen:
There is no reason to wakeup all threads while there are events available, as that would make same event to be consumed by multiple threads at the same time and result in errors.

### DIFF
--- a/sys/kern/kern_event.c
+++ b/sys/kern/kern_event.c
@@ -1881,7 +1881,7 @@ kqueue_wakeup(struct kqueue *kq)
 
 	if ((kq->kq_state & KQ_SLEEP) == KQ_SLEEP) {
 		kq->kq_state &= ~KQ_SLEEP;
-		wakeup(kq);
+		wakeup_one(kq);
 	}
 	if ((kq->kq_state & KQ_SEL) == KQ_SEL) {
 		selwakeuppri(&kq->kq_sel, PSOCK);


bdrewery:
I don't know about the change, but the fallout appears to be "Thundering Head" rather than "multiple threads consume it".  The waiter is:

<pre>
kq->kq_state |= KQ_SLEEP;
error = msleep_sbt(kq, &kq->kq_lock, PSOCK | PCATCH,
    "kqread", asbt, rsbt, C_ABSOLUTE);
</pre>

It will lock on `kq->kq_lock` before returning.  If it cannot acquire the lock it just goes back to waiting for the next wakeup.

andy-yx-chen:
That would be acceptable if the waked up thread will sleep again, however, the problem is for socket listener, the event will keep available until an accept call is made, that means many threads will be waken up and get the same event, but only one could successfully accept an event while others will be suspended on accept call forever, here is an example,
http://www.andytech.net/a_292.html

bdrewery:
@kostikbel Mind reviewing this?

kostikbel:
I already wrote all I think about this in PR 209558.

1. The change breaks the interface, since level-triggered events stop reporting randomly.
2. The thundering herd problem is only there when several threads share one kqueue, which is weird and untypical API use on its own.

andy-yx-chen:
what is the recommended way to use kqueue with a thread pool?

andy-yx-chen:
Again **what I am going to improve is the scenario of event based kqueue on multicore systems**
Copy my reply in PR 209558 to here,my two cents
1. the good stuff for kqueue is it could be used to served almost all sorts of events, like file access, timer, socket etc. that leads to a very common design that having a kqueue event based asynchronous event processing framework, like libuv, as well as nginx event module. For FreeBSD, in order to implement this, most likely that people would do is having a kqueue to hold all handles and one thread (because of this issue) to poll events from event queue, then, since most of the servers would not just have one CPU core, in order to leverage all CPU resources, the framework may have a thread pool (size == # of CPUs), then the polling thread needs to do event dispatching to thread pool, which works fine, however, a little bit odd, comparing to how Windows IOCP does and linux's EPOLL, in which all threads in thread pool keep reading from IOCP (or EPOLL, http://lxr.free-electrons.com/source/include/linux/wait.h#L230), and process any events that are available (no job re-dispatching required)

2. echoing to the point of "thread exits before event has been processed", even you wake up all threads, all of them may also die before reach kqueue, which is not a good reason to keep this behavior. moreover, we can make a kernel to be robust, say, impossible to be crashed by usermode's fault, but it sounds  impossible to me that the kernel can take care of usermode's errors, all the kernel needs to guarantee is ensure the system is still good even there are tens of thousands of errors in usermode.

3. moreover, having a thread to handle a socket is good, however, for many cases, server applications would serve hundreds of connections in one second, which doesn't means we need to create hundreds of threads for an application, and even worse that might causes unbalanced CPU load, such as two cores, with three requests, in async mode, it's very likely that core #1 takes req 1 and 3, while core #2 takes req 2, while req 1 is very heavy, while req #2 is lightweight, then request #3 may get stuck (because of core #1 is busy on req 1, though req 2 has already finished by core #2)

I believe many async IO applications/frameworks are having a very different way for kqueue because of this (my point #1) such as asio c++ library http://think-async.com/

Async IO is extreme important for today's server applications, which also have a very strong requirement not to create too many threads (#of threads == #of cpu cores), having this fix would help many developers to write asyc io applications/frameworks in the same way as windows and linux, which I think is valuable

cemeyer:
> what is the recommended way to use kqueue with a thread pool?

One kqueue per thread, with each monitored object assigned to only a single kqueue.  That could mean a single kqueue event loop sharding work across a thread pool, or several event loops with independent event sources.

andy-yx-chen:
@cemeyer that makes sense, however is a little bit different than other platforms

PR #73, opened 2016-05-31
align text in etc/motd

BenMorganIO:
A completely aestetic improvement, but I noticed that when the text was printed, some of it wasn't aligned anymore.


emaste:
I believe this is intentionally not aligned so that the (rather long) questions list URL does not wrap over 80 columns.

BenMorganIO:
Hmm ok. The text is 84 characters long. Maybe just remove 4 spaces?

bsdimp:
I'm thinking that since this was intentional, and has to be misaligned, that no change should be made. Thanks for the submission, however. Normally, keeping things neat is quite important. It's just that not wrapping past 80 columns was more so.

PR #76, opened 2016-06-07
Add support for Vodafone/Huawei K3771.

djzort:
See also http://www.draisberghof.de/usb_modeswitch/device_reference.txt


djzort:
anyone care to look?

mrkiko:
We're lucky: this modem is supported on Linux by cdc_ether.
This means it should be possible to add it's pid and get it working, right?
See commit e2949080792256d1c979aaf30ecd4cab42829f87 in Linux kernel git tree.

djzort:
This modem is a serial one, its just a matter of adding the usb id's to the driver (which is what this pull request does)

Linux does the usb modeswitch in user space, freebsd does it in the driver. The usb_modeswitch project maintains the list of usb id's (as linked above) - so there is a good argument to simply input all those ID's straight in to the freebsd driver.

mrkiko:
Looking at Linux commit message however, my impression was that this modem isn't purely serial.
And if you can use it in NDIS, it maybe better.

Different modems then use different protocols... serial mode is supported most of the times I think, but I am not sure it's always supported.

djzort:
i do believe that some of the modems can operate in different modes. for me though, this model works perfectly in Linux as a ttyUSB device - in freebsd nothing. So i submitted this patch.

mrkiko:
And you're right.
Thank you for your work.


Enrico Mioso
Mobile Phone Number: +393807096934 ( +Telegram :) )
My Tox ID is: 7C593F402A3C8632D87AB4B948D492294C39A6A614464ECF843CA3429FB023284180472C7475

I like / recommend usage of open messaging standards when possible.

On Sun, 14 Aug 2016, Dean Hamstead wrote:

==Date: Mon, 15 Aug 2016 06:51:41
==From: Dean Hamstead <notifications@github.com>
==Reply-To: freebsd/freebsd
==    <reply+0101e3b98e95f91c10e73762ce40ad15d03fd3ce01b9447e92cf0000000113c90c5
==    d92a169ce09772639@reply.github.com>
==To: freebsd/freebsd <freebsd@noreply.github.com>
==Cc: mrkiko <mrkiko.rs@gmail.com>, Comment <comment@noreply.github.com>
==Subject: Re: [freebsd/freebsd] Add support for Vodafone/Huawei K3771. (#76)
==
==
==i do believe that some of the modems can operate in different modes. for me
==though, this model works perfectly in Linux as a ttyUSB device - in freebsd
==nothing. So i submitted this patch.
==
==—
==You are receiving this because you commented.
==Reply to this email directly, view it on GitHub, or mute the
==thread.[AQHjuW1-LKHHb-tUa7NHogiKaAEMXxhiks5qf_BdgaJpZM4IvcDC.gif]
==
==
==

sterglee:
Hi all,

I repeatedly installed freebsd  on iMac, on an external USB hard disk,
and at unpredictable times, I got a serious problem on booting it,
specifically:

Mounting from ufs:/dev/da1p1 failed with error 19

mountroot>



Unfortunately, I cannot find a way to recover the installation.

I tried without success:
kern.cam.boot_delay="10000"

and some other suggestions from Google searching,
but unfortunately, they do not succeed.

What can happen?



Best Regards
Stergios

djzort:
this is not the correct forum for this question.

sys/dev/usb/usbdevs
```diff
@@ -2406,7 +2406,9 @@ product HUAWEI E1752		0x1446	3G modem
 product HUAWEI K4505		0x1464	3G modem
 product HUAWEI K3765		0x1465	3G modem
 product HUAWEI E1820		0x14ac	E1820 HSPA+ USB Slider
+product HUAWEI K3771		0x14c4  K3771 Initial
```
larsengels: I think this should be K3771_INIT

bsdimp:
Pull Request Accepted and Committed to FreeBSD.
https://svnweb.freebsd.org/changeset/base/314407

djzort:
awesome thanks

PR #80, opened 2016-07-12
Allow user to select the boot pool if more than one found on the boot device

2asoft:
Currently the first zfs pool is chosen as the boot pool.
There may be cases (I am one such case) when a single disk is used to boot multiple machines, each from a separate zfs pool. The user thus needs to be able to select which pool to boot from.


2asoft:
This wasnt actually closed. @uqs commit something from openzfs that referenced openzfs issue #80, and magically this gets closed :(

PR #81, opened 2016-07-26
Fix support of user-class option in dhclient

baarde:
`dhclient` does not send the `user-class` option (`77`) because it's not in the default dhcp option priority list, as reported in [Bug 184117](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=184117#c1).


cemeyer:
FreeBSD doesn't take PRs directly, sorry.  But I've posted the change in our code review system and have taken assignment of the bug.  Here's the link: https://reviews.freebsd.org/D7911

baarde:
Thank you. I'll close this PR.

PR #84, opened 2016-08-18
Refresh geom labels

benrubson:
Hello,

This change solves the following bug :
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=162690
It adds a "refresh" option to glabel command which refreshes labels from devices.

Would be glad to have it merged with 10.3 and 11 !

Thank you very much 👍 

Best regards,

Ben


benrubson:
patch submitted directly into bug report

PR #86, opened 2016-08-18
Refresh geom labels (stable/10)

benrubson:
Same as #84 but in stable/10 branch.

Thank you 👍 

Ben


benrubson:
// not needed as we can simply MFC #84.

PR #87, opened 2016-08-22
Update sshd_config

benrubson:
Hello,

This PR enforces ciphers/algorithms/MACs so that weak/unsecured ones are disabled by default.
Would then be nice to have this in FreeBSD-11.

Many thanks 👍 

Ben


benrubson:
Related bug report 212390 :
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=212390

benrubson:
`Resolution: --- → Works As Intended`

PR #94, opened 2016-10-30
only sort smaller section via recursion

jaypatrickhoward:
Please see:

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=213922


bsdimp:
Uploaded to bugzilla, closing this request, since its desired to play out there.

PR #96, opened 2016-11-08
renumber copyright clause 4

jschauma:
It seems that when the copyright was changed from the 4-clause to the
3-clause BSD license, that only the third clause was renamed, leaving the
4th clause in place _as the 4th clause_.  That is, copyright clauses were
then '1. Redistributions...', '2. Redistributions...', and '4.
Neither...'.

This commit renumbers the 3rd clause to be labeled '3. Neither...'.


bsdimp:
Yes. Permission was granted to remove clause 3. At the time, it was unclear if that also included renumbering the remaining clauses or not. In a legal document, one generally doesn't renumber when you remove a section once the instrument has been executed. However, since then the consensus has been that I'm being overly pedantic and renumbering is fine.

bsdimp:
merged, after fixing conflicts.

PR #98, opened 2016-11-14
Update user.subr

oleantus:
Adding the ability to copy from the skel directory not only individual .dot files, but also directories containing files such as KDE directory with configuration files.

### DIFF
--- a/usr.sbin/bsdconfig/usermgmt/share/user.subr
+++ b/usr.sbin/bsdconfig/usermgmt/share/user.subr
@@ -145,12 +145,12 @@ f_user_copy_dotfiles()
 		[ -e "$file" ] || continue # no-match
 
 		f_shell_escape "$file" "_file"
-		f_eval_catch $funcname cp "cp -n '%s' '%s'" \
+		f_eval_catch $funcname cp "cp -R -n '%s' '%s'" \
 			"$_file" "$_user_home_dir/${_file#dot}"
 		retval=$?
 		[ $retval -eq $SUCCESS ] || break
 		f_eval_catch $funcname chown \
-			"chown -h '%i:%i' '%s'" \
+			"chown -R -h '%i:%i' '%s'" \
 			"$_user_uid" "$_user_gid" \
 			"$_user_home_dir/${_file#dot}"
 		retval=$?


emaste:
Paging @allanjude @devinteske

allanjude:
Are the files in the subdirectory going to be named 'dot.foo' ?
If so, we'll need to loop over those separately, to remove the 'dot' part.

I am not sure what the expected behaviour is here.

oleantus:
> Are the files in the subdirectory going to be named 'dot.foo' ?

Yes.

> If so, we'll need to loop over those separately, to remove the 'dot' part.

Don't need this.

I am developer of ulbsd.ru (freebsd clone). I am tested this in our product. It's worked!

emaste:
I don't like the inconsistency of renaming skel's dot.foo files to .foo, but not files in subdirectories. That said, I think we should probably just move away from the "dot" prefix in skel, rather than extending this change to rename files in subdirectories too.

oleantus:
> I think we should probably just move away from the "dot" prefix in skel, rather than extending this change to rename files in subdirectories too.

It is a good idea. But now this patch solves the problem with the subdirectories in the skel directory.

yaneurabeya:
I like this change.

As far as moving away from `dot.` prefixed paths, I think that there's a lot more work that would be required to make this a reality, and there might be unnecessary breakage due to shifting away from `dot.` prefixing.

bsdimp:
I'll commit this patch as is. Most .foo/bar setups don't have .foo/.bar, so the need for dot.bar is much reduced.

yaneurabeya:
@bsdimp: please check with @devinteske

devinteske:
Reviewing

devinteske:
As discussed on IRC with imp, Allan Jude, John Baldwin, and myself...

bsdconfig mirrors the actions of pw(8) with respect to skel (please see the pw(8) manual with respect to skeleton files and the effect the `-m' parameter has on copying of skel files).

At the very least, the suggested change here has to be accompanied by a corresponding change to several other tools, including but not limited to pw(8), adduser(8), and their respective manuals.

bsdimp:
Redirected to bugzilla. https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=217602

PR #99, opened 2016-11-14
Update tws.h

jcatrysse:
`#define TWS_MAX_NUM_LUNS`
defines the maximum number of LUN's on the device and to my knowledge it shouldn't be limited to only 16 LUN's. In my config it was provoking problems because not all the disks (volumes) were detected. I changed the value to 32 in my config and everything seems to work now. However, I think the value should be 255.

I am not aware of the reason why it was 16 in the first place.


bsdimp:
This is a duplicate of pull request #100 which has a better description.

PR #100, opened 2016-11-16
Update tw_cl_share.h

jcatrysse:
TW_CL_MAX_NUM_LUNS should not be 16 but I presume 255. I have a 3ware controller with more than 16 volumes (LUN's) and otherwise all LUN's above the 16'th are not working.

### DIFF
--- a/sys/dev/twa/tw_cl_share.h
+++ b/sys/dev/twa/tw_cl_share.h
@@ -68,7 +68,7 @@
 #define TW_CL_MAX_NUM_UNITS		32	/* max # of units we support */
 #endif /* TW_OSL_ENCLOSURE_SUPPORT */
 
-#define TW_CL_MAX_NUM_LUNS		16	/* max # of LUN's we support */
+#define TW_CL_MAX_NUM_LUNS		255	/* max # of LUN's we support */
 #define TW_CL_MAX_IO_SIZE		0x20000	/* 128K */
 
 /*


bsdimp:
This change looks good to me. It doesn't increase the size of the data structures and shouldn't add much to the scanning time. I think we should accept and merge.

AraHaan:
Yeah this looks promising also does anyone know if FreeBSD has resource sections embedded into compiled assemblies just like how Windows has it but in a little different way? I like the idea of their ''.rsrc'' section makeup and wish other platforms could follow something similar. I been also trying to work into my own fork of the Linux kernel as well as it could benefit people and also deprecate the need for the ''<binary name>.desktop'' files for things like showing an icon on it if people wants the assembly or assemblies to have an icon similar to how Windows handles them.

bsdimp:
Author: imp
Date: Wed Mar  1 04:40:57 2017
New Revision: 314469
URL: https://svnweb.freebsd.org/changeset/base/314469

Log:
  Update tw_cl_share.h to allow 255 LUNs

  TW_CL_MAX_NUM_LUNS should not be 16 but I presume 255. I have a 3ware
  controller with more than 16 volumes (LUN's) and otherwise all LUN's
  above the 16'th are not working.

  Submitted by: jcatrysse <j.catrysse@proximedia.be>
  Pull Request: https://github.com/freebsd/freebsd/pull/100

Modified:
  head/sys/dev/twa/tw_cl_share.h

PR #102, opened 2016-11-22
Modify comment in "usr.bin/apply/apply.c"

SeongWook-Hong:
I think that the step number '3' is more suitable by a sequence.


bsdimp:
duplicate. thanks though.

PR #103, opened 2016-12-04
REMOVE #if 0 code

kloun:


bsdimp:
Thanks for taking the time to submit this change. However, I think it's incorrect. This is the version control information that shows the origin of pwd.c. We should reject this pull request.

PR #104, opened 2016-12-06
Fix include/arpa/nameser_compat.h by adding T_DNAME definition.

jrushford:
T_DNAME is missing from include/arpa/nameser_compat.h.  Ran across this when adding DNAME support to apache trafficserver, FreeBSD builds failed and had to add a workaround.

### DIFF
--- a/include/arpa/nameser_compat.h
+++ b/include/arpa/nameser_compat.h
@@ -175,6 +175,7 @@ typedef struct {
 #define T_ATMA		ns_t_atma
 #define T_NAPTR		ns_t_naptr
 #define T_A6		ns_t_a6
+#define T_DNAME		ns_t_dname
 #define T_OPT		ns_t_opt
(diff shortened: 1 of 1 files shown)


bsdimp:
Change looks right in that it does what it says it does. But I'm curious. What needs this? What program or class of programs is being fixed by this? That would be helpful to include in the commit message.

bsdimp:
Committed to head. Missed what it fixed, so no need to reply.
Author: imp
Date: Tue Feb 28 20:34:25 2017
New Revision: 314408
URL: https://svnweb.freebsd.org/changeset/base/314408

Log:
  Fix include/arpa/nameser_compat.h by adding T_DNAME definition.
  Fixes Apache trafficserver

  Submitted by: John J. Rushford <https://github.com/jrushford>
  Pull Request: https://github.com/freebsd/freebsd/pull/104

PR #105, opened 2017-01-06
if_rsu: add board type

idwer:


erikarn:
Committed in r311830. Thanks!

PR #107, opened 2017-03-24
Replace bubble sort to selection sort for sysinit table.

Wuyang-Chung:
Change bubble sort to selection sort for better performance since it will do less variable swap operations.

### DIFF
--- a/sys/net/if_epair.c
+++ b/sys/net/if_epair.c
@@ -831,7 +831,8 @@ epair_clone_create(struct if_clone *ifc, char *name, size_t len, caddr_t params)
 	ifp->if_start = epair_start;
 	ifp->if_ioctl = epair_ioctl;
 	ifp->if_init  = epair_init;
-	ifp->if_snd.ifq_maxlen = ifqmaxlen;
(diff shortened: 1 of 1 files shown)


cemeyer:
Thanks for the suggestion.

One thing to keep in mind is that bubble sort is a stable sort, keeping any existing order in the input among equal-comparing entries.

The second thing to think about is:  the sysinit list has a small number of entries and is sorted once.  Can you demonstrate that any improvement is needed there?

The third thing to consider is that selection sort is also an O(N²) sort.  Why do you expect it to be better than bubble sort?

Wuyang-Chung:
From the definition of bubble sort in wiki, the implementation in the original code is not bubble sort. It is more like selection sort but not exactly selection sort. I am just making the code to use the exact definition of selection sort.

Assuming that the order of the sysinit table is pretty random, This implementation will have less variable swap operations so it should be a tiny faster then the original code. Not much improvement though. In the original code, it might need to do the variable swap operations several times within the inner for loop. In this implement, it only needs to do it once outside of the inner for loop.

PR #109, opened 2017-04-09
Fix bug 218512.

fhriley:
Add test cases to reproduce all the problems (and then some)
noted in bug 218512.

Allow setting passphrases and changing iterations in all cases
except when two passphrases are set or being set. In that case,
disallow changing iterations.

Modify the behavior of setkey's '-i' to be deterministic. If a
user uses setkey and does not specify '-i', geli will find how
many iterations it can do in 2 seconds. It is possible for
separate invocations of setkey to come up with different
iterations, which could result in a setkey failure because
iterations is not allowed to be changed. This is fixed by
using the iterations value from the metadata if it has already
been set, and then falling back to the calculation if no
iterations have been set in the metadata.

Update the geli man page to reflect the above changed behavior.

Add the md_passphrases field to the metadata and increase the geli
version to v8. This allows geli to know for certain if a passphrase
is set on the provider. The current implementation uses
md_iterations for this, but that only means a passphrase has been
set at some point in the past. The metadata will automatically be
updated from v7 to v8 on the first metadata disk write.

Adding md_passphrases has a side effect of aligning the remaining
metadata fields to 4 bytes, which should be better than the odd
byte alignment they currently have.

Use the md_passphrases field to determine if a passphrase is set on
a provider.

Use the md_passphrases field to determine if no passphrases are
remaining on a provider and the iterations field in the metadata
can be cleared.

Fix geli dump to output iterations correctly. Add md_passphrases
to dump.

### DIFF
--- a/sbin/geom/class/eli/geli.8
+++ b/sbin/geom/class/eli/geli.8
@@ -303,7 +303,11 @@ Number of iterations to use with PKCS#5v2 when processing User Key
 passphrase component.
 If this option is not specified,
 .Nm
-will find the number of iterations which is equal to 2 seconds of crypto work.
+will use the number of iterations that has been previously set in the
+.Nm
+metadata.
+If that has not been set, it will find the number of iterations
+which is equal to 2 seconds of crypto work.
 If 0 is given, PKCS#5v2 will not be used.
 PKCS#5v2 processing is performed once, after all parts of the passphrase
 component have been read.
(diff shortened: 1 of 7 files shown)


cemeyer:
Fixed in a slightly different way, see discussion on Bugzilla: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=218512 .  (This can be closed.)

emaste:
https://svnweb.freebsd.org/changeset/base/317246

allanjude:
The correct reference is: 1fbe8b4
This was auto-closed by a reference to an OpenZFS PR with the same ID

PR #110, opened 2017-04-15
Implement Bug 218679

fhriley:
Add a 'verify' command to geli. This allows a user to verify a
key and/or passphrase without having to call commands that will
modify the system (such as attach or setkey).

Update geli man page to add new command.

### DIFF
--- a/sbin/geom/class/eli/geom_eli.c
+++ b/sbin/geom/class/eli/geom_eli.c
@@ -75,6 +75,7 @@ static void eli_resize(struct gctl_req *req);
 static void eli_version(struct gctl_req *req);
 static void eli_clear(struct gctl_req *req);
 static void eli_dump(struct gctl_req *req);
+static void eli_verify(struct gctl_req *req);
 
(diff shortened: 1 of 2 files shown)


cemeyer:
+cc @allanjude

bsdimp:
This looks good to my eye, but I'd feel better about merging it if it were reviewed by someone on the security team.

sbin/geom/class/eli/geom_eli.c
```diff
@@ -1724,3 +1736,52 @@ eli_dump(struct gctl_req *req)
...
+	if (nargs != 1) {
+		gctl_error(req, "Invalid number of arguments.");
+		return;
+	}
+	prov = gctl_get_ascii(req, "arg0");
+
+	if (eli_metadata_read(req, prov, &md) == -1)
+		return;
+
+	if (md.md_keys == 0) {
```
delphij: This code is mostly a copy of eli_setkey_detached.  Could you please extract the common portion out instead?

oshogbo:
We have right now a 'dry-run' option in the attach method.

PR #112, opened 2017-04-21
Add missing argument for H_SET operation of the history function.

abhinav-upadhyay:
The H_SET operation of the history() function takes an int argument which is the
position of the item to which the cursor should be moved to.


cemeyer:
Thanks!  I've gone ahead and committed this as r317570.  Github pull requests are unmonitored, though, so in the future please file a PR on Bugzilla instead.

brooksdavis:
Committed upstream.

brooksdavis:
For the record d63632bb1e2401ccde034e11e47174201c3811c6.

PR #113, opened 2017-07-24
Uniformize KASSERT use in PI mutex functions

pmarillo:
Only one function in kern_umtx.c directly calls panic().
Shouldn't it use the KASSERT macro instead, like any other function in this file?


kostikbel:
Committed as r325386, using MPASS() instead of KASSERT().

PR #114, opened 2017-08-21
Correct typo in ctladm's usage screen

peterpakos:
I've just noticed a small typo in ctladm's usage screen.


jeamland:
I've committed this as Subversion r322804. Thanks!

PR #116, opened 2017-09-29
New memcpy

jonwoodruff:
When we're happy with this branch, we should merge it with main.  The memcpy itself has been tested in the three major configurations in the baremetal test suite, and I have built a disk image with this memcpy and successfully booted.  Also I have checked tha our benchmarks do not have obvious regressions, and that the performance issue with the SHA benchmark is resolved using this memcpy.

At the moment we are still using the old memcpy for memmove_c, and should probably also use the new memcpy for this one.


jonwoodruff:
Sorry created against wrong repository

PR #117, opened 2017-10-03
TD_TID is not used after loaded to %eax. This line is redundant.

Wuyang-Chung:


kostikbel:
Committed as r325387.

PR #123, opened 2017-11-06
Save 3 bytes of code space each time ALLOCPAGES gets expanded

Wuyang-Chung:
Trade "movl $((foo)*PAGE_SIZE),%ecx" with "movl %ecx, %eax" and save 3 bytes of code space each time ALLOCPAGES gets expanded.


Wuyang-Chung:
ALLOCPAGES is expanded 5 times, so 15 bytes can be saved.

PR #124, opened 2017-11-06
Change variable foff to type vm_ooffset_t.

Wuyang-Chung:
In i386, vm_offset_t is 32-bit and vm_ooffset_t is 64-bit. foff should be 64-bit instead of 32-bit. All the places where foff appears are expecting a variable of type vm_ooffset_t instead of vm_offset_t.


markjdb:
Fixed in r325528.

PR #125, opened 2017-11-09
Suggest changing to selection sort

Wuyang-Chung:
I have tested both bubble sort and selection sort on my test machine. Selection sort runs faster. The test method is as follows:
1. Read the tsc before and after the for loop
2. Calculate the tsc difference
3. Reboot the system to get more tsc difference samples of the for loop.

Here is my test result. I reboot the system 4 times for each sorting algorithm and the average of tsc difference are
bubble sort: 7,220,292
selection sort: 3.976,695

Selection sort takes only 55% of the time compared to bubble sort.

### DIFF
--- a/sys/kern/init_main.c
+++ b/sys/kern/init_main.c
@@ -229,13 +229,11 @@ restart:
 		struct sysinit *save;
 
 		min = sipp;
-		for (xipp = sipp + 1; xipp < sysinit_end; xipp++) {
+		for (xipp = sipp + 1; xipp < sysinit_end; xipp++)
 			if ((*xipp)->subsystem < (*min)->subsystem ||
 			    ((*xipp)->subsystem == (*min)->subsystem &&
-			     (*xipp)->order < (*min)->order)) {
-			        min = xipp;
-			}
-		}
+			     (*xipp)->order < (*min)->order))
+				min = xipp;
 		save = *sipp;
 		*sipp = *min;
 		*min = save;
--- a/sys/kern/init_main.c
+++ b/sys/kern/init_main.c
@@ -203,10 +203,7 @@ symbol_name(vm_offset_t va, db_strategy_t strategy)
 void
 mi_startup(void)
 {
-
 	struct sysinit **sipp;	/* system initialization*/
-	struct sysinit **xipp;	/* interior loop of sort*/
-	struct sysinit *save;	/* bubble*/
 
 #if defined(VERBOSE_SYSINIT)
 	int last;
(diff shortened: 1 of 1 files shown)


bsdimp:
Is qsort profitable here?

sys/kern/init_main.c
```diff
@@ -223,19 +220,25 @@ mi_startup(void)
 
 restart:
 	/*
-	 * Perform a bubble sort of the system initialization objects by
+	 * Perform a selection sort of the system initialization objects by
 	 * their subsystem (primary key) and order (secondary key).
 	 */
-	for (sipp = sysinit; sipp < sysinit_end; sipp++) {
+	for (sipp = sysinit; sipp < sysinit_end-1; sipp++) {
+		register struct sysinit **xipp;	/* interior loop of sort*/
```
emaste: please don't use `register`

sys/kern/init_main.c
```diff
@@ -223,19 +220,25 @@ mi_startup(void)
 
 restart:
 	/*
-	 * Perform a bubble sort of the system initialization objects by
+	 * Perform a selection sort of the system initialization objects by
 	 * their subsystem (primary key) and order (secondary key).
 	 */
-	for (sipp = sysinit; sipp < sysinit_end; sipp++) {
+	for (sipp = sysinit; sipp < sysinit_end-1; sipp++) {
```
emaste: style(9) comment, spaces around the `-`

DimitryAndric:
While this looks nice, isn't this a largely unnecessary micro-optimization?  How many times is this code called, and is it performance-critical at all?

emaste:
It's called once, at system startup time - but I think it's still worth saving a tiny bit of boot time if it's trivial to do so -- bubble and selection sort are about equal complexity (in terms of understanding their algorithms) so no downside there from a switch.

That said, we "could" do the sorting at build time instead, saving even more time.

bsdimp:
It would be hard to sort at build time since this is a linker set that's added to by loaded modules.

Wuyang-Chung:
qsort is a recursive function and it is recommended not to use recursive function in kernel.

cemeyer:
The order of magnitude of boot time saved, when converted from cycles to seconds, is 0.001s on a 3 GHz CPU, or 0.03s on a lowly 100Mhz CPU.  This is maybe not a useful optimization.

(Also, qsort does not have to be implemented recursively — it can trivially be implemented with a separate heap stack and no recursion.)

PR #127, opened 2017-11-17
Added a Makefile and atf-script which contains test-cases for head(1).

fjs-github:
I added a Makefile and atf-script which contains test-cases for head(1).
I'm new to this and not sure if there are additional files which are needed to integrate the test for head(1) into the test-suite. I used the test-cases for tail(1) as a guide.

There's no particular reason that I selected head(1). I read on the mailing list to just pick a utility which doesn't yet have test-cases.

### DIFF
--- a/usr.bin/head/tests/head_test.sh
+++ b/usr.bin/head/tests/head_test.sh
@@ -113,8 +113,9 @@ read_from_stdin_head() {
 	atf_set "descr" "Test head(1)'s reading of stdin"
 }
 read_from_stdin_body() {
-	jot -b test 2 > outfile
-	jot -b test 10 | head -2 > expectfile
+	#head(1) defaults to head -n 10 if no args are given.
+	jot -b test 10 > outfile
+	jot -b test 20 | head > expectfile
 	atf_check cmp outfile expectfile 
 }
 
--- a/usr.bin/head/tests/head_test.sh
+++ b/usr.bin/head/tests/head_test.sh
@@ -1,0 +1,100 @@
+#! /usr/libexec/atf-sh
+# Copyright (c) 2016 Alan Somers
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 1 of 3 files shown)


usr.bin/head/tests/head_test.sh
```diff
@@ -0,0 +1,100 @@
+#! /usr/libexec/atf-sh
```
asomers: No need for a shebang.  Make will add it.

usr.bin/head/tests/head_test.sh
```diff
@@ -0,0 +1,100 @@
...
+# Copyright (c) 2016 Alan Somers
```
asomers: Copyright Alan Somers?  I'm flattered, but you should probably change it.

usr.bin/head/tests/head_test.sh
```diff
@@ -0,0 +1,100 @@
...
+}
+
+atf_test_case sparse_file_text_at_beginning
+sparse_file_text_at_beginning_head() {
+	atf_set "descr" "Test head(1)'s handling of a sparse file with text at the beginning of the file"
+}
+sparse_file_text_at_beginning_body () {
+	jot -b test 10 > outfile
+	truncate -s +1K outfile
+	atf_check -s exit:0 -o ignore head -c 2048 outfile
```
asomers: 2048 bytes is much more than the size of the file.  It would be a more interesting test if you told head to stop reading somewhere in the middle of the sparse section.  In that case,  you should also verify that it does indeed read only the requested amount.

jillest:
Thanks for helping getting things tested.

A change to etc/mtree/BSD.tests.dist is required so that the directory /usr/tests/usr.bin/head is created.

fjs-github:
Thanks for the feedback Alan. I'll make the changes you suggested and resubmit the branch.

fjs-github:
Made changes per review/comments by asomers. 
Please review and let me know if you think there still needs to be changes made.
--------------------------------------   
1. Removed the shebang line.
2. Changed the name on the copyright.
3. Changed the code in sparse_file_text_at_beginning_body so that head(1) stops reading in the sparse section of the file.
4. Added test cases for invalid args.
5. Added a test case for reading from stdin.

asomers:
Most of these changes look good.  However, I disagree with using `-2` in the final test.  That form is deprecated by POSIX, and isn't mentioned in the FreeBSD man page.  You should take it out.  What I meant in my previous comment was to test reading from stdin when `-` is the only comment.  However, I now see POSIX lists that behavior as optional.  FreeBSD doesn't support it, and on reflection I agree with that approach.  Linux does, but that makes it impossible to use head on a file that's actually named `-`.

So it would be best to simply remove the arguments to head in `read_from_stdin_body`.  Just test that head reads from stdin when there's no file argument.  Also, don't forget to add the entry to `etc/mtree/BSD.tests.dist` as suggested by @jillest .

fjs-github:
Per review by asomers, I changed the code in read_from_stdin to remove the deprecated option syntax.

I'll add the test for head to BSD.tests.dist but I thought I should wait to see if this branch gets accepted and merged.

asomers:
Nope.  Commiting this as is, without BSD.tests.dist, would break the build.

fjs-github:
Added an entry for head(1) to BSD.tests.dist

asomers:
https://svnweb.freebsd.org/base?view=revision&revision=326039

PR #128, opened 2017-11-18
Redundant or (|) operation.

Wuyang-Chung:
The original statement is
regs->tf_eflags = PSL_USER | (regs->tf_eflags & PSL_T);

Before the assignment statement, regs->tf_eflags is 0 (because of bzero), so
    regs->tf_eflags = PSL_USER | (0 & PSL_T);
can be reduced to
    regs->tf_eflags = PSL_USER;


Wuyang-Chung:
The same thing also happened in sys/amd64/amd64/machdep.c and sys/pc98/pc98/machdep.c

bsdjhb:
Hmm, this is possibly a bug instead.  Single-stepping across an exec() system call should probably "work".  This has been broken since the line was added in r4201 though.  Let me write a test case though and probably fix this the right way (of preserving PSL_T rather than clobbering it)

larsengels:
@bsdjhb Have you made any progress on this?

PR #129, opened 2017-12-02
Remove archaic, difficult to maintain, unlikely to be used code from …

Sebbyastian:
…OpenSSL

"dead meat belongs in frying pans or ovens" -- xmj, sorry, but I'm sure you'll agree with this.

For a start, it's extremely unlikely that anyone is going to try to compile FreeBSD with Microsoft Visual C++ 5.0.

Secondly, there's no logical cause and effect here. The function the code tries to communicate with (defined here: <https://msdn.microsoft.com/en-us/library/windows/desktop/ms647550(v=vs.85).aspx>) can't see this variable. If it can, it can only do so by inspecting some hard-coded value from "the stack" or something "magical" (A.K.A. undefined behaviour). That makes this a nightmare to maintain, as introducing new variables before it will cause it to break, for a compiler nobody even uses anymore!

Finally, throughout my research (yes, I did do research, contrary to the assumption otherwise), I found no evidence that this "hack" would resolve the problem. I crafted searches with different combinations of keywords such as "MSVC++ 5.0", "printf", "double", "MT" ... and even though I found evidence that the functions (on Windows, which is unimportant) indeed don't support floating points, I could find no other reference to this "hack" or "work-around" (those two are keywords, too, in case you'd like to try to verify my research).

### DIFF
--- a/crypto/openssl/crypto/cryptlib.c
+++ b/crypto/openssl/crypto/cryptlib.c
@@ -117,10 +117,6 @@
 #include "cryptlib.h"
 #include <openssl/safestack.h>
 
-#if defined(OPENSSL_SYS_WIN32) || defined(OPENSSL_SYS_WIN16)
-static double SSLeay_MSVC5_hack = 0.0; /* and for VC1.5 */
-#endif
-
 DECLARE_STACK_OF(CRYPTO_dynlock)
 
 /* real #defines in crypto.h, keep these upto date */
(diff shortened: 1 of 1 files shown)


asomers:
I don't disagree with you, but FreeBSD is the wrong place to make this change.  FreeBSD is a downstream consumer of OpenSSL.  We regularly merge the latest version.  Cosmetic changes like this one only serve to make the merge process more difficult.  Instead, the only non-merge changes that we should be making to OpenSSL are things that are purely FreeBSD-specific, like its build system.

Plus, it looks like upstream already independently made the same change:

https://github.com/openssl/openssl/commit/156561b0ade98b22df0e3ebc63682e54129c2cb4#diff-aa4b8bc0c4623928e19882b06d2825db

Sebbyastian:
> We regularly merge the latest version.

The upstream change you quoted is over two years old... How "regularly"?

Please pull the most recent soon. I'd like to treat my analysis of the source for FreeBSD more seriously, but this is getting in the way...

asomers:
Upstream maintains a master branch and a few stable branches.  FreeBSD head is currently tracking OpenSSL's 1.0.2 branch, and the change in question hasn't yet been merged to that branch.

https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/crypto/cryptlib.c#L120

It *has* been merged to the 1.1.0 branch.  Do you think we should upgrade to 1.1.0?  Do you want to be involved?  If so, contact jkim@FreeBSD.org .

Sebbyastian:
Well, 'tis the season, so if you'd like to make that choice...

I'm happy to commit the change upstream, or... not sure how much work would need to be done to use 1.1.0, but... again, 'tis the season...

If you can't decide, use a random number generator for me. 0 = don't upgrade, 1 = upgrade.

PR #135, opened 2018-03-28
Add PNP info to puc, bce

lakhanshiva:
I have added PNP info to puc pci driver. My mentor is Warner Losh (@bsdimp ). Kindly review.


cemeyer:
Commit message on b153f00897d71bedec91f1ac5356af970dfde919 is missing a space between 'Add' and 'PNP'.  Please try to keep commit messages tidy.  It is common to refer to drivers by their manual page section (4 for drivers), so for example 'puc(4)' and 'bce(4)'.

sys/dev/bce/if_bce.c
```diff
@@ -529,6 +529,7 @@ MODULE_DEPEND(bce, miibus, 1, 1, 1);
 
 DRIVER_MODULE(bce, pci, bce_driver, bce_devclass, NULL, NULL);
 DRIVER_MODULE(miibus, bce, miibus_driver, miibus_devclass, NULL, NULL);
+MODULE_PNP_INFO("U16:vendor;U16:device;D:#", pci, bce, bce_devs, sizeof(bce_devs[0]), nitems(bce_devs) - 1);
```
cemeyer: The descriptor string is wrong.  `bce_type` is:

```
struct bce_type {
        u_int16_t bce_vid;
        u_int16_t bce_did;
        u_int16_t bce_svid;
        u_int16_t bce_sdid;
        const char *bce_name;
};
```

So you've got vendor and device correct, but the next part should be "V16:subvendor;V16:subdevice;D:#".

Note that the subvendor/subdevice matching is `V` rather than `U` — PCI_ANY_ID (0xffff) is used for some entries and means, match any value in this field.  That's what the 'V' character signifies as well — `0xffff` matches any.

The bus, module, table entry size and count looks correct.

sys/dev/puc/puc_pci.c
```diff
@@ -198,3 +198,5 @@ static driver_t puc_pci_driver = {
 };
 
 DRIVER_MODULE(puc, pci, puc_pci_driver, puc_devclass, 0, 0);
+MODULE_PNP_INFO("U16:vendor;U16:device;D:#", pci, puc, puc_pci_devices,
```
cemeyer: This descriptor string is wrong in exactly the same way as the one for `bce(4)`.  You need to actually look at the struct definition for the devices array to determine the correct format string.

For this driver, the device table data itself will need some modification because the table assumes that a subvendor ANY match means the subdevice id is ignored, and so 0 is often used for table subdevice id when subvendor is ANY.  I think many subdevice id zeros in this driver need to be converted to 0xffff.

PR #136, opened 2018-03-29
Adding PNP info to puc, bce

lakhanshiva:
@bsdimp , @cemeyer

### DIFF
--- a/usr.bin/grep/zgrep.sh
+++ b/usr.bin/grep/zgrep.sh
@@ -106,6 +106,10 @@ do
 	    silent=1
 	    shift
 	    ;;
+	-r|-R)
+	    echo "${prg}: the ${1} flag is not currently supported" >&2
+	    exit 1
+	    ;;
 	-V|--version)
 	    exec ${grep} -V
 	    ;;


sys/dev/puc/puc_pci.c
```diff
@@ -198,3 +198,4 @@ static driver_t puc_pci_driver = {
 };
 
 DRIVER_MODULE(puc, pci, puc_pci_driver, puc_devclass, 0, 0);
+MODULE_PNP_INFO("U16:vendor;U16:device;V16:subvendor;V16:subdevice;D:#", pci, puc, puc_pci_devices, sizeof(puc_pci_devices[0]), nitems(puc_pci_devices) - 1);
```
cemeyer: This change fails to adapt the puc_pci_devices table as described.  Additionally, long lines should be wrapped — see style(9).

bsdimp: I'd  be more inclined to, in this case, leave out the subvendor and subdevice. They aren't needed to differentiate which cards the driver might load for, just to configure parameters of those cards. Therefore, it's easier to just ignore them (use U16:#; for each of the fields to ignore). That also solves the style issue.

sys/dev/bce/if_bce.c
```diff
@@ -529,7 +529,7 @@ MODULE_DEPEND(bce, miibus, 1, 1, 1);
 
 DRIVER_MODULE(bce, pci, bce_driver, bce_devclass, NULL, NULL);
 DRIVER_MODULE(miibus, bce, miibus_driver, miibus_devclass, NULL, NULL);
-
+MODULE_PNP_INFO("U16:vendor;U16:device;V16:subvendor;V16:subdevice;D:#", pci, bce, dce_devs, sizeof(bce_devs[0]), nitems(bce_devs) - 1);
```
bsdimp: I think this line is now also too long.

sys/dev/puc/puc_pci.c
bsdimp: Actually cem, I think we should just ignore them. They don't change which driver gets the card, but do change how things are configured. I'd rather under-specify a smidge here than do the churn on the table.
"U16:vendor;U16:device;U16:#;U16:#;D:#;" seems like a much better choice for the descriptor string in this case.

cemeyer: That's a good point.  Matching on vendor+devid is probably good enough -- even if it is false positive and some other driver is needed for the specific subvendor/subdevice, the harm is low -- one extra driver is loaded.

Should they be anonymous '#'s, or just remove them (and 'D:#') from the format string entirely?

bsdimp: I think they should be anonymous #'s because I like to keep the description if possible. I'm still on the fence on its usefulness, and until I jump one way or the other, I'd like to tag as much as possible.

lakhanshiva: I placed the anonymous #'s for now. For subvendor and subdevice, is it right to use U16 instead of V16 (which cemeyer has suggested) ? 

The puc_pci_devices array is a puc_cfg struct defined as extern const struct puc_cfg puc_pci_devices[];

The puc_cfg struct is
struct puc_cfg {
uint16_t vendor;
uint16_t device;
uint16_t subvendor;
uint16_t subdevice;
const char desc;
int clock;
int8_t ports;
int8_t rid; / Rid of first port /
int8_t d_rid; / Delta rid of next ports /
int8_t d_ofs; / Delta offset of next ports */
puc_config_f *config_function;
};

How can i adapt this PNP_INFO to puc_pci_devices table ?

bsdimp: The PNP_INFO need only describe the first few items in the table. Since we encode a count of items, and their size as well, we need not describe each entry in the table.

cemeyer:
> For subvendor and subdevice, is it right to use U16 instead of V16 (which cemeyer has suggested) ?

Doesn't matter if they will be ignored ('#' name).  They have the same width — U vs V just affects the match behavior.

sys/dev/bce/if_bce.c
```diff
@@ -529,7 +529,8 @@ MODULE_DEPEND(bce, miibus, 1, 1, 1);
 
 DRIVER_MODULE(bce, pci, bce_driver, bce_devclass, NULL, NULL);
 DRIVER_MODULE(miibus, bce, miibus_driver, miibus_devclass, NULL, NULL);
-
+MODULE_PNP_INFO("U16:vendor;U16:device;U16:#;U16:#;D:#", pci, bce,
+		dce_devs, sizeof(bce_devs[0]), nitems(bce_devs) - 1);
```
bsdimp: Not quite right: dce_devs here I think should be bce_devs.

sys/dev/puc/puc_pci.c
```diff
@@ -198,3 +198,6 @@ static driver_t puc_pci_driver = {
 };
 
 DRIVER_MODULE(puc, pci, puc_pci_driver, puc_devclass, 0, 0);
+MODULE_PNP_INFO("U16:vendor;U16:device;U16:#;U16:#;D:#", pci, puc,
+		puc_pci_devices, sizeof(puc_pci_devices[0]),
+		nitems(puc_pci_devices) - 1);
```
bsdimp: This doesn't compile because the size of puc_pci_devices isn't known here.

lakhanshiva: I see, that might be because this table is not present directly in the same file. It is present in another one. What could be done here to correct this ?

bsdimp: I think you'll need to rearrange the code a little so that puc_pci_devices is included in pcu_pci.c somehow. I'd recommend doing it with a #include of the current file that has it since I believe that file is shared with NetBSD.

lakhanshiva: Included pucdata.c which contains puc_pci_devices

bsdimp: missed the reply to this until now...

This looks like it could work, but pucdata.c is already built in the module and specified in the kernel config files (see sys/conf/files for lines containing pucdata.c). So, adjustments in those areas are needed:

diff --git a/sys/conf/files b/sys/conf/files
index d680535ae0b9..b2da16980cee 100644
--- a/sys/conf/files
+++ b/sys/conf/files
@@ -2771,7 +2771,6 @@ dev/puc/puc.c                     optional puc
 dev/puc/puc_cfg.c              optional puc
 dev/puc/puc_pccard.c           optional puc pccard
 dev/puc/puc_pci.c              optional puc pci
-dev/puc/pucdata.c              optional puc pci
 dev/quicc/quicc_core.c         optional quicc
 dev/ral/rt2560.c               optional ral
 dev/ral/rt2661.c               optional ral
diff --git a/sys/modules/puc/Makefile b/sys/modules/puc/Makefile
index 042ae102abef..882e33343067 100644
--- a/sys/modules/puc/Makefile
+++ b/sys/modules/puc/Makefile
@@ -4,7 +4,7 @@
 .PATH: ${SRCTOP}/sys/dev/puc

 KMOD=  puc
-SRCS=  puc.c puc_cfg.c puc_pci.c puc_pccard.c pucdata.c
+SRCS=  puc.c puc_cfg.c puc_pci.c puc_pccard.c
 SRCS+= bus_if.h device_if.h serdev_if.c serdev_if.h \
        card_if.h pci_if.h

If you do any future pull requests, please make sure they compile.

PR #138, opened 2018-05-02
Fix FAT32 scheme-specific type

jlduran:


asomers:
Do you have a reference?

jlduran:
https://github.com/freebsd/freebsd/blob/47aa838516f2954372f66cf72bd285aa4c070ac8/release/tools/arm.subr#L68

asomers:
That's not a reference, just an example use.  A reference would be the gpart code that handles partition types, or better yet the FAT standard (is there one?) or the GPT standard, if relevant.

jlduran:
My bad, `0x0B` is the right file system type:

https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-xp/bb457122(v=technet.10)

jlduran:
Wonder if after https://reviews.freebsd.org/D15266 I should fix and re-submit the PR.

PR #140, opened 2018-05-04
 hda pin fixes for realtek ALC260 and ALC268

kstaring:
This pull request also contains the ALC880 patches.
If there's a more preferable way to submit these patches, please let me know. Also, if the new structures should be changed, please don't hesitate to say so.
There'll be more where this came from, hope to be as efficient as possible.


bsdimp:

PR #142, opened 2018-05-18
Fix the last day of Julian calendar for BG.

terzievk:
Based on multiple sources including the State Gazette the last day of Julian calendar in Bulgaria was 31.03.1916.


bsdimp:
Verified as correct and committed to FreeBSD as r333798.

PR #149, opened 2018-05-18
snd_hda(4) pin fixes realtek ALC286

kstaring:
As with previous patches, earlier branches are also contained in this one. Also, pin patches were ported provided the Linux sources were clear-- entries that looked suspiciously like human error (e.g. double entries, unknown pci vendor ids) weren't ported.


bsdimp:

PR #150, opened 2018-05-18
snd_hda(4) pin fixes realtek ALC293

kstaring:
As with previous patches, earlier branches are also contained in this one. Also, pin patches were ported provided the Linux sources were clear-- entries that looked suspiciously like human error (e.g. double entries, unknown pci vendor ids) weren't ported.


kstaring:
PS: I can also combine all of these pull requests into one big one now, or when Realtek's ready. Continuing by generating a pull request per ALC branch is also fine by me-- whatever's best for a committer.

seanbruno:
Nah, this is kind of awesome in this manner.  I'm testing each one right now for compilation only.

It'll make it a lot easier to bisect in SVN if you continue the process you started here (one patch per device ID).

ykla:
Do you have ALC233?

bsdimp:

PR #154, opened 2018-06-07
Remove duplicated lines in contrib/tzcode/stdtime/private.h

timClicks:
This pull request eliminates some duplication in the source.


bsdimp:
d912068ad826 in main pulls this in, closing out this pull request.
For the record: it looks like this was a mis-merge from the vendor branch somewhere along the way.
The 'somewhere' is so far back I can't find it more specifically than the import of 2009e.
So removing this in -current won't affect future vendor merges.

PR #156, opened 2018-06-10
Allow to run SSHd in Installer

benrubson:
Hi,

This PR should allow to flawlessly run SSHd in Installer, thus allowing remote connections during installation, for convenient manual operations.
It then solves https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=228863

Thank you very much 👍 

Ben

### DIFF
--- a/release/packages/ssh.ucl
+++ b/release/packages/ssh.ucl
@@ -15,4 +15,14 @@ licenses = [ ISCL ]
 desc = <<EOD
 %DESC%
 EOD
+scripts: {
+    post-install = <<EOD
+    sed -i "" 's+keyfile="/etc/ssh+mkdir -p /var/ssh; keyfile="/var/ssh+' ${PKG_ROOTDIR}/etc/rc.d/sshd
+    sed -i "" 's+^[#[:blank:]]*HostKey[[:blank:]]*/etc/ssh+HostKey /var/ssh+' ${PKG_ROOTDIR}/etc/ssh/sshd_config
+    sed -i "" 's+^[#[:blank:]]*PasswordAuthentication[[:blank:]].*+PasswordAuthentication no+' ${PKG_ROOTDIR}/etc/ssh/sshd_config
+    sed -i "" 's+^[#[:blank:]]*ChallengeResponseAuthentication[[:blank:]].*+ChallengeResponseAuthentication no+' ${PKG_ROOTDIR}/etc/ssh/sshd_config
+    sed -i "" 's+^[#[:blank:]]*PermitRootLogin[[:blank:]].*+PermitRootLogin yes+' ${PKG_ROOTDIR}/etc/ssh/sshd_config
+    sed -i "" 's+^[#[:blank:]]*AuthorizedKeysFile[[:blank:]].*+AuthorizedKeysFile /var/ssh/%u/authorized_keys+' ${PKG_ROOTDIR}/etc/ssh/sshd_config
+EOD
+}
 


ngie-eign:
@benrubson: this change seems really reasonable. I have a couple comments:
1. It would be nice if there was an unprivileged user with a default password that could su to root to avoid opening a direct security hole with the root user.
2. The commits should be merged into a single commit. The individuals merging the changes can squash the commits/edit the commit message, but they can also be merged via `git rebase -i`.

asomers:
I disagree with ngie.  Even if the unprivileged user is disabled by default, it's bad form to have a default password.  Also, why use `toor` instead of `root`?

ngie-eign:
> I disagree with ngie. Even if the unprivileged user is disabled by default, it's bad form to have a default password. Also, why use `toor` instead of `root`?

Technically the installer has a default password, "", though. What I'm advocating for is moving away from this default password to a non-nul string.

emaste:
> What I'm advocating for is moving away from a default password of a nul string to a non-nul string.

I would advocate for an empty (i.e., no) password - it makes the context security abundantly clear. IMO it's much more likely that a default non-empty password would remain unchanged.

benrubson:
Thank you all for your feedbacks.
I also agree with the current empty password.
As SSHd is off by default, impact of such a password is quite limited.
User knows what he does launching SSHd, and for the short period of time it will be up.

> Also, why use `toor` instead of `root`?

To "limit" the impact of the empty password.
Think about an online remote server, and SSH robots, which generally use `root` as their first intrusion test user.

We could also use an alternate port to mitigate further...

bsdimp:
On Fri, Dec 21, 2018 at 3:16 AM Ben RUBSON <notifications@github.com> wrote:

> Thank you all for your feedbacks.
> I also agree with the current empty password.
> As SSHd is off by default, impact of such a password is quite limited.
> User knows what he does launching SSHd, and for the short period of time
> it will be up.
>
> Also, why use toor instead of root?
>
> To "limit" the impact of the empty password.
> Think about an online remote server, and SSH robots, which generally use
> root as their first intrusion test user.
>
> We could also use an alternate port to mitigate further...
>

Why not prompt for a password to temporarily use. Then it's safe and you
don't have to mitigate the risks and you don't have to resort to lame
arguments that this is safe because it's short, or using some 'alternative'
that's nothing more than Security Through Obscurity.

Warner

benrubson:
> I would advocate for a 'enable sshd in installer' menu item...
> Why not prompt for a password to temporarily use.

I just wanted to keep this as simple as possible :)

> But I've not looked into how hard that would be to pull off.

And I've no idea (iat least for now) how to implement this :)

benrubson:
Some news regarding this topic ?
Many thanks :+1:

benrubson:
Some news regarding this useful topic please ?
Many thanks :+1:

lwhsu:
If I understand the patch correctly, unconditionally open ssh, even for am unprivileged user, while installation makes me nervous. Having an option to enable it and prompt to enter or generate a random password would be much better.

benrubson:
Thank you @lwhsu for your prompt reply 👍
So, the patch modifies the SSHd configuration so that it can correctly start, changing for example the hosts key path to a writable dir. And so that we can connect to it.

But it does not start SSHd automatically.
You'll still have to launch a shell from the installer, and then start SSHd by yourself.
So only users who know what they do will do this :)

lwhsu:
Oh sorry that I overlooked it. Then I don't think we need `PermitEmptyPasswords` here, users know how to enable sshd should also know how to change password to login, this keeps consistency as the default installation. We can even drop `PermitRootLogin`.

benrubson:
It won't work, as `/etc` is read-only ! You won't then be able to add a user / set a password.
This is also why we need `PermitRootLogin` and `PermitEmptyPasswords` :)
And this is also why we move host-keys path from `/etc` to `/tmp`, so that they can be generated when SSHd starts.

lwhsu:
Understood, I'm fine with an default empty password, but unable to change it doesn't sound a good practice. Is there any way we can solve this?

benrubson:
OK, I've found a better secured way to achieve this.
I've then completely disabled SSH password authentication.
And I enabled SSH keys authentication, moving the `authorized_keys` file location to a writable directory.
Then, thanks to `fetch`, we can easily download a public key and then flawlessly connect through SSH to the installer.
Should suit all comments above 👍

lwhsu:
Thanks, this sounds good to me. I am thinking perhaps we can have something like:
`AuthorizedKeysFile /var/ssh-keys/%u`
so it will not become all uses sharing the same key file.  This might not be necessary in this use case, but it would be better to align with the best practice of the most cases.

I hope others in this thread have more comments and review.

benrubson:
Many thanks for your prompt feedback @lwhsu 👍
I just did the `%u` subdir modification, sounds good to me too.

lwhsu:
Glad to help. It looks good to me. I didn't work and am not familiar in this area so I would like to have more feedback before merging this to the tree. Thanks for your work and please wait a little more.

benrubson:
@ngie-eign @asomers @emaste @bsdimp may I ask you your thoughts regarding this please ?
Goal is to have this in 12.2 installer.
Many thanks 👍

benrubson:
_I just merged the various commits into one commit only to make things proper._

lwhsu:
@benrubson how about asking on freebsd-hackers@ or freebsd-arch@ mailing list?

benrubson:
@lwhsu I just sent a ping to sysinstall@, I'll gently do to the mailing lists you advised if we don't get some news :) Many thx 👍

benrubson:
Done [here](https://lists.freebsd.org/pipermail/freebsd-hackers/2020-March/055711.html) @lwhsu, thank you 👍

benrubson:
@lwhsu, without any additional feedback, could we consider merging this tiny patch ?
Thank you very much 👍

lwhsu:
@benrubson yeah, I'm keeping an eye on this.  I like the idea basically, but unfortunately I'm not quite available right now so I cannot have thorough review and test (and commit) this patch.  Could you wait a little more or check if other developers can handle this?  I'm happy to check this later (before 12.2 release process starts.)

benrubson:
@lwhsu good news you keep this PR in mind, if you plan to handle it before 12.2 release process starts, then it's perfect 👍
Thank you again !

PR #157, opened 2018-06-17
Stable/11

myvker:


asomers:
This change is way too big to review via GitHub (which is just a read-only mirror for FreeBSD).  Please close the PR, and create a review at https://reviews.freebsd.org/ .  When you do, please explain your rationale for selecting these commits to MFC.  They all seem to be unrelated to Docker.

PR #159, opened 2018-07-04
Update kern_exec.c

Wuyang-Chung:
Let 'destp' align on pointer boundary will make copyout run faster.


brooksdavis:
On the copyout I'm most familiar with (MIPS) this might save 30 instructions in the worst case.  I'm skeptical this is worth doing.

Wuyang-Chung:
On i386 and amd64 platform copyout uses string move instructions movsl and movsq respectively to copy data from source to destination. String move instruction run faster when the data are properly aligned.

brooksdavis:
Committed in r336238.  Thanks!

PR #160, opened 2018-07-07
There should be a ';' at the end of struct definition.

Wuyang-Chung:


Wuyang-Chung:
No I didn't have any build error. I just noticed that there is a C grammar error.

brooksdavis:
These documentary declarations have been removed on all current branches.

PR #161, opened 2018-07-12
Add AMD X370 AHCI and XHCI PCI IDs

myfreeweb:
Previously posted as https://reviews.freebsd.org/D15398 — no one answered, let's try here :)


markjdb:
This was committed as r336225.

PR #165, opened 2018-07-19
Remove %%fs and %%gs in the counter add function

Wuyang-Chung:
%%fs (i386) and %%gs (amd64) is confusing in the counter addition function. Since we can get the absolute address of the counter by calling zpcpu_get function, I think it's better to use that function instead of using %%fs and %%gs.

### DIFF
--- a/sys/amd64/include/counter.h
+++ b/sys/amd64/include/counter.h
@@ -83,10 +83,12 @@ counter_u64_zero_inline(counter_u64_t c)
 static inline void
 counter_u64_add(counter_u64_t c, int64_t inc)
 {
+	uint64_t *counter_addr;
 
-	__asm __volatile("addq\t%1,%%gs:(%0)"
+	counter_addr = zpcpu_get(c);
+	__asm __volatile("addq\t%1,(%0)"
 	    :
-	    : "r" ((char *)c - (char *)&__pcpu[0]), "ri" (inc)
+	    : "r" (counter_addr), "ri" (inc)
 	    : "memory", "cc");
 }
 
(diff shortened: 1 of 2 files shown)


markjdb:
The amd64 change, at least, seems wrong to me. The executing thread might be migrated to a different CPU after fetching the counter address and before the increment, in which case it becomes possible for two threads to non-atomically update the same counter. The current implementation is atomic because it consists of a single instruction.

Wuyang-Chung:
Thanks for the comment. So it is the same reason that in i386 implementation it calls zpcpu_get within critical_enter and critical_exit.
I checked the generated assembly output of counter_u64_add in amd64 implementation, it is actually not one assembly instruction. Below is the generated assembly output.
	movslq	%edi, %rdi
	subq	$__pcpu, %rdi
	addq	$1,%gs:(%rdi)

jillest:
The point is that reading the current CPU's PCPU base pointer, reading the data, performing the operation and writing the result are all in one instruction, and cannot be interrupted by a context switch. That some preparation (independent of the current CPU's PCPU base pointer) is required, does not change that.

In the i386 case, the cmpxchg8b instruction (with more preparation) acts as a 64-bit read-modify-write add instruction if the value in %edx:%eax matches, and clears the zero flag and loads %edx:%eax otherwise. If the thread is migrated to a different CPU between `1:` and the cmpxchg8b, and the new CPU's counter value happens to be the same as the old CPU's counter value, the cmpxchg8b will still succeed, and this is perfectly fine.

Wuyang-Chung:
Thank you all for your feedback. Although %%fs and %%gs are confusing, it is a correct implementation.

PR #168, opened 2018-10-05
fix hn_set_hlen so it doesn't return NULL on UDPv6

kevin-j-morse:
As it was, this function discards UDPv6 packets which breaks IPv6 on Hyper-V servers.

### DIFF
--- a/sys/dev/hyperv/netvsc/if_hn.c
+++ b/sys/dev/hyperv/netvsc/if_hn.c
@@ -861,7 +861,7 @@ hn_set_hlen(struct mbuf *m_head)
 
 		PULLUP_HDR(m_head, ehlen + sizeof(*ip6));
 		ip6 = mtodo(m_head, ehlen);
-		if (ip6->ip6_nxt != IPPROTO_TCP) {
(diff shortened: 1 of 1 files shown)


sys/dev/hyperv/netvsc/if_hn.c
```diff
@@ -861,7 +861,7 @@ hn_set_hlen(struct mbuf *m_head)
 
 		PULLUP_HDR(m_head, ehlen + sizeof(*ip6));
 		ip6 = mtodo(m_head, ehlen);
-		if (ip6->ip6_nxt != IPPROTO_TCP) {
+		if ((ip6->ip6_nxt != IPPROTO_TCP) & (ip6->ip6_nxt != IPPROTO_UDP)) {
```
ngie-eign: Did you mean to do a bitwise AND here? Should this the below pattern instead (I removed the superfluous parentheses)?

```if (ip6->ip6_nxt != IPPROTO_TCP && ip6->ip6_nxt != IPPROTO_UDP) {```

ngie-eign:
It turns out, this was resolved in parallel in 9de81ae09f353f9149e29dbfe48005e7346e99e9 (r339585).

PR #169, opened 2018-10-13
Address the fact that errors from `copyout(9)` are not properly percolated up on success

ngie-eign:
This change makes a best effort attempt to ensure that the error returned from `fo_sendfile`
is preserved for diagnostic reasons.

Bug:	232210
Submitted by:	ngie

### DIFF
--- a/sys/kern/kern_sendfile.c
+++ b/sys/kern/kern_sendfile.c
@@ -951,7 +951,7 @@ sendfile(struct thread *td, struct sendfile_args *uap, int compat)
 	struct uio *hdr_uio, *trl_uio;
 	struct file *fp;
 	off_t sbytes;
-	int error;
+	int copyout_error, error;
 
 	/*
 	 * File offset must be positive.  If it goes beyond EOF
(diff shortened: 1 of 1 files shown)


ngie-eign:
Hi @glebius ! I was wondering if you might be able to review this change.

sys/kern/kern_sendfile.c
```diff
@@ -1008,7 +1008,7 @@ sendfile(struct thread *td, struct sendfile_args *uap, int compat)
 	fdrop(fp, td);
 
 	if (uap->sbytes != NULL)
-		copyout(&sbytes, uap->sbytes, sizeof(off_t));
+		error = copyout(&sbytes, uap->sbytes, sizeof(sbytes));
```
bsdimp: This change is fine, but a nop because sbytes is off_t

markjdb: How can that be? An error from the copyout() will now be returned when it wasn't before.

That said, the current behaviour doesn't obviously seem like an oversight: if error == 0 and the copyout() fails, the system call still performed the requested operation. sbytes is purely an output parameter. I think the change is reasonable though.

ngie-eign: > @markjdb: How can that be? An error from the copyout() will now be returned when it wasn't before.

I think @bsdimp was referring to the `sizeof(off_t)` -> `sizeof(bytes)` portion of the change.

The reason why I proposed that change is that it feels really wrong doing a sizeof of a type, which can change for `sbytes` in the future, and the copyout will break.

In terms of the issue, here's a catch with this diff that I'm not comfortable with and I'm going back and forth with @glebius on: if the operation before the `copyout` fails, but the copyout succeeds, the copyout succeeding will mask the earlier failure.

I think it's better to expose the `copyout` error as long as it doesn't mask errors outside of EAGAIN/EINTR.

ngie-eign:
@bsdimp: could you please review this PR?

PR #240, opened 2018-11-30
Clean up the tests/sys/audit integration work

ngie-eign:
There were some issues with the integration that became apparent on my system as I have MK_ACCT set to no in src.conf(5).

This pull request fixes the integration work, in addition to the fact that similar issues existed with MK_AUDIT == no, but in a more fundamental manner.

Submitted by: ngie

### DIFF
--- a/tests/sys/audit/Makefile
+++ b/tests/sys/audit/Makefile
@@ -49,6 +49,7 @@ SRCS.miscellaneous+=		utils.c
 TEST_METADATA+= timeout="30"
 TEST_METADATA+= required_user="root"
 TEST_METADATA+= is_exclusive="true"
+TEST_METADATA+=	required_files="/etc/rc.d/auditd"
 
 WARNS?=	6
 
--- a/tests/sys/Makefile
+++ b/tests/sys/Makefile
@@ -6,7 +6,7 @@ TESTSDIR=		${TESTSBASE}/sys
 
 TESTS_SUBDIRS+=		acl
 TESTS_SUBDIRS+=		aio
-TESTS_SUBDIRS+=		audit
+TESTS_SUBDIRS+=		${_audit}
 TESTS_SUBDIRS+=		auditpipe
 TESTS_SUBDIRS+=		capsicum
 TESTS_SUBDIRS+=		${_cddl}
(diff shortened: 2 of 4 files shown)


ngie-eign:
CC: @asomers, @emaste: could you please review/merge the PR?

asomers:
Is `MK_ACCT` really the right knob?  Shouldn't it be `MK_AUDIT`?

tests/sys/audit/Makefile
```diff
@@ -49,6 +49,7 @@ SRCS.miscellaneous+=		utils.c
 TEST_METADATA+= timeout="30"
 TEST_METADATA+= required_user="root"
 TEST_METADATA+= is_exclusive="true"
+TEST_METADATA+=	required_files="/etc/rc.d/accounting /etc/rc.d/auditd"
```
asomers: This metadata should be set for the `acct_success` testcase only, not for the entire test program.

ngie-eign:
> Is `MK_ACCT` really the right knob? Shouldn't it be `MK_AUDIT`?

`MK_ACCT` is what controls `/etc/rc.d/accounting` being installed. From `libexec/rc/rc.d/Makefile`:
```
136 .if ${MK_ACCT} != "no"
137 CONFGROUPS+=    ACCT
138 ACCT+=          accounting
139 ACCTPACKAGE=    acct
140 .endif
```
There isn't a separate knob for auditd (but there should be -- you're right!).

asomers:
Uh, `WITHOUT_AUDIT` is documented in `src.conf(5)`.  Am I missing something?

ngie-eign:
@asomers: `MK_AUDIT` isn't conditionally installing `/etc/rc.d/audit*` is what I was implying.

asomers:
Ahh, I see.  That would be a bug in libexec/rc/rc.d/Makefile.  We should fix that too.

tests/sys/audit/Makefile
ngie-eign: Uh, actually, pretty much everything calls `setup(..)`, which prods the accounting/auditd services, so I'm 99% sure this is valid as-is :(...

ngie-eign:
@asomers: I applied your comments. Please review/merge https://github.com/freebsd/freebsd/pull/242 first -- thanks!

tests/sys/audit/Makefile
asomers: `setup` starts `auditd`, but it doesn't have anything to do with `accounting`.

ngie-eign: Ah... ok. I see what you meant! I'll work on fixing that.

ngie-eign:
@asomers: can this commit be merged?

asomers:
Submitted in 3 revisions as r342154, r342172, and r342173.

PR #243, opened 2018-12-02
Add [initial] functional tests for sendfile(2) as lib/libc/sys/sendfile

ngie-eign:
Add [initial] functional tests for sendfile(2) as lib/libc/sys/sendfile
  
These testcases exercise a number of functional requirements for sendfile(2).

The testcases use IPv4 and IPv6 domain sockets with TCP, and were confirmed
functional on UFS and ZFS. UDP address family sockets cannot be used per the
sendfile(2) contract, thus using UDP sockets is outside the scope of
testing the syscall in positive cases. As seen in 
`:s_negative_udp_socket_test`, UDP is used to test the sendfile(2) contract
to ensure that EINVAL is returned by sendfile(2).

The testcases added explicitly avoid testing out `SF_SYNC` due to the
complexity of verifying that support. However, this is a good next logical
item to verify.

The `hdtr_positive*` testcases work to a certain degree (the header
testcases pass), but the trailer testcases do not work (it is an expected
failure). In particular, the value received by the mock server doesn't match
the expected value, and instead looks something like the following (using
python array notation):

`trailer[:]message[1:]`

instead of:

`message[:]trailer[:]`

This makes me think there's a buffer overrun issue or problem with the
offset somewhere in the sendfile(2) system call, but I need to do some
other testing first to verify that the code is indeed sane, and my
assumptions/code isn't buggy.

TODO items:
1. Fix the header/trailer testcases.
2. Implement proper linear probing for available ports when setting up sockets.
3. Handle short recv(2)'s in `server_cat(..)`.
4. Add `SF_SYNC` support.
5. Add some more negative tests outside the scope of the functional contract.

Submitted by:   ngie
Sponsored by:   Netflix, Inc

### DIFF
--- a/lib/libc/tests/sys/Makefile
+++ b/lib/libc/tests/sys/Makefile
@@ -8,6 +8,7 @@ PACKAGE=			tests
 ATF_TESTS_C+=			brk_test
 .endif
 ATF_TESTS_C+=			queue_test
+ATF_TESTS_C+=			sendfile_test
 
 # TODO: clone, lwp_create, lwp_ctl, posix_fadvise, recvmmsg,
 # swapcontext
--- a/lib/libc/tests/sys/sendfile_test.c
+++ b/lib/libc/tests/sys/sendfile_test.c
@@ -1,0 +1,1146 @@
+/*-
+ * Copyright (c) 2018 Enji Cooper.
+ * All rights reserved.
+ *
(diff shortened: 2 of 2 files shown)


ngie-eign:
@bsdimp, @glebius: these are the test cases I was working on for the Netflix technical exercise.

This requires https://github.com/freebsd/freebsd/pull/169 to be merged first.

PR #385, opened 2019-01-30
Repeated excessive logging about unsuccessful port reset during device enumeration

maxsteciuk:
Reduce log level for message about unsuccesful port reset which ends up flooding syslog

The following sysctl configuration also helps to get rid of excessive logging but the downside that certain USB device (Create SoundBlaster) are not even getting recognized by USB hub controller.

> hw.usb.disable_enumeration=1


bsdimp:
I've forward ported this to -current. I've created a review https://reviews.freebsd.org/D30621 to talk about it.
Sorry this has languished for so long.

bsdimp:
@hselasky OK'd it, so this landed as 5c447fe67792

PR #387, opened 2019-02-04
Make the file more consistent

scootergrisen:
MacBook not Macbook or macbook
"no accent keys" consistently instead og "without accent keys"
An attempt was made to sort the lines and keep :en: strings first.


bsdphk:
Already committed by @emaster as 46035750a41891d55fd5a41df65ab84440f6ca91

PR #390, opened 2019-02-13
Adding felinity to shutdown

LadySerenaKitty:
Propose adding "nao", "mew", and "meow" to shutdown command.

### DIFF
--- a/sbin/shutdown/shutdown.c
+++ b/sbin/shutdown/shutdown.c
@@ -436,7 +436,10 @@ getoffset(char *timearg)
 
 	(void)time(&now);
 
-	if (!strcasecmp(timearg, "now")) {		/* now */
+	if (!strcasecmp(timearg, "now")
+	    || !strcasecmp(timearg, "nao")
+	    || !strcasecmp(timearg, "mew")
+	    || !strcasecmp(timearg, "meow")) {		/* now */
 		offset = 0;
 		shuttime = now;
 		return;
--- a/sbin/shutdown/shutdown.c
+++ b/sbin/shutdown/shutdown.c
@@ -436,10 +436,10 @@ getoffset(char *timearg)
 
 	(void)time(&now);
 
-	if (!strcasecmp(timearg, "now")
-	    || !strcasecmp(timearg, "nao")
-	    || !strcasecmp(timearg, "mew")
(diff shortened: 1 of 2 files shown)


sbin/shutdown/shutdown.c
```diff
@@ -436,7 +436,10 @@ getoffset(char *timearg)
 
 	(void)time(&now);
 
-	if (!strcasecmp(timearg, "now")) {		/* now */
+	if (!strcasecmp(timearg, "now")
+	    || !strcasecmp(timearg, "nao")
```
bdrewery: From _style(9)_:
```
     Indentation is an 8 character tab.  Second level indents are four spaces.  If you have to wrap a long statement, put the operator at the end of the line.

             while (cnt < 20 && this_variable_name_is_too_long &&
                 ep != NULL)
```

So the `||` should all be at the end of the previous lines.

caladri: For that matter, probably a good time to move from !strcasecmp(...) to strcasecmp(...) == 0, since the result of str*cmp(3) is a sorting order, not a boolean.

I would suggest making this locale-sensitive, too, so that e.g. French locales can use miaou.

LadySerenaKitty: @bdrewery Simple fix!  Want me to update it? 😸

@caladri This code change is simple. 😸 I'm not sure how to make it locale-sensitive, tho. 😿

caladri: Search around for the implementation of YESEXPR, and add a NOWEXPR and a MEOWEXPR and treat NOWEXPR or MEOWEXPR as adequate impetus to shut down now.

I'd also like to see "shutdown whenever" use a random time period.

LadySerenaKitty: I can do that!  As for `shutdown whenever`, write the modification and submit a PR.

LadySerenaKitty: * Updated code to meet the requirements of `style(9)`.
* Looked into L10N, discovered I'm not qualified to work on this.
* Updated `man(8)` page to reflect the new keywords.
* `shutdown whenever` is outside the scope of this PR

HBelusca:
What's the purpose?

zmarffy:
Need explanations.

PR #394, opened 2019-04-06
Fix memory leak in pw.

xaxo05:
Fix memory leak in pw due to improper use of free() to release the memory used for the password and group structures when invoked with the -V argument. The following leaks memory "valgrind pw -V /etc usershow root", however this does not "valgrind pw /etc usershow root".

### DIFF
--- a/usr.sbin/pw/pw_vpw.c
+++ b/usr.sbin/pw/pw_vpw.c
@@ -47,6 +47,25 @@ static FILE * pwd_fp = NULL;
 static int pwd_scanflag;
 static const char *pwd_filename;
 
+static void
+free_gr(struct group *gr)
+{
+	free(gr->gr_name);
+	free(gr->gr_passwd);
+	free(gr->gr_mem);
+}
+
+static void
+free_pw(struct passwd *pw)
+{
+	free(pw->pw_name);
+	free(pw->pw_passwd);
+	free(pw->pw_class);
+	free(pw->pw_gecos);
+	free(pw->pw_dir);
+	free(pw->pw_shell);
+}
+
 void
 vendpwent(void)
 {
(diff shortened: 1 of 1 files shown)


allanjude:
This is a tricky bit of code. But in libutil, the pw and gr structs are allocated in a special way. Instead of just allocating the struct, a single allocation for the struct, plus enough storage for the known contents of each member, is allocated here:
https://github.com/freebsd/freebsd/blob/c125d4a078b7a032c543d05536e21ae7b33c4c75/lib/libutil/pw_util.c#L601

So, a single free can actually be used. It really does seem like there should be some annotation or something to make it obvious, since the original allocation is hidden in pw_dup() called by pw_scan(), and is an over-allocation.

xaxo05:
I later tried to see if what I wrote worked and it didn't, I got some invalid frees and ended up exactly there. I should have compiled before submitting the pull request, though. That is why I closed the pull request. However, it still leaks, and if you try "pw -V /etc showuser -a" that is a massive leak. My thought was adding a global flag variable to pw to check whether you executed pw with -V or -R, as files_password() and files_group() are used in that case. Then one needs to free() after each GETPWENT() and GETGRENT() call. Those are macros hiding the dispatch table. I will reopen the pull request when I do that.

xaxo05:
Please, consider the new proposed solution.

lwhsu:
@xaxo05 if it is possible for you, please consider file a bug report in the issue tracker: https://bugs.freebsd.org and put your proposed patch to the code review system: https://reviews.freebsd.org .  Currently we don't have a complete workflow with github so please use those channels to receive wider discussion.  Also, thanks very much for working on this.

xaxo05:
Ok, I will do that. Thank you.

xaxo05:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=237064

PR #396, opened 2019-04-25
Update vm_map.c

Wuyang-Chung:
Should it be <= *addr instead of length?

### DIFF
--- a/sys/vm/vm_map.c
+++ b/sys/vm/vm_map.c
@@ -1881,7 +1881,7 @@ again:
 	} else if ((cow & MAP_REMAP) != 0) {
 		if (*addr < vm_map_min(map) ||
 		    *addr + length > vm_map_max(map) ||
-		    *addr + length <= length) {
+		    *addr + length <= *addr) {
 			rv = KERN_INVALID_ADDRESS;
 			goto done;
 		}


markjdb:
I believe the purpose of that condition is to check for overflow. Is there a case where testing <= length is not sufficient?

Wuyang-Chung:
I think that code is used to check illegal addresses. That check exists in several places in the code. In other places, the check condition is
    if (start < vm_map_min(map) || end > vm_map_max(map) || start > end)
In this case, start is *addr, end is *addr+length, so the last check condition should be *addr > *addr + length.

You can see the illegal address check in the beginning of vm_map_insert and the macro VM_MAP_RANGE_CHECK.

Wuyang-Chung:
The last check " *addr + length <= length" should be a wraparound check. And a wraparound check should be " *addr + length <= *addr".

markjdb:
Consider mathematical integers a, b >= 0. Suppose a < n and b < n with a+b >= n, i.e., there is some overflow mod n. (a+b)%n is equal to b-(n-a)=a+b-n. a-n < 0 so a+b-n < b. Similarly, a+b-n < a.

*addr and len are unsigned. So, I believe the checks *addr + len <= len and *addr + len <= *addr are equivalent for detecting overflow. Is there something I'm missing?

Note, the current code does not catch the (non-overflow) case where length == 0. But this is handled later in the function.

Wuyang-Chung:
You are right. I didn't think it thoroughly. The checks are equivalent. Thanks for the explanation.

PR #397, opened 2019-05-06
Remove an unnecessary comparison operation when nprocs_new < maxproc - 10

Wuyang-Chung:
In the original code even if nprocs_new < maxproc -10, the program will still want to test if nprocs_new >= maxproc. That test is not necessary.
With the new code the unnecessary test will be eliminated and the generated assembly code is smaller.

### DIFF
--- a/sys/kern/kern_fork.c
+++ b/sys/kern/kern_fork.c
@@ -883,7 +883,7 @@ fork1(struct thread *td, struct fork_req *fr)
 	 * processes; don't let root exceed the limit.
 	 */
 	nprocs_new = atomic_fetchadd_int(&nprocs, 1) + 1;
-	if (nprocs_new >= maxproc - 10)
+	if (nprocs_new >= maxproc - 10) {
 		if (priv_check_cred(td->td_ucred, PRIV_MAXPROC) != 0 ||
 		    nprocs_new >= maxproc) {
 			error = EAGAIN;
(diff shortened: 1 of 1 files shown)


mjguzik:
It does not do that. The C programming language short-circuits expressions of the sort, i.e. nprocs_new > maxproc - 10 evaluating to false causes the rest to not be evaluated. Most notably the call to priv_check is not being executed, nor is the later comparison. iow, the proposed change is a no-op.

markjdb:
I don't think the comment about short-circuiting applies here: the expression is of the form ((A && B) || C).

Looking at the fork1() disassembly on my system, we do indeed test nprocs_new >= maxproc if nprocs_new < maxproc - 10.

sys/kern/kern_fork.c
```diff
@@ -883,19 +883,19 @@ fork1(struct thread *td, struct fork_req *fr)
...
-	if ((nprocs_new >= maxproc - 10 &&
-	    priv_check_cred(td->td_ucred, PRIV_MAXPROC) != 0) ||
-	    nprocs_new >= maxproc) {
-		error = EAGAIN;
-		sx_xlock(&allproc_lock);
-		if (ppsratecheck(&lastfail, &curfail, 1)) {
-			printf("maxproc limit exceeded by uid %u (pid %d); "
-			    "see tuning(7) and login.conf(5)\n",
-			    td->td_ucred->cr_ruid, p1->p_pid);
+	if (nprocs_new >= maxproc - 10)
```
allanjude: I would prefer to see explicit curly braces {} here

Wuyang-Chung:
I have checked the assembly output from both gcc and clang with -O2 and both will check nprocs_new >= maxproc again even if nprocs_new < maxproc - 10.

markjdb:
Committed in r347227.

PR #398, opened 2019-05-10
Remove obsolete 10 and 10/100 Ethernet NIC drivers per fcp-0101.

brooksdavis:
The following drivers will be removed: 
ae, bm, cs, de, ed, ep, ex, fe, pcn, sf, sn, tl, tx, txp, vx, wb, xe

Each will be removed in a seperate commit with pre- and post- commits 
bounding the removals.  A bump of __FreeBSD_version is included in the 
final commit.

**This pull request exists to complement a phabricator review by exposing the planned individual commits.**


brooksdavis:
Review at https://reviews.freebsd.org/D20230

brooksdavis:
This has been committed to SVN.

PR #401, opened 2019-06-01
fix i386 msun cexp tests

moritzbuhl:
adding another volatile helped
only tested on openbsd (with freebsds libm tho)


moritzbuhl:
Also I just copied this change over from my editor into the GitHub thing. I can't promise that it will compile.

DimitryAndric:
This looks OK to me, and it appears to work on 13-CURRENT with clang 8.  Though I wonder why the same trick isn't needed to for the `test_tol()` macros?

DimitryAndric:
Committed in d36465065, with some minor changes.

PR #403, opened 2019-06-08
bsnmpd(1) fixes made during the Vienna Hackathon 2019

shteryana:
(Ob)use GitHub to make the work public before it gets committed to SVN


shteryana:
Patches have been committed to head & stable/12

PR #407, opened 2019-07-09
getgroup and setgroup with negative int

moritzbuhl:
I was running this test on OpenBSD recently: https://github.com/NetBSD/src/blob/trunk/tests/lib/libc/sys/t_getgroups.c
getgroups/setgroups want an int and therefore casting it to u_int resulted in
`getgroups(-1, ...)` will not `return -1` and set `errno = EINVAL` but instead just work.
I haven't tested this on FreeBSD but the code already was in 4.4BSD to my knowledge.

http://cvsweb.netbsd.org/bsdweb.cgi/src/sys/kern/kern_prot.c (8 years ago)

This change is not tested. I suggest try this:

```
#include <sys/syslimits.h>

#include <stdio.h>
#include <strings.h>
#include <unistd.h>


int
main(void)
{
        gid_t gidset[NGROUPS_MAX];
        bzero(gidset, NGROUPS_MAX * sizeof(gid_t));
        printf("%d %d %d\n", gidset[0], gidset[1], gidset[2]);
        int a = getgroups(-1, gidset);
        printf("%d\n", a);
        printf("%d %d %d\n", gidset[0], gidset[1], gidset[2]);
        return 0;
}
```

### DIFF
--- a/sys/kern/kern_prot.c
+++ b/sys/kern/kern_prot.c
@@ -278,7 +278,7 @@ sys_getegid(struct thread *td, struct getegid_args *uap)
 
 #ifndef _SYS_SYSPROTO_H_
 struct getgroups_args {
-	u_int	gidsetsize;
+	int	gidsetsize;
 	gid_t	*gidset;
 };
 #endif
@@ -286,7 +286,7 @@ int
 sys_getgroups(struct thread *td, struct getgroups_args *uap)
 {
 	struct ucred *cred;
-	u_int ngrp;
+	int ngrp;
 	int error;
 
 	cred = td->td_ucred;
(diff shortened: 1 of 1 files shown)


moritzbuhl:
Also this change was edited in the github web editor.

markjdb:
This looks reasonable to me.

bsdimp:
OK. There's a few problems with this. Normally, I'd bounce it back, but since it's been so long since submitted, I'm correcting these issues. 
First, it doesn't update syscalls.master and regenerate sys/sysproto.h, so it's effectively a nop.
Second, it doesn't remove the annotation from contrib/netbsd-tests/lib/libc/sys/t_getgroups.c that it's broken
Third, it didn't mention bug 189941 in the commit message.

I've opened https://reviews.freebsd.org/D30617 for review.

PR #408, opened 2019-07-24
off is zero. It can be removed.

Wuyang-Chung:
off = trunc_page(offset + filsz) - trunc_page(offset + filsz) and it is zero.

### DIFF
--- a/sys/kern/imgact_elf.c
+++ b/sys/kern/imgact_elf.c
@@ -610,8 +610,7 @@ __elfN(load_section)(struct image_params *imgp, vm_ooffset_t offset,
 	 * segment in the file is extended to provide bss.  It's a neat idea
 	 * to try and save a page, but it's a pain in the behind to implement.
 	 */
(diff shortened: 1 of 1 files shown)


markjdb:
I committed the removal of `off`, thanks. It's not obvious to me that the compiler won't perform the micro-optimization in the other commit; did you verify that we get some change in the compiled kernel?

Wuyang-Chung:
I write a small test program and compare the assembly output for both statements. I didn't expect that the compiler can optimize the code as efficient as the original one. The compiler is really doing a good job.

emaste:
> The compiler is really doing a good job.

Indeed, the compiler can do some quite interesting optimizations.

> The proposed one is more easy for me to understand the code. :)

This is key, we should aim for readability/understandability - especially when the compiler generates the same code either way. I'm not sure which I find more understandable though; I read the current code as "excluding the already copied portion" and your proposed change as "the partial page content"

Wuyang-Chung:
On a second thought the original one is better. :)

PR #410, opened 2019-08-22
Update vm_glue.c

Wuyang-Chung:
Incorrect function name in printf and panic format string. Can use `__func__` to avoid this issue in the future.


allanjude:
I have staged this for commit in our review system: https://reviews.freebsd.org/D21361

PR #414, opened 2019-11-12
Don't left-shift negative numbers in certain circumstances in the various fdlibm hypot() functions

jswalden:
Various paths through `hypot(x, y)` will multiply `x` and `y` by a power of two, perform the calculation in a range where IEEE-754 provides greater precision, then undo the multiplication to determine the true result.  Undoing that multiplication is implemented as `t1*w`, where `t1=2**k`.

`2**k` is often computed by taking the high word of `1.0`, then adding `k<<20` (for doubles or long doubles) or `k<<23` (for floats) to it, then overwriting that high word.  But when `k` is negative this left-shifts a negative value -- and that's undefined behavior in many editions of C and C++.

This patch should fix all `hypot` implementations to compute `2**k` without triggering this particular bit of undefined behavior.  I've only _very lightly_ tested out the `double` change, in SpiderMonkey's JavaScript engine; the other functions' changes have more or less only been eyeballed.  I am also unsure whether the particular ways of spelling things out here are stylistically preferred; the fdlibm code has all sorts of magic numbers in it, and few idioms for hiding them away behind structured access functions, so the desired idioms are not entirely obvious.  Careful examination appreciated!

### DIFF
--- a/lib/msun/src/e_hypot.c
+++ b/lib/msun/src/e_hypot.c
@@ -118,10 +118,9 @@ __ieee754_hypot(double x, double y)
 	    w  = sqrt(t1*y1-(w*(-w)-(t1*y2+t2*b)));
 	}
 	if(k!=0) {
-	    u_int32_t high;
-	    t1 = 1.0;
-	    GET_HIGH_WORD(high,t1);
-	    SET_HIGH_WORD(t1,high+(k<<20));
+	    int32_t exponent = 1023 + k;
+	    t1 = 0.0;
+	    SET_HIGH_WORD(t1,exponent<<20);
 	    return t1*w;
 	} else return w;
 }
--- a/lib/msun/src/e_hypotf.c
+++ b/lib/msun/src/e_hypotf.c
@@ -77,7 +77,7 @@ __ieee754_hypotf(float x, float y)
 	    w  = __ieee754_sqrtf(t1*y1-(w*(-w)-(t1*y2+t2*b)));
 	}
 	if(k!=0) {
-	    SET_FLOAT_WORD(t1,0x3f800000+(k<<23));
+	    SET_FLOAT_WORD(t1,(127+k)<<23);
 	    return t1*w;
 	} else return w;
 }
--- a/lib/msun/src/e_hypotl.c
+++ b/lib/msun/src/e_hypotl.c
@@ -115,10 +115,8 @@ hypotl(long double x, long double y)
 	    w  = sqrtl(t1*y1-(w*(-w)-(t1*y2+t2*b)));
 	}
 	if(k!=0) {
-	    u_int32_t high;
-	    t1 = 1.0;
-	    GET_HIGH_WORD(high,t1);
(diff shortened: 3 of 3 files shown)


lwhsu:
Thanks very much for submitting this! I'm sorry that we currently don't have a standard workflow for github. Can I suggest you that filing a ticket and/or review in our issue tracker and code review system?
https://bugs.freebsd.org/
https://reviews.freebsd.org/

Also, for the changes in this field, I would recommend to discuss on the freebsd-numerics mailing list:
https://lists.freebsd.org/mailman/listinfo/freebsd-numerics

jswalden:
Probably, sure -- although this hasn't been necessary for similar issues I've fixed in the past, as https://github.com/freebsd/freebsd/pull/411 and https://github.com/freebsd/freebsd/pull/412 and https://github.com/freebsd/freebsd/pull/130 show.  Has something changed procedurally recently?

lwhsu:
No, there are some developers (including me) watching this repository and trying to help the work from contributors get into the tree. I'm just suggesting the current best route to get most developers' attention and have the work merged. We're working on improving the whole workflow but that may not be available very soon.

lwhsu:
To be clear, nothing has been changed recently, and this PR is absolutely fine. It might just take a bit longer to have the right people notice this and follow up. Again, thanks very much for your work!

jswalden:
Ah, cool.  I'm fine with "a bit longer" here, for this sort of change, for the most part.

But seeing as I've made other changes to this code in the past, so there's non-negligible chance of me having to do so again in the future, I guess I may as well try to get over the phabricator hurdle.  (I don't like it, myself, and I wish Mozilla hadn't foisted it on me.  But so it goes.)  I'll try to see if I can turn that around in the next little bit...

lwhsu:
Thanks for the understanding. We're working on workflow improvement but there is no ETA at the moment. We're searching any possibilities so please let us know if you have any suggestions.  I personally guess we would have some better github integration in the future, but that needs carefully planned.

To working with FreeBSD's review system, hope this could help you: https://wiki.freebsd.org/Phabricator
BTW, I still strongly suggest that posting to -numerics@ mailing list to have more attention from the people in this field: https://lists.freebsd.org/mailman/listinfo/freebsd-numerics

jswalden:
Okay -- phabricator rev [submitted](https://reviews.freebsd.org/D22354), mail [sent](https://lists.freebsd.org/pipermail/freebsd-numerics/2019-November/000922.html) to the mailing list.  Help appreciated directing this to appropriate reviewer(s)!

DimitryAndric:
FWIW I'm fine with reviewing stuff on GitHub, but in particular for libm related things, the audience is mostly on the traditional mailing lists.  Even Phabricator is way too modern and fancy for most. ;)

DimitryAndric:
Thanks, committed in https://svnweb.freebsd.org/changeset/base/355120 (git commit is still being imported)

DimitryAndric:
Fixed by c5111f3a65b2c2755b3332eaaa580171f19803a5.

PR #416, opened 2019-11-14
initial linker support userland-DTrace aarch64

klaus4:
initially solves :
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=239805

furthermore it would be nice if we could work on pid-provider-support for aarch64 :
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=241028#c0


klaus4:
https://reviews.freebsd.org/D22360

klaus4:
https://github.com/freebsd/freebsd/commit/2ade50c4604d177da75b805a24d65bf06af13bf4

PR #417, opened 2019-11-28
Update vm_map.c

Wuyang-Chung:
The valid range of pindex is 0 to (object->size - 1) so the equal sign should be needed here.


markjdb:
I think you are correct, but I believe that the bug is harmless. That is, if pindex == object->size then we will set psize = 0 and the rest of vm_map_pmap_enter() will do nothing.

markjdb:
Committed in r355399, thanks.

PR #419, opened 2019-12-04
Update vm_map.c

Wuyang-Chung:
Should fault_type be initialized here since fault_type will be changed if the entry is wired and the program decided to jump to RetryLookup*? That is if the entry is wired entry, fault_type will be changed. But if the program later decides to jump to RetryLockup*, then the original contents from fault_typea will be lost.


markjdb:
Nice find. I think you are right.

markjdb:
Committed in r355469, thank you.

PR #420, opened 2019-12-19
Update vm_map_submap in vm_map.c

Wuyang-Chung:
vm_map_submap is destined to to fail if vm_map_lookup_entry returns FALSE or entry->end < end.
I think we only need to do vm_map_clip_start and vm_map_clip_end when vm_map_lookup_entry successes and entry->end >= end.


markjdb:
Sorry, I didn't see this one. I think the change is fine, though FreeBSD does not really use submaps for anything.

markjdb:
I created https://reviews.freebsd.org/D23299 based on this patch.

PR #422, opened 2020-01-22
Remove redundant code in __elfN(map_insert)

Wuyang-Chung:
Also
* Do more check on the executable file.
* Add a new error return constant ESUCCESS to make the return code more clear.

### DIFF
--- a/sys/kern/imgact_elf.c
+++ b/sys/kern/imgact_elf.c
@@ -600,7 +600,7 @@ __elfN(load_section)(struct image_params *imgp, vm_ooffset_t offset,
 	}
 
 	/*
-	 * If the code reaches here it means that this is a data section and
+	 * If the program reaches here it means that this is a data section and
 	 * data section always has WRITE permission.
 	 */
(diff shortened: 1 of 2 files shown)


markjdb:
Nice, I guess this code became dead with r344705.

I'm not convinced that the ESUCCESS addition is really warranted: it is non-standard (so it should at least be defined under #ifndef _POSIX_SOURCE), and there is a ton of code in the tree that understands that a return value 0 means "success". Adding ESUCCESS would mean that there's now two ways of indicating success, which is itself confusing.

Wuyang-Chung:
For some functions, the return value of 0 means failure and the return value of 1 or non-zero means success. So when I see a function that returns 0, I can not immediately know whether it is a success or failure. By returning ESUCCESS, I think it will make the code more clear. This is inspired by KERN_SUCCESS. KERN_SUCCESS also has the value of 0. Instead of returning 0, returning KERN_SUCCESS makes the code more easy to understand.

PR #424, opened 2020-02-29
Fix confusing libfuse_max_write default

felixbuenemann:
Remove redundant 0x1000 (4096) - 4096 in the libfuse_max_write.
The + 0x1000 was probably added to represent the fuse message
header size, which is 4 KB, but that is not included in max_write,
explaining why the - 4096 was added later.

### DIFF
--- a/tests/sys/fs/fusefs/utils.cc
+++ b/tests/sys/fs/fusefs/utils.cc
@@ -54,10 +54,9 @@ using namespace testing;
 
 /*
  * The default max_write is set to this formula in libfuse, though
- * individual filesystems can lower it.  The "- 4096" was added in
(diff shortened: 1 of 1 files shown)


felixbuenemann:
Also note that newer Linux kernels support up to 256 pages, but that has to be negotiated using CAP_MAX_PAGES and requires protocol version 7.28 or later, so the FreeBSD port which supports up to 7.23 can't currently take advantage of that.

asomers:
Yeah, it looks confusing.  But that's because it's copied from libfuse, where the definition is indeed confusing.  And since that code is trying to imitate libfuse, it's actually more understandable overall to stick with libfuse's confusing definition.

PR #426, opened 2020-03-08
usr.sbin/bhyve: Add the block_rbd driver for Rados access

wjwithagen:
usr.sbin/bhyve: Add the block_rbd driver for Rados access

### DIFF
--- a/usr.sbin/bhyve/block_if.c
+++ b/usr.sbin/bhyve/block_if.c
@@ -25,827 +25,194 @@
  * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
  * SUCH DAMAGE.
  *
- * $FreeBSD$
+ * $FreeBSD: head/usr.sbin/bhyve/block_if.c 356523 2020-01-08 22:55:22Z vmaffione $
  */
 
 #include <sys/cdefs.h>
-__FBSDID("$FreeBSD$");
+__FBSDID("$FreeBSD: head/usr.sbin/bhyve/block_if.c 356523 2020-01-08 22:55:22Z vmaffione $");
 
 #include <sys/param.h>
-#ifndef WITHOUT_CAPSICUM
-#include <sys/capsicum.h>
-#endif
-#include <sys/queue.h>
-#include <sys/errno.h>
-#include <sys/stat.h>
-#include <sys/ioctl.h>
-#include <sys/disk.h>
-
-#include <assert.h>
-#ifndef WITHOUT_CAPSICUM
-#include <capsicum_helpers.h>
-#endif
-#include <err.h>
-#include <fcntl.h>
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
-#include <pthread.h>
-#include <pthread_np.h>
-#include <signal.h>
-#include <sysexits.h>
+#include <errno.h>
+#include <dlfcn.h>
+#include <regex.h>
 #include <unistd.h>
+#include <stdio.h>
+#include <string.h>
 
-#include <machine/atomic.h>
+#include <sys/linker_set.h>
 
-#include "bhyverun.h"
-#include "debug.h"
-#include "mevent.h"
 #include "block_if.h"
+#include "debug.h"
 
-#define BLOCKIF_SIG	0xb109b109
-
-#define BLOCKIF_NUMTHR	8
-#define BLOCKIF_MAXREQ	(BLOCKIF_RING_MAX + BLOCKIF_NUMTHR)
-
-enum blockop {
-	BOP_READ,
-	BOP_WRITE,
-	BOP_FLUSH,
-	BOP_DELETE
-};
-
-enum blockstat {
-	BST_FREE,
-	BST_BLOCK,
-	BST_PEND,
-	BST_BUSY,
-	BST_DONE
-};
-
-struct blockif_elem {
-	TAILQ_ENTRY(blockif_elem) be_link;
-	struct blockif_req  *be_req;
-	enum blockop	     be_op;
-	enum blockstat	     be_status;
-	pthread_t            be_tid;
-	off_t		     be_block;
-};
-
-struct blockif_ctxt {
-	int			bc_magic;
-	int			bc_fd;
-	int			bc_ischr;
-	int			bc_isgeom;
-	int			bc_candelete;
-	int			bc_rdonly;
-	off_t			bc_size;
-	int			bc_sectsz;
-	int			bc_psectsz;
-	int			bc_psectoff;
-	int			bc_closing;
-	pthread_t		bc_btid[BLOCKIF_NUMTHR];
-	pthread_mutex_t		bc_mtx;
-	pthread_cond_t		bc_cond;
-
-	/* Request elements and free/pending/busy queues */
-	TAILQ_HEAD(, blockif_elem) bc_freeq;       
-	TAILQ_HEAD(, blockif_elem) bc_pendq;
-	TAILQ_HEAD(, blockif_elem) bc_busyq;
-	struct blockif_elem	bc_reqs[BLOCKIF_MAXREQ];
-};
-
-static pthread_once_t blockif_once = PTHREAD_ONCE_INIT;
-
-struct blockif_sig_elem {
-	pthread_mutex_t			bse_mtx;
-	pthread_cond_t			bse_cond;
-	int				bse_pending;
-	struct blockif_sig_elem		*bse_next;
-};
-
-static struct blockif_sig_elem *blockif_bse_head;
-
-static int
-blockif_enqueue(struct blockif_ctxt *bc, struct blockif_req *breq,
-		enum blockop op)
-{
-	struct blockif_elem *be, *tbe;
-	off_t off;
-	int i;
-
-	be = TAILQ_FIRST(&bc->bc_freeq);
-	assert(be != NULL);
-	assert(be->be_status == BST_FREE);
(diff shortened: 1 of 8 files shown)


usr.sbin/bhyve/block_if.h
```diff
@@ -75,4 +138,55 @@ int	blockif_delete(struct blockif_ctxt *bc, struct blockif_req *breq);
...
+
+        int	(*bb_delete)(blockif_ctxt_t *bc, blockif_req_t *breq);
+
+        int	(*bb_cancel)(blockif_ctxt_t *bc, blockif_req_t *breq);
+
+        int	(*bb_close)(blockif_ctxt_t *bc);
+
+	/* Room for backend-specific data. */
+	void *bb_opaque;
+};
```
trociny: If we are going to have backends outside of the base code, I think we should have a header file with this structure defined somewhere in /usr/include.

usr.sbin/bhyve/block_if.h
```diff
@@ -75,4 +138,55 @@ int	blockif_delete(struct blockif_ctxt *bc, struct blockif_req *breq);
...
+
+/* Cheating for legacy code were we need the blocklocal_backend block */
+block_backend_t blocklocal_backend;
```
trociny: I can't understand this comment.

usr.sbin/bhyve/block_if.c
```diff
@@ -25,827 +25,155 @@
...
-		else if (sscanf(cp, "sectorsize=%d/%d", &ssopt, &pssopt) == 2)
-			;
-		else if (sscanf(cp, "sectorsize=%d", &ssopt) == 1)
-			pssopt = ssopt;
-		else {
-			EPRINTLN("Invalid device option \"%s\"", cp);
-			goto err;
+	block_backend_t **bbe = NULL;
+	blockif_ctxt_t *ret = NULL;
+        char *optrest; 		/* if found, optstr without scheme: */
```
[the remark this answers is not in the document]
wjwithagen: Yeah, all indent/space/tab regime is a hassle. But I usually fix 'm on the final run.

usr.sbin/bhyve/block_if.c
```diff
@@ -25,827 +25,155 @@
...
+
+	optrest = strchr(optstr, ':');
```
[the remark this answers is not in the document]
wjwithagen: That would be a nice assumption to add.
But add things like: `-_`?? Or only `[a-zA-Z0-9]` ??

usr.sbin/bhyve/block_if.c
```diff
@@ -25,827 +25,155 @@
...
-				goto err;
+		/*
+		 * Local access has a pattern like:
+		 *   3:0,virtio-blk,file/somewhere/guest.img
+		 *   3:0,virtio-blk,/dev/xxxx
+                 * Or new style:
+		 *   3:0,virtio-blk,file:file/somewhere/guest.img
+		 *   3:0,virtio-blk,file:/dev/xxxxx
+		 * Requesting a RADOS block device should look something like:
+		 *   3:0,virtio-blk,rbd:pool/image,option_1,option_2=val,....
```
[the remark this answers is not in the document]
wjwithagen: I'll change it to a more generic description of what is matched.

usr.sbin/bhyve/block_if.h
```diff
@@ -57,12 +80,52 @@ struct blockif_req {
...
-struct blockif_ctxt;
-struct blockif_ctxt *blockif_open(const char *optstr, const char *ident);
+typedef struct block_backend block_backend_t;
+struct blockif_ctxt {
+        int                     bc_magic;
+        union {
+                int             bc_fd;
+        	/* For data specific for this instance of the backend*/
+	        void           *bc_desc;
+	} desc;
```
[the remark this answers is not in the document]
wjwithagen: Every slot in in the PCI-table needs a to be connected to a driver....
So IMHO the a slot is connected to backend instance that goes with the data of that PCI slot.
For example 
 - every file: slot needs its own backend struc to register the filedescriptor.
 - every rbd: slot needs its onw backend to point to an rbd descriptor.
Other things that will differ per PCI-slot is the size.

Then that backend also points to the backend driver that it is using, like file: or rbd:

I've made it a pointer, since most of the backend_descriptors will be a allocated structure that is usually more that just an integer.

usr.sbin/bhyve/block_if.h
```diff
@@ -57,12 +80,52 @@ struct blockif_req {
...
+
+        block_backend_t	       *be;
+
+        /* Request elements and free/pending/busy queues */
+        TAILQ_HEAD(, blockif_elem) bc_freeq;
+        TAILQ_HEAD(, blockif_elem) bc_pendq;
+        TAILQ_HEAD(, blockif_elem) bc_busyq;
+        struct blockif_elem  bc_reqs[BLOCKIF_MAXREQ];
+};
+typedef struct blockif_ctxt blockif_ctxt_t;
```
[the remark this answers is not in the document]
wjwithagen: 'mmmm, oke. I guess they have a rationale for that.
personally I find it rather clumsy.
Guess it is time again to read style(9)

usr.sbin/bhyve/block_if.h
wjwithagen: I think that that is going to be a good fix for not having to include everything and the kitchensick in the port.
something like `/usr/include/bhyve/block_if.h` ??

usr.sbin/bhyve/block_if.h
wjwithagen: I'll try to refrase this, but it is a poor attempt not to create a `block_local.h` file with only one thing in it: the address of the jump table.
And that is needed in `block_if.c`, to be able to use it in case of a legacy PCI-slot specification without `file:`
The other way around would be to prefix legacy with `file:`, and run it through the regular backend-search loop. But detecting legacy by scanning the string is rather complex or error prone.

usr.sbin/bhyve/block_if.c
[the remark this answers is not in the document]
trociny: Any way is ok for me.

usr.sbin/bhyve/block_if.h
trociny: yes

usr.sbin/bhyve/block_if.h
trociny: Actually, the "way around" looks fine to me. Something like this:
```
backend_name = "file";
optrest = strchr(optstr, ':');
if (optrest != NULL) {
  *optrest = '\0';
  backend_name = optstr;
  optstr = ++optrest;
}
/* search for backend using backend_name */
```

usr.sbin/bhyve/block_if.h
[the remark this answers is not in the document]
trociny: Ok, I forgot what `bb_opaque` was for. Anyway, I don't think we need a union, just a `void *` pointer, so every driver will cast it to its own struct or whatever it wants.

usr.sbin/bhyve/block_local.c
```diff
@@ -0,0 +1,828 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause-FreeBSD
+ *
+ * Copyright (c) 2013  Peter Grehan <grehan@freebsd.org>
```
asomers: You should update the copyright date and author

usr.sbin/bhyve/block_rbd.c
```diff
@@ -0,0 +1,972 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause-FreeBSD
+ *
+ * Copyright (c) 2013  Willem Jan Withagen <wjw@digiware.nl>
```
asomers: You should update this copyright date, too,

usr.sbin/bhyve/block_rbd.c
```diff
@@ -0,0 +1,972 @@
...
+        return (-1);
+	}
+
+    return (r);
+}
+
+static ssize_t
+block_rbd_delete(block_rbd_inst_t *bd, size_t nbytes, off_t offset)
+{
+	struct block_rbd_inst *ri = bd;
```
asomers: Irregular indentation here and elsewhere in this function.

usr.sbin/bhyve/block_rbd.c
```diff
@@ -0,0 +1,972 @@
...
+		}
+		if (buf == NULL) {
+			if ((len = block_rbd_writev(bc->desc.bc_desc, br->br_iov, br->br_iovcnt,
+				    br->br_offset)) < 0)
+				err = errno;
+			else
+				br->br_resid -= len;
+			break;
+		}
+        /* Have yet to see that buf != NULL for iovcnt > 1 */
```
asomers: More indentation problems.

asomers:
> I think the rbd driver does not need threads. The reason why they were needed for local driver is because it operated with sych read/write/etc functions. In the rbd case we can fully utilize aio functions and just had a queue for requests. But this can be improved later. As a POC I think the rbd driver is in a good shape.

I'm a little confused.  Where does AIO come in?  librados isn't actually providing a block device, so I assume you aren't talking about using POSIX AIO.  Are you talking about a different Ceph AIO API?

bsdimp:
This seems stalled... what's the latest status? It's also conflicting with block_if.c, so at the very least needs a rebase.

wjwithagen:
Hi Warner,

Have not spent time on this for a while, but getting new versions of ceph into Ports. After which I ll rebase this code. Need to do so for my EuroBSD talk

—WjW

bsdimp:
It's been over a year, with no rebase. Closing. Please update and reopen if you find time.

PR #428, opened 2020-04-01
added Novatel Wireless MiFi 8800/8000 support

rootless4real:
added Novatel Wireless MiFi 8800/8000 support.  An entry was needed to recognize the "0xef" presented by the device so it would be flagged and a ueX device would be created.


rootless4real:
Please add this so support for these popular LTE modems (Verizon, Sprint, T-Mobile, AT&T) is available.

lwhsu:
https://bugs.freebsd.org/245152

lwhsu:
Merged in https://svnweb.freebsd.org/changeset/base/365606

bsdimp:
Has been committed to base

PR #431, opened 2020-05-19
Correcting comment about "sched_interact_score".

wiklam:
Correcting description about calculating interactive score when run time is greater than sleep time.

### DIFF
--- a/sys/kern/sched_ule.c
+++ b/sys/kern/sched_ule.c
@@ -1524,9 +1524,9 @@ sched_initticks(void *dummy)
  * When a thread's run time is greater than its sleep time the
  * calculation is:
  *
- *                           scaling factor 
- * interactivity score =  ---------------------    + scaling factor
- *                        run time / sleep time
+ *                                                 scaling factor 
+ * interactivity score = 2 * scaling factor  -  ---------------------
+ *                                              run time / sleep time
  */
 static int
 sched_interact_score(struct thread *td)


lwhsu:
After reading the sched_interact_score() code, I think the original comment is correct. Can you explain the rationale of your patch more?

wiklam:
The comment is describing code in line 1547.
```
	if (ts->ts_runtime > ts->ts_slptime) {
		div = max(1, ts->ts_runtime / SCHED_INTERACT_HALF);
		return (SCHED_INTERACT_HALF +
		    (SCHED_INTERACT_HALF - (ts->ts_slptime / div)));
	}
```
and we return (2 * scaling factor - sleep time/div), where div = run time/scaling factor (we use max to avoid dividing by zero).

Otherwise according to the old comment sched_interact_score() function would be discontinuous according to the ratio between sleep time and run time.
Example:  when sleep time == run time our interactive score is equal scaling factor, but when our run time will become little bigger then sleep time the interactive score would be 2 * scaling factor instead scaling factor + epsilon.

bsdimp:
I think this is a good change.

jrtc27:
This looks right to me. Note that `SCHED_INTERACT_HALF` is the "scaling factor".

jrtc27:
Reviewed By: jrtc27

bsdimp:
Merged to freebsd main at 43521b46fc78.

PR #433, opened 2020-07-30
script(1): Enable proper playback [-p] of curses sessions

8vasu:
The current manpage of script(1) says the following.

> Certain interactive commands, such as vi(1), create garbage in the typescript file.  The script utility works best with commands that do not manipulate the screen.  The results are meant to emulate a hardcopy terminal, not an addressable one.

script(1) is currently able to record [-r] curses sessions perfectly; only the playback functionality [-p] has the above restriction; this patch removes this restriction by appropriately setting terminal attributes.

### DIFF
--- a/usr.bin/script/script.c
+++ b/usr.bin/script/script.c
@@ -428,6 +428,34 @@ consume(FILE *fp, off_t len, char *buf, int reg)
 } while (0/*CONSTCOND*/)
 
 static void
+termset(void)
+{
+	struct termios traw;
+
+	ttyflg = isatty(STDOUT_FILENO);
+	if (!ttyflg)
+		return;
+
+	if (tcgetattr(STDOUT_FILENO, &tt) == -1)
+		err(1, "tcgetattr");
+
+	traw = tt;
+	cfmakeraw(&traw);
+	traw.c_lflag |= ISIG;
+	if (tcsetattr(STDOUT_FILENO, TCSANOW, &traw) == -1)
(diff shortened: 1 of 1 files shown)


8vasu:
Additional notes:

-  I used https://github.com/iiSeymour/game-of-life to reproduce the problem; authored by GitHub user iiSeymour, this, in the words of the author, is "a python implementation of Conway's Game of Life using the curses module with a small evolutionary twist"; it was written for Linux; therefore, one must call "python gol.py" on FreeBSD to bypass the "#!".

- This is a duplicate of https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=248372, which was also created by me.
I created this to maximize the chances of getting noticed. If the patch on Bugzilla is accepted, I will close this pull request, and vice versa.

- I submitted a similar patch for "scriptreplay(1)" from util-linux; it has now been merged by the maintainer of the project.

8vasu:
Further info:

- To reproduce the problem, please run the script(1) session in a terminal that does not have post processing turned off.

- NetBSD has accepted a similar patch. They modified it slightly: they call atexit(termreset) instead of the several termreset() calls that are present in this patch. While not necessary, upon request, I can edit this patch to reflect that change.

8vasu:
Patch was updated with atexit(termreset) instead of multiple calls of termreset.

8vasu:
Comfortable with FreeBSD Bugzilla. Closing.

PR #437, opened 2020-08-19
pci/ich handle ich_init() error

lzto:
ich_init() can only return 0 and ENXIO(6) and the caller is checking
failure using == -1

### DIFF
--- a/sys/dev/sound/pci/ich.c
+++ b/sys/dev/sound/pci/ich.c
@@ -1191,15 +1191,17 @@ ich_pci_resume(device_t dev)
 {
 	struct sc_info *sc;
 	int i;
+	int err;
 
 	sc = pcm_getdevinfo(dev);
 
 	ICH_LOCK(sc);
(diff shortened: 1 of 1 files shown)


sys/dev/sound/pci/ich.c
```diff
@@ -1196,7 +1196,7 @@ ich_pci_resume(device_t dev)
 
 	ICH_LOCK(sc);
 	/* Reinit audio device */
-    	if (ich_init(sc) == -1) {
+  if (ich_init(sc) != 0) {
```
bsdjhb, line 1199: Perhaps ideally the error from `ich_init` would be returned directly, but this fix is also fine.

lzto: yes, I agree with you. I only did one line change since it is returning ENXIO anyway.

lzto:
Hi, I did another revision, it now returns error directly.

PR #440, opened 2020-10-01
Improved docs

aditya11k:


markjdb:
I think "top level" is fairly standard terminology and "superior" is not, so it's not clear why this is an improvement.

PR #441, opened 2020-10-12
Add VMware ESXi ARM Fling support

darkain:
Now that there is an ARM64 version of VMware, I've added several of the drivers from AMD64 GENERIC over to ARM64 GENERIC. These drivers all work out of the box, unmodified, under VMware ESXi ARM Fling.


darkain:
Tracked as bug: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=250308

darkain:
These have been merged in, and work! :D

PR #445, opened 2020-12-05
usr.sbin/uefisign: Process zero-sized sections

kjcamann:
uefisign(8) tries to locate and sanity-check the sizes of all
PE sections, to compute a checksum. If a PE section contains only
uninitialized data (like .bss in ELF) then the "size of raw data"
and "pointer to raw data" section header fields are both supposed
to be zero, a condition the current code doesn't expect and which
causes it die. Most EFI binaries built these days contain one such
section, so this fixes uefisign for those binaries. Tested by signing
and booting with the FreeBSD Lua loader, which is currently broken
without this fix.

### DIFF
--- a/usr.sbin/uefisign/pe.c
+++ b/usr.sbin/uefisign/pe.c
@@ -232,7 +232,7 @@ parse_section_table(struct executable *x, off_t off, int number_of_sections)
 	range_check(x, off, sizeof(*psh) * number_of_sections,
 	    "section table");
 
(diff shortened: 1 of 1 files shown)


kjcamann:
A link to the PE format spec, search for "PointerToRawData":

https://docs.microsoft.com/en-us/windows/win32/debug/pe-format

markjdb:
@trasz could you take a look at this?

bsdimp:
12db51d20823a5e already fixes this issue.
5deb35002545 lands the second part.

PR #447, opened 2020-12-27
gconcat online append

main--:
I first submitted this patch [here](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=235473) and [here](https://reviews.freebsd.org/D19227) a while ago, but that evidently stalled. So here is another attempt.


bsdimp:
I'm going to merge the first two commits. They are a good cleanup. That should make the review easier.
The next one has a static function that's not called, so won't compile until we merge the 4th one.
This commit is missing documentation updates for the new functionality.

main--:
Thank you! Regarding documentation, do you need me to update anything other than the gconcat(8) manpage?

bsdimp:
> Thank you! Regarding documentation, do you need me to update anything other than the gconcat(8) manpage?

That's what I had in mind. That's the usual place to update geom things. Thanks!

main--:
Uh ... whoops?

main--:
Reopened as #472.

PR #448, opened 2020-12-28
macOS Finder .DS_Store

fel1x-developer:
It can be controversy, because it is for people who browse source on mac.
It thought about this pull request several times, but in Freebsd wiki, GitHub is for browsing and cloning, 
so I think .DS_Store should be added to .gitignore.


fel1x-developer:
Thank you for your recommendation.

PR #450, opened 2021-01-17
grep: fix LINKS in Makefile

arp242:
zstdegrep was listed twice, instead of zstdfgrep


uqs:
Thanks for the fix! (I wish GH would detect and auto-close the PR, hmm).

PR #451, opened 2021-01-23
User/claplace/gicv3 mbi

claplace:
These two commits add GICv3 support for ESXi Arm Fling virtual machines:
- The first one adds MSIs using SPI interrupts, with the SPI range described in the device tree.
- The second one is updating GICD_IROUTERn writes to be 64-bit instead of 32-bit. While 32-bit accesses are allowed, the register is 64-bit. And current ESXi Arm fling ignores 32-bit writes.


emaste:
Committed to main, thank you for the submission!

PR #454, opened 2021-01-29
Use the default Git history depth

fkorotkov:
Which is `50`. I saw a few errors like `Failed to force reset to SHA: object not found!` which seems is happening because the SHA is not available because there were two commits pushed almost simultaneously and the second from the top fails with this error because the SHA is not in the history.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -7,9 +7,6 @@ freebsd_instance:
   cpu: 8
   memory: 24G
 
-env:
-  CIRRUS_CLONE_DEPTH: 1
-
 task:
   only_if: $CIRRUS_BRANCH != 'svn_head'
   timeout_in: 120m


fkorotkov:
Seems then the clone time it getting just too high. 🤔

emaste:
Great, thank you! I'm not sure where I picked that up from, I'm worried I might have propagated it to other projects and will have a few cleanups to submit.

fkorotkov:
@emaste it seems affecting the `clone` time in a weird way. On my PR build it was just taking too long seems because of the calculations on GitHub side. Maybe they have some optimizations around single commit history clones. [The `main` build](https://cirrus-ci.com/task/5528735170428928) is already taking 4 minutes and counting instead of "just" 2. Might need to revert it back. :-(

fkorotkov:
I was poking around because I just switched Cirrus to optimize 4 and 8 cores VM tasks to use compute-optimized instances and wondered how much faster FreeBSD CI will be.

emaste:
Oh yeah, 1:54 vs 6:27 :(

So it seems right now have have to choose long clones or occasional failures.

kevans91:
It's not doing something silly like clone --no-single-branch, right?

fkorotkov:
It does a single branch shallow clone so I think it might be computation on GH side.

fkorotkov:
@emaste should I revert the change or you will do it? In other news the optimization to use compute optimized VMs for 4 or 8 core configurations seems dropped CI time for this repo by approximately 25% from an hour to 45 minutes.

emaste:
I'll revert. I wonder if there's a way if the clone can be made more robust against `SHA: object not found` (retry?)

emaste:
I reapplied this change for now as we were seeing a reasonably high failure rate - e.g. 7 out of 37 runs failed when I looked recently. Let me know if there's anything I can do to help debug or try to raise the priority of this issue with the git implementation you're using.

PR #455, opened 2021-02-10
msun: Fix incorrect results in `hypotl` near underflow

ararslan:
This patch is the work of @kargl, fixing a bug reported by @zimmermann6 in https://github.com/JuliaMath/openlibm/issues/224. @kargl's authorship has been preserved in the commit.

---

In the edited code, `t1` is intended to scale subnormal numbers, but `t1=1` is needed instead of `t1=0` when preparing the scaling.

With this test program, compiled with `-fno-builtin`:
```c
#include <stdio.h>
#include <math.h>

int main() {
    long double x = 0x3.6526795f4176ep-16384l;
    long double y = 0x3.ffffffffffffep-16352l;
    long double z = hypotl(x, y);
    printf("z=%La\n", z);
    return 0;
}
```
msun generates `z=0x0p+0`, whereas GNU libc generates `z=0xf.ffffffffffff8p-16354`.

The result from msun is incorrect, as `hypot(x, y) >= |y|`, and thus cannot be zero for nonzero `y`.

This issue was reported by zimmermann6 on GitHub to openlibm, which repackages parts of msun. The test case is their work.

### DIFF
--- a/lib/msun/src/e_hypotl.c
+++ b/lib/msun/src/e_hypotl.c
@@ -82,7 +82,7 @@ hypotl(long double x, long double y)
 	        man_t manh, manl;
 		GET_LDBL_MAN(manh,manl,b);
 		if((manh|manl)==0) return a;
-		t1=0;
+		t1=1;
 		SET_HIGH_WORD(t1,ESW(MAX_EXP-2));	/* t1=2^(MAX_EXP-2) */
 		b *= t1;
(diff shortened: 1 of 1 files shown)


bsdimp:
d3338f3355a612cf385632291f46c5777bba8d18 by dim@ fixed this shortly after the pull request was opened and merged to openlibm...

PR #461, opened 2021-03-31
Fix. cd-inverted and wp-inverte return all time true.

priittrees:
true   ^ 100b = true
false  ^ 100b = true


bsdimp:
Tweaked a bit in cfae21201a2b: used !! instead of bool cast because that's portable to some older compilers that some embedded platforms still occasionally use (though not the in tree ones). Word-smithed the commit message a bit as well.

PR #462, opened 2021-04-11
contrib: Fix typos in 'host' man page

chriswells0:
Fixes very minor typos in the docs for the host utility.


uqs:
Thanks for this submission. The problem is in "contributed" software however and it should be reported upstream, so that we don't diverge from the original sources.

@dag-erling it seems that https://tx97.net/ldns-host/ is dead. Can you point us to the proper upstream?

PR #463, opened 2021-04-13
bhyve: enhance debug info for memory range clash

RobCrowston:
Explain what the two clashing regions are.

### DIFF
--- a/usr.sbin/bhyve/mem.c
+++ b/usr.sbin/bhyve/mem.c
@@ -109,9 +109,11 @@ mmio_rb_add(struct mmio_rb_tree *rbt, struct mmio_rb_range *new)
 
 	if (overlap != NULL) {
 #ifdef RB_DEBUG
-		printf("overlap detected: new %lx:%lx, tree %lx:%lx\n",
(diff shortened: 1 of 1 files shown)


jake-87:
Are we sure that that "claimed" is the right wording? I would lean more towards "reserved", as that seems to be a more accurate description of the message.

usr.sbin/bhyve/mem.c
```diff
@@ -109,9 +109,11 @@ mmio_rb_add(struct mmio_rb_tree *rbt, struct mmio_rb_range *new)
 
 	if (overlap != NULL) {
 #ifdef RB_DEBUG
-		printf("overlap detected: new %lx:%lx, tree %lx:%lx\n",
+		printf("overlap detected: new %lx:%lx, tree %lx:%lx, '%s' "
+		       "claims region already claimed for '%s'\n",
```
jake-87, line 113: Reword to "claims region already reserved for"

jake-87 (changes requested):
Reword section for clariry

bsdimp:
This was approved in https://reviews.freebsd.org/D29696 and I landed as efec757b2042

PR #468, opened 2021-06-03
Optimized multiple string functions

gAlfonso-bit:
Removed old lingering variables meant for GCC 3 and added restrict qualifiers to arguments for strtok and memccpy as per The Open Group Base Specifications Issue 7, 2018 edition

### DIFF
--- a/lib/libc/string/memccpy.3
+++ b/lib/libc/string/memccpy.3
@@ -39,7 +39,7 @@
 .Sh SYNOPSIS
 .In string.h
 .Ft void *
-.Fn memccpy "void *dst" "const void *src" "int c" "size_t len"
+.Fn memccpy "void * __restrict dst" "const void * __restrict src" "int c" "size_t len"
 .Sh DESCRIPTION
 The
 .Fn memccpy
--- a/lib/libc/string/memccpy.c
+++ b/lib/libc/string/memccpy.c
@@ -38,13 +38,13 @@ __FBSDID("$FreeBSD$");
 #include <string.h>
 
 void *
-memccpy(void *t, const void *f, int c, size_t n)
+memccpy(void * __restrict t, const void * __restrict f, int c, size_t n)
 {
 
 	if (n) {
 		unsigned char *tp = t;
 		const unsigned char *fp = f;
-		unsigned char uc = c;
+		const unsigned char uc = (unsigned char)c;
 		do {
 			if ((*tp++ = *fp++) == uc)
 				return (tp);
--- a/lib/libc/string/memset.c
+++ b/lib/libc/string/memset.c
@@ -58,7 +58,7 @@ bzero(void *dst0, size_t length)
 #include <string.h>
 
 #define	RETURN	return (dst0)
-#define	VAL	c0
+#define	VAL	(u_char)c0
 #define	WIDEVAL	c
 
 void *
--- a/lib/libc/string/strtok.3
+++ b/lib/libc/string/strtok.3
@@ -55,9 +55,9 @@
 .Sh SYNOPSIS
 .In string.h
 .Ft char *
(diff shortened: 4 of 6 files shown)


lib/libc/string/memset.c
```diff
@@ -58,7 +58,7 @@ bzero(void *dst0, size_t length)
 #include <string.h>
 
 #define	RETURN	return (dst0)
-#define	VAL	c0
+#define	VAL	(u_char)c0
```
bsdimp, line 61: This seems out of place now. Can you split this out of the commit it is in and say it's an explicit cast because you are assigning through a u_char pointer?

lib/libc/string/memccpy.c
```diff
@@ -38,13 +38,13 @@ __FBSDID("$FreeBSD$");
...
 void *
-memccpy(void *t, const void *f, int c, size_t n)
+memccpy(void * __restrict t, const void * __restrict f, int c, size_t n)
 {
 
 	if (n) {
 		unsigned char *tp = t;
 		const unsigned char *fp = f;
-		unsigned char uc = c;
+		const unsigned char uc = (unsigned char)c;
```
bsdimp, line 47: This is arguably a separate commit, but just amending the commit message to say you're also explicitly casting 'c' to unsigned char while you are here will suffice.

bsdimp:
Thanks for updating the docs and the commit message.
I'd add a note to the strspn.c commit message that you're removing old code meant for gcc3 as well, like you did in the comment here.
I made a couple of other suggestions to the code...
This is getting quite close to commit... thanks for the changes so far...

lib/libc/string/strtok.3
```diff
@@ -55,9 +55,9 @@
 .Sh SYNOPSIS
 .In string.h
 .Ft char *
-.Fn strtok "char *str" "const char *sep"
+.Fn strtok "char * __restrict str" "const char * __restrict sep"
```
bsdimp, line 58: These here and below should be documented as `restrict` not `__restrict`

lib/libc/string/memccpy.3
```diff
@@ -39,7 +39,7 @@
 .Sh SYNOPSIS
 .In string.h
 .Ft void *
-.Fn memccpy "void *dst" "const void *src" "int c" "size_t len"
+.Fn memccpy "void * __restrict dst" "const void * __restrict src" "int c" "size_t len"
```
bsdimp, line 42: should be documented as `restrict` not `__restrict`.

bsdimp (changes requested):
The changes themselves look good.
However, the branch needs to be linear history (no merge commits) and have the changes separated logically: (1) for the restrict changes (both code and docs) (2) One for the explicit casts (3) strspn.c updates (4) mode_t changes for strmode (5) libc_nonshared (which might not be accepted, I need to confirm one way or another). Thanks!

tools/build/cross-build/include/linux/string.h
```diff
@@ -52,7 +52,7 @@ __BEGIN_DECLS
 size_t strlcpy(char *dst, const char *src, size_t siz);
 size_t strlcat(char *dst, const char *src, size_t siz);
 char *strnstr(const char *str, const char *find, size_t str_len);
-void strmode(/* mode_t*/ int mode, char *str);
+void strmode(mode_t mode, char *str);
```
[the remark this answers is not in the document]
jrtc27: Hmm, with glibc at least it's an unsigned int, not a signed int...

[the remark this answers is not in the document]
jrtc27: (for glibc's Linux and Hurd ports; for GNU/kFreeBSD it follows FreeBSD and uses an unsigned short)

PR #470, opened 2021-06-04
Optimized multiple string functions

gAlfonso-bit:
Removed old lingering variables meant for GCC 3 and added restrict qualifiers to arguments for strtok and memccpy as per The Open Group Base Specifications Issue 7, 2018 edition


lib/libc/string/strspn.c
```diff
@@ -39,34 +39,24 @@ __FBSDID("$FreeBSD$");
...
-	int idx;
 
 	if(*s == '\0')
 		return (0);
 
 #if LONG_BIT == 64	/* always better to unroll on 64-bit architectures */
 	tbl[3] = tbl[2] = tbl[1] = tbl[0] = 0;
 #else
-	for (idx = 0; idx < sizeof(tbl) / sizeof(tbl[0]); idx++)
+	for (unsigned int idx = 0; idx < sizeof(tbl) / sizeof(tbl[0]); idx++)
```
jrtc27, line 51: size_t, and I'm unconvinced we need the manual unrolling, every compiler of interest on godbolt will unroll this loop, it's such a basic optimisation to make here.

lib/libc/string/strlcat.c
```diff
@@ -32,23 +32,22 @@ __FBSDID("$FreeBSD$");
...
+	if (n == 0)
 		return(dlen + strlen(src));
 	while (*src != '\0') {
-		if (n != 0) {
+		if (--n != 0) {
 			*dst++ = *src;
-			n--;
 		}
 		src++;
 	}
```
jrtc27, line 53: Why? This has done nothing interesting to the code, and I'd expect it to compile to identical assembly.

lib/libc/string/strlcpy.c
```diff
@@ -30,20 +30,20 @@ __FBSDID("$FreeBSD$");
...
 	}
 
 	/* Not enough room in dst, add NUL and traverse rest of src. */
-	if (nleft == 0) {
-		if (dsize != 0)
+	if (dsize == 0) {
+		if (odsize != 0)
 			*dst = '\0';		/* NUL-terminate dst */
 		while (*src++)
 			;
```
jrtc27, line 49: Why? This is just unnecessary churn.

tools/build/cross-build/include/linux/string.h
```diff
@@ -49,8 +49,8 @@
 #include <sys/cdefs.h>
 
 __BEGIN_DECLS
-size_t strlcpy(char *dst, const char *src, size_t siz);
-size_t strlcat(char *dst, const char *src, size_t siz);
+size_t strlcpy(char * __restrict dst, const char * __restrict src, size_t siz);
+size_t strlcat(char * __restrict dst, const char * __restrict src, size_t siz);
```
jrtc27, line 53: These are prototypes for glibc's functions (or another Linux libc), not FreeBSD's, used when building native tools at the start of the build.

PR #472, opened 2021-06-06
gconcat online append

main--:
Rebased #447 (sadly github won't let me reopen that one)

### DIFF
--- a/lib/geom/concat/gconcat.8
+++ b/lib/geom/concat/gconcat.8
@@ -46,6 +46,11 @@
 .Ar name
 .Ar prov ...
 .Nm
+.Cm append
+.Op Fl hv
+.Ar name
+.Ar prov
+.Nm
 .Cm stop
 .Op Fl fv
 .Ar name ...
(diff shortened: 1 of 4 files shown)


gAlfonso-bit:
Glad it’s not just me having errors despite the tests passing locally. I was going crazy

emaste:
You can check the status of the same CI task that runs against PRs at https://cirrus-ci.com/github/freebsd/freebsd-src/main
It is currently passing but was recently broken for a while.

bsdimp:
I think I like the code, but I'll need to study is tad bit more... one issue I have is the commit messages for the commits could use some work. https://www.pullrequest.com/blog/writing-a-great-pull-request-description/ or https://docs.freebsd.org/en/articles/committers-guide/#commit-log-message for some good guidelines, or ask here if you are unsure after reading them.

main--:
Great! With the preliminary changes already merged, and given the fact that the first commit just adds an unused function, do you think it makes sense to squash these three commits into one?

bsdimp:
> Great! With the preliminary changes already merged, and given the fact that the first commit just adds an unused function, do you think it makes sense to squash these three commits into one?

Yes. I believe so.

main--:
Done!

bsdimp:
For future reference: the .Dd date in the man page should have been updated.
Also, it would have been nice to rebase this to a more recent main. An unrelated failure is why the workflow shows X.
However, I was able to do both of those trivially. Doing some final checks and will report back if anything comes up.

bsdimp:
I've pushed this in as d575e81fbcfa

main--:
Thank you!

bsdimp:
> Thank you!

You bet. If you have any further issues, please open a new pull request :). thanks for your submission.

PR #473, opened 2021-06-07
bsdinstall: set SCRIPT=/path/to/script env var to be able to automate…

eoli3n:
### automated bsdinstall to a jail

It doesn't seem to be the cleanest way, but it works with my simple example.

```bash
$ cat << EOF > /usr/local/jails/jail.template
DISTRIBUTIONS="base.txz"
export nonInteractive="YES"

#!/bin/sh
sysrc ifconfig_eb0_bind="inet 192.168.0.101 netmask 255.255.255.0"
sysrc defaultrouter="192.168.0.254"
EOF

$ zfs create -o mountpoint=/usr/local/jails zroot/jails
$ zfs create zroot/jails/jail1

$ env SCRIPT=/usr/local/jails/jail.template bsdinstall jail /usr/local/jails/jail1
```

### DIFF
--- a/usr.sbin/bsdinstall/scripts/jail
+++ b/usr.sbin/bsdinstall/scripts/jail
@@ -52,10 +52,15 @@ error() {
 	fi
 }
 
-
 rm -rf $BSDINSTALL_TMPETC
 mkdir $BSDINSTALL_TMPETC
 mkdir -p $1 || error "mkdir failed for $1"
+
+if [ -n "$SCRIPT" ]
+then
+        split -a 2 -p '^#!.*' "$SCRIPT" $TMPDIR/bsdinstall-installscript-
+        . $TMPDIR/bsdinstall-installscript-aa
+fi
 
 test ! -d $BSDINSTALL_DISTDIR && mkdir -p $BSDINSTALL_DISTDIR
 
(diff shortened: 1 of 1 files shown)


usr.sbin/bsdinstall/scripts/jail
```diff
@@ -52,11 +52,16 @@ error() {
...
 
-
 rm -rf $BSDINSTALL_TMPETC
 mkdir $BSDINSTALL_TMPETC
 mkdir -p $1 || error "mkdir failed for $1"
 
+if [ -n "$SCRIPT" ]
+then
+        split -a 2 -p '^#!.*' "$SCRIPT" $TMPDIR/bsdinstall-installscript-
+        . $TMPDIR/bsdinstall-installscript-aa
```
bsdimp, line 62: What's the benefit of stripping the initial comments? #!... would be treated as a comment by /bin/sh. Can't you just .  the "$SCRIPT"?

eoli3n: I didn't write the line, it comes from https://github.com/freebsd/freebsd-src/blob/main/usr.sbin/bsdinstall/scripts/script#L93

eoli3n: and see ``SCRIPTING`` part : https://www.unix.com/man-page/freebsd/8/bsdinstall/

allanjude: The point is to split the 'script' file into two parts
everything before the sh-bang #! is variables to set, and everything after it is a 'post install' script to run at the end.

usr.sbin/bsdinstall/scripts/jail
```diff
@@ -69,20 +74,23 @@ if [ ! -f $BSDINSTALL_DISTDIR/MANIFEST -a -z "$BSDINSTALL_DISTSITE" ]; then
...
-	EXTRA_DISTS=$(echo $DISTMENU | xargs dialog \
-	    --backtitle "FreeBSD Installer" \
-	    --title "Distribution Select" --nocancel --separate-output \
-	    --checklist "Choose optional system components to install:" \
-	    0 0 0 \
-	2>&1 1>&3)
-	for dist in $EXTRA_DISTS; do
-		export DISTRIBUTIONS="$DISTRIBUTIONS $dist.txz"
-	done
+    if [ -z "$SCRIPT" ]
```
allanjude, line 81: It might make more sense here to check for the nonInteractive variable instead.

usr.sbin/bsdinstall/scripts/jail
```diff
@@ -108,20 +116,46 @@ fi
...
-bsdinstall rootpass || error "Could not set root password"
+if [ -z "$SCRIPT" ]
+then
+    bsdinstall rootpass || error "Could not set root password"
+fi
+
+if [ -n "$SCRIPT"]
+then
+    # Configure bootloader if needed
+    bsdinstall bootconfig
```
allanjude, line 127: I don't think it makes sense to install the boot config in a jail

eoli3n: It comes from https://github.com/freebsd/freebsd-src/blob/main/usr.sbin/bsdinstall/scripts/script#L151

usr.sbin/bsdinstall/scripts/jail
bsdimp: yea, these scripts are 'special' and 'different' in ways that are opaque...  OK. this is fine then...

usr.sbin/bsdinstall/scripts/jail
eoli3n: done

usr.sbin/bsdinstall/scripts/jail
eoli3n: done

usr.sbin/bsdinstall/scripts/jail
```diff
@@ -108,20 +116,43 @@ fi
...
-bsdinstall rootpass || error "Could not set root password"
+if [ -z "$SCRIPT" ]
+then
+    bsdinstall rootpass || error "Could not set root password"
+fi
+
+if [ -n "$SCRIPT" ]
+then
+    # Finalize install
+    bsdinstall config
```
allanjude, line 127: `bsdinstall config` is already run on line 142

usr.sbin/bsdinstall/scripts/jail
```diff
@@ -108,20 +116,43 @@ fi
 
 bsdinstall checksum || error "Distribution checksum failed"
 bsdinstall distextract || error "Distribution extract failed"
-bsdinstall rootpass || error "Could not set root password"
+if [ -z "$SCRIPT" ]
```
allanjude, line 119: If this is to be conditional, it should use nonInteractive as well, so you can still script it to prompt if you wish

allanjude (changes requested):

bsdimp:
Landed as 6f4c1456b51d in current.

PR #474, opened 2021-06-08
Change “compiled” to “assembled”

gAlfonso-bit:
Assembly files are assembled, not compiled.


bsdimp:
Committed as b0e54e61b390 without an author name set because there was no valid email address with these commits. I used submitted by instead.

PR #475, opened 2021-06-12
syslogd: fix WITHOUT_INET builds

akhramov:
Since 2d82b47 syslogd can't be built with `WITHOUT_INET` or
`WITHOUT_INET6` build variables set, because `iovlist_truncate` is not
defined but used.

This change wraps the problematic `iovlist_truncate` call within ifdef
directive.


bsdimp:
This is correct. I did a minor tweak to the commit message to explain why this is ok better. `iovlist_truncate` is compiled out for this situation, and should have been noted in the commit message. Thanks!
f06fa6f88733

bsdimp:
Commit f06fa6f88733 is what's needed.

PR #478, opened 2021-06-16
Simplify fuse_device_filt_write to directly return 1

gAlfonso-bit:
It always returns 1, so why bother having a variable?

We should directly return 1


asomers (approved):
That was fast, Alfonso.  You must be watching the commits mailing list like a hawk.

asomers:
And to answer your question, the variable exists because I copied it from a few lines above.

asomers:
Merged in 9b876fbd504e5c718d8d0275b32d806ab14558c8

PR #479, opened 2021-06-16
top(1): support command name and argument grepping

jgrafton:
This feature is similar to the grep functionality in OpenBSD's top command.

### DIFF
--- a/usr.bin/top/top.c
+++ b/usr.bin/top/top.c
@@ -53,6 +53,7 @@ typedef void sigret_t;
 static char stdoutbuf[Buffersize];
 
 static int fmt_flags = 0;
+int show_args = false;
 int pcpu_stats = false;
 
 /* signal handling routines */
@@ -907,6 +908,19 @@ restart:
 				}
 				break;
 
+			    case CMD_grep: /* grep command name */
+				new_message(MT_standout,
+				    "Grep command name: ");
+				if (readline(tempbuf1, sizeof(tempbuf1), false) > 0) {
+					free(ps.command);
+					if (tempbuf1[0] == '+' && tempbuf1[1] == '\0') {
+						ps.command = NULL;
+					} else if ((ps.command = strdup(tempbuf1)) == NULL)
+						quit(1);
+				}
+				clear_message();
+				break;
+
 			    case CMD_displays:	/* change display count */
 				new_message(MT_standout,
 					"Displays to show (currently %s): ",
@@ -1025,6 +1039,7 @@ restart:
 				break;
 			    case CMD_showargs:
 				fmt_flags ^= FMT_SHOWARGS;
+				show_args = fmt_flags & FMT_SHOWARGS;
 				new_message(MT_standout | MT_delayed,
 				    " %sisplaying process arguments.",
 				    fmt_flags & FMT_SHOWARGS ? "D" : "Not d");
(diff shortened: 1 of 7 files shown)


bsdimp:
This passes my initial traige, and also looks cool :)

jgrafton:
@allanjude mentioned this feature on _BSD Now_ and I thought it sounded like a great challenge to tackle ;)

usr.bin/top/top.c
```diff
@@ -907,6 +908,20 @@ main(int argc, const char *argv[])
 				}
 				break;
 
+			    case CMD_grep: /* grep command name */
+				new_message(MT_standout,
+				    "Grep command name: ");
+				if (readline(tempbuf1, sizeof(tempbuf1), false) > 0)
+				{
```
bsdimp, line 915: This { needs to be at the end of the prior line.

Also, if I do /getty<CR> I see just the getty routines. However, if I do /<CR> next, it doesn't reset the display to all processes because readline is return  -1 in this case...

bsdimp:
So this needs an update to proceed.

bsdimp:
Also, the cmd_match appears to be lifted 100% from OpenBSD w/o credit in the commit message.

usr.bin/top/top.c
bsdimp: And there's a resource leak, both of these can be fixed by free'ing ps.command and setting it to  NULL before the readline...

usr.bin/top/top.c
```diff
@@ -907,6 +908,20 @@ main(int argc, const char *argv[])
...
+					if (tempbuf1[0] == '+' &&
+						tempbuf1[1] == '\0')
+						ps.command = NULL;
+					else if ((ps.command = strdup(tempbuf1)) == NULL)
+						quit(1);
+					clear_message();
```
bsdimp, line 921: And this clear_message() needs to be outside the if, because we need to do it either way.

jgrafton: ah, good catch

usr.bin/top/top.c
jgrafton: Resetting the display to all processes is handled by /+  That's the behavior in OpenBSD.  I like it because it allows you to back out of a search.

The missing free is fixed.  A const qualifier was added to the process_select struct in fc36f5a7cb968 though I'm unsure why.

bsdimp:
I'm still thinking about the const char *command -> char *command change...  Otherwise I think this may be good. The code snagged from OpenBSD doesn't exactly match our current preferred style, but I think it's OK as it's in the style top was written in...

usr.bin/top/top.c
```diff
@@ -907,6 +908,19 @@ main(int argc, const char *argv[])
 				}
 				break;
 
+			    case CMD_grep: /* grep command name */
+				new_message(MT_standout,
+				    "Grep command name: ");
+				if (readline(tempbuf1, sizeof(tempbuf1), false) > 0) {
+					if (tempbuf1[0] == '+' && tempbuf1[1] == '\0') {
+						free(ps.command);
```
bsdimp, line 916: This free needs to be moved up to before the if since we need to do this in both cases.
Otherwise we leak the previous ps.command

bsdimp (changes requested):

usr.bin/top/top.c
jgrafton: Ah, yes, you're right.  It needs to be freed regardless if display is reset or a new value is entered.  Thanks for the quick review, Warner!

usr.bin/top/top.c
bsdimp: The const can be removed, I think.

bsdimp:
This looks good at first blush. I'll take a look at it (in that I'll import it into a trial branch and compile it, test it, etc)

bsdimp:
Landed as a00d703f2f43

PR #480, opened 2021-06-16
Optimize and modernize string functions

gAlfonso-bit:


bsdimp:
I noticed on another pull request that your author name is:

Author: gAlfonso-bit <83477269+gAlfonso-bit@users.noreply.github.com>

which is a bit too anonymous for this project. Do you have something better you can either set your author name to, or that you can provide for use to use for your submissions? By better, it should ideally be your real name with an email address that works.

gAlfonso-bit:
So the strmode changes are breaking the boot. So I’ll work on that in another PR. Everything else I’ll make into new commits and a new PR, but other than that, the code itself is ready for review @bsdimp

PR #483, opened 2021-06-18
Make hacking

bsdimp:
So, if I make a pull request, will github automation run for me, a long time contributor?
If you are 'minding the store' before I get to this, this is just a test...


bsdimp:
this test has shown what I need to show.

emaste:
Aside, Cirrus-CI is passing but the cross-build GitHub jobs are not, we need to fix or disable them.

PR #485, opened 2021-06-20
release: Remove C-like string comparison operator

jlduran:
Per `[(1)`, i.e. `test(1)`, the string comparison operator should be `=`, not `==` in `sh(1)` scripts.

No functional change.


bsdimp:
Committed as 3f197b101f71 after the re signed off on it.

PR #486, opened 2021-06-20
Add var db zoneinfo

jlduran:
Per `tzsetup(8)`, `/etc/localtime` and `/var/db/zoneinfo` go hand in hand.

(Submitted as two separate commits on purpose.)

### DIFF
--- a/tools/tools/sysbuild/sysbuild.sh
+++ b/tools/tools/sysbuild/sysbuild.sh
@@ -243,7 +243,7 @@ ports_recurse() (
 
 ports_build() (
 
-	ports_recurse . $PORTS_WE_WANT 
+	ports_recurse . $PORTS_WE_WANT
 
 	if [ "x${PKG_DIR}" != "x" ] ; then
 		mkdir -p ${PKG_DIR}
@@ -262,7 +262,7 @@ ports_build() (
 
 			(
 			cd $p
-			ports_make clean all install 
+			ports_make clean all install
 			) > _.$b 2>&1 < /dev/null
 			continue
 		fi
@@ -340,7 +340,7 @@ ports_prefetch() (
 		) > /${ldir}/_.prefetch.$b 2>&1
 	done
 	echo "Done" >> /${ldir}/_.prefetch
-	) 
+	)
 )
 
 #######################################################################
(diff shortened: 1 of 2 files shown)


usr.sbin/bsdinstall/scripts/jail
```diff
@@ -137,6 +137,7 @@ trap error SIGINT	# SIGINT is bad again
 bsdinstall config  || error "Failed to save config"
 cp /etc/resolv.conf $1/etc
 cp /etc/localtime $1/etc
+cp /var/db/zoneinfo $1/var/db
```
bsdimp, line 140: Two coments: first, these are two different programs, so should have two separate commits
Second, in the sysbuild patch you checked to see if this exists, but here you do not, why?

tools/tools/sysbuild/sysbuild.sh
```diff
@@ -623,14 +627,14 @@ cp /tmp/_sb_log ${SBMNT}/tmp
...
 	cp $c_arg ${SBMNT}/root
-	chroot ${SBMNT} sh /root/$0 -c /root/`basename $c_arg` $use_pkg chroot_script 
+	chroot ${SBMNT} sh /root/$0 -c /root/`basename $c_arg` $use_pkg chroot_script
 else
 	chroot ${SBMNT} sh /root/$0 $use_pkg chroot_script
 fi
 cp ${SBMNT}/tmp/_sb_log /tmp
 
 log_it create all mountpoints
-grep -v '^[ 	]*#' ${SBMNT}/etc/fstab |
```
bsdimp, line 633: while it's fine this time, in the future, can you separate out whitespace from non-whitespace changes?

usr.sbin/bsdinstall/scripts/jail
jlduran: I was just following the previous "style".

I would prefer to check if the file exists first as well. However, since this is `bsdinstall(8)` one might think the file is not there or you want to overwrite anyway.

tools/tools/sysbuild/sysbuild.sh
jlduran: It is easy for me to force push a third commit with just whitespace changes. I'll do that. Thank you!

usr.sbin/bsdinstall/scripts/jail
bsdimp: I was thinking more basic: what if the host doesn't have it... but now that I'm thinking about it, there's bigger issues if the host doesn't have it...

bsdimp:
Landed as 91604330e607 and 6ce785c50bfa

bsdimp:
Looks like I was too quick to pick up the forced push with white space change separation.... thanks for trying :)

PR #487, opened 2021-06-21
geom_disk: use a preallocated geom_event for disk destruction.

bsdimp:
Preallocate a geom_event (using the new geom_alloc_event) when we create
a disk. When we create the disk, we're going to be in a sleepable
context, so we can always allocate this extra bit of memory. Then use
this preallocated memory to free the disk. CAM can try to free the disk
from an unsleepable context if there was I/O outstanding when the disk
was destroyted (say because the SIM said it had gone away). The I/O
context isn't sleepable. Rather than trying to invent a retry mechanism
and making sure all the other geom_disk consumers did it properly,
preallocating the event ensure that the geom_disk will be properly torn
down, even when there's memory pressure when the disk departs.

Sponsored by:		Netflix

### DIFF
--- a/sys/geom/geom.h
+++ b/sys/geom/geom.h
@@ -269,6 +269,9 @@ int g_media_changed(struct g_provider *pp, int flag);
 int g_media_gone(struct g_provider *pp, int flag);
 void g_orphan_provider(struct g_provider *pp, int error);
 void g_waitidlelock(void);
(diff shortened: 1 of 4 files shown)


bsdimp:
Note: another test.

bsdimp:
OK. This test has shown what I need to show.

PR #489, opened 2021-06-23

gAlfonso-bit:
Now the function is proper and working.

### DIFF
--- a/tools/build/cross-build/include/linux/string.h
+++ b/tools/build/cross-build/include/linux/string.h
@@ -52,7 +52,7 @@ __BEGIN_DECLS
 size_t strlcpy(char *dst, const char *src, size_t siz);
 size_t strlcat(char *dst, const char *src, size_t siz);
 char *strnstr(const char *str, const char *find, size_t str_len);
-void strmode(/* mode_t*/ int mode, char *str);
+void strmode(mode_t mode, char *str);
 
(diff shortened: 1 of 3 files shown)


tools/build/cross-build/include/linux/string.h
```diff
@@ -52,7 +52,7 @@ __BEGIN_DECLS
 size_t strlcpy(char *dst, const char *src, size_t siz);
 size_t strlcat(char *dst, const char *src, size_t siz);
 char *strnstr(const char *str, const char *find, size_t str_len);
-void strmode(/* mode_t*/ int mode, char *str);
+void strmode(mode_t mode, char *str);
```
bsdimp, line 55: mode_t isn't necessarily defined in the cross compilation environments. The current definition matches glibc on various linux machines we need to bootstrap from. So I think this change should be dropped.

gAlfonso-bit: It is though. If it weren’t, none of the cross build tests would have passed

gAlfonso-bit: In fact this change is needed for the cross-build to work at all because it turns out, strmode isn’t in glibc at all! It’s not even in the comments of glibc. In fact, searching for strmode in glibc gives 0 results.

PR #492, opened 2021-06-23

gAlfonso-bit:


gAlfonso-bit:
@bsdimp is this good?

PR #493, opened 2021-06-25

gAlfonso-bit:
Finally, we have the correct function definition for strmode.


bsdimp:
849dcdb1c0c50b and 0d9e8c286df6c have been pushed to main.

PR #494, opened 2021-06-25
Add restrict qualifiers and optimize string functions

gAlfonso-bit:


gAlfonso-bit:
@bsdimp ready for review

lib/libc/string/bcopy.c
```diff
@@ -101,18 +104,19 @@ bcopy(const void *src0, void *dst0, size_t length)
 		 * Copy whole words, then mop up any trailing bytes.
 		 */
 		t = length / wsize;
-		TLOOP(*(word *)(void *)dst = *(const word *)(const void *)src;
+		TLOOP(*(word *)dst = *(const word *)src;
 		    src += wsize; dst += wsize);
 		t = length & wmask;
 		TLOOP(*dst++ = *src++);
+#ifndef MEMCOPY
```
jillest, line 111: In practice, programs do depend on `memcpy()` working on overlapping areas sensibly, even though it is wrong. Therefore, breaking this is unwise.

At least on x86, I expect that the gain from not handling overlaps is quite small, except if modern CPUs have really lowered the setup cost of REP MOVSB so much that it can always be used. In such a case, small copies are best done by loading the data into registers and then storing it at the destination (taking advantage of unaligned access to reduce branches), which naturally handles overlaps, while the overhead of a branch is limited in large copies.

gAlfonso-bit: Which programs would that be, if I may ask? Can I have an example?

vixie:
On Sat, Jun 26, 2021 at 01:43:46PM -0700, Jilles Tjoelker wrote:
> @jillest commented on this pull request.
> 
> >  		    src += wsize; dst += wsize);
> 
> In practice, programs do depend on `memcpy()` working on overlapping areas
> sensibly, even though it is wrong. Therefore, breaking this is unwise.

i disagree strongly. if a program expects something the API definition does
not promise, then the implementation should behave in a deliberately breaking
way. allowing code to make nonstandard / nonportable assumptions does nobody
any favours. see also this:

https://www.youtube.com/watch?v=Vr_T5hFtyoM

-- 
Paul Vixie

gAlfonso-bit:
I think I crossed the threshold for amount of changes appropriate for this PR.

So I’m going to leave it at that.

PR #496, opened 2021-06-30
Fix potential overflow when comparing wide char types

AtariDreams:
Comparing wide char types may result in an overflow when subtracting them if one value happens to be extremely large and the other isn’t, as the difference is evaluated as an unsigned int and then returned as an int.

Although this isn’t a problem for small numbers, such as 0 and 1 as the result will just be 0xFFFFFFFF, or -1, but if the subtraction is 0xFFFFFFFF and 2, the result will turn out to be -3, despite the first value being bigger when interpreted as an unsigned int.

### DIFF
--- a/lib/libc/string/wcscasecmp.c
+++ b/lib/libc/string/wcscasecmp.c
@@ -39,7 +39,9 @@ wcscasecmp(const wchar_t *s1, const wchar_t *s2)
 		c1 = towlower(*s1);
 		c2 = towlower(*s2);
 		if (c1 != c2)
-			return ((int)c1 - c2);
+			return ((unsigned int)c1 > (unsigned int)c2 ? 1 : -1);
+		if (c1 == L'\0')
+			return (0);
 	}
-	return (-*s2);
+	return (-1);
 }
--- a/lib/libc/string/wcscmp.c
+++ b/lib/libc/string/wcscmp.c
@@ -48,9 +48,11 @@ int
 wcscmp(const wchar_t *s1, const wchar_t *s2)
 {
 
-	while (*s1 == *s2++)
-		if (*s1++ == '\0')
+	while (*s1 == *s2) {
+		if (*s1 == L'\0')
 			return (0);
+		++s1, ++s2;
+	}
 	/* XXX assumes wchar_t = int */
-	return (*(const unsigned int *)s1 - *(const unsigned int *)--s2);
+	return (*(const unsigned int *)s1 > *(const unsigned int *)s2 ? 1 : -1);
 }
--- a/lib/libc/string/wcsncasecmp.c
+++ b/lib/libc/string/wcsncasecmp.c
@@ -41,9 +41,9 @@ wcsncasecmp(const wchar_t *s1, const wchar_t *s2, size_t n)
 		c1 = towlower(*s1);
 		c2 = towlower(*s2);
 		if (c1 != c2)
-			return ((int)c1 - c2);
-		if (--n == 0)
+			return ((unsigned int)c1 > (unsigned int)c2 ? 1 : -1);
+		if (c1 == L'\0' || --n == 0)
 			return (0);
 	}
-	return (-*s2);
+	return (-1);
 }
--- a/lib/libc/string/wcsncmp.c
+++ b/lib/libc/string/wcsncmp.c
@@ -45,13 +45,14 @@ wcsncmp(const wchar_t *s1, const wchar_t *s2, size_t n)
 	if (n == 0)
 		return (0);
 	do {
-		if (*s1 != *s2++) {
+		if (*s1 != *s2) {
 			/* XXX assumes wchar_t = int */
-			return (*(const unsigned int *)s1 -
-			    *(const unsigned int *)--s2);
+			return (*(const unsigned int *)s1 >
+			    *(const unsigned int *)s2 ? 1 : -1);
 		}
-		if (*s1++ == 0)
+		if (*s1 == L'\0')
 			break;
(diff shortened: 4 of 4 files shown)


lib/libc/string/strlen.c
```diff
@@ -126,5 +126,4 @@ strlen(const char *str)
 	}
 
 	/* NOTREACHED */
-	return (0);
```
bsdimp, line 129: This is not right. Many compilers will warn if you don't end the function like this, even if it is nominally not reached.

lib/libc/string/wcscasecmp.c
```diff
@@ -35,11 +35,13 @@ wcscasecmp(const wchar_t *s1, const wchar_t *s2)
 {
 	wchar_t c1, c2;
 
-	for (; *s1; s1++, s2++) {
+	for (;; s1++, s2++) {
```
bsdimp, line 38: There's no real reason for this change and the companion if below.

lib/libc/string/wcscasecmp.c
```diff
@@ -35,11 +35,13 @@ wcscasecmp(const wchar_t *s1, const wchar_t *s2)
...
 		c1 = towlower(*s1);
 		c2 = towlower(*s2);
 		if (c1 != c2)
-			return ((int)c1 - c2);
+			return ((unsigned int)c1 > (unsigned int)c2 ? 1 : -1);
+		if (c1 == L'\0')
+			return (0);
 	}
-	return (-*s2);
+	/* NOTREACHED */
```
bsdimp, line 46: This is not right, and many compilers will warn about this.

lib/libc/string/wcscasecmp.c
```diff
@@ -35,11 +35,13 @@ wcscasecmp(const wchar_t *s1, const wchar_t *s2)
 {
 	wchar_t c1, c2;
 
-	for (; *s1; s1++, s2++) {
+	for (;; s1++, s2++) {
 		c1 = towlower(*s1);
 		c2 = towlower(*s2);
 		if (c1 != c2)
-			return ((int)c1 - c2);
+			return ((unsigned int)c1 > (unsigned int)c2 ? 1 : -1);
```
bsdimp, line 42: I fail to see the benefit from this cast. What's the reason for it?

lib/libc/string/wcscmp.c
```diff
@@ -48,9 +48,11 @@ int
 wcscmp(const wchar_t *s1, const wchar_t *s2)
 {
 
-	while (*s1 == *s2++)
-		if (*s1++ == '\0')
+	while (*s1 == *s2) {
+		if (*s1 == L'\0')
 			return (0);
+		++s1, ++s2;
+	}
```
bsdimp, line 55: So what's the benefit from making this more complex?

lib/libc/string/wcsncasecmp.c
```diff
@@ -37,13 +37,13 @@ wcsncasecmp(const wchar_t *s1, const wchar_t *s2, size_t n)
...
 		c2 = towlower(*s2);
 		if (c1 != c2)
-			return ((int)c1 - c2);
-		if (--n == 0)
+			return ((unsigned int)c1 > (unsigned int)c2 ? 1 : -1);
+		if (c1 == L'\0' || --n == 0)
 			return (0);
 	}
-	return (-*s2);
+	/* NOTREACHED */
```
bsdimp, line 48: Again, this is not correct.

lib/libc/string/strlen.c
gAlfonso-bit: Not every function in this codebase has this and it’s been compiling fine for years so I don’t know.

lib/libc/string/wcscasecmp.c
gAlfonso-bit: Because wide char types are assumed to be 32 bit. This casting wouldn’t be needed if these were regular chars

lib/libc/string/wcscmp.c
gAlfonso-bit: More complex is doing the ++ and then the — in the return statement, not this.

lib/libc/string/wcscasecmp.c
gAlfonso-bit: Yes there is because if the leading bit is 1, the return value will be positive instead of negative.

gAlfonso-bit:
Remember we are comparing the values as unsigned instead of signed. Wide char types that are likely sign extended will end up resulting in a negative return value if we just compare without casting or specifying this. 

That’s why this PR was opened in the first place. Because of this, the -*s2 doesn’t work

gAlfonso-bit:
If this is intended behavior, I’ll just close the PR.

lib/libc/string/wcscasecmp.c
bsdimp, line 38: This is wrong and I won't merge it.

lib/libc/string/wcscmp.c
bsdimp: this change is churn, I'm not going to risk merging it.

lib/libc/string/wcsncasecmp.c
```diff
@@ -37,13 +37,13 @@ wcsncasecmp(const wchar_t *s1, const wchar_t *s2, size_t n)
 
 	if (n == 0)
 		return (0);
-	for (; *s1; s1++, s2++) {
+	for (;; s1++, s2++) {
```
bsdimp, line 40: This improperly removes the termination condition.  It's easier to understand

PR #497, opened 2021-06-30
Remove incorrect __restricted labels to avoid UB.

gAlfonso-bit:
strcspn isn’t supposed to be restricted, and by making it so, it opens the door to UB, as the standard describes the function as finding a complimentary substring, which by definition may include part of the string being compared to.

### DIFF
--- a/sys/libkern/strcspn.c
+++ b/sys/libkern/strcspn.c
@@ -32,7 +32,7 @@ __FBSDID("$FreeBSD$");
 #include <sys/limits.h>
 
 #define	IDX(c)	((u_char)(c) / LONG_BIT)
-#define	BIT(c)	(1UL << ((u_char)(c) % LONG_BIT))
+#define	BIT(c)	((u_long)1 << ((u_char)(c) % LONG_BIT))
 
 size_t 
 strcspn(const char *s, const char *charset) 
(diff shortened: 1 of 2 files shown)


gAlfonso-bit:
This issue doesn’t affect the strspn because that doesn’t have the restricted qualifier

bsdimp:
I don't think this is a UB issue. However, __restrict isn't in usermode version of this and it's not in the standard.

sys/libkern/strcspn.c
```diff
@@ -32,10 +32,10 @@ __FBSDID("$FreeBSD$");
 #include <sys/limits.h>
 
 #define	IDX(c)	((u_char)(c) / LONG_BIT)
-#define	BIT(c)	((u_long)1 << ((u_char)(c) % LONG_BIT))
+#define	BIT(c)	(1UL << ((u_char)(c) % LONG_BIT))
```
bsdimp, line 35: This is an unrelated change.

gAlfonso-bit:
Fixed.

bsdimp:
https://cgit.freebsd.org/src/commit/?id=56d33e86b74b197a36f42255824b56715c96a596 landed this, closing.

PR #499, opened 2021-07-02
passwd: remove the hardcoded local / NIS user check

Trumeet:
This check was first introduced as the process of PAMify passwd back in 2002 (see 5f0ae68a1890ba5b50257f7195ebdeaa789c9079). However, passwd already used pam for password modification and thus the password change process should be handled by specific pam modules. Therefore, the check to make passwd only available to local or NIS users is nonsense.

On my systems, passwd works perfectly to change users' LDAP passwords using nss-pam-ldapd or their Kerberos passwords using pam-krb5 in the ports after removing this check, so I believe it has nothing to do with passwd.

### DIFF
--- a/usr.bin/passwd/passwd.c
+++ b/usr.bin/passwd/passwd.c
@@ -122,9 +122,9 @@ main(int argc, char *argv[])
 		    pwd->pw_name);
 		break;
 	default:
-		/* XXX: Green men ought to be supported via PAM. */
-		errx(1, 
-	  "Sorry, `passwd' can only change passwords for local or NIS users.");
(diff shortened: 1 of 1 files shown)


bsdimp:
Feedback from maintainer: "PAM does not know where the
account came from and therefore cannot know which pam_sm_chauthtok() to
call, even if it had a mechanism to choose different paths for different
users, which it does not.  The correct solution is to use a separate
utility for remote users, in this case ldappasswd(1) which comes with
OpenLDAP."

so I'm inclined to close this soon unless that will be addressed...

PR #504, opened 2021-07-06
Split definitions of memcpy and memmove

AtariDreams:
Memmove is a valid way to implement memcpy. However, memcpy isn’t a valid way to implement memmove. 

For this reason, I switched them around. No ABI breakage should occur

### DIFF
--- a/lib/libc/string/bcopy.c
+++ b/lib/libc/string/bcopy.c
@@ -40,7 +40,7 @@ __FBSDID("$FreeBSD$");
 
 #include <sys/types.h>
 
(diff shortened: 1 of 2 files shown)


AtariDreams:
@emaste thoughts?

bsdimp:
This change should have more discussion around it. Since none has happened here in the 18 months it's been open, I'm going to close this request. Please discuss on hackers@freebsd.org or arch@freebsd.org to get consensus for this change in the community, and to get better review of the actual changes.

PR #507, opened 2021-07-07
Use unsigned char * to represent caddr_t [DO NOT MERGE]

gAlfonso-bit:
40 years ago, when Unix was first being developed, way before the C language was even standardized, a type needed to exist for a general core pointer. 

This pointer needed to represent the smallest unit of RAM. The caddr_t type was thus created.

However, it’s intended purpose was a core pointer, not explicitly a char pointer, and multiple standard c functions such as mmap use void * instead of caddr_t. For this reason, mostly because void* supersedes caddr_t, it would be a better typedef than char*, especially considering caddr_t isn’t used in modern code anymore in favor of void*.

This isn’t a change to be made lightly, so I’m labeling it as Do Not Merge for the time being, as this seems to be something that should have the opinion and scrutiny of all maintainers

### DIFF
--- a/sys/sys/types.h
+++ b/sys/sys/types.h
@@ -73,8 +73,8 @@ typedef	__uint64_t	u_quad_t;	/* quads (deprecated) */
 typedef	__int64_t	quad_t;
 typedef	quad_t *	qaddr_t;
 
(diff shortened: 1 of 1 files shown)


gAlfonso-bit:
Closing until further notice

PR #510, opened 2021-07-16
NanoBSD: Fix the pattern for changing PermitRootLogin

jlduran:
The previous pattern had the (harmless) side-effect of duplicating the entry in the config file.

### DIFF
--- a/tools/tools/nanobsd/defaults.sh
+++ b/tools/tools/nanobsd/defaults.sh
@@ -728,7 +728,7 @@ cust_comconsole ( ) (
 # Allow root login via ssh
 
 cust_allow_ssh_root ( ) (
-	sed -i "" -e '/PermitRootLogin/s/.*/PermitRootLogin yes/' \
+	sed -i "" -E 's/^#?PermitRootLogin.*/PermitRootLogin yes/' \
 	    ${NANO_WORLDDIR}/etc/ssh/sshd_config
 )
 


jlduran:
I see cross-builds are failing for macOS here on GitHub. This is unrelated to this PR.

I guess the cross-bootstrap-tools.yml file needs to be updated now that macOS instances are using clang-12, but @uqs or @arichardson will know better.

bsdimp:
> I see cross-builds are failing for macOS here on GitHub. This is unrelated to this PR.

Yes. I'd planned on testing this by generating a bootable nanobsd image, since there's no way it could break other things.

> I guess the cross-bootstrap-tools.yml file needs to be updated now that macOS instances are using clang-12, but @uqs or @arichardson will know better.

They will. I'll let them sort it out.

arichardson:
> I see cross-builds are failing for macOS here on GitHub. This is unrelated to this PR.
> 
> I guess the cross-bootstrap-tools.yml file needs to be updated now that macOS instances are using clang-12, but @uqs or @arichardson will know better.

I noticed that in CheriBSD too yesterday. It appears that homebrew now customizes the lld version string so bsd.linker.mk no longer detects the linker as lld. Should be trivial to fix but I am away until Monday.

jlduran:
> Yes. I'd planned on testing this by generating a bootable nanobsd image, since there's no way it could break other things.

Nice!

_Wishful thinking:_ @bsdimp is going to generate a bootable nanobsd image using GitHub actions for testing... :star_struck:

bsdimp:
Landed as 73358cd2cfccf9aafde069ffcb8185a0f3420ffd

PR #511, opened 2021-07-16
NanoBSD/rescue: Update OpenSSH configuration files

jlduran:


bsdimp:
silly question... is there some way to just copy these from the base system instead of keeping our own, gotta update but never do files?

jlduran:
Not sure, since this is essentially an overlay directory, I would guess duplication is desired.

But I see your point, next in line is `/etc/ttys`, for example, (although I'm afraid to submit a patch for that one).

I might have a few ideas for NanoBSD in general that could help consolidating some of these issues. But I am waiting on poudriere-image changes to avoid duplicate work.

bsdimp:
Will do ttys in a separate commit.
Closed in 9f6c794ee2ed91f65b570176d3eb729777817bc4
Went ahead and did the update to the OpenSSH in the interim as well, as a separate commit.

PR #512, opened 2021-07-20
Convert isipv6 to bool since it is used as one

AtariDreams:
I also removed an unused hw_tls struct member that I found was completely unused.

### DIFF
--- a/sys/netinet/tcp_output.c
+++ b/sys/netinet/tcp_output.c
@@ -229,9 +229,7 @@ tcp_default_output(struct tcpcb *tp)
 #endif
 #ifdef INET6
 	struct ip6_hdr *ip6 = NULL;
-	int isipv6;
-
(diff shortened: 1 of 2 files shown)


sys/netinet/tcp_stacks/tcp_rack.h
```diff
@@ -340,9 +340,6 @@ struct rack_fast_send_blk {
 	struct udphdr *udp;
 	struct mbuf *m;
 	uint32_t o_m_len;
-	uint32_t rfo_apply_push : 1,
```
bsdimp, line 343: These are still used in the tree.

gAlfonso-bit: Alright

gAlfonso-bit:
Fixed

emaste:
headline says `hw_tls` but the change is to `isipv6`

PR #515, opened 2021-07-23
Use C++ headers and casting instead of C in C++ files

AtariDreams:
C++ gives us a lot of features like C++ headers and booleans. Let’s use that instead of integers in C++ code.

Redundant “extern C” statements were also removed since those header files themselves have macros and statements for “extern C”


emaste:
CI builds are currently broken due to 6cafdee, although it looks like it is failing earlier due to a bug in this set of changes

PR #517, opened 2021-07-27
Update ftpd.c to use va_end()

Coldfirex:
va_start opened but not closed by va_end()

pointed out by cppcheck


Coldfirex:
For bug report: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=257317

asomers:
Acutally, that function is dead code.  It's been #ifdef'd out ever since SVN r13139 in 1996.  I'll just delete the whole thing.

PR #519, opened 2021-07-31
Use C++ features and headers in C++ code instead of C

AtariDreams:
We don’t need to use integers as booleans or the C standard headers when we have the C++ standard headers and booleans

### DIFF
--- a/lib/csu/tests/cxx_constructors.cc
+++ b/lib/csu/tests/cxx_constructors.cc
@@ -35,25 +35,25 @@ __FBSDID("$FreeBSD$");
 #include <sys/types.h>
 #include <sys/wait.h>
 
-#include <errno.h>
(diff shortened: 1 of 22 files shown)


supahero1:
lib/csu/tests/cxx_constructors.cc:56
`1` substituted for `false`?

bsdimp:
Too many changes in this to effectively review. Please split it up into logical changes.

bsdimp:
This pull request is too wide-ranging. The changes are likely be good, but it's impossible to land as is. Please create a new pull request per directory as there are different reviewers needed for different parts of the tree.  Some of the changes add std:: (which is likely right, but not mentioned) and noexcept (which may or may not be right, I didn't study the code to know). at the very least, this information should be reflected i the commit messages. Thanks.

PR #523, opened 2021-08-07
Don't reset aac_reset_adapter on timedout

mdedetrich:
Credit for this change goes to `SilverNetworks` from https://www.truenas.com/community/threads/idle-reboot-bug-aacraid0-command-0xfffffe00007bdfa0-timeout-after-3857-seconds-shutting-down-controller-done.86761/

The premise of the problem is that if for some reason a disk device behind the aacraid driver isn't sending/receiving any IO, then the `timedout` condition triggers and the call to `aac_reset_adapter(sc)` causes the FreeBSD system to lock up.

This leads to a lot of unintentional behavior, for example in my personal usecase a hard drive sitting behind this raid controller failed a S.M.A.R.T. check. Since the S.M.A.R.T check failed FreeBSD stopped writing/reading data from the disk (as it should) however this invariably trigger the `timedout` which then caused the system to hang. You experience the same problem if for whatever other reason you don't write/read from a disk behind the raid controller (i.e. you have some disks sitting in the system and they aren't mounted but still recognized).

### DIFF
--- a/sys/dev/aacraid/aacraid.c
+++ b/sys/dev/aacraid/aacraid.c
@@ -2184,8 +2184,13 @@ aac_timeout(struct aac_softc *sc)
 		}
 	}
 
-	if (timedout) 
-		aac_reset_adapter(sc);
+    if (timedout)
+        /*
+         * Resetting the adapter causes FreeNAS 11.x/12.x to panic with a "command not in queue" error.
+         * Let's not reset the adapter and carry on as if nothing happened.
+         * This is the command we are leaving out:
+         * aac_reset_adapter(sc);
+         */
 	aacraid_print_queues(sc);
 }
 


sys/dev/aacraid/aacraid.c
```diff
@@ -2184,8 +2184,13 @@ aac_timeout(struct aac_softc *sc)
...
 
-	if (timedout) 
-		aac_reset_adapter(sc);
+    if (timedout)
+        /*
+         * Resetting the adapter causes FreeNAS 11.x/12.x to panic with a "command not in queue" error.
+         * Let's not reset the adapter and carry on as if nothing happened.
+         * This is the command we are leaving out:
+         * aac_reset_adapter(sc);
+         */
```
bsdimp, line 2193: Wouldn't it be better to fix the state machine that leads aac to think a command isn't in the queue to guard against leaks and such when there's a timeout?

mdedetrich: Definitely agreed however this is a lot more work to do correctly and with the current state of the driver (i.e. its basically freezing FreeBSD whenever data is not being written/read from the drive for a period of time) at least in my opinion it makes sense to disable this now while working on a complete fix in the future.

What are your thoughts on this?

mdedetrich: I have been looking at the code and another thing to add is that according to the original author of the driver (which since has been untouched), its okay if the `timedout` doesn't even execute, i.e.

https://github.com/freebsd/freebsd-src/blob/main/sys/dev/aacraid/aacraid.c#L1040-L1046

At least to me this indicates that disabling the `aac_reset_adapter(sc)` in this specific circumstance is not any less safe than the current situation.

> Wouldn't it be better to fix the state machine that leads aac to think a command isn't in the queue to guard against leaks and such when there's a timeout?

I can definitely say that in my case (and in the other case mentioned in the forum above), if you disable  `aac_reset_adapter(sc)` there is no memory leak, talking about systems that have run for months here. What I didn't mention before is that at least in my case the card is being run in HBA mode (i.e. all RAID/smart cache etc etc functionality is disabled and the card is just exposing the hardrives as a JBOD to the system) so maybe a better way forward is to detect this mode on the card and don't call `aac_reset_adapter(sc)` if you are running in HBA mode?

bsdimp: Looking at the original link, we have a command that's been queued that's on the 'busy' queue, but it isn't flagged as being on the busy queue. If you don't call `aac_reset_adapter` how does it get off the busy queue? the bit is set when it's enqueued by `aac_enqueue_busy()`. What happens to that request? Why did that request timeout? Is there insufficient locking between the timeout routine and the ISR that would normally process this?

Looking at the code, the panic is coming form the ISR while we're resetting the controller. This suggests to me that the 'wedged' command is canceled in the aac's firmware and is reporting a completion to the ISR. The ISR looks like it's doing an `aac_remove_busy` (the source of the panic) after `aac_reset_adapter` code has already removed it from the busy list. This suggests the real bug here may be a mismatch between the driver's expectation for these busy commands and the firmware's. This is resulting in the mismatch. So we complete it once in aac_reset_adapter, then we get an interrupt telling us the command is complete and try to process it again. Only now the command is who knows what because the first time it was completed it was returned to the pool. So the panic is catching a real bug.

Now, what to do about it. This only fires after a really long time. 3121 seconds in the reported case. So why is that
command wedged for the better part of an hour? And why didn't it trigger after 180 seconds like it's supposed to. The aac_timeout function is only called from the aac_command_thread which only calls it if we've not got a command
ack back after 20 seconds (so it waits for 20s of idle, which likely is a bit long. If this is an easy panic to reproduce, then changing AAC_PERIODIC_INTERVAL from 20 in aacraid_var.h to 1 and seeing if that changes behavior (I think it may only make the time it takes to timeout closer to 180s, but I could be wrong). The other thing you could try is to #if 0 out the the "Abort all pending commands" while loop in aac_reset_adapter which would 'trust' the firmware will return completion records for them eventually. This may also solve the problem while still allowing the other cleanup to happen.

So I believe you that this "fixes" things for you, but over time those wedged commands will add up, stealing command slots which will reduce the number of concurrent I/Os that can be done.

bsdimp:
https://people.freebsd.org/~imp/aacraid.diff has the above instructions reduced to a patch :)

bsdimp:
Had anybody tried my alternative to see if that also fixes the problem? It's more consistent with the underlying problem and should solve the issue as well...

mdedetrich:
I commented in the forum thread, I would like to do this but there is no documentation on how to make a TrueNAS build with a patched kernel so I created a ticket on this.

Once I figure this out I will test it on my systems

sys/dev/aacraid/aacraid.c
bsdimp: Since there's no new information, I can only walk away from this, or provide the following bandaid
```
diff --git a/sys/dev/aacraid/aacraid.c b/sys/dev/aacraid/aacraid.c
index 51a633e70515..1df9ab44fb78 100644
--- a/sys/dev/aacraid/aacraid.c
+++ b/sys/dev/aacraid/aacraid.c
@@ -227,6 +227,10 @@ MALLOC_DEFINE(M_AACRAIDBUF, "aacraid_buf", "Buffers for the AACRAID driver");
 /* sysctl node */
 SYSCTL_NODE(_hw, OID_AUTO, aacraid, CTLFLAG_RD | CTLFLAG_MPSAFE, 0,
     "AACRAID driver parameters");
+static bool aacraid_reset_on_timeout = true;
+SYSCTL_BOOL(_hw_aacraid, OID_AUTO, reset_on_timeout, CTLFLAG_RWTUN,
+    &aacraid_reset_on_timeout, 0,
+    "Reset the controller on timeout");
 
 /*
  * Device Interface
@@ -2184,7 +2188,7 @@ aac_timeout(struct aac_softc *sc)
                }
        }
 
-       if (timedout) 
+       if (timedout && aac_reset_on_timeout)
                aac_reset_adapter(sc);
        aacraid_print_queues(sc);
 }
```
which creates a tunable `hw.aacraid.reset_on_timeout` that can be set to '0' to disable things.

This should apply back to at least FreeBSD 12. 

I don't run FreeNAS, so I can't help on build issues there...

bsdimp:
OK. After reading the thread followups, I'm just going to close this. It's been too long, nobody cares to even try the patch I came upwith.

mdedetrich:
@bsdimp I am sorry for letting this go. I am honestly not that familiar with TrueNAS, there wasn't any documentation on how to run a version of TrueNAS with a patched kernel and I tried to create a ticket for it but I didn't get the most welcoming answers.

Ontop of that I have very little time, I am maintaining a lot of open source projects including a massive non trivial one. Maybe at some point in the future when I have more capacity I can look into it but I apologize for not being able to test it.

PR #524, opened 2021-08-11
rc: Fix firstboot fs mount logic

jlduran:
This change was introduced in 40adda8665bbcda1bad7e3a6a3c1031027faf39a, and partially fixed in 1ce07411fae8986f3acc293c9fb9ebee5211056a.

Restore the ability to remove the `firstboot_sentinel` file on read-only mounted file systems.


jlduran:
Move to phabricator https://reviews.freebsd.org/D34166.

PR #529, opened 2021-08-20
crypto: fix MAC null check

simran-kathpalia:
Fixed an uninitialized memory bug in crypto device driver reported by syzkaller. Added a check if MAC(digest) is provided by userspace or not and errors out if no MAC is provided but hash and CRYPTO_OP_VERIFY_DIGEST is specified.
To be reviewed by @markjdb


markjdb:
I committed a derivative of this patch here: https://cgit.freebsd.org/src/commit/?id=7c2f227a17ded0934c5941c7911797edb7d770a2

Thanks!

PR #531, opened 2021-08-30
GitHub: Fix cross-builds

jlduran:
It's time to bother (bi-weekly) @arichardson for GitHub cross-builds, sorry.

This is more an FYI than an actual PR, but this is what I did:

Add `__is_aligned` definition to cross-build's `sys/cdefs.h`

This was introduced in 4e5d32a445f90d37966cd6de571978551654e3f3.

PS: I did not want to hinder @DimitryAndric warning-silencing spree, that's why I decided to wait a while to submit.

### DIFF
--- a/tools/build/cross-build/include/common/sys/cdefs.h
+++ b/tools/build/cross-build/include/common/sys/cdefs.h
@@ -255,3 +255,22 @@
 #define	__BSD_VISIBLE		1
 #define	__ISO_C_VISIBLE		2011
 #define	__EXT1_VISIBLE		1
+
+/* Alignment builtins for better type checking and improved code generation. */
+/* Provide fallback versions for other compilers (GCC/Clang < 10): */
+#if !__has_builtin(__builtin_is_aligned)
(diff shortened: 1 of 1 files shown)


DimitryAndric:
Ah, I wasn't aware of this compat header. Should be no problem to commit this.

arichardson:
Seems reasonable although maybe we should just copy-paste all the alignment macros from cdefs here.

bsdimp:
I wonder if there's not some way that we could bring in the latest sys/cdefs.h entirely, or have a common subset we can share. cdefs.h is in need of a good scrubbing and I'd hate to have to do that multiple places...
Not for this commit, but for future work maybe?

arichardson:
Merged in 94d9439b6be6bd5ef9febfaf38128e0cad91476d, but I don't seem to have permissions to mark this PR as closed.

jlduran:
Thank you! 

> I don't seem to have permissions to mark this PR as closed

I think it’s because you are not a member of FreeBSD on GitHub 🙃.

PR #532, opened 2021-09-01
Update if_ether.c

orange30:
Free memory before return from arprequest_internal.


bsdimp:
This looks sane to me.

orange30:
I am reading the code of arp, when I read the function in_arpinput, I find it also need free mbuf before return. @bsdimp  Please help to check it, thanks.

markjdb:
Looks ok. It appears that in_arpinput() has the same bug: if arp_fillheader() fails, it returns without a `goto drop`.

orange30:
See th pr #534 please.

PR #533, opened 2021-09-01
Fix some improper locking bugs

ycaibb:
Fix some improper locking bugs.

Signed-off-by: Ryan Cai <ycaibb@gmail.com>

### DIFF
--- a/sys/net/route/route_ctl.c
+++ b/sys/net/route/route_ctl.c
@@ -266,6 +266,8 @@ get_info_weight(const struct rt_addrinfo *info, uint32_t default_weight)
 	/* Keep upper 1 byte for adm distance purposes */
 	if (weight > RT_MAX_WEIGHT)
 		weight = RT_MAX_WEIGHT;
+	else if (weight == 0)
+		weight = default_weight;
 
 	return (weight);
 }


ycaibb:
Hi, I found some errors(could be small) due to the unreleased locks. Thank you very much for your checking!

bsdimp:
> Hi, I found some errors(could be small) due to the unreleased locks. Thank you very much for your checking!

These all look to be locked on exit bugs. How'd you find those: that class of bugs don't tend to cause problems.

Also, the ones in contrib or sys/contrib likely should be resolved with our upstream first, which may mean splitting this PR if they have vastly different response times.

ycaibb:
@bsdimp OK, let me remove the patches in the contrib or sys/contrib. I will also report this to the upstream.

They are indeed not harmful bugs since the program exits and all of the program resources will be cleaned up by some common OS( other OS systems (e.g., embedded systems) do not automatically free some resources at the exit).
On the other hand, there is no harm to write unlock statement before the program exit for better resource management and for good code symmetry. Also, taking locks before program exit might slow down termination.

ycaibb:
@bsdimp @bsdjhb 
Can you take a look？ Thank you very much.

PR #534, opened 2021-09-01
 update if_ether.c: squash commit 6e49ae2 and c6e3c75 into one

duzhikuo:
Free memory before return from arprequest_internal.
In in_arpinput(), if arp_fillheader() fails, it should use goto drop.

### DIFF
--- a/sys/netinet/if_ether.c
+++ b/sys/netinet/if_ether.c
@@ -418,6 +418,7 @@ arprequest_internal(struct ifnet *ifp, const struct in_addr *sip,
 	linkhdrsize = sizeof(linkhdr);
 	error = arp_fillheader(ifp, ah, 1, linkhdr, &linkhdrsize);
 	if (error != 0 && error != EAFNOSUPPORT) {
+		m_freem(m);
 		ARP_LOG(LOG_ERR, "Failed to calculate ARP header on %s: %d\n",
 		    if_name(ifp), error);
 		return (error);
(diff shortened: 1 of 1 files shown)


emaste:
FWIW you can update an existing pull request (e.g., with a squashed commit)

orange30:
> FWIW you can update an existing pull request (e.g., with a squashed commit)

Thanks a lot，I will use the squashed commit next time. :blush:

bsdimp:
This looks better than #532 in that it is more complete :).

AlexanderChernikov (approved):
LGTM.

orange30:
If we set net.link.ether.inet.proxyall=1, the ip on lo0 can reply arp; but if we donot set proxyall to be 1, it can not relay arp for this ip and we cannot use this ip.
Sometimes we need set the ip on lo0 and want to use it, but we donot need the function of arp proxy.
Is it a problem? 
@bsdimp @AlexanderChernikov @markjdb

markjdb:
I committed the fix for the memory leaks, thanks.

I'm not sure about your proxyall question. It is not clear to me how you are using lo0. I suggest asking on the freebsd-net mailing list: https://lists.freebsd.org/subscription/freebsd-net .

duzhikuo:
> I committed the fix for the memory leaks, thanks.
> 
> I'm not sure about your proxyall question. It is not clear to me how you are using lo0. I suggest asking on the freebsd-net mailing list: https://lists.freebsd.org/subscription/freebsd-net .

Ok, thanks！

PR #536, opened 2021-09-07
top(1): indicate how to reset grep string

jgrafton:
Indicate how to reset grep string during a search.


bsdimp:
No Brainer Yes. Committed and queued for mfc to 12/13 4ccbbe5f0689

bsdimp:
(as well as PR 479 which had dropped off my radar)

PR #537, opened 2021-09-10
Fix some improper locking bugs

ycaibb:
Fix two improper locking bugs.

Signed-off-by: ycaibb <ycaibb@gmail.com>


ycaibb:
I removed the patches in the contrib or sys/contrib. Thank you very much for your checking! Update of #533

PR #539, opened 2021-09-17
e1000: Sync DPDK ich8lan

kev009:
This contains a probable regression for suspend/resume and other issues on I219, break it out so users can help test.

Also of note, I left out two changes to the FreeBSD driver to try and opportunistically use ULP https://cgit.freebsd.org/src/diff/sys/dev/e1000/if_em.c?id=a4378873e9ce1b35b55378c21f8eae69e58c2525

@ricera

### DIFF
--- a/sys/dev/e1000/if_em.c
+++ b/sys/dev/e1000/if_em.c
@@ -1111,9 +1111,6 @@ em_if_attach_pre(if_ctx_t ctx)
 
 	em_print_fw_version(sc);
 
-	/* Disable ULP support */
-	e1000_disable_ulp_lpt_lp(hw, true);
-
 	/*
 	 * Get Wake-on-Lan and Management info for later use
 	 */
--- a/sys/dev/e1000/e1000_ich8lan.c
+++ b/sys/dev/e1000/e1000_ich8lan.c
@@ -72,6 +72,7 @@
 
 #include "e1000_api.h"
 
+static s32  e1000_oem_bits_config_ich8lan(struct e1000_hw *hw, bool d0_state);
 static s32  e1000_acquire_swflag_ich8lan(struct e1000_hw *hw);
 static void e1000_release_swflag_ich8lan(struct e1000_hw *hw);
 static s32  e1000_acquire_nvm_ich8lan(struct e1000_hw *hw);
@@ -330,7 +331,9 @@ static s32 e1000_init_phy_workarounds_pchlan(struct e1000_hw *hw)
 	 * so forcibly disable it.
 	 */
 	hw->dev_spec.ich8lan.ulp_state = e1000_ulp_state_unknown;
-	e1000_disable_ulp_lpt_lp(hw, TRUE);
+	ret_val = e1000_disable_ulp_lpt_lp(hw, true);
+	if (ret_val)
+		ERROR_REPORT("Failed to disable ULP\n");
 
 	ret_val = hw->phy.ops.acquire(hw);
 	if (ret_val) {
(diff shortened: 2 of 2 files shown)


kev009:
@ricera is there any way to get guidance from intel on this file?  The datasheets for the client parts are not very enlightening and it's hard to tell whether FreeBSD or OOT or DPDK is authoritative WRT clients.

ricera:
> @ricera is there any way to get guidance from intel on this file? The datasheets for the client parts are not very enlightening and it's hard to tell whether FreeBSD or OOT or DPDK is authoritative WRT clients.

Basically not from within because of its age; but have you looked for the "software developers manuals" instead for those parts? In the past, that contained the stuff that was interesting to drivers, and that stuff wasn't included in the datasheets. Here's one for some old parts, but I don't know exactly what's in ich8lan: https://www.intel.com/content/dam/www/public/us/en/documents/manuals/pcie-gbe-controllers-open-source-manual.pdf

ricera:
For ICH8/9/10, it looks like you'd need "Intel® ICH 8/9/10 and 82566/82567/82562V: Developers Manual", but I can't get a link to it atm

kev009:
> For ICH8/9/10, it looks like you'd need "Intel® ICH 8/9/10 and 82566/82567/82562V: Developers Manual", but I can't get a link to it atm

The problem in particular is with I218/I219 ULP there isn't really documentation on this stuff.  I can't tell if the FreeBSD or DPDK code is newer in e1000_enable_ulp_lpt_lp().

ricera:
> > For ICH8/9/10, it looks like you'd need "Intel® ICH 8/9/10 and 82566/82567/82562V: Developers Manual", but I can't get a link to it atm
> 
> The problem in particular is with I218/I219 ULP there isn't really documentation on this stuff. I can't tell if the FreeBSD or DPDK code is newer in e1000_enable_ulp_lpt_lp().

I would assume the DPDK version is newer, and I also think we might have excluded ULP stuff in the FreeBSD driver.

kev009:
I was able to find it on archive.org http://web.archive.org/web/20170226164709/https://www.intel.com/content/dam/doc/manual/i-o-controller-hub-8-9-10-82566-82567-82562v-software-dev-manual.pdf

This will be useful but doesn't seem to cover the LPT+ chipsets.

ricera:
Ok, I compared the freebsd in-kernel version with the DPDK and internal versions of the file, and it definitely looks like the DPDK version matches the internal one more closely; it's probably the more definitive file to use.

As for I219-ish documentation, that might be when Intel stopped publishing external software documentation for them, I dunno. I have an internal one, but I haven't found any external-facing one yet.

sys/dev/e1000/if_em.c
```diff
@@ -1093,9 +1093,6 @@ em_if_attach_pre(if_ctx_t ctx)
 		}
 	}
 
-	/* Disable ULP support */
-	e1000_disable_ulp_lpt_lp(hw, true);
```
kev009, line 1097: This is redundant, it is called in ich8lan setup via  e1000_setup_init_funcs() above.

kev009:
I've got this working on I219 now with https://reviews.freebsd.org/D32087(!)

kev009:
@ricera I feel like I still don't have enough information to make any decisions on the majority of the changes

In particular, this changeset results in:
- a hardware initialization failure message when booting I219 (2) without a cable/link
- When AMT is enabled on the system firmware, the link flaps regularly.  This does not happen without this changeset.

The LSC to CSC change is interesting, Linux and any intel.com Linux and BSD drivers seem to not carry it.

kev009:
After much trial and error the link flaps appear to be related to a particular old cable.  Will need to trace through the shared code to see what is causing the hw init failure message.

bsdimp:
This branch looks like it may no longer be relevant. Is it still active? Or can it be closed?

kev009:
Committed relevant parts as d36fbdb08ff5

PR #540, opened 2021-09-23
e1000: fix I219 hang on reset

kev009:
Clear the rings before reset to avoid a HW hang.

Inspired by em-7.7.8 and DPDK (1fc9701238edcf0541289b9ae15565b6d9d7ab30)

@ricera

### DIFF
--- a/sys/dev/e1000/if_em.c
+++ b/sys/dev/e1000/if_em.c
@@ -37,7 +37,8 @@
 /*********************************************************************
  *  Driver version:
  *********************************************************************/
-char em_driver_version[] = "7.6.1-k";
+char em_driver_version[] = "7.7.8-fbsd";
+char igb_driver_version[] = "2.5.19-fbsd";
 
 /*********************************************************************
  *  PCI Device ID Table
@@ -339,6 +340,7 @@ static int	em_get_regs(SYSCTL_HANDLER_ARGS);
 
 static void	lem_smartspeed(struct e1000_softc *);
 static void	igb_configure_queues(struct e1000_softc *);
+static void	em_flush_desc_rings(struct e1000_softc *);
 
 
 /*********************************************************************
@@ -588,7 +590,7 @@ static struct if_shared_ctx igb_sctx_init = {
 	.isc_ntxqs = 1,
 	.isc_admin_intrcnt = 1,
 	.isc_vendor_info = igb_vendor_info_array,
-	.isc_driver_version = em_driver_version,
+	.isc_driver_version = igb_driver_version,
 	.isc_driver = &igb_if_driver,
 	.isc_flags = IFLIB_NEED_SCRATCH | IFLIB_TSO_INIT_IP | IFLIB_NEED_ZERO_CSUM,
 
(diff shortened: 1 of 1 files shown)


sys/dev/e1000/if_em.c
```diff
@@ -2461,6 +2468,113 @@ igb_init_dmac(struct adapter *adapter, u32 pba)
...
+	struct e1000_hw		*hw = &adapter->hw;
+	struct tx_ring		*txr = &adapter->tx_queues->txr;
+	struct e1000_tx_desc	*txd;
+	u32			tctl, txd_lower = E1000_TXD_CMD_IFCS;
+	u16			size = 512;
+
+	tctl = E1000_READ_REG(hw, E1000_TCTL);
+	E1000_WRITE_REG(hw, E1000_TCTL, tctl | E1000_TCTL_EN);
+
+	txd = &txr->tx_base[txr->tx_cidx_processed];
```
kev009, line 2494: Not sure this is correct, although it seems like we just need a scratch descriptor prior to the head.

sys/dev/e1000/if_em.c
```diff
@@ -2461,6 +2468,113 @@ igb_init_dmac(struct adapter *adapter, u32 pba)
...
+
+	/* Just use the ring as a dummy buffer addr */
+	txd->buffer_addr = txr->tx_paddr;
+	txd->lower.data = htole32(txd_lower | size);
+	txd->upper.data = 0;
+
+	/* flush descriptors to memory before notifying the HW */
+	wmb();
```
kev009, line 2502: These membars are from the OOT driver and are macro'd out to nothing in e1000_osdep.h,

sys/dev/e1000/if_em.c
```diff
@@ -37,7 +37,8 @@
 /*********************************************************************
  *  Driver version:
  *********************************************************************/
-char em_driver_version[] = "7.6.1-k";
+char em_driver_version[] = "7.7.8-fbsd";
+char igb_driver_version[] = "2.5.19-fbsd";
```
kev009, line 41: Incrementing these to avoid confusion in users; we are roughly on par with these OOT versions.

bsdimp:
what's the status of this pull request? IT looks like it's stalled and hasn't landed in main yet.

kev009:
@bsdimp it needs review, hopefully from intel

sys/dev/e1000/if_em.c
```diff
@@ -588,7 +590,7 @@ static struct if_shared_ctx igb_sctx_init = {
 	.isc_ntxqs = 1,
 	.isc_admin_intrcnt = 1,
 	.isc_vendor_info = igb_vendor_info_array,
-	.isc_driver_version = em_driver_version,
+	.isc_driver_version = igb_driver_version,
```
ricera, line 593: that's a good catch

sys/dev/e1000/if_em.c
```diff
@@ -1907,6 +1909,10 @@ em_if_stop(if_ctx_t ctx)
 
 	INIT_DEBUGOUT("em_if_stop: begin");
 
+	/* I219 needs special flushing to avoid hangs */
+	if (sc->hw.mac.type >= e1000_pch_spt && sc->hw.mac.type < igb_mac_min)
```
ricera, line 1913: Looking here: https://git.dpdk.org/dpdk/tree/drivers/net/e1000/em_ethdev.c#n728, this check was updated to apply this workaround to cnp in addition to spt, but I see that the device ID list only goes up to (7) for DPDK, and not up to the current (23) that we're up to, so maybe this check doesn't need to change and we should apply it to everything spt and newer.

ricera (approved):
I guess it's alright? It looks like the workaround code has been in the OOT driver since 2015, but somehow didn't get upstreamed, or got removed?

kev009:
Merged as ae1dca798e0f826de46f4ec11914ba4c91928d7a

bsdimp:
thanks @kev009 for closing these

PR #541, opened 2021-09-28
Fix bug in syslogd that caused -N to drop SecureMode if specified after -s

JFranklin13:
This commit should fix the unintuitive behavior where specifying "-ss -N" as opts would drop SecureMode to 1. It might not be the typical use case but I think it's more in line with what a user would expect, plus it's more sane as a security feature being impacted by the order of args looks bad.


markjdb:
Committed, thanks.

PR #545, opened 2021-10-02
Fix a typo

tomhukins:


bsdimp:
This looks good. I'll merge it, but in the future using something more descriptive like "Spell BeagleBone correctly"  that's a little more specific would be good for a single file error like this.

bsdimp:
Pushed in as 202692b1a733

PR #548, opened 2021-10-05
Remove obsolete amd(8) rc.conf configuration

tomhukins:
The script that used these was removed in 13f7dbe822d5f along with amd itself.


emaste:
I'll grab this, and update `.Dd` also

PR #554, opened 2021-10-10
Prune old support

bsdimp:
This is mostly a test of our current CI stuff...

### DIFF
--- a/Makefile.inc1
+++ b/Makefile.inc1
@@ -682,8 +682,8 @@ PACKAGE=	kernel
 
 BOOTSTRAPPING?=	0
 # Keep these in sync
-MINIMUM_SUPPORTED_OSREL?= 1002501
-MINIMUM_SUPPORTED_REL?= 10.3
+MINIMUM_SUPPORTED_OSREL?= 1102501
+MINIMUM_SUPPORTED_REL?= 11.3
 
 # Common environment for world related stages
 CROSSENV+=	\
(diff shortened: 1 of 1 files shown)


bsdimp:
Pushed as a420a672bc53 since this built

arichardson:
I'm pretty sure these are needed for Linux+macos builds since the host tools don't support all calls used in the build. We should really add a full cross-build buildworld CI.

bsdimp:
> I'm pretty sure these are needed for Linux+macos builds since the host
> tools don't support all calls used in the build. We should really add a
> full cross-build buildworld CI
>
 What's a good test then? Full buildworld on Linux? I thought that was part
of the build on github already...

Warner

arichardson:
I only enabled buildkernel since buildworld took quite a long time on the GitHub hosted builders. I haven't tried recently but it might make sense to add buildworld (-DWITHOUT_TOOLCHAIN to avoid building LLVM) to the GitHub action to see how long it takes.

bsdimp:
OK. I was mistaken about what was being tested and showing green. I'll toss
a tree onto one of my ubuntu VMs to build before I commit again

emaste:
We could add a full cross-build test in Cirrus-CI if we want, but would want to make it on demand (as with the arm64 smoke test) to avoid consuming too much of Cirrus-CI's free resources.

PR #556, opened 2021-10-16
g_class.config is not used

wy-chung:
Looks like g_class.config is not used. Should it be removed?


bsdimp:
These two commits should have been submitted squashed. Please do that next time.
Otherwise, this is a good change. config appears to be write only.
Tweaked the commit message a little.
landed as 9cb485d18f01900800efe4ec3cfcfadd335da858

PR #557, opened 2021-10-17
Change the name of the second parameter of biowait to wmesg

wy-chung:
The second parameter of biowait is actually wait message instead of wait channel. The wait channel is the first parameter.


bsdimp:
8587d752558c landed this
I had to tweak the commit message. github doesn't use the pull request's title for any of the commit messages, so when I cherry-pick the change, it just had the message N/A. Not a big deal, but something to consider in the future.

PR #562, opened 2021-11-17
Coretemp prochot is not critical

sadaszewski:
According to the [Intel manual](https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-3b-part-2-manual.pdf) the Thermal Status (0) and Thermal Status Log (1) bits report only a high temperature on the CPU, not a *critical* temperature as suggested in the coretemp driver. I would remove the warning or check for the actual Critical Temperature Log (5) bit. Looking at the Log bit might be better than Critical Temperature Status (4), as AFAIK the critical temperature waives guarantees of correct function, therefore the CPU could have for example written some wrong values into memory at that point and the OS should be stopped ASAP as the state is no longer reliable.

### DIFF
--- a/sys/dev/coretemp/coretemp.c
+++ b/sys/dev/coretemp/coretemp.c
@@ -53,7 +53,6 @@ __FBSDID("$FreeBSD$");
 
 #define	TZ_ZEROC			2731
 
-#define THERM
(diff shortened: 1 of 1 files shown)


bsdimp:
pushed as 8362905cb69e

PR #563, opened 2021-11-17
Generate freebsd32 sysent files from the default ABI

brooksdavis:
This set of changes allows freebsd32 generated files to be generated from sys/kern/syscalls.master. It's an extend set of changes based my work in CheriBSD to support freebsd64. makesyscalls.lua learns about pad registers for 64-bit arguments on 32-bit systems as well as about various ABI changes due to long, pointer, and time_t types. I annotate syscalls.master with a new set of _Contains_ attributes indicating which types the object a pointer points to contains.

The changes for the pads in *systrace_args.c in "freebsd32: generate splits of 64-bit arguments" are likely the most controversial. There are a number of options for this, but the current one should work and have performance identical to the previous one at -O2.

The final regen isn't quite a no-op, but I think it's close enough.


sys/sys/sysproto.h
```diff
@@ -2476,7 +2476,7 @@ struct freebsd4_sigaction_args {
 	char oact_l_[PADL_(struct sigaction *)]; struct sigaction * oact; char oact_r_[PADR_(struct sigaction *)];
 };
 struct freebsd4_sigreturn_args {
-	char sigcntxp_l_[PADL_(const struct ucontext4 *)]; const struct ucontext4 * sigcntxp; char sigcntxp_r_[PADR_(const struct ucontext4 *)];
+	char sigcntxp_l_[PADL_(const struct freebsd4_ucontext *)]; const struct freebsd4_ucontext * sigcntxp; char sigcntxp_r_[PADR_(const struct freebsd4_ucontext *)];
```
kostikbel, line 2479: This is generated file

sys/kern/systrace_args.c
```diff
@@ -9562,7 +9562,7 @@ systrace_return_setargdesc(int sysnum, int ndx, char *desc, size_t descsz)
 	/* sendto */
 	case 133:
 		if (ndx == 0 || ndx == 1)
-			p = "int";
+			p = "ssize_t";
 		break;
 	/* shutdown */
 	case 134:
```
kostikbel, line 9569: This is generated file. It seems you did not removed it from several commits.

sys/compat/freebsd32/freebsd32_proto.h
```diff
@@ -149,10 +149,10 @@ struct freebsd32_msgsys_args {
...
-	char a2_l_[PADL_(uint32_t)]; uint32_t a2; char a2_r_[PADR_(uint32_t)];
-	char a3_l_[PADL_(uint32_t)]; uint32_t a3; char a3_r_[PADR_(uint32_t)];
-	char a4_l_[PADL_(uint32_t)]; uint32_t a4; char a4_r_[PADR_(uint32_t)];
+	char which_l_[PADL_(int)]; int which; char which_r_[PADR_(int)];
+	char a2_l_[PADL_(int)]; int a2; char a2_r_[PADR_(int)];
+	char a3_l_[PADL_(int)]; int a3; char a3_r_[PADR_(int)];
+	char a4_l_[PADL_(int)]; int a4; char a4_r_[PADR_(int)];
 };
 struct freebsd32_ntp_adjtime_args {
 	char tp_l_[PADL_(struct timex32 *)]; struct timex32 * tp; char tp_r_[PADR_(struct timex32 *)];
```
kostikbel, line 158: Again, the files are generated

kostikbel (commented):
I think this series introduces some rules how types must be named in syscalls.master.  For instance, it requires u_long and not unsigned long, unless I mistaken.  Would be useful to list the allowed types spelling explicitly.

There are several top-level variables that contain list of syscalls like abi_change.  Could you please add comments explaining when to use them?  I mean, when I add a new syscall, how to handle compat32.

Please move all generated files changes to dedicated commits.  In fact, I do not see why not regenerate once at the end of the series.

sys/compat/freebsd32/freebsd32_misc.c
```diff
@@ -2473,6 +2473,105 @@ freebsd11_freebsd32_fhstat(struct thread *td,
...
+	unsigned int :(8 / 2) * (16 - (int)sizeof(struct timespec32));
+	unsigned int :(8 / 2) * (16 - (int)sizeof(struct timespec32));
+};
+
+static void
+freebsd11_cvtnstat32(struct stat *sb, struct nstat32 *nsb)
+{
+	bzero(nsb, sizeof(*nsb));
+	CP(*sb, *nsb, st_dev);
+	CP(*sb, *nsb, st_ino);
```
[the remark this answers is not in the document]
brooksdavis: That wasn't true of `freebsd11_cvtnstat`. My next push will switch `freebsd11_cvtnstat` to call `freebsd11_cvtstat` to avoid duplicate code.

sys/compat/freebsd32/freebsd32_misc.c
```diff
@@ -2473,6 +2473,105 @@ freebsd11_freebsd32_fhstat(struct thread *td,
...
+	fflags_t  st_flags;		/* user defined flags for file */
+	__uint32_t st_gen;		/* file generation number */
+	struct timespec32 st_birthtim;	/* time of file creation */
+	/*
+	 * See comment in the definition of struct freebsd11_stat
+	 * in sys/stat.h about the following padding.
+	 */
+	unsigned int :(8 / 2) * (16 - (int)sizeof(struct timespec32));
+	unsigned int :(8 / 2) * (16 - (int)sizeof(struct timespec32));
+};
```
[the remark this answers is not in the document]
brooksdavis: I've moved it.

sys/i386/include/sigframe.h
```diff
@@ -85,7 +85,7 @@ struct sigframe4 {
 		__siginfohandler_t	*sf_action;
 		__sighandler_t		*sf_handler;
 	} sf_ahu;
-	struct ucontext4 sf_uc;		/* = *sf_ucontext */
+	struct freebsd4_ucontext sf_uc;		/* = *sf_ucontext */
```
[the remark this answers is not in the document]
brooksdavis: Both sigcontext4 and ucontext4 need to be defined in a header for sys/i386/i386/genassym.c. freebsd4_ucontext could probably move to an i386-specific header.

brooksdavis:
I've addressed most comments. I've mostly moved regeneration outside of commits containing changes, but I've kept a number of regen commits.  In particular, any change that results in a broken tree is immediately followed by a regen so `git bisect skip` jumps out of the hole of broken commits.  I've also left ones that are important to see how the changes to makesyscalls.lua affect generated code.

sys/compat/freebsd32/syscalls.conf
```diff
@@ -9,5 +9,51 @@ syscallprefix="FREEBSD32_SYS_"
...
+# sigqueue - union sival requires special handing on big-endian
+# sigreturn - struct osigcontext
+# sysarch - always MD
+# _umtx_lock _umtx_unlock _umtx_op - arguments require special handling
+syscall_abi_change="fcntl ksem_init ksem_open mprotect msgsys semsys shmsys sigqueue sigreturn sysarch _umtx_lock _umtx_unlock _umtx_op"
+
+# System calls that appear to require handling, but don't:
+# aio_cancel - the aiocb pointer is just a virtual address
+# thr_* - the lack of sign extension of long args is OK
+# sctp_generic_* - iovec differencs handled internally
```
kostikbel, line 43: differences

sys/kern/syscalls.master
```diff
@@ -230,26 +258,26 @@
 		int ptrace(
 		    int req,
 		    pid_t pid,
-		    _Inout_opt_ caddr_t addr,
+		    _Inout_opt_ _Contains_ptr_ caddr_t addr,
```
kostikbel, line 261: Why does ptrace addr requires Contrains_ptr annotation?

sys/kern/syscalls.master
```diff
@@ -388,15 +416,15 @@
...
+		    _In_opt_ _Contains_long_ptr_ const struct sigaltstack *ss,
+		    _Out_opt_ _Contains_long_ptr_ struct sigaltstack *oss
 		);
 	}
 54	AUE_IOCTL	STD|CAPENABLED {
 		int ioctl(
 		    int fd,
 		    u_long com,
-		    _Inout_opt_ char *data
+		    _Inout_opt_ _Contains_long_ptr_timet_ char *data
```
kostikbel, line 427: Same question, why Contains is needed for data?

sys/i386/include/sigframe.h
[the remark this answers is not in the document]
kostikbel: Do you plan to, or should I do the move after your series is landed?

sys/kern/syscalls.master
brooksdavis: So the fact that ioctl requires handling is detected by makesyscalls.lua. `data` can and does point to anything.

sys/kern/syscalls.master
brooksdavis: `long` might be more accurate here, but if FreeBSD ptrace add definitely points to things containing pointers.

sys/kern/syscalls.master
brooksdavis: To elaborate, I'd prefer not to require additions to syscalls.conf to indicate that syscalls need ABI-specific handling. Some can't seem to be avoided for freebsd32 (they all happen to be masked on CheriBSD because all syscalls for integer-pointer ABIs require handling.)

sys/kern/syscalls.master
brooksdavis: Upon further reflection `_Contains_long_ptr_` is the correct annotation for addr since it could point to a struct ptrace_io_desc which contains both.

sys/kern/syscalls.master
brooksdavis: I've decided to drop `time_t_`. It seems likely such an ioctl exists, but freebsd32_ioctl only handles ioctl commands that contain longs and pointers and those likely should be pushed deeper into the kernel where there isn't a risk (however small) of the command being conflated with another one.

brooksdavis:
Fixed the comment type and tweaked the `_Contains_` annotations for ioctl and ptrace. I've also updated the author emails on stuff I'd cherry-picked from CheriBSD. Once approved I'll do a pass of adding Reviewed by lines before pushing to main.

sys/kern/syscalls.master
```diff
@@ -49,6 +49,24 @@
 ;		definition to syscall.h besides adding a sysent.
 ;	NOTSTATIC syscall is loadable
 ;	CAPENABLED syscall is allowed in capability mode
+;
+; To support programatic generation of both the default ABI and 32-bit compat
```
kevans91, line 53:
```suggestion
; To support programmatic generation of both the default ABI and 32-bit compat
```

sys/kern/syscalls.master
```diff
@@ -49,6 +49,24 @@
...
+; (freebsd32) we impose a number of restictions on the types of system calls.
```
kevans91, line 54:
```suggestion
; (freebsd32) we impose a number of restrictions on the types of system calls.
```

kevans91:
Just a couple of spelling nits; consider me `Reviewed-by` for the makesyscalls.lua changes (I didn't look too closely at the syscalls.master changes this time outside of the amended description at the top for what makesyscalls.lua is expecting).

Just a general note that I'll likely start refactoring makesyscalls.lua into a libmakesys.lua + makesyscalls.lua driver after this change lands. If there are any Cheri-specific changes adjacent to these that I could make some room for, I'm more than happy to try and accommodate that.

bsdimp:
On Mon, Nov 22, 2021, 12:28 PM Kyle Evans ***@***.***> wrote:

> Just a couple of spelling nits; consider me Reviewed-by for the
> makesyscalls.lua changes (I didn't look too closely at the syscalls.master
> changes this time outside of the amended description at the top for what
> makesyscalls.lua is expecting).
>
> Just a general note that I'll likely start refactoring makesyscalls.lua
> into a libmakesys.lua + makesyscalls.lua driver after this change lands. If
> there are any Cheri-specific changes adjacent to these that I could make
> some room for, I'm more than happy to try and accommodate that
>

I'd love to be able to use it to generate data for bsd-user system call
tracing... so I love that direction

Warner

>

kevans91:
> I'd love to be able to use it to generate data for bsd-user system call tracing... so I love that direction Warner
> […](#)

Yeah, that was the main motivation... I've also idly tossed around the idea of a linter that just inspects the table for some constraints like Brooks is adding regarding types to use and warns. It could probably use cpp internally, even, and sanity check typedefs or other concerns with the generated output.

kostikbel (approved):

brooksdavis:
merged as 88dfcfa2a08a05070767f39160834f5a7277b4d6..b92f4ae58a57fe736a2147662f7e2900b961c202

PR #565, opened 2021-11-21
usb/u3g driver, add support for Quectel EM05

yjqg6666:
Add USB driver support for Quectel EM05. Tested and works like a charm on ppp.


hselasky:
Should be merged now. Please close!

yjqg6666:
@hselasky Would this patch MFC/back-port to Stable/13 ?

hselasky:
Yes, see:

https://cgit.freebsd.org/src/commit/?id=9e89269ee7e5f31aa60ae8c645821d4ca08f5ae6

--HPS

yjqg6666:
Cool！ I would test it later on the latest commit. It works on stable/13 with up-to-yesterday commits and. this patch.

PR #566, opened 2021-11-21
powerpc/aim: ifdef DDB-only pmap information functions

sterlingjensen:
INVARIANTS and DDB can be enabled independently.


bsdimp:
Can I get an email address that isn't @noreply.github.com?

Also, the title of this likely should be:
```
powerpc/aim: ifdef DDB-only pmap information functions
```

sterlingjensen:
The spam isn't worth it and I relinquishment all claims of ownership on the two ifdef statements - I'm equally happy with an uncredited change. The edits flag is check - I welcome its use.

bsdimp:
I'll commit with a 'Submitted by: Sterling Jenson' then w/o an email. Thanks for your contribution!

bsdimp:
47796d47ecec45fab8a0e8357da756554e8272f7 pushed this in.

PR #567, opened 2021-11-23
Assorted syscall.master cleanups

brooksdavis:
This set of changes includes some trivial improvements and fixes I deferred in the recent freebsd32 work, plus changes to completely eliminate the use of syscall aliases (the three items after the `}` in some syscalls.master entries). Fixing syscall(2) and __syscall(2) required a new type in makesyscalls.lua. I've not attempted to remove support for syscall aliases yet. It could potentially be disruptive to downstream consumers and it wasn't immediately obvious where to start pulling on the string to remove it.

### DIFF
--- a/sys/kern/kern_exit.c
+++ b/sys/kern/kern_exit.c
@@ -205,12 +205,12 @@ exit_onexit(struct proc *p)
 /*
  * exit -- death of process.
  */
-void
-sys_sys_exit(struct thread *td, struct sys_exit_args *uap)
+int
+sys_exit(struct thread *td, struct exit_args *uap)
 {
 
 	exit1(td, uap->rval, 0);
-	/* NOTREACHED */
+	__unreachable();
 }
 
 /*
--- a/sys/kern/uipc_syscalls.c
+++ b/sys/kern/uipc_syscalls.c
@@ -78,10 +78,6 @@ static int recvit(struct thread *td, int s, struct msghdr *mp, void *namelenp);
 
 static int accept1(struct thread *td, int s, struct sockaddr *uname,
 		   socklen_t *anamelen, int flags);
-static int getsockname1(struct thread *td, struct getsockname_args *uap,
-			int compat);
-static int getpeername1(struct thread *td, struct getpeername_args *uap,
-			int compat);
 static int sockargs(struct mbuf **, char *, socklen_t, int);
 
 /*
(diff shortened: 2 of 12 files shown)


sys/kern/uipc_syscalls.c
```diff
@@ -1331,21 +1335,19 @@ kern_getsockopt(struct thread *td, int s, int level, int name, void *val,
 	return (error);
 }
 
-/*
- * getsockname1() - Get socket name.
- */
 static int
-getsockname1(struct thread *td, struct getsockname_args *uap, int compat)
+user_getsockname(struct thread *td, int fdes, struct sockaddr *asa,
+    socklen_t *alen, int compat)
```
kostikbel, line 1340: Change compat type to bool, while there?

sys/kern/kern_exit.c
```diff
@@ -205,12 +205,13 @@ exit_onexit(struct proc *p)
...
  */
-void
-sys_sys_exit(struct thread *td, struct sys_exit_args *uap)
+int
+sys_exit(struct thread *td, struct exit_args *uap)
 {
 
 	exit1(td, uap->rval, 0);
 	/* NOTREACHED */
+	__unreachable();
```
kostikbel, line 214: exit1() is declared with __dead2, so this should be excessive.  I think the lint NOTREACHED annotation is useless as well.

kostikbel (approved):

sys/kern/kern_exit.c
brooksdavis: clang objected to the lack of a return(some_int) without this addition. I'd debated with myself about the NOTREACHED and will remove it.

sys/kern/uipc_syscalls.c
brooksdavis: will do

sys/kern/kern_exit.c
bsdimp: We removed lint(1) from the system years ago, and these markers were for lint(1) and a few other older / ancient analysis tools that are no longer relevant. I see no reason to keep it, honestly, except fear of breaking something unknown... At this point, i can't even guess what that might be.

sys/kern/kern_exit.c
```diff
@@ -205,12 +205,13 @@ exit_onexit(struct proc *p)
 /*
  * exit -- death of process.
  */
-void
-sys_sys_exit(struct thread *td, struct sys_exit_args *uap)
+int
+sys_exit(struct thread *td, struct exit_args *uap)
```
bsdimp, line 209: This is a FreeBSD wart, but I've not chased down why we did it. All historic unixes through 4.4 used the bare exit()

bsdimp:
I looked through this the best I could, and couldn't find anything for what it's worth.

brooksdavis:
merged as 33f9ea209e954cfa3a89d05b400198bf14dd7ad6..0e765d9b08dcc2386b799815f2c835c97a93614c

PR #568, opened 2021-11-24
ipsec: fix a logic error in key_do_getnewspi

weliu:


lwhsu:
Ping @bu7cher for fcf596178b5f2be36424ecbc1b6a3224b29c91d2

mjguzik:
@bu7cher the patch look correct. I'll commit it next week if you don't get around to it.

@weliu thanks for the patch and apologies for the delay!

mjguzik:
Landed in https://cgit.freebsd.org/src/commit?id=509f1a0f405daf0e2bdaa5b366d21b77052149c3

PR #569, opened 2021-11-26
uchcom: Add USB product ID for CH341A

jlduran:
Add support for Qinheng's (aka WinChipHead) CH341A.

Naming and description obtained from NetBSD.

Datasheet: http://www.wch-ic.com/downloads/CH341DS1_PDF.html


jlduran:
The code to make CH341A work is not yet ready for submission. Close for now, and resubmit when ready.

PR #571, opened 2021-12-15
USB 32-bit compat

brooksdavis:
Replace the COMPAT_32BIT hacks in usb ioctls with actual 32-bit compatibility. We'll need this on CheriBSD because making pointers always 64-bit won't help when native pointers are 128-bit.

This compiles, but is completely untested. I'm not really sure how to go about doing that. I have reasonable confidence that the first commit is correct. Lower confidence that the fifo support is correct as it's more complex.

### DIFF
--- a/sys/dev/hid/hidraw.c
+++ b/sys/dev/hid/hidraw.c
@@ -41,6 +41,9 @@ __FBSDID("$FreeBSD$");
 #include "opt_hid.h"
 
 #include <sys/param.h>
+#ifdef COMPAT_FREEBSD32
+#include <sys/abi_compat.h>
+#endif
 #include <sys/bus.h>
 #include <sys/conf.h>
 #include <sys/fcntl.h>
@@ -114,6 +117,31 @@ struct hidraw_softc {
 
 	struct cdev *dev;
 };
+
+#ifdef COMPAT_FREEBSD32
+struct hidraw_gen_descriptor32 {
+	uint32_t hgd_data;	/* void * */
+	uint16_t hgd_lang_id;
+	uint16_t hgd_maxlen;
+	uint16_t hgd_actlen;
+	uint16_t hgd_offset;
+	uint8_t hgd_config_index;
+	uint8_t hgd_string_index;
+	uint8_t hgd_iface_index;
+	uint8_t hgd_altif_index;
+	uint8_t hgd_endpt_index;
+	uint8_t hgd_report_type;
+	uint8_t reserved[8];
+};
+#define	HIDRAW_GET_REPORT_DESC32 \
+    _IOC_NEWTYPE(HIDRAW_GET_REPORT_DESC, struct hidraw_gen_descriptor32)
+#define	HIDRAW_GET_REPORT32 \
+    _IOC_NEWTYPE(HIDRAW_GET_REPORT, struct hidraw_gen_descriptor32)
+#define	HIDRAW_SET_REPORT_DESC32 \
+    _IOC_NEWTYPE(HIDRAW_SET_REPORT_DESC, struct hidraw_gen_descriptor32)
+#define	HIDRAW_SET_REPORT32 \
+    _IOC_NEWTYPE(HIDRAW_SET_REPORT, struct hidraw_gen_descriptor32)
+#endif
 
 static d_open_t		hidraw_open;
 static d_read_t		hidraw_read;
@@ -507,11 +535,35 @@ hidraw_write(struct cdev *dev, struct uio *uio, int flag)
 	return (error);
 }
 
+#ifdef COMPAT_FREEBSD32
+static void
+update_hgd32(const struct hidraw_gen_descriptor *hgd,
+    struct hidraw_gen_descriptor32 *hgd32)
+{
+	/* Don't update hgd_data pointer */
+	CP(*hgd, *hgd32, hgd_lang_id);
+	CP(*hgd, *hgd32, hgd_maxlen);
+	CP(*hgd, *hgd32, hgd_actlen);
+	CP(*hgd, *hgd32, hgd_offset);
+	CP(*hgd, *hgd32, hgd_config_index);
+	CP(*hgd, *hgd32, hgd_string_index);
+	CP(*hgd, *hgd32, hgd_iface_index);
+	CP(*hgd, *hgd32, hgd_altif_index);
+	CP(*hgd, *hgd32, hgd_endpt_index);
+	CP(*hgd, *hgd32, hgd_report_type);
+	/* Don't update reserved */
+}
+#endif
+
 static int
 hidraw_ioctl(struct cdev *dev, u_long cmd, caddr_t addr, int flag,
     struct thread *td)
 {
 	uint8_t local_buf[HIDRAW_LOCAL_BUFSIZE];
+#ifdef COMPAT_FREEBSD32
+	struct hidraw_gen_descriptor local_hgd;
+	struct hidraw_gen_descriptor32 *hgd32 = NULL;
+#endif
 	void *buf;
 	struct hidraw_softc *sc;
 	struct hidraw_gen_descriptor *hgd;
@@ -526,6 +578,32 @@ hidraw_ioctl(struct cdev *dev, u_long cmd, caddr_t addr, int flag,
 	sc = dev->si_drv1;
 	if (sc == NULL)
 		return (EIO);
+
+#ifdef COMPAT_FREEBSD32
+	hgd = (struct hidraw_gen_descriptor *)addr;
+	switch (cmd) {
+	case HIDRAW_GET_REPORT_DESC32:
+	case HIDRAW_GET_REPORT32:
+	case HIDRAW_SET_REPORT_DESC32:
+	case HIDRAW_SET_REPORT32:
+		cmd = _IOC_NEWTYPE(cmd, struct hidraw_gen_descriptor);
+		hgd32 = (struct hidraw_gen_descriptor32 *)addr;
+		hgd = &local_hgd;
+		PTRIN_CP(*hgd32, *hgd, hgd_data);
+		CP(*hgd32, *hgd, hgd_lang_id);
+		CP(*hgd32, *hgd, hgd_maxlen);
+		CP(*hgd32, *hgd, hgd_actlen);
+		CP(*hgd32, *hgd, hgd_offset);
+		CP(*hgd32, *hgd, hgd_config_index);
+		CP(*hgd32, *hgd, hgd_string_index);
+		CP(*hgd32, *hgd, hgd_iface_index);
+		CP(*hgd32, *hgd, hgd_altif_index);
+		CP(*hgd32, *hgd, hgd_endpt_index);
+		CP(*hgd32, *hgd, hgd_report_type);
+		/* Don't copy reserved */
+		break;
+	}
+#endif
 
 	/* fixed-length ioctls handling */
 	switch (cmd) {
@@ -562,15 +640,19 @@ hidraw_ioctl(struct cdev *dev, u_long cmd, caddr_t addr, int flag,
 		mtx_lock(&sc->sc_mtx);
 		sc->sc_state.uhid = true;
 		mtx_unlock(&sc->sc_mtx);
-		hgd = (struct hidraw_gen_descriptor *)addr;
 		if (sc->sc_rdesc->len > hgd->hgd_maxlen) {
 			size = hgd->hgd_maxlen;
 		} else {
 			size = sc->sc_rdesc->len;
 		}
 		hgd->hgd_actlen = size;
+#ifdef COMPAT_FREEBSD32
+		if (hgd32 != NULL)
+			update_hgd32(hgd, hgd32);
+#endif
 		if (hgd->hgd_data == NULL)
 			return (0);		/* descriptor length only */
+
 		return (copyout(sc->sc_rdesc->data, hgd->hgd_data, size));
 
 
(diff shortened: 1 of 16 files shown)


sys/dev/usb/usb_generic.c
```diff
@@ -1192,54 +1203,175 @@ ugen_fs_copy_in(struct usb_fifo *f, uint8_t ep_index)
...
+    struct usb_fs_endpoint* fs_ep, bool copy_arrays)
+{
+	void *uptr;
+#ifdef COMPAT_FREEBSD32
+	struct usb_fs_endpoint32 fs_ep32;
+	uint32_t *uptr32;
+#endif
+	int error;
+
+	switch(f->fs_ep_sz) {
```
jrtc27, line 1243: Missing space (repeated on line 1285)

sys/dev/usb/usb_generic.c
```diff
@@ -943,6 +950,31 @@ ugen_do_request(struct usb_fifo *f, struct usb_ctl_request *ur)
...
+
+	error = ugen_do_request(f, &ur);
+
+	/* Don't update ucr_data pointer */
+	CP(ur, *ur32, ucr_flags);
+	CP(ur, *ur32, ucr_actlen);
+	CP(ur, *ur32, ucr_addr);
+	CP(ur, *ur32, ucr_request);
+
+	return(error);
```
jrtc27, line 974:
```suggestion
	return (error);
```

sys/dev/usb/usb_generic.c
```diff
@@ -1192,54 +1203,175 @@ ugen_fs_copy_in(struct usb_fifo *f, uint8_t ep_index)
...
+{
+	void *uptr;
+#ifdef COMPAT_FREEBSD32
+	struct usb_fs_endpoint32 fs_ep32;
+	uint32_t *uptr32;
+#endif
+	int error;
+
+	switch(f->fs_ep_sz) {
+	case sizeof(struct usb_fs_endpoint):
```
jrtc27, line 1244: Is there a reason you chose this over if/else if/else? Not aware of this being a common pattern in FreeBSD so far?

jrtc27 (commented):
Is the ambiguous return value for any of these fuword's a concern (i.e. should fueword be used instead)? Especially with the value being put in an unsigned variable in some places despite the return type being signed.

sys/dev/hid/hidraw.c
```diff
@@ -527,6 +579,32 @@ hidraw_ioctl(struct cdev *dev, u_long cmd, caddr_t addr, int flag,
...
+		CP(*hgd32, *hgd, hgd_actlen);
+		CP(*hgd32, *hgd, hgd_offset);
+		CP(*hgd32, *hgd, hgd_config_index);
+		CP(*hgd32, *hgd, hgd_string_index);
+		CP(*hgd32, *hgd, hgd_iface_index);
+		CP(*hgd32, *hgd, hgd_altif_index);
+		CP(*hgd32, *hgd, hgd_endpt_index);
+		CP(*hgd32, *hgd, hgd_report_type);
+		/* Don't copy reserved */
+		break;
```
hselasky, line 604: Style: Should maybe add:

default:
break;

at the end there.

sys/dev/usb/input/uhid.c
```diff
@@ -550,13 +550,30 @@ uhid_ioctl(struct usb_fifo *fifo, u_long cmd, void *addr,
...
+	switch (cmd) {
+	case USB_GET_REPORT_DESC32:
+	case USB_GET_REPORT32:
+	case USB_SET_REPORT32:
+		ugd32 = addr;
+		ugd = &local_ugd;
+		usb_gen_descriptor_from32(ugd, ugd32);
+		cmd = _IOC_NEWTYPE(cmd, struct usb_gen_descriptor);
+		break;
+	}
```
hselasky, line 572: default:
break;

sys/dev/usb/input/uhid_snes.c
```diff
@@ -281,13 +281,30 @@ uhid_snes_ioctl(struct usb_fifo *fifo, u_long cmd, void *data, int fflags)
...
+	switch (cmd) {
+	case USB_GET_REPORT_DESC32:
+	case USB_GET_REPORT32:
+	case USB_SET_REPORT32:
+		ugd32 = data;
+		ugd = &local_ugd;
+		usb_gen_descriptor_from32(ugd, ugd32);
+		cmd = _IOC_NEWTYPE(cmd, struct usb_gen_descriptor);
+		break;
+	}
```
hselasky, line 303: default:
break;

sys/dev/usb/usb_dev.c
```diff
@@ -1660,6 +1666,12 @@ usb_static_ioctl(struct cdev *dev, u_long cmd, caddr_t data, int fflag,
 			err = usb_read_symlink(u.urd->urd_data,
 			    u.urd->urd_startentry, u.urd->urd_maxlen);
 			break;
+#ifdef COMPAT_FREEBSD32
+		case USB_READ_DIR32:
+			err = usb_read_symlink(PTRIN(u.urd32->urd_data),
+			    u.urd32->urd_startentry, u.urd32->urd_maxlen);
+			    break;
```
hselasky, line 1673: break; has extra indention.

sys/dev/usb/usb_dev.h
```diff
@@ -123,6 +123,7 @@ struct usb_fifo {
 	void   *priv_sc0;		/* client data */
 	void   *priv_sc1;		/* client data */
 	void   *queue_data;
+	size_t fs_ep_sz;
```
hselasky, line 126: Could you use usb_size_t here? It makes it easier to compile the USB code for non-FreeBSD platforms and bootloaders?

sys/dev/usb/usb_generic.c
```diff
@@ -1223,54 +1203,175 @@ ugen_fs_copy_in(struct usb_fifo *f, uint8_t ep_index)
...
+	 * Avoid userspace DoS.  16 is selected because it's the
+	 * largest number any in-tree driver uses (most use 0-2).
+	 * It might be that MAXPHYS/sizeof(void *) would be better.
+	 */
+	if (fs_ep->nFrames > 16)
+		return (EINVAL);
+
+	switch(f->fs_ep_sz) {
+	case sizeof(struct usb_fs_endpoint):
+		fs_ep->ppBuffer = malloc(sizeof(*fs_ep->ppBuffer) *
```
hselasky, line 1287: The maximum number of frames / buffers is given when USB_FS_OPEN is called.
Please allocate this buffer there. Yes, it will be shared between threads, but only one thread should submit transfers at a time.

Or have a mixed stack / malloc usage, that when nFrames is small use stack, else malloc.

This function is in the fast path and can be called thousands of times per second and I don't want to see malloc there!
```
                        /* update number of frames */
                        u.popen->max_frames =
                            f->fs_xfer[u.popen->ep_index]->nframes;

```

sys/dev/usb/usb_generic.c
```diff
@@ -1223,54 +1203,175 @@ ugen_fs_copy_in(struct usb_fifo *f, uint8_t ep_index)
...
+#ifdef COMPAT_FREEBSD32
+	case sizeof(struct usb_fs_endpoint32):
+		u.fs_ep_uptr32 = (struct usb_fs_endpoint32 *)
+		    ugen_fs_ep_uptr(f, ep_index);
+		aFrames_uptr = &u.fs_ep_uptr32->aFrames;
+		isoc_time_complete_uptr = &u.fs_ep_uptr32->isoc_time_complete;
+		status_uptr = &u.fs_ep_uptr32->status;
+		break;
+#endif
+	}
```
hselasky, line 1343: default:
break;

sys/dev/usb/usb_generic.c
```diff
@@ -1402,31 +1490,14 @@ ugen_fs_copy_out(struct usb_fifo *f, uint8_t ep_index)
...
-	if (error)
-		goto done;
-
-	/* update "status" */
-	error = copyout(&fs_ep.status, &fs_ep_uptr->status,
-	    sizeof(fs_ep.status));
-done:
+	if (error != 0)
+		error = ugen_fs_update(&fs_ep, f, ep_index);
+	free(fs_ep.ppBuffer, M_USB);
```
hselasky, line 1500: free and malloc should not be in the fast path.

hselasky:
@brooksdavis :
Changes are mostly good, except for added malloc/free in the fast path. That must be fixed somehow, either by pre-allocation, or by copying in pointer by pointer maybe in chunks of 16 at a time on the stack. They are always processed sequentially so when "n % 16" == 0, then you simply copy in the next chunk min(16, nFrames - offset).

sys/dev/usb/usb_generic.c
```diff
@@ -1223,54 +1203,175 @@ ugen_fs_copy_in(struct usb_fifo *f, uint8_t ep_index)
 			}
 
 			/* move data */
-			error = copyin(uaddr, kaddr, length);
+			error = copyin(fs_ep.ppBuffer[n], kaddr, length);
```
hselasky, line 1206: I suggest something like this instead:

```
void *ppBuffer[16]; /* temporary array on stack */

if ((n % 16) == 0) {
   /* update temporary array every 16th interval */
   delta = MIN(16, xfer->nframes - n);
   error = copyin(ppBuffer, ....);
   ....
}
error = copyin(ppBuffer[n % 16], kaddr, length);
....
```

sys/dev/hid/hidraw.c
brooksdavis: Definitely not. Maybe:
```
default:
panic(...);
```
but not a default case that does nothing for a programming error.

sys/dev/usb/usb_generic.c
brooksdavis: There will be three values in CheriBSD so a case is more natural and this will produce less churn.

sys/dev/hid/hidraw.c
brooksdavis: Sorry, in this case one would be fine, but IMO utterly pointless. This is your code though so if you want pointless default cases I can add them.

sys/dev/usb/usb_generic.c
brooksdavis: If there's going to be a `default:` here it must be a panic.

brooksdavis:
I've made most of the requested changes.

I dropped the ppBuffer buffer allocation in favor of an accessor that copies in the value and translates it if required. Hopefully on systems without COMPAT_FREEBSD32 defined the compiler will collapse it to an fueword call.

I'd been waffling on fuword vs fueword and switch to fueword. I think all were safe in practice but fueword isn't much more verbose.

I've not added extra `default: break;` statements to the case statements for thunks, but could do so. I did add `default: panic(...)` to switches on struct size. That's probably a bit overkill.

sys/dev/hid/hidraw.c
hselasky: OK, as long as the compiler doesn't complain.

sys/dev/usb/usb_generic.c
```diff
@@ -1066,6 +1067,38 @@ ugen_fs_set_complete(struct usb_fifo *f, uint8_t index)
...
+
+	u.ppBuffer = buffer;
+	switch (f->fs_ep_sz) {
+	case sizeof(struct usb_fs_endpoint):
+		if (fueword(u.ppBuffer + n, (long *)uptrp) != 0)
+			return (EFAULT);
+		return (0);
+#ifdef COMPAT_FREEBSD32
+	case sizeof(struct usb_fs_endpoint32):
+		if (fueword32(u.ppBuffer + n, &uptr32) != 0)
```
hselasky, line 1092: I guess:
u.ppBuffer -> u.ppBuffer32

brooksdavis: oops, will fix

hselasky: Patches are looking good otherwise.

hselasky: Do you have something to test with?

sys/dev/usb/usb_generic.c
```diff
@@ -1402,31 +1503,13 @@ ugen_fs_copy_out(struct usb_fifo *f, uint8_t ep_index)
...
-	    &fs_ep_uptr->isoc_time_complete,
-	    sizeof(fs_ep.isoc_time_complete));
-	if (error)
-		goto done;
-
-	/* update "status" */
-	error = copyout(&fs_ep.status, &fs_ep_uptr->status,
-	    sizeof(fs_ep.status));
-done:
+	if (error != 0)
```
hselasky, line 1511: I think you mean "if (error == 0)" here?

hselasky:
One more issue.

You can test using a memory stick and /usr/src/tools/tools/usbtest (string descriptor test and read-only test) should cover the code you've patched.

brooksdavis:
I've tested with an emulated umass device on qemu and see the same results for
 - 64-bit usbtest on freebsd/main without patches
 - 64-bit usbtest with patches
 - 32-bit usbtest with patches

sys/dev/usb/usb_generic.c
```diff
@@ -1066,6 +1067,38 @@ ugen_fs_set_complete(struct usb_fifo *f, uint8_t index)
...
+#endif
+	} u;
+#ifdef COMPAT_FREEBSD32
+	uint32_t uptr32;
+#endif
+
+	u.ppBuffer = buffer;
+	switch (f->fs_ep_sz) {
+	case sizeof(struct usb_fs_endpoint):
+		if (fueword(u.ppBuffer + n, (long *)uptrp) != 0)
```
hselasky, line 1087: Should this be: (uintptr_t *) ?

hselasky:
Spelling:
usb: add 32-bit compat for fifos 
usb: add 32-bit compat for FIFOs 

I'm good with the current changes.

brooksdavis:
merged as 45b48cbc2b5819cd6e3dee3632d66e55d5d7c101..ea972feea2d2589ee1317a19eeada0598332defa

PR #572, opened 2021-12-17
Fix "Mesurement" typo in usr.sbin/acpi/acpiconf/acpiconf.c

mattaudesse:


bsdimp:
pushed as 9d4785813870

PR #573, opened 2021-12-24
Remove the function vattr_null

wy-chung:
Remove the function vattr_null and initialize the variable va_null when it is defined.


wy-chung:
Initialize the variable 'va_null' at compile time so that we don't have to initialize it in vfs_register.

PR #575, opened 2022-01-11
report full error string on SSL_connect() failure in rpc.tlsclntd

vladak:
This change reports full OpenSSL error string on `SSL_connect()` failure in `rpc.tlsclntd`. Looks like this:
```
Jan 10 18:44:21 freebsd rpc.tlsclntd[12198]: rpctls_connect: SSL_connect failed -1: error:1408F10B:SSL routines:ssl3_get_record:wrong version number
```
I realize that:
  - this is not thread safe
  - there is lot of other places in the code that should receive the same treatment
  - it prints only the top level error from the OpenSSL ERR stack


vladak:
@bsdjhb might be interested in this one.

bsdimp:
This looks good to me, so I'm going to go ahead and push this since @bsdjhb has had long enough to object :)

PR #577, opened 2022-01-29
fix: Missing comma in a list in `event_rpcgen.py`

mrshu:
* The comma has most probably been left out unintentionally, leading to string concatenation between the two lines and potential trouble.

### DIFF
--- a/contrib/ntp/sntp/libevent/event_rpcgen.py
+++ b/contrib/ntp/sntp/libevent/event_rpcgen.py
@@ -811,7 +811,7 @@ class EntryStruct(Entry):
             '  had_error = 1;',
             '  goto done;',
             '}',
-            'done:'
+            'done:',
             'if (tmp != NULL)',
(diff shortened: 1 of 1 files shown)


emaste:
can you double check against this upstream change https://github.com/libevent/libevent/commit/6a3ca347379d28f8d6f8f9a621867860b6097b15

bsdimp:
This code is imported from the ntp project. Since this issue is very minor (the string pasting results in code that's fine), please take it up with them. FreeBSD will get the bug fix next time we import from upstream.

bsdimp:
> can you double check against this upstream change [libevent/libevent@6a3ca34](https://github.com/libevent/libevent/commit/6a3ca347379d28f8d6f8f9a621867860b6097b15)

This commit looks to be a super-set of this change. I closed this because I think that we should ingest this via the two upstreams that provide this..

PR #580, opened 2022-02-22
libefivar: Correct the string expression of UTF8 vendor device path

jlduran:
REF: https://bugzilla.tianocore.org/show_bug.cgi?id=1225

According to UEFI spec, the string expression of UTF8 vendor device node should be displayed as: `VenUtf8()`. Current code displays it as: `VenUft8()` by mistake when convert device path node to text.

This commit is to fix this bug.

Obtained from:	https://github.com/tianocore/edk2/commit/959be180e185869b71db9dad34c3f4bba660d724


bsdimp:
Landed as 91a35e5803be

PR #581, opened 2022-02-25
Update libefivar

jlduran:
@bsdimp: This is a commit-by-commit update of libefivar.

Maybe not all commits are wanted, but surely, there are some interesting ones to be cherry-picked (almost all are atomic).

Some formatting, despite FreeBSD-Update instructions, was done in order to reduce the diffs and have a clear view of the differences.


bsdimp:
These all look decent to me, so I've gone ahead and merged them all. Nothing looks terrible, even the cleanup you did for diff reduction. I've tweaked FreeBSD-updating and the commit message to that. Other than tagging all the others with this pull request, I've let them go in unchanged. Thanks for catching up with EDK2.

PR #584, opened 2022-03-02
Fix sbintime conversion divisors

jarekpelczar:
Functions which convert from ms, us and ns divided by SBT_1S constants instead of
dividing by proper conversion factor.


amotin (approved):
The comment about "divided by" is confusing, but the patch looks making sense.

bsdimp:
Divisor is the wrong name here. It's ms_per_sec, us_per_sec and ns_per_sec

bsdimp:
by chance do you have a atf test program to test this condition?

PR #585, opened 2022-03-02
bit_count: Fix invalid calculation when _start is not zero

jarekpelczar:
_bit_make_mask calculation fails when using e.g.

mask = bits 0 to 200 are set
bit_count(mask, 10, 30) returns 20, but should return 30


jarekpelczar:
Need to check again

PR #589, opened 2022-03-11
kern: Remove duplicate header include

jlduran:


bsdimp:
Landed as df949e762c67b4b4109dd16b0179f17faa925cd2.... since we're sorting the include files anyway, I removed the duplicate at the same time.

PR #590, opened 2022-03-13
linuxulator: add sendfile fallback for non-socket FDs

udp:
In Linux < 2.6.33, sendfile could only be used to send from a file to a socket. This is the behaviour implemented by the FreeBSD Linuxulator. However, since 2.6.33, sendfile can send from any FD to any other FD, which is not implemented by FreeBSD sendfile or by the Linuxulator.

I ran into this problem trying to run the game "Factorio" using the Linuxulator, which relies on sendfile to copy files on disk. This patch changes the Linuxulator sendfile such that it will use FreeBSD sendfile where possible, but otherwise fallback on a simple read/write loop.

I would very much appreciate some feedback, as I have never contributed to the FreeBSD kernel before. In particular, any suggestion for a better way to choose the buffer size than the arbitrary 8 KB I have hardcoded would be welcome.

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=262535

### DIFF
--- a/sys/compat/linux/linux_socket.c
+++ b/sys/compat/linux/linux_socket.c
@@ -54,6 +54,8 @@ __FBSDID("$FreeBSD$");
 #include <sys/syslog.h>
 #include <sys/un.h>
 #include <sys/unistd.h>
+#include <sys/protosw.h>
+#include <sys/vnode.h>
 
 #include <security/audit/audit.h>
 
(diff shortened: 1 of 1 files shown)


sys/compat/linux/linux_socket.c
```diff
@@ -2058,20 +2060,162 @@ linux_getsockopt(struct thread *td, struct linux_getsockopt_args *args)
...
+			goto cleanup;
+		}
+	}
+
+cleanup:
+	*bytes_read = bytes_sent;
+	free(tmpbuf, M_TEMP);
+	return (error);
+}
+
```
jake-87, line 2199: Should probably `#undef` earlier macro of TMPBUF_SIZE to ensure no issues

udp:
Also uploaded at https://reviews.freebsd.org/D34555

bsdimp:
what's the status of this? I see a lot of feedback on D34555, but no updates after a point...

bsdimp:
I'm going to close this. there's no activity on the Phab reivew after the initial feedback. This isn't ready to land.
Once it is ready, please go ahead and re-open the pull request. I'll leave the phab open.

PR #591, opened 2022-03-14
Add `NULL` check to prevent `panic` in `mixer.c`

qiu-x:
Without this check a kernel panic is triggered after running the following program:
```c
#include <fcntl.h> 
#include <stdio.h> 
#include <stdlib.h> 
#include <sys/soundcard.h>
#include <unistd.h> 

int
main()
{
	int fd;
	fd = open("/dev/mixer", O_RDWR);
	oss_mixerinfo mi;
	ioctl(fd, SNDCTL_MIXERINFO, &mi);
	return (0);
}
```


qiu-x:
Also uploaded at https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=262671

PR #592, opened 2022-03-16
Fix _REVISION to be always major.minor

astralblue:
Introduced in commit 613fe53, _REVISION was redefined to be just the
major version (14) instead of major.minor (14.0) iff:

  - PKG_VERSION was not overridden (default), and
  - Branch was -CURRENT, -STABLE, or -PRERELEASE.

This introduced side effects in other parts of the code which expected
_REVISION to be major.minor:

  - make-pkg-package.sh target triple became amd64-portbld-freebsd14
    (should be amd64-portbld-freebsd14.0);
  - Versioned cross dev symlink name became amd64-freebsd14-&lt;tool&gt;
    (should be amd64-freebsd14.0-&lt;tool&gt;)

Also, compounded with commit ea9a92d (which was introduced afterward
and removed minor version from _REVISION incorrectly using :S with a
regex) this also caused pkgbase repo to use a wrong PKG_ABI with a minor
version (FreeBSD:14.0:amd64) when a custom PKG_VERSION was specified.
(Note, without a custom PKG_VERSION the bug was not triggered because in
that case _REVISION was already major-only.)

This commit fixes both problems by introducing and using MAJOR_REVISION
instead of redefining _REVISION.  Existing uses of _REVISION now see
major.minor again, except PKG_ABI (ex: FreeBSD:14:amd64, as described
above) and PKG_VERSION (ex: 14.snap20220311121531) need only the major
version and now use MAJOR_REVISION instead.

### DIFF
--- a/Makefile.inc1
+++ b/Makefile.inc1
@@ -537,12 +537,12 @@ SRCRELDATE!=	awk '/^\#define[[:space:]]*__FreeBSD_version/ { print $$3 }' \
 VERSION=	FreeBSD ${_REVISION}-${_BRANCH:C/-p[0-9]+$//} ${TARGET_ARCH} ${SRCRELDATE}
 .export VERSION
 .endif
+MAJOR_REVISION=	${_REVISION:R}
 
 .if !defined(PKG_VERSION)
 _STRTIMENOW=	%Y%m%d%H%M%S
 _TIMENOW=	${_STRTIMENOW:gmtime}
 .if ${_BRANCH:MCURRENT*} || ${_BRANCH:MSTABLE*} || ${_BRANCH:MPRERELEASE*}
-_REVISION:=	${_REVISION:R}
 EXTRA_REVISION=	.snap${_TIMENOW}
 .elif ${_BRANCH:MALPHA*}
 EXTRA_REVISION=	.a${_BRANCH:C/ALPHA([0-9]+).*/\1/}.${_TIMENOW}
@@ -553,7 +553,7 @@ EXTRA_REVISION=	.rc${_BRANCH:C/RC([0-9]+).*/\1/}.${_TIMENOW}
 .elif ${_BRANCH:M*-p*}
 EXTRA_REVISION=	p${_BRANCH:C/.*-p([0-9]+$)/\1/}
 .endif
-PKG_VERSION:=	${_REVISION}${EXTRA_REVISION:C/[[:space:]]//g}
+PKG_VERSION:=	${MAJOR_REVISION}${EXTRA_REVISION:C/[[:space:]]//g}
 .endif
 .endif	# !defined(PKG_VERSION)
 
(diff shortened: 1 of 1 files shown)


astralblue:
Fixes bugs [#262600](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=262600) and [#262601](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=262601).

bsdimp:
This seems sane to me... @emaste @evadot ?

evadot:
Probably ok but I haven't tested and reviewing makefile stuff always makes my head hurt :)

emaste:
I think it's fine

Makefile.inc1
```diff
@@ -553,7 +553,7 @@ EXTRA_REVISION=	.rc${_BRANCH:C/RC([0-9]+).*/\1/}.${_TIMENOW}
 .elif ${_BRANCH:M*-p*}
 EXTRA_REVISION=	p${_BRANCH:C/.*-p([0-9]+$)/\1/}
 .endif
-PKG_VERSION:=	${_REVISION}${EXTRA_REVISION:C/[[:space:]]//g}
+PKG_VERSION:=	${MAJOR_REVISION}${EXTRA_REVISION:C/[[:space:]]//g}
```
bsdimp, line 556: This looks fine: It's exactly what it was before.

Makefile.inc1
```diff
@@ -1885,7 +1885,7 @@ _pkgbootstrap: .PHONY
 # Determine PKG_ABI from newvers.sh if not already set.
 #
 .if !defined(PKG_ABI) && (make(create-world-packages-jobs) || make(create-kernel-packages*) || make(real-update-packages) || make(sign-packages))
-PKG_ABI=${_TYPE}:${_REVISION:S/\..*$//}:${TARGET_ARCH}
+PKG_ABI=${_TYPE}:${MAJOR_REVISION}:${TARGET_ARCH}
```
bsdimp, line 1888: This looks fine as well.... The ABI is supposed to only be the major number.

Makefile.inc1
```diff
@@ -537,12 +537,12 @@ SRCRELDATE!=	awk '/^\#define[[:space:]]*__FreeBSD_version/ { print $$3 }' \
 VERSION=	FreeBSD ${_REVISION}-${_BRANCH:C/-p[0-9]+$//} ${TARGET_ARCH} ${SRCRELDATE}
 .export VERSION
 .endif
+MAJOR_REVISION=	${_REVISION:R}
 
 .if !defined(PKG_VERSION)
 _STRTIMENOW=	%Y%m%d%H%M%S
 _TIMENOW=	${_STRTIMENOW:gmtime}
 .if ${_BRANCH:MCURRENT*} || ${_BRANCH:MSTABLE*} || ${_BRANCH:MPRERELEASE*}
-_REVISION:=	${_REVISION:R}
```
bsdimp, line 545: This changes what xdev links happen, and it's my belief they are now more correct since they are supposed to have major.minor in them....

There's only one other use of _REVSION other than below I could find:
```
package-pkg: .PHONY
        rm -rf /tmp/ports.${TARGET} || :
        env ${WMAKEENV:Q} SRCDIR=${.CURDIR} PORTSDIR=${PORTSDIR} REVISION=${_REVISION} \
                PKG_CMD=${PKG_CMD} PKG_VERSION=${PKG_VERSION} REPODIR=${REPODIR} \
                WSTAGEDIR=${WSTAGEDIR} \
                sh ${.CURDIR}/release/scripts/make-pkg-package.sh

```
and make-pkg-package.sh just uses it for
```
        CONFIGURE_ARGS="--host=$(uname -m)-portbld-freebsd${REVISION}" \
```
which should have major.minor in it

bsdimp:
I took a close look, and I'm sure the xdev stuff is now more correct. I added notes to the review, and then realized I was largely replicating the commit message (which I didn't study before I made the comments, honest).
Landed as a1e3fb8fa0a4652bc345bd284c32d6630a8dbe08

PR #595, opened 2022-04-10
Strip trailing slashes from DESTDIR

mariusvw:
Solves duplicate slashes in paths

```
Scanning //usr/share/certs/trusted for certificates...
Scanning //usr/local/share/certs for certificates...
```


emaste:
CC @kevans91

bsdimp:
landed as 17720d0bd0e8

jrtc27:
This also points at the fact Makefile.inc1 should be using ${_INSTALL_DDIR} rather than ${DESTDIR}/${DISTDIR} for this, but certctl can be used in other contexts so worth it being robust too.

PR #597, opened 2022-04-15
More complete -m32 support for amd64

brooksdavis:
This set of patches makes am amd64 sysroot sufficiently compatible to build C startup (csu) code and all libraries in the base system for i386 using the -m32 flag.  The patches use two methods to do this depending on header complexity:
 -  Making a single combined header installed in `/usr/include/x86` which is included by stubs in the amd64 and i386 `machine` directories. Some existing headers do this.
 - Adding a new `i386` include directory on amd64 where the i386 header is installed and included in place of the contents of the amd64 version when targeting i386.

### DIFF
--- a/sys/i386/include/cpufunc.h
+++ b/sys/i386/include/cpufunc.h
@@ -98,7 +98,6 @@ clts(void)
 static __inline void
 disable_intr(void)
 {
-
 	__asm __volatile("cli" : : : "memory");
 }
 
@@ -147,14 +146,12 @@ cpuid_count(u_int ax, u_int cx, u_int *p)
 static __inline void
 enable_intr(void)
 {
-
 	__asm __volatile("sti");
 }
 
 static __inline void
 cpu_monitor(const void *addr, u_long extensions, u_int hints)
 {
-
 	__asm __volatile("monitor"
 	    : : "a" (addr), "c" (extensions), "d" (hints));
 }
@@ -162,28 +159,24 @@ cpu_monitor(const void *addr, u_long extensions, u_int hints)
 static __inline void
 cpu_mwait(u_long extensions, u_int hints)
 {
-
 	__asm __volatile("mwait" : : "a" (hints), "c" (extensions));
 }
 
 static __inline void
 lfence(void)
 {
-
 	__asm __volatile("lfence" : : : "memory");
 }
 
 static __inline void
 mfence(void)
 {
-
 	__asm __volatile("mfence" : : : "memory");
 }
 
 static __inline void
 sfence(void)
 {
-
 	__asm __volatile("sfence" : : : "memory");
 }
 
--- a/Makefile.inc1
+++ b/Makefile.inc1
@@ -1062,6 +1062,9 @@ _bootstrap-tools:
 	    -p ${WORLDTMP}/usr >/dev/null
 	${WORLDTMP_MTREE} -f ${.CURDIR}/etc/mtree/BSD.include.dist \
 	    -p ${WORLDTMP}/usr/include >/dev/null
+.if ${TARGET_ARCH} == "amd64"
+	mkdir -p ${WORLDTMP}/usr/include/i386
+.endif
 	ln -sf ${.CURDIR}/sys ${WORLDTMP}
 .if ${MK_DEBUG_FILES} != "no"
 	${WORLDTMP_MTREE} -f ${.CURDIR}/etc/mtree/BSD.debug.dist \
@@ -1395,6 +1398,9 @@ distributeworld installworld stageworld: _installcheck_world .PHONY
 	    -p ${DESTDIR}/${DISTDIR}/${dist}/usr >/dev/null
 	${DESTDIR_MTREE} -f ${.CURDIR}/etc/mtree/BSD.include.dist \
 	    -p ${DESTDIR}/${DISTDIR}/${dist}/usr/include >/dev/null
+.if ${TARGET_ARCH} == "amd64"
+	-mkdir  ${DESTDIR}/${DISTDIR}/usr/include/i386
+.endif
 .if ${MK_DEBUG_FILES} != "no"
 	${DESTDIR_MTREE} -f ${.CURDIR}/etc/mtree/BSD.debug.dist \
 	    -p ${DESTDIR}/${DISTDIR}/${dist}/usr/lib >/dev/null
@@ -1423,6 +1429,9 @@ distributeworld installworld stageworld: _installcheck_world .PHONY
 	    sed -e 's#^\./#./${dist}/usr/#' >> ${METALOG}
 	${IMAKEENV} ${DISTR_MTREE} -C -f ${.CURDIR}/etc/mtree/BSD.include.dist | \
 	    sed -e 's#^\./#./${dist}/usr/include/#' >> ${METALOG}
+.if ${TARGET_ARCH} == "amd64"
+	echo "./${dist}/usr/include/i386 type=dir uname=root gname=wheel mode=0755" >> ${METALOG}
+.endif
 .if defined(_LIBCOMPAT)
 	${IMAKEENV} ${DISTR_MTREE} -C -f ${.CURDIR}/etc/mtree/BSD.lib${libcompat}.dist | \
 	    sed -e 's#^\./#./${dist}/usr/#' >> ${METALOG}
@@ -2852,6 +2861,9 @@ native-xtools-install: .PHONY
 	    -p ${NXBDESTDIR}/usr >/dev/null
 	${DESTDIR_MTREE} -f ${.CURDIR}/etc/mtree/BSD.include.dist \
 	    -p ${NXBDESTDIR}/usr/include >/dev/null
+.if ${TARGET_ARCH} == "amd64"
+	mkdir -p ${NXBDESTDIR}/usr/include/i386
+.endif
 	${_+_}cd ${.CURDIR}; ${NXBMAKE} \
 	    DESTDIR=${NXBDESTDIR} \
 	    -DNO_ROOT \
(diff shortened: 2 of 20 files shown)


brooksdavis:
Oops, spotted two more commits I'll pull out of this series:
 - Only include things for btxld on supported platforms
 - a.out.h: expose only nlist on non-x86

Both relate to an attempt to clean up use of machine/reloc.h that I'm going to do separately.

sys/x86/include/db_machdep.h
```diff
@@ -0,0 +1,120 @@
...
+#define	_X86_DB_MACHDEP_H_
+
+#include <machine/frame.h>
+#include <machine/trap.h>
+
+typedef	vm_offset_t	db_addr_t;	/* address - unsigned */
+#ifdef __amd64__
+typedef	long		db_expr_t;	/* expression - signed */
+#else
+typedef	int		db_expr_t;	/* expression - signed */
```
bsdjhb, line 39: Does it break a lot if you just use long on i386 as well?

sys/x86/include/i386/npx.h
```diff
@@ -0,0 +1,99 @@
...
+
+/*
+ * 287/387 NPX Coprocessor Data Structures and Constants
+ * W. Jolitz 1/90
+ */
+
+#ifndef _X86_I386_NPX_H_
+#define	_X86_I386_NPX_H_
+
+#include <x86/fpu.h>
```
bsdjhb, line 46: This change kind of surprises me as userland shouldn't care about the #ifdef _KERNEL bits?  Shouldn't the existing amd64 machine/fpu.h already work in userland as it #includes <x86/fpu.h> which has the the #ifdef's to work for both?

sys/x86/include/gdb_machdep.h
```diff
@@ -0,0 +1,101 @@
...
+ * NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+ * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * $FreeBSD$
+ */
+
+#ifndef _X86_GDB_MACHDEP_H_
```
bsdjhb, line 31: This one also surprises me.  I assumed this was a kernel-only header and not used by userland?

sys/x86/include/i386/md_var.h
```diff
@@ -0,0 +1,84 @@
...
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ *
+ * $FreeBSD$
+ */
+
+#ifndef _X86_I386_MD_VAR_H_
```
bsdjhb, line 34: This is another kernel-only header I thought?

sys/x86/include/i386/counter.h
```diff
@@ -0,0 +1,199 @@
...
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ *
+ * $FreeBSD$
+ */
+
+#ifndef __X86_I386_COUNTER_H__
```
bsdjhb, line 31: Kernel only?

sys/x86/include/i386/pcpu.h
```diff
@@ -0,0 +1,205 @@
...
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ *
+ * $FreeBSD$
+ */
+
+#ifndef _X86_I386_PCPU_H_
```
bsdjhb, line 31: I wondered if it might be possible to leave the #ifdef KERNEL bits in sys/i386/include/pcpu.h still, but perhaps that isn't much cleaner, and probably a bit harder to maintain going forward.

sys/x86/include/runq.h
```diff
@@ -0,0 +1,65 @@
...
+#else
+#define	RQB_LEN		(1)		/* Number of priority status words. */
+#define	RQB_L2BPW	(6)		/* Log2(sizeof(rqb_word_t) * NBBY)). */
+#endif
+#define	RQB_BPW		(1<<RQB_L2BPW)	/* Bits in an rqb_word_t. */
+
+#ifdef __i386__
+#define	RQB_BIT(pri)	(1 << ((pri) & (RQB_BPW - 1)))
+#else
+#define	RQB_BIT(pri)	(1ul << ((pri) & (RQB_BPW - 1)))
```
bsdjhb, line 46: This version would work on i386 as well (the ul wouldn't hurt)

sys/x86/include/runq.h
```diff
@@ -0,0 +1,65 @@
...
+#ifdef __i386__
+#define	RQB_FFS(word)	(ffs(word) - 1)
+#else
+#define	RQB_FFS(word)	(bsfq(word))
+#endif
+
+/*
+ * Type of run queue status word.
+ */
+#ifdef __i386__
```
bsdjhb, line 59: This could perhaps use `unsigned long` on both platforms?

bsdjhb (commented):
So I have one general comment perhaps which is I understand why you did it this way, but the actual moving of files in the kernel is somewhat annoying from a future maintenance perspective.  I'm also thinking about other platforms in the future (e.g. if we wanted to support -m32 on arm).  One possible alternative is to still patch all the amd64 headers, but instead of moving the i386 headers, add an explicit list in include/Makefile (or include/Makefile.amd64) of headers to copy from sys/i386/include to DESTDIR/usr/include/x86/i386.  This would mean that you can leave the i386 headers in their current location and only need to add an i386 header to a variable in the Makefile along with the simple patch to the amd64 header to add the fallback.

My other suggestion is that there are several kernel-only headers I noted that I'm surprised are used.  It might be good in the commit logs to explain what bits of the userland build use them as you did for some of the later commits for headers used by procstat's ZFS bits.  (It might also be nice if the ZFS folks could fix the procstat bits to not need the full-on _KERNEL treatment but add some _WANT_ZFS_FOO knobs or the like to limit the exposure of kernel-only bits to userspace).

sys/x86/include/db_machdep.h
brooksdavis: I tested and hit two cases (`db_printf("%r",...)` and a type mismatch with `size_t`). Changing this to `register_t` might work.

sys/x86/include/i386/npx.h
brooksdavis: Indeed, I might have been a little too much in the zone. I'll retest without this change.

brooksdavis:
I've reworked this to create a /usr/include/i386 directory on amd64 which contains only the files we actually require. The amd64 files then include them in place of the entire file contents when in -m32 mode. I also re-faulted in the files we actually required so the total set of files is reduced. I'd gotten a bit ahead of myself converting some of the files.

The current set doesn't include kernel support as I forgot to test before pushing.

sys/x86/include/i386/counter.h
brooksdavis: procstat ZFS support :(

sys/x86/include/i386/md_var.h
brooksdavis: More procstat zfs support.

bsdjhb (approved):
Thanks, this approach seems cleaner to me.  We could also use this to add in real lib32/-m32 support on aarch64 in the future if we wanted.

brooksdavis:
rebased, added revised by.

sys/i386/include/cpufunc.h
```diff
@@ -147,43 +146,37 @@ cpuid_count(u_int ax, u_int cx, u_int *p)
...
 mfence(void)
 {
-
 	__asm __volatile("mfence" : : : "memory");
 }
 
 static __inline void
 sfence(void)
 {
-
```
bsdimp, line 186: This seems unrelated, but the change itself is good.

bsdimp:
I don't see how to tick the 'approved' box, though I approve it.

Reviewed by: imp

sys/i386/include/cpufunc.h
brooksdavis: This dates to a prior effort to merge these two files and I kept it because it's a basically sensible change so didn't see a need to throw it out.

brooksdavis:
Merged to freebsd

PR #598, opened 2022-04-17
Modularize uuencode and uudecode by wrapping them in bintrans.c

pstef:
The program will be installed as bintrans, uuencode, uudecode,
b64encode, and b64decode and will be responsible for running the coders
according to their historical behavior.

Additionally, bintrans will be able to take a parameter designating
the coder and accept all its options in this form:
bintrans <coder> [options]
and the behavior should be the same as if
<coder> [options]
was invoked.
This has the advantage that adding coders won't require installing them
as binaries.

Move uudecode files to uuencode since the latter is the one that
provides the manual page.


pstef:
This was to trigger CI but it's been revealed to me that I don't need PRs and can trigger this in my fork.

PR #599, opened 2022-06-06
.github: Cross-build aarch64 as well as amd64 kernels.

bsdjhb:


.github/workflows/cross-bootstrap-tools.yml
```diff
@@ -27,6 +28,10 @@ jobs:
             compiler: clang-12
             cross-bindir: /usr/local/opt/llvm/bin
             pkgs: bmake libarchive
+	  - arch: amd64
+	    target: amd64
```
emaste, line 32: probably `target_arch` for consistency?

emaste:
Does this result in 6 builds? (ie amd64 and arm64 on the three existing os/compiler combinations)

bsdjhb:
Right now it results in zero builds, need to debug this more.  I did rename to target_arch locally.  If we could drop clang 9 and just do a single build on ubuntu I think this would be simpler to deal with.

bsdjhb:
Well, progress it seems.  I had misparsed the earlier syntax error and might see if I can revert the 'host' helper variable (though I suspect I need it).

bsdjhb:
The dtc fix (which we already had downstream in CheriBSD) is needed to fix the build of dtc on with newer versions of G++/libstdc++ used on the newer ubuntu 22.04 image.  aarch64 didn't build with clang 9 nor clang 10 as they didn't support the PAC registers and didn't really seem useful to keep old versions of the compilers around.  If we really wanted the old versions I could bring them back but use an exclude clause to only use them for x86.

brooksdavis:
12.3 has LLVM 10. Officially we don't require support for that upgrade path.

PR #603, opened 2022-06-28
D35560

ehem:
As per: https://reviews.freebsd.org/D35560


bsdimp:
Resolved conflicts, dropped psm changes (it's vendor code, despite not living in sys/contrib), dropped netmap as already done.
Built GENERIC on amd64.

PR #604, opened 2022-07-08
Simplify pps_capture() and pps_event()

sebhub:
This patch set tries to simplify the pps_capture() and pps_event() processing.

I am not sure if the rounding change is an improvement.

### DIFF
--- a/sys/kern/kern_tc.c
+++ b/sys/kern/kern_tc.c
@@ -1761,6 +1761,7 @@ void
 pps_capture(struct pps_state *pps)
 {
 	struct timehands *th;
+	struct timecounter *tc;
 
 	KASSERT(pps != NULL, ("NULL pps pointer in pps_capture"));
 	th = timehands;
@@ -1769,7 +1770,8 @@ pps_capture(struct pps_state *pps)
 #ifdef FFCLOCK
 	pps->capffth = fftimehands;
 #endif
-	pps->capcount = th->th_counter->tc_get_timecount(th->th_counter);
+	tc = th->th_counter;
+	pps->capcount = tc->tc_get_timecount(tc);
 	atomic_thread_fence_acq();
 	if (pps->capgen != th->th_generation)
 		pps->capgen = 0;
--- a/sys/kern/kern_tc.c
+++ b/sys/kern/kern_tc.c
@@ -1772,14 +1772,14 @@ pps_capture(struct pps_state *pps)
 #endif
 	tc = th->th_counter;
 	pps->capcount = tc->tc_get_timecount(tc);
-	atomic_thread_fence_acq();
-	if (pps->capgen != th->th_generation)
-		pps->capgen = 0;
 }
 
 void
 pps_event(struct pps_state *pps, int event)
 {
+	struct timehands *capth;
+	struct timecounter *captc;
+	uint64_t capth_scale;
 	struct bintime bt;
 	struct timespec ts, *tsp, *osp;
 	u_int tcount, *pcount;
@@ -1798,9 +1798,17 @@ pps_event(struct pps_state *pps, int event)
 	/* Nothing to do if not currently set to capture this event type. */
 	if ((event & pps->ppsparam.mode) == 0)
 		return;
+
+	/* Make a snapshot of the captured timehand */
+	capth = pps->capth;
+	captc = capth->th_counter;
+	capth_scale = capth->th_scale;
+	tcount = capth->th_offset_count;
+	bt = capth->th_bintime;
+
 	/* If the timecounter was wound up underneath us, bail out. */
-	if (pps->capgen == 0 || pps->capgen !=
-	    atomic_load_acq_int(&pps->capth->th_generation))
+	atomic_thread_fence_acq();
+	if (pps->capgen == 0 || pps->capgen != capth->th_generation)
 		return;
 
 	/* Things would be easier with arrays. */
@@ -1838,24 +1846,18 @@ pps_event(struct pps_state *pps, int event)
 	 * If the timecounter changed, we cannot compare the count values, so
 	 * we have to drop the rest of the PPS-stuff until the next event.
 	 */
-	if (pps->ppstc != pps->capth->th_counter) {
-		pps->ppstc = pps->capth->th_counter;
+	if (__predict_false(pps->ppstc != captc)) {
+		pps->ppstc = captc;
 		*pcount = pps->capcount;
 		pps->ppscount[2] = pps->capcount;
 		return;
 	}
 
 	/* Convert the count to a timespec. */
-	tcount = pps->capcount - pps->capth->th_offset_count;
-	tcount &= pps->capth->th_counter->tc_counter_mask;
-	bt = pps->capth->th_bintime;
-	bintime_addx(&bt, pps->capth->th_scale * tcount);
+	tcount = pps->capcount - tcount;
+	tcount &= captc->tc_counter_mask;
+	bintime_addx(&bt, capth_scale * tcount);
 	bintime2timespec(&bt, &ts);
-
-	/* If the timecounter was wound up underneath us, bail out. */
-	atomic_thread_fence_acq();
-	if (pps->capgen != pps->capth->th_generation)
-		return;
 
 	*pcount = pps->capcount;
 	(*pseq)++;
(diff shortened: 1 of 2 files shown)


delphij (approved):
This looks reasonable to me.

The parameter passed to `tc_get_timecount` was essentially a `this` pointer and shall be the same time counter object that the caller is querying.

bsdimp:
This looks reasonable to my eye as well... But I'm curious why you got rid of the atomic fences. What makes this safe now that they are gone?

sebhub:
I would like to update the pull request next week with a slightly improved version since we figured out why the existing code works. It is basically a combination of two bugs which lead to a working solution.

sebhub:
> This looks reasonable to my eye as well... But I'm curious why you got rid of the atomic fences. What makes this safe now that they are gone?

The acquire fence is still there:
```
	/* If the timecounter was wound up underneath us, bail out. */
	atomic_thread_fence_acq();
	if (pps->capgen == 0 || pps->capgen != capth->th_generation)
		return;
```
Compared to the existing code, the generation number (sequence lock) is only checked once.

The code is now similar to the other code blocks, for example:
```
static __inline void
getthmember(void *out, size_t out_size, u_int off)
{
	struct timehands *th;
	u_int gen;

	do {
		th = timehands;
		gen = atomic_load_acq_int(&th->th_generation);
		memcpy(out, (char *)th + off, out_size);
		atomic_thread_fence_acq();
	} while (gen == 0 || gen != th->th_generation);
}
```

sebhub:
I updated the pull request. The main update is in commit "Simplify the nsec calculation in pps_event()" which contains now also an explanation why the current approach worked.

bsdimp:
> I updated the pull request. The main update is in commit "Simplify the nsec calculation in pps_event()" which contains now also an explanation why the current approach worked.

Thanks! I'll look through this in the coming days...  I'm sick today, but I hope to get to it this week and let you know if I have questions.

joycemaferko:
Some additional context on the tests we ran:

Hardware: Cortex M4-based STM32F411 Discovery Board; u-blox NEO 6M GPS module; External GPS antenna.

Approach: Use a hardware unit to capture the time point of the PPS signal and convert it to a timecounter time point. Measure the delta between these values (in nanoseconds) over time. 

Here are two graphs which depict the control loop in action—one before the change (pps_error_original) and one after (pps_error_simplified). The results are essentially the same. The initial spike in the graphs is due to the phase offset, and this is gradually corrected. In both graphs, the delta values stabilize to 1billion (+/-  ~25ns) after around 50 seconds of run time.
pps_error_original
pps_error_simplified

sys/kern/kern_tc.c
```diff
@@ -1769,7 +1770,8 @@ pps_capture(struct pps_state *pps)
 #ifdef FFCLOCK
 	pps->capffth = fftimehands;
 #endif
-	pps->capcount = th->th_counter->tc_get_timecount(th->th_counter);
+	tc = th->th_counter;
+	pps->capcount = tc->tc_get_timecount(tc);
```
bsdimp, line 1774: This looks good to me.

sys/kern/kern_tc.c
```diff
@@ -1838,25 +1846,19 @@ pps_event(struct pps_state *pps, int event)
...
+	tcount = pps->capcount - tcount;
+	tcount &= captc->tc_counter_mask;
+	bintime_addx(&bt, capth_scale * tcount);
 	bintime2timespec(&bt, &ts);
 
-	/* If the timecounter was wound up underneath us, bail out. */
-	atomic_thread_fence_acq();
-	if (pps->capgen != pps->capth->th_generation)
-		return;
-
```
bsdimp, line 1857: I kinda think I understand why you can remove this check...  My hesitance suggests a question though: why is this OK? Is it because we (a) read everything before and (b) do the fence_acq() before? How do we know that the read above of the generation is done such that all the stores from other threads are now visible to us to detect the collision?

sys/kern/kern_tc.c
```diff
@@ -1769,15 +1770,16 @@ pps_capture(struct pps_state *pps)
 #ifdef FFCLOCK
 	pps->capffth = fftimehands;
 #endif
-	pps->capcount = th->th_counter->tc_get_timecount(th->th_counter);
-	atomic_thread_fence_acq();
-	if (pps->capgen != th->th_generation)
-		pps->capgen = 0;
```
sebhub, line 1775: Here we can remove the check because we have the same condition in the check in line 1811 (after changes).

sys/kern/kern_tc.c
```diff
@@ -1796,9 +1798,17 @@ pps_event(struct pps_state *pps, int event)
...
+	captc = capth->th_counter;
+	capth_scale = capth->th_scale;
+	tcount = capth->th_offset_count;
+	bt = capth->th_bintime;
+
 	/* If the timecounter was wound up underneath us, bail out. */
-	if (pps->capgen == 0 || pps->capgen !=
-	    atomic_load_acq_int(&pps->capth->th_generation))
+	atomic_thread_fence_acq();
+	if (pps->capgen == 0 || pps->capgen != capth->th_generation)
```
sebhub, line 1811: Here the generation number is checked once and all the captured state is saved to local variables.

sys/kern/kern_tc.c
```diff
@@ -1836,25 +1846,19 @@ pps_event(struct pps_state *pps, int event)
...
-	tcount = pps->capcount - pps->capth->th_offset_count;
-	tcount &= pps->capth->th_counter->tc_counter_mask;
-	bt = pps->capth->th_bintime;
-	bintime_addx(&bt, pps->capth->th_scale * tcount);
+	tcount = pps->capcount - tcount;
+	tcount &= captc->tc_counter_mask;
+	bintime_addx(&bt, capth_scale * tcount);
 	bintime2timespec(&bt, &ts);
 
-	/* If the timecounter was wound up underneath us, bail out. */
```
sebhub, line 1853: This check can be removed because we saved the captured state in local variables.

bsdimp:
> Some additional context on the tests we ran:

The old-time time-geek in me wants to see the graphs in steady state. The impulse response at startup looks quite good, though.  If it's super easy, can you regen the graphs from about 30-100s? If not, that's cool, it won't affect this change landing I don't think.

sys/kern/kern_tc.c
bsdimp: > This check can be removed because we saved the captured state in local variables.

OK. That all makes sense, and confirmed what I thought was going on, but hadn't studied the code enough to be positive about.

You've run this in a FreeBSD kernel, right?

sebhub: No, we didn't do the test runs on FreeBSD. We run the tests using the real-time operating system RTEMS. We ported the FreeBSD timecounters to RTEMS in 2015 while we added SMP support to RTEMS. During this work we also found some concurrency issues in the early implementations of the timecounters: https://github.com/freebsd/freebsd-src/commit/2c6946dca29a0481f1af751eda1ed9d97cdb92d5.

Our policy is to first integrate changes in FreeBSD and then back port them to RTEMS. We regularly synchronize the port with FreeBSD: https://github.com/RTEMS/rtems/commits/master/cpukit/score/src/kern_tc.c.

For the timecounters, we have a functional specification with validation tests: https://github.com/RTEMS/rtems-central/tree/master/spec/score/timecounter. This pull request is related to an attempt to also specify and validate the PPS synchronization. To simplify the specification it is beneficial to eliminate as many conditional expressions as possible. We also have to eliminate all dead code paths.

joycemaferko:
> > Some additional context on the tests we ran:
> 
> The old-time time-geek in me wants to see the graphs in steady state. The impulse response at startup looks quite good, though. If it's super easy, can you regen the graphs from about 30-100s? If not, that's cool, it won't affect this change landing I don't think.

Sure! Here are the two new graphs. On the Y axis, "0" is 1billion ns...you can see the values fluctuate between ~+/- 25ns from there.
pps_error_original_2
pps_error_simplified_2

bsdimp:
Thanks for helping me to understand it... and for the extra cool graph...
Landed... With a prototype version of my ghpr script...

sebhub:
Thanks for the review and integration.

PR #605, opened 2022-07-11
Correct jail_max_af_ips sysctl name

rwp0:
The sysctl is named `security.jail.jail_max_af_ips` with the `jail_` prefix in the final part.


bsdimp:
Looks good, verified in the kernel sources and sysctl output. Landed as cfd6acbb5591

PR #606, opened 2022-07-12
`-v` flag for ls, `versionsort(3)` for `scandir(3)`, and `strverscmp(3)`

obiwac:
Adds a `-v` flag for `ls`, which sorts entries following a more natural order, rather than lexicographically ordering as it usually does (e.g. "ent1 ent2 ent10" as opposed to "ent1 ent10 ent2").

This is done through a new `strverscmp(3)` function in libc, a GNU extension which I implemented by reading the glibc manual page for it. I took the opportunity to also add a `versionsort(3)` function for `scandir(3)` ;)

(A thorough review is probably necessary, because I'm not sure to have done everything the most "correct" way, but everything is working fine AFAICT and I wrote a few tests.)

I've requested my account to be activated for Phabricator, which will probably be a more appropriate place to upload my patch than a GitHub PR. I'll close this when done.

### DIFF
--- a/bin/ls/cmp.c
+++ b/bin/ls/cmp.c
@@ -65,6 +65,20 @@ revnamecmp(const FTSENT *a, const FTSENT *b)
 }
 
 int
+verscmp(const FTSENT *a, const FTSENT *b)
+{
+
+	return (strverscmp(a->fts_name, b->fts_name));
+}
+
+int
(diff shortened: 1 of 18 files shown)


obiwac:
[D35783](https://reviews.freebsd.org/D35783)

PR #608, opened 2022-07-19
Fix declaration of unmapped_buf

sebhub:
For architectures with a small-data area, the __read_mostly section must
present at the object declaration.

### DIFF
--- a/sys/sys/buf.h
+++ b/sys/sys/buf.h
@@ -522,7 +522,8 @@ extern int	bdwriteskip;
 extern int	dirtybufferflushes;
 extern int	altbufferflushes;
 extern int	nswbuf;			/* Number of swap I/O buffer headers. */
-extern caddr_t	unmapped_buf;	/* Data address for unmapped buffers. */
+extern caddr_t __read_mostly unmapped_buf; /* Data address for unmapped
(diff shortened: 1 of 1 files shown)


emaste:
CC @mjguzik

mjguzik:
I don't see how that would happen.

I have to note very few archs have linker script support for these annotations and I'm guessing this is why it's not working out for whatever arch you are looking at.

sebhub:
I don't use the FreeBSD kernel directly. I use a port of FreeBSD kernel code to RTEMS. I got an error on powerpc. The problem is when you don't add the __read_mostly to the declaration, then the compiler will generated small-data relocations for load/store instructions. In the definition the object is placed in the __read_mostly section and not the small-data area. If these two sections are to far away in the executable, then the linker cannot resolve the small-data relocation.

sebhub:
Is there a general issue with this patch? From my point of view it fixes an obvious issue for all targets with small-data area.

emaste:
@mjguzik even if this is unlikely to be an issue in practice on FreeBSD targets the declaration and definition should still match, yeah?

emaste:
Merged

PR #609, opened 2022-07-26
sys: rename preprocessed assembly-language files to .S

ehem:
Even though assembly-language .s files aren't supposed to be run through
cpp, clearly all of these are being preprocessed (#include is a strong
indicator).  A few potentially non-preprocessed files remain named .s.

Differential Revision: https://reviews.freebsd.org/D35908


ehem:
Simply rebased since the main branch got pushed.  Clearly all the checks passed though.

emaste:
This pull request was for the .s to .S rename; other changes should be in other pull requests? Anyhow no harm in grabbing other commits from here as you review/approve.

bsdimp:
Yea, they should be, but the overhead for the two I just landed in asking for them to be so was too large for me :)

bsdimp:
I've committed the last of the original changes from this that are no-brainers to commit.
What's the status of the rest of them? They appear to be in review, under active development, etc. That generally suggests that we close the PR unless there's a reason to keep it open...

bsdimp:
Closing this pull request... It's expanded somewhat beyond it's original remit... All changes that don't need discussion and/or additional work have landed.

PR #611, opened 2022-08-07
arc4random: Make rekey interval less predictable 

Keelan10:
This is from arc4random.c where the rekey interval is made to be less predictable.


jlduran:
Close by e9a2e4d1d28bf608a2ec915e25df9200af98e3b8?

PR #617, opened 2022-10-12
blacklistd: Reduce diffs with upstream

jlduran:
@emaste: Here are a few commits that further reduce the diffs with upstream blacklistd.

Thank you!

### DIFF
--- a/contrib/blacklist/libexec/blacklistd-helper
+++ b/contrib/blacklist/libexec/blacklistd-helper
@@ -81,7 +81,7 @@ add)
 		    /sbin/pfctl -a "$2/$6" -f -
 		# insert $ip/$mask into per-protocol/port anchored table
 		/sbin/pfctl -qa "$2/$6" -t "port$6" -T add "$addr/$mask" && \
-		    /sbin/pfctl -q -k $addr && echo OK
+		    /sbin/pfctl -qk "$addr" && echo OK
 		;;
 	esac
 	;;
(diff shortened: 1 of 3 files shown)


contrib/blacklist/libexec/blacklistd-helper
```diff
@@ -81,7 +81,7 @@ add)
 		    /sbin/pfctl -a "$2/$6" -f -
 		# insert $ip/$mask into per-protocol/port anchored table
 		/sbin/pfctl -qa "$2/$6" -t "port$6" -T add "$addr/$mask" && \
-		    /sbin/pfctl -q -k $addr && echo OK
+		    /sbin/pfctl -qk "$addr" && echo OK
```
emaste, line 84: All else being equal I like `-q -k` better, but even more than that I like avoiding diffs with upstream.

jlduran: 😂I had that exact same thought on the change below, but @zoulasc knows better.

zoulasc: he does not, he just followed suit with the rest of the code that combined options. he can change it.

jlduran: No worries. Thank you, we’ll abide!

jlduran:
We are only a couple of commits behind upstream, the overall spirit of this pull request has been achieved.
Thank you both @emaste and @zoulasc for your cooperation!

PR #618, opened 2022-10-16
fix xargs exit code when run in parallel mode

liu-du:
This fixes [bug #267110](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=267110) 

When xargs runs in parallel mode (using -P flag), it may sometimes return incorrect exit code. The following code shows current (incorrect) behaviour: 
* `/bin` directory exists, `/foo` directory does not exist. 
* `echo /bin /foo | xargs -n1 -P2 test -d` sometimes returns 0 sometimes returns 1:

```
root@freebsd:~ # test -d /bin
root@freebsd:~ # echo $?
0
root@freebsd:~ # test -d /foo
root@freebsd:~ # echo $?
1
root@freebsd:~ # echo /bin /foo | xargs -n1 -P2 test -d
root@freebsd:~ # echo $?
0
root@freebsd:~ # echo /bin /foo | xargs -n1 -P2 test -d
root@freebsd:~ # echo $?
1
root@freebsd:~ # echo /bin /foo | xargs -n1 -P2 test -d
root@freebsd:~ # echo $?
0
```

### DIFF
--- a/usr.bin/xargs/tests/regress.sh
+++ b/usr.bin/xargs/tests/regress.sh
@@ -1,6 +1,6 @@
 # $FreeBSD$
 
-echo 1..16
+echo 1..20
 
 REGRESSION_START($1)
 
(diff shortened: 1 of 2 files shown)


usr.bin/xargs/tests/regress.sh
```diff
@@ -21,4 +21,9 @@ REGRESSION_TEST(`0L', `xargs -0 -L2 echo <${SRCDIR}/regress.0.in')
 REGRESSION_TEST(`0P1', `xargs -0 -P1 echo <${SRCDIR}/regress.0.in')
 REGRESSION_TEST(`quotes', `xargs -n1 echo <${SRCDIR}/regress.quotes.in')
 
+REGRESSION_TEST_FREEFORM(`parallel1', `echo /    /    | xargs -n1 -P2 test -d; [ $? = 0 ]')
+REGRESSION_TEST_FREEFORM(`parallel2', `echo /    /foo | xargs -n1 -P2 test -d; [ $? = 1 ]')
+REGRESSION_TEST_FREEFORM(`parallel3', `echo /foo /    | xargs -n1 -P2 test -d; [ $? = 1 ]')
+REGRESSION_TEST_FREEFORM(`parallel4', `echo /foo /bar | xargs -n1 -P2 test -d; [ $? = 1 ]')
```
mjguzik, line 27: Names like foo and bar are definitely a non-starter. Instead you should use /var/empty for a directory which does exit and /var/empty/nodir (or whatever else of that sort) which does not.

mjguzik (commented):
commit message needs to be headed with: xargs: fix exit code when using -P
[or similar, key point being xargs: at the front]

Past that looks good, thanks for fixing.

usr.bin/xargs/tests/regress.sh
liu-du: thank you, fixed

liu-du:
thank you for review! I fixed the directory name and squashed the commits.

mjguzik:
What do you want in 'author'? As is github suggests "liu-du <duliujimmy@hotmail.com>"

liu-du:
thank you, yes that is fine

mjguzik:
ok, I'll tinker a little bit with it and commit tomorrow.

mjguzik:
landed in 0ca740d9a639ab635f5a28be9051d0124a9544a1

liu-du:
thank you @mjguzik

PR #619, opened 2022-10-22
Attempt to fix strfmon(3)

jlduran:
I believe `strfmon`'s behavior should be fully tested (therefore documented) and fixed.  I present my suggestions for you to pick:

1. Add a series of tests in order to avoid regressions and find edge cases.
2. A few _de rigueur_ white space/spelling fixes.
3. A series of fixes.
4. Documentation improvements (date NOT bumped).
5. There is a commit named DROP used only to run the tests.

`strfmon_l` is not working at the moment as it should. It will be addressed separately.

I believe the original `XXX` comments from commit 9d430a5991d3f64a75fee951a1efab3593207832 can be removed afterwards.

~~Herald~~ (me) added the last 3 committers: @emaste, @bsdjhb, @cemeyer.

PR: [267282](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=267282)

PR #620, opened 2022-10-29
Fix strfmon_l(3)

jlduran:
Following up #619, fix `strfmon_l` as well.

1. Minor improvements to `strfmon_l(3)` man page.
2. Correctness improvement to the current test.
3. Add a test for `strfmon_l`.
4. Fix `strfmon_l`.
5. Remove `XXX` marks.
6. DROP commit. Used only for testing using Cirrus CI runners.

Tagging @kostikbel, as he kindly reviewed #619.

PR: [267410](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=267410)

### DIFF
--- a/lib/libc/stdlib/strfmon.c
+++ b/lib/libc/stdlib/strfmon.c
@@ -71,9 +71,9 @@ __FBSDID("$FreeBSD$");
 		PRINT(*tmps++);					\
 } while (0)
 
-#define	GET_NUMBER(VAR) do {					\
+#define	GET_NUMBER(VAR, LOC) do {					\
 	VAR = 0;						\
-	while (isdigit((unsigned char)*fmt)) {			\
+	while (isdigit_l((unsigned char)*fmt, LOC)) {			\
 		if (VAR > INT_MAX / 10)				\
 			goto e2big_error;			\
 		VAR *= 10;					\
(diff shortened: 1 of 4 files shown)


lib/libc/stdlib/strfmon.c
```diff
@@ -128,7 +128,6 @@ vstrfmon_l(char * __restrict s, size_t maxsize, locale_t loc,
 
 	char		*tmpptr;	/* temporary vars */
 	int		sverrno;
-	FIX_LOCALE(loc);
```
kostikbel, line 131: Can you explain the move of FIX_LOCALE() call from there to strfmon_l()? In other words, why does strfmon() not need the fixup?

kostikbel (commented):
There are some style changes I make while looking at the code, all of them are line wraps.  Please apply them to the commit
[style](https://kib.kiev.ua/git/gitweb.cgi?p=deviant3.git;a=commit;h=e2df5ae5cd69930a08e1f0444dea471fe06637d3)

lib/libc/stdlib/strfmon.c
jlduran: Maybe I misinterpreted:

> Any public function that accepts a `locale_t` parameter must use the `FIX_LOCALE()` macro on it before using it.  For efficiency, this can be emitted (sic) in functions which *only* use their locale parameter as an argument to another public function, as the callee will do the `FIX_LOCALE()` itself.

And moved it to:

https://github.com/freebsd/freebsd-src/blob/70059838ae4ca5621c4c3c5cdf978a0db5bf838e/lib/libc/stdlib/strfmon.c#L662

I'll put it back to where it was before.

kostikbel: Yes, but you left strfmon() without FIX_LOCALE(), this is what I am asking about.

Even if the move is legitimate (and I misunderstand it), it should be a separate change with appropriate commit message explaining why.

jlduran: The original reasoning was:
1. `strfmon` uses `__get_locale()` to supply the `locale_t` parameter, no need to `FIX_LOCALE()`.
2. `strfmon_l` do need `FIX_LOCALE()`. So, move the `FIX_LOCALE()` to the corresponding function.

Anyhow, it's reverted back to `vstrfmon_l`, where it was originally.

kostikbel: Ok, please provide this move as a separate change then.  You might just mail me the git-format patch after this series goes in.

jlduran:
Closing. These changes have already been MFCd.
Thank you!

PR #621, opened 2022-10-29
xlocale(3): Link man pages

jlduran:
Essentially this is a followup of c889dd01b9ada00728601de06e8b087a8c751cbb. 

From `xlocale(3)`, this is just `ctype_l`. Others may need a similar treatment, but I'll wait for reviews of this one first.


jlduran:
Thank you!

PR #628, opened 2022-11-14
Fix possible overflow when comparing wide char types

AtariDreams:
Because wide char types is assumed to be an int, and the return type is an int, subtracting as unsigned integers can cause issues if the value in s1 is large while the value in s2 is small, as the result will overflow and the wrong result will be given.

Example: If *s1 evaluates to a number with its sign bit set, but *s2 evaluates to 1 or a small number, the result may still have its significant bit set, which will return a number smaller than 0 when interpreted as an integer.

### DIFF
--- a/lib/libc/string/wcscasecmp.c
+++ b/lib/libc/string/wcscasecmp.c
@@ -41,7 +41,7 @@ wcscasecmp(const wchar_t *s1, const wchar_t *s2)
 		c1 = towlower(*s1);
 		c2 = towlower(*s2);
 		if (c1 != c2)
-			return ((int)c1 - c2);
+			return (int)((unsigned int)c1 - (unsigned int)c2);
 	}
 	return (-*s2);
 }
--- a/lib/libc/string/wcscmp.c
+++ b/lib/libc/string/wcscmp.c
@@ -51,8 +51,8 @@ wcscmp(const wchar_t *s1, const wchar_t *s2)
 {
 
 	while (*s1 == *s2++)
-		if (*s1++ == '\0')
+		if (*s1++ == L'\0')
 			return (0);
-	/* XXX assumes wchar_t = int */
-	return (*(const unsigned int *)s1 - *(const unsigned int *)--s2);
+
+	return (int)((unsigned int)*s1 - (unsigned int)(*(s2 - 1)));
 }
--- a/lib/libc/string/wcsncasecmp.c
+++ b/lib/libc/string/wcsncasecmp.c
@@ -43,9 +43,9 @@ wcsncasecmp(const wchar_t *s1, const wchar_t *s2, size_t n)
 		c1 = towlower(*s1);
 		c2 = towlower(*s2);
 		if (c1 != c2)
-			return ((int)c1 - c2);
+			return (int)((unsigned int)c1 - (unsigned int)c2);
 		if (--n == 0)
 			return (0);
 	}
-	return (-*s2);
+	return (int)(-*s2);
 }
--- a/lib/libc/string/wcsncmp.c
+++ b/lib/libc/string/wcsncmp.c
@@ -47,12 +47,9 @@ wcsncmp(const wchar_t *s1, const wchar_t *s2, size_t n)
 	if (n == 0)
 		return (0);
 	do {
-		if (*s1 != *s2++) {
-			/* XXX assumes wchar_t = int */
-			return (*(const unsigned int *)s1 -
-			    *(const unsigned int *)--s2);
-		}
-		if (*s1++ == 0)
+		if (*s1 != *s2++)
+			return (int)((unsigned int)*s1 - (unsigned int)(*--s2));
+		if (*s1++ == L'\0')
 			break;
 	} while (--n != 0);
 	return (0);


lib/libc/string/wcscmp.c
```diff
@@ -50,9 +51,11 @@ int
...
 {
 
-	while (*s1 == *s2++)
-		if (*s1++ == '\0')
+	while (*s1 == *s2) {
+		if (*s1 == L'\0')
 			return (0);
+		++s1, ++s2;
+	}
 	/* XXX assumes wchar_t = int */
```
bsdimp, line 59: this could likely be removed.

lib/libc/string/wcscasecmp.c
```diff
@@ -41,7 +42,9 @@ wcscasecmp(const wchar_t *s1, const wchar_t *s2)
 		c1 = towlower(*s1);
 		c2 = towlower(*s2);
 		if (c1 != c2)
-			return ((int)c1 - c2);
+			return ((uintmax_t)c1 > (uintmax_t)c2 ? 1 : -1);
```
bsdimp, line 45: so why is the cast needed here? they are the same time, why can't the comparison be made without one?
This is odd-ball enough that a comment about the relatively obscure corner of 'C' that necessitates this would be good too.

lib/libc/string/wcsncmp.c
```diff
@@ -38,22 +38,21 @@ __RCSID("$NetBSD: wcsncmp.c,v 1.3 2001/01/05 12:13:13 itojun Exp $");
...
 wcsncmp(const wchar_t *s1, const wchar_t *s2, size_t n)
 {
 
-	if (n == 0)
-		return (0);
-	do {
-		if (*s1 != *s2++) {
+	for (; n; --n) {
+		if (*s1 != *s2) {
 			/* XXX assumes wchar_t = int */
```
bsdimp, line 50: This comment can also be removed. The code assumes that uchar is unsigned and the size of an int before, but after not so much. Same question / comment comment as above.

bsdimp:
I understand why we shouldn't subtract them, but I don't understand the uintmax_t patches.

lib/libc/string/wcscasecmp.c
AtariDreams: Answer: w_char is not guaranteed to be an integer, but yeah I think this could be done better and it will.

bsdimp:
I'm on the road the next couple of days, but first blush this answers my concerns.

lib/libc/string/wcsncmp.c
```diff
@@ -44,16 +44,14 @@ int
...
```
bsdimp, line 49: is this comment still relevant?

lib/libc/string/wcsncmp.c
```diff
@@ -44,16 +44,14 @@ int
...
-			return (*(const unsigned int *)s1 -
-			    *(const unsigned int *)--s2);
+			return (*s1 - *s2);
 		}
-		if (*s1++ == 0)
+		if (*s1 == L'\0')
 			break;
-	} while (--n != 0);
+		++s1, ++s2;
```
bsdimp, line 54: these belong in the for loop. That's more consistent with the style and now that they are unconditional, it's easy to move them there.

lib/libc/string/wcscmp.c
```diff
@@ -41,6 +41,7 @@ __RCSID("$NetBSD: wcscmp.c,v 1.3 2001/01/05 12:13:12 itojun Exp $");
 #endif /* LIBC_SCCS and not lint */
 __FBSDID("$FreeBSD$");
 
+#include <stdint.h>
```
bsdimp, line 44: I don't think this is needed here. I don't see any new intXX_t or PRxxx being introduced.

bsdimp:
OK. I've had most of my questions answered, there's still a few minor style issues to resolve.

bsdimp:
Also, the commit message is > 80 columns long and doesn't follow the conventions in https://docs.freebsd.org/en/articles/committers-guide/#commit-log-message

AtariDreams:
@bsdimp Fixed

AtariDreams:
Fixed up!

AtariDreams:
@bsdimp

jrtc27:
Why does the control flow need restructuring for all of these? Changes should be minimal, not gratuitous.

AtariDreams:
> Why does the control flow need restructuring for all of these? Changes should be minimal, not gratuitous.

Thank you for your feedback. I have adjusted my changes accordingly.

bsdimp:
Please separate the different changes here. There's type changes, flow control changes, etc. Please do each one separately in case one of them breaks something so we have to revert as little as possible.

AtariDreams:
Each case is proven to not break actually. They are functionally equivalent, but sure.

ngie-eign:

jrtc27:
You still have gratuitous changes

AtariDreams:

Done!

jrtc27 (changes requested):
This still needlessly messes with control flow. For the last time, keep your changes minimal.

AtariDreams:
> This still needlessly messes with control flow. For the last time, keep your changes minimal.

Done!

lib/libc/string/strncmp.c
```diff
@@ -46,7 +46,7 @@ strncmp(const char *s1, const char *s2, size_t n)
 	do {
 		if (*s1 != *s2++)
 			return (*(const unsigned char *)s1 -
-				*(const unsigned char *)(s2 - 1));
```
jrtc27, line 49: Gratuitous

lib/libc/string/wcscasecmp.c
```diff
@@ -41,7 +41,7 @@ wcscasecmp(const wchar_t *s1, const wchar_t *s2)
 		c1 = towlower(*s1);
 		c2 = towlower(*s2);
 		if (c1 != c2)
-			return ((int)c1 - c2);
```
jrtc27, line 44: Some architectures have wchar_t as int, not unsigned int, and so are still identical to the original code here. Presumably to be 100% free from UB this should be performing the arithmetic on unsigned ints and then (implicitly?) casting back to int.

lib/libc/string/wcscmp.c
```diff
@@ -49,10 +49,9 @@ __FBSDID("$FreeBSD$");
 int
 wcscmp(const wchar_t *s1, const wchar_t *s2)
 {
-
```
jrtc27, line 52: Gratuitous

lib/libc/string/wcscmp.c
```diff
@@ -49,10 +49,9 @@ __FBSDID("$FreeBSD$");
...
 	while (*s1 == *s2++)
-		if (*s1++ == '\0')
```
jrtc27, line 54: Gratuitous (yes, maybe better, but this is not a "possible overflow when comparing wide char types")

lib/libc/string/wcscmp.c
```diff
@@ -49,10 +49,9 @@ __FBSDID("$FreeBSD$");
...
+		if (*s1++ == L'\0')
 			return (0);
-	/* XXX assumes wchar_t = int */
-	return (*(const unsigned int *)s1 - *(const unsigned int *)--s2);
```
jrtc27, line 57: This is worse, it's introducing signed overflow on architectures with wchar_t typedef'ed to (signed) int

lib/libc/string/wcsncasecmp.c
```diff
@@ -43,7 +43,7 @@ wcsncasecmp(const wchar_t *s1, const wchar_t *s2, size_t n)
 		c1 = towlower(*s1);
 		c2 = towlower(*s2);
 		if (c1 != c2)
-			return ((int)c1 - c2);
```
jrtc27, line 46: Same as wcscasecmp

lib/libc/string/wcsncmp.c
```diff
@@ -43,16 +43,12 @@ __FBSDID("$FreeBSD$");
 int
 wcsncmp(const wchar_t *s1, const wchar_t *s2, size_t n)
 {
-
```
jrtc27, line 46: All the same comments as before: gratuitous changes and regresses wchar_t-is-int architectures

jrtc27 (changes requested):

AtariDreams:
@jrtc27 Fixed!

lib/libc/string/wcscmp.c
jrtc27: Not sure why you think this is resolved when you've neither replied nor dropped the change? IMO this change is fine but either it should be in a separate commit or this one should have its message altered to reflect the actual changes.

lib/libc/string/wcscmp.c
```diff
@@ -51,8 +51,8 @@ wcscmp(const wchar_t *s1, const wchar_t *s2)
 {
 
 	while (*s1 == *s2++)
-		if (*s1++ == '\0')
+		if (*s1++ == L'\0')
 			return (0);
-	/* XXX assumes wchar_t = int */
```
jrtc27, line 56: This still assumes wchar_t fits in an (unsigned) int, but I guess other places don't have an XXX so it's fine to drop...

lib/libc/string/wcscmp.c
```diff
@@ -51,8 +51,8 @@ wcscmp(const wchar_t *s1, const wchar_t *s2)
...
-	return (*(const unsigned int *)s1 - *(const unsigned int *)--s2);
+
+	return (int)((unsigned int)*s1 - (unsigned int)(*(s2 - 1)));
```
jrtc27, line 57: The --s2 is stupid code but better to leave it as is for this change, especially if you're not changing wcsncmp to match

lib/libc/string/wcscasecmp.c
```diff
@@ -41,7 +41,7 @@ wcscasecmp(const wchar_t *s1, const wchar_t *s2)
...
+			return (int)((unsigned int)c1 - (unsigned int)c2);
```
jrtc27, line 44: style(9) requires parentheses around the entire expression including the cast, though IDK if we prefer the explicit cast like this or just let it be implicitly cast on return

lib/libc/string/wcscmp.c
```diff
@@ -51,8 +51,8 @@ wcscmp(const wchar_t *s1, const wchar_t *s2)
 {
 
 	while (*s1 == *s2++)
-		if (*s1++ == '\0')
+		if (*s1++ == L'\0')
 			return (0);
-	/* XXX assumes wchar_t = int */
-	return (*(const unsigned int *)s1 - *(const unsigned int *)--s2);
+
```
jrtc27, line 56: Don't insert random blank lines

lib/libc/string/wcsncasecmp.c
```diff
@@ -43,9 +43,9 @@ wcsncasecmp(const wchar_t *s1, const wchar_t *s2, size_t n)
...
+			return (int)((unsigned int)c1 - (unsigned int)c2);
 		if (--n == 0)
 			return (0);
 	}
-	return (-*s2);
+	return (int)(-*s2);
```
jrtc27, line 50: This is totally unnecessary, there's already an implicit cast here to int...

lib/libc/string/wcscmp.c
bsdimp: Also style(9) requires a set of parens around this whole thing.

bsdimp:
This has been in the changes requested state for months. I'm closing it. If the requested changes are made, it can be reopened and/or another review opened.

PR #631, opened 2022-11-26
Give every printf like function __restrict for the format value

AtariDreams:
The format value has to be a string literal, every time. Otherwise, you are not using these functions correctly. To reinforce this fact, I put __restrict over every non-contrib example of this I could find.


AtariDreams:
@bsdimp Could you please look at this when you have the time?

igalic:
passing on some feedback from IRC:

Probably most cases there should indeed use bare `restrict`.

But anything that ends up in `/usr/include` needs to be C89 compatible, and thus needs `__restrict`

bsdimp:
This change is impossibly big to look at. there's no way I can give meaningful feedback to this. Can you break it down to something smaller? Like maybe batches of 15 or so files? This is also the sort of change that has low value, but high risk of one of these screwing up something subtle we won't find out about for a while...

bsdimp:
Closing pull request: It needs to be split up better please. 372 files is impossible to review as one chunk. It would be better to create multiple pull requests that have 10-20 files each in them, one commit per directory with a good commit message in each of them. Otherwise, this would take me too long to make conforming. Thanks!

PR #640, opened 2023-01-25
Clarify hardpps() parameter name and comment

sebhub:
Since the following commit

commit 32c203577a5e89ed1e5849d1a5e81abf501cfb50
Author: Poul-Henning Kamp <phk@FreeBSD.org>
Date:   Thu Mar 11 15:09:51 1999 +0000

    Make even more of the PPSAPI implementations generic.

    FLL support in hardpps()

    Various magic shuffles and improved comments

    Style fixes from Bruce.

the "nsec" parameter of hardpps() is a time difference and no longer a time point.  Change the name to "delta_nsec" and adjust the comment.

Remove comment about a clock tick adjustment which is no longer in the code.

### DIFF
--- a/sys/kern/kern_ntptime.c
+++ b/sys/kern/kern_ntptime.c
@@ -733,11 +733,11 @@ hardupdate(offset)
  * variables, except for the actual time and frequency variables, which
  * are determined by this routine and updated atomically.
  *
- * tsp  - time at PPS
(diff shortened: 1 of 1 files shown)


bsdimp:
This looks good to me.

bsdimp:
pulled in as c7c53e3ca631

PR #643, opened 2023-02-09
uniq(1): use strtonum to parse options

daniel-tameling:
Previously strtol was used and the result was directly cast to an int without checking for an overflow. 
So one can do things like
`echo 'hello\nbello' | uniq -s 4294967296`
which is equivalent to
`echo 'hello\nbello' | uniq -s 0`
With the program aborts with:
`uniq: character skip value is too large: 4294967296`


bsdimp:
This looks good to me.

bsdimp:
Landed as e052829e3e16dfd82d0adcbb69fd0e30f47a3a6c

PR #644, opened 2023-02-09
ADD x86 MSR Data Operand Independent Timing Mode

paepckehh:
Add msr register MSR_IA32_UARCH_MISC_CTL 
available in Intel Skylake and later CPUs.

This register will be needed in a next step to be able to mitigate a well know  Intel CPU vulnerability. 

https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/best-practices/data-operand-independent-timing-isa-guidance.html

This is a first step to address.
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=269219

The register is there. So add it. Allow to disable it if needed individually.  How to handle it in general in FreeBSD (buildtime/runtime/defaults) may need a some discussion.

### DIFF
--- a/sys/x86/include/specialreg.h
+++ b/sys/x86/include/specialreg.h
@@ -544,6 +544,7 @@
 #define	MSR_IA32_FEATURE_CONTROL 0x03a
 #define	MSR_IA32_SPEC_CTRL	0x048
 #define	MSR_IA32_PRED_CMD	0x049
+#define MSR_IA32_UARCH_MISC_CTL	0x00001b01 /* Data Operand Independent Timing Mode */
 #define	MSR_BIOS_UPDT_TRIG	0x079
 #define	MSR_BBL_CR_D0		0x088
 #define	MSR_BBL_CR_D1		0x089


bsdimp:
hmmm... this just adds the constant to the specialregs.h file. What makes it useful from there? I'm not sure I understand what's here, so maybe something is missing?

paepckehh:
>  [...] so maybe something is missing?

Yes, the decision of the FreeBSD Core Team how to handle the default value :) 

I'm afraid that thos discussion (similar to Linux, since mid 2022)  could take a while. So I wanted to reflect the (unchangeable, fixed) reality that hardware defines here and inspire people to explore the full solution-space we have here (sysctl/kern-userland-split-brain/build-time-options/app-specific-control).

My personal favorite (eg. personal plan) is to just pin it down to zero at build-time, +1 LOC. Done. This would discourage this kind of behavour in harware design, or at least encourage better designs (eg. exclude AES-NI) in the future.

But looking at the Linux Team discussion, there is a lot of different solutions possible. So my hope was, providing the basic framwork (and even beating linux here) could attract some attention and spark some better ideas and code here, than just statically nail it down.

paepckehh:
> Are you talking to the core team?

Ed Maste subscribed to the incident. I don't know the exact technical definition of the FreeBSD Foundation,  but you are the a well known core (technical term) committer.  From my POV, whatever the final solution is, this ancor is needed - anyhow. Rolling it back and forth from the main includes with every future solution approach/rollback is avoidable.

The main intention of this commit is to **start** ( and enable ) the search for a solution. It will not hurt anyone and may spark some ideas/attention.

My personal approach does not fit for everyone. (There is a wide range of EFI/UEFI/BIOS/ME solutions that lock and monitor the MSR register in the background). Others might be unhappy about the performance hit. And just permanently ignore it might be not a good approach either. Intel already announced keep the option open to push this also to older architectures via microcode updates - and this may break existing validated security equipment assumptions.

PR #645, opened 2023-02-10
add Huawei E3372h NCM Mode DeviceID and UPDATING entry

paepckehh:
* add missing entry for E3372h Model already in NCM mode
reported by: David BOYER

* add updating section to help other users to migrate


bsdimp:
fixed with c92790b3fb9f push

bsdimp:
there was some trailing whitespace in the UPDATING file, but I tweaked that before committing

PR #646, opened 2023-02-10
add to cdce.4 section about NCM network mode

paepckehh:
add a new section how to use NCM network mode and specify the required kernel modules


bsdimp:
Fixed with push c92790b3fb9f

PR #647, opened 2023-02-15
Allow to manage KBD_DELAY[1|2] 

paepckehh:
Allow to configure KBD_DELAY[1|2] via KERNCONF to bypass (strict) fixed value option sets within kbdcontrol and allow to manage without user land for embedded/security appliances, similar to the other existing KBD_ kernel options.

If this gets accepted, I can add a new PR to enhance kbdcontrol as well. 
Promise! You will love your new snappy & fast interactive console setup options!

No change regarding the default behavior,  anything pure optional.


bsdimp:
Usually there's not this extra layer of indirection. Usually kernel options are exactly what's used in the kernel with the
```
#ifndef FOO_BAR
#define FOO_BAR 27
#endif 
```
Having KBD_DELAY1 vs KD_DELAY1 seems needlessly confusing given this pattern. It's also harder to grep for the effect of indirect options.

paepckehh:
> Usually there's not this extra layer of indirection. Usually kernel options are exactly what's used in the kernel with the
> 
> ```
> #ifndef FOO_BAR
> #define FOO_BAR 27
> #endif 
> ```
> 
> Having KBD_DELAY1 vs KD_DELAY1 seems needlessly confusing given this pattern. It's also harder to grep for the effect of indirect options.

Yes, I agree with you! But the existing interface for exposed keyboard kernel options clearly uses KBD_ as a prefix. I tried to avoid confusion by introducing another prefix, KB_, here.

https://github.com/freebsd/freebsd-src/blob/main/sys/conf/options

```
[...]
# kbd options
KBD_DISABLE_KEYMAP_LOAD	opt_kbd.h
KBD_INSTALL_CDEV	opt_kbd.h
KBD_MAXRETRY		opt_kbd.h
KBD_MAXWAIT		opt_kbd.h
KBD_RESETDELAY		opt_kbd.h
KBDIO_DEBUG		opt_kbd.h
[...]
```

There is no doubt that the right thing to do is to migrate KB_DELAY* to KBD_DELAY* and then expose KBD_DELAY* without any indirect intermediaries. I would really like to do that. There are currently four instances where KB_DELAY is used in /sys. Adding it to options would lead to six modifications in two commits, so a slightly bigger commit.

If you agree, I would love to cancel this PR and open another one for the complete solution. As KB_DELAY* is not exposed via options yet, there should be no fallout from a cleanup.

bsdimp:
This looks good. Two things would make it perfect.

(1) Add an entry to UPDATING. KD_DELAYx wasn't an official option, so this should be enough to prevent POLA
(2) Add documentation to share/man/man4/atkbdc.4 similar to what's there for the other kernel options.

PR #648, opened 2023-02-16
310.accounting: Verify process accounting is active before log rotation.

jgrafton:
Check if process accounting is active in the periodic script.  If not, exit the periodic script indicating process accounting is not active instead of attempting to rotate the logs.  onerotate_log turns accounting back on regardless of what accounting_enable in rc.conf is set to.

PR: 267464


bsdimp:
This looks good to me.

bsdimp:
https://reviews.freebsd.org/D37434 has another 'looks good' but wanted more review.

igalic (approved):
I like the description in the Phabricator review much better, but regardless, 👍

PR #651, opened 2023-02-20
netlink: Fix "version introduced" documentation

tomhukins:
netlink(4) and associated features will exist in FreeBSD 14.0 but they will also exist in 13.2, an older version, from commits such as 02b958b1 and b309249b.

This commit needs merging to stable/13 and releng/13.2.

### DIFF
--- a/share/man/man3/snl.3
+++ b/share/man/man3/snl.3
@@ -297,7 +297,7 @@ and
 The
 .Dv SNL
 library appeared in
-.Fx 14.0 .
+.Fx 13.2 .
 .Sh AUTHORS
 This library was implemented by
 .An Alexander Chernikov Aq Mt melifaro@FreeBSD.org .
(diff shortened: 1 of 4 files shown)


AlexanderChernikov:
Thank you, Tom! I indeed missed the fact that netlink manpages should be updated to reflect it's appearance in 13.2.
I'll keep this change request open till I get the commit landed in 13.2.

bsdimp:
I've merged this and will make sure it gets into 13.2. So closing to reduce clutter...

bsdimp:
> Thank you, Tom! I indeed missed the fact that netlink manpages should be updated to reflect it's appearance in 13.2. I'll keep this change request open till I get the commit landed in 13.2.

Ah, looks like this was committed to stable/13 directly... I've committed it to main and can do the merge request to re@ for releng/13.2. I have that staged locally.

tomhukins:
The [13.2 release schedule](https://www.freebsd.org/releases/13.2R/schedule/) suggests that RC2 is due today so it would be great if either @bsdimp or @AlexanderChernikov could check again with re@ about getting this merged soon.

Apologies for pestering, but I think it would be helpful to provide correct information in 13.2-RELEASE.

bsdimp:
I'd planned on submitting it today since a number of work things commanded my attention this past week..

PR #652, opened 2023-02-21
sbin: Fix some typos

ElyesH:


igalic:
given that none of these typos have much in common, you should split this PR into multiple commits for each sub-directory, with each of them being a logical unit, as per https://docs.freebsd.org/en/articles/contributing/#contrib-general

emaste:
As an example I grabbed the typo fixes under include/ and pushed them as ce7db385f50a

bsdimp:
I'm planning on landing this pull request with a new script that marks each commit with the meta data needed to track it and have those hashes in freeBSD's repo show up here automatically.

bsdimp:
Ed and I chatted about this and we'll likely combine some of the individual commits to balance between too many changes clogging our email over typos vs too large of a change making MFC hard.

ElyesH:
please see https://github.com/freebsd/freebsd-src/pull/653 
Thx

PR #653, opened 2023-02-22
Fix some typos

ElyesH:


sbin/nvmecontrol/nvmecontrol.h
```diff
@@ -85,7 +85,7 @@ void print_intel_add_smart(const struct nvme_controller_data *cdata __unused, vo
  * 128-bit integer augments to standard values. On i386 this
  * doesn't exist, so we use 64-bit values. So, on 32-bit i386,
  * you'll get truncated values until someone implement 128bit
```
igalic, line 87: implements

igalic (approved):
👍

hope nobody's cli tooling relies on the typos in the outputs that are fixed now 🤣

emaste:
> hope nobody's cli tooling relies on the typos in the outputs that are fixed now rofl

We should actually look in some detail at the changes in user-facing messages. Typos in comments we can fix without any worry.

igalic:
if we break this pr into two: fix typos in comments, and fix typos in output, that would make assessment a little easier

bsdimp:
I've looked at the code changes, and they are fine, including the user facing messages. they are all error messages and I don't think anybody parses them.
I have a staging branch that I'm building before pushing.

bsdimp:
landed ed96987b4cfa

PR #654, opened 2023-02-22
usr.sbin: Remove trailing semicolon

ElyesH:
Macros shouldn't use trailing semicolon.


bsdimp:
Please separate this into logical units: bhyve, lpr, pmccontrol and ppp as separate commits. I'll push that.

Since this was submitted, we've added a note about strongly suggesting style changes be done only when prepping additional work in an area, so I'll let this one in under the old rules, but please consider that in the future.

bsdimp:
This looks good. I've merged it.

PR #655, opened 2023-02-22
Add a check that the alignment requested by memalign is not zero

paulfloyd:
If it is zero then the current code will trigger a division by zero SIGFGE and terminate.

See https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=269688

### DIFF
--- a/lib/libc/gen/memalign.c
+++ b/lib/libc/gen/memalign.c
@@ -35,5 +35,12 @@ __FBSDID("$FreeBSD$");
 void *
 memalign(size_t align, size_t size)
 {
-	return (aligned_alloc(align, size ? roundup(size, align) : size));
+	/*
+	 * if align is zero then we can't roundup
+	 * in that case leave it to malloc
+	 */
+	if (align != 0)
+		return (aligned_alloc(align, roundup(size, align)));
+	else
+		return (malloc(size));
 }
--- a/lib/libc/gen/memalign.c
+++ b/lib/libc/gen/memalign.c
@@ -35,5 +35,5 @@ __FBSDID("$FreeBSD$");
 void *
 memalign(size_t align, size_t size)
 {
-	return (aligned_alloc(align, roundup(size, align)));
+	return (aligned_alloc(align, size ? roundup(size, align) : size));
 }


lib/libc/gen/memalign.c
```diff
@@ -35,5 +35,5 @@ __FBSDID("$FreeBSD$");
 void *
 memalign(size_t align, size_t size)
 {
-	return (aligned_alloc(align, roundup(size, align)));
+	return (aligned_alloc(align, size ? roundup(size, align) : size));
```
emaste, line 38: probably `size != 0 ? ...` for style(9)

emaste:
I've staged it in my wip branch: https://github.com/emaste/freebsd/commit/2359351d53d4b0c5344dc255cc422f2f7291664d

bsdimp:
Is freebsd:#655 enough? In the past I've used the full url to distinguish between different possible pull request sources

emaste:
> Is freebsd:#655 enough? In the past I've used the full url to distinguish between different possible pull request sources

The commit contains the full URL; GitHub rendered a munged version.

emaste (changes requested):
the faulting condition is with `align = 0` not size; a reasonable condition might be
```
if (align == 0)
        align = 1;
```

paulfloyd:
Sorry for being hasty and getting most things wrong!

return (aligned_alloc(align, align ? roundup(size, align) : 1));

would probably be better. I don't think that the value after the : matters much, aligned_alloc will return NULL and set EINVAL.

paulfloyd:
Long context:
I'm working on overhauling the handling of aligned (and sized) allocations in Valgrind. On the horizon (C23) are new sized and sized/aligned free functions (not implemented yet afaik). In preparation for those I'd like to clear up the existing aligned allocation functions.

At the moment Valgrind just does more or less what GNU libc does. It has one test for memalign which is badly written (it doesn't cover the error described below and doesn't check for NULL returns or EINVAL)

It doesn't help that GNU libc is also a mess so it isn't a very good reference. The GNU libc manpage says

```
The obsolete function memalign() allocates size bytes and returns a pointer to the allocated memory.  The memory address will be a multiple of alignment, which must be a power of two.
```
That is almost entirely not what GNU libc does

- If the alignment is less than the natural alignment of `malloc` [2*sizeof(size_t_] then just call malloc
- If alignment is less than MINSIZE [16/32 for i386/amd64] then set it to MINSIZE
- If the alignment is huge (over half the size of the address space) then set EINVAL and return NULL
- If the alignment isn't a power of two, bump it up to the next power of two

What do other platforms do?

macOS doesn't implement `memalign`. Easy!

Solaris.
```
       The memalign() function allocates size bytes on a  specified  alignment
       boundary and returns a pointer to the allocated block. The value of the
       returned address is guaranteed to be an even multiple of alignment. The
       value  of  alignment must be a power of two and must be greater than or
       equal to the size of a word.
```

This time Sun lib C does do just about what the man page says. It does work for an alignment of 4 on amd64 which isn't quite what I'd expect but it does fail for 0, less than 4 and non-multiples of 2.

Illumos hasn't changed the above.

Finally there is MUSL libc. This allows a size of 0 but not any non-zero non-multiple of 2.

Unfortunately that's not very conclusive.

At the end of the day I'd say that there are two possibilities

- just pass an alignment of zero on to `aligned_alloc` which will return NULL and set errno to EINVAL which would be more like Solaris/Illumos
- just call `malloc` with the size which would be more like glibc and musl

The other 3 I've described are incompatible so I don't see much point in aiming for full compatibility with any of them.

paulfloyd:
And here's the glibc implementation

https://elixir.bootlin.com/glibc/glibc-2.37/source/malloc/malloc.c#L3504

Illumos

https://github.com/illumos/illumos-gate/blob/master/usr/src/lib/libc/port/gen/memalign.c

MUSL

https://github.com/esmil/musl/blob/master/src/malloc/memalign.c

mjguzik:
The most pragmatic option to is to be least glibc-incompatible, iow if I'm reading this right it would be just call malloc for 0 size.

paulfloyd:
I'd be happy with something like

```
if (align != 0)
   return (aligned_alloc(align, roundup(size, align)));
else
   return malloc(size);
```

That leaves aligned_alloc to handle everything but the alignment of zero.

mjguzik:
looks good to me, I doubt anyone is going to object.

if you post that patch it has 'Reviewed by: mjg' from me. since someone else got involved (emaste@) i'm gonna wait for him to coment just in case.

emaste:
I'd be fine with the patch a couple of comments above. If you refresh the pull request @mjguzik or I can bring it in.

bsdimp:
I'm also fine with the above suggestion... I'll leave it to others to land this.

emaste:
I staged a version in https://github.com/emaste/freebsd/commit/11506ebd6bf75e4a7e82ddbb50a7258a66f53849 (waiting for CI) but can grab yours instead and push a few hours from now.

mjguzik:
commit message should note this is glibc-compatible. a sentence to that extent can be added without you pushing.

emaste:
Merged to main thank you

PR #656, opened 2023-02-23
Useless return for void function

ElyesH:


emaste:
Please prefix the commit subject line with the functional area, e.g. for https://github.com/freebsd/freebsd-src/pull/656/commits/982dd0439ef584e19905748020949a8da9643ae8 could be `bhyve: Remove useless return at the end of void function`

igalic (approved):
👍

bsdimp:
This is a good test of my pull request script :)
While the changes are correct, this PR is definitely on that is just style for style sake and doesn't appear to be preparatory for additional work. while we've not formalized policies around this yet, we don't want to get overwhelmed with style commits and miss something due to the volume. It's a good learning experience, though, for both of us so I'm going to go ahead and land this PR.

PR #657, opened 2023-02-23
veriexec: fix declaration without a prototype

igalic:
veriexec_usage takes no parameters, so be explicit about it.

Reported by: gbe@

### DIFF
--- a/sbin/veriexec/veriexec.c
+++ b/sbin/veriexec/veriexec.c
@@ -50,7 +50,7 @@ const char *Cdir = NULL;
  * @return always returns code 0
  */
 static int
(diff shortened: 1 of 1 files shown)


gbergling:
I can confirm that the bugfix is working.

mjguzik:
i don't think changes like this need a review. but if you are worried, freel free to commit with "Reviewed by: mjg"

mjguzik (approved):
lemme officially slap it

mjguzik:
the change landed

igalic:
with a much better commit message, too!

PR #660, opened 2023-02-23
add altlog_jaillist to syslogd's rc script

igalic:
syslogd has an altlog_proglist stanza, which can place the log socket into a program's chroot directory, so it can still log.

This patch adds an equivalent to that for jails.
If used with jails, we lookup the each jail's "path" with jls appending it to the sockfile and, now also privsockfile.

Caveats:

- syslogd must be restarted
- -H must be added to syslogd_flags before this is actually useful

Differential Revision: https://reviews.freebsd.org/D27411


igalic:
Opening this here as well, to get more eyes on it.

igalic:
closing this based on @mjguzik's feedback on Phabricator.

I'm going to have to rethink this completely.

PR #661, opened 2023-02-23
kldxref: skip .pkgsave files

igalic:
This should help people transitioning from traditional setups to pkgbase experience a lot less friction.

We do this by skipping all files containing two dots.

Differential Revision: https://reviews.freebsd.org/D27959


igalic:
Opening this here as well to get more eyes / opinions on it.

bsdimp:
Seems good to me

bsdimp:
landed as 773c13c686e4b6ae9dbbc150b342b82c3f47d73a

PR #662, opened 2023-02-23
rc: ignore .pkgsave files

igalic:
the local parts of rc already skip .sample files;
we add .pkgsave to the list, and add logic for base

thanks to @RhodiumToad for getting this started.

Differential Revision: https://reviews.freebsd.org/D27962

### DIFF
--- a/libexec/rc/rc
+++ b/libexec/rc/rc
@@ -94,7 +94,9 @@ fi
 # Do a first pass to get everything up to $early_late_divider so that
 # we can do a second pass that includes $local_startup directories
 #
-files=`rcorder ${skip} ${skip_firstboot} /etc/rc.d/* 2>/dev/null`
+unset system_rc
+find_system_scripts
+files=`rcorder ${skip} ${skip_firstboot} ${system_rc} 2>/dev/null`
 
 _rc_elem_done=' '
 for _rc_elem in ${files}; do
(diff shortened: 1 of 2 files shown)


igalic:
This revision seems to have been forgotten, so I'm opening it here as well to get more eyes / opinions on it.

igalic:
with https://github.com/freebsd/freebsd-src/pull/661 merged, maybe we can take a look at this now, too?

bsdimp:
Is there some reason you didn't say system_rc=$(find_system_scripts) instead of the global variable... It makes it simpler.

jillest:
> Is there some reason you didn't say system_rc=$(find_system_scripts) instead of the global variable... It makes it simpler.

I can see some reasons:

- The above `find_local_scripts_new` returns a result in a variable in the same way.
- The `$()` syntax runs the function in a subshell which might have both performance and correctness implications. For example, an error will truncate the list at the point of the error but will not cause the outer script to abort.

RhodiumToad:
> Is there some reason you didn't say system_rc=$(find_system_scripts) instead of the global variable... It makes it simpler.

I just copied the existing style for ```find_local_scripts_new```

Probably best to check whether there was some particular reason why that function does it before changing both.

bsdimp:
This looks in decent shape. Only one issue in the review still pending, right?

igalic:
merged in https://github.com/freebsd/freebsd-src/commit/3693d9140e05aba9942232df13468f51a6cde136

PR #663, opened 2023-02-24
jail: convert several functions from int to bool

igalic:
these functions exclusively return (0) and (1), so convert them to bool

We also convert some networking related jail functions from int to bool some of which were returning an error that was never used.

Differential Revision: https://reviews.freebsd.org/D29659

### DIFF
--- a/sys/sys/jail.h
+++ b/sys/sys/jail.h
@@ -413,24 +413,24 @@ struct vfsconf;
  */
 #define jailed(cred)	(cred->cr_prison != &prison0)
 
-int jailed_without_vnet(struct ucred *);
+bool jailed_without_vnet(struct ucred *);
 void getcredhostname(struct ucred *, char *, size_t);
 void getcreddomainname(struct ucred *, char *, size_t);
 void getcredhostuuid(struct ucred *, char *, size_t);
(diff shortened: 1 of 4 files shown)


igalic:
An earlier version of this Pull Request was approved by jamie@, but it was never merged.
Rebasing and putting on GitHub to get more eyes on it again.

sys/sys/jail.h
```diff
@@ -413,17 +413,17 @@ struct vfsconf;
  */
 #define jailed(cred)	(cred->cr_prison != &prison0)
 
-int jailed_without_vnet(struct ucred *);
+boolt jailed_without_vnet(struct ucred *);
```
jlduran, line 416: There seems to be a typo here. Should be just `bool`, not `boolt`.

igalic: thanks for catching that

fixed!

jlduran: You are welcome, now we just need the changes to make it compile.

igalic: looking okay now.
And would look a lot better with #668 merged.

ngortheone (approved):

bsdimp:
looks good... sorry it took so long to get into the tree...

PR #664, opened 2023-02-24
apic: prevent divide by zero in CPU frequency init

igalic:
If a CPU for some reason returns 0 as CPU frequency, we currently panic on the resulting divide by zero when trying to initialize the CPU(s) via APIC

PR: 269767

### DIFF
--- a/sys/x86/x86/local_apic.c
+++ b/sys/x86/x86/local_apic.c
@@ -953,7 +953,7 @@ lapic_calibrate_initcount_cpuid_vm(void)
 
 	/* Record divided frequency. */
 	count_freq = freq / lapic_timer_divisor;
-	return (true);
+	return (count_freq != 0);
 }
 
 static uint64_t


igalic:
this fix has been suggested by @dwmw2.
I'm opening it as draft pull request to test it first

igalic:
tested, boots

This now produces a lock order reversal in `if_virtio`, and 100% CPU load on one CPU by `rand_harvest`, but it boots.

igalic:
merged in https://github.com/freebsd/freebsd-src/commit/499171a98c8813e4dc6e085461d5c47750efa555

dwmw2:
This sanity check on the data obtained from CPUID 0x40000010 is sensible enough, but I have questions...

Why do we assume that 0x40000010 even exists? There's code elsewhere to look at the 'max leaf' we get back from 0x40000000 and store that, but I don't see us checking that before assuming that it's at least 0x10? 

Also, what if the KVM hypervisor leaves aren't at 0x40000000? If Hyper-V is present, *it* will be at 0x40000000 and the KVM leaves will be shifted up to 0x40000100 (thus won't we need to use 0x40000110 for this?).

bsdimp:
@dwmw2 Followup commits to make this code more correct are welcome.  We likely should check that to make sure it's valid.

PR #667, opened 2023-02-24
[WIP] daemon improvements

ngortheone:
Code readability and consistency improvements. The change is broken down into small commits, each commit has a commit message with rationale for the change.

cc @kevans91 @bsdimp

### DIFF
--- a/cddl/contrib/opensolaris/cmd/lockstat/sym.c
+++ b/cddl/contrib/opensolaris/cmd/lockstat/sym.c
@@ -42,6 +42,7 @@
 #include <libelf.h>
 #include <link.h>
 #include <elf.h>
+#include <gelf.h>
 #ifdef illumos
 #include <sys/machelf.h>
 
@@ -53,7 +54,6 @@
 #include <sys/linker.h>
 #endif
 #include <sys/cpuvar.h>
-
 
 typedef struct syment {
 	uintptr_t	addr;
@@ -72,11 +72,6 @@ static char maxsymname[64];
 #define	elf_getshdr elf32_getshdr
 #endif
 #endif
-
-#define __sElfN(x)       typedef __CONCAT(__CONCAT(__CONCAT(Elf,__ELF_WORD_SIZE),_),x) x
-__sElfN(Sym);
-__sElfN(Shdr);
-#define	elf_getshdr		__elfN(getshdr)
 
 static void
 add_symbol(char *name, uintptr_t addr, size_t size)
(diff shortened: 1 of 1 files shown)


igalic (changes requested):
this (a46e046) is, unfortunately, running counter to our style(9) definition

usr.sbin/daemon/daemon.c
```diff
@@ -670,24 +682,21 @@ handle_chld(int signo __unused)
...
 }
 
 static void
 reopen_log(struct log_params *lpp)
 {
-	int outfd;
-
 	do_log_reopen = 0;
-	outfd = open_log(lpp->outfn);
+	int outfd = open_log(lpp->outfn);
```
kevans91, line 699: This one's a no-go, declarations must be at the top of the containing block.

usr.sbin/daemon/daemon.c
```diff
@@ -139,27 +141,43 @@ usage(int exitcode)
...
+        int dosyslog = 0;
+        int child_eof = 0;
+        int stdmask = STDOUT_FILENO | STDERR_FILENO;
+        int logpri = LOG_NOTICE;
+        int logfac = LOG_DAEMON;
+        int log_reopen = 0;
+        int pfd[2] = { -1, -1 };
+        int outfd = -1;
+
+        sigset_t mask_susp;
```
kevans91, line 165: Same comment as below about declarations

usr.sbin/daemon/daemon.c
```diff
@@ -139,27 +141,43 @@ usage(int exitcode)
...
+	sigset_t mask_orig;
+
+	sigset_t mask_read;
+	sigemptyset(&mask_read);
+
+	sigset_t mask_term;
+	sigemptyset(&mask_term);
+
+
+        struct log_params logpar = { 0 };
```
kevans91, line 186: Same comment as below about declarations

usr.sbin/daemon/daemon.c
```diff
@@ -292,22 +301,15 @@ main(int argc, char *argv[])
 	 * and syslog.
 	 */
 	if (pidfile || ppidfile || restart || outfd != -1 || dosyslog) {
-		struct sigaction act_term, act_chld, act_hup;
```
kevans91, line 271: I have mixed feelings about this one, given that they're really only used within this block

kevans91:
I don't think I understand the last one; daemon shouldn't be running forever in the non-forking mode, it just doesn't need to monitor the spawned process so it jumps right into `execvp`  (note that `pid` will not be reassigned from -1 if none of these options are set).

ngortheone:
@kevans91 disregard my last commit, I misinterpreted the code a bit. I will revert it soon.
But it does leave a big question open, which I will share in email

PR #668, opened 2023-02-25
GitHub: Silence clang-12 cross builds

jlduran:
The main idea is to stir a discussion, rather than present this pull request as the actual fix.

Can we just remove clang-12 from the test matrix?

cc/ @jrtc27, @emaste, @bsdjhb (educated guess of reviewers)

### DIFF
--- a/.github/workflows/cross-bootstrap-tools.yml
+++ b/.github/workflows/cross-bootstrap-tools.yml
@@ -2,7 +2,7 @@ name: Cross-build Kernel
 
 on:
   push:
-    branches: [ main, 'stable/13' ]
+    branches: [ main, 'stable/13', silence-github-clang-12 ]
   pull_request:
     branches: [ main ]
 
--- a/tools/build/cross-build/include/linux/unistd.h
+++ b/tools/build/cross-build/include/linux/unistd.h
@@ -41,17 +41,9 @@
 /* Ensure that unistd.h pulls in getopt */
 #define __USE_POSIX2
 #endif
-/*
- * Before version 2.25, glibc's unistd.h would define the POSIX subset of
- * getopt.h by defining __need_getopt,  including getopt.h (which would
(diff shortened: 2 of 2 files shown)


bsdimp:
I think we can just drop clang12 from the mix... unless the change to linux/unistd.h really fixes it, in which case we should do that...

lwhsu:
> I think we can just drop clang12 from the mix... unless the change to linux/unistd.h really fixes it, in which case we should do that...

#678 seems working. BTW, if we're going to drop clang 12 support, we need to update https://docs.freebsd.org/en/articles/committers-guide/#_current_compiler_versions

bsdimp:
> > I think we can just drop clang12 from the mix... unless the change to linux/unistd.h really fixes it, in which case we should do that...
> 
> #678 seems working. BTW, if we're going to drop clang 12 support, we need to update https://docs.freebsd.org/en/articles/committers-guide/#_current_compiler_versions

Yea, I think we should commit this fix, but keep clang12 since we need it still for stable/12 til its EOL. Might as well test it with a recent-ish ubuntu too.

jlduran:
You are right! it should be kept until EOY.

I just tested `ubuntu-latest` + `clang-15` and it works as expected. I'll submit a separate pull request for evaluation, but before that, how should the final cross-build matrix be?:

| OS version    | clang-12 | clang-13 | clang-14 | clang-15 |
|:--------------|:--------:|:--------:|:--------:|:--------:|
| macos-latest  | ✅       | ✅       | ✅       | ✅       |
| ubuntu-latest | ✅       | ✅       | ✅       | ✅       |

bsdimp:
Got the OK from @jrtc27 on IRC, with an appropriate commit message.

bsdimp (approved):
I'll steal much of the comment to do the push...

PR #670, opened 2023-02-26
if_re: Generate an address if there is none in the EEPROM

evgeni:
There exists hardware that has no ethernet address burned into the EEPROM. Loading if_re on such a HW brings the device up with '00:00:00:00:00:00' as the address, and that doesn't get you too far in a real network.

PR: [262406](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=262406)
Differential Revision: https://reviews.freebsd.org/D34485
Signed-off-by: Evgeni Golov <evgeni@debian.org>

### DIFF
--- a/sys/dev/re/if_re.c
+++ b/sys/dev/re/if_re.c
@@ -1674,6 +1674,11 @@ re_attach(device_t dev)
 		goto fail;
 	}
 
+	/* If address was not found, create one based on the hostid and name. */
+	if (ETHER_IS_ZERO(eaddr)) {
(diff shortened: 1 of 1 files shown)


igalic:
I would also reference the review, even tho it has been neglected
https://reviews.freebsd.org/D34485

that way, when this gets merged, it should be closed automatically, referencing the commit

evgeni:
Thanks! I have updated the commit message to refer to that and also update the BZ reference with the 'PR: ...' notation.

igalic:
excellent!
please also update the Pull Request description to match the commit message

PR #673, opened 2023-02-27
fix psm/atkbc build after c51978f4b2f0, unlock new settings from NOTES via userland

paepckehh:
* fixes the build of the legacy psm/atkbc driver after c51978f4b2f0 reported by Mark Millard / discord
* unlocks the use of the new NOTES featured kbd settings via 'kbdcontrol -r ultra' from userland as well

ref. [D38806](https://reviews.freebsd.org/D38806)


paepckehh:
needs a refactor, cancel here

PR #676, opened 2023-02-28
arm64/gicv3: Allow us to inherit the LPI mappings

bsdimp:
Due to bugs in the gicv3 hardware, you can't change the LPI mappings. Add code to read them in if we were booted with LPI enabled and use those instead of what we would allocate.

Be more pedanic about which bits we trust from region registers.

When forced to reuse an address, make sure its reserved.

Ensure what we do the gicv3 workaround only when it is safe to do so. Panic if we're forced to re-use an address that wasn't reserved early in boot. We have no good way if that address was reserved for this purpose and we won't have two users of that address.

When we're kexec'd from Linux, we have to reuse the memory that was assigned to the GICv3 ITS since the device can't be disabled enough to safely change these values without undefined behavior.

Sponsored by:		Netflix

### DIFF
--- a/sys/arm64/arm64/gicv3_its.c
+++ b/sys/arm64/arm64/gicv3_its.c
@@ -47,6 +47,7 @@ __FBSDID("$FreeBSD$");
 #include <sys/malloc.h>
 #include <sys/module.h>
 #include <sys/mutex.h>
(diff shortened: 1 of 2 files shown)


bsdimp:
Testing templates... didn't seem to work.

PR #677, opened 2023-03-01
ftgpio: Fix panic at attach and use better device description

stephane-rochoy-stormshield:
panic: ftgpio_group_get_ioreg: invalid register 0 for group 0

Successfully tested via `gioctl` to drive a LED via GPIO56 and GPIO57.


bsdimp:
Thanks! Fixed.

PR #679, opened 2023-03-02
xlocale: garbage collect references to strtoq_l/strtouq_l

valpackett:
These were explicitly never implemented (see lib/libc/locale/DESIGN.xlocale), but were referenced in the manpage and the symbol map.

Fixes:          3c87aa1d3dc ("Implement xlocale APIs from Darwin")
Reported by:    ld.lld 16 being --no-undefined-version by default
Reviewed by:    theraven
Sponsored by:   https://www.patreon.com/valpackett
Differential Revision: https://reviews.freebsd.org/D38408


emaste:
939b24b0abbd703948981c9bcb07afd7b01ca3b1

PR #682, opened 2023-03-04
acpica: do not print warning for missing _ADR

yuripv:
Started seeing the following after updating to VMware ESXi 8.0:

pcib2: <ACPI Host-PCI bridge> on acpi0
pcib2: could not evaluate _ADR - AE_NOT_FOUND
pci2: <ACPI PCI bus> on pcib2
vmx0: <VMware VMXNET3 Ethernet Adapter> ...

The virtual NIC works fine, and the code comment suggests that missing _ADR is not something fatal, skip printing the message if status is AE_NOT_FOUND.


bsdimp (approved):
I think this is good...

bsdimp:
109abf744bf7 landed this.

PR #683, opened 2023-03-05
consolidate kb interfaces (phase one)

paepckehh:
* remove 38 lines of redundant code
* consolidate & sync tables content
* add warning when fallback interface is used
* except bug fixes & warning, no change in behavior planned

Sponsored by: 		DSS GmbH
Pull Request:		https://github.com/freebsd/freebsd-src/pull/683
Differential Revision: 	https://reviews.freebsd.org/D38818

### DIFF
--- a/sys/sys/kbio.h
+++ b/sys/sys/kbio.h
@@ -70,6 +70,13 @@ struct keyboard_info {
 };
 typedef struct keyboard_info keyboard_info_t;
 
+/* keyboard repeat rate mapping table */
+static const int delays[]  = { 250, 500, 750, 1000 };
+static const int rates[] = { 34,  38,  42,  46,  50,
+	55,  59,  63, 68,  76,  84,  92, 100, 110, 118, 126,
+	136, 152, 168, 184, 200, 220, 236, 252, 272, 304, 336,
+	368, 400, 440, 472, 504 };
+
 /* add/remove keyboard to/from mux */
 #define KBADDKBD	_IOW('K', 68, keyboard_info_t)	/* add keyboard */
 #define KBRELKBD	_IOW('K', 69, keyboard_info_t)	/* release keyboard */
--- a/sys/sys/kbio.h
+++ b/sys/sys/kbio.h
@@ -71,8 +71,8 @@ struct keyboard_info {
 typedef struct keyboard_info keyboard_info_t;
 
 /* keyboard repeat rate mapping table */
-static const int delays[]  = { 250, 500, 750, 1000 };
-static const int rates[] = { 34,  38,  42,  46,  50,
+static const int kbdelays[]  = { 250, 500, 750, 1000 };
+static const int kbrates[] = { 34,  38,  42,  46,  50,
 	55,  59,  63, 68,  76,  84,  92, 100, 110, 118, 126,
 	136, 152, 168, 184, 200, 220, 236, 252, 272, 304, 336,
 	368, 400, 440, 472, 504 };
--- a/usr.sbin/kbdcontrol/kbdcontrol.c
+++ b/usr.sbin/kbdcontrol/kbdcontrol.c
@@ -1053,9 +1053,9 @@ set_keyrates(char *opt)
 	int r, d;
 
 	if (!strcmp(opt, "slow")) {
-		delay = 1000, repeat = 504 , d = 3, r = 31;
+		delay = 1000; repeat = 504; d = 3; r = 31;
 	} else if (!strcmp(opt, "normal")) {
-		delay = 500, repeat = 126, d = 1, r = 15;
+		delay = 500; repeat = 126; d = 1; r = 15;
 	} else if (!strcmp(opt, "fast")) {
 		delay = repeat = d = r = 0;
 	} else {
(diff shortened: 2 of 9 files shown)


usr.sbin/kbdcontrol/kbdcontrol.c
```diff
@@ -1058,14 +1053,11 @@ set_keyrates(char *opt)
 	int r, d;
 
 	if (!strcmp(opt, "slow")) {
-		delay = 1000, repeat = 500;
-		d = 3, r = 31;
+		delay = 1000, repeat = 504 , d = 3, r = 31;
```
bsdimp, line 1056: I know the old code uses , operators, but it would be better to avoid that and use normal ; here and below...

bsdimp:
generally, I like this. a minor issue or two... sorry for the delay...  not entirely sure I like the static in the .h file, but after sleeping on it a few days I have nothing better to suggest.

paepckehh:
CodeStyle updates done.

igalic (changes requested):
style

usr.sbin/kbdcontrol/kbdcontrol.c
```diff
@@ -1058,14 +1053,11 @@ set_keyrates(char *opt)
 	int r, d;
 
 	if (!strcmp(opt, "slow")) {
-		delay = 1000, repeat = 500;
-		d = 3, r = 31;
+		delay = 1000; repeat = 504; d = 3; r = 31;
 	} else if (!strcmp(opt, "normal")) {
-		delay = 500, repeat = 125;
-		d = 1, r = 15;
+		delay = 500; repeat = 126; d = 1; r = 15;
```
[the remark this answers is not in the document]
paepckehh: Yes, absolutely. I agree with you about modernizing and linting the code base. I was just trying to fit into the existing situation, to make it easier to read the code block. (updated)

paepckehh:
Code style updates for improved readability. (done)

usr.sbin/kbdcontrol/kbdcontrol.c
```diff
@@ -1085,14 +1086,15 @@ set_keyrates(char *opt)
...
-			if (repeat <= repeats[n])
+			if (repeat <= rates[n])
 				break;
 		r = n;
 	}
 
 	arg[0] = delay;
 	arg[1] = repeat;
 	if (ioctl(0, KDSETREPEAT, arg)) {
+		warn("fallback, setting keyboard rate via legacy interface (KDSETRAD), will be removed soon");
```
bsdimp, line 1097: How does one avoid this message? The message itself should have better instructions. The legacy interface won't be completely removed, just compiled out by default soon. I don't see the path to not get this message as well: if we're going to say it all the time, it's kinda pointless.

paepckehh: As far as I can see, even the darkest legacy ATKBC, ADB (lol) and all flavors of custom GPIO keyboards support the new interface. 

Therefore, the message is only for people with real niche custom driver (if any).  With the static mapping tables in the header and seven live code examples, it is not that complicated to migrate.

How often do you call KBDControl and adjust the repeat rate parameter manually during the day back and forth ( on a legacy non-supported custom keyboard driver device ) that necessitates the need to suppress this message? 

Being annoying is the whole point of that line. If this is a problem, I can move it into a (silent) comment.

sys/sys/kbio.h
```diff
@@ -70,6 +70,13 @@ struct keyboard_info {
 };
 typedef struct keyboard_info keyboard_info_t;
 
+/* keyboard repeat rate mapping table */
+static const int delays[]  = { 250, 500, 750, 1000 };
```
jrtc27, line 74: In other cases we have an opt-in macro guarding these so there's no pollution of the identifier space. Though even then these probably should be given less-generic names.

jrtc27:
Can you please not make 8 pushes in the space of 30 mins? Each one sends an email to the PR's subscribers.

sys/sys/kbio.h
paepckehh: Jessica, updated to avoid namespace pollution via rates and delays. Even though a certain 'pollution' via kbio.h was intended to prevent people from producing this mess again, I did not consider a more global perspective. Updated and solved.

paepckehh:
> Can you please not make 8 pushes in the space of 30 mins? Each one sends an email to the PR's subscribers.

Sorry for the fallout! It was not intentional. I am adjusting my tool chain now. Seem even the new beta GitHub interface is not better than the old one; Phabricator produced a correct consolidated diff/update.

paepckehh:
... anything missing?

bsdimp:
This is a bit of a mess... there's lots of commits with bad commit messages, there's some style issues (preexisting in many cases) and a couple of places that really should have nitems.
There's also other cleanups that can be done, which my tardiness in review no doubt delayed.
So I'm going to do those tweaks... and push once my build finishes.

bsdimp:
We'll work on those details next time...  sorry for the delay... I lost track of this one for some reason...

bsdimp:
landed as 761ab48d6af8
Sorry again for the delay.

PR #684, opened 2023-03-05
daemon: state management refactoring

ngortheone:
This PR is a refactoring of the state management in daemon.

Currently state management in daemon is sub-optimal for a number of reasons:
- most of the state is concentrated in `main()` - this triggers the following problems
- state initialization and main operation loop are mangled together
- management of the supervised process' desired state (i.e. RUN|TERMINATE) is implicit and transitions are managed by goto jumps to various places in main.

Daemon has a number of outstanding issues that have not been yet resolved. Adding new features or fixing existing bugs is problematic with due to the above outlined issues. I decided to perform code refactoring to resolve current code structure issues to make further improvements easier.

What was done:
- all state moved to `struct daemon_state` to simplify passing it around
- decouple initialization logic (`main()` from main operation loop (`deamon_mainloop()`)
- get rid of goto jumps and replace them with function calls

Daemon's open issues that I am planning to fix once code cleanup is finished:

-  [Bug #268580](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=268580)
-  [Bug #236117](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=236117)
-  [Bug #254511](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=254511)
-  [Bug #212829](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=212829)

P.S. All commits are atomic and each passes kyua tests

### DIFF
--- a/usr.sbin/daemon/daemon.c
+++ b/usr.sbin/daemon/daemon.c
@@ -80,6 +80,7 @@ struct daemon_state {
 	struct pidfh *child_pidfh;
 	struct log_params * logparams;
 	int keep_cur_workdir;
+	int restart_delay;
 	bool supervision_enabled;
 	bool child_eof;
 	bool restart_enabled;
@@ -163,7 +164,6 @@ main(int argc, char *argv[])
 {
 	char *p = NULL;
 	int ch = 0;
-	int restart_delay = 1;
 	int stdmask = STDOUT_FILENO | STDERR_FILENO;
 	struct log_params logparams = {
 		.syslog_enabled = false,
@@ -188,6 +188,7 @@ main(int argc, char *argv[])
 		.child_eof = false,
 		.restart_enabled = false,
 		.keep_cur_workdir = 1,
+		.restart_delay = 1,
 	};
 	sigset_t mask_orig;
 	sigset_t mask_read;
@@ -278,8 +279,8 @@ main(int argc, char *argv[])
 			break;
 		case 'R':
 			state.restart_enabled = true;
-			restart_delay = strtol(optarg, &p, 0);
-			if (p == optarg || restart_delay < 1) {
+			state.restart_delay = strtol(optarg, &p, 0);
+			if (p == optarg || state.restart_delay < 1) {
 				errx(6, "invalid restart delay");
 			}
 			break;
@@ -491,7 +492,7 @@ restart:
 
 	}
 	if (state.restart_enabled && !terminate) {
-		daemon_sleep(restart_delay, 0);
+		daemon_sleep(state.restart_delay, 0);
 	}
 	if (sigprocmask(SIG_BLOCK, &mask_term, NULL)) {
 		warn("sigprocmask");
--- a/usr.sbin/daemon/daemon.c
+++ b/usr.sbin/daemon/daemon.c
@@ -76,6 +76,7 @@ struct daemon_state {
 	struct pidfh *child_pidfh;
 	struct log_params * logparams;
 	bool supervision_enabled;
+	bool child_eof;
 };
 
 static void restrict_process(const char *);
@@ -154,7 +155,6 @@ usage(int exitcode)
 int
 main(int argc, char *argv[])
 {
-	bool child_eof = false;
 	bool restart_enabled = false;
 	char *p = NULL;
 	const char *child_pidfile = NULL;
@@ -181,6 +181,7 @@ main(int argc, char *argv[])
 		.child_pidfh = NULL,
 		.logparams = &logparams,
 		.supervision_enabled = false,
+		.child_eof = false,
 	};
 	sigset_t mask_orig;
 	sigset_t mask_read;
@@ -447,7 +448,7 @@ restart:
 		 * to depart. To handle the second option, a different
 		 * approach would be needed (procctl()?)
 		 */
-		if (child_gone && child_eof) {
+		if (child_gone && state.child_eof) {
 			break;
 		}
 
@@ -455,7 +456,7 @@ restart:
 			daemon_terminate(&state, 1);
 		}
 
-		if (child_eof) {
+		if (state.child_eof) {
 			if (sigprocmask(SIG_BLOCK, &mask_susp, NULL)) {
 				warn("sigprocmask");
 				daemon_terminate(&state, 1);
@@ -475,7 +476,7 @@ restart:
 			daemon_terminate(&state, 1);
 		}
 
-		child_eof = !listen_child(state.pipe_fd[0], &logparams);
+		state.child_eof = !listen_child(state.pipe_fd[0], &logparams);
 
 		if (sigprocmask(SIG_UNBLOCK, &mask_read, NULL)) {
 			warn("sigprocmask");
--- a/usr.sbin/daemon/daemon.c
+++ b/usr.sbin/daemon/daemon.c
@@ -74,6 +74,7 @@ struct daemon_state {
 	int pipe_fd[2];
 	const char *child_pidfile;
 	const char *parent_pidfile;
+	const char *title;
 	struct pidfh *parent_pidfh;
 	struct pidfh *child_pidfh;
 	struct log_params * logparams;
@@ -159,7 +160,6 @@ int
 main(int argc, char *argv[])
 {
 	char *p = NULL;
-	const char *title = NULL;
 	const char *user = NULL;
 	int ch = 0;
 	int keep_cur_workdir = 1;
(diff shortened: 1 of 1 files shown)


usr.sbin/daemon/daemon.c
```diff
@@ -501,23 +529,14 @@ main(int argc, char *argv[])
...
-exit:
-	close(logparams.output_fd);
-	close(pipe_fd[0]);
-	close(pipe_fd[1]);
-	if (logparams.syslog_enabled) {
-		closelog();
-	}
-	pidfile_remove(child_pidfh);
-	pidfile_remove(parent_pidfh);
-	exit(1); /* If daemon(3) succeeded exit status does not matter. */
```
igalic, line 520: leave this comment here

usr.sbin/daemon/daemon.c
```diff
@@ -742,3 +761,21 @@ reopen_log(struct log_params *logparams)
...
+	close(state->logparams->output_fd);
+	close(state->pipe_fd[0]);
+	close(state->pipe_fd[1]);
+	if (state->logparams->syslog_enabled) {
+		closelog();
+	}
+	pidfile_remove(state->child_pidfh);
+	pidfile_remove(state->parent_pidfh);
+
+	/* If daemon(3) succeeded exit status does not matter. */
```
igalic, line 779: and remove it here

usr.sbin/daemon/daemon.c
```diff
@@ -742,3 +761,21 @@ reopen_log(struct log_params *logparams)
 	logparams->output_fd = outfd;
 }
 
+static _Noreturn void
+daemon_terminate(struct daemon_state * state, int exitcode)
+{
+	if (state == NULL) {
+		exit(1);
```
igalic, line 768: why aren't you passing on the exit code here?

usr.sbin/daemon/daemon.c
```diff
@@ -534,14 +553,14 @@ daemon_sleep(time_t secs, long nsecs)
 
 static void
 open_pid_files(const char *pidfile, const char *ppidfile,
```
igalic, line 555: why are these two not part of the state?

igalic (commented):
first review pass

usr.sbin/daemon/daemon.c
ngortheone: this is abnormal behavior, state should never be null, so we exit with 1 if that happens

usr.sbin/daemon/daemon.c
ngortheone: they should be, I was trying to be minimalist with my change and moved only those variables that are needed to git rid of `goto exit`

ngortheone: another answer to this question is this - these are only necessary to until those files are opened and we have their descriptors, so storing names and descriptors feels a bit redundant..

ngortheone:
cc @kevans91

ngortheone:
cc @bsdimp

usr.sbin/daemon/daemon.c
```diff
@@ -742,3 +743,20 @@ reopen_log(struct log_params *logparams)
 	logparams->output_fd = outfd;
 }
 
+static _Noreturn void
+daemon_terminate(struct daemon_state * state, int exitcode)
+{
+	if (state == NULL) {
```
bsdimp, line 762: Can state be NULL?  If not, this should be an assert

ngortheone: it should never be, this is simply defensive programming, can easily be removed in the current state of the code state can't be NULL

usr.sbin/daemon/daemon.c
```diff
@@ -534,14 +535,14 @@ daemon_sleep(time_t secs, long nsecs)
...
 	int serrno;
 
 	if (pidfile) {
-		*pfh = pidfile_open(pidfile, 0600, &fpid);
-		if (*pfh == NULL) {
+		state->child_pidfh = pidfile_open(pidfile, 0600, &fpid);
+		if (state->child_pidfh == NULL) {
 			if (errno == EEXIST) {
 				errx(3, "process already running, pid: %d",
 				    fpid);
```
bsdimp, line 517: Why does this exit path not call the common exit path code?
A comment about why not would be good to have here.
Though it looks to be only to generate better error messages (which is good) and different return codes which are documented, so a comment here would clear that up... and a quick back reference below...

usr.sbin/daemon/daemon.c
```diff
@@ -551,10 +552,10 @@ open_pid_files(const char *pidfile, const char *ppidfile,
 	}
 	/* Do the same for the actual daemon process. */
 	if (ppidfile) {
-		*ppfh = pidfile_open(ppidfile, 0600, &fpid);
-		if (*ppfh == NULL) {
+		state->parent_pidfh = pidfile_open(ppidfile, 0600, &fpid);
+		if (state->parent_pidfh == NULL) {
 			serrno = errno;
-			pidfile_remove(*pfh);
+			pidfile_remove(state->child_pidfh);
```
bsdimp, line 527: Same here...

usr.sbin/daemon/daemon.c
```diff
@@ -341,46 +342,19 @@ main(int argc, char *argv[])
...
-		sigaddset(&act_term.sa_mask, SIGCHLD);
-
-		act_chld.sa_handler = handle_chld;
-		sigemptyset(&act_chld.sa_mask);
-		sigaddset(&act_chld.sa_mask, SIGTERM);
-
-		act_hup.sa_handler = handle_hup;
-		sigemptyset(&act_hup.sa_mask);
 
 		/* Block SIGTERM to avoid racing until we have forked. */
```
bsdimp: why not move this one too?

usr.sbin/daemon/daemon.c
```diff
@@ -693,12 +693,15 @@ setup_signals(struct daemon_state *state)
 		daemon_terminate(state, 1);
 	}
 
-	/* Setup SIGHUP */
```
bsdimp: This looks like it changes behavior. Why is it OK to just return here rather than just call daemon_terminate?

bsdimp (commented):
Got through the first five commits and stopped. This suggests that 20 is too many for this series to be reviewed, and maybe it would be better to have a couple of pull requests that are more like 5 each or so rather than 20. 20 is good for purely mechanical changes, but these changes move things around and are more complex 'non functional' changes which none-the-less require some thought for each one before it's clear it's good.

My suggestion is to do these 5 or so at a time so we can get the incremental stuff done more quickly than looking at the cumulative diff for all 20 and going 'woof, that's too much'.

I know this is a hassle for you, or could be, but let's see if that helps us get this cleared up more quickly, or if my hunch here isn't right. Thanks!

usr.sbin/daemon/daemon.c
ngortheone: The change at hand does not touch the logic of the function, I was simply renaming and relocating the variable. So my answer to the question "Why does this exit path not call the common exit path code?" is I don't know. But I will take this as a TODO to clean this up or clarify with comments

bsdimp: That makes sense. I'm happy to defer this to a TODO for future enhancement.

bsdimp:
Also, thinking about it, I could also try to land these commits N at a time if you rebase as I do that.

usr.sbin/daemon/daemon.c
bsdimp: An assert is the typical way to say 'this can't happen so tell me if it does' in FreeBSD.

usr.sbin/daemon/daemon.c
ngortheone: setting signal handlers happens once at the very beginning of the supervisor's startup.

the usage of `sigprocmask` is different, daemon does `SIG_UNBLOCK` as soon as fork succeeds later in this function. So this does not belong at `setup_signals()` I think keeping BLOCK/UNBLOCK pair in the same function is better for clarity.

See:  line 419
https://github.com/freebsd/freebsd-src/pull/684/files/f3f7fc66d65e178f8f53e0987d3bedca58604d59..2cbe6eb7ffab54ee544e213ede53a328ae3909e0#diff-e92ba17e18b6d814df32ae1bcf967fa671e44757395813d70a5305b74b556ff8R419

ngortheone:
@bsdimp a smaller portion of these changes moved here

https://github.com/freebsd/freebsd-src/pull/687

PR #685, opened 2023-03-06
libbe: Avoid double printing cloning errors.

jgrafton:
be_clone calls be_clone_cb and both call set_error on the return error path.  set_error prints the error resulting in a double print. be_clone_cb should just return the error code and allow be_clone to print it.

PR: 265248
Reported by: Graham Perrin

### DIFF
--- a/lib/libbe/be.c
+++ b/lib/libbe/be.c
@@ -745,11 +745,11 @@ be_clone_cb(zfs_handle_t *ds, void *data)
 
 	/* construct the boot environment path from the dataset we're cloning */
 	if (be_get_path(ldc, dspath, be_path, sizeof(be_path)) != BE_ERR_SUCCESS)
-		return (set_error(ldc->lbh, BE_ERR_UNKNOWN));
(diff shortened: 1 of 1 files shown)


bsdimp:
I was premature in my review...
There's 3 other set_error() calls in be_clone_cb(). I think they should all be fixed up.

jgrafton:
> I was premature in my review... There's 3 other set_error() calls in be_clone_cb(). I think they should all be fixed up.

Agreed.  I originally limited the change scope to the referenced PR but it's an obvious fix.

jlduran:
Doesn't this line also need to just return the `err`?:
https://github.com/freebsd/freebsd-src/blob/dbbd66466cb8f8538b7a918c8217fbba44a4035b/lib/libbe/be.c#L784

jgrafton:
> Doesn't this line also need to just return the `err`?:

Yeah, you're correct.  I can't think of any reason it should set the error in be_clone_cb at all.

bsdimp:
OK. looks good.

PR #686, opened 2023-03-11
Adding secure_getenv() into stdlib

seafork:
This commit adds the secure_getenv() function to stdlib.h. The rational behind this is the fact that many Linux applications use this function instead of getenv() as it's widely consider a, "best practice". As it follows, adding this to our stdlib would assist with the porting process. Also, having an easy function that promotes safety is a plus.

(Also, sorry for the blank pull request. Qutebrowser decided to "sabotage" me)
Signed-off-by: Lucy Marsh - seafork@disroot.org


bsdimp:
Seems reasonable.  It's implemented they way I'd do it.. :)

lib/libc/stdlib/getenv.c
```diff
@@ -448,6 +448,17 @@ getenv(const char *name)
 }
 
 
+/*
+ * Runs getenv() unless the current process is tainted by uid or gid changes, in which case it will return NULL.
+ */
+char *
+secure_getenv(const char *name)
+{
+    if (issetugid())
```
bsdimp, line 457: Any other conditions that should return null? The linux man page returns null for other conditions...

lib/libc/stdlib/getenv.3
```diff
@@ -79,6 +81,17 @@ to by the
 function.
 .Pp
 The
+.Fn secure_getenv
+function wraps the
+.Fn getenv
+function. As such, it will return
+.Fa NULL
+if the current process is tainted by uid or gid changes. Otherwise, it
```
bsdimp, line 89: Please start new sentences on a new line (due to a quirk in mandoc).
Please consider using igor (pkg install igor) and mandoc -Tlint to verify the changes.
You should have two warnigs about the useless style macro Tn, unrelated to this change.
Also set the .Dd at the top to be today's date.

mjguzik:
manpage should indicate this is for compat with glibc, and more importantly that 1. glibc performs linux-specific checks not done here 2. the current list of restrictions is subject to be extended down the road

lib/libc/stdlib/getenv.c
bsdimp: If you adopt @mjguzik 's man page suggestions, this becomes a non-issue.

seafork:
Added the requested changes to the man page.

bsdimp:
OK, we're getting close. There's three more things, plus a couple of comments.

First, we need to add a MLINKS line to the makefile so we install a link for secure_getenv.3 (see how we do that in the Makefile.inc in the stdlib directory. See how it's done for putenv, etc.

Second, we need to add this to the symbol versioning in stdlib/Symbol.map in FBSD_1.7 section.

Third, this is one commit worth of change, so please squish all these commits into one. I'd suggest a commit message like the following
```
Add GNU glibc compatible secure_getenv
    
Add most-glib/msl compatible secure_getenv. This only triggers for
issetugid() indicate the process is tainted.
```
for the combined series.

lib/libc/stdlib/getenv.3
```diff
@@ -32,7 +32,7 @@
 .\"     @(#)getenv.3	8.2 (Berkeley) 12/11/93
 .\" $FreeBSD$
 .\"
-.Dd November 7, 2021
+.Dd March 11, 2023
 .Dt GETENV 3
 .Os
 .Sh NAME
```
bsdimp, line 38: need to add secure_getenv alphabetically to this section.

seafork:
Hi, as a relative git newbie, I believe I have made one of the great big "git sins" with trying to rewrite history. I've used Github's sync feature (which uses rebase under the hood) and as a result I don't think there's any easy way for me to squash multiple commits without messing up something. I can cherry pick the changes into another branch; however, that would require another pull request. If you're fine with that, we could link back to this one for purpose of archiving and bypass all of this. If, however, there is something better or you would like me to do something different, I would be more than happy to so! Sorry for all the mess.

bsdimp:
> Hi, as a relative git newbie, I believe I have commited of the great big "git sins" with trying to rewrite history. I've used Github's sync feature (which I believe uses rebase under the hood) and as a result I don't think there's any easy way for me to squash multiple commits without messing up something. I can cherry pick the changes into another branch; however, that would require another pull request. If you're fine with that, we could link back to this one for purpose of archiving and bypass all of this. If, however, there is something better or you would like me to do something different, I would be more than happy to do so! Sorry for all the mess.

I can squash all the commits into one if that's hard to do...  Usually you'll need to do a forced push to make it work out, but if that's a problem (or you'd rather not do that), then I can squash things all down into one if everything else is in line. If it's one or two things, it isn't a big deal for me to do, but if it winds up being a lot then it takes too much time... So maybe focus on the other stuff and I'll work out that one detail...

bsdimp:
OK. I've squashed the changes, updated to today's date, fixed the Makefile.inc (when you add a link you need to add them in pair -- it's why the workflow failed the cirrus-ci test), fixed the comment on secure_getenv (to wrap it to < 80 columns) and have started a make buildworld successfully. I've used a mix of my own commit message and the one suggested by this pull request. I also approved the cross build run so future submissions will run automatically (and I know all of its jobs will fail quickly enough).

So I'll push the modified request once 'buildworld' finishes, unless I hit more issues....

seafork:
Awesome. Thank you so much for all of your help! I don't know why I didn't pick up on the pair part of the `Makefile.inc`, but I'm glad that you were kind enough to fix it for me. If there's anything more that needs to be done, I will more than gladly do so. Thank you again for all of your help with this PR!

bsdimp:
OK. I pushed this in. I had to correct the indenting... our style wants tabs, and it was 4 spaced. I didn't notice before.

Thanks for your submission. Hope it's the first of many and that you've learned a lot... This was a challenging one because of all the details that go along with adding something to libc.

jlduran:
GitHub cross-builds on macOS are failing after 68ca8363c7a19d5351dc2b10568cbf2403e07e33. Should we add something similar to [`tools/build/cross-build/include/mac/stdlib.h`](https://github.com/freebsd/freebsd-src/blob/78e6c3aacc7fb9181a0fd77dc0bdf4b29e2f1d3f/tools/build/cross-build/include/mac/stdlib.h)?

seafork:
Yeah, I just checked it out and it is most certainly is a problem. With that being said, I do believe that your suggestion would fix it. I'll push a PR for it in a minute or so!

jlduran:
There's no need, thank you! It was fixed in 0ec03c0b10b125f5996d59fee904248978981898.

seafork:
That's what I get for not checking my email enough... Thanks you for being patient!

PR #690, opened 2023-03-15
added proposal for 64bit bus buffer

elgithubo:
for review

### DIFF
--- a/tools/bus_space/bus.c
+++ b/tools/bus_space/bus.c
@@ -145,7 +145,7 @@ bs_read(int rid, off_t ofs, void *buf64, ssize_t bufsz)
 		return (0);
 	}
 	ofs += r->ofs;
-	if (r->ptr != MAP_FAILED) {
+	if (r->ptr64 != MAP_FAILED) {
 		ptr = r->ptr + ofs;
 		switch (bufsz) {
 		case 1:
--- a/tools/bus_space/bus.c
+++ b/tools/bus_space/bus.c
@@ -159,7 +159,6 @@ bs_read(int rid, off_t ofs, void *buf64, ssize_t bufsz)
 			break;
 		default:
 			*((uint64_t *)buf64)[] = *((volatile uint64_t *)ptr64)[];
-			errno = EIO;
 			return (0);
 		}
 	} else {
@@ -259,7 +258,6 @@ bs_write(int rid, off_t ofs, void *buf64, ssize_t bufsz)
 			break;
 		default:
 			*((volatile uint64_t *)ptr64)[] = *((uint64_t *)buf64)[];
-			errno = EIO;
 			return (0);
 		}
 	} else {
--- a/tools/bus_space/bus.c
+++ b/tools/bus_space/bus.c
@@ -132,6 +132,7 @@ int
 bs_read(int rid, off_t ofs, void *buf, ssize_t bufsz)
 {
 	struct resource *r;
+	volatile void *ptr64[];
 	volatile void *ptr;
 	off_t o;
 	ssize_t s;
@@ -158,6 +159,7 @@ bs_read(int rid, off_t ofs, void *buf, ssize_t bufsz)
 			break;
 		default:
 			errno = EIO;
+			*((uint64_t *)buf64)[] = *((volatile uint64_t *)ptr)[];
 			return (0);
 		}
 	} else {
@@ -227,10 +229,11 @@ bs_unmap(int rid)
 }
 
 int
-bs_write(int rid, off_t ofs, void *buf, ssize_t bufsz)
+bs_write(int rid, off_t ofs, void *buf64, ssize_t bufsz)
 {
 	struct resource *r;
 	volatile void *ptr;
+	volatile void *ptr64[];
 	off_t o;
 	ssize_t s;
 
@@ -242,7 +245,7 @@ bs_write(int rid, off_t ofs, void *buf, ssize_t bufsz)
 		return (0);
 	}
 	ofs += r->ofs;
-	if (r->ptr != MAP_FAILED) {
+	if (r->ptr64 != MAP_FAILED) {
 		ptr = r->ptr + ofs;
 		switch (bufsz) {
 		case 1:
@@ -255,6 +258,7 @@ bs_write(int rid, off_t ofs, void *buf, ssize_t bufsz)
 			*((volatile uint32_t *)ptr) = *((uint32_t *)buf);
 			break;
 		default:
+			*((volatile uint64_t *)ptr64)[] = *((uint64_t *)buf64)[];
 			errno = EIO;
 			return (0);
 		}
--- a/tools/bus_space/bus.c
+++ b/tools/bus_space/bus.c
@@ -157,8 +157,8 @@ bs_read(int rid, off_t ofs, void *buf, ssize_t bufsz)
 			*((uint32_t *)buf) = *((volatile uint32_t *)ptr);
 			break;
 		default:
-			errno = EIO;
 			*((uint64_t *)buf64)[] = *((volatile uint64_t *)ptr)[];
+			errno = EIO;
 			return (0);
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


bsdimp:
So what's the point? Tools/bus_space doesn't really do anything and I don't see where this is headed.

elgithubo:
bus_space did not support 64bit up to now and bus_space does something (I can look it up if you can't).
It builds on amd64 already.
Up to now it returned an error if more than 32bit typesize was used.

bsdimp:
OK. That makes sense. I can comment on what's there.

elgithubo:
Thank you and I would like to apologize for being rude on my first src pull request.

tools/bus_space/bus.c
```diff
@@ -227,10 +228,11 @@ bs_unmap(int rid)
 }
 
 int
-bs_write(int rid, off_t ofs, void *buf, ssize_t bufsz)
+bs_write(int rid, off_t ofs, void *buf64, ssize_t bufsz)
```
bsdimp, line 231: same comments all the way through for bs_write as bs_read.
To support 64 bit types you just need a case 8: in both of these functions.

bsdimp (commented):
Glad you are adding support for 64-bit operation to this userland tool.
The commit messages need to provide that context. They likely should be squashed down into one commit for the pull request, since they don't separate things out into different logical commits. Seems like what you want, though is something more like:
```
diff --git a/FreeBSD/tools/bus_space/bus.c b/FreeBSD/tools/bus_space/bus.c
index 1b0d2ca59c3d..8b47c32e1783 100644
--- a/FreeBSD/tools/bus_space/bus.c
+++ b/FreeBSD/tools/bus_space/bus.c
@@ -153,8 +153,8 @@ bs_read(int rid, off_t ofs, void *buf, ssize_t bufsz)
                case 2:
                        *((uint16_t *)buf) = *((volatile uint16_t *)ptr);
                        break;
-               case 4:
-                       *((uint32_t *)buf) = *((volatile uint32_t *)ptr);
+               case 8:
+                       *((uint64_t *)buf) = *((volatile uint64_t *)ptr);
                        break;
                default:
                        errno = EIO;
@@ -254,6 +254,9 @@ bs_write(int rid, off_t ofs, void *buf, ssize_t bufsz)
                case 4:
                        *((volatile uint32_t *)ptr) = *((uint32_t *)buf);
                        break;
+               case 8:
+                       *((volatile uint64_t *)ptr) = *((uint64_t *)buf);
+                       break;
                default:
                        errno = EIO;
                        return (0);
```

elgithubo:
Wow! Thank you for your great feedback.
I will resolve the feedback and then reply here again until no further objections are found.
Then I would create a PR in Bugzilla if required.
Is that okay?

bsdimp:
So we're in a bit of an experimental phase with github pull requests.
I'm trying to only do pull requests that meet the criteria in CONTRIBUTIONS.md.
This doesn't quite meet those (since it's not ready or nearly ready).
However, this looks simple enough, and there's someone who is providing you feedback (me) so that catches an exception in our current policy (so there's no need to go elsewhere yet).
So proceed here. Once you have all the feedback worked out, there's no need to go to bugzilla. I'll just land the change from here. If things stall out, we'll reevaluate.

igalic (commented):
👀

tools/bus_space/bus.c
```diff
@@ -255,6 +256,7 @@ bs_write(int rid, off_t ofs, void *buf, ssize_t bufsz)
 			*((volatile uint32_t *)ptr) = *((uint32_t *)buf);
 			break;
 		default:
+			*((volatile uint64_t *)ptr)[] = *((uint64_t *)buf)[];
```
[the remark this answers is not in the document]
elgithubo: I have performed an intense code cleaning. Could you please check once again?

elgithubo:
I would feel ready to test build on <= 32bit now, too.

tools/bus_space/bus.c
```diff
@@ -156,9 +156,12 @@ bs_read(int rid, off_t ofs, void *buf, ssize_t bufsz)
 		case 4:
 			*((uint32_t *)buf) = *((volatile uint32_t *)ptr);
 			break;
+		case 8:
+			*((uint64_t *)buf)[] = *((volatile uint64_t *)ptr)[];
```
bsdimp, line 160: remove the '[]'. It won't compile with them there.

tools/bus_space/bus.c
```diff
@@ -156,9 +156,12 @@ bs_read(int rid, off_t ofs, void *buf, ssize_t bufsz)
...
+			break;
 		default:
-			errno = EIO;
 			return (0);
+			errno = EIO;
```
bsdimp, line 164: errno needs to be set before the return.

tools/bus_space/bus.c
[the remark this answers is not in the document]
bsdimp: You need to remove the [] here as well.

tools/bus_space/bus.c
elgithubo: On amd64 it compiles. Instead of dimension N then you have dimension 1 when removing the []. Can you send me the testbuild logs, then I'll check it.

bsdimp: it surprises me that it builds.... It is semantically incorrect` even if it does. buf points to the 8 byte thing you want to interact with, not to an array of pointers to such things.

jlduran:
```
===> C (all)
cc  -O2 -pipe -fno-common -I/usr/src/tools/bus_space/C/..   -g -gz=zlib -MD  -MF.depend.bus.o -MTbus.o -std=gnu99 -Wno-format-zero-length -fstack-protector-strong -Wsystem-headers -Werror -Wall -Wno-format-y2k -W -Wno-unused-parameter -Wstrict-prototypes -Wmissing-prototypes -Wpointer-arith -Wreturn-type -Wcast-qual -Wwrite-strings -Wswitch -Wshadow -Wunused-parameter -Wcast-align -Wchar-subscripts -Wnested-externs -Wold-style-definition -Wno-pointer-sign -Wdate-time -Wmissing-variable-declarations -Wthread-safety -Wno-empty-body -Wno-string-plus-int -Wno-unused-const-variable -Wno-error=unused-but-set-variable  -Qunused-arguments    -c /usr/src/tools/bus_space/C/../bus.c -o bus.o
/usr/src/tools/bus_space/C/../bus.c:160:23: error: expected expression
                        *((uint64_t *)buf)[] = *((volatile uint64_t *)ptr)[];
                                           ^
/usr/src/tools/bus_space/C/../bus.c:160:55: error: expected expression
                        *((uint64_t *)buf)[] = *((volatile uint64_t *)ptr)[];
                                                                           ^
/usr/src/tools/bus_space/C/../bus.c:261:32: error: expected expression
                        *((volatile uint64_t *)ptr)[] = *((uint64_t *)buf)[];
                                                    ^
/usr/src/tools/bus_space/C/../bus.c:261:55: error: expected expression
                        *((volatile uint64_t *)ptr)[] = *((uint64_t *)buf)[];
                                                                           ^
4 errors generated.
*** Error code 1

Stop.
make[1]: stopped in /usr/src/tools/bus_space/C
*** Error code 1

Stop.
make: stopped in /usr/src/tools/bus_space
```

tools/bus_space/bus.c
elgithubo: swapped lines have been pushed.

tools/bus_space/bus.c
bsdimp: Yea, that's the error that I thought should happen.... The [] need to be removed for this pull request to move forward.

elgithubo:
Sorry to say, but I am indisponiert, since I cannot log into my server. After I have set it up, I have to finish a AI bootstrap case for my elaio which includes outbound ipv6 verification since I am unsure if it works without AWS ipv6 ::/0 (or 0/::!) security's definition setup and I even may not notice the verification run completion due to the setup of dovecot being delayed and sendmail not seeming reliable to me which results in a delay of probably weeks of the fixing of this commit's local directory compilation via make (and therefore also via ANSI C compiler since the c++ <something>-17 command line switch is there not available afaik).
Sorry!

bsdimp:
This has been in progress for months. When it's fixed, it can be reopened.

PR #692, opened 2023-03-16
two fixes for regex

github-cygwin:
While merging the latest FreeBSD regex into Cygwin, I stumbled over two
problems which I thought make sense upstream as well.

Hope that's ok,
Corinna

### DIFF
--- a/lib/libc/regex/regcomp.c
+++ b/lib/libc/regex/regcomp.c
@@ -329,7 +329,7 @@ regcomp_internal(regex_t * __restrict preg,
 		computejumps(p, g);
 		computematchjumps(p, g);
 		if(g->matchjump == NULL && g->charjump != NULL) {
-			free(g->charjump);
+			free(&g->charjump[CHAR_MIN]);
 			g->charjump = NULL;
 		}
 	}
--- a/lib/libc/regex/regcomp.c
+++ b/lib/libc/regex/regcomp.c
@@ -1138,7 +1138,7 @@ p_b_term(struct parse *p, cset *cs)
 #else
 			if (MB_CUR_MAX > 1) {
 #endif
-				(void)REQUIRE(start <= finish, REG_ERANGE);
+				(void)REQUIRE(p_range_cmp(start, finish) <= 0, REG_ERANGE);
 				CHaddrange(p, cs, start, finish);
 			} else {
 				(void)REQUIRE(p_range_cmp(start, finish) <= 0, REG_ERANGE);


lib/libc/regex/regcomp.c
```diff
@@ -1138,7 +1138,7 @@ p_b_term(struct parse *p, cset *cs)
 #else
 			if (MB_CUR_MAX > 1) {
 #endif
-				(void)REQUIRE(start <= finish, REG_ERANGE);
+				(void)REQUIRE(p_range_cmp(start, finish) <= 0, REG_ERANGE);
```
jrtc27, line 1141: This doesn't make sense, `start <= finish` is a precondition for CHaddrange to do anything, and it's defined to operate on the integer range not the localised character range

lib/libc/regex/regcomp.c
```diff
@@ -329,7 +329,7 @@ regcomp_internal(regex_t * __restrict preg,
 		computejumps(p, g);
 		computematchjumps(p, g);
 		if(g->matchjump == NULL && g->charjump != NULL) {
-			free(g->charjump);
+			free(&g->charjump[CHAR_MIN]);
```
jrtc27, line 332: This change matches the `&g->charjump[-(CHAR_MIN)]` when allocated (though stylistically I'd rather those were `g->charjump +/- CHAR_MIN` to be clear they're offsetting rather than trying to refer to a single element, even if semantically the two forms are defined to be equivalent).

lib/libc/regex/regcomp.c
github-cygwin: You're right, I didn't look into that closely enough, my apologies.

The test in CHaddrange is wrong, too.  The entire range processing should take the collation order into account, so a complete patch
should drop the check in CHaddrange and make sure that CHIN1 checks
the ranges against the collation order either.

I'll apply and test a patch in my Cygwin clone first and come back
later with a patch proposal.

jrtc27: Wouldn't it then become the same as the else branch? Anyway, I don't know this code well, I'm just noting that changing code in one branch to be the same as in the other branch is suspicious. If you think there's a bug here, perhaps you could provide some failing test cases? Regardless, maybe one for @kevans91 to look at as the most recent person to significantly touch regex(3).

github-cygwin: The else branch only handles single-byte codesets, and the difference is that in one case, CHaddrange allowes to define real start-finish ranges while in the other case the chars are added one by one to the bmp member, handling single-byte chars as a bit array.

Actually, from my POV, the entire special handling of single-byte codesets and the bmp member should go away.  Having two different ways of handling characters is an old optimization effort which doesn't make much sense anymore these days.  It only results in having more code basically doing the same thing.

Also, the bmp stuff is a problem as soon as the code starts to handle real unicode equivalences and collating sequences, which I intend to add to the code at least as part of the Cygwin code base.

For instance, "[[.ch.]-f]".  In the Czech locale, the character sequence "ch" is handled as a single character, and it's ordered between c and d.
So in Czech, the above range would match "ch", "d", "e", "f", but not "cg" or "ci".  With the bmp, this can't be adequately represented.

I already added support for unicode equivalence and collating sequences to the FreeBSD based fnmatch(3) and glob(3) in Cygwin.

regex(3) is the last candidate and the most complicated, IMHO.


Thanks,
Corinna

yuripv: The problem with ranges is that they really make sense only in C locale; once you start adding collation into the picture, you have to compromise; currently it's:

1. for multi-byte locales, just add the range as is; this works for [a-z] NOT matching A-Z (which was reported a lot, and still the case for e.g. fnmatch())
2. for single-byte locales, collation is used, but [a-z] will match anything in A-Z range (see e.g. https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=257972)

If you have a fix for this, I will be happy to look at it.

bsdimp:
So what's the status of this... The review seems mixed, but I may be misreading things.

github-cygwin:
> So what's the status of this... The review seems mixed, but I may be misreading things.

THe first patch still fixes freeing memory in an error condition.
As I noted in another reply, stylistically this is the same expression
as used in regfree.

The second patch is in limbo. Per
https://github.com/freebsd/freebsd-src/pull/692#discussion_r1139952897
I was planning to do this in the Cygwin code base first, but my time
to work on Cygwin was extremly limited. I have a bit of code locally,
but it's not yet finished.

bsdimp:
Reading the reviews, it looks like changes are needed. If the original author doesn't have time to make the changes soon, I'll close this review. It can be reopened when they have time.

bsdimp:
OK. I think we should land the first patch (any objections @jrtc27 ?) and then close this pull request until the second part is completed, when it could be a new pull request.

My apologies for letting this linger so long.

jrtc27:
> OK. I think we should land the first patch (any objections @jrtc27 ?) and then close this pull request until the second part is completed, when it could be a new pull request.
> 
> My apologies for letting this linger so long.
> 
> Note to self (or others): the Ready tag is for the one change.

First patch can have my Reviewed by:

bsdimp:
619f455b8fc9 landed. First commit only. Per agreement, closing PR. Second commit will be reworked and possibly resubmitted in the future.

PR #694, opened 2023-03-18
daemon: repace goto exit with daemon_terminate()

ngortheone:
Start breaking down big main()
Remove goto exit label and replace it with a function that does cleanup.

### DIFF
--- a/usr.sbin/daemon/daemon.c
+++ b/usr.sbin/daemon/daemon.c
@@ -730,7 +730,6 @@ reopen_log(struct daemon_state *state)
 static void
 daemon_state_init(struct daemon_state *state)
 {
-	memset(state, 0, sizeof(struct daemon_state));
 	*state = (struct daemon_state) {
 		.pipe_fd = { -1, -1 },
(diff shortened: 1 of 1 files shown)


ngortheone:
cc @kevans91 @bsdimp

usr.sbin/daemon/daemon.c
```diff
@@ -762,3 +755,19 @@ daemon_state_init(struct daemon_state *state)
 		.output_filename = NULL,
 	};
 }
+
+
+static _Noreturn void
+daemon_terminate(struct daemon_state *state)
+{
+	assert(state);
```
kevans91, line 763: Pointer should be explicitly compared against NULL when used like this

ngortheone: ack, will fix in a bit

ngortheone: fixed

ngortheone:
committed by @kevans91  
https://reviews.freebsd.org/rGcf6356fd471cdf4e5c52550f63599881d722962c

PR #696, opened 2023-03-20
local-unbound-setup.sh: Prevents updating resolv.conf by means of resolvconf.

petruchito:
How it was done before, it broke the `resolvconf -u` command. `Resolvconf` was trying and failing to create `/dev/null.bak`. 
Now it keeps `resolv.conf` constant and pointing to the local resolver, but user is still able to force `forwarders.conf` update.

### DIFF
--- a/usr.sbin/unbound/setup/local-unbound-setup.sh
+++ b/usr.sbin/unbound/setup/local-unbound-setup.sh
@@ -195,7 +195,9 @@ do_not_edit() {
 gen_resolvconf_conf() {
 	local style="$1"
 	do_not_edit
-	echo "resolv_conf=\"/dev/null\" # prevent updating ${resolv_conf}"
+  echo "name_servers_append=\"127.0.0.1\""
+  echo "resolv_conf_local_only=\"yes\""
+  echo "resolv_conf_options=\"edns0\""
 	if [ "${style}" = "dynamic" ] ; then
 		echo "unbound_conf=\"${forward_conf}\""
 		echo "unbound_pid=\"${pidfile}\""


usr.sbin/unbound/setup/local-unbound-setup.sh
```diff
@@ -195,7 +195,9 @@ do_not_edit() {
 gen_resolvconf_conf() {
 	local style="$1"
 	do_not_edit
-	echo "resolv_conf=\"/dev/null\" # prevent updating ${resolv_conf}"
+  echo "name_servers_append=\"127.0.0.1\""
+  echo "resolv_conf_local_only=\"yes\""
+  echo "resolv_conf_options=\"edns0\""
```
cschuber, line 200: 058a4e3419425 removed this because:

    Prevent resolvconf from updating /etc/resolv.conf.  As Jakob Schlyter
    pointed out, having additional nameservers listed in /etc/resolv.conf
    can break DNSSEC verification by providing a false positive if unbound
    returns SERVFAIL due to an invalid signature.  The downside is that
    the domain / search path won't get updated either, but we can live
    with that.
    
    Approved by:    re (blanket)

cschuber (changes requested):

usr.sbin/unbound/setup/local-unbound-setup.sh
petruchito: There won't be additional nameservers, `resolv_conf_local_only` makes resolvconf only add `127.0.0.1` and it's also smart enough not to add the localhost to `forward.conf`.

cschuber: > There won't be additional nameservers, `resolv_conf_local_only` makes resolvconf only add `127.0.0.1` and it's also smart enough not to add the localhost to `forward.conf`. Resulting `resolv.conf` is similar to what the script was generating except commented out old records.
> 
> ```
> petruchito@delta:~ % cat -n /etc/resolv.conf
>      1	# Generated by resolvconf
>      2	nameserver 127.0.0.1
>      3	options edns0
>      4	
> ```

In other words this reverts 058a4e3 because 058a4e3 removed the add of nameserver 127.0.0.1. Correct?

petruchito: Yes. 
I just thought I should also have added `echo "resolv_conf=\"${resolv_conf}\""` for non-standard `resolv.conf` locations.

cschuber:
The best path forward is to either email des@ directly or open a phabricator review and assign des@ as a reviewer (because I don't think he reviews P/R's). His reason for his patch is sound and I think he should review this patch. Does this sound fair?

petruchito:
Absolutely. Must it be done by a maintainer or I can open a phabricator review? (Gonna have to find out how then.)

cschuber:
> Absolutely. Must it be done by a maintainer or I can open a phabricator review? (Gonna have to find out how then.)

Thank you. Though I'm a committer and the person who has imported unbound from vendor into base for the last few years, I'm not comfortable touching this particular piece of code without DES reviewing or approving it. His reason for it is sound and he should review/approve it.

dag-erling:
This does not fix the problem, it just creates new ones. See [D39262](https://reviews.freebsd.org/D39262).

PR #698, opened 2023-03-21
fdescfs: Fix reference count leak

CTurt:
The reference used to be dropped in the `vn_vget_ino_gen` -> `fdesc_get_ino_alloc` path, however the `vn_vget_ino_gen` function has early return paths (depending on result of either `vfs_busy` or `VN_IS_DOOMED`), where the `fdesc_get_ino_alloc` callback, and thus the `fdrop` would not be reached.

### DIFF
--- a/sys/fs/fdescfs/fdesc_vnops.c
+++ b/sys/fs/fdescfs/fdesc_vnops.c
@@ -267,7 +267,7 @@ fdesc_get_ino_alloc(struct mount *mp, void *arg, int lkflags,
 	} else {
 		error = fdesc_allocvp(a->ftype, a->fd_fd, a->ix, mp, rvp);
 	}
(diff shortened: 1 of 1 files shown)


markjdb:
Thanks for the report. I think the patch isn't quite right: it's not safe to drop the reference while the directory vnode is locked. If we end up releasing the last reference, we'll lock the corresponding vnode, and some comments in fdesc_lookup() suggest that this can result in a lock order reversal.

I wrote a (much uglier) patch which addresses the leak while avoiding this problem: https://reviews.freebsd.org/D39189

emaste:
Issue fixed in https://github.com/freebsd/freebsd-src/commit/0f5b6f9a041e9cca3b376f6ec909374938887a3b

PR #700, opened 2023-03-22
arpa: garbage collect ns_newmsg/ns_rdata decls

valpackett:
These were brought in by the libbind import,
but these functions were never actually implemented anywhere, only header declarations and symbol map entries were imported.

Fixes:         046c3635cd, e45764721a
Reported by:   ld.lld 16 being --no-undefined-version by default
Sponsored by:  https://www.patreon.com/valpackett
Reviewed by:   @emaste
Differential Revision: https://reviews.freebsd.org/D38407

### DIFF
--- a/include/arpa/nameser.h
+++ b/include/arpa/nameser.h
@@ -133,17 +133,6 @@ typedef struct __ns_msg {
 	const u_char	*_msg_ptr;
 } ns_msg;
 
-/*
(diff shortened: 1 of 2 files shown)


emaste:
Merged in https://github.com/freebsd/freebsd-src/commit/f9c7fb7caed063757c7bb6a08cb5e81b2bbb5a6e

PR #701, opened 2023-03-23
daemon: use kqueue for all events

ngortheone:
Refactor daemon to use `kqueue/kevent` instead of signals.

This changes allows to simplify the code in several ways:
- the execution flow is now linear, no async events
- several variables became redundant and got removed (including all global variables)
- all event handling is now concentrated inside of the event loop, which makes code reading and comprehension easier.

### DIFF
--- a/usr.sbin/daemon/daemon.c
+++ b/usr.sbin/daemon/daemon.c
@@ -34,6 +34,7 @@
 __FBSDID("$FreeBSD$");
 
 #include <sys/param.h>
+#include <sys/event.h>
 #include <sys/mman.h>
 #include <sys/wait.h>
 
@@ -59,11 +60,14 @@ __FBSDID("$FreeBSD$");
 
 #define LBUF_SIZE 4096
 
+enum daemon_mode {
+	MODE_DAEMON = 0,   /* simply daemonize, no supervision */
+	MODE_SUPERVISE,    /* initial supervision state */
+	MODE_TERMINATING,  /* user requested termination */
+	MODE_NOCHILD,      /* child is terminated, final state of the event loop */
+};
+
 struct daemon_state {
-	sigset_t mask_orig;
-	sigset_t mask_read;
-	sigset_t mask_term;
-	sigset_t mask_susp;
 	int pipe_fd[2];
 	char **argv;
 	const char *child_pidfile;
@@ -74,6 +78,8 @@ struct daemon_state {
 	const char *user;
 	struct pidfh *parent_pidfh;
 	struct pidfh *child_pidfh;
+	enum daemon_mode mode;
+	int pid;
 	int keep_cur_workdir;
 	int restart_delay;
 	int stdmask;
@@ -81,33 +87,25 @@ struct daemon_state {
 	int syslog_facility;
 	int keep_fds_open;
 	int output_fd;
-	bool supervision_enabled;
-	bool child_eof;
 	bool restart_enabled;
 	bool syslog_enabled;
 	bool log_reopen;
 };
 
-static void setup_signals(struct daemon_state *);
 static void restrict_process(const char *);
-static void handle_term(int);
-static void handle_chld(int);
-static void handle_hup(int);
 static int  open_log(const char *);
 static void reopen_log(struct daemon_state *);
 static bool listen_child(int, struct daemon_state *);
 static int  get_log_mapping(const char *, const CODE *);
 static void open_pid_files(struct daemon_state *);
 static void do_output(const unsigned char *, size_t, struct daemon_state *);
-static void daemon_sleep(time_t, long);
+static void daemon_sleep(struct daemon_state *);
 static void daemon_state_init(struct daemon_state *);
 static void daemon_eventloop(struct daemon_state *);
 static void daemon_terminate(struct daemon_state *);
-
-static volatile sig_atomic_t terminate = 0;
-static volatile sig_atomic_t child_gone = 0;
-static volatile sig_atomic_t pid = 0;
-static volatile sig_atomic_t do_log_reopen = 0;
+static void daemon_exec(struct daemon_state *);
+static bool daemon_is_child_dead(struct daemon_state *);
+static void daemon_set_child_pipe(struct daemon_state *);
 
 static const char shortopts[] = "+cfHSp:P:ru:o:s:l:t:m:R:T:h";
 
@@ -172,6 +170,10 @@ main(int argc, char *argv[])
 
 	daemon_state_init(&state);
 
+	/* Signals are processed via kqueue */
+	signal(SIGHUP, SIG_IGN);
+	signal(SIGTERM, SIG_IGN);
+
 	/*
 	 * Supervision mode is enabled if one of the following options are used:
 	 * --child-pidfile -p
@@ -207,7 +209,7 @@ main(int argc, char *argv[])
 				errx(5, "unrecognized syslog facility");
 			}
 			state.syslog_enabled = true;
-			state.supervision_enabled = true;
+			state.mode = MODE_SUPERVISE;
 			break;
 		case 'm':
 			state.stdmask = strtol(optarg, &p, 10);
@@ -223,19 +225,19 @@ main(int argc, char *argv[])
 			 * daemon could open the specified file and set it's
 			 * descriptor as both stderr and stout before execve()
 			 */
-			state.supervision_enabled = true;
+			state.mode = MODE_SUPERVISE;
 			break;
 		case 'p':
 			state.child_pidfile = optarg;
-			state.supervision_enabled = true;
+			state.mode = MODE_SUPERVISE;
 			break;
 		case 'P':
 			state.parent_pidfile = optarg;
-			state.supervision_enabled = true;
+			state.mode = MODE_SUPERVISE;
 			break;
 		case 'r':
 			state.restart_enabled = true;
-			state.supervision_enabled = true;
+			state.mode = MODE_SUPERVISE;
 			break;
 		case 'R':
 			state.restart_enabled = true;
@@ -251,11 +253,11 @@ main(int argc, char *argv[])
 				errx(4, "unrecognized syslog priority");
 			}
 			state.syslog_enabled = true;
-			state.supervision_enabled = true;
+			state.mode = MODE_SUPERVISE;
 			break;
 		case 'S':
 			state.syslog_enabled = true;
-			state.supervision_enabled = true;
+			state.mode = MODE_SUPERVISE;
 			break;
 		case 't':
 			state.title = optarg;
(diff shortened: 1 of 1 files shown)


ngortheone:
cc @kevans91

usr.sbin/daemon/daemon.c
```diff
@@ -172,6 +164,10 @@ main(int argc, char *argv[])
 
 	daemon_state_init(&state);
 
+	/* Signals are processed via kqueue */
+	signal(1, SIG_IGN);
```
igalic, line 168: please use SIGHUP here instead of 1

usr.sbin/daemon/daemon.c
```diff
@@ -416,6 +405,10 @@ daemon_eventloop(struct daemon_state *state)
 				close(state->pipe_fd[1]);
 			}
 		}
+
+		/* Ignored signals remain ignored after execve, unignore them */
+		signal(1, SIG_DFL);
```
igalic, line 410: again, please use names instead of numbers here

usr.sbin/daemon/daemon.c
```diff
@@ -304,246 +307,218 @@ main(int argc, char *argv[])
...
+	close(state->pipe_fd[0]);
+	state->pipe_fd[0] = -1;
 }
 
 static void
-daemon_sleep(time_t secs, long nsecs)
+daemon_sleep(struct daemon_state *state)
 {
-	struct timespec ts = { secs, nsecs };
+	struct timespec ts = { state->restart_delay, 0 };
```
igalic, line 510: this is very clever.

will kqueue… queue events for us long enough?

i often set the restart delay to 30 seconds

usr.sbin/daemon/daemon.c
```diff
@@ -613,19 +588,18 @@ restrict_process(const char *user)
  * Return value of false is assumed to mean EOF or error, and true indicates to
```
igalic, line 588: is there anything special to do in case of error?
if so, you might create an Enum with SUCCESS, ERROR, EOF

igalic (commented):
I like this!

this feels like a big improvement

usr.sbin/daemon/daemon.c
```diff
@@ -304,246 +307,218 @@ main(int argc, char *argv[])
...
+
+	// TODO: add feature to avoid backgrounding
+	// i.e. --foreground, -f
+	//if (daemon(state.keep_cur_workdir, state.keep_fds_open) == -1) {
+	//	warn("daemon");
+	//	daemon_terminate(&state);
+	//}
+
+	if (state.mode == MODE_LEGACY) {
+		printf("executing legacy mode\n");
```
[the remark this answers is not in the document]
ngortheone: This is when no supervision happens. Daemon simply does execve() doing some minor preparations before:
- calls daemon(3) to go to background
- does something to stdin/stdout / redirect output to a file
- write pid file
- .. something else I might not remember atm. 

I called it `legacy` for 2 reasons: it existed in daemon before I touched it and apparently lots of ports use daemon in this mode, secondly - it is really not the best way to do supervision.

usr.sbin/daemon/daemon.c
```diff
@@ -304,246 +307,218 @@ main(int argc, char *argv[])
...
+		daemon_exec(&state);
 	}
+
 	/* Write out parent pidfile if needed. */
 	pidfile_write(state.parent_pidfh);
+	printf("written parent pidfile %s\n", state.parent_pidfile);
```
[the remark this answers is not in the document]
ngortheone: no, this is my makeshift debugging. all `printf()` will be removed in final version.
I needed this to really make sure it does the right thing. And this was better than stepping through in a debugger.

usr.sbin/daemon/daemon.c
ngortheone: I dunno why `nanosleep` was used. It was always using 0 nanoseconds.
In the future this can probably be replaced with sleep()...

usr.sbin/daemon/daemon.c
ngortheone: A nice improvement for the future. I was avoiding `listen_child` function for now, it still exists in its initial form (except for int->bool return type change)

usr.sbin/daemon/daemon.c
ngortheone: Or better with timers+kq as you point out

usr.sbin/daemon/daemon.c
```diff
@@ -304,246 +307,218 @@ main(int argc, char *argv[])
...
-			}
-			while (!terminate && !child_gone) {
-				sigsuspend(&state->mask_orig);
-			}
-			if (sigprocmask(SIG_UNBLOCK, &state->mask_susp, NULL)) {
-				warn("sigprocmask");
-				daemon_terminate(state);
-			}
+	while (state->mode != MODE_NOCHILD) {
+		printf("-- event loop tick.\n");
```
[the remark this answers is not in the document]
ngortheone: this is just debug stuff, will be removed.

igalic (approved):
to me this looks really good now.

i say we merge this, and then start hacking away at the TODOs in the code comments

usr.sbin/daemon/daemon.c
```diff
@@ -33,6 +33,7 @@
 #include <sys/cdefs.h>
 __FBSDID("$FreeBSD$");
 
+#include <sys/event.h>
```
kevans91, line 36: Per style(9), `sys/param.h` or `sys/types.h` always come first, followed by the rest in alphabetical order.

usr.sbin/daemon/daemon.c
```diff
@@ -304,246 +306,208 @@ main(int argc, char *argv[])
...
-			err(1, "pipe");
-		}
-		/*
-		 * Spawn a child to exec the command.
-		 */
-		child_gone = 0;
-		pid = fork();
+        int kq;
+	int ret;
+	struct kevent event;
```
kevans91, line 368: This should go above `int` vars

usr.sbin/daemon/daemon.c
```diff
@@ -304,246 +306,208 @@ main(int argc, char *argv[])
...
 
-		if (sigprocmask(SIG_BLOCK, &state->mask_read, NULL)) {
-			warn("sigprocmask");
-			daemon_terminate(state);
+		if (event.flags & EV_ERROR) {
+			errx(EXIT_FAILURE, "Event error: %s", strerror(event.data));
 		}
 
-		state->child_eof = !listen_child(state->pipe_fd[0], state);
+		switch(event.filter) {
```
kevans91, line 438:
```suggestion
		switch (event.filter) {
```

usr.sbin/daemon/daemon.c
```diff
@@ -304,246 +306,208 @@ main(int argc, char *argv[])
...
-			warn("sigprocmask");
-			daemon_terminate(state);
-		}
+			switch (event.ident) {
+			case SIGCHLD:
+				if (daemon_is_child_dead(state)) {
+					/* child is dead, read all until EOF */
+					state->pid = -1;
+					state->mode = MODE_NOCHILD;
+					while(listen_child(state->pipe_fd[0], state));
```
kevans91, line 447:
```suggestion
					while (listen_child(state->pipe_fd[0], state));
```

usr.sbin/daemon/daemon.c
```diff
@@ -613,6 +577,7 @@ restrict_process(const char *user)
  * Return value of false is assumed to mean EOF or error, and true indicates to
  * continue reading.
  */
+// TODO: simplify signature - state contains pipefd
```
kevans91, line 580: Wrong comment style, if it's going to stay. Ideally just merge it into the block just above with a blank line separating.

usr.sbin/daemon/daemon.c
```diff
@@ -780,25 +705,23 @@ daemon_state_init(struct daemon_state *state)
...
 
 static _Noreturn void
 daemon_terminate(struct daemon_state *state)
 {
 	assert(state != NULL);
-	close(state->output_fd);
-	close(state->pipe_fd[0]);
-	close(state->pipe_fd[1]);
+
+	if (state->output_fd >= 0 ) {
```
kevans91, line 715:
```suggestion
	if (state->output_fd >= 0) {
```

usr.sbin/daemon/daemon.c
```diff
@@ -812,3 +735,52 @@ daemon_terminate(struct daemon_state *state)
 	 */
 	exit(1);
 }
+
+/*
+ * Returns true if SIGCHILD came from state->pid
+ * TODO: this function is potentially buggy.
```
kevans91, line 741: I would probably omit this particular commentary, just a note that it could hang here if it's signaled for a reason other than child dying (e.g., ptrace attach).

usr.sbin/daemon/daemon.c
ngortheone: fixed

usr.sbin/daemon/daemon.c
ngortheone: fixed

usr.sbin/daemon/daemon.c
ngortheone: fixed

usr.sbin/daemon/daemon.c
ngortheone: fixed

usr.sbin/daemon/daemon.c
ngortheone: fixed

usr.sbin/daemon/daemon.c
ngortheone: fixed

usr.sbin/daemon/daemon.c
ngortheone: fixed

usr.sbin/daemon/daemon.c
```diff
@@ -304,246 +306,208 @@ main(int argc, char *argv[])
...
+
+	if (pipe(state->pipe_fd)) {
+		err(1, "pipe");
 	}
 
-	/* fork failed, this can only happen when supervision is enabled */
-	if (pid == -1) {
-		warn("fork");
-		daemon_terminate(state);
+	kq = kqueue();
```
kevans91, line 381: I think we'll always fork + close it at this point, but you might consider using the very recently added `kqueuex` w/ `KQUEUE_CLOEXEC` here

ngortheone: done

ngortheone:
https://reviews.freebsd.org/rG8935a3993219be76c7ea03e9ad4509657d08af6c
commited by @kevans91

PR #705, opened 2023-04-01
nanobsd: pcengines: Use kldstat -m instead of -n

jlduran:
`kldstat(8)` displays the status of any files dynamically linked into the kernel.

Per `kldstat(8)`, `-n` will display the status of a filename, `-m` will display the status of a modname.

We are checking by modnames, not filenames.  Use `-m` instead.

### DIFF
--- a/tools/tools/nanobsd/pcengines/test.sh
+++ b/tools/tools/nanobsd/pcengines/test.sh
@@ -29,10 +29,10 @@ _run () {
 }
 
 _init () {
-  kldstat -n kqemu || kldload kqemu
-  kldstat -n aio || kldload aio
-  kldstat -n if_tap || kldload if_tap
-  kldstat -n if_bridge || kldload if_bridge
+  kldstat -m kqemu || kldload kqemu
+  kldstat -m aio || kldload aio
+  kldstat -m if_tap || kldload if_tap
+  kldstat -m if_bridge || kldload if_bridge
   sysctl net.link.tap.up_on_open=1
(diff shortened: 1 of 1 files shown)


tools/tools/nanobsd/pcengines/test.sh
```diff
@@ -29,10 +29,10 @@ _run () {
 }
 
 _init () {
-  kldstat -n kqemu || kldload kqemu
-  kldstat -n aio || kldload aio
-  kldstat -n if_tap || kldload if_tap
-  kldstat -n if_bridge || kldload if_bridge
+  kldstat -m kqemu || kldload kqemu
```
gmshake, line 32: Can you please try 'kldload -nq kqemu' ?

jlduran:
I must had an old borked setup.
For the obsolete kqemu to exist, you need to use the emulators/qemu pkg, however commit https://github.com/freebsd/freebsd-ports/commit/cb11558b75bc84cf6c824c94aad9b343fea1224b bumped the version 0.11.1, which was the last one that included kqemu (as noted in the qemu-devel pkg-message) or the emulators/kqemu-kmod pkg, removed in https://github.com/freebsd/freebsd-ports/commit/5cc0ee357bc56bdb3ee306cc8ba20ec745ee4d92.
I have recreated it, and both `-m` and `-n` work. The "quiet" `-q` option is nice to have, but I feel an obsolete script like this  should not be touched.
Thank you for reviewing, I will however close this pull request.

PR #708, opened 2023-04-04
nvme: add locking around nvme consumer registration

IanMarcoMoffett:
Tested on a thinkpad T470

### DIFF
--- a/sys/dev/nvme/nvme.c
+++ b/sys/dev/nvme/nvme.c
@@ -33,6 +33,7 @@ __FBSDID("$FreeBSD$");
 #include <sys/bus.h>
 #include <sys/conf.h>
 #include <sys/module.h>
+#include <sys/mutex.h>
 
 #include <vm/uma.h>
 
@@ -46,6 +47,7 @@ struct nvme_consumer {
 	nvme_cons_fail_fn_t	fail_fn;
 };
 
+static struct mtx nvme_consumer_mutex;
 struct nvme_consumer nvme_consumer[NVME_MAX_CONSUMERS];
 #define	INVALID_CONSUMER_ID	0xFFFF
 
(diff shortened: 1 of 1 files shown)


bsdimp:
Why is this needed?

IanMarcoMoffett:
> Why is this needed?

Well I thought I'd do what the TODO says.

It's also probably a good idea to do that to reduce the risk of a race condition causing undefined behavior.

IanMarcoMoffett:
> Why is this needed?

Additionally, I saw no locks in place and the code that I added ensures that modification to that data structure is synchronized which is needed because it's a shared structure.

sys/dev/nvme/nvme.c
```diff
@@ -310,18 +314,18 @@ nvme_register_consumer(nvme_cons_ns_fn_t ns_fn, nvme_cons_ctrlr_fn_t ctrlr_fn,
 {
 	int i;
 
-	/*
-	 * TODO: add locking around consumer registration.
-	 */
 	for (i = 0; i < NVME_MAX_CONSUMERS; i++)
 		if (nvme_consumer[i].id == INVALID_CONSUMER_ID) {
+			mtx_lock(&nvme_consumer_mutex);
```
gmshake, line 319: This is wrong.

Assume two threads concurrently register consumer, they may still ref to the same nvme_consumer[i] , then this lock does not serve any purpose.

sys/dev/nvme/nvme.c
```diff
@@ -310,18 +314,18 @@ nvme_register_consumer(nvme_cons_ns_fn_t ns_fn, nvme_cons_ctrlr_fn_t ctrlr_fn,
 {
 	int i;
 
-	/*
-	 * TODO: add locking around consumer registration.
```
gmshake, line 314: I think the comment is meant to prevent concurrent registration.

gmshake (changes requested):

PR #709, opened 2023-04-04
copy_file_range.2: Document EXDEV error

jlduran:
Two commits:
- _De rigueur_ mandoc lint pass
- Document `EXDEV` error code in `copy_file_range(2)` (with date bump)


mjguzik:
the error is *not* correct and is only being temporarily returned as a bug, this will be fixed soon(tm)

that said, sorry but closing this one :)

PR #711, opened 2023-04-06
ifconfig tests: Add test for create if_bridge without unit number

jlduran:
https://reviews.freebsd.org/D39396


jlduran:
Sorry, wrong base.

PR #712, opened 2023-04-12
Update UPDATING in main

grahamperrin:
Uppercase H for the _FreeBSD Handbook_.

Change lowercase current to uppercase CURRENT. 

Remove the white space that precedes the link to the relevant section of the Handbook. This is subtle; it improves usability for people who habitually triple-click to select a chunk of text (compared to a double-click to select a word).

### DIFF
--- a/UPDATING
+++ b/UPDATING
@@ -1,4 +1,4 @@
-Updating Information for users of FreeBSD 14.0-CURRENT.
+Updating Information for users of FreeBSD-CURRENT.
 
 This file is maintained and copyrighted by M. Warner Losh <imp@freebsd.org>.
 See end of file for further details.  For commonly done items, please see the
--- a/UPDATING
+++ b/UPDATING
@@ -4,7 +4,7 @@ This file is maintained and copyrighted by M. Warner Losh <imp@freebsd.org>.
 See end of file for further details.  For commonly done items, please see the
 COMMON ITEMS: section later in the file.  These instructions assume that you
 basically know what you are doing.  If not, then please consult the FreeBSD
-Handbook:
+handbook:
 
     https://docs.freebsd.org/en/books/handbook/cutting-edge/#makeworld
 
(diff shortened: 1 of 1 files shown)


emaste:
IMO The whitespace before the URL is intentional, to make it stand out.

UPDATING
```diff
@@ -1,12 +1,12 @@
-Updating Information for FreeBSD current users.
+Updating Information for users of FreeBSD 14.0-CURRENT.
```
bsdimp, line 1: We never change this on main. NAK.

UPDATING
```diff
@@ -1,12 +1,12 @@
...
 
 This file is maintained and copyrighted by M. Warner Losh <imp@freebsd.org>.
 See end of file for further details.  For commonly done items, please see the
 COMMON ITEMS: section later in the file.  These instructions assume that you
 basically know what you are doing.  If not, then please consult the FreeBSD
-handbook:
+Handbook:
 
-    https://docs.freebsd.org/en/books/handbook/cutting-edge/#makeworld
+https://docs.freebsd.org/en/books/handbook/cutting-edge/#makeworld
```
bsdimp, line 9: Don't delete this leading whitespace.

bsdimp:
Only the capital h looks good to me.

grahamperrin:
> IMO The whitespace before the URL is intentional, to make it stand out.

I think it's fairly outstanding as a separate paragraph for the address alone, but no argument. 

image

bsdimp:
NAK means "I'm explicitly saying no." so no to adding the version of FreeBSD to the -current UPDATING page...

The extra spacing would be fine if there's blank lines...  the URL I think is now longer than it was when it was added (or I'm just confused).

grahamperrin:
> … the URL I think is now longer than it was when it was added …

I don't know the back history, but URLs that included `en_US.ISO8859-1` were longer. See for example: 

https://github.com/freebsd/freebsd-src/pull/713/files#diff-7a99811ebf6b6268a2581fcf050df0a2c325fdbf1c02481dad16afd38f4bd6a1R9

PR #716, opened 2023-04-13
ping: pr_iph() improvements

jlduran:
Very early on, the `Src`/`Dst` IP addresses were printed in hex notation (`%08x`), which will always be 8-characters wide.  It was later changed to use a dot-decimal notation.  Depending on the IP address length, the `Src` and `Dst` headers may require a different padding.  Use the source and destination IP lengths as padding for the headers.

Also, print an `Opts` (options) header, if there are options present.  It has been abbreviated to `Opts` to match the length of the previous `Data` header, removed in ef9e6dc7eebe9830511602904d3ef5218d964080.

Print the header info such that no trailing spaces are produced.  As some git workflows may automatically trim them, and make the tests fail (see 25b86f8559c2e7076daff56933217e95cd4398d4).

Before
```
Vr HL TOS  Len   ID Flg  off TTL Pro  cks      Src      Dst
 4  f  00 007c 0001   0 0000  40  01 d868 192.0.2.1  192.0.2.2␣
```

After
```
Vr HL TOS  Len   ID Flg  off TTL Pro  cks       Src       Dst
 4  f  00 007c 0001   0 0000  40  01 d868 192.0.2.1 192.0.2.2
```

And with options:

Before
```
Vr HL TOS  Len   ID Flg  off TTL Pro  cks      Src      Dst
 4  f  00 007c 0001   0 0000  40  01 d868 192.0.2.1  192.0.2.2 01...
```

After
```
Vr HL TOS  Len   ID Flg  off TTL Pro  cks       Src       Dst Opts
 4  f  00 007c 0001   0 0000  40  01 d868 192.0.2.1 192.0.2.2 01...
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,54 +1,24 @@
 # $FreeBSD$
 
-compute_engine_instance:
-  # Image list available via
-  # gcloud compute images list --project freebsd-org-cloud-dev --no-standard-images
-  platform: freebsd
-  image_project: freebsd-org-cloud-dev
-  image: freebsd-13-1-release-amd64
-  cpu: 8
-  memory: 8G
-  disk: 40
+freebsd_instance:
+   image_family: freebsd-14-0-snap
 
 task:
-  matrix:
-  - name: World and kernel amd64 build and boot smoke test
-    env:
-      TARGET: amd64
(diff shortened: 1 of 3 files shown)


jlduran:
Sorry, wrong base

PR #717, opened 2023-04-15
Fix saving STRUCTURED-DATA in RFC 5424 log format

nuclight:
When saving log files in RFC 5424 format presence of STRUCTURED-DATA in message caused line to not appear in log file at all, because a variable forgotten to initialize. This could be tested by (slightly modified) example from RFC:

    echo '<165>1 2003-10-11T22:14:15.003Z mymachine.example.com su 12345 ID47 [exampleSDID@32473 iut="3" eventSource="Application" eventID="1011"][id@2 test="tast"] BOM"su root" failed for lonvick on /dev/pts/8" ' | nc -w1 -Uu /var/run/log

While here, update stale information in comment to logmsg() that RFC 5424 log format was not supported.

### DIFF
--- a/usr.sbin/syslogd/syslogd.c
+++ b/usr.sbin/syslogd/syslogd.c
@@ -1123,6 +1123,7 @@ parsemsg_rfc5424(const char *from, int pri, char *msg)
 	FAIL_IF("STRUCTURED-NAME", start == msg);			\
 } while (0)
 	IF_NOT_NILVALUE(structured_data) {
+		structured_data = msg;
 		/* SD-ELEMENT. */
 		while (*msg == '[') {
 			++msg;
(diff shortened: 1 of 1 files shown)


igalic:
>  This could be tested by (slightly modified) example from RFC:

*\*looks at the gaping hole where tests should be\**

maybe take a look at https://github.com/freebsd/freebsd-src/blob/HEAD/usr.sbin/daemon/tests/daemon_test.sh as an example for how to add tests, if you think it's worth your time

nuclight:
Well, for tests the real implementation of structured data must be done, e.g. in property filters: currently `syslogd` just checks this field for syntax validity and saves it as raw string - no real logic is present, so nothing to test.

bsdimp:
Looks good to my eye as well.
pushed as a81b559dad5b
Thanks for your contribution. Sorry it took so long

PR #720, opened 2023-04-19
sysctl: Fix vfs.cache.stats.* name typos

ihoro:
Two vfs.cache.stats names are fixed:
- s/.dotdothis/.dotdothits/
- s/.posszaps/.poszaps/


mjguzik:
thanks, landed in https://cgit.freebsd.org/src/commit/?id=0e0c47ecd624da976bf7e1ff391cb4a00c76b200

PR #724, opened 2023-04-21
cat: ensure serrno is always set

AtariDreams:
Set serrno to errno before the loop begins for the edge case that res0 is immediately NULL.

### DIFF
--- a/bin/cat/cat.c
+++ b/bin/cat/cat.c
@@ -436,7 +436,6 @@ udom_open(const char *path, int flags)
 	 */
 	bzero(&hints, sizeof(hints));
 	hints.ai_family = AF_LOCAL;
-	fd = -1;
 
 	if (fileargs_realpath(fa, path, rpath) == NULL)
 		return (-1);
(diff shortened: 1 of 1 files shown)


ngie-eign:

AtariDreams:
> Please change the email address to a non-users.noreply.github.com based address.

Done!

bin/cat/cat.c
```diff
@@ -447,11 +446,14 @@ udom_open(const char *path, int flags)
 		errno = EINVAL;
 		return (-1);
 	}
-	cap_rights_init(&rights, CAP_CONNECT, CAP_READ, CAP_WRITE,
-	    CAP_SHUTDOWN, CAP_FSTAT, CAP_FCNTL);
+	cap_rights_init(&rights, CAP_CONNECT, CAP_READ, CAP_WRITE, CAP_SHUTDOWN,
+	    CAP_FSTAT, CAP_FCNTL);
```
bsdimp, line 450: This change is cosmetic. Please submit is separately.

bin/cat/cat.c
```diff
@@ -447,11 +446,14 @@ udom_open(const char *path, int flags)
...
+
+	/* Default error if something goes wrong. */
+	serrno = EINVAL;
+
 	for (res = res0; res != NULL; res = res->ai_next) {
-		fd = socket(res->ai_family, res->ai_socktype,
-		    res->ai_protocol);
+		fd = socket(res->ai_family, res->ai_socktype, res->ai_protocol);
```
bsdimp, line 456: ditto

bsdimp:
OK. Please drop the style changes I've highlighted. the other style changes should be mentioned in the commit message. Like "Simplify structure of code because fd < 0 test is not needed if we return early." or some such.

AtariDreams:
@bsdimp Done!

bsdimp:
landed as 38f57faa06cf

PR #725, opened 2023-04-21
cross-build: Add common compatibility headers after OS-specific ones

arichardson:
The current order is rather counter-intuitive and it appears to work fine if we swap the order.


bsdimp:
Looks good to me, and I think that's the right order to do things since it allows #include_next in the OS specific files to do stuff AFTER the #include_next, which can sometimes be important.

PR #729, opened 2023-05-01
Manual page improvements to mpr(4), mps(4) and mpsutil(8)

tomhukins:


bsdimp:
these look good... maybe a bit too small per commit, but it's a bit of a hassle to fix that and it's not worth a round trip...af32969ddbb5

tomhukins:
> maybe a bit too small per commit

Thanks for the feedback, @bsdimp.  I remember an earlier pull request of mine receiving the suggestion that I separate changes into smaller commits, but I guess I overdid it this time.

bsdimp:
yea, it's a balance...  it wasn't that bad, but next time a small bit bigger wouldn't be bad...

PR #731, opened 2023-05-04
ntpd: Support IPv6-only hosts

igalic:
0.pool.* returns only IPv4 addresses.
2.pool.* returns both, IPv6 and IPv4 addresses.

conservatively extend our IPv4 only pool configuration by adding a second pool, which also returns IPv6 addresses.

PR: 270536
Reported by:  Lapo Luchini <lapo@lapo.it>
Differential Revision: https://reviews.freebsd.org/D39954
Pull Request: https://github.com/freebsd/freebsd-src/pull/731

### DIFF
--- a/usr.sbin/ntp/ntpd/ntp.conf
+++ b/usr.sbin/ntp/ntpd/ntp.conf
@@ -20,24 +20,27 @@
 tos minclock 3 maxclock 6
 
 #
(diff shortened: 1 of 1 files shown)


tomhukins (approved):

igalic:
merged in fff1551c0357cc43f6138fecfad1d79d8d8baa15

PR #733, opened 2023-05-06
newfs: prefer unsigned index over signed

AtariDreams:
We can just use a for loop starting at 0 instead of a while loop starting at -1.


igalic (approved):
👍

ngie-eign:

AtariDreams:

Done!

bsdimp:
I think it's good. I've asked kirk to take a look.

bsdimp:
Pushed as 430168942f2b

bsdimp:
> +1

Doh! I missed this review when I pushed. I'm sorry.

PR #734, opened 2023-05-06
newfs: nextnum should be a u_int32_t, not an int

AtariDreams:
The function that uses nextnum expects to return a u_int32_t, not a mere int, so let's make nextnum a u_int32_t instead.

### DIFF
--- a/sbin/newfs/mkfs.c
+++ b/sbin/newfs/mkfs.c
@@ -1230,7 +1230,7 @@ ilog2(int val)
 static u_int32_t
 newfs_random(void)
 {
-	static int nextnum = 1;
+	static u_int32_t nextnum = 1;
 
 	if (Rflag)
 		return (nextnum++);


ngie-eign:

AtariDreams:

Done!

sbin/newfs/mkfs.c
```diff
@@ -1230,7 +1230,7 @@ ilog2(int val)
 static u_int32_t
 newfs_random(void)
 {
-	static int nextnum = 1;
+	static u_int32_t nextnum = 1;
```
bsdimp, line 1233: This should be uint32_t. u_int32_t is the obsolete form.

bsdimp: ah, the rest of the file uses u_int32_t, so I'll leave that be.

bsdimp:
27cebb4eac3d

PR #735, opened 2023-05-07
Mark usage function as __dead2 in programs where it does not return

AtariDreams:
In most cases, usage does not return, so mark them as __dead2. For the cases where they do return, they have not been marked __dead2. Yes, all cases were checked.

### DIFF
--- a/bin/chflags/chflags.c
+++ b/bin/chflags/chflags.c
@@ -59,7 +59,7 @@ __FBSDID("$FreeBSD$");
 
 static volatile sig_atomic_t siginfo;
 
-static void usage(void);
+static void usage(void) __dead2;
 
 static void
 siginfo_handler(int sig __unused)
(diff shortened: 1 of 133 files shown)


delphij (approved):
Looks good to me.

ngie-eign:

bsdimp:
So what does this offer over and above the _Noreturn marker for exit()?

AtariDreams:
> So what does this offer over and above the _Noreturn marker for exit()? In the past we've not added this to the entire call tree that doesn't return, just the leaves, which the compiler can then infer is no-return... Help me understand what benefit there is here.

For consistency with the other functions already marked as dead2

bsdimp:
OK. That makes sense. Most of the current ones don't really need that since the compiler can guess.
Consistency is good. _Noreturn is the new standard for this stuff, but it's likely not worth it to do that here.

bsdimp:
landed. But for next time:
Break changes up into smaller chunks. I broke this into 15 changes with 10-25 files each.
Also had to fix the commit message to be < 80 characters.

PR #736, opened 2023-05-07
Prefer returning from main over calling exit in main

AtariDreams:
This is redundant otherwise.

### DIFF
--- a/bin/cat/cat.c
+++ b/bin/cat/cat.c
@@ -230,8 +230,7 @@ main(int argc, char *argv[])
 		scanfiles(argv, 0);
 	if (fclose(stdout))
 		err(1, "stdout");
-	exit(rval);
-	/* NOTREACHED */
+	return rval;
 }
 
 static void
--- a/bin/chflags/chflags.c
+++ b/bin/chflags/chflags.c
@@ -212,7 +212,7 @@ main(int argc, char *argv[])
 	}
 	if (errno)
 		err(1, "fts_read");
-	exit(rval);
+	return (rval);
 }
 
 static void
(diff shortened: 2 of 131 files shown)


bsdimp:
Why? Seems like a lot of churn for no gain.

AtariDreams:
> Why? Seems like a lot of churn for no gain.

Gain in terms of, less function calls to do, as well as less confusion for the compiler. Yeah I agree not every instance should be changed, but exit(0) at least in main makes less sense than return 0;

bsdimp:
Any compier that doesn't know exit doesn't return is worthless. We annotate exit for exactly this reason.

I'm strongly inclined to say "no" here...

delphij:
> > Why? Seems like a lot of churn for no gain.
> 
> Gain in terms of, less function calls to do, as well as less confusion for the compiler. Yeah I agree not every instance should be changed, but exit(0) at least in main makes less sense than return 0;

`exit()` is always called regardless if one chooses to `return()` in `main()` or not (see `__libc_start1` in `lib/libc/csu/libc_start1.c` for details).  However, except for a few special cases (for instance, `main()` is called by a different caller, for example when a command is also used as built-in of `sh(1)` where it's desirable to run the command within the same process), calling `exit()` may actually be intentional, and not doing so could confuse a dynamic memory analyzer.

For example, if one allocates memory in `main()` and use an on-stack pointer to reference the memory block, calling `exit()` is slightly different from `return` because for the latter the stack frame would be gone and the memory is leaked, and for the `exit()` case no `free()` is needed (and they shouldn't be used in these cases).

emaste:
@delphij gives a good explanation that provides justification for why a change like this shouldn't be done as a global sweep. If we are doing to make changes like this they should be done on a case-by-case basis and will need to confirm correctness and that it's an improvement in readability.

PR #737, opened 2023-05-08
vfs: fix description comment of vfs_lookup()

ihoro:


mhorne (approved):
Hi, thanks for the submission. 

This change looks good, and I am happy to merge it, but one question: did you audit any other details of the comment? The entire comment is quite old (1994/1995), and I believe the high-level description still applies but maybe some other details are stale.

If not it is okay, I will just make a note of it in the commit message before pushing.

ihoro:
Actually, this is a good question. I simply corrected obvious thing which caught my eyes. From a bit deeper overview I could propose the subsequent commit of changes.

sys/kern/vfs_lookup.c
```diff
@@ -842,34 +842,35 @@ vfs_lookup_failifexists(struct nameidata *ndp)
  *
  * The pathname is pointed to by cn_nameptr and is of length ni_pathlen.
  * The starting directory is taken from ni_startdir. The pathname is
- * descended until done, or a symbolic link is encountered. The variable
- * ni_more is clear if the path is completed; it is set to one if a
+ * descended until done, or a symbolic link is encountered. The cn_flags
+ * has ISLASTCN or'ed if the path is completed or ISSYMLINK or'ed if a
  * symbolic link needing interpretation is encountered.
```
ihoro, line 847: It looks that some time ago it was passed as ni_more variable as a flag, today it works with cn_flags bits instead.

sys/kern/vfs_lookup.c
```diff
@@ -842,34 +842,35 @@ vfs_lookup_failifexists(struct nameidata *ndp)
...
- * If flag has LOCKPARENT or'ed into it, the parent directory is returned
- * locked. If flag has WANTPARENT or'ed into it, the parent directory is
+ * If cn_flags has LOCKPARENT or'ed into it, the parent directory is returned
+ * locked. If it has WANTPARENT or'ed into it, the parent directory is
  * returned unlocked. Otherwise the parent directory is not returned. If
- * the target of the pathname exists and LOCKLEAF is or'ed into the flag
+ * the target of the pathname exists and LOCKLEAF is or'ed into the cn_flags
  * the target is returned locked, otherwise it is returned unlocked.
- * When creating or renaming and LOCKPARENT is specified, the target may not
- * be ".".  When deleting and LOCKPARENT is specified, the target may be ".".
```
ihoro, line 859: Both statements seem to be outdated, probably it makes sense to simply remove them. As I see, at least two cases (RENAME+dot and DELETE+dot) work another way today - it simply returns with EINVAL regardless locking option.

sys/kern/vfs_lookup.c
```diff
@@ -842,34 +842,35 @@ vfs_lookup_failifexists(struct nameidata *ndp)
...
  *
  * Overall outline of lookup:
  *
+ *	handle degenerate case where name is null string
```
ihoro, line 861: Previously it was mentioned to be done upon each dirloop iteration, but today it's done once before the loop.

sys/kern/vfs_lookup.c
```diff
@@ -842,34 +842,35 @@ vfs_lookup_failifexists(struct nameidata *ndp)
...
+ *
  * dirloop:
  *	identify next component of name at ndp->ni_cnd.cn_nameptr
- *	handle degenerate case where name is null string
+ *	handle .. special cases related to capabilities, chroot, jail
```
ihoro, line 865: Just to mention a bit more cases related to the well-known subsystems.

sys/kern/vfs_lookup.c
```diff
@@ -842,34 +842,35 @@ vfs_lookup_failifexists(struct nameidata *ndp)
...
  *	if .. and crossing mount points and on mounted filesys, find parent
  *	call VOP_LOOKUP routine for next component name
  *	    directory vnode returned in ni_dvp, unlocked unless LOCKPARENT set
  *	    component vnode returned in ni_vp (if it exists), locked.
  *	if result vnode is mounted on and crossing mount points,
  *	    find mounted on vnode
  *	if more components of name, do next level at dirloop
+ *	if VOP_LOOKUP returned ERELOOKUP, repeat the same level at dirloop
```
ihoro, line 873: Probably, it's useful to mention a case when dirloop iteration can be repeated for the same component name again. This mechanism was added for specific autofs case: https://github.com/freebsd/freebsd-src/commit/213ed8385587440539d32a2198056db72a10dba4

sys/kern/vfs_lookup.c
bsdimp: let's keep this as a followup... I'm going to merge this, but feel free to followup on this detail

bsdimp:
7b5a1c39f1e6 landed.

PR #738, opened 2023-05-08
bug: efi_print_global only checks for the first 5 letters of "Driver"

AtariDreams:
As a result, it is only really checking for the word Drive, making "Drive" appended to anything else considered for efi env.

### DIFF
--- a/stand/efi/libefi/env.c
+++ b/stand/efi/libefi/env.c
@@ -537,7 +537,7 @@ efi_print_global(const CHAR16 *varnamearg, uint8_t *data, UINTN datasz)
 		goto done;
 	}
 	if (strncmp("Boot", var, 4) == 0 ||
-	    strncmp("Driver", var, 5) == 0 ||
+	    strncmp("Driver", var, 6) == 0 ||
 	    strncmp("SysPrep", var, 7) == 0 ||
 	    strncmp("OsRecovery", var, 10) == 0) {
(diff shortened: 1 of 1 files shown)


delphij (changes requested):
Hi, nice find!  To make the code more readable, could you please change the length numbers to `strlen()`s instead?  (I am pretty sure that the compiler would turn these into number constants at build time).

bsdimp:
Looks good to me... The practical effect of this, however, is limited since we'd filter more things through this gate, but nothing standard conflicts... It's a good fix, and I'll land it.

bsdimp:
So I can't find the name/email that I've used in the past to land your commits, and this commit has the 'github' anonymous email address.

bsdimp:
Oh, wait, in the past I've used
Author: Alfonso <gfunni234@gmail.com>
is that still good?

AtariDreams:
> Oh, wait, in the past I've used Author: Alfonso [gfunni234@gmail.com](mailto:gfunni234@gmail.com) is that still good?

This is still good.

AtariDreams:
@bsdimp

ngie-eign:

AtariDreams:

Done!

kevans91:
Merged as 39ae24e3bf1c8e7d053d0249a6bc88f65eff6de1. Thanks!

PR #739, opened 2023-05-08
Correct size parameter to strncmp

AtariDreams:
Wrong value passed to strncmp means only enable and disable are accepted, when enabled and disabled are supposed to also be accepted.

### DIFF
--- a/usr.sbin/mptutil/mpt_volume.c
+++ b/usr.sbin/mptutil/mpt_volume.c
@@ -246,9 +246,9 @@ volume_cache(int ac, char **av)
 	Settings = volume->VolumeSettings.Settings;
 
 	NewSettings = Settings;
-	if (strncmp(av[2], "enable", sizeof("enable")) == 0)
+	if (strncmp(av[2], "enable", strlen("enable")) == 0)
 		NewSettings |= 0x01;
-	if (strncmp(av[2], "disable", sizeof("disable")) == 0)
+	else if (strncmp(av[2], "disable", strlen("disable")) == 0)
 		NewSettings &= ~0x01;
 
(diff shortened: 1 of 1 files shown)


usr.sbin/mptutil/mpt_volume.c
```diff
@@ -246,9 +246,9 @@ volume_cache(int ac, char **av)
 	Settings = volume->VolumeSettings.Settings;
 
 	NewSettings = Settings;
-	if (strncmp(av[2], "enable", sizeof("enable")) == 0)
+	if (strncmp(av[2], "enable", sizeof("enable") - 1) == 0)
```
delphij, line 249: Could this be changed to use `strlen()` instead?

delphij (changes requested):

usr.sbin/mptutil/mpt_volume.c
ngie-eign: @AtariDreams : I agree with @delphij. Is it ok if we/I apply his suggestion and merge the change on your behalf?

ngie-eign: FWIW, it's functionally the same as written, unless `sizeof(av[2])` becomes a `char*` quantity (in which case it will always chop the string off at 8 characters on 64-bit platforms :D...).

AtariDreams: Yes that is okay! @ngie-eign

ngie-eign (approved):

delphij (approved):

lwhsu:
merged as 5a9c724847f9b4e3831aa2c16276cc2ae20a99cc .

PR #743, opened 2023-05-16
passwd.5: Update man page after e0155c6

jlduran:
Two commits:
1. Lint, by removing trade name macros.
2. Update the man page after e0155c6.

cc/ @bapt 

---

NB: I do not understand why not check for `validchars` instead of `badchars`... anyhow, the man page needs updating.


bapt:
merged thank you, as the badchar vs valid char, I have no idea, but yes probably worth reverting the logic.

PR #747, opened 2023-05-19
ifconfig: skip calling fnmatch once the result no longer matters

AtariDreams:
Because fnmatch has no side effects, we can safely avoid calling fnmatch if the end result does not matter anyway (the compiler cannot see this, so it calls fnmatch in the event it has side-effects).


ngie-eign:

AtariDreams:
> Please change the email address to a non-users.noreply.github.com based address.

Done!

bsdimp:
This looks good, but for future reference please format the commit message that follows https://docs.freebsd.org/en/articles/committers-guide/#_ports_considerations (except the port-specific recommendation to use cat/port), namely commit messages should be < 72 characters..

bsdimp:
21c32cebf81a pushed

PR #748, opened 2023-05-19
Prefer unsigned variables over signed variables if they will be treated as such

AtariDreams:
There is no reason for variables to be signed if they will be passed to functions that expect unsigned or have explicitly unsigned operations done to them.

### DIFF
--- a/lib/libifconfig/libifconfig.c
+++ b/lib/libifconfig/libifconfig.c
@@ -494,7 +494,7 @@ int
 ifconfig_get_groups(ifconfig_handle_t *h, const char *name,
     struct ifgroupreq *ifgr)
 {
-	int len;
+	unsigned int len;
 
 	memset(ifgr, 0, sizeof(*ifgr));
 	strlcpy(ifgr->ifgr_name, name, IFNAMSIZ);
--- a/sbin/ifconfig/af_nd6.c
+++ b/sbin/ifconfig/af_nd6.c
@@ -99,7 +99,7 @@ void
 setnd6defif(if_ctx *ctx, const char *dummyaddr __unused, int d)
 {
 	struct in6_ndifreq ndifreq = {};
-	int ifindex;
+	unsigned int ifindex;
 	int error;
 
 	strlcpy(ndifreq.ifname, ctx->ifname, sizeof(ndifreq.ifname));
--- a/sbin/ifconfig/ifconfig.c
+++ b/sbin/ifconfig/ifconfig.c
@@ -782,7 +782,7 @@ list_interfaces_ioctl(if_ctx *ctx)
 	struct ifaddrs *ifap, *sifap, *ifa;
 	struct ifa_order_elt *cur, *tmp;
 	char *namecp = NULL;
-	int ifindex;
+	size_t ifindex;
 	struct ifconfig_args *args = ctx->args;
 
 	if (getifaddrs(&ifap) != 0)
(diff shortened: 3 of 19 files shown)


igalic (changes requested):
can you please make one atomic, descriptive commit per subsystem

pstef:
To quote style(9):

>      The project prefers the use of ISO/IEC 9899:1999 (“ISO C99”) unsigned
>      integer identifiers of the form uintXX_t rather than the older BSD-style
>      integer identifiers of the form u_intXX_t.  New code should use the
>      former, and old code should be converted to the new form if other major
>      work is being done in that area and there is no overriding reason to
>      prefer the older BSD-style.  Like white-space commits, care should be
>      taken in making uintXX_t only commits.

AtariDreams:
> To quote style(9):
> 
> > ```
> >  The project prefers the use of ISO/IEC 9899:1999 (“ISO C99”) unsigned
> >  integer identifiers of the form uintXX_t rather than the older BSD-style
> >  integer identifiers of the form u_intXX_t.  New code should use the
> >  former, and old code should be converted to the new form if other major
> >  work is being done in that area and there is no overriding reason to
> >  prefer the older BSD-style.  Like white-space commits, care should be
> >  taken in making uintXX_t only commits.
> > ```

Fixed

bsdimp:
I'm inclined to close this too. The author field is wrong, despite warnings in the past about it. It's still one huge commit, despite requests to make it smaller. I'd be inclined to close this pull request and request that each change be submitted separately with some justification for each one besides a preference. Many of these are likely good, and would go in on their own, if they were presented individually.

However, one big batch like this is impossible: it creates a lot of extra work for the people landing the code. Extra work that starves out more useful work and the return on investment for it is low. We have a general bias against purely stylistic changes because in bulk they can be used to hide one or two real changes that might create a bug. In bulk they are also hard to evaluate individually. The project discourages active style sweeps unless you are working in a related area (eg, ifconfig clean can include style fixes, but can't be only style fixes).

I'm not  going to close this one just yet... But plan to do so shortly... I'd suggest that @AtariDreams be more strategic in the future. Target ifconfig in one pull request, pfctl in another, etc. And to space them out a bit...

igalic:
just to elaborate on @bsdimp's

> The author field is wrong, despite warnings in the past about it.

```mbox
From: Rose <83477269+AtariDreams@users.noreply.github.com> 
```

Rose is a perfectly fine name. However, the project would like to be able to contact people who worked on code or docs via email, and that seems pretty much impossible with an email address like that.

bsdimp:
Rose has fixed the author in the past with an acceptable email address for me. We really like to have that, or an explicit statement to use something else if there's a good reason that's a problem (you can email me privately if you need to)... I can also use the old address, but that's more manual steps and extra time for me to process the pull request...

AtariDreams:
@bsdimp OK

igalic:
> can you please make one atomic, descriptive commit per subsystem

this still applies

bsdimp:
There's a couple of things:
(1) As requested elsewhere one commit per subsystem.
(2) All the u_int -> unsigned int changes are bogus. It's a misapplication of style(9). There's a lot of needless churn here
(3) This falls under the 'style commit for style commit sake' at the moment. Some justification for such a sweeping change is needed. Please explain the actual bug being fixed, ideally on a per-change basis.

bsdimp:
Changes requested months ago. Closing. If the changes happen, then please resubmit. But bear in mind that we'll want proof of testing all these utilities to prove there's no regression.

PR #749, opened 2023-05-21
Add id vendor for USB modem Quectel and align with quectel's marketing names.

eros23:
Request for adding 3g/LTE/5G modems from vendor quectel.
I need to get them to work on devices like opnsense, pfsense without recompiling the kernel driver.

### DIFF
--- a/sys/dev/mlx4/mlx4_core/mlx4_main.c
+++ b/sys/dev/mlx4/mlx4_core/mlx4_main.c
@@ -1164,6 +1164,8 @@ err_sup:
 	return err;
 }
 
+static DEFINE_MUTEX(set_port_type_mutex);
+
 static ssize_t set_port_type(struct device *dev,
 			     struct device_attribute *attr,
 			     const char *buf, size_t count)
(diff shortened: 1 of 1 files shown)


bsdimp:
Thanks for updating the manual. That's often forgotten. The new ids look good to my eye as well. This is almost ready, but I have a question and a suggestion for the commit message.

It looks like you've changed some of the old ones too. And the commit message is a little short.

Maybe update the commit message with why for the old ones? Eg "align with quectel's marketing names." Or something like that. That would address both of these things.

eros23:
Thank you.
do you say to defer all the new commit or do I just change the title above?
the old names referred to specific boards, for example EP06_E referred to the European model but with the same board they also refer to EUX etc .
I have seen, from their manuals, that they have given this type of nomenclature.

bsdimp:
The title update is enough. I'll land this when I get home in a few hours

eros23:
Sorry, don't send anything. I ate a "T" of a device. I check everything and completely redo the new commit, with lots of details.

bsdimp:
Why was this closed? It looks good to my eye and I'd hoped to commit it this week after BSDcan and some other travel I had to do.

bsdimp:
Oh, replaced by #750.

PR #750, opened 2023-05-21
Add vendor id for Quectel USB modem and align with quectel marketing names

eros23:
Added support for 3G/4G-LTE/5G USB modems and alignment with vendor trade names.

Most devices work fine with ppp.


eros23:
@bsdimp I had to correct a "T" on a template. this is the correct one.
These are the versions of the 3G/4G-LTE/5G modems available, according to the vendor.

bsdimp:
Looks good to me. Pushed as d3a83456e1e3

PR #751, opened 2023-05-25
ci: Make quick checker GitHub PR dst branch agnostic

ihoro:
This is for GitHub CI Workflow testing in case of a pull request. See https://reviews.freebsd.org/D40261.

### DIFF
--- a/.github/workflows/prereq.yml
+++ b/.github/workflows/prereq.yml
@@ -1,0 +1,34 @@
+---
+name: Quick Checker
+
+on:
+  push:
+    branches: [ 'ci-*' ]
+  pull_request:
+    branches: [ 'main' ]
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 500
+      - name: Pull pain
+        run: git fetch --no-tags --no-recurse-submodules --depth=500 origin main
+      - name: Environment
+        run: |
+          echo $GITHUB_REF_NAME
+          git rev-list -1 $GITHUB_REF_NAME
+          git rev-list -1 origin/main
+      - name: Recent enough main?
+        run: git merge-base origin/main $GITHUB_REF_NAME
+      - name: Check Email
+        run:  sh ./tools/build/check-email.sh origin/main..$GITHUB_REF_NAME
+      - name: Check Style
+        run:  ./tools/build/checkstyle9.pl -color=always origin/main..$GITHUB_REF_NAME
--- a/.github/workflows/prereq.yml
+++ b/.github/workflows/prereq.yml
@@ -2,6 +2,8 @@
 name: Quick Checker
 
 on:
+  push:
+    branches: [ 'ci-*' ]
   pull_request:
 
 permissions:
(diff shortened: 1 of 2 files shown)


.github/workflows/prereq.yml
```diff
@@ -18,13 +20,24 @@ jobs:
...
+          if test -n "$GITHUB_BASE_REF"; then
+            BASEB=$(git rev-list -1 HEAD^1)
+            HEADB=$(git rev-list -1 HEAD^2)
+          else
+            git fetch --no-tags --no-recurse-submodules --depth=500 origin main
+            BASEB=origin/main
+            HEADB=HEAD
+          fi
+          echo "BASEB: $BASEB"
+          echo "HEADB: $HEADB"
```
bsdimp, line 35: these two lines look redundant with the following two.. You'll see the values in the logs, would you not

.github/workflows/prereq.yml
```diff
@@ -18,13 +20,24 @@ jobs:
...
-          git rev-list -1 HEAD^2
+          echo "GITHUB_REF_NAME: $GITHUB_REF_NAME"
+          echo "GITHUB_BASE_REF: $GITHUB_BASE_REF"
+          echo "HEAD: $(git rev-list -1 HEAD)"
+          if test -n "$GITHUB_BASE_REF"; then
+            BASEB=$(git rev-list -1 HEAD^1)
+            HEADB=$(git rev-list -1 HEAD^2)
+          else
+            git fetch --no-tags --no-recurse-submodules --depth=500 origin main
+            BASEB=origin/main
```
bsdimp, line 31: Also, this has a problem I'm not sure how to address.
If my github main branch is super old (because I didn't push the main ref, only my branch's ref),
then this gives crazy results sometimes (though enough bad things happen and are caught by the 23 bad email addresses we have in the repo today).
Also, we only fetch main's refs, not the branches...

.github/workflows/prereq.yml
```diff
@@ -18,13 +20,24 @@ jobs:
...
-          echo $GITHUB_REF_NAME
-          git rev-list -1 HEAD
-          git rev-list -1 HEAD^1
-          git rev-list -1 HEAD^2
+          echo "GITHUB_REF_NAME: $GITHUB_REF_NAME"
+          echo "GITHUB_BASE_REF: $GITHUB_BASE_REF"
+          echo "HEAD: $(git rev-list -1 HEAD)"
+          if test -n "$GITHUB_BASE_REF"; then
+            BASEB=$(git rev-list -1 HEAD^1)
+            HEADB=$(git rev-list -1 HEAD^2)
```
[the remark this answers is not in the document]
ihoro: If I got it right then I will try to guess that, probably, it was read as `HEAD~1` and `HEAD~2`.

Upon pull request GitHub runs the job against a merge commit, where the first parent `HEAD^1` is `$GITHUB_BASE_REF` (e.g. freebsd/freebsd-src/main branch) and the second parent `HEAD^2` is `$GITHUB_HEAD_REF` (e.g. contributor/freebsd-src/some-fix branch). The reasoning details can be found here: https://reviews.freebsd.org/D40261#916852.
With fetch-depth=500 we get both parents 500 commits deep, i.e. both base and head branches are already fetched by actions/checkout. The only nuance is that `$GITHUB_BASE_REF` and `$GITHUB_HEAD_REF` references are not present within a runner's git repo. The only reference we have is named something like `refs/remotes/pull/<PRID>/merge` and points to this merge commit. We could invoke additional git fetch, but it won’t bring anything new, both branches' heads are already fetched, they are `HEAD^1` and `HEAD^2`.

This PR #751 is the test itself to show how it works: https://github.com/freebsd/freebsd-src/actions/runs/5082844700/jobs/9133072060.

Yes, it may look tricky from `prereq.yml` reader perspective. Hence, it looks like we have at least two options: leave it shorter without additional git fetch commands (maybe with some comment added), or additionally invoke git fetch `$GITHUB_BASE_REF` to make it more obvious. I think your experience has better vision which way fits.

.github/workflows/prereq.yml
ihoro: The following two lines depict the official way of env vars passing to the next steps of a job: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable.
Probably, it was meant that reporting the values in the Environment step could be omitted due to they can be spotted in the commands of the next steps. If we prefer such way, then yes, the lines 34-35 can be omitted.

.github/workflows/prereq.yml
ihoro: Yes, a tricky question. I personally do not see an optimal solution for now. Probably, this additional option can be considered: https://reviews.freebsd.org/D40261#916935

.github/workflows/prereq.yml
```diff
@@ -0,0 +1,44 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 500
```
jlduran, line 20: I still believe that we should pull in the entire repo:
```yaml
fetch-depth: 0
```
It currently takes about 4 minutes, so it's not that bad.

bsdimp:
Thanks for the discussion. I'll fold it into my work. Closing for now. I'll submit a pull request with what I land on.

PR #754, opened 2023-05-27
amdtemp: Fix missing 49 degree offset on current EPYC CPUs

valpackett:
On an EPYC 7313P, the temperature reported by amdtemp was off, because the offset was not applied. Turns out it needs to be applied with one more condition: https://lkml.org/lkml/2023/4/13/1095

Tested by:      mike.jakubik@gmail.com
Sponsored by:   https://www.patreon.com/valpackett

### DIFF
--- a/sys/dev/amdtemp/amdtemp.c
+++ b/sys/dev/amdtemp/amdtemp.c
@@ -166,6 +166,12 @@ static const struct amdtemp_product {
 #define	AMDTEMP_17H_CUR_TMP		0x59800
 #define	AMDTEMP_17H_CUR_TMP_RANGE_SEL	(1u << 19)
 /*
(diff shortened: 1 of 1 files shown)


MikeJakubik:
Hi,

I am the one that reported the issue of Discord and tested the patch, works fine on a recent build of 13-STABLE  running on a Lenovo SR635, can also test on a SR655 if need be. Would love to see this make it into stable as this is throwing off our monitoring software (Netdata).

Thank You.

valpackett:
@emaste ping

mhorne (approved):
Seems good to me. I will grab this and push to `main`, and then stable branches after a few days.

bsdimp:
This has landed in main, so closing.

PR #755, opened 2023-05-28
usr.bin/rctl/rctl.8 document `group` subject to hierarchical resource limits

photograveur:
According to its source code, `rctl(8)` seems to support `group`/`g` subjects. This information was missing in the manual page.

### DIFF
--- a/usr.bin/rctl/rctl.8
+++ b/usr.bin/rctl/rctl.8
@@ -24,7 +24,7 @@
 .\"
 .\" $FreeBSD$
 .\"
-.Dd February 26, 2018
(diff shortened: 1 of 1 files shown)


bsdimp:
This looks like it was basically good... why was this closed?

photograveur:
> This looks like it was basically good... why was this closed?

Sorry, I should have looked at the kernel’s sources. There are evidently some preparations in `rctl`, but the kernel’s facilities do not support group subjects.

bsdimp:
Gotcha. Thanks for the explanation.

PR #758, opened 2023-05-31
unbound: Request the OpenSSL 1.1 API

khorben:
OPENSSL_API_COMPAT can be used to specify the OpenSSL API version in use for the purpose of hiding deprecated interfaces and enabling the appropriate deprecation notices.

This change is a NFC while we're still using OpenSSL 1.1.1 but will avoid deprecation warnings upon the switch to OpenSSL 3.0.

A future update may migrate to use the OpenSSL 3.0 APIs.

PR:		271615


khorben:
Review requested at https://reviews.freebsd.org/D40361.

bsdimp:
OBE

PR #759, opened 2023-06-01
stand/lua: per-product conf if requested via product_vars

stephane-rochoy-stormshield:
This patch allow for picking loader configuration files according to the product (in the sense of `smbios.system.product`).

If `product_vars` is set, it must be a space separated list of environment variable names to walk through to guess the product. And if a product can be guessed, prepend:
- `/boot/PRODUCT.conf` to `loader_conf_files`
- and `/boot/loader.conf.d/PRODUCT` to `loader_conf_dirs`

I made `product_vars` a list to let have a single value suitable for a range of different products. 

For example, you can use it with `smbios.system.product uboot.m_product`  if you have U-Boot exposing the product this way. 

I plan to propose applying this idea to pick DTB on ARM platforms too.

### DIFF
--- a/stand/defaults/loader.conf.5
+++ b/stand/defaults/loader.conf.5
@@ -21,7 +21,7 @@
 .\" LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
-.Dd January 9, 2024
+.Dd January 18, 2024
 .Dt LOADER.CONF 5
 .Os
 .Sh NAME
@@ -131,6 +131,10 @@ Space separated list of directories to process for configuration files.
 The lua-based loader will process files with a
 .Dq .conf
 suffix that are placed in these directories.
+Files found here are processed after the ones listed in
+.Va loader_conf_files
+but before the ones found in
+.Va local_loader_conf_files .
 .It Ar loader_conf_files
 Defines additional configuration files to be processed right after the
 present file.
@@ -138,6 +142,13 @@ present file.
 should be treated as write-only.
 One cannot depend on any value remaining in the loader environment or carried
 over into the kernel environment.
+.It Ar local_loader_conf_files
+Space separated list of additional configuration files to be processed at last,
+i.e., after
+.Va loader_conf_files
+and
+.Va loader_conf_dirs
+are processed.
 .It Ar product_vars
 When set, must be a space separated list of environment variable names to walk
 through to guess product information.
@@ -274,6 +285,14 @@ default settings can be ignored.
 The few of them which are important
 or useful are:
 .Bl -tag -width bootfile -offset indent
+.It Va local_loader_conf_files
+.Pq Dq /boot/loader.conf.local
+Ensure
+.Va loader.conf.local
+can always be used to override settings from files found in
+.Va loader_conf_files
+and
+.Va loader_conf_dirs .
 .It Va bitmap_load
 .Pq Dq NO
 If set to
@@ -454,13 +473,18 @@ It is not available in the default Forth-based loader.
 .Sh FILES
 .Bl -tag -width /boot/defaults/loader.conf -compact
 .It Pa /boot/defaults/loader.conf
-default settings \(em do not change this file.
+Default settings \(em do not change this file.
 .It Pa /boot/loader.conf
-user defined settings.
+User defined settings.
 .It Pa /boot/loader.conf.lua
-user defined settings written in lua.
+User defined settings written in lua.
+.It Pa /boot/loader.conf.d/*.conf
+User defined settings split in separate files.
+.It Pa /boot/loader.conf.d/*.lua
+User defined settings written in lua and split in separate files.
 .It Pa /boot/loader.conf.local
-machine-specific settings for sites with a common loader.conf.
+Machine-specific settings for sites with a common loader.conf. Allow to override
+settings defined in other files.
 .El
 .Sh SEE ALSO
 .Xr loader.conf.lua 5 ,
--- a/stand/defaults/loader.conf.5
+++ b/stand/defaults/loader.conf.5
@@ -21,7 +21,7 @@
 .\" LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
-.Dd July 31, 2021
+.Dd January 9, 2024
 .Dt LOADER.CONF 5
 .Os
 .Sh NAME
@@ -138,6 +138,38 @@ present file.
 should be treated as write-only.
 One cannot depend on any value remaining in the loader environment or carried
 over into the kernel environment.
+.It Ar product_vars
+When set, must be a space separated list of environment variable names to walk
+through to guess product information.
+The order matters as reading a config file override the previously defined
+values.
+Undefined variables are silently ignored.
+.Pp
+When product information can be guessed, for each product information found,
+append
+.Pa /boot/loader.conf.d/PRODUCT
+to
+.Ar loader_conf_dirs .
+It can be typically used as follow:
+.Bd -literal
+smbios.system.planar.maker="PLANAR_MAKER"
+smbios.system.planar.product="PLANAR_PRODUCT"
+smbios.system.product="PRODUCT"
+uboot.m_product="M_PRODUCT"
+product_vars="smbios.system.planar.maker smbios.system.planar.product smbios.system.product uboot.m_product"
+.Ed
+.Pp
+to read files found in the following directories, in that order:
+.Bl -bullet -compact
+.It
+.Pa /boot/loader.conf.d/PLANAR_MAKER
+.It
+.Pa /boot/loader.conf.d/PLANAR_PRODUCT
+.It
+.Pa /boot/loader.conf.d/PRODUCT
+.It
+.Pa /boot/loader.conf.d/M_PRODUCT
+.El
 .It Ar kernel
 Name of the kernel to be loaded.
 If no kernel name is set, no additional
--- a/stand/lua/config.lua
+++ b/stand/lua/config.lua
@@ -658,12 +658,37 @@ function config.readConf(file, loaded_files)
 
 	if load_conf_dirs then
 		local loader_conf_dirs = getEnv("loader_conf_dirs")
+
+		-- If product_vars is set, it must be a list of environment variable names
+		-- to walk through to guess product information. The order matters as
+		-- reading a config files override the previously defined values.
+		--
+		-- If product information can be guessed, for each product information
+		-- found, also read config files found in /boot/loader.conf.d/PRODUCT/.
+		local product_vars = getEnv("product_vars")
+		if product_vars then
+			local product_conf_dirs = ""
+			for var in product_vars:gmatch("%S+") do
+				local product = getEnv(var)
+				if product then
+					product_conf_dirs = product_conf_dirs .. " /boot/loader.conf.d/" .. product
+				end
+			end
+
+			if loader_conf_dirs then
+				loader_conf_dirs = loader_conf_dirs .. product_conf_dirs
+			else
+				loader_conf_dirs = product_conf_dirs
+			end
+		end
+
 		if loader_conf_dirs ~= nil then
 			for name in loader_conf_dirs:gmatch("[%w%p]+") do
 				if lfs.attributes(name, "mode") ~= "directory" then
 					print(MSG_FAILDIR:format(name))
 					goto nextdir
 				end
+
 				for cfile in lfs.dir(name) do
 					if cfile:match(".conf$") then
 						local fpath = name .. "/" .. cfile
--- a/stand/lua/config.lua
+++ b/stand/lua/config.lua
@@ -630,8 +630,7 @@ function config.readConf(file, loaded_files)
 		return
 	end
 
-	-- We'll process loader_conf_dirs at the top-level readConf
-	local load_conf_dirs = next(loaded_files) == nil
+	local top_level = next(loaded_files) == nil -- Are we the top-level readConf?
 	print("Loading " .. file)
 
 	-- The final value of loader_conf_files is not important, so just
@@ -656,7 +655,7 @@ function config.readConf(file, loaded_files)
 		end
 	end
 
-	if load_conf_dirs then
+	if top_level then
 		local loader_conf_dirs = getEnv("loader_conf_dirs")
 
 		-- If product_vars is set, it must be a list of environment variable names
@@ -682,6 +681,7 @@ function config.readConf(file, loaded_files)
 			end
 		end
 
+		-- Process "loader_conf_dirs" extra-directories
 		if loader_conf_dirs ~= nil then
 			for name in loader_conf_dirs:gmatch("[%w%p]+") do
 				if lfs.attributes(name, "mode") ~= "directory" then
@@ -698,6 +698,15 @@ function config.readConf(file, loaded_files)
 					end
 				end
 				::nextdir::
+			end
+		end
+
+		-- Always allow overriding with local config files, e.g.,
+		-- /boot/loader.conf.local.
+		local local_loader_conf_files = getEnv("local_loader_conf_files")
+		if local_loader_conf_files then
+			for name in local_loader_conf_files:gmatch("[%w%p]+") do
+				config.readConf(name, loaded_files)
 			end
 		end
 	end
--- a/UPDATING
+++ b/UPDATING
@@ -27,6 +27,27 @@ NOTE TO PEOPLE WHO THINK THAT FreeBSD 15.x IS SLOW:
 	world, or to merely disable the most expensive debugging functionality
 	at runtime, run "ln -s 'abort:false,junk:false' /etc/malloc.conf".)
 
+20240118:
+	Loader now also read configuration files listed in local_loader_conf_files.
+	Files listed here are the last ones read. And /boot/loader.conf.local was
+	moved from loader_conf_files to local_loader_conf_files leaving only
+	loader.conf and device.hints in loader_conf_files by default.
+
+	The following sequencing is applied:
+
+	1. Bootstrap:
+	    /boot/defaults/loader.conf
+
+	2. Read loader_conf_files files:
+	    /boot/device.hints
+	    /boot/loader.conf
+
+	3. Read loader_conf_dirs files:
+	    /boot/loader.conf.d/*.conf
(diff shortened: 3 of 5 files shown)


bsdimp:
I think I like it. Will review in detail when I'm at my desk.

stand/lua/config.lua
```diff
@@ -578,8 +578,35 @@ function config.readConf(file, loaded_files)
...
+	if product ~= nil then
+		local product_conf_file = string.format("/boot/%s.conf", product)
+		if lfs.attributes(product_conf_file, "mode") == "file" then
+			if loader_conf_files == nil then
+				loader_conf_files = product_conf_file
+			else
+				loader_conf_files = product_conf_file .. " " .. loader_conf_files
+			end
+		end
+	end
```
bsdimp, line 609: I like this, but think it would be a little better if you had the "for vars" loop around basically this entire section so you could do that for multiple variables that match. I might want to have certain variables set for all of my Dell servers, regardless of model, as well as overrides for a specific motherboard. If I set product_vars to "smbios.planar.maker smbios.planar.product" then I could do just that if the loop were around this whole thing.

stand/defaults/loader.conf.5
```diff
@@ -140,6 +140,27 @@ present file.
...
+and
+.Pa /boot/loader.conf.d/PRODUCT
+to
+.Ar loader_conf_dirs
+.El
+
+It can be typically used as follow:
+.Pp
+.Dl product_vars="smbios.system.product"
+
```
bsdimp, line 163: If you take up my suggestion below, then you'd need to tweak this to note it is done for all matches.

bsdimp:
My first impressions were right: I do like this. I have a suggestion to make it a little more general without a lot of work. Please consider it and let me know what you think.

stand/defaults/loader.conf.5
bsdimp: oh, and bump .Dd at the top please.

bsdimp:
The commit message and getting all the details I check there right.

stephane-rochoy-stormshield:
As far as I understand, your suggestion is about giving the possibility to aggregate config files using product information instead of just trying to pick one. This is clearly a good suggestion that I'm willing to implement :smile: 

So continuing with your Dell servers:

    product_vars="smbios.system.maker smbios.system.product"
    smbios.planar.maker="Dell Inc."
    smbios.planar.product="0H28RR"

should lead to reading the following files and directories:

    /boot/Dell Inc..conf
    /boot/0H28RR.conf
    /boot/loader.conf.d/Dell Inc./*.conf
    /boot/loader.conf.d/0H28RR/*.conf

If I'm correct, you're probably guessing the two questions I'll ask:
- Should I somehow sanitize/normalize the values, e.g., convert `Dell Inc.` into `Dell_Inc`? (It don't seems to be technically required, but the resulting name looks weird to me :wink:)
- As reading order would now matters, shouldn't I append the new path instead of prepending them?

stephane-rochoy-stormshield:
[8e6b04f](https://github.com/freebsd/freebsd-src/pull/759/commits/8e6b04f3dbb75a685dbe23910e356362ba8dcc8b) is a PoC of your suggestion (without normalizing values).

stand/lua/config.lua
```diff
@@ -578,8 +578,35 @@ function config.readConf(file, loaded_files)
...
+
 	-- Going to process "loader_conf_files" extra-files
 	local loader_conf_files = getEnv("loader_conf_files")
+	if products ~= nil then
+		local product_loader_conf_files = "/boot/" .. table.concat(products, ".conf /boot/") .. ".conf"
+		if loader_conf_files ~= nil then
+			loader_conf_files = loader_conf_files .. " " .. product_loader_conf_files
+		else
+			loader_conf_files = product_loader_conf_files
+		end
```
bsdimp, line 608: Maybe we should append both the .conf and .lua versions of the file. We'll silently ignore the one that doesn't exist. We don't need to do the existence check that I thought we might because config.readConf will silently ignore missing files.

stephane-rochoy-stormshield:
My understanding is that we start by processing files and then directories. While we would like to finish by processing files and directories for each product var in order. A naive approach could be to simply move product vars processing as `readConf`'s last step. 

Kyle's patch you're referring to is 3cb2f5f, i.e., support for `.lua` config files, right? If yes, are you suggesting that we should just have `product_vars=smbios.planar.maker` append `Dell Inc..conf.lua` to `loader_conf_files` with something like:

```lua
local product = getEnv("smbios.planar.product")
if product = "X10SRH-CF" then
    # serial(0) is COM1 and serial(1) is COM2 
elsif product = "X9SRL-F" then
    # serial(1) is COM1 and serial(2) is COM2
end
```

Or are you more thinking about using `.lua` to populate `loader_conf_files` and `loader_conf dirs`? E.g., have `loader_conf_files` contains some `platform.conf.lua` with the following logic:

```lua
local maker = getEnv("smbios.planar.maker ")
if maker ~= nil then
    loader_conf_files.append("/boot/" .. maker .. ".conf")
    loader_conf_files.append("/boot/" .. maker .. ".conf.lua")
    loader_conf_dirs.append("/boot/loader.conf.d/" .. maker)

    local product = getEnv("smbios.planar.product")
    loader_conf_files.append("/boot/" .. product .. ".conf")
    loader_conf_files.append("/boot/" .. product .. ".conf.lua")
    loader_conf_dirs.append("/boot/loader.conf.d/" .. maker .. "/" .. product)
end
```

stephane-rochoy-stormshield:
The fact that `readConf` is a recursive function makes things a bit difficult IMO. So I simplified a bit the patch to only use `product_vars` to extend `loader_conf_dirs` and leave `loader_conf_files` untouched. 

It allow to hook into the `load_conf_dirs` code path that is expected to be processed at the top-level `readConf`.

If I was correct about Kyle's patch you were referring to, `.lua` files should be handled just fine through `processFile` normally.

Let me know what do you think @bsdimp :)

bsdimp:
I like this. I have a few nits now that I like this form. Sorry I haven't commented sooner, it's been a bust month.

stand/defaults/loader.conf.5
```diff
@@ -23,7 +23,7 @@
 .\" SUCH DAMAGE.
 .\"
 .\" $FreeBSD$
-.Dd July 31, 2021
+.Dd June 2, 2023
```
bsdimp, line 26: Today's date, though if that's the only issue I fix that on commit :)

stand/defaults/loader.conf.5
```diff
@@ -140,6 +140,32 @@ present file.
 should be treated as write-only.
 One cannot depend on any value remaining in the loader environment or carried
 over into the kernel environment.
+.It Ar product_vars
+When set, must be a space separated list of environment variable names to walk
+through to guess product information. The order matters as reading a config file
+override the previously defined values.
+
```
bsdimp, line 147: Turn this blank line into .Pp

stand/defaults/loader.conf.5
```diff
@@ -140,6 +140,32 @@ present file.
 should be treated as write-only.
 One cannot depend on any value remaining in the loader environment or carried
 over into the kernel environment.
+.It Ar product_vars
+When set, must be a space separated list of environment variable names to walk
+through to guess product information. The order matters as reading a config file
+override the previously defined values.
```
bsdimp, line 146: start a new sentence on a new line.

stand/defaults/loader.conf.5
```diff
@@ -140,6 +140,32 @@ present file.
...
+
+If product information can be guessed, for each product information found,
+append
+.Pa /boot/loader.conf.d/PRODUCT
+to
+.Ar loader_conf_dirs .
+
```
bsdimp, line 153: Blank line -> .Pp

bsdimp:
anyway, I'll push this in with these final tweaks.

stephane-rochoy-stormshield:
I fixed loader.conf(5) :)

stand/lua/config.lua
```diff
@@ -657,12 +677,40 @@ function config.readConf(file, loaded_files)
...
+
+		-- If product_vars is set, it must be a list of environment variable names
+		-- to walk through to guess product information. The order matters as
+		-- reading a config files override the previously defined values.
+		--
+		-- If product information can be guessed, for each product information
+		-- found, also read config files found in /boot/loader.conf.d/PRODUCT/.
+		local product_vars = getEnv("product_vars")
+		if product_vars ~= nil then
+			print("product_vars: " .. product_vars) -- XXX
```
kevans91, line 689: Noting that this and the one below at 698 should go away pre-push

stand/lua/config.lua
```diff
@@ -657,12 +677,40 @@ function config.readConf(file, loaded_files)
...
 		local loader_conf_dirs = getEnv("loader_conf_dirs")
+
+		-- If product_vars is set, it must be a list of environment variable names
+		-- to walk through to guess product information. The order matters as
+		-- reading a config files override the previously defined values.
+		--
+		-- If product information can be guessed, for each product information
+		-- found, also read config files found in /boot/loader.conf.d/PRODUCT/.
+		local product_vars = getEnv("product_vars")
+		if product_vars ~= nil then
```
kevans91, line 688: At some point I think I'd like to do a sweep of all of the lua in the tree and convert explicit comparisons against `nil` to `foo` or `not foo` as appropriate... probably 99% of occurrences today can be converted because they don't need to care about `nil` vs `false`.

kevans91:
> The fact that `readConf` is a recursive function makes things a bit difficult IMO. So I simplified a bit the patch to only use `product_vars` to extend `loader_conf_dirs` and leave `loader_conf_files` untouched.
> 

FWIW, I'm open to the suggestion that how we do all of this can be improved... a lot of this was written in the same frantic way that I normally write new lua stuff; i.e., it ends up functional but maybe not the most ideal. The main complication we need to deal with is that each configuration file can specify more loader_conf_file and loader_conf_dirs to process after we're done, and I think we need to do them as they're specified (i.e. if the top-level is `loader_conf_files="/boot/loader.conf /boot/loader.local.conf"` and /boot/loader.conf specifies `loader_conf_files="/boot/loader.kevans.conf"`, we need to read `loader.kevans.conf` before `loader.conf.local` to get the ordering right.

stephane-rochoy-stormshield:
> > The fact that `readConf` is a recursive function makes things a bit difficult IMO. So I simplified a bit the patch to only use `product_vars` to extend `loader_conf_dirs` and leave `loader_conf_files` untouched.
> 
> FWIW, I'm open to the suggestion that how we do all of this can be improved... a lot of this was written in the same frantic way that I normally write new lua stuff; i.e., it ends up functional but maybe not the most ideal. The main complication we need to deal with is that each configuration file can specify more loader_conf_file and loader_conf_dirs to process after we're done, and I think we need to do them as they're specified.
> 
> i.e. if the top-level is `loader_conf_files="/boot/loader.conf /boot/loader.local.conf"` and /boot/loader.conf specifies `loader_conf_files="/boot/loader.kevans.conf"`, we need to read `loader.kevans.conf` before `loader.local.conf` to get the ordering right.

I see. But what about `loader_conf_dirs`? The `load_conf_dirs = next(loaded_files) == nil` line let me think that we gather modifications of `loader_conf_dirs` from the recursive calls to `readConf` but we'll read the corresponding files only once. Which means modifications of `loader_conf_dirs` from this files are ignored, right?

stephane-rochoy-stormshield:
@bsdimp @kevans91 I played a bit more with this patch and found a problem: as `loader_conf_files` are processed before `loader_conf_dirs`, a file like `loader.conf.local` is unable to override something found in `loader.conf.d/`.

I'm thinking about introducing a specific variable, `local_loader_conf_files`, to address this issue. It would avoid having to deal with sequencing the files found in `loader.conf.d/`, IMO.

What to you guys think?

bsdimp:
> @bsdimp @kevans91 I played a bit more with this patch and found a problem: as `loader_conf_files` are processed before `loader_conf_dirs`, a file like `loader.conf.local` is unable to override something found in `loader.conf.d/`.
> 
> I'm thinking about introducing a specific variable, `local_loader_conf_files`, to address this issue. It would avoid having to deal with sequencing the files found in `loader.conf.d/`, IMO.
> 
> What to you guys think?

I think I like the idea... But could you describe a little better what the sequencing would then become?

stephane-rochoy-stormshield:
> But could you describe a little better what the sequencing would then become?

With the following in `defaults/loader.conf`:

    product_vars="smbios.system.maker smbios.system.product"
    smbios.planar.maker="Dell Inc."
    smbios.planar.product="0H28RR"

    loader_conf_files="/boot/device.hints /boot/loader.conf"
    loader_conf_dirs="/boot/loader.conf.d"
    local_loader_conf_files="/boot/loader.conf.local"

You would get the following sequencing (assuming no config file modify
`loader_conf_files`):

1. Bootstrap:

       /boot/defaults/loader.conf

2. Read `loader_conf_files` files:

       /boot/device.hints
       /boot/loader.conf

3. Read `loader_conf_dirs` files (with per-product files coming last but honoring the order used in `product_vars`):

       /boot/loader.conf.d/*.conf
       /boot/loader.conf.d/Dell Inc./*.conf
       /boot/loader.conf.d/0H28RR/*.conf

4. Read `local_loader_conf_files` files:

       /boot/loader.conf.local

bsdimp:
That sounds like a good sequence. It's what I'd hoped you'd say without realizing it in my first question. Is that's what's implemented here, or are you going to implement that and force-push to update this?

stephane-rochoy-stormshield:
>  Is that's what's implemented here, or are you going to implement that and force-push to update this?

With, [6931996](https://github.com/freebsd/freebsd-src/pull/759/commits/69319966d22d852c395bfb07b8fc5e6b934dbfdf), it's what's implemented. But I would really appreciate a second look (or even a third by @kevans91 if possible) ;)

stand/defaults/loader.conf.5
```diff
@@ -138,6 +138,35 @@ present file.
...
+.Pp
+When product information can be guessed, for each product information found,
+append
+.Pa /boot/loader.conf.d/PRODUCT
+to
+.Ar loader_conf_dirs .
+It can be typically used as follow:
+.Pp
+.Dl product_vars="smbios.system.planar.maker smbios.system.planar.product smbios.system.product uboot.m_product"
+
```
kevans91, line 156: Not 100% sure on the markup here; is `mandoc -Tlint` OK with this?

kevans91 (approved):
This looks fine to me! Not 100% sure about extra blank lines in the manpage outside of a literal block, but if mandoc's linter is OK with it then I think it's fine...

bsdimp (commented):
These got stuck

stand/lua/config.lua
```diff
@@ -657,12 +677,40 @@ function config.readConf(file, loaded_files)
...
+
+			local product_conf_dirs = ""
+			for var in product_vars:gmatch("%S+") do
+				local product = getEnv(var)
+				if product ~= nil then
+					product_conf_dirs = product_conf_dirs .. " /boot/loader.conf.d/" .. product
+				end
+			end
+			print("product_conf_dirs: " .. product_conf_dirs) -- XXX
+
```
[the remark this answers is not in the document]
stephane-rochoy-stormshield: I don't understand, I don't see this this call to `print`  in recent commits, (e.g., 4671394b6522) oO

stand/defaults/loader.conf.5
```diff
@@ -140,6 +140,32 @@ present file.
...
+It can be typically used as follow:
+.Pp
+.Dl product_vars="smbios.system.planar.maker smbios.system.planar.product smbios.system.product"
+
+to read files found in the following directories, in that order:
+
+.Bl -bullet -compact
+.It
+.Pa /boot/loader.conf.d/PLANAR_MAKER
```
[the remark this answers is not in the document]
stephane-rochoy-stormshield: Combining `.It` and `.Pa` don't seems be OK. `.It Pa /boot/loader.conf.d/PLANAR_MAKE` trigger the following error:

    mandoc: loader.conf.5:174:2: WARNING: empty list item: Bl -bullet It
    mandoc: loader.conf.5:174:2: ERROR: skipping all arguments: It Pa

bsdimp:
OK. I like the design. I like where we're at. I have one concern, though.

This commit splits loader_conf_files into loader_conf_files and local_loader_conf_files. Fair enough. It moves loader.conf.local to this variable. Also seems fair.

This is a user visible change, so that commit needs a better commit message. It needs to be explicit that it is doing
these things. It also needs an UPDATING entry. I was going to write these last two things when I saw the problem.

If I have a loader_conf_files in my loader.conf that copies the old value from defaults and adds something:
`loader_conf_files="/boot/device.hints /boot/loader.conf /boot/loader.conf.local /boot/netflix-magic.conf"`
and there's a `/boot/loader.conf.local`, then won't it run twice with these patches? And possibly undo `netflix-magic.conf`? Or will this fail 'safe'ish and process it twice, but the second time is a nop because of the interlock that only allows config.readConf to read a file once?

And also, I think we need to document this change in config.lua.8, which currently says:
```
         config.readConf(file, loaded_files)
                                             Process file as a configuration
                                             file (e.g., as loader.conf) and
                                             then processing files listed in
                                             loader_conf_files variable (see
                                             loader.conf(5)).  The caller may
                                             optionally pass in a table as the
                                             loaded_files argument, which uses
                                             filenames as keys and any non-nil
                                             value to indicate that the file
                                             named by the key has already been
                                             loaded and should not be loaded
                                             again.
```
to add a mention of `local_loader_conf_files`?

So three requested changes: update config.lua.8, update the commit message of the second commit and write an UPDATING entry.

Finally, not an issue, per se, for this bug, but loader.conf.5 is shared between lua and forth. Do we add a bugs at the bottom saying some items might not be available for forth? Or ????

Sorry to take so long on this feedback.

kevans91:
> If I have a loader_conf_files in my loader.conf that copies the old value from defaults and adds something: `loader_conf_files="/boot/device.hints /boot/loader.conf /boot/loader.conf.local /boot/netflix-magic.conf"` and there's a `/boot/loader.conf.local`, then won't it run twice with these patches? And possibly undo `netflix-magic.conf`? Or will this fail 'safe'ish and process it twice, but the second time is a nop because of the interlock that only allows config.readConf to read a file once?
> 

The latter, the `loaded_files` that we pass around tracks every file we've been asked to process since we started at the root in /boot/defaults/loader.conf. We create the table at the very top call and it's effectively pass-by-reference, so no matter the hierarchy it's consistent with what we've actually read all the way through to the end.

Agree on the rest... we should generally note specifically what's not available in 4thloader.

stephane-rochoy-stormshield:
I tweaked the manpages (config.lua(8) but also loader.conf(5)), added an entry to `UPDATING` and reworked the commit messages.

Regarding the divergence with 4thloader, how should it be documented? I was unable to find some example in loader.conf(5).

bsdimp:
I think this is ready to go. @kevans91 do you agree?

kevans91:
Yes, I'm happy

bsdimp:
d3d0b735571d
Due to conflicts in dates, I forgot to run the tagging script, so these landed w/o the usual Pull Request tags.

PR #763, opened 2023-06-05
hier(7): improvement, modernisation

grahamperrin:
hier(7): improvement, modernisation

Consistent use of lowercase, spacing between sections, etc.

Cease mentioning floppy disks.

De-list /usr/share/misc/fonts/, which has been ??? (without a description) for twenty-seven years.

Change zpool to pool.

Uppercase PPP for Point-to-Point Protocol.

A few other improvements to wording, including avoidance of the phrase pre-fab.

Update the descriptions of: 

* /tmp/
* /usr/share/misc/
* /var/preserve/
* /var/tmp/
* /var/tmp/vi.recover/.

Refer to vi(1) instead of ex(1).

https://bugs.freebsd.org/261349

PR:                      261349
Pull request:            https://github.com/freebsd/freebsd-src/pull/763

----

The 30th May edition, rendered: <https://reviews.freebsd.org/paste/raw/572/>

### DIFF
--- a/share/man/man7/hier.7
+++ b/share/man/man7/hier.7
@@ -182,10 +182,15 @@ statically-linked programs for emergency recovery; see
 .It Pa /root/
 home directory of the root user
 .It Pa /sbin/
-system programs and administration utilities
-fundamental to both single-user and multi-user environments
+system programs and administration utilities that are fundamental to
+single-user and multi-user modes
 .It Pa /tmp/
-temporary files that are not guaranteed to persist across system reboots
+temporary files that may be removed by
+.Xr rc 8 ;
+see the
+.It Va clear_tmp_enable
+variable of
+.Xr rc.conf 5
 .It Pa /usr/
 contains the majority of user utilities and applications
 .Pp
@@ -590,7 +595,8 @@ line printer spooling directories
 .El
 .Pp
 .It Pa tmp/
-temporary files that are not removed during boot
+temporary files that are not removed by
+.Xr rc 8
 .Pp
 .Bl -tag -width "vi.recover/" -compact
 .It Pa vi.recover/
--- a/share/man/man7/hier.7
+++ b/share/man/man7/hier.7
@@ -414,7 +414,7 @@ console screen maps
 .Pp
 .It Pa sysroot/
 files necessary for the -sysroot compiler/linker argument to build non-native
-binaries.
+binaries
 .Pp
 .Bl -tag -width Fl -compact
 .It Pa VERSION/
--- a/share/man/man7/hier.7
+++ b/share/man/man7/hier.7
@@ -28,7 +28,7 @@
 .\"	@(#)hier.7	8.1 (Berkeley) 6/5/93
 .\" $FreeBSD$
 .\"
-.Dd June 7, 2023
+.Dd June 10, 2023
 .Dt HIER 7
 .Os
 .Sh NAME
--- a/share/man/man7/hier.7
+++ b/share/man/man7/hier.7
@@ -72,7 +72,7 @@ and
 .Xr dtc 1
 .It Pa zfs/
 .Xr zfs 8
-zpool cache files
+pool cache files
 .El
 .It Pa /compat/
 normally a link to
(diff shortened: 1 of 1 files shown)


bsdimp:
Am I blind? I only see capitalizing in this PR

grahamperrin:
You're not blind :-) it's a draft …

share/man/man7/hier.7
```diff
@@ -581,9 +581,7 @@ system messages database;
 see
 .Xr msgs 1
 .It Pa preserve/
-temporary home of files preserved after an accidental death
-of an editor;
-see
+temporary storage for files following a crash of an editor; see
 .Xr vi 1
```
grahamperrin, line 550: [vi(1)](https://man.freebsd.org/cgi/man.cgi?query=vi&sektion=1&manpath=freebsd-release) makes no mention of _preserve_. 

Let's have a truer description for `/var/preserve/`.

grahamperrin: … how about: 

> reserved for historical reasons

----

For what it's worth, in a Linux context, <https://refspecs.linuxfoundation.org/FHS_2.3/fhs-2.3.html#VARLIBLTEDITORGTEDITORBACKUPFILESAN> includes this rationale under `/var/lib/<editor>`: 

> Previous Linux releases, as well as all commercial vendors, use `/var/preserve` for vi or its clones. However, each editor uses its own format for these crash-recovery files, so a separate directory is needed for each editor.
> 
> Editor-specific lock files are usually quite different from the device or resource lock files that are stored in `/var/lock` and, hence, are stored under `/var/lib`.

grahamperrin:
For convenience, a rendered view of the changed page: <https://reviews.freebsd.org/paste/raw/575/>

mhorne (commented):
LGTM, thanks for your work.

I checked the rendered version as well. To summarize the spacing rules for the document, we insert a blank line:
1. After every top-level directory entry
2. Before and after every next-level directory list

... is that right?

share/man/man7/hier.7
```diff
@@ -45,34 +45,34 @@ user utilities fundamental to both single-user and multi-user environments
...
 .It Pa overlays/
-Compiled flattened device tree (FDT) overlays; see
+compiled flattened device tree (FDT) overlays; see
 .Xr fdt 4
 and
 .Xr dtc 1
 .It Pa zfs/
 .Xr zfs 8
-zpool cache files
+pool cache files
```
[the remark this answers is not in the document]
grahamperrin:
```text
% file /boot/zfs/zpool.cache 
/boot/zfs/zpool.cache: , name offset 0x636f7070
% 
```

The cache file relates: 

* to a _pool_, or _pools_, of data
* not to the [zpool(8)](https://openzfs.github.io/openzfs-docs/man/8/zpool.8.html) command.

grahamperrin:
> … a blank line:
> 
> * After every top-level directory entry
> 
> * Before and after every next-level directory list

That's a fair summary. Never more than one blank line. 

----

In a nutshell: where I found the spacing used in parts (not all) of the current edition, I did find it much easier to read. So I used the same approach to spacing throughout the page, in its entirety. 

Maybe easiest to see the difference with these two eighty-column windows. The window to the right is what's drafted, with extra spacing: 

image

----

Now, I see that word _pre-fab_. Not ideal. I'll take this PR back to draft for a few more tweaks.

mhorne (approved):

grahamperrin:
@mhorne thanks, there remains a question above https://github.com/freebsd/freebsd-src/pull/763#pullrequestreview-1463104984 re: the description for `/var/preserve/`, which currently misdirects to the manual page for vi(1).

share/man/man7/hier.7
mhorne: Some small note about how it is unused but historical would be adequate IMO.

Probably we could go further and purge it from the src tree, including the (useless?) clean-preserve `periodic(8)` script. I don't think there would be any harm done by losing this historical information either.

grahamperrin: Thanks; 

* [`5ba7c58` (#763)](https://github.com/freebsd/freebsd-src/pull/763/commits/5ba7c585d11a38083de155d0495077c3e387bce6)
* FreeBSD bug [271933 – Consider removal of /usr/share/misc/fonts/, /var/preserve/ and the clean-preserve periodic(8) script](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=271933).

PR #764, opened 2023-06-05
man(1): Support custom sections

mohd-akram:
PR: [271830](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=271830)


bsdimp:
I like the concept, and I think the execution is decent. Can anybody think of a reason not to do this? Or if there's documentation that needs to be updated for the change?

bsdimp:
OK. Went ahead and pushed. I don't think there's any docs that require changing (the man manpage doesn't need it).

PR #765, opened 2023-06-05
sendit: check msg_controllen before returning EINVAL

ngortheone:
Avoid retuning an undocumented EINVAL on sendmsg() when:
```
    msg_control != NULL and msg_controllen == 0
```
Zero length should indicate to the kernel that there is no control data to transmit, same as NULL pointer to msg_control buffer.

### DIFF
--- a/sys/kern/uipc_syscalls.c
+++ b/sys/kern/uipc_syscalls.c
@@ -613,7 +613,7 @@ kern_socketpair(struct thread *td, int domain, int type, int protocol,
 		struct unpcb *unp, *unp2;
 		unp = sotounpcb(so1);
 		unp2 = sotounpcb(so2);
-		/* 
+		/*
 		 * No need to lock the unps, because the sockets are brand-new.
 		 * No other threads can be using them yet
 		 */
@@ -685,7 +685,7 @@ sendit(struct thread *td, int s, struct msghdr *mp, int flags)
 		to = NULL;
 	}
 
-	if (mp->msg_control) {
+	if (mp->msg_control != NULL && mp->msg_controllen > 0) {
 		if (mp->msg_controllen < sizeof(struct cmsghdr)
 #ifdef COMPAT_OLDSOCK
 		    && (mp->msg_flags != MSG_COMPAT ||


kevans91:
CC @markjdb

ngortheone:
just for reference, Linux seems to be doing the same thing (more or less)
https://github.com/torvalds/linux/blob/master/net/socket.c#L2472

markjdb:
I don't have any objection to the change. One downside is that the kernel will no longer tell you you made a mistake if you forget to set `msg_controllen`. There are applications which rely on this behaviour from Linux, so having compatibility might outweigh this downside.

Regarding the patch, freebsd32_sendmsg() should be updated as well. The comment near the beginning of linux_sendmsg_common() should also be updated.

igalic:
> Also, I think documenting EINVAL is also worth doing, but in a separate commit.

no harm doing that in the same PR tho

kevans91:
> I don't have any objection to the change. One downside is that the kernel will no longer tell you you made a mistake if you forget to set `msg_controllen`. There are applications which rely on this behaviour from Linux, so having compatibility might outweigh this downside.
>

I'd like to see the results of, say, a console message output every time we see a `msg_controllen == 0 && msg_control != NULL` while running some applications for a while that are well-known for their relatively heavy cmsg usage and are also largely built/tested on Linux. e.g., IIRC, Firefox/Chromium. Mostly curious to see if maybe we're seeing this today in some common cases but the result is a manifestation that we aren't immediately recognizing as a bug, or haven't tracked it down as such.

kevans91:
That is to say, I don't see any reason to reject being compatible off-hand, but it'd be nice to gather a little more information since we're not in any rush to land it.

markjdb:
> > I don't have any objection to the change. One downside is that the kernel will no longer tell you you made a mistake if you forget to set `msg_controllen`. There are applications which rely on this behaviour from Linux, so having compatibility might outweigh this downside.
> 
> I'd like to see the results of, say, a console message output every time we see a `msg_controllen == 0 && msg_control != NULL` while running some applications for a while that are well-known for their relatively heavy cmsg usage and are also largely built/tested on Linux. e.g., IIRC, Firefox/Chromium. Mostly curious to see if maybe we're seeing this today in some common cases but the result is a manifestation that we aren't immediately recognizing as a bug, or haven't tracked it down as such.

Something like this should tell you: `dtrace -n 'fbt::sendit:entry /args[2]->msg_controllen == 0 && args[2]->msg_control != NULL/{printf("%s", execname);}'`

kevans91:
Ah, good shout- easy enough to run for a bit.

Doing a quick survey of https://grep.app, I'm a couple of pages into the search `msg_controllen`, and it seems like this should be exceedingly rare for an application to rely on -- 0-assignments are nearly always paired with a NULL `msg_control`, there's at least one common use in openssl that sets it to 0 initially but it's added to before the final `sendmsg(2)` call. The vast majority of references are either assigning it `sizeof(...)` or `CMSG_SPACE(...)`.

There is an interesting one in the [linux-test-project](https://github.com/linux-test-project/ltp/blob/master/utils/sctp/lib/sendmsg.c#L87); it sets `msg_controllen` to 0 and seemingly leaves `msg_control` to be stack garbage since the kernel won't be attempting to dereference it with `msg_controllen` properly set, but that seems to also be quite rare.

kevans91:
> > Something like this should tell you: dtrace -n 'fbt::sendit:entry /args[2]->msg_controllen == 0 && args[2]->msg_control != NULL/{printf("%s", execname);}'
> 
> I doubt anything in base does it today, because it would trigger EINVAL (and I fail to think of a good reason to do this intentionally or to rely on this for something)

As noted, I'm much less interested in base and much more interested in what happens in ports.

kevans91:
Data point, though not a very interesting one: I ran the above dtrace for 3 days with pretty heavy chromium/thunderbird usage and came up with exactly 0 hits. I whipped up a quick test program to confirm that it was working as expected, and indeed it it was. I'm really not sure based on grep.app inspection and just running some practical applications that this is a behavior that any real applications actually depend on, and if we change it now going in the reverse direction, if we decide it was a mistake, is a lot harder to do.

bsdimp:
So is this patch good, bad or ugly? Commentary to date suggests that this is an open question and that maybe it isn't best as a pull request. @kevans91 @markjdb : what do you think?

markjdb:
> So is this patch good, bad or ugly? Commentary to date suggests that this is an open question and that maybe it isn't best as a pull request. @kevans91 @markjdb : what do you think?

The patch itself is incomplete and shouldn't be committed as it is. See my comments earlier; freebsd32 and linux compat wrappers need updating. It would also be lovely to have a small test case for this, perhaps in tests/sys/kern/unix_dgram.c. :)

I don't feel too strongly about the change one way or the other. It probably won't break anything. It would be nice to see a specific case where it helps something though, I don't see any examples.

bsdimp:
I'm going to close this, without prejudice. It's not yet finished, and should be re-opened when it's done and the reservations expressed are addressed in some way...

PR #766, opened 2023-06-05
<bsd.progs.mk>: Allow using SRCS for common sources

VexedUXR:
Instead of having to do:
```
PROGS=	program1 program2
SRCS.program1=	program1.c common.c
SRCS.program2=	program2.c common.c
```
This can now be done instead:
```
PROGS=	program1 program2
SRCS=	common.c
SRCS.program1=	program1.c
SRCS.program2=	program2.c
```
This shouldn't effect any existing projects using <bsd.progs.mk>.

Signed-off-by: Ahmad Khalifa [ahmadkhalifa570@gmail.com](mailto:ahmadkhalifa570@gmail.com)

Edit: Add Signed-off-by


igalic:
cc @sgerraty

bsdimp:
I'm unsure.  This is likely good, but we've also discouraged .progs use. I'm learning towards approving.  It looks good, but this is well into make magic and I'm unsure what it might break. I think nothing, but om unsure

sgerraty:
I've not looked at what has been done to bsd.progs.mk recently, but contrib/bmake/mk/progs.mk already allows what the description above appears to want.  So given this change would appear to bring it back closer to the original, don't imagine it would break anything, but you'd have to ask the folk who made the changes introducing PROG_OVERRIDE_VARS

sgerraty:
After looking at some of the history of bsd.progs.mk I would note that the diffs would be much easier to follow if the variable lists used the one per line format

VexedUXR:
Alright, so the only edge case I found was that something like this is valid only when SRCS is in `PROG_OVERRIDE_VARS`
```
PROGS=	program1 program2
SRCS=	common.c
SRCS.program1=	program1.c common.c
SRCS.program2=	program2.c common.c
```
With these changes this obviously produces a duplicate symbol error, however I dont see why someone using progs.mk would do this.

> the diffs would be much easier to follow if the variable lists used the one per line format

As for this, should I commit this in this PR?

sgerraty:
On Thu, 29 Jun 2023 11:44:21 -0700, VexedUXR writes:
>Alright, so the only edge case I found was that something like this is valid o
>nly when SRCS is in `PROG_OVERRIDE_VARS`
>```
>PROGS=	program1 program2
>SRCS=	common.c
>SRCS.program1=	program1.c common.c
>SRCS.program2=	program2.c common.c
>```
>With these changes this obviously produces a duplicate symbol error, however I
> dont see why someone using progs.mk would do this.

Agreed.

>> the diffs would be much easier to follow if the variable lists used the one 
>per line format
>
>As for this, should I commit this in this PR?

It is perhaps a broader topic, and best done as a standalone commit

bsdimp:
@sgerraty you ok with this? I kinda think you are (but have suggestions for follow on work), but wanted to make sure before I push it.

sgerraty:
> @sgerraty you ok with this? I kinda think you are (but have suggestions for follow on work), but wanted to make sure before I push it.

Yes I'm ok with it, and yes my other comments were for future work

bsdimp:
Great! Pushed in. Thanks for the submission and reviews everybody!

bdrewery:
FYI https://reviews.freebsd.org/D55711

PR #767, opened 2023-06-06
od(1): fix skip error handling

mohd-akram:
PR: [271832](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=271832)

The current check doesn't seem to make sense. The file size is subtracted from skip in doskip(), so if it's nonzero and we reached the end, throw an error.


mohd-akram:
This was fixed in https://github.com/freebsd/freebsd-src/commit/1232c7884d4a9bc69d4dd71cd293f92bb1ae6197.

PR #769, opened 2023-06-07
ntp: Fixed undefined behavior of the shift operator

d06alexandrov:
L_LINT macro is used with negative numbers (e.g `L_LINT(time_freq, -MAXFREQ)`) so it could cause undefined behavior.
So it should be similar to the L_RSHIFT(v, n) macro.


emaste:
CC @cschuber 

(gcc12 build failure in Cirrus-CI is due to PR271903 and can be ignored)

cschuber:
This has been committed in af9ce4e9bb7d.

PR #760, opened 2023-06-01
Update OpenSSL to version 3.0.9 in the base system

khorben:
This branch is the closest I am currently to a functional update to OpenSSL 3.0.9 in FreeBSD's base system. It was started from an update to the `vendor/openssl-3.0` branch. (see https://reviews.freebsd.org/D40365)

The thorough review expected should include:
* The configuration applied (default from the corresponding `security/openssl30` port with the legacy provider enabled)
* Security patches up to date (the 3.0.9 version was released on May 30th, 2023)
* Other patches in the corresponding port may be relevant too (I have not included them)
* API changes reflected accurately in `secure/lib/libcrypto/Version.map` and `secure/lib/libssl/Version.map`
* Whether the choice of `30` as `SHLIB_MAJOR` is good for the `.so` files (upstream's `3` has already been obsoleted in FreeBSD's base system)
* Functional tests on Tier-1 architectures (amd64, aarch64...)
* Relevant and functional `ossl-modules` providers
* Patches to software using OpenSSL are correct (currently libarchive, dumpon, and Kerberos)
* This does not touch the assembly OpenSSL files that were moved into `sys/crypto/openssl` to avoid any trouble with the kernel, and re-imports them into `secure/lib/libcrypto/arch` instead as per `Makefile.asm`; security fixes beware of both locations.

In most software users of OpenSSL, a compatibility compilation flag was used in order to expose and use the former OpenSSL 1.1 API, which is still provided by OpenSSL 3.0 on request.

PR: 271615
Sponsored by: The FreeBSD Foundation

PR #770, opened 2023-06-07
khorben/openssl 3.0.9 rebased

khorben:
This is a candidate to replace #760, rebased on a newer commit from `main`.


emaste:
b077aed33b7b6aefca7b17ddb250cf521f938613

PR #772, opened 2023-06-07
mount_nullfs(8): refer to nullfs(5)

tomhukins:


sbin/mount_nullfs/mount_nullfs.8
```diff
@@ -48,6 +48,7 @@
 The
 .Nm
 utility creates a
+.Xr nullfs 5
 null layer, duplicating a sub-tree of the file system
```
igalic, line 52: i would remove the following null:
```suggestion
layer, duplicating a sub-tree of the file system
```

igalic (changes requested):
can you add the xref to the see also section?

sbin/mount_nullfs/mount_nullfs.8
tomhukins: Good idea.  I approve of removing redundant words.  I've just pushed a modified version of my patch that includes your suggestion.

tomhukins:
> can you add the xref to the see also section?

For long manual pages, I can see the advantage of listing all related documentation together, but for a short document like this it seems unnecessary.  I don't feel strongly about this, however, and the [Documentation Primer](https://docs.freebsd.org/en/books/fdp-primer/manual-pages/) doesn't seem to state a preference for either approach.

debdrup:
Well, mdoc(7) covers what should be in the SEE ALSO section - and based on that, I'd say it should go in.

In addition the documentation primer isn't really meant to have every single rule, and I think it's important to consider whether someone looking at the documentation would find it useful to have nullfs(5) mentioned in the SEE ALSO section of the mount_nullfs(8) manual page - which I definitely would argue is the case.

So please go ahead and add it.
While you're at it, make sure to also test with `textproc/igor` and `mandoc -Tlint`.

That way I can commit and push it sooner. :)

tomhukins:
> Well, mdoc(7) covers what should be in the SEE ALSO section

Good point, thank you.

> and based on that, I'd say it should go in.

I've reread the _SEE ALSO_ section of mdoc(7) several times.  I'm a native English speaker and I see nothing suggesting that everything mentioned elsewhere in each manual page should be mentioned yet again here.  In practice, I don't see manual pages following that principle either.

> it's important to consider whether someone looking at the documentation would find it useful to have nullfs(5) mentioned in the SEE ALSO section of the mount_nullfs(8) manual page - which I definitely would argue is the case.

Why would you argue this?  My original patch ensured it was mentioned right at the start of the _DESCRIPTION_ section inlined with the relevant text.  What does mentioning it a second time accomplish?

> So please go ahead and add it. While you're at it, make sure to also test with `textproc/igor` and `mandoc -Tlint`.

I have run both of these commands against every version of the patch submitted in this pull request and seen no errors.  Did I miss something?

> That way I can commit and push it sooner. :)

I still think you're making the wrong decision, but I've amended my patch because referring to nullfs(5) twice, as you want, seems better than not referring to it at all.

PR #773, opened 2023-06-08
uipc_syscalls: removed unnecessary check in accept1() function

d06alexandrov:
Lines 289 and 298 already contain checks of uname and error values.
So the expression `(error == 0 && uname != NULL)` is always true.

### DIFF
--- a/sys/kern/uipc_syscalls.c
+++ b/sys/kern/uipc_syscalls.c
@@ -298,15 +298,13 @@ accept1(struct thread *td, int s, struct sockaddr *uname, socklen_t *anamelen,
 	if (error != 0)
 		return (error);
 
-	if (error == 0 && uname != NULL) {
 #ifdef COMPAT_OLDSOCK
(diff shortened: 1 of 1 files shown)


bsdimp:
Looks good to my eye. Approving CI run.

emaste:
LGTM.

Out of idle curiosity, what prompted you to look at this function?

emaste:
gcc12 build failure in Cirrus-CI is due to PR271903 and can be ignored

d06alexandrov:
@emaste just trying to fix some static analyzer's warnings in another FreeBSD-based project)

bsdimp:
`git diff -w HEAD^` confirms that just the if was removed.
checkstyle9.pl passes (but for something like this, existing style would trump any failure here)
GENERIC kernel on amd64 builds.
I agree with the static analysis tool's flagging this as an unnecessary test.
6016aedba1e3

PR #775, opened 2023-06-09
Add boot1.efi to CLEANFILES

VexedUXR:


bsdimp:
Seems ok. Why doesn't loader.efi and friends need this too?

VexedUXR:
loader.efi already has it. 
Just checked gptboot.efi tough, it dosent. Should I add a commit for gptboot too?

bsdimp:
OK. That all makes sense to me, and checks out.
a8c1c0b727ef

PR #776, opened 2023-06-11
fix qcom_cpu_kpssv2 compile on ASUS_AC1300

igalic:
follow afdb42987ca82869eeaecf6dc25c2b6fb7b8370e and change the return type
from boolean_t to bool for the definition, too.

PR: 271932
Reported by: Mina Galić
Fixes: afdb42987ca82869eeaecf6dc25c2b6fb7b8370e
Pull Request: https://github.com/freebsd/freebsd-src/pull/776


bsdimp:
This looks good. Need to tweak the commit message a little (project custom is to not use reported by when you also are fixing the bug) and I reworded the commit message to use only 12 of the characters of the hash and to make it flow a little better. 3ed646d89aa4

PR #777, opened 2023-06-14
loader.efi: Remove redundant error message

VexedUXR:
`efi_copy_init` already prints an error message (with more information) if it fails.


bsdimp:
This compiles
This looks like it would work: all paths out of efi_copy_init() print errors (though there's two calls to it that might fail w/o noticing, which is a different bug).
This passes checkstyle9.pl, as well as my eyeball check.
Normal booting still works.
No man page update is needed.
Approved.
780332f1c1cf

PR #778, opened 2023-06-16
Simplify buffer count calculation in qlnx_os.c

shiftee:


sys/dev/qlnx/qlnxe/qlnx_os.c
```diff
@@ -446,17 +445,12 @@ qlnx_num_tx_compl(qlnx_host_t *ha, struct qlnx_fastpath *fp,
 {
 	u16 hw_bd_cons;
 	u16 ecore_cons_idx;
-	uint16_t diff;
 
 	hw_bd_cons = le16toh(*txq->hw_cons_ptr);
 
 	ecore_cons_idx = ecore_chain_get_cons_idx(&txq->tx_pbl);
-	if (hw_bd_cons < ecore_cons_idx) {
```
bsdimp, line 454: how do we know this is always false?

shiftee: It isn't always false.
The hardware counter will rollover to zero before the software one but subtracting them will always give the correct value.
e.g. 0 - 0xFFFF = 1 (for unsigned 16 bit integer type)

shiftee:
Fixed style errors found by tools/checkstyle9.pl and rebased on latest upstream code

sys/dev/qlnx/qlnxe/qlnx_os.c
bsdimp: That makes sense.

PR #779, opened 2023-06-17
qlnxe: Fix building as in-kernel component

shiftee:
I am not in a position to test this, but the driver now compiles when built as an in-kernel component.

I have also tested that building as a module still works.

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=268354


bsdimp:
Please add to NOTES for all the places the module builds, or at least those that these changes support.
I'll look in more detail when I have more time, but that stood out to me.

shiftee:
I have added it to the amd64 NOTES file.
I also fixed the error which caused the original CI failure with clang, although I was unable to reproduce the error locally.

Ximalas:
I have four Dell PowerEdge R740 with two "QLogic 10GE 2P QL41162HxRJ-DE Adapter" each. Neither server is in prod atm. Note, these NICs are the RJ45 version, and I was able to receive packets at 1GbE speed, but the NICs doesn't seem to transmit at such a slow speed. In a few days I'll have a 10GbE RJ45 switch ready and can play with both speeds, 1Gbe and 10GbE.

bsdimp:
Here's the relevant output of tools/checkstyle9.pl. The ^ points to the wrong place though. There's supposed to be a space at the end of the macro: `);}` -> `); }`
```
1/8 Checking commit 7a05644c39ac (qlnxe: Remove -Wno-case-qual from module build)
ERROR: "foo* bar" should be "foo *bar"
#39: FILE: sys/dev/qlnx/qlnxe/ecore_dbg_fw_funcs.c:1919:
+static const struct dbg_bus_block* get_dbg_bus_block_desc(struct ecore_hwfn *p_hwfn,

ERROR: "foo* bar" should be "foo *bar"
#50: FILE: sys/dev/qlnx/qlnxe/ecore_dbg_fw_funcs.c:1928:
+static const struct dbg_bus_line* get_dbg_bus_line_desc(struct ecore_hwfn *p_hwfn,

5/8 Checking commit d4fa301475bf (qlnxe: Fix unused but set variable warning)
ERROR: space required after that ',' (ctx:VxV)
#25: FILE: sys/dev/qlnx/qlnxe/bcm_osal.h:483:
+#define OSAL_SPIN_LOCK_IRQSAVE(x,y) {(void)y; mtx_lock(x);}
                                 ^

ERROR: space required after that ';' (ctx:VxV)
#25: FILE: sys/dev/qlnx/qlnxe/bcm_osal.h:483:
+#define OSAL_SPIN_LOCK_IRQSAVE(x,y) {(void)y; mtx_lock(x);}
                                                          ^

ERROR: space required after that ',' (ctx:VxV)
#26: FILE: sys/dev/qlnx/qlnxe/bcm_osal.h:484:
+#define OSAL_SPIN_UNLOCK_IRQSAVE(x,y) {(void)y; mtx_unlock(x);}
                                   ^

ERROR: space required after that ';' (ctx:VxV)
#26: FILE: sys/dev/qlnx/qlnxe/bcm_osal.h:484:
+#define OSAL_SPIN_UNLOCK_IRQSAVE(x,y) {(void)y; mtx_unlock(x);}
                                                              ^
```

LINT does build on amd64.

The two CI failures look to be false negatives: They can't find libpam, which this has nothing to do with.

The author name on these commits is valid (and appears to be correct).

So we're getting close.

shiftee:
Fixed style errors found by tools/checkstyle9.pl and rebased on latest upstream code

shiftee:
I have modified some #defines to match those used when building the driver as a module.
There should be no difference between them now.
Hopefully this will resolve the ifconfig hangs seen by @Ximalas

bsdimp:
checking builds now...
If they succeed, I'll commit... It's passed the style check and I've not noticed anything bad.

bsdimp:
restarting... is this ready?

bsdimp:
Looks good locally, I'll push unless the github CI workflows have real issues.

bsdimp:
6f9388d50588 lands this

PR #780, opened 2023-06-19
geli: fix setkey behavior on detached providers

igalic:
wipe out cached_passphrase (using memset(3) for clarity) to allow
geli(8) setkey to correctly function on detached providers.

PR: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=254966
Reported by:  rob2g2
Submitted by:  Arjan de Vet <freebsd@devet.org>
Pull Request: https://github.com/freebsd/freebsd-src/pull/780

### DIFF
--- a/lib/geom/eli/geom_eli.c
+++ b/lib/geom/eli/geom_eli.c
@@ -1363,6 +1363,12 @@ eli_setkey_detached(struct gctl_req *req, const char *prov,
 		return;
 	}
 
+	/*
+	 * The previous eli_genkey() set cached_passphrase, we do not want
+	 * to use that for the new passphrase so always prompt for it
+	 */
+	memset(cached_passphrase, 0, sizeof(cached_passphrase));
+
 	/* Generate key for Master Key decryption. */
 	if (eli_genkey_single(req, md, key, false) == NULL) {
 		explicit_bzero(key, sizeof(key));


emtiu:
For steps to reproduce the bug, see: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=254966#c2

tomhukins:
I don't feel qualified to review the code, but the commit message begins with `geli(8): `, following the convention for commits to manual pages, as opposed to `geli: `, the convention for source code commits.  As this is a code change, not a documentation change, you might want to amend the commit message.

bsdimp:
I think this is right, but I'd feel happier with a security reviewer like @cperciva or @bsdjhb

igalic:
@bsdjhb recommends to ping @pjd and @allanjude

Koston-0xDEADBEEF:
> Do not submit pull requests that are security-related

https://github.com/freebsd/freebsd-src/blob/main/CONTRIBUTING.md

allanjude (approved):
Reviewed-by: allanjude

lib/geom/eli/geom_eli.c
```diff
@@ -1365,6 +1365,12 @@ eli_setkey_detached(struct gctl_req *req, const char *prov,
 		return;
 	}
 
+	/*
+	 * The previous eli_genkey() set cached_passphrase, we do not want
+	 * to use that for the new passphrase so always prompt for it
+	 */
+	explicit_bzero(cached_passphrase, sizeof(cached_passphrase));
```
bsdjhb, line 1372: In this case a plain`memset` is probably ok.  Typically one uses `explicit_bzero` when a secret is no longer needed and you want to prevent it leaking into another context, not to invalidate a cached secret that is intended normally to be reused.  For this case I think `explicit_bzero` might be a bit confusing / misleading as it is not serving it's normal role.

igalic: aye, that's good feedback. will fix

igalic: okay, so, after consulting the rest of the code, there's not a single `memset`, or even `bzero` in all of lib/geom/eli/

bsdjhb: Older BSD code uses `bzero`, but newer code uses `memset` since `memset` is POSIX and `bzero` is not.  (`bzero` in the kernel is just mapped to `memset` via a macro.)

igalic:
@bsdimp, reviews are in and fixes are made as requested. I believe this is ready to merge

markjdb:
This patch doesn't actually fix the problem reported in the bug report. The memset() happens too early, before `cached_passphrase` is initialized. The patch in bugzilla is correct.

I wrote a regression test for this. I'll commit the original patch and test case shortly.

markjdb:
Landed in 88d13bf33754.

PR #781, opened 2023-06-19
netlink: add unregister call when cleanup

f0rm2l1n:
For protocols that use netlink (generic and route for now), the unint handler seems to have forgotten to call unregister, which will cause the assertion the next time the module is loaded.

This patch adds unregister call to netlink_unregister_proto() for those handlers to avoid bad things happen.

Fixes: 7e5bf68495cc ("netlink: add netlink support")

### DIFF
--- a/sys/netlink/netlink_generic.c
+++ b/sys/netlink/netlink_generic.c
@@ -297,6 +297,7 @@ SYSINIT(genl_load_all, SI_SUB_PROTO_DOMAIN, SI_ORDER_THIRD, genl_load_all, NULL)
 static void
 genl_unload(void *u __unused)
 {
+	netlink_unregister_proto(NETLINK_GENERIC);
 	EVENTHANDLER_DEREGISTER(genl_family_event, family_event_tag);
 	genl_unregister_family(CTRL_FAMILY_NAME);
 	NET_EPOCH_WAIT();
--- a/sys/netlink/netlink_route.c
+++ b/sys/netlink/netlink_route.c
@@ -135,6 +135,7 @@ static void
 rtnl_unload(void *u __unused)
 {
 	netlink_callback_p = nlbridge_orig_p;
+	netlink_unregister_proto(NETLINK_ROUTE);
 	rtnl_ifaces_destroy();
 	rtnl_neighs_destroy();
 


AlexanderChernikov:
Thank you for the report! I tested netlink module load/unload on recent -head w/witness etc and it didn’t result in panic. I’ll take a look at the code tomorrow, meanwhile could you provide the panic trace / message you’re observing?

f0rm2l1n:
> Thank you for the report! I tested netlink module load/unload on recent -head w/witness etc and it didn’t result in panic. I’ll take a look at the code tomorrow, meanwhile could you provide the panic trace / message you’re observing?

Thanks for replying. Unfortunately I have not give it a dynamic debugging chance yet. All above are currently static. 😢
Though I think obviously something is wrong as this netlink_unregister_proto() function defined but not used is just too weird. Still my bad not actually trigger this dynamically. I will do the test asap and may resend the report next time when I actually witness the "bad thing". 😃

bsdimp:
waiting on @AlexanderChernikov before landing, asking for changes or rejecting. Tagging with needs-review

AlexanderChernikov (approved):

AlexanderChernikov:
> > Thank you for the report! I tested netlink module load/unload on recent -head w/witness etc and it didn’t result in panic. I’ll take a look at the code tomorrow, meanwhile could you provide the panic trace / message you’re observing?
> 
> Thanks for replying. Unfortunately I have not give it a dynamic debugging chance yet. All above are currently static. 😢 Though I think obviously something is wrong as this netlink_unregister_proto() function defined but not used is just too weird. Still my bad not actually trigger this dynamically. I will do the test asap and may resend the report next time when I actually witness the "bad thing". 😃
Sorry for delaying the review. So Netlink original model was to create protocols per use-case (`NETLINK_NFLOG`, `NETLINK_SCSITRANSPORT ` etc). The associated design issues (dedicated userland socket per protocol, static headers) resulted in this model to be obsoleted by the generic netlink approach, where a single socket can be used to communicate with any "family" inside genetlink. As of now, FreeBSD implementation consists of `NETLINK_ROUTE` and `NETLINK_GENERIC` protocols. It's likely that most new development will go in the `NETLINK_GENERIC` families, effectively keeping the actual _protocol_ to the existing two. As both are embedded in the module, "unregister" happens automatically with the module unload - with the module itself.

Anyway, for consistency it certainly make sense to call `netlink_unregister_proto()` in both cases.

emaste:
> Anyway, for consistency it certainly make sense to call `netlink_unregister_proto()` in both cases.

Should I take that as an `Reviewed by: melifaro` and merge it to main?

igalic:
given the github approved ☑ I would say so, yes

PR #787, opened 2023-06-30
libcrypto: fix the legacy and FIPS providers

khorben:
These commits should help fix issues with various applications using OpenSSL, especially in relation to cryptography algorithms deemed legacy since OpenSSL 3 (e.g., RC4) or when enforcing FIPS compliance.

It is not exactly clear to me why this required the change in `crypto/openssl/crypto/bn/bn_const.c`, however without it the FIPS module fails with:
```
00B0AD22712F0000:error:12800067:DSO support routines:dlfcn_load:could not load the shared library:/usr/src/crypto/openssl/crypto/dso/dso_dlfcn.c:118:filename(/usr/lib/ossl-modules/fips.so): /usr/lib/ossl-modules/fips.so: Undefined symbol "ossl_bignum_modp_1536_p"
```


cschuber:
+ @emaste

madpilot78:
I have installed FreeBSD head with this patch applied and it works fine.

Programs requiring it are now able to load the legacy provider and work flawlessly.

Thanks!

khorben:
Cc @markjdb

khorben:
Cc @ngie-eign

crypto/openssl/crypto/bn/bn_const.c
```diff
@@ -82,10 +82,12 @@ BIGNUM *BN_get_rfc2409_prime_1024(BIGNUM *bn)
  * RFC2312 specifies a generator of 22.
  */
 
+#ifndef FIPS_MODULE
```
markjdb, line 85: Why exactly is the ifdef needed? It's not mentioned in the commit log.

khorben: I mentioned it in the pull request here:

> It is not exactly clear to me why this required the change in `crypto/openssl/crypto/bn/bn_const.c`, however without it the FIPS module fails with:
> ```
> 00B0AD22712F0000:error:12800067:DSO support routines:dlfcn_load:could not load the shared library:/usr/src/crypto/openssl/crypto/dso/dso_dlfcn.c:118:filename(/usr/lib/ossl-modules/fips.so): /usr/lib/ossl-modules/fips.so: Undefined symbol "ossl_bignum_modp_1536_p"
> ```

For some reason it builds fine in the `security/openssl30` port, without this patch. I tried to compare compiler flags, but did not see what could influence that.

markjdb: Sorry I missed that.

The same ifdef wraps `modp_1536_p` in bn_dh.c, so this makes some sense. Still doesn't explain why upstream openssl doesn't have this problem.

Could you please amend the commit log message to explain this a bit further?

khorben: Done! (in bb47d4b4c79a35d4ba477a8979c42e6d72b73dd9)

markjdb:
Merged, thank you.

markjdb:
I ended up reverting this change due to some build breakage on arm64. I suspect that the solution is to move the include paths from libcrypto/Makefile to Makefile.common. I'll run a tinderbox with that change.

markjdb:
Re-pushed with build fixes applied. There is still an issue in that the fips module doesn't load on arm64 due to a missing symbol, but the legacy provider loads.

khorben:
I just confirmed that the FIPS module can be configured to load correctly, with this pull-up request applied, on my local amd64 machine:

* Enabling the FIPS provider in `openssl.cnf` disables the default module, so make sure it has `activate = 1` in its section.
* The default module is required for `openssl fipsinstall`, otherwise no HMAC provider is available to generate the corresponding configuration file. (Defaults to `fips.cnf`)
* The output of `openssl fipsinstall` (the configuration file) needs to be installed in e.g., `/etc/ssl/fipsmodule.cnf` and included by `openssl.cnf` in order for the FIPS provider to work. (Check the provider's section name to be correct and matching that of `fipsmodule.cnf`, e.g., `fips_sect`)
* The configuration file depends on the binary code of the `fips.so` provider module, therefore in order for FreeBSD to ship a working FIPS provider by default, `openssl fipsinstall` (or an equivalent) has to be executed to generate it once all of OpenSSL is done building.

grahamperrin:
Post-closure <https://github.com/freebsd/freebsd-src/pull/787#issuecomment-1622414189>: 

> … the fips module doesn't load on arm64 due to a missing symbol, but the legacy provider loads.

Thanks. 

Will there be a bug report (for tracking), or (for documentation) should we prepare for an erratum at <https://www.freebsd.org/releases/14.0R/errata/>?

markjdb:
> Will there be a bug report (for tracking), or (for documentation) should we prepare for an erratum at https://www.freebsd.org/releases/14.0R/errata/?

I submitted https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=272454

I'm not sure exactly what the problem is there, we should fix it before the release.

PR #788, opened 2023-07-03
CONTRIBUTING.md: minor changes

WuerfelDev:
- Misspelled GitHub
- angle bracket not escaped correctly
- ~Double white spaces~
- White spaces at EOL

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,14 +15,14 @@ FreeBSD accepts source code contributions using one of several methods:
 - Direct access to the [Git repository](https://cgit.freebsd.org/src/)
 
 The preferred method depends on a few factors including the size or scope of
-the change.  GitHub pull requests are preferred for relatively straightforward
+the change. GitHub pull requests are preferred for relatively straightforward
 changes where the contributor already has a GitHub account.
 
 ## GitHub Pull Requests
 
 Presently, GitHub 'freebsd-src' repository is one of the publish-only services
 for the FreeBSD 'src' repository the project uses to promote collaboration and
-contribution.  Pull requests that need little developer time, are generally
+contribution. Pull requests that need little developer time, are generally
 small, and have limited scope should be submitted. Do not submit pull requests
 that are security-related, problem reports, works in progress, changes that are controversial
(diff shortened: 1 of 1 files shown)


jlduran:
These are nice fixes, thank you!
Paging @bsdimp or @emaste.
The only thing I would suggest, is leaving the two spaces after the periods, maybe out of nostalgia, or maybe I could say that the CommonMark specification uses two spaces after periods.
It's up to them to decide.

bsdimp:
Two spaces are the one true way.
Too many don't follow the way.
I can go either way, but I always use two spaces.

WuerfelDev:
I reinserted those spaces. Never used double spaces on purpose before but I guess it helps with readability

igalic:
fwiw, in the docs and on the website, we ask people to put one sentence per line.
This makes it easy to read during review, and easy to diff on change, because changes often happen at a sentence level

bsdimp:
> fwiw, in the docs and on the website, we ask people to put one sentence per line. This makes it easy to read during review, and easy to diff on change, because changes often happen at a sentence level

I thought we didn't do that for markdown documents because it messed up the rendering

jlduran:
> fwiw, in the docs and on the website, we ask people to put one sentence per line.
> This makes it easy to read during review, and easy to diff on change, because changes often happen at a sentence level

Some Markdown (in GitHub's case CommonMark) parsers may interpret the the sentence on a new line as a line break.
For example, this sentence will appear on a new line, i.e., it will insert a `<br>` between the two lines, which is almost visually like a new paragraph, it may not be what I originally intended.  I would have wanted it to look like this sentence.

If an empty line in between is used, then a paragraph is rendered.

bsdimp:
OK. One last issue. The author name is listed as
```
Author: WuerfelDev <WuerfelDev@users.noreply.github.com>
```
which is a nonsense email address. Do you have a real email address you'd like me to use? or I could do `Submitted by: github user WuerfelDev` if you like.

bsdimp:
Thanks! Pushed! 
And at the last second, went ahead with the one space patch. I appreciate you reverting it, but I think that it's likely best to go with that trend.

PR #789, opened 2023-07-03
man(1): use gzcat for .gz files

mohd-akram:
POSIX zcat appends the .Z suffix to file arguments causing the command to fail with .gz files.


markjdb:
I believe this change has no effect on FreeBSD, since our zcat(1) is apparently not compliant and doesn't have the behaviour of appending .Z to paths specified in its arguments. Though, our gzcat(1) does indeed append ".gz" if an attempt to open a file returns ENOENT. Probably it's a bug that it doesn't try ".Z" when invoked as zcat.

markjdb:
Landed in `main`, thanks.

PR #792, opened 2023-07-04
Make memcpy C99 compliant

AtariDreams:
memcpy should have restrict in its prototype definition because that is the C99 standard.

We also must not break the assumptions made by making memcpy restrict.

### DIFF
--- a/lib/libc/string/bcopy.c
+++ b/lib/libc/string/bcopy.c
@@ -56,10 +56,11 @@ typedef	intptr_t word;		/* "word" used for optimal copy speed */
 void *
 #ifdef MEMCOPY
 memcpy
+(void *__restrict dst0, const void *__restrict src0, size_t length)
 #else
 memmove
-#endif
 (void *dst0, const void *src0, size_t length)
+#endif
 #else
 #include <strings.h>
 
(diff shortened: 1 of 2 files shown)


bsdimp:
So again this does too much per commit.
You restructure code gratuitously
The description is insufficient.  What assumptions and why? You can and should fix the bad assumption in one commit then change the prototype in the next. This will help with review and bysecting. Also a good description of what the broken assumption will allow us to make sure your understanding matches ours. Since this is memcpy we have to be extra picky to make sure nothing is broken by these changes.

AtariDreams:
So let's break down each change 1-by-1

dst0 == src0 cannot happen because the whole function would be undefined anyway, so remove the check for that.

Similarly, the check for positioning of the two pointers does not matter if we know they do not overlap, so the entire bottom else statement becomes irrelevant.

Finally, memmove does not have the restrict qualifier, so it should not have those.

brooksdavis:
This change is technically correct, but not worth it. This C memcpy() has been memmove() since the 4.4 lite import (and the amd64 implementation preserves this feature) so it's likely that this change would cause quite a few failures to save one branch in a function mostly called with larger sizes (since it gets elided by the compiler in small cases). There might be some limited code correctness diagnostic value in declaring the function restrict per C99 and *not* using that definition when compiling, but I suspect most compilers that can diagnose this restrict case is going to do it with or without the declarations.

bsdimp:
I'm going to close this. We're not going to change the semantics of memcpy/memmove.

PR #796, opened 2023-07-07
etc/rc.d/routing: use find_system_scripts

brooksdavis:
In 3693d9140e05aba9942232df13468f51a6cde136 /etc/rc switched to using find_system_scripts rather than directly including /etc/rc.d/* in the list of scripts to run in order to skip .pkgsave files.  Follow suit in etc/rc.d/routing.


bsdimp:
This looks good, but the hash isn't right.

brooksdavis:
Oops, must have gone distracted and forgot what was in the paste buffer (that was an unrelated CheriBSD hash). Should be fixed now.

PR #798, opened 2023-07-10
Mmccam vs minimal

igalic:
- switch amd64's GENERIC-MMCCAM to GENERIC
- restore the old GENERIC-MMCCAM as MINIMAL-MMCCAM
- remove CAMDEBUG_FLAGS, while keeping CAMDEBUG itself.

### DIFF
--- a/sys/amd64/conf/GENERIC-MMCCAM
+++ b/sys/amd64/conf/GENERIC-MMCCAM
@@ -12,12 +12,8 @@ ident		GENERIC-MMCCAM
 
 options 	MMCCAM
 
-# Add CAMDEBUG stuff
+# Allow for CAM debugging
 options 	CAMDEBUG
-options 	CAM_DEBUG_FLAGS=(CAM_DEBUG_INFO|CAM_DEBUG_PROBE|CAM_DEBUG_PERIPH)
-
-# pass(4) device
-device		pass
 
 nodevice	mmc
 nodevice	mmcsd
--- a/sys/amd64/conf/GENERIC-MMCCAM
+++ b/sys/amd64/conf/GENERIC-MMCCAM
@@ -1,33 +1,23 @@
-# MMCCAM is the kernel config for doing MMC on CAM development
-# and testing on bhyve
+#
+# GENERIC-MMCCAM
+#
+# Custom kernel for GENERIC plus MMCCAM as opposed to the prior MMC stack.
+#
 
 #NO_UNIVERSE
 
-include         MINIMAL
+include		GENERIC
 
 ident		GENERIC-MMCCAM
-
-# Access GPT-formatted and labeled root volume
-options 	GEOM_LABEL
-
-# UART -- for bhyve console
-device          uart
-
-# VirtIO support, needed for bhyve
-device		virtio			# Generic VirtIO bus (required)
-device		virtio_pci		# VirtIO PCI device
-device		vtnet			# VirtIO Ethernet device
-device		virtio_blk		# VirtIO Block device
-device		virtio_scsi		# VirtIO SCSI device
-device		virtio_balloon		# VirtIO Memory Balloon device
-
-# CAM-specific stuff
-device		pass
-device		scbus
-device		da
 
 options 	MMCCAM
 
 # Add CAMDEBUG stuff
 options 	CAMDEBUG
 options 	CAM_DEBUG_FLAGS=(CAM_DEBUG_INFO|CAM_DEBUG_PROBE|CAM_DEBUG_PERIPH)
+
+# pass(4) device
+device		pass
+
+nodevice	mmc
+nodevice	mmcsd
--- a/sys/amd64/conf/MINIMAL-MMCCAM
+++ b/sys/amd64/conf/MINIMAL-MMCCAM
@@ -29,6 +29,5 @@ device		da
 
 options 	MMCCAM
 
-# Add CAMDEBUG stuff
+# Allow for CAM debugging
 options 	CAMDEBUG
-options 	CAM_DEBUG_FLAGS=(CAM_DEBUG_INFO|CAM_DEBUG_PROBE|CAM_DEBUG_PERIPH)
--- a/sys/amd64/conf/MINIMAL-MMCCAM
+++ b/sys/amd64/conf/MINIMAL-MMCCAM
@@ -1,0 +1,34 @@
+#
+# MINIMAL-MMCCAM
+#
+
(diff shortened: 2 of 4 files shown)


igalic:
a previous attempt at this was made as https://reviews.freebsd.org/D40909

bsdimp:
This looks good to my eye. We need to land mmc cam as the default for 15 thought.

igalic:
> We need to land mmc cam as the default for 15 though

I am not going to submit a "GENERIC-MMC-CLASSIC" patch until branch, and the current PR needs to be merged for 14

(and maybe MFC'd)

sys/amd64/conf/GENERIC-MMCCAM
```diff
@@ -1,34 +1,24 @@
-# MMCCAM is the kernel config for doing MMC on CAM development
-# and testing on bhyve
+#
+# GEMERIC-MMCCAM
```
jlduran, line 2: s/GEMERIC/GENERIC/

igalic: should fix it here https://github.com/freebsd/freebsd-src/blob/5d0826017fc9b84f50ef19b0d452bc0590072f2d/sys/arm/conf/GENERIC-MMCCAM#L2 as well then

igalic: done.

igalic:
@bsdimp ping

bsdimp:
I like this, but have been kinda swamped. Will try to land it in next pull request pass.

igalic:
@bsdimp , now that 14 is in ALPHA, we should get this merged before we branch proper.

igalic:
> > We need to land mmc cam as the default for 15 though
> 
> I am not going to submit a "GENERIC-MMC-CLASSIC" patch until we branch, and the current PR needs to be merged for 14
> 
> (and maybe MFC'd)

@bsdimp now that 14 is branched and this still isn't merged, I might as well do what you suggested.

igalic:
i think this will probably need a RELNOTES entry.

sys/amd64/conf/GENERIC-MMCCAM
```diff
@@ -1,33 +1,23 @@
...
-# CAM-specific stuff
-device		pass
-device		scbus
-device		da
-
 options 	MMCCAM
 
 # Add CAMDEBUG stuff
 options 	CAMDEBUG
 options 	CAM_DEBUG_FLAGS=(CAM_DEBUG_INFO|CAM_DEBUG_PROBE|CAM_DEBUG_PERIPH)
```
bsdimp, line 17: I'd drop the last of these lines. CAMDEBUG is fine.. it make it easy to turn on debugging. CAM_DEBUG_FLAGS make things too chatty by default. Yes, the one that's there should likely have that done to it too.

sys/amd64/conf/GENERIC-MMCCAM
```diff
@@ -1,33 +1,23 @@
...
+
+# pass(4) device
+device		pass
```
bsdimp, line 20: Is pass not in generic?

sys/amd64/conf/MINIMAL
```diff
@@ -36,6 +36,7 @@ makeoptions	WITH_CTF=1		# Run ctfconvert(1) for DTrace support
 options 	SCHED_ULE		# ULE scheduler
 options 	NUMA			# Non-Uniform Memory Architecture support
 options 	PREEMPTION		# Enable kernel thread preemption
+options 	VIMAGE			# Subsystem virtualization, e.g. VNET
```
bsdimp, line 39: this seems unrelated.

sys/amd64/conf/MINIMAL-MMCCAM
```diff
@@ -0,0 +1,33 @@
+# MMCCAM is the kernel config for doing MMC on CAM development
```
bsdimp, line 1: I don't think this is needed. Can you remove it form your submission?

sys/i386/conf/MINIMAL
```diff
@@ -41,6 +41,7 @@ makeoptions	WITH_CTF=1		# Run ctfconvert(1) for DTrace support
 
 options 	SCHED_ULE		# ULE scheduler
 options 	PREEMPTION		# Enable kernel thread preemption
+options 	VIMAGE			# Subsystem virtualization, e.g. VNET
```
bsdimp, line 44: again, unrelated change.

sys/amd64/conf/MINIMAL
igalic: that's why it's a standalone commit

bsdimp:
left some comments. And I'm on the fence if you should turn off CAM_DEBUG_FLAGS with this commit or as a separate commit.

sys/i386/conf/MINIMAL
igalic: do you want me to create a separate pull request for that?

igalic:
thank you for the feedback
I'll try to make something sensible of it tomorrow

sys/amd64/conf/GENERIC-MMCCAM
igalic: removed, in its own commit, so you can decide whether that's a good idea or not as you review this PR

sys/amd64/conf/GENERIC-MMCCAM
igalic: it is, but it's not in MINIMAL, which is what amd64's GENERIC-MMCCAM kernel was based on.

sys/amd64/conf/MINIMAL
igalic: spun out into #962

sys/i386/conf/MINIMAL
igalic: spun out into #962

bsdimp:
I think this is close from my quick glance now. I'll take a more detailed look. Sorry this has lingered.

igalic:
don't worry, I've lost all feeling for time about 5 years ago. I didn't notice this has been open for 8 months, until somebody pointed it out

bsdimp:
OK. Now that I've had time to look at this with a clear head it's mostly good. And I'll land the good parts.
I'm going to drop the MINIMAL-MMCCAM config, though, since I don't think we have a huge need to preserve the old behavior. We can revisit that detail later in the name of getting the rest in easily enough. 834517eff8aa.

PR #799, opened 2023-07-16
Prefer fputc over fputs where possible

AtariDreams:
fputc is meant for single characters, fputs is for strings.

### DIFF
--- a/usr.bin/truss/syscalls.c
+++ b/usr.bin/truss/syscalls.c
@@ -1179,7 +1179,7 @@ print_iovec(FILE *fp, struct trussinfo *trussinfo, uintptr_t arg, int iovcnt)
 		return;
 	}
 
-	fputs("[", fp);
+	fputc('[', fp);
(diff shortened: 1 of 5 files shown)


usr.bin/truss/syscalls.c
```diff
@@ -1518,10 +1518,10 @@ print_cmsgs(FILE *fp, pid_t pid, bool receive, struct msghdr *msghdr)
 			print_gen_cmsg(fp, cmsghdr);
 			break;
 		}
-		fputs("}", fp);
+		fputc('}', fp);
 		first = false;
 	}
-	fputs("}", fp);
+		fputc('}', fp);
```
igalic, line 1524: what's up with the whitespace change here?

igalic (changes requested):
please split this into one commit per Subsystem.

bsdimp:
NAK. This is churn for no good effect. It's not worth the risk.

PR #800, opened 2023-07-20
Adhesive bandage fix for libcrypto with gcc12

jlduran:
Sprinkle `-z noexecstack` into libcrypto's LDFLAGS when compiling with GCC.

---

(To illustrate a point in https://reviews.freebsd.org/D41101)


emaste:
See https://reviews.freebsd.org/D41102

jlduran:
Future-proof, global implementation in: https://reviews.freebsd.org/D41120.

PR #801, opened 2023-07-26
Remove the unnecessary goto parse from the mixer binary

RHL120:
In usr.sbin/mixer/mixer.c:120 there is an if statement that executes a goto to the label `parse` which occurs right after another if statement. As far as I can tell, the first if statement and it's corresponding goto and label can be removed if the negation of what it checks is checked in the if statement that follows. This is what this PR does.

### DIFF
--- a/usr.sbin/mixer/mixer.c
+++ b/usr.sbin/mixer/mixer.c
@@ -117,15 +117,12 @@ main(int argc, char *argv[])
 
 	initctls(m);
 
-	if (dflag && set_dunit(m, dunit) < 0)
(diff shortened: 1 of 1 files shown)


markjdb:
I don't think `if (((!dflag || set_dunit(m, dunit) >= 0) && sflag)) {` is particularly easy to read, so it's not clear why that version is preferable. I don't think it's worth changing the code this way to eliminate a trivial goto.

bsdimp:
NAK. This makes the code harder to follow.

PR #804, opened 2023-07-27
src.conf.5: Regenerate makeman after enabling the building of lib32 on arm64

jlduran:
Regenerate makeman after enabling the building of lib32 on arm64.


emaste:
81a7ce44047f

PR #806, opened 2023-07-28
Typo fix on if_vge.c

priittrees:


igalic:
would be cool to do another run with PVS Studio.
It catches these kinds of things

emaste:
Merged in 0a5d2802b41fd216d9a345f749af1a6ccbe9f382

PR #817, opened 2023-08-09
RNDA-197

Pawel-Oriol-Mobica:
RNDA-197 In addition, the patch fixes (purely by chance) two additional errors :  
implicit declaration of "config_mountroot" (RNDA-199) and "loadfirmware" .
Changed Makefile to display all the errors as well.
Reduntant headers where only commented out to keep with the practice of predecessors (that way it's easier to follow the changes for the time being).


Pawel-Oriol-Mobica:
Accidental push on upstream - apologies.

PR #818, opened 2023-08-11
ls.1 : Mention -v incompatability with -s and -t

concussious:
Github CI check fails on something unrelated. I think actually this needs to be a traditional patch going through bugzilla, so I'm going to close this.

### DIFF
--- a/bin/ls/ls.1
+++ b/bin/ls/ls.1
@@ -460,9 +460,9 @@ options all override each other; the last one specified determines
 the file time used.
 .Pp
 The
-.Fl S
+.Fl S , t
 and
-.Fl t
+.Fl v
 options override each other; the last one specified determines
 the sort order used.
 .Pp
--- a/bin/ls/ls.1
+++ b/bin/ls/ls.1
@@ -928,7 +928,7 @@ command appeared in
 The
 .Fl v
 option was added in
-.Fx 14.0 .
+.Fx 13.2 .
 .Sh BUGS
 To maintain backward compatibility, the relationships between the many
 options are quite complex.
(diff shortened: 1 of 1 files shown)


emaste:
Pull request is fine for this kind of change, but are you sure it's accurate?

in ls.c I see:
```
                /* The -t and -S options override each other. */
                case 'S':
                        f_sizesort = 1;
                        f_timesort = 0;
                        break;
                case 't':
                        f_timesort = 1;
                        f_sizesort = 0;
                        break;
...
                case 'v':
                        f_verssort = 1;
                        break;
```

so it looks like `S` and `t` do indeed override each other, but `v` is handled separately.

I think the logic is actually the first one of these wins:
- `S` if it is the last of any `S` and `t` options (regardless of `v`)
- `v` anywhere (regardless of any `S` and `t`)
- `t` if it is the last of any `S` and `t` options, and there is no `v`

`v` was introduced in e2662256cdbc6ca9be49a66837f9725081cfa9d3

concussious:
> Pull request is fine for this kind of change, but are you sure it's accurate?

I tried to add -v into my workflow and discovered that they couldn't be used together.
Since they can't be used together, I think ls.c should probably then read:

`                /* The -t and -S options override each other. */
                case 'S':
                        f_sizesort = 1;
                        f_timesort = 0;
                        f_verssort = 0;
                        break;
                case 't':
                        f_timesort = 1;
                        f_sizesort = 0;
                        f_verssort = 0;
                        break;
...
                case 'v':
                        f_verssort = 1;
                        f_timesort = 0;
                        f_sizesort = 0;
                        break;
`
Thanks.

PR #820, opened 2023-08-17
arm/allwinner: simple-framebuffer on A20 and Olimex-SOM-EVB support

n-p-soft:
The goal of this driver is to use the framebuffer set up by the bootloader (u-boot); settings are kept but the framebuffer itself is moved to another usable address into memory. I know that a similar work is done in -CURRENT under dev/vt but I think it may need deep changes to run on A20 (probably other) boards. So I chose to create a special device tree binding named "allwinner,sun7i-a20-simplefb" for this driver.

Also added are a kernel configuration file (OLIMEX-SOM-EVB) and a device tree file (olimex-som-evb.dts) needed for the Olimex A20 SOM EVB board to work, as an example. This board is now running with a graphical console.

Tested using releng/13.2, u-boot 2019.07,  ubldr.bin.


sys/arm/allwinner/a20/a20_simplefb.c
```diff
@@ -0,0 +1,327 @@
...
+ * IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ *
+ * $FreeBSD$
```
igalic, line 25: this needs ti go

sys/arm/allwinner/a20/a20_simplefb.c
```diff
@@ -0,0 +1,327 @@
...
+ * used for its framebuffer as 'reserved', we may lose this area when allocating
+ * the new framebuffer. But for now (2023/08), some versions of loader.efi
+ * which show this area as reserved fail to start FreeBSD for another reason.
+ * The best way to start FreeBSD in this case is to use u-boot + ubldr.bin
+ * instead of loader.efi: the u-boot framebuffer will simply be overwritten
+ * by the kernel and our new framebuffer will go on.
+ */
+
+#include <sys/cdefs.h>
+__FBSDID("$FreeBSD$");
```
igalic, line 54: same with these yeti lines

igalic (changes requested):
I don't have much to say except:

sys/arm/allwinner/a20/a20_simplefb.c
n-p-soft: Done, thank you

evadot:
It would be better to fix efifb to work on armv7 if you care about display,.
Also there is already a DTS for this board and we already compile a DTB for it, why creating a new one ?

n-p-soft:
The existing DTB did not work for me (but it works under Linux) due to some changes in the main include coming from Linux (sun7i-a20.dtsi); plus bindings missing, etc.The problem for the framebuffer is that u-boot puts it at the end of the memory. Efi adds a reserved area that overlaps with it. So the quickest way for me was to move the u-boot buffer elsewhere (there are problems about reserved areas, I tried multiple ways but did not manage to keep the buffer) and not use efi at all. Probably the address chosen by u-boot is card-dependent..Le 19 sept. 2023 18:19, Emmanuel Vadot ***@***.***> a écrit :
It would be better to fix efifb to work on armv7 if you care about display,.
Also there is already a DTS for this board and we already compile a DTB for it, why creating a new one ?

evadot:
> The existing DTB did not work for me (but it works under Linux) due to some changes in the main include coming from Linux (sun7i-a20.dtsi); 

 What changes exactly ? It did work for me when I added support for this board a long time ago.

> plus bindings missing, etc.

 Which ones ?

> The problem for the framebuffer is that u-boot puts it at the end of the memory. Efi adds a reserved area that overlaps with it. So the quickest way for me was to move the u-boot buffer elsewhere (there are problems about reserved areas, I tried multiple ways but did not manage to keep the buffer) and not use efi at all.

 EFI is the only really supported way of booting armv7 nowadays for FreeBSD.

> Probably the address chosen by u-boot is card-dependent..Le 19 sept. 2023 18:19, Emmanuel Vadot ***@***.***> a écrit : It would be better to fix efifb to work on armv7 if you care about display,. Also there is already a DTS for this board and we already compile a DTB for it, why creating a new one ? —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>

n-p-soft:
It seems that the Linux kernel changed the names of many pins in the dtsi.Le 20 sept. 2023 10:01, Emmanuel Vadot ***@***.***> a écrit :

The existing DTB did not work for me (but it works under Linux) due to some changes in the main include coming from Linux (sun7i-a20.dtsi);

What changes exactly ? It did work for me when I added support for this board a long time ago.

plus bindings missing, etc.

Which ones ?

The problem for the framebuffer is that u-boot puts it at the end of the memory. Efi adds a reserved area that overlaps with it. So the quickest way for me was to move the u-boot buffer elsewhere (there are problems about reserved areas, I tried multiple ways but did not manage to keep the buffer) and not use efi at all.

EFI is the only really supported way of booting armv7 nowadays for FreeBSD.

Probably the address chosen by u-boot is card-dependent..Le 19 sept. 2023 18:19, Emmanuel Vadot @.> a écrit : It would be better to fix efifb to work on armv7 if you care about display,. Also there is already a DTS for this board and we already compile a DTB for it, why creating a new one ? —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: @.>

bsdimp:
I've tagged this as changes-required because it's not suitable in its current form, though I'm not at all clear which changes we'll need. It might also be invalid (per make it work with efi comments). Either way, it's not ready to go.

bsdimp:
This has been changes requested for a couple months now. I'm going to close it as inactive. If you revisit this, please feel free to reopen this pull request, or to create a new one.

PR #821, opened 2023-08-17
arm/allwinner: clock fixes and EARLY_PRINTF support

n-p-soft:
Code was added to allow EARLY_PRINTF on UART0. Two fixes in the ng-clocks source code (missing variables initialization).


evadot:
We already have early_printf for 32bits Allwinner in https://cgit.freebsd.org/src/tree/sys/dev/uart/uart_dev_snps.c#n55

n-p-soft:
I missed the early debug, but the clocks fixes are needed.Le 19 sept. 2023 18:16, Emmanuel Vadot ***@***.***> a écrit :
We already have early_printf for 32bits Allwinner in https://cgit.freebsd.org/src/tree/sys/dev/uart/uart_dev_snps.c#n55

PR #822, opened 2023-08-18
auto_master.5: improve description, bump date

concussious:
This patch bumps the date and changes the document description of File Formats Manual page auto_master.5 from:

`auto_master – auto_master and map file format`

to:

`auto_master – autofs automounter configuration file format`

This is more descriptive and serves to save a step when looking for auto_master but you forgot or never knew what it's called so you try "automount" or "autofs". This page is linked on the others, but with this change they will all come up together when searched for with either term. Also no keywords have been removed from the manual's name.

### DIFF
--- a/share/man/man5/auto_master.5
+++ b/share/man/man5/auto_master.5
@@ -1,0 +1,373 @@
+.\" Copyright (c) 2014 The FreeBSD Foundation
+.\"
+.\" This software was developed by Edward Tomasz Napierala under sponsorship
+.\" from the FreeBSD Foundation.
+.\"
+.\" Redistribution and use in source and binary forms, with or without
+.\" modification, are permitted provided that the following conditions
+.\" are met:
+.\" 1. Redistributions of source code must retain the above copyright
+.\"    notice, this list of conditions and the following disclaimer.
+.\" 2. Redistributions in binary form must reproduce the above copyright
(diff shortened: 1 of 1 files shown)


share/man/man5/auto_master.5
```diff
@@ -0,0 +1,373 @@
...
+file.
+.It Pa /etc/autofs/
+Directory containing shell scripts to implement special maps and directory
+services.
+.El
+.Sh SEE ALSO
+.Xr autofs 5 ,
+.Xr automount 8 ,
+.Xr automountd 8 ,
+.Xr autounmountd 8
```
igalic, line 367: please add a reference to this man page in the above man pages

igalic (changes requested):
my eyes are too sticky to look at the rest, but here's one comment so far

igalic:
i just realised this patch is supposed to change two lines.

where did that file go?
and when?

share/man/man5/auto_master.5
concussious: Thank you. Currently on 13.2-RELEASE, this page is already referenced in all of the above man pages.

auto_master.5 is not listed in the tree on github for some reason.

igalic:
it looks like auto_master.5 has been in usr.sbin/autofs/auto_master.5 since 2014.

PR #826, opened 2023-08-20
freebsd-update(8): add a BUGS section

grahamperrin:
Add the section.

Describe, but do not refer to, bug 204549.

Whilst here: wrap to 80 columns (80 bytes line length), as suggested by the
'-T lint' output mode of mandoc(7).

https://bugs.freebsd.org/204549
https://bugs.freebsd.org/273155

PR: 204549 273155


grahamperrin:
### Rendered

```text
FREEBSD-UPDATE(8)       FreeBSD System Manager's Manual      FREEBSD-UPDATE(8)

NAME
     freebsd-update – fetch and install binary updates to FreeBSD

SYNOPSIS
     freebsd-update [-F] [-b basedir] [--currently-running release]
                    [-d workdir] [-f conffile] [-j jail] [-k KEY]
                    [--not-running-from-cron] [-r newrelease] [-s server]
                    [-t address] command ...

DESCRIPTION
     The freebsd-update tool is used to fetch, install, and rollback binary
     updates to the FreeBSD base system.

BINARY UPDATES AVAILABILITY
     Binary updates are not available for every single FreeBSD version and
     architecture.

     In general, binary updates are available for ALPHA, BETA, RC, and RELEASE
     versions of FreeBSD, e.g.:
           FreeBSD 13.1-ALPHA3
           FreeBSD 13.1-BETA2
           FreeBSD 13.1-RC1
           FreeBSD 13.1-RELEASE
     They are not available for branches such as PRERELEASE, STABLE, and
     CURRENT, e.g.:
           FreeBSD 13.0-PRERELEASE
           FreeBSD 13.1-STABLE
           FreeBSD 14.0-CURRENT

     In particular, the FreeBSD Security Team only builds updates for releases
     shipped in binary form by the FreeBSD Release Engineering Team.

OPTIONS
     The following options are supported:

     -b basedir     Operate on a system mounted at basedir.  (default: /, or
                    as given in the configuration file.)

     -d workdir     Store working files in workdir.  (default:
                    /var/db/freebsd-update/, or as given in the configuration
                    file.)

     -f conffile    Read configuration options from conffile.  (default:
                    /etc/freebsd-update.conf)

     -F             Force freebsd-update fetch to proceed in the case of an
                    unfinished upgrade.

     -j jail        Operate on the given jail specified by jid or name.  (The
                    version of the installed userland is detected and the
                    --currently-running option is no more required.)

     -k KEY         Trust an RSA key with SHA256 of KEY.  (default: read value
                    from configuration file.)

     -r newrelease  Specify the new release (e.g., 11.2-RELEASE) to which
                    freebsd-update should upgrade (upgrade command only).

     -s server      Fetch files from the specified server or server pool.
                    (default: read value from configuration file.)

     -t address     Mail output of cron command, if any, to address.
                    (default: root, or as given in the configuration file.)

     --not-running-from-cron
                    Force freebsd-update fetch to proceed when there is no
                    controlling tty(4).  This is for use by automated scripts
                    and orchestration tools.  Please do not run freebsd-update
                    fetch from crontab(5) or similar using this flag, see:
                    freebsd-update cron

     --currently-running release
                    Do not detect the currently-running release; instead,
                    assume that the system is running the specified release.
                    This is most likely to be useful when upgrading jails.

COMMANDS
     The command can be any one of the following:

     fetch     Based on the currently installed world and the configuration
               options set, fetch all available binary updates.

     cron      Sleep a random amount of time between 1 and 3600 seconds, then
               download updates as if the fetch command was used.  If updates
               are downloaded, an email will be sent (to root or a different
               address if specified via the -t option or in the configuration
               file).  As the name suggests, this command is designed for
               running from cron(8); the random delay serves to minimize the
               probability that a large number of machines will simultaneously
               attempt to fetch updates.

     upgrade   Fetch files necessary for upgrading to a new release.  Before
               using this command, make sure that you read the announcement
               and release notes for the new release in case there are any
               special steps needed for upgrading.  Note that this command may
               require up to 500 MB of space in workdir depending on which
               components of the FreeBSD base system are installed.

     updatesready
               Check if there are fetched updates ready to install.  Returns
               exit code 2 if there are no updates to install.

     install   Install the most recently fetched updates or upgrade.  Returns
               exit code 2 if there are no updates to install and the fetch
               command wasn't passed as an earlier argument in the same
               invocation.

     rollback  Uninstall the most recently installed updates.

     IDS       Compare the system against a "known good" index of the
               installed release.

     showconfig
               Show configuration options after parsing conffile and command
               line options.

TIPS
     •   If your clock is set to local time, adding the line

               0 3 * * * root /usr/sbin/freebsd-update cron

         to /etc/crontab will check for updates every night.  If your clock is
         set to UTC, please pick a random time other than 3AM, to avoid overly
         imposing an uneven load on the server(s) hosting the updates.

     •   In spite of its name, freebsd-update IDS should not be relied upon as
         an "Intrusion Detection System", since if the system has been
         tampered with it cannot be trusted to operate correctly.  If you
         intend to use this command for intrusion-detection purposes, make
         sure you boot from a secure disk (e.g., a CD).

ENVIRONMENT
     PAGER  The pager program used to present various reports during the
            execution.  (Default: “/usr/bin/less”.)

            PAGER can be set to “cat” when a non-interactive pager is desired.

FILES
     /etc/freebsd-update.conf  Default location of the freebsd-update
                               configuration file.

     /var/db/freebsd-update/   Default location where freebsd-update stores
                               temporary files and downloaded updates.

SEE ALSO
     freebsd-version(1), uname(1), freebsd-update.conf(5), nextboot(8)

AUTHORS
     Colin Percival <cperciva@FreeBSD.org>

BUGS
     In patch level situations – for example, 13.2-RELEASE-p1 up to
     13.2-RELEASE-p2: if any previous modification to a file in /etc/ will
     conflict with an available update, then freebsd-update will make no
     attempt to merge.  Instead: freebsd-update will print a list of affected
     locally-modified files.

FreeBSD 14.0-ALPHA2             August 20, 2023            FreeBSD 14.0-ALPHA2
```

bsdimp:
Apart from the stale date, which I'll fix, I think this is fine and ready to go. The style commits are separated out from the content changes in the commits.

bsdimp:
rebased, folded some changes and reworded commit messages...
0b287d19e4ce

PR #827, opened 2023-08-23
document pmap_kextract/vtophys

igalic:
The main body of this pull request is documenting pmap_kextract(9) function and its vtophys(9) alias.

The need to understand pmap_kextract(9) arose from trying to understand virtio code which I am trying to document, and as such this a small side-step in my cloud-init project, where lately I have been working on a virtio vsock driver.

Sponsored by: The FreeBSD Foundation

### DIFF
--- a/share/man/man9/pmap.9
+++ b/share/man/man9/pmap.9
@@ -23,7 +23,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd August 30, 2016
+.Dd August 24, 2023
 .Dt PMAP 9
 .Os
 .Sh NAME
@@ -97,6 +97,7 @@ operation.
 .Xr pmap_init2 9 ,
 .Xr pmap_is_modified 9 ,
 .Xr pmap_is_prefaultable 9 ,
+.Xr pmap_kextract 9 ,
 .Xr pmap_map 9 ,
 .Xr pmap_mincore 9 ,
 .Xr pmap_object_init_pt 9 ,
--- a/share/man/man9/pmap_kextract.9
+++ b/share/man/man9/pmap_kextract.9
@@ -52,8 +52,7 @@ However, if supplied with an illegitimate value for
 .Fa va
 it will return 0 on arm64, and
 .Xr panic 9
-on armv6 and riscv.
-Note that such platform dependent behaviours are subject to change.
+on arm and riscv.
 .Sh SEE ALSO
 .Xr pmap 9
 .Sh AUTHORS
--- a/share/man/man9/pmap_kextract.9
+++ b/share/man/man9/pmap_kextract.9
@@ -49,12 +49,12 @@ associated with the kernel virtual address
 .Fn pmap_kextract
 generally does not fail.
 However, if supplied with an illegitimate value for
-.Fa va
-it will return 0 on arm64, and
-.Xr panic 9
-on arm and riscv.
+.Fa va ,
+the function may return zero, an invalid non-zero value, or call
+.Xr panic 9 .
 .Sh SEE ALSO
-.Xr pmap 9
+.Xr pmap 9 ,
+.Xr pmap_extract 9
 .Sh AUTHORS
 .An -nosplit
 This manual page was written by
--- a/share/man/man9/pmap_kextract.9
+++ b/share/man/man9/pmap_kextract.9
@@ -12,7 +12,7 @@
 .Sh NAME
 .Nm pmap_kextract ,
 .Nm vtophys
-.Nd extract from kernel page table to a physical address
+.Nd extract a physical address from the kernel page table
 .Sh SYNOPSIS
 .In sys/param.h
 .In vm/vm.h
@@ -28,17 +28,23 @@
 .Sh DESCRIPTION
 The
 .Fn pmap_kextract
-function, and its alias
-.Fn vtophys ,
-map a kernel virtual address to a physical page.
+function retrieves the underlying physical memory address corresponding to the given kernel virtual address
+.Fa va .
+The value of
+.Fa va
+must correlate to an active mapping in the kernel address space.
+.Pp
+.Fn vtophys
+is an alias for
+.Fn pmap_kextract
+and behaves identically.
 .Sh RETURN VALUES
 The
 .Fn pmap_kextract
-function will return the physical page address
-.Ft ( vm_paddr_t )
+function will return the physical address
+.Pq Vt vm_paddr_t
 associated with the kernel virtual address
-.Fa va
-inside the physical map of the address space.
+.Fa va .
 In case of failure,
 .Fn pmap_kextract
 will
--- a/share/man/man9/pmap_kextract.9
+++ b/share/man/man9/pmap_kextract.9
@@ -45,13 +45,15 @@ function will return the physical address
 .Pq Vt vm_paddr_t
 associated with the kernel virtual address
 .Fa va .
-In case of failure,
+.Pp
 .Fn pmap_kextract
-will
+generally does not fail.
+However, if supplied with an illegitimate value for
+.Fa va
+it will return 0 on arm64, and
 .Xr panic 9
-on armv6 and riscv, where most other platforms would
-.Xr KASSERT 9 .
-The function will return 0 on arm64.
+on armv6 and riscv.
+Note that such platform dependent behaviours are subject to change.
 .Sh SEE ALSO
 .Xr pmap 9
 .Sh AUTHORS
--- a/share/man/man9/pmap_kextract.9
+++ b/share/man/man9/pmap_kextract.9
@@ -1,0 +1,58 @@
+.\"
+.\" SPDX-License-Identifier: BSD-2-Clause
+.\"
+.\" Copyright (c) 2023 The FreeBSD Foundation
+.\"
+.\" This manual page was written by Mina Galić <FreeBSD@igalic.co> under
+.\" sponsorship from the FreeBSD Foundation.
+.\"
+.Dd August 24, 2023
+.Dt PMAP_KEXTRACT 9
+.Os
+.Sh NAME
+.Nm pmap_kextract ,
+.Nm vtophys
+.Nd extract from kernel page table to a physical address
+.Sh SYNOPSIS
+.In sys/param.h
+.In vm/vm.h
+.In vm/pmap.h
+.Ft vm_paddr_t
+.Fo pmap_kextract
+.Fa "vm_offset_t va"
+.Fc
+.Ft vm_paddr_t
+.Fo vtophys
+.Fa "vm_offset_t va"
+.Fc
+.Sh DESCRIPTION
+The
+.Fn pmap_kextract
+function, and its alias
+.Fn vtophys ,
+map a kernel virtual address to a physical page.
+.Sh RETURN VALUES
+The
+.Fn pmap_kextract
+function will return the physical page address
+.Ft ( vm_paddr_t )
+associated with the kernel virtual address
+.Fa va
+inside the physical map of the address space.
+In case of failure,
+.Fn pmap_kextract
+will
+.Xr panic 9
+on armv6 and riscv, where most other platforms would
+.Xr KASSERT 9 .
+The function will return 0 on arm64.
+.Sh SEE ALSO
+.Xr pmap 9
+.Sh AUTHORS
+.An -nosplit
+This manual page was written by
+.An Mina Galić Aq Mt FreeBSD@igalic.co ,
+based on the
+.Xr pmap_extract 9
+page written by
+.An Bruce M Simpson Aq Mt bms@spc.org .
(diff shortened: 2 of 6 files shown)


igalic:
cc @pauamma who previously reviewed this on Phabricator, before i broke and abandoned those revisions

share/man/man9/pmap_kextract.9
```diff
@@ -0,0 +1,56 @@
...
+function will return the physical page address associated with the
+kernel virtual address
+.Fa va
+inside the physical map of the address space.
+In case of failure,
+.Fn pmap_kextract
+will return
+.Dv NULL .
+.Sh SEE ALSO
+.Xr mutex 9 ,
```
pauamma, line 47: Why is this needed?

sys/arm/arm/pmap-v6.c
```diff
@@ -1090,6 +1090,8 @@ pmap_kextract(vm_offset_t va)
 	pt1_entry_t pte1;
```
igalic, line 1090: i should probably copy the comment above this function definition to all implementations.

igalic (commented):
another possible improvement.

pauamma (approved):
Other maybe than the mutex reference, this LGTM and seems to match the intent of the other 3 changes. Consistency with actual source code is beyond me.

share/man/man9/pmap_kextract.9
igalic: that looks like a left-over from copying pmap_extract.9

share/man/man9/pmap.9
grahamperrin, line 1: Need to update the `.Dd` at [`87805c5` (#827)](https://github.com/freebsd/freebsd-src/pull/827/commits/87805c5d374ec884d509eda382cd71ec3b2bd5e8#diff-9eeb8f3f0acc15f299de1876624dc36eca4e0778280f6e901828f6a021b0b773R26).

grahamperrin (changes requested):

share/man/man9/pmap_kextract.9
```diff
@@ -0,0 +1,56 @@
...
+.Fa "vm_offset_t va"
+.Fc
+.Ft vm_paddr_t
+.Fo vtophys
+.Fa "vm_offset_t va"
+.Fc
+.Sh DESCRIPTION
+The
+.Fn pmap_kextract
+function and its alias
```
grahamperrin, line 31:
```suggestion
function, and its alias
```

share/man/man9/pmap_kextract.9
```diff
@@ -0,0 +1,56 @@
...
+.Fn vtophys
```
grahamperrin, line 32:
```suggestion
.Fn vtophys ,
```

grahamperrin (changes requested):
Maybe add commas. 

I'm not sure that its necessary to mention the alias here (it's already a `.Nm` under _NAME_).

share/man/man9/pmap_kextract.9
```diff
@@ -0,0 +1,56 @@
...
+map a kernel virtual address to a physical page.
+.Pp
+.Sh RETURN VALUES
```
grahamperrin, line 35:
```suggestion
.Sh RETURN VALUES
```

Resolve part of <https://github.com/freebsd/freebsd-src/pull/827#issuecomment-1690346027>.

----

From <https://man.freebsd.org/cgi/man.cgi?query=mdoc&sektion=7&manpath=freebsd-release#MACRO_REFERENCE>: 

> … Paragraph breaks are not needed before or after Sh or Ss macros …

grahamperrin (changes requested):

share/man/man9/pmap_kextract.9
```diff
@@ -0,0 +1,55 @@
...
+function will return the physical page address associated with the
+kernel virtual address
+.Fa va
+inside the physical map of the address space.
+In case of failure,
+.Fn pmap_kextract
+will return
+.Dv NULL .
+.Sh SEE ALSO
+.Xr mutex 9 ,
```
igalic, line 46:
```suggestion
```

share/man/man9/pmap.9
igalic: changing the date to today feels overly optimistic.

grahamperrin: <https://docs.freebsd.org/en/books/fdp-primer/book/#manual-pages-markup-guidelines> 

> … Trivial changes like spelling or punctuation fixes that do not affect usage …

As the newness of pmap_kextract(9) is significant, so I leant towards treating its `.Xr` in pmap(9) as significant. The cross-reference does not affect usage, however …

… for a person who's accustomed to reading pmap(9) in isolation, a change of date might be a trigger to realise that _something_ significant has changed. 

----

I have minimal experience with manual page review, so I'll leave this for you and other reviewers to decide. 

Thanks

sys/amd64/amd64/pmap.c
```diff
@@ -3933,12 +3933,20 @@ pmap_extract_and_hold(pmap_t pmap, vm_offset_t va, vm_prot_t prot)
...
+ *		Extract from the kernel page table the physical address
+ *		that is mapped by the given virtual address "va".
+ */
 vm_paddr_t
 pmap_kextract(vm_offset_t va)
 {
 	pd_entry_t pde;
 	vm_paddr_t pa;
 
+	pa = 0;
```
jrtc27, line 3948: This is dead, pa is always written to below

share/man/man9/pmap_kextract.9
```diff
@@ -0,0 +1,54 @@
...
+The
+.Fn pmap_kextract
+function will return the physical page address associated with the
+kernel virtual address
+.Fa va
+inside the physical map of the address space.
+In case of failure,
+.Fn pmap_kextract
+will return
+.Dv NULL .
```
jrtc27, line 44: NULL is of type `void *` but it returns vm_paddr_t, an integer type

share/man/man9/pmap_kextract.9
```diff
@@ -0,0 +1,54 @@
...
+.Sh RETURN VALUES
+The
+.Fn pmap_kextract
+function will return the physical page address associated with the
+kernel virtual address
+.Fa va
+inside the physical map of the address space.
+In case of failure,
+.Fn pmap_kextract
+will return
```
jrtc27, line 43: I believe the current contract is "you must be sure it's valid', so panicking is a legitimate implementation, and more helpful perhaps than returning 0 (do we allow 0 as a legitimate physical address?..)

sys/arm/arm/pmap-v6.c
```diff
@@ -1107,9 +1109,6 @@ pmap_kextract(vm_offset_t va)
 		pte2 = pte2_load(pt2map_entry(va));
 		pa = pte2_pa(pte2) | (va & PTE2_OFFSET);
 	}
-	else {
```
jrtc27, line 1110: Stylistically, else pa = 0; is clearer

jrtc27 (changes requested):
If you want to push for this, it needs to be in Phabricator and reviewed by the likes of kib/markj/alc

share/man/man9/pmap_kextract.9
```diff
@@ -0,0 +1,58 @@
...
+The
+.Fn pmap_kextract
+function, and its alias
+.Fn vtophys ,
+map a kernel virtual address to a physical page.
+.Sh RETURN VALUES
+The
+.Fn pmap_kextract
+function will return the physical page address
+.Ft ( vm_paddr_t )
```
yuripv, line 38: Use `.Vt` for variable types, i.e. `.Pq Vt vm_paddr_t`.

share/man/man9/pmap_kextract.9
```diff
@@ -0,0 +1,58 @@
...
+associated with the kernel virtual address
+.Fa va
+inside the physical map of the address space.
+In case of failure,
```
yuripv, line 42: Please take a look at how this is done in pmap_extract.9 that you are referencing -- there should be a `.Sh RETURN VALUES` section documenting what is returned on success and failure (e.g. address and NULL).  Also I don't think saying it will panic (or fail assertion) makes sense, it shouldn't do that and looking at the code it does not for bad value issues (something being unimplemented or missing from hardware is out of scope of this man page).

yuripv (changes requested):

share/man/man9/pmap_kextract.9
```diff
@@ -0,0 +1,58 @@
...
+.Sh DESCRIPTION
+The
+.Fn pmap_kextract
+function, and its alias
+.Fn vtophys ,
+map a kernel virtual address to a physical page.
+.Sh RETURN VALUES
+The
+.Fn pmap_kextract
+function will return the physical page address
```
mhorne, line 37:
```suggestion
function will return the physical address
```

share/man/man9/pmap_kextract.9
```diff
@@ -0,0 +1,58 @@
...
+.Ft ( vm_paddr_t )
+associated with the kernel virtual address
+.Fa va
+inside the physical map of the address space.
```
mhorne, line 41:
```suggestion
.Fa va .
```

share/man/man9/pmap_kextract.9
mhorne: I agree here, it would be enough to say that zero/NULL is returned in the case of failure.

Better yet, we should state explicitly in the description the requirement that `va` **must** represent a "valid" kernel address. As in, the value of `va` is obtained by legitimate means (e.g. a pointer value from a malloc'd object). To pass an arbitrary/random value within the KVA range _might_ appear to succeed (if the virtual mapping exists), but it is a programming error subject to "undefined behavior" such as a kernel panic.

I've included this in my suggested text above.

share/man/man9/pmap_kextract.9
```diff
@@ -0,0 +1,58 @@
...
+.\" This manual page was written by Mina Galić <FreeBSD@igalic.co> under
+.\" sponsorship from the FreeBSD Foundation.
+.\"
+.Dd August 24, 2023
+.Dt PMAP_KEXTRACT 9
+.Os
+.Sh NAME
+.Nm pmap_kextract ,
+.Nm vtophys
+.Nd extract from kernel page table to a physical address
```
mhorne, line 15:
```suggestion
.Nd extract a physical address from the kernel page table
```

mhorne (changes requested):

share/man/man9/pmap_kextract.9
igalic:
```suggestion
.Pq Vt vm_paddr_t
```

share/man/man9/pmap_kextract.9
igalic: Thank you for your review.
I've incorporated your changes, and updated this section, too.
If you're still unhappy with it, we can scrap the entire In case of failure paragraph.

share/man/man9/pmap_kextract.9
```diff
@@ -0,0 +1,58 @@
...
+.Ft vm_paddr_t
+.Fo vtophys
+.Fa "vm_offset_t va"
+.Fc
+.Sh DESCRIPTION
+The
+.Fn pmap_kextract
+function, and its alias
+.Fn vtophys ,
+map a kernel virtual address to a physical page.
```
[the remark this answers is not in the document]
igalic: I've committed this with you as Co-Author.

share/man/man9/pmap_kextract.9
```diff
@@ -0,0 +1,66 @@
...
+.Fa va .
+.Pp
+.Fn pmap_kextract
+generally does not fail.
+However, if supplied with an illegitimate value for
+.Fa va
+it will return 0 on arm64, and
+.Xr panic 9
+on armv6 and riscv.
+Note that such platform dependent behaviours are subject to change.
```
jrtc27, line 56: What's the point of documenting it? Nobody should be relying on the behaviour. And this is incomplete anyway.

share/man/man9/pmap_kextract.9
```diff
@@ -0,0 +1,66 @@
...
+associated with the kernel virtual address
+.Fa va .
+.Pp
+.Fn pmap_kextract
+generally does not fail.
+However, if supplied with an illegitimate value for
+.Fa va
+it will return 0 on arm64, and
+.Xr panic 9
+on armv6 and riscv.
```
jrtc27, line 55: armv6 is a MACHINE_ARCH not a MACHINE (which would be arm) like arm64 and riscv (and armv7 exists)

sys/arm64/arm64/pmap.c
```diff
@@ -1949,6 +1949,10 @@ pmap_klookup(vm_offset_t va, vm_paddr_t *pa)
 	return (true);
 }
 
+/*
+ *  Extract from the kernel page table the physical address
```
jrtc27, line 1953: Why the change in style for arm64 and riscv?

sys/amd64/amd64/pmap.c
```diff
@@ -3933,6 +3933,12 @@ pmap_extract_and_hold(pmap_t pmap, vm_offset_t va, vm_prot_t prot)
 	return (m);
 }
 
+/*
+ *	Routine:	pmap_kextract
+ *	Function:
+ *		Extract from the kernel page table the physical address
```
jrtc27, line 3939: Can this be made to more closely follow the wording for pmap_extract? Something more like:

> Extract the physical page address associated with the given kernel virtual address.

mhorne (approved):
Thanks! I can merge this.

igalic:
merged in 74e4a8d208f0b3cf2525e3786f3efba71fcdb752

PR #842, opened 2023-09-13
axge: Skip dummy packet headers

brokad:
The AX88179A (such as that can be found in some recent Amazon Basics dongles) interweaves dummies alongside valid packet headers in USB bulk IN transfer data. This was probably done for backward compatibility with existing drivers.

However current driver records these dummy headers as dropped frames, leading to stats misreporting one Ierr per Ipkt.

This skips those dummy headers silently, thereby not generating Ierrs for them.

The patch should be compatible with older versions of the chip, as actual dropped frames should have non-zero `pktlen`.

I happen to have both versions of the chip on hand, and the patch was smoke-tested on each, with the interface working with no noticeable change from main in cursory `iperf3` tests.


emaste:
Can you confirm `Damien Broka <git@damien.sh>` is the desired/correct author info?

emaste (approved):

PR #844, opened 2023-09-15
libcrypto: link engines and the legacy provider to libcrypto

khorben:
The most efficient way to ship OpenSSL's legacy provider module and engines is to have them link to libcrypto.so. This can break the build since they are created in a sub-directory of secure/lib/libcrypto, and may be ready to link before libcrypto.so is available.

This commit introduces a LIBCRYPTO_WITHOUT_SUBDIRS define, ensuring libcrypto.so builds in its usual early phase without any OpenSSL provider module or engines. They are then completed as expected later.


emaste:
Merged in 1a18383a52bc373e316d224cef1298debf6f7e25

PR #849, opened 2023-09-20
arm/allwinner: fix variables initialization in clkng

n-p-soft:


evadot:
Thanks, will get this in the tree.

PR #850, opened 2023-09-21
makefs: Handle missing link keyword in mtree spec

stephane-rochoy-stormshield:
Small fix initially reported back in 2019 via [bug#241178](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=241178).


emaste:
Apologies that I lost track of the patch from the PR, thank you for following up. I've pushed now in d26c1a0f8bceb6748436fe3fc57bac3ef8179dd3 (with a small tweak to the error message).

PR #853, opened 2023-09-25
Fix "version introduced" in numerous manual pages

tomhukins:


share/man/man4/ftgpio.4
```diff
@@ -46,7 +46,7 @@ is a driver for the GPIO controller found on Fintek Super I/O chips.
 The
 .Nm
 manual page first appeared in
-.Fx 14.0 .
+.Fx 13.2 .
```
emaste, line 49: Double-check this one, I don't see it in 13.2?

tomhukins: Thank you for your review, @emaste.  It seems `git cherry-pick` did something unexpected with changes I made to `share/man/man4/rtw88fw.4`, which no longer exists in `main` and applied them to this file instead.

I can amend my commit to no longer contain this change, but that won't correct `rtw88fw.4` in `stable/14` and `stable/13`.  Does that seem like the best approach to you?

tomhukins: In the absence of any further feedback, I have pushed an updated version of this patch that contains everything from the previous version except my incorrect changes to `rtw88fw.4 ` in dd05d70.

tomhukins:
@emaste would you be happy to merge this now that I have corrected the mistake you spotted in October?

PR #854, opened 2023-09-28
axge: Add support for AX88179A

brokad:
Though not mentioned anywhere in documentation or driver source code, the AX88179A works fine with axge, as it's mostly the same chip as the AX88179.

The AX88179A has two firmware modes, one of which is backward compatible with existing AX88178A/179 driver (also see 70fbcd451b68b7f6038d8a602cd8d5e1bb890f1d). The active firmware mode can be controlled through a register.

Update axge(4) man page to mention 179A support. Also ensure that, when bound to a AX88179A, the driver activates the compatible firmware mode, which prevents an issue I've encountered where a soft-reset will boot up the chip in the wrong firmware.


sys/dev/usb/net/if_axge.c
```diff
@@ -70,13 +71,14 @@
...
 
 static const STRUCT_USB_HOST_ID axge_devs[] = {
-#define	AXGE_DEV(v,p) { USB_VP(USB_VENDOR_##v, USB_PRODUCT_##v##_##p) }
-	AXGE_DEV(ASIX, AX88178A),
-	AXGE_DEV(ASIX, AX88179),
-	AXGE_DEV(BELKIN, B2B128),
-	AXGE_DEV(DLINK, DUB1312),
-	AXGE_DEV(LENOVO, GIGALAN),
-	AXGE_DEV(SITECOMEU, LN032),
+#define	AXGE_DEV(v,p,i,...) { USB_VPI(USB_VENDOR_##v, USB_PRODUCT_##v##_##p, i), __VA_ARGS__ }
```
markjdb, line 74: This line is too long, just write

```
#define AXGE_DEV(v, p, i, ...)<tab>\
<tab>{ USB_VPI(...
```

sys/dev/usb/net/if_axge.c
```diff
@@ -410,6 +412,11 @@ axge_chip_init(struct axge_softc *sc)
 	axge_write_cmd_1(sc, AXGE_ACCESS_MAC, AXGE_CLK_SELECT,
 	    AXGE_CLK_SELECT_ACS | AXGE_CLK_SELECT_BCS);
 	uether_pause(&sc->sc_ue, hz / 10);
+
+	if ((sc->sc_flags & AXGE_FLAG_179A) != 0) {
+		/* Ensure 179A runs in compat 178A/179 fw mode. */
```
markjdb, line 417: I think a more detailed comment here would be good. In particular, what do we lose by enabling compat mode?

markjdb (commented):
Overall seems reasonable to me, just a couple of comments.

markjdb:
Thanks for the patch.

PR #858, opened 2023-10-06
Cirrus CI: Only trigger on pull requests

jlduran:
Since Cirrus Labs is limiting their free usage tier [^1], limit CI runs on pull requests only.  Otherwise, we might deplete our monthly quota within a few days.

[^1]: https://cirrus-ci.org/blog/2023/07/17/limiting-free-usage-of-cirrus-ci/

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -23,6 +23,7 @@ task:
       TOOLCHAIN: llvm15
       TOOLCHAIN_PKG: ${TOOLCHAIN}-lite
   - name: amd64-llvm16 World and kernel build and boot smoke test
+    only_if: $CIRRUS_BRANCH =~ 'pull/.*'
     env:
       TARGET: amd64
       TARGET_ARCH: amd64
@@ -63,7 +64,7 @@ task:
       TOOLCHAIN_PKG: ${TOOLCHAIN}
       EXTRA_MAKE_FLAGS: -s
   - name: amd64-gcc12 World and kernel build and boot smoke test (FreeBSD repo)
-    only_if: $CIRRUS_REPO_FULL_NAME == 'freebsd/freebsd-src'
+    only_if: $CIRRUS_REPO_FULL_NAME == 'freebsd/freebsd-src' && $CIRRUS_BRANCH =~ 'pull/.*'
     env:
       TARGET: amd64
       TARGET_ARCH: amd64


jlduran:
cc/ @emaste

Pull request as an idea/suggestion exercise only.

emaste:
I pushed 1/2 of this now to help conserve our free runs. I think we want to adjust the `only_if` for the `amd64-llvm16` job though, because there are downstreams that still want to run it on all pushes.

Perhaps `$CIRRUS_REPO_FULL_NAME != 'freebsd/freebsd-src' ||  $CIRRUS_BRANCH =~ 'pull/.*'` (modulo syntax errors)?

jlduran:
Great! I believe a similar commit to all active branches is also in order?

jlduran:
> Perhaps `$CIRRUS_REPO_FULL_NAME != 'freebsd/freebsd-src' || $CIRRUS_BRANCH =~ 'pull/.*'` (modulo syntax errors)?

It could also work. The project ran out of credits on day 3!

Closing it for now. I'll explore other options over the weekend.

jlduran:
Just i case I forget, the "other" options are at the moment:

- nightly cron CI jobs (https://cirrus-ci.org/guide/writing-tasks/#cron-builds)
- Experiment with @davidchisnall 's https://github.com/davidchisnall/container-vm-scripts, and use GitHub

davidchisnall:
The podman scripts are not great, but we've had a lot of success with github-act-runner on self-hosted infrastructure.  @nwf wrote some slightly terrifying scripts that take the GitHub Actions messages into an Azure message queue and then pulls them out in the head node of a cluster.  Extending that to start a pristine bhyve VM to run src builds wouldn't be too hard.  

Cirrus can also use an Azure subscription (not clear if it can use VMs / scale sets, or just ACI, which would be helpful only for cross-build tests), I'm not sure what the project did with the $20K of Azure credit that Microsoft donated but that might be a good use.

davidchisnall:
(Assuming you meant to link to https://github.com/davidchisnall/act-pot/ which is a set of scripts for running GitHub Actions in jails using pot, not https://github.com/davidchisnall/container-vm-scripts, which is a set of scripts for building VMs to use with Podman)

jlduran:
Yes, sorry for the confusion.
I believe ultimately joining [ci.freebsd.org](https://ci.freebsd.org) in one way or another, should be the way to go in the long run, with something similar to what you have pointed out, i.e., using a [self-hosted runner](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners) or https://cirrus-ci.org/guide/supported-computing-services/.
Thank you for the insight.

davidchisnall:
There's also the [Jenkins GitHub Plugin](https://plugins.jenkins.io/github/), which integrates with GitHub status notifications.

PR #863, opened 2023-10-10
Summarized ping Phabricator reviews

jlduran:
- https://reviews.freebsd.org/D42114
- https://reviews.freebsd.org/D38528
- https://reviews.freebsd.org/D39485
- https://reviews.freebsd.org/D38484
- https://reviews.freebsd.org/D39126
- https://reviews.freebsd.org/D39492
- https://reviews.freebsd.org/D39561

### DIFF
--- a/sbin/ping/main.c
+++ b/sbin/ping/main.c
@@ -35,6 +35,8 @@
 #include <netinet/in.h>
 
 #include <err.h>
+#include <math.h>
+#include <signal.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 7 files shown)


jlduran:
Rebase after 7fd2c91a291bd518e012b438d6ca6fdd04d39dbf.

jlduran:
> It seems that the tests in test_ping.py can't be run as non-root since they create an interface. Could you please update them to set `required_user = root`?

Nice catch! It's a bad habit. I believe there is an option to allow for de-escalating privileges inside the vnet, maybe these are a good candidate afterwards. At any rate, fixed!

markjdb:
This looks good to me.

As a side note, I tried running the ping tests on arm64 under QEMU. All of the test_ping.py tests timed out. I confirmed that they pass on hardware, but they're really quite slow. Merely listing the tests takes 30s in QEMU, 4s on my microsoft devkit, 2.6s on my desktop. That's really quite slow. The tests themselves take several minutes to run but that shouldn't be necessary. I wonder if there's some low-hanging fruit here?

markjdb:
Some profiling suggests that scapy import times account for a lot of the overhead. During import, scapy execs ifconfig quite a few times per interface, so if many interfaces are present this can be rather slow. I think but am not certain that that's the source of most of the overhead.

markjdb:
> Some profiling suggests that scapy import times account for a lot of the overhead. During import, scapy execs ifconfig quite a few times per interface, so if many interfaces are present this can be rather slow. I think but am not certain that that's the source of most of the overhead.

Hmm, this might be a red herring. If I run the tests serially, then they don't time out in QEMU. For some reason, though, if I run four tests in parallel in a VM with 4 vCPUs, I see very high system CPU usage and timeouts. This might be a kernel problem. It would still be good if we could improve scapy import times somehow though...

markjdb:
Merged, thank you.

jlduran:
Yes, I have "documented" this fact when we decided to use atf_python instead of atf-sh. I'll look into "caching" scapy or something... I can replicate the slowness using a bhyve VM with a single core and 1GB RAM to run these tests. Thank you!

jlduran:
combined

igalic:
over in cloud-init land, i wrote an [ifconfig parser](https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/parsers/ifconfig.py)  for BSDs that takes the entire input of `ifconfig -a` and LRU caches it, processing it only once. calls for specific a interface's information are processed from that cache. iOf course that requires [lugging around state](https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/networking.py#L179), but maybe that's for the better

PR #865, opened 2023-10-12
atf_python: Miscellaneous minor improvements

jlduran:
Sorry, but these are too small for me to manage in Phabricator.

This is the sad part of GitHub, the CODEOWNERS file is not 100% effective:
cc/ @markjdb @AlexanderChernikov @ngie-eign @emaste

### DIFF
--- a/tests/atf_python/sys/net/vnet.py
+++ b/tests/atf_python/sys/net/vnet.py
@@ -180,6 +180,8 @@ class IfaceFactory(object):
 
     @staticmethod
     def is_autodeleted(iface_name: str) -> bool:
+        if iface_name == "lo0":
+            return False
         iface_type = re.split(r"\d+", iface_name)[0]
         return iface_type in IfaceFactory.AUTODELETE_TYPES
 
--- a/tests/atf_python/sys/net/vnet.py
+++ b/tests/atf_python/sys/net/vnet.py
@@ -82,13 +82,9 @@ class VnetInterface(object):
     def set_jailed(self, jailed: bool):
         self.jailed = jailed
 
-    def run_cmd(
-        self,
-        cmd,
-        verbose=False,
-    ):
+    def run_cmd(self, cmd, verbose=False):
         if self.vnet_name and not self.jailed:
-            cmd = "jexec {} {}".format(self.vnet_name, cmd)
+            cmd = "/usr/sbin/jexec {} {}".format(self.vnet_name, cmd)
         return run_cmd(cmd, verbose)
 
     @classmethod
@@ -189,7 +185,7 @@ class IfaceFactory(object):
     def cleanup_vnet_interfaces(self, vnet_name: str) -> List[str]:
         """Destroys"""
         ifaces_lst = ToolsHelper.get_output(
-            "/usr/sbin/jexec {} ifconfig -l".format(vnet_name)
+            "/usr/sbin/jexec {} /sbin/ifconfig -l".format(vnet_name)
         )
         for iface_name in ifaces_lst.split():
             if not self.is_autodeleted(iface_name):
@@ -197,7 +193,7 @@ class IfaceFactory(object):
                     print("Skipping interface {}:{}".format(vnet_name, iface_name))
                     continue
             run_cmd(
-                "/usr/sbin/jexec {} ifconfig {} destroy".format(vnet_name, iface_name)
+                "/usr/sbin/jexec {} /sbin/ifconfig {} destroy".format(vnet_name, iface_name)
             )
 
     def cleanup(self):
@@ -231,7 +227,7 @@ class VnetInstance(object):
 
     def run_vnet_cmd(self, cmd):
         if not self.attached:
-            cmd = "jexec {} {}".format(self.name, cmd)
+            cmd = "/usr/sbin/jexec {} {}".format(self.name, cmd)
         return run_cmd(cmd)
 
     def disable_dad(self):
@@ -269,7 +265,7 @@ class VnetFactory(object):
 
     @staticmethod
     def _wait_interfaces(vnet_name: str, ifaces: List[str]) -> List[str]:
-        cmd = "jexec {} /sbin/ifconfig -l".format(vnet_name)
+        cmd = "/usr/sbin/jexec {} /sbin/ifconfig -l".format(vnet_name)
         not_matched: List[str] = []
         for i in range(50):
             vnet_ifaces = run_cmd(cmd).strip().split(" ")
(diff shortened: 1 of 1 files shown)


tests/atf_python/sys/net/vnet.py
```diff
@@ -180,6 +180,7 @@ def create_iface(self, alias_name: str, iface_name: str) -> List[VnetInterface]:
 
     @staticmethod
     def is_autodeleted(iface_name: str) -> bool:
+        return "lo0" not in iface_name
```
markjdb, line 180: Doesn't this unconditionally return, i.e., the subsequent lines are never executed?

igalic: yeah, this would need an `if` to be actually conditional.

jlduran: Yes. The overall way it works is that for each interface, it checks if `is_autodeleted`, so I wan it to return `False` if the interface is `lo0`, it will call this method once for each interface.
Previously any loopback interface was added to the list of interfaces to be automatically deleted.
This change will produces the `lo0` to be skipped, for example:
```
...
sbin/ping/test_ping.py::TestPing::test_pinger[_0_0_special_wrong] ==== vnet cleanup ====
# topology_id: 'TestPing:test_pinger[_0_0_special_wrong]'
Skipping interface pytest:TestPing:test_pinger__0_0_special_wrong:lo0
run: '/usr/sbin/jexec pytest:TestPing:test_pinger__0_0_special_wrong ifconfig epair1a destroy'
run: '/usr/sbin/jexec pytest:TestPing:test_pinger__0_0_special_wrong ifconfig tun0 destroy'
run: '/usr/sbin/jail -r pytest:TestPing:test_pinger__0_0_special_wrong'
PASSED

```

jlduran: I'll re-submit.

tests/atf_python/sys/net/vnet.py
```diff
@@ -180,6 +180,8 @@ def create_iface(self, alias_name: str, iface_name: str) -> List[VnetInterface]:
 
     @staticmethod
     def is_autodeleted(iface_name: str) -> bool:
+        if iface_name == "lo0":
+            return False
         iface_type = re.split(r"\d+", iface_name)[0]
         return iface_type in IfaceFactory.AUTODELETE_TYPES
```
markjdb, line 183: I think the AUTODELETE_TYPES thing is not a great idea in general. It requires the test framework to maintain a list of interface types which will get stale over time. I'm not sure why it doesn't include `if_bridge` or `if_wg` for example.

Is the intent here to destroy all interfaces belonging to a vnet before destroying the vnet itself?

tests/atf_python/sys/net/vnet.py
```diff
@@ -189,15 +185,15 @@ def is_autodeleted(iface_name: str) -> bool:
...
+            "/usr/sbin/jexec {} /sbin/ifconfig -l".format(vnet_name)
         )
         for iface_name in ifaces_lst.split():
             if not self.is_autodeleted(iface_name):
                 if iface_name not in self._list_ifaces():
                     print("Skipping interface {}:{}".format(vnet_name, iface_name))
                     continue
             run_cmd(
-                "/usr/sbin/jexec {} ifconfig {} destroy".format(vnet_name, iface_name)
+                "/usr/sbin/jexec {} /sbin/ifconfig {} destroy".format(vnet_name, iface_name)
```
markjdb, line 196: Is your intent just to fix the inconsistency that some places specify a full path and others don't? I don't think it matters too much one way or the other, but the commit log message should be a bit more clear.

markjdb (approved):
I think I'm ok with the change. I'll try running the test suite with this applied.

tests/atf_python/sys/net/vnet.py
jlduran: > I think the AUTODELETE_TYPES thing is not a great idea in general. It requires the test framework to maintain a list of interface types which will get stale over time. I'm not sure why it doesn't include `if_bridge` or `if_wg` for example.

Probably because there are no tests that need these types at the moment.

> Is the intent here to destroy all interfaces belonging to a vnet before destroying the vnet itself?

Yes, similarly to what is done in atf-sh (https://github.com/freebsd/freebsd-src/commit/80fc25025ffcb0d369fc0b6d4d272ad6fd3f53c3)

tests/atf_python/sys/net/vnet.py
jlduran: Right, I'll change the commit message. Chances are very low that someone overrides these in the PATH, especially within a VNET.
These are all minor changes, maybe changing all string formatting to f-strings (as these are arguably more readable and slightly better performant) is next. They can all be dropped if you don't consider it is worth the effort.

tests/atf_python/sys/net/vnet.py
markjdb: Hmm, ok. The problem that motivated commit 80fc250 is now fixed, so I'm still kind of skeptical, but alright.

tests/atf_python/sys/net/vnet.py
markjdb: I think the changes are fine, I'm asking questions in part because I'm not too familiar with atf_python.

tests/atf_python/sys/net/vnet.py
jlduran: Ahh.. OK, I believe @AlexanderChernikov's words were at the time:
> There is a known problem resulting from a corner case in the kernel delayed object reclamation model.

I'll try test destroying everything from the host... but I remember half-epairs remaining and something similar to https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=270628 when testing with VLANs.

tests/atf_python/sys/net/vnet.py
jlduran: I have no problem at all with questions!
As we already know, it makes pytest slow once you add `--atf` to its parameters, as you well pointed out, if there are low hanging-fruits that could it speed up... I'm just trying to comb it a little.
pytest runs twice per test:
1. To run the test:
```
# pytest -vv -p no:cacheprovider -s --atf --confcutdir=/usr/tests --atf-source-dir=/usr/tests/atf_python <name of single test here>
```
2. To cleanup:
```
# pytest -vv -p no:cacheprovider -s --atf --confcutdir=/usr/tests --atf-cleanup --atf-source-dir=/usr/tests/atf_python <name of single test here>
```
In case you want to experiment with pytest directly.
I have not found anything obvios at the moment, but one minor thing that I found confusing when creating atf_python tests was that the debugging information is not immediately clear, I'm trying to familiarize with pytest's internals a little more as well.

markjdb:
Merged, thank you.

PR #868, opened 2023-10-15
Optimize bitwise operations where possible

AtariDreams:
We can use compiler intrinsics to do this instead of doing it manually.

### DIFF
--- a/sys/dev/ixl/i40e_common.c
+++ b/sys/dev/ixl/i40e_common.c
@@ -3800,7 +3800,7 @@ static void i40e_parse_discover_capabilities(struct i40e_hw *hw, void *buff,
 				     enum i40e_admin_queue_opc list_type_opc)
 {
 	struct i40e_aqc_list_capabilities_element_resp *cap;
-	u32 valid_functions, num_functions;
+	u32 num_functions;
 	u32 number, logical_id, phys_id;
 	struct i40e_hw_capabilities *p;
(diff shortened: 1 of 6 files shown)


sys/dev/ixl/i40e_common.c
```diff
@@ -4124,13 +4124,7 @@ static void i40e_parse_discover_capabilities(struct i40e_hw *hw, void *buff,
...
 	}
 
-	valid_functions = p->valid_functions;
-	num_functions = 0;
-	while (valid_functions) {
-		if (valid_functions & 1)
-			num_functions++;
-		valid_functions >>= 1;
-	}
+	num_functions = __builtin_popcount(p->valid_functions);
```
ricera, line 4127: You should use bitcount32() here, instead.

ricera (changes requested):
I'm going to echo what @igalic says and that this PR should have a separate commit for each subsystem that it touches, as well as each commit explaining what is improved i.e. in Linux kernel terms, why would someone merge this commit into their tree?

I gave feedback on the ixl change since I maintain that, and I think there's a decent reason there -- it's easier to understand what a single `bitcount32()` call is doing instead of several lines that include a for loop and shifts.

bsdimp:
Absent a specific bug these are fixing, or performance measurements that outline the benfit, this falls under the wide ranging change rule that we just don't have the time to properly evaluate.

PR #869, opened 2023-10-17
Use C99 flexible array

ElyesH:
zero-length arrays have been deprecated since last millennium.
https://gcc.gnu.org/onlinedocs/gcc/Zero-Length.html


igalic (approved):
👍 thanks for taking this on

igalic:
@ElyesH could you do a rebase on this branch so we get a fresh CI build?
None of the CI failures are related to your patch.

igalic:
see also dc0b4094abf6784bf1a9492c2fea3fb91116b014

ElyesH:
ping

sys/ofed/include/rdma/rdmavt_qp.h
```diff
@@ -179,7 +179,7 @@ struct rvt_swqe {
 struct rvt_rwqe {
 	u64 wr_id;
 	u8 num_sge;
-	struct ib_sge sg_list[0];
+	struct ib_sge sg_list0[];
```
ehem, line 182: Typo, which will break things!

ehem (commented):
Basically okay and some good cleanup to match the standards FreeBSD is supposed to target.  One typo which almost certainly breaks things though.

jrtc27:
This doesn't actually build. Flexible arrays are only allowed in the outermost aggregate, yet there are cases where these aggregates are embedded in others and therefore are invalid. Please build test your patches.

sys/ofed/include/rdma/ib_verbs.h
```diff
@@ -1710,7 +1710,7 @@ struct ib_flow_eth_filter {
 	__be16	ether_type;
 	__be16	vlan_tag;
 	/* Must be last */
-	u8	real_sz[0];
```
jrtc27, line 1713: As the name implies, these are never non-zero length. These are just so the code can calculate the size of the struct without its tail padding. Kind of ugly, but that's how the code works.

bsdimp:
I'm going to close this. It's scope is too large. As @jrtc27 said, this needs to be done more carefully.  Many of these changes are good, but separating the wheat from the chaff is too hard. This means this is not a good candidate for a pull request.

PR #870, opened 2023-10-17
Remove $FreeBSD$: one line sh-pattern

jlduran:
Remove /^\s*#[#!]?\s*\$FreeBSD\$.*$\n/

cc/ @bsdimp


usr.bin/mkimg/tests/mkimg_test.sh
```diff
@@ -86,9 +86,7 @@ mkimg_rebase()
     fi
 
     if test $update = yes; then
-	# Prevent keyword expansion when writing the keyword.
-	(echo -n '# $'; echo -n FreeBSD; echo '$') > $baseline
```
bsdimp, line 90: I deleted this three times today....

bsdimp (approved):
This is good

emaste (approved):

bsdimp:
Pushed as 3b5c9845727a with a re-worded commit message.
Thanks for saving me the trouble of doing this... In testing I removed the pattern from the shell script one time too few :(.

PR #871, opened 2023-10-18
devfs.8: Remove references to /usr/share/examples/etc/devfs.conf

jlduran:
This file does not exist, remove it from the list of files to avoid confusion.  The example file is just /etc/devfs.conf.

---

I was not able to find it in the git log.
Reading [share/examples/etc/README.examples](https://github.com/freebsd/freebsd-src/blob/main/share/examples/etc/README.examples), it seems as if at some point all /etc's conf files were copied there in case a recovery was necessary.

### DIFF
--- a/sbin/devfs/devfs.8
+++ b/sbin/devfs/devfs.8
@@ -23,7 +23,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd December 1, 2020
+.Dd October 18, 2023
 .Dt DEVFS 8
 .Os
 .Sh NAME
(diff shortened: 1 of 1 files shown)


mhorne (approved):
Nice find, I will merge this.

This bit about `/usr/share/examples/etc` containing untouched versions of configuration files is interesting...

That functionality appears quite old, with the `README.examples` file dating back to 1994 at least (d1e0c8254c1a). I was not able to determine how long this has been broken, but some relevant make logic was removed in 2019 (a8b94f3d7980ab609a4f6dac11f23eecbdcce892), and it seems to have been non-functional at that point.

@jlduran do you have any interest in investigating further? At this point I do not think it is worth significant effort to restore the functionality, but some more cleanup (at least this readme) is probably needed.

jlduran:
> @jlduran do you have any interest in investigating further? At this point I do not think it is worth significant effort to restore the functionality, but some more cleanup (at least this readme) is probably needed.

I find it interesting... the original idea kinda complements rescue(8).

The way /usr/share/examples/etc currently is —as you well pointed out— is almost useless (maybe a template for `wpa_supplicant.conf` at most). I'm currently setting up a local Jenkins instance, so it'll take me some time to investigate. Sadly it is not the only forsaken directory in the /usr/share/examples' treasure trove.

markjdb:
Can this PR be closed?

PR #872, opened 2023-10-18
Remove unreachable breaks

rilysh:
Most binutils, `usage()` function exits with `exit(<EXIT_CODE>)` and so `break` after `usage()` function becomes unreachable. This PR nitpick on this.


rilysh:
I've spilled changes in three multiple commits for each one. Here's the single version in one patch: https://paste.debian.net/plain/1295460

clausecker:
This change should be combined with including `<stdnoreturn.h>` and marking the `usage` function as `noreturn`.

rilysh:
> This change should be combined with including `<stdnoreturn.h>` and marking the `usage` function as `noreturn`.

`_Noreturn` was first introduced in C11, but the codebase is most likely compatible with C99. I generally don't see a point in doing this. There's not much going on there to hint the compiler for any optimization. However, I'll happily add them, if one or more maintainers also approved to add this.

clausecker:
We have a shim for `_Noreturn` in `<sys/cdefs.h>` to support pre-C11 compilers, so it'll be fine.

rilysh:
Thanks for the clarification @clausecker! One more question, should I apply the `noreturn` attribute to all binutils or should I just stick with current changed files?

bsdimp:
> Thanks for the clarification @clausecker! One more question, should I apply the `noreturn` attribute to all binutils or should I just stick with current changed files?

I'd stick to the directory you've already done. Once all the fussy details are done, I'll land that and you can do other binaries, one commit per directory. 

I'm surprised these annotations are needed. Can't the compiler infer that because exit() is called always and that's already marked?

clausecker:
@rilysh I don't know really.  Personally I think it might be helpful to be more consistent with `noreturn`.  The main advantage is that it communicates intent to the compiler and can help in case by accident such functions can actually return in one way or another.

Whatever I write is just my personal opinion.  I am not a src committer and am not familiar enough with the source to make a decision.

rilysh:
> > Thanks for the clarification @clausecker! One more question, should I apply the `noreturn` attribute to all binutils or should I just stick with current changed files?
> 
> I'd stick to the directory you've already done. Once all the fussy details are done, I'll land that and you can do other binaries, one commit per directory.
> 
> I'm surprised these annotations are needed. Can't the compiler infer that because exit() is called always and that's already marked?

There don't seem to be any clear description from GCC/Clang side, but as far as I can see, `noreturn` attribute is quite explicit and a lot of code base aren't yet ready for this change.

For example
```c
#include <stdlib.h>

void f(int n)
{
        if (n > 0)
              abort();
}
```

This function will have undefined behaviour if n <= 0. Compiler will not generate an warning message for this and it still may return to the caller. If we mark this function as `noreturn` this will confirm from our side that this function will never return anything, but in fact it still returns that undefined state and compiler will emit an warning for it.

There are some other reasons too, which I'm not sure about. For now the best is to explicitly adding `noreturn` than relying on the compiler.

rilysh:
@bsdimp Should I start dropping commits per-directory (on this current PR) or should I create another PR?

rilysh:
@bsdimp Any updates on this?

bsdimp:
OK. This change is probably fine. The _Noreturn stuff, I don't know and is 'mission creep' from the original PR so make it separate. This is a good change to learn how to do things, but I don't want to encourage wholesale changes like this.
Since it's been here a while, I'll go ahead and fix two minor nits I noticed with the commit messages: (1) there needs to be a blank line after the first line in these messages to set off the 'subject' of the commit from the rest. and (2) All the merge with FreeBSD commits should have been a rebase + forced push. Next time I'll bounce...

bsdimp:
merged.

PR #873, opened 2023-10-18
Optimize bitwise operations in pfctl

AtariDreams:
We can use compiler intrinsics to do this instead of doing it manually.

### DIFF
--- a/sbin/pfctl/pf_print_state.c
+++ b/sbin/pfctl/pf_print_state.c
@@ -447,17 +447,17 @@ print_state(struct pfctl_state *s, int opts)
 int
 unmask(struct pf_addr *m, sa_family_t af)
 {
-	int i = 31, j = 0, b = 0;
+	int b = 0;
+	unsigned int i = 0;
 	u_int32_t tmp;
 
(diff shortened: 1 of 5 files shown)


AtariDreams:
@bsdimp Thoughts?

igalic:
cc @kprovost

bsdimp:
that's a lot of code churn that takes some concentration to see if they are the same.
What's the measured improvement in speed?

bsdjhb:
This is too noisy a diff.  Most if it is converting signed to unsigned.  For the code in question, it would likely be easier to read/understand if it just used builtin_popcount or bitcount32 from <sys/bitcount.h>:

```
     return (bitcount32(mask[0]) + bitcount32(mask[1]) + bitcount32(mask[2]) + bitcount32(mask[3]));
```

or the like.  That said, it's also not performance critical, so readability is more important than speed here.

bsdimp:
Since this isn't fixing a specific bug, and there's no demonstrated performance win, I'm going to close it. We don't have the time to process changes like this and there's a high risk of regression.

PR #876, opened 2023-10-22
Make bpf interop consistent with if_loop

PHSpline:
The pseudo_AF_HDRCMPLT check is already being done in if_loop and just needed to be ported over to if_ic, if_wg, if_disc, if_gif, if_gre, if_me, if_tuntap and ng_iface.  This is needed in order to allow these interfaces to work properly with e.g., tcpreplay.

PR:	256587

### DIFF
--- a/sys/dev/iicbus/if_ic.c
+++ b/sys/dev/iicbus/if_ic.c
@@ -363,8 +363,8 @@ icoutput(if_t ifp, struct mbuf *m, const struct sockaddr *dst,
 	u_char *cp;
 	u_int32_t hdr;
 
-	/* BPF writes need to be handled specially. */ 
-	if (dst->sa_family == AF_UNSPEC)
+	/* BPF writes need to be handled specially. */
+	if (dst->sa_family == AF_UNSPEC || dst->sa_family == pseudo_AF_HDRCMPLT)
 		bcopy(dst->sa_data, &hdr, sizeof(hdr));
 	else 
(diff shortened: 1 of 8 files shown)


markjdb:
I think this patch looks good. I wonder why we set a special `sa_family` to begin with - it looks like BPF can just set `bd_hdrcmplt` and avoid using pseudo_AF_HDRCMPLT. That would let us avoid adding these new branches to the data path.

bsdimp:
> I think this patch looks good. I wonder why we set a special `sa_family` to begin with - it looks like BPF can just set `bd_hdrcmplt` and avoid using pseudo_AF_HDRCMPLT. That would let us avoid adding these new branches to the data path.

I'm a little unclear what you are saying here @markjdb. Are you requesting this change? Or are you saying this change is good and that might be a good followup?

bsdimp:
(I ticked needs-review so I'll check back for Mark's answer later)

markjdb:
> > I think this patch looks good. I wonder why we set a special `sa_family` to begin with - it looks like BPF can just set `bd_hdrcmplt` and avoid using pseudo_AF_HDRCMPLT. That would let us avoid adding these new branches to the data path.
> 
> I'm a little unclear what you are saying here @markjdb. Are you requesting this change? Or are you saying this change is good and that might be a good followup?

I'm not requesting the change, I just think it's a better way to go in the future. I'll merge this patch now.

markjdb:
Merged, thank you.

PR #877, opened 2023-10-27
isp(4): Rework firmware handling/loading + ispfw(4): Update and add firmware

jpulz:
Rework of firmware handling especially for adapters with primary and secondary firmware region in flash.

Please see commit messages and https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=273263 for further details.

This fixes PR 273263


share/man/man4/isp.4
```diff
@@ -80,12 +80,13 @@ FC-Tape is automatically enabled when connecting controller that supports
 it to a target that supports it.
 It may be disabled using configuration and hint options described below.
 .Sh FIRMWARE
-Firmware loading is supported if the
+Firmware loading is supported and handled by
+.Xr firmware 9 . The correct firmware is either loaded automatically, if
```
igalic, line 84: please leave directives on their own lines, and start new sentences on new lines

jpulz: Did you tried this? I did before deciding to do it this way.

The way  I did it:
```roff
Firmware loading is supported and handled by
.Xr firmware 9 . The correct firmware is
```
and the output produced by mandoc(1):
```
Firmware loading is supported and handled by firmware(9). The correct firmware is
                                                         ^
```
There is only one space between sentence end and next sentence.

The _more correct_ way you proposed (and I'm aware of) **Xr** macro on own line with sentence ending dot and following sentence on the next line:
```roff
Firmware loading is supported and handled by
.Xr firmware 9 .
The correct firmware is
``` 
and the output produced by mandoc(1):
```
Firmware loading is supported and handled by firmware(9).  The correct firmware is
                                                         ^^
```
You see, there is an ugly and wrong double space between sentence end and next sentence.

If you do the same with a "," (comma) instead of a "." (dot), everything is fine and no double spaces appear in the output.
So what's the resolution here? Do it _correct_ with **Xr** macro on it's own line with the sentence ending dot and live with the ugly and wrong double space?
Maybe there is some trick I'm not aware of to make mandoc(1) behave correctly in this situation?

jpulz: Fixed.

markjdb:
The firmware updates seem uncontroversial; the other patch is rather large. It would be best to split the style fixes into a second patch, it's rather tedious to read otherwise.

@amotin @bsdimp is this something you could review?

sys/dev/isp/ispreg.h
```diff
@@ -266,6 +266,25 @@ typedef struct {
 #define ISP28XX_BASE_ADDR	0x7f7d0000
 #define ISP28XX_FLT_ADDR	(0x11000 / 4)
 
+#if 0
```
markjdb, line 269: Why is this zeroed out?

sys/dev/ispfw/asm_2500.h
```diff
@@ -16079,6 +16079,6 @@ static const uint32_t isp_2500_risc_code[] = {
 	0x00000000, 0x00000000, 0x00000000, 0x00000000,
 	0x00000000, 0x00000000, 0x00000000, 0x00000000,
 	0x00000000, 0x00000000, 0x000f1f01, 0xc3002003,
-	0xb9f6ad13, 0xffffb7ea
+	0xb9f6ad13, 0xffffb7ea, 0xffee7e37
```
markjdb, line 16082: What is this change?

sys/dev/isp/ispreg.h
jpulz: This block was already added for my upcoming rewrite of the NVRAM and FLT parts.
As it does not belong to this pull request, I've removed it for now.

sys/dev/ispfw/asm_2500.h
jpulz: Nice catch - this actually belongs to the previous commit. Fixed this by rebasing.
Background:
The firmware of those adapters is part of a so called MBI (MultiBootImage) that consists of many parts.
In a first attempt I've extracted the firmware "manually" by finding the starting point and doing some calculations to find the end. This was fine so far, but bad for new releases of the MBI/firmware in the future.
After I finished all the other parts of isp(4) and the pull request was out, I wrote a small python script to automatically extract the firmware part from the MBI.
While comparing the manually and automatically extracted firmware I found that there is a one DWORD difference in the firmware for the 25xx adapters.
I triple checked everything and I'm now sure that I made an off by one error in my first calculations and manually extracted one DWORD less out of the MBI.
So this was just to fix my calculation error and is now included in the previous firmware commit.

jpulz:
> The firmware updates seem uncontroversial; the other patch is rather large. It would be best to split the style fixes into a second patch, it's rather tedious to read otherwise.
> 
> @amotin @bsdimp is this something you could review?

Mark, thanks for your thoughts and comments.
I've removed most of the style fixes for now. I've only left trailing whitespace removals and two ```sizeof value -> sizeof(value)``` changes.
I've removed some stuff I already added/changed for future work too, as it is currently unrelated and not needed.
This makes the diff a bit smaller and hopefully easier to review.

sys/dev/isp/isp.c
```diff
@@ -432,11 +559,28 @@ isp_reset(ispsoftc_t *isp, int do_load_defaults)
...
+	mbs.param[11] = 0;
+	if (IS_27XX(isp))
+		mbs.param[4] |= 0x08;	/* NVME_ENABLE_FLAG */
 	isp_mboxcmd(isp, &mbs);
 	if (mbs.param[0] != MBOX_COMMAND_COMPLETE)
 		return;
+	fcp->fw_ability_mask = (mbs.param[3] << 16) | mbs.param[2];
+	isp_prt(isp, ISP_LOGDEBUG0, "Firmware ability mask: 0x%x",
+	    fcp->fw_ability_mask);
+	if (IS_27XX(isp)) {
```
amotin, line 573: Why not IS_26XX() here?  My 26xx+27xx specs do mention it.

sys/dev/isp/isp.c
```diff
@@ -452,103 +596,160 @@ isp_reset(ispsoftc_t *isp, int do_load_defaults)
 	isp->isp_fwrev[1] = mbs.param[2];
 	isp->isp_fwrev[2] = mbs.param[3];
 	isp->isp_fwattr = mbs.param[6];
-	isp->isp_fwattr |= ((uint64_t) mbs.param[15]) << 16;
-	if (isp->isp_fwattr & ISP2400_FW_ATTR_EXTNDED) {
-		isp->isp_fwattr |=
-		    (((uint64_t) mbs.param[16]) << 32) |
-		    (((uint64_t) mbs.param[17]) << 48);
+	if (isp->isp_fwattr & ISP_FW_ATTR_EXTNDED) {
+		isp->isp_fwattr_h = mbs.param[15];
```
amotin, line 600: Why did you move this under ISP_FW_ATTR_EXTNDED?  According to my specs the flag covers only registers 15 and 16.

sys/dev/isp/isp.c
```diff
@@ -4662,15 +4858,468 @@ isp_parse_flt_2xxx(ispsoftc_t *isp, uint8_t *flt_data)
...
+}
+
+static bool
+isp_check_image_status_signature(struct isp_image_status *image_status)
+{
+	unsigned long signature = le32toh(image_status->signature);
+
+	return
+	    signature != ISP27XX_IMG_STATUS_SIGN &&
+	    signature != ISP28XX_IMG_STATUS_SIGN;
```
amotin, line 4891: Why no parentheses?

sys/dev/isp/isp.c
```diff
@@ -4662,15 +4858,468 @@ isp_parse_flt_2xxx(ispsoftc_t *isp, uint8_t *flt_data)
...
+	    active_regions->global == ISP27XX_SECONDARY_IMAGE ?
+		"secondary" : "invalid",
+	    active_regions->global);
+}
+
+static bool isp_risc_firmware_invalid(ispsoftc_t *isp, uint32_t *dword)
+{
+	return
+	    !(dword[4] | dword[5] | dword[6] | dword[7]) ||
+	    !(~dword[4] | ~dword[5] | ~dword[6] | ~dword[7]);
```
amotin, line 5149: Also misses parentheses.  And since it is not boolean ' == 0' would be better.

sys/dev/isp/ispvar.h
```diff
@@ -551,11 +598,12 @@ struct ispsoftc {
 #define	ISP_CFG_LPORT_ONLY	0x02	/* insist on {N/F}L-Port connection */
 #define	ISP_CFG_NPORT_ONLY	0x04	/* insist on {N/F}-Port connection */
 #define	ISP_CFG_LPORT		0x06	/* prefer {N/F}L-Port connection */
+					/* 0x06 seems wrong to me */
```
amotin, line 601: Its not wrong, since reads are masked by ISP_CFG_PORT_PREF.  It is one 3-bit fields, not three 1-bit fields.

sys/dev/isp/ispvar.h
```diff
@@ -613,18 +661,24 @@ struct ispsoftc {
...
-#define	ISP_FW_MINOR(code)		((code >> 16) & 0xff)
-#define	ISP_FW_MICRO(code)		((code >>  8) & 0xff)
-#define	ISP_FW_REVX(xp)			((xp[0]<<24) | (xp[1] << 16) | xp[2])
+#define	ISP_FW_REV(major, minor, micro)				\
+	((major << 16) & 0xff0000 | (minor << 8) & 0x00ff00 | micro & 0x0000ff)
+#define	ISP_FW_MAJOR(code)		((code & 0xff0000) >> 16)
+#define	ISP_FW_MINOR(code)		((code & 0x00ff00) >> 8)
+#define	ISP_FW_MICRO(code)		(code & 0x0000ff)
+#define	ISP_FW_REVX(xp)						\
+	((xp[0] << 16) & 0xff0000 | (xp[1] << 8) & 0x00ff00 | xp[2] & 0x0000ff)
```
amotin, line 670: While you are editing them, macro arguments should be in parentheses.  And would you apply masks and shifts in different order the strings could be shorter.  Not sure why did you change them.

amotin (commented):
Thank you for working on this!  I haven't tested it or looked too deep, but still some comments:

sys/dev/isp/isp.c
```diff
@@ -321,41 +383,113 @@ isp_reset(ispsoftc_t *isp, int do_load_defaults)
...
+			    fcp->fw_ispfwrev[1], fcp->fw_ispfwrev[2]);
+			isp_prt(isp, ISP_LOGCONFIG,
+			    "Firmware revision (ispfw) %u.%u.%u (%x).",
+			    fcp->fw_ispfwrev[0], fcp->fw_ispfwrev[1],
+			    fcp->fw_ispfwrev[2], fcp->fw_ispfwrev[3]);
+		} else {
+			snprintf(fcp->fw_version_ispfw,
+			    sizeof(fcp->fw_version_ispfw), "load error");
+			isp_prt(isp, ISP_LOGERR,
+			    "Error loading ispfw(4) firmware %s", fwname);
```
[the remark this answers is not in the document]
jpulz: It may be either due to firmware absence or ```firmware_get()``` failed for whatever reason.
As ```firmware_get()``` is unspecific and just returns ```NULL``` on any failure we can not further distinguish.
Could be changed to ```ISP_LOGWARN``` if this fits better in your eyes.

sys/dev/isp/isp.c
```diff
@@ -4662,15 +4858,468 @@ isp_parse_flt_2xxx(ispsoftc_t *isp, uint8_t *flt_data)
...
+	int rval = ISP_SUCCESS;
+	unsigned int segments, fragment;
+	unsigned long i;
+	unsigned int j;
+	unsigned long dlen;
+	uint32_t *dcode;
+	uint32_t risc_addr, risc_size = 0;
+
+	isp_prt(isp, ISP_LOGDEBUG0,
+	    "FW: Loading firmware from flash (0x%x).", faddr);
```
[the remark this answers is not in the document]
jpulz: There is an `isp_read_flash_data()` directly after it where we read the first 8 DWORDs out of flash and do integrity verification and get the version number from.
So we do read, but not the full firmware.
The following should be more correct:
```
isp_prt(isp, ISP_LOGDEBUG0,
    "Accessing flash firmware at 0x%x.", faddr);
```

sys/dev/isp/isp.c
```diff
@@ -432,11 +559,28 @@ isp_reset(ispsoftc_t *isp, int do_load_defaults)
 	MBSINIT(&mbs, MBOX_EXEC_FIRMWARE, MBLOGALL, 5000000);
 	mbs.param[1] = code_org >> 16;
 	mbs.param[2] = code_org;
-	if (!IS_26XX(isp))
-		mbs.param[3] = loaded_fw ? 0 : 1;
+	mbs.param[3] = loaded_fw ? 0 : 1;
```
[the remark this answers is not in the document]
jpulz: As I don't have any specs at all, my whole work is based on reading and understanding the Linux driver provided by Marvell/QLogic.
There is no differentiating for this register in the Linux driver. It's just `mcp->mb[3] = 0;` there, no matter if firmware was loaded or not and what adapter/chip generation.

sys/dev/isp/isp.c
jpulz: Again, no specs here just the Linux driver.
```
if (IS_QLA27XX(ha) || IS_QLA28XX(ha)) {
                ha->max_supported_speed = mcp->mb[2] & (BIT_0|BIT_1);
                ql_dbg(ql_dbg_mbx, vha, 0x119b, "max_supported_speed=%s.\n",
....
```
So I assumed that only 27xx+ support reporting the max supported speed.

sys/dev/isp/isp.c
jpulz: You're right. This should be changed to
```
isp->isp_fwattr = mbs.param[6];
isp->isp_fwattr_h = mbs.param[15];
if (isp->isp_fwattr & ISP_FW_ATTR_EXTNDED) {
        isp->isp_fwattr_ext[0] = mbs.param[16];
        isp->isp_fwattr_ext[1] = mbs.param[17];
}
```
The Linux driver isn't using this flag at all. They decide on the FWI2 (firmware interface 2 ???) capability of the adapter, which is true for all 24xx+.

sys/dev/isp/isp.c
jpulz: Sorry, missed that. Is
```
return
    ((signature != ISP27XX_IMG_STATUS_SIGN) &&
    (signature != ISP28XX_IMG_STATUS_SIGN));
```
better?

sys/dev/isp/ispvar.h
jpulz: The old code was wrong for the micro version:
```
#define ISP_FW_REV(maj, min, mic)        ((maj << 24) | (min << 16) | mic)
#define ISP_FW_MICRO(code)               ((code >>  8) & 0xff)
```
Additionally, I don't think we will see a firmware version with major|minor|micro > 255 anytime soon. To represent 255.255.255 shifting like this `((maj << 16) | (min << 8) | mic)` is enough.

Here is a second try taking your notes into account:
```
#define ISP_FW_REV(maj, min, mic)       ((maj << 16) | (min << 8) | mic)
#define ISP_FW_MAJOR(code)              ((code >> 16) & 0xff)
#define ISP_FW_MINOR(code)              ((code >> 8) & 0xff)
#define ISP_FW_MICRO(code)              (code & 0xff)
#define ISP_FW_REVX(xp)                 ((xp[0] << 16) | (xp[1] << 8) | xp[2])
```
What do you think about it?

sys/dev/isp/isp.c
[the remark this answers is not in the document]
amotin: It would not change anything, since both WARN and ERR are enabled by default.  I'd prefer to to be rephrased somehow to not look like an error, since I guess in most cases it will just mean that ispfw(4) is not loaded, which is not an error.

sys/dev/isp/isp.c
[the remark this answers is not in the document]
amotin: OK.

sys/dev/isp/isp.c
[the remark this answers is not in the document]
amotin: I haven't read the code closely to be sure loaded_fw is always set for the later HBAs.  If that is the case, it may be work OK and only slightly confusing logically.

sys/dev/isp/isp.c
amotin: I would remove the newline after "return".

sys/dev/isp/isp.c
amotin: style(9) explicitly tells "Do not use ! for tests unless it is a boolean".  `!`, `!!`, etc are considered Linux'isms.

sys/dev/isp/ispvar.h
amotin: To be safe it should be like this (see additional parentheses around `code`):
```
#define ISP_FW_MAJOR(code)              (((code) >> 16) & 0xff)
```

sys/dev/isp/isp.c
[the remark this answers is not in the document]
jpulz: > t would not change anything, since both WARN and ERR are enabled by default.

You're right. So
```
			isp_prt(isp, ISP_LOGWARN,
			    "Unable to load ispfw(4) firmware %s", fwname);
```
or better hide this behind `LOGDEBUG0`?
As we don't talk about an error anymore, those lines above
```
			snprintf(fcp->fw_version_ispfw,
			    sizeof(fcp->fw_version_ispfw), "load error");
```
can be removed to, as `fcp->fw_version_ispfw` is already set to `not loaded` which is still valid.

sys/dev/isp/isp.c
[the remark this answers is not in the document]
jpulz: Until now, `loaded_fw` would only be set for **24xx and 25xx** when an ispfw(4) was actually loaded. In this case a `MBOX_VERIFY_CHECKSUM` was issued first and a `MBOX_EXEC_FIRMWARE` right after it with `mbs.param[3]` set based on the value of `loaded_fw`.
For 26xx+ there was no ispfw(4) firmware until now. so `loaded_fw` is always `0` and the code explicitely issued a `MBOX_LOAD_FLASH_FIRMWARE`. As `MBOX_VERIFY_CHECKSUM` is hidden behind a `if (loaded_fw)` this was not issued for 26xx+ and the `MBOX_EXEC_FIRMWARE` got issued without setting `mbs.param[3]` to anything at all.

sys/dev/isp/ispvar.h
jpulz: So
```c
#define ISP_FW_REV(maj, min, mic)       (((maj) << 16) | ((min) << 8) | (mic))
#define ISP_FW_MAJOR(code)              (((code) >> 16) & 0xff)
#define ISP_FW_MINOR(code)              (((code) >> 8) & 0xff)
#define ISP_FW_MICRO(code)              ((code) & 0xff)
#define ISP_FW_REVX(xp)                 (((xp[0]) << 16) | ((xp[1]) << 8) | (xp[2]))
```
then?

sys/dev/isp/isp.c
jpulz: Ok

sys/dev/isp/ispvar.h
amotin: In the last case parentheses should be around `xp`, not `xp[0]`.

jpulz: Ok, got it now.

sys/dev/isp/isp.c
```diff
@@ -4662,15 +4858,468 @@ isp_parse_flt_2xxx(ispsoftc_t *isp, uint8_t *flt_data)
...
+		isp_prt(isp, ISP_LOGERR,
+		    "Unable to verify the integrity of flash firmware image.");
+		isp_prt(isp, ISP_LOGERR,
+		    "Firmware data: 0x%08x 0x%08x 0x%08x 0x%08x.",
+		    dcode[0], dcode[1], dcode[2], dcode[3]);
+		return (ISP_FUNCTION_FAILED);
+	} else {
+		for (i = 0; i < 4; i++)
+			fcp->fw_flashrev[i] =
+				be32toh((uint32_t)dcode[4 + i]);
```
amotin, line 5210: Why do you need a type conversion here? It is already uint32_t after `[]`.

sys/dev/isp/isp.c
amotin: The bit manipulations look weird, but a bit better.  You do not need one set of parentheses in the first line.

sys/dev/isp/isp.c
jpulz: You're right again. Found two more of those and fixed it here.

sys/dev/isp/isp.c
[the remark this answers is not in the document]
jpulz: I changed this back to be spec conform.
I've tested this with and without the change and could not see any difference.

sys/dev/isp/isp.c
jpulz: Tested this with a 26xx based adpater and you (and the specs) are right.
Changed it to `IS_26XX()`

sys/dev/isp/isp.c
jpulz: fixed

sys/dev/isp/isp.c
jpulz: fixed

sys/dev/isp/isp.c
```diff
@@ -434,9 +561,27 @@ isp_reset(ispsoftc_t *isp, int do_load_defaults)
 	mbs.param[2] = code_org;
 	if (!IS_26XX(isp))
 		mbs.param[3] = loaded_fw ? 0 : 1;
+	mbs.param[4] = 0;
+	mbs.param[11] = 0;
+	if (IS_27XX(isp))
+		mbs.param[4] |= 0x08;	/* NVME_ENABLE_FLAG */
```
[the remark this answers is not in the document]
jpulz: You mean like this - only set this for 27xx and up and ignore for all others?
```c
        if (!IS_26XX(isp))
                mbs.param[3] = loaded_fw ? 0 : 1;
        if (IS_27XX(isp))
                mbs.param[4] |= 0x08;   /* NVME_ENABLE_FLAG */
        mbs.param[11] = 0;
```

sys/dev/isp/isp.c
```diff
@@ -395,32 +518,36 @@ isp_reset(ispsoftc_t *isp, int do_load_defaults)
...
 		mbs.param[2] = code_org;
 		isp_mboxcmd(isp, &mbs);
 		if (mbs.param[0] != MBOX_COMMAND_COMPLETE) {
-			isp_prt(isp, ISP_LOGERR, dcrc);
+			isp_prt(isp, ISP_LOGERR, "%s: 0x%x", dcrc,
+			    (mbs.param[2] << 16 | mbs.param[1]));
 			return;
 		}
+	} else {
+		if (IS_26XX(isp)) {
```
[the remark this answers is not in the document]
jpulz: Ok, will change it to match your suggestion.

sys/dev/isp/isp.c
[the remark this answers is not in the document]
amotin: You set mbs.param[4], then mbs.param[11] and then again mbs.param[4].  It is just weird.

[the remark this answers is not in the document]
jpulz: Ok, got it.

One last force push for those two changes.

markjdb:
Looks like this can be merged. Is there any reason to hold off?

amotin:
@markjdb Yea, should be merged. I was thinking to, but forgot. If somebody has a magic button to merge from here, please push it.

markjdb:
Done now, thanks @jpulz and @amotin.

bsdimp:
Why reopen?

markjdb:
> Why reopen?

I left a comment, which apparently appears only in the commit and not the PR: https://github.com/freebsd/freebsd-src/commit/10ed63fc06cb9902cc783ce8d0086c9aa97ed1e1#r135869967

Essentially, gcc12 found what looks like a minor bug.

markjdb:
> > Why reopen?
> 
> I left a comment, which apparently appears only in the commit and not the PR: [10ed63f#r135869967](https://github.com/freebsd/freebsd-src/commit/10ed63fc06cb9902cc783ce8d0086c9aa97ed1e1#r135869967)
> 
> Essentially, gcc12 found what looks like a minor bug.

I went ahead and fixed it in 91d2a093dfaccdf288db663b6e9d47539ea1c1bf.

jpulz:
On Tue, 2 Jan 2024, Mark Johnston wrote:

>>> Why reopen?
>>
>> I left a comment, which apparently appears only in the commit and not the PR: [10ed63f#r135869967](https://github.com/freebsd/freebsd-src/commit/10ed63fc06cb9902cc783ce8d0086c9aa97ed1e1#r135869967)
>>
>> Essentially, gcc12 found what looks like a minor bug.
>
> I went ahead and fixed it in 91d2a093dfaccdf288db663b6e9d47539ea1c1bf.

Mark,

thanks for taking care of this PR and for the fix.

Joerg

-- 
The beginning is the most important part of the work.
 				-Plato

PR #878, opened 2023-10-27
Update the GitHub actions

zxombie:
Add stable/14 to the pushed branches to build and allow manually running actions.

For manually run actions the `workflow_dispatch` change needs to be in the default branch of a clone.


bsdimp (approved):
lgtm

PR #879, opened 2023-10-28
traceroute/traceroute6: Implement ECN bleaching detection

jlduran:
Given the importance of the Explicit Congestion Notification (ECN) field in low-latency applications, have traceroute/traceroute6 easily report if some routers in the path are _bleaching_ or _mangling_ this field by setting the `-E` flag.

Inspired by: Darwin

---

Unsure about how to proceed with the code in contrib. It is already different from upstream.

Using checkstyle9.pl, fixed only a subset of the style(9) violations, in order to avoid a larger diff.


sys/netinet/ip6.h
```diff
@@ -97,10 +97,12 @@ struct ip6_hdr {
 #if BYTE_ORDER == BIG_ENDIAN
 #define IPV6_FLOWINFO_MASK	0x0fffffff	/* flow info (28 bits) */
 #define IPV6_FLOWLABEL_MASK	0x000fffff	/* flow label (20 bits) */
+#define IPV6_FLOW_ECN_MASK	0x00300000	/* the 2 ECN bits */
```
jlduran, line 100: I wonder if we should go with the "FLOW" and name it `IPV6_FLOWECN_MASK` instead?

markjdb:
Seems reasonable. Are we sure that the modifications in Darwin are released under the same license as the original code?

I don't see a description of ECN bleaching anywhere. It'd be nice to briefly explain it in the man pages.

It appears our traceroute doesn't track upstream anymore; we should probably move it out of contrib, but that's a separate matter. I think modifying it as you did is ok.

sys/netinet/ip6.h
markjdb: I suspect that compatibility is a bit more useful than consistency here.

markjdb: ... which is to say that I think it's ok as-is. :)

jlduran:
> Seems reasonable. Are we sure that the modifications in Darwin are released under the same license as the original code?

Hmm, not sure, I believe so, given the original license... https://github.com/apple-oss-distributions/network_cmds/commit/c1efe97740c238a20ed1e4bf7fbc41e1a70203fc#diff-cf63b30829ee1ebc11a69178e0517df298151227ac017663ae68e4185ee72221

> I don't see a description of ECN bleaching anywhere. It'd be nice to briefly explain it in the man pages.

Yes, I'll take a stab at it. (I'll paraphrase https://datatracker.ietf.org/doc/html/rfc8087#section-3.5).

> It appears our traceroute doesn't track upstream anymore; we should probably move it out of contrib, but that's a separate matter. I think modifying it as you did is ok.

Nice! Maybe a unification of the code (traceroute/traceroute6) as well?

markjdb:
> > Seems reasonable. Are we sure that the modifications in Darwin are released under the same license as the original code?
> 
> Hmm, not sure, I believe so, given the original license... [apple-oss-distributions/network_cmds@c1efe97#diff-cf63b30829ee1ebc11a69178e0517df298151227ac017663ae68e4185ee72221](https://github.com/apple-oss-distributions/network_cmds/commit/c1efe97740c238a20ed1e4bf7fbc41e1a70203fc#diff-cf63b30829ee1ebc11a69178e0517df298151227ac017663ae68e4185ee72221)

Right, so these modifications are licensed under the APSL, which I think has some negative implications per https://docs.freebsd.org/en/articles/license-guide/ . That doc doesn't quite make it clear whether APSL-licensed code can be added to userspace.

It's not 100% clear to me. Maybe @bsdimp or @emaste can clarify the project's stance here? Maybe I'm misunderstanding something.

jlduran:
OK, I see... I'll wait for an explanation to close or update this PR.
It is only a nice-to-have, but not essential. Thank you!

bsdimp:
```
				if (ecn) {
					u_char input_ecn = hip->ip_tos & IPTOS_ECN_MASK;
					u_char output_ecn = outip->ip_tos & IPTOS_ECN_MASK;

					if (input_ecn == output_ecn) {
						Printf(" (ecn=passed)");
					} else if (input_ecn == IPTOS_ECN_CE) {
						Printf(" (ecn=mangled)");
					} else if (input_ecn == IPTOS_ECN_NOTECT) {
						Printf(" (ecn=bleached)");
					}
```
RFC -3168 however uses different terms. 01 and 10 are ECN(0) and ECN(1) code points. 00 is NOTECT and section 5 uses the term "erase" (in quotes like that) rather than bleached. And that 11 is 'CE' meaning congestion exists to the end node. This suggests that we should report ecn=congested. There's one other possiblity, where the router swaps ECN(0) to ECN(1) or vice versa that currently prints nothing. The older RFC-2481 may have used different terms, which is what the Apple code might be interpreting. RFC-3168 dates from 2001, so I'm not sure how relevant the older bits are. I also think that there needs to be an else "ecn=swapped" when ECN goes from ECN(0) to ECN(1).

https://datatracker.ietf.org/doc/html/rfc3168 is a handy reference.

So we may need to rework this such that the original license is no longer relevant.

jlduran:
Second iteration:
- Incorporate some of the suggestions.
- The code is now different from Darwin. I've even found a few minor bugs that I need to report (red tape alert).
- I left the terms _bleached_ and _mangled_ [^1] [^2].
- Added _congestion_, in the rare event we receive `11`.
- Explain a bit more how it works in the manual page.
- Add a verbose commit message.

[^1]: https://datatracker.ietf.org/doc/html/rfc8087#section-3.5
[^2]: https://arxiv.org/pdf/2208.14523.pdf

bsdimp:
Thanks for the pointers. It's helped me update my understanding. And yes, I agree with them. I like the new commit messages as well...

bsdimp (approved):
I think we may have a winner. @markjdb ?

jlduran:
- Use the past tense (`congestion` -> `congested`)

markjdb (commented):
Mostly looks good to me, thank you. I left some comments on the traceroute6.c patch.

usr.sbin/traceroute6/traceroute6.c
```diff
@@ -584,6 +590,11 @@ main(int argc, char *argv[])
 		exit(1);
 	}
 
+	if (ecn) {
+		tclass = 0;
+		tclass |= IPTOS_ECN_ECT1;
```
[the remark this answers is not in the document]
jlduran: Hmmm... Thinking about it, I do _not_ want them to be mutually exclusive, but rather complementary:
- For IPv4, I want to keep the DSCP bits, and override the ECN ones
- For IPv6, I want to keep the DS bits, and override the ECN ones

But, yes, this has to be documented in the man page. Nice catch!

jlduran:
Third iteration:
- Address style suggestions
- Make the `-E` flag be able to work together with `-t`, the classic example is when debugging ECN bleaching for traffic that gets routed differently based on its classification. Also document this possibility.
- Return the ECN value explicitly in `packet_ok()`.

jlduran:
- Prefer a switch over to if/else.

contrib/traceroute/traceroute.8
```diff
@@ -66,6 +66,22 @@ default.
...
+and report if the hop has bleached
+.Pq Dv 00
+or mangled
+.Pq Dv 10
+them, or if it is experiencing congestion
+.Pq Dv 11 .
+Otherwise, report that it passed the bits appropriately.
+If
+.Fl t
+is also specified, the corresponding ECN bits will be replaced.
```
markjdb, line 84: Why do you say "replaced"? From my reading of the code, if `-t` is also specified we OR the ECN flag into the corresponding IP header field.

jlduran: In the third iteration, I zero the ECN bits before `OR`ing:
- IPv4:
   https://github.com/freebsd/freebsd-src/blob/6c00b180fedf9c4896dbfe523a5bb5235ade299f/contrib/traceroute/traceroute.c#L793-L794
- IPv6:
   https://github.com/freebsd/freebsd-src/blob/6c00b180fedf9c4896dbfe523a5bb5235ade299f/usr.sbin/traceroute6/traceroute6.c#L593-L598

The final crafted DSCP+ECN (originally ToS) (IPv4) or Traffic Class (IPv6) field when `-E` is passed should always be:
```
+-+-+-+-+-+-+-+-+
| Traffic Class |
+-+-+-+-+-+-+-+-+
|x x x x x x|0 1|
+-+-+-+-+-+-+-+-+
```

markjdb: I get it now, thank you.

markjdb (approved):
Looks good to me, thank you!

markjdb:
@bsdimp feel free to merge this if you're happy with the updates. I can merge as well.

markjdb:
Done!

jlduran:
Thank you!

PR #880, opened 2023-10-30
Update regdomain.xml

JonathanDLee24:
I wanted to bring this to your attention. I noticed and learned that the regdomain.xml file shows very low values for 11ng and 11na. This file sets the max power range from as low as 17 to a max value of 30. Is this of concern to lower the max power on WiFi cards to such a low value? I assume this limits use of this strictly to studio apartments. I myself have owned many AP devices that gave the consumer greater control over the transmit and receive power. Understandably many APs give consumers a level of control for AP power levels. The power levels in many other AP systems allow consumer control, why has this been limited to such a low value? FCC controls the frequency range correct. 17 percent seems like it has been degraded for a unknown or accidental reason. . .

17 percent of max power utilization to me seems to really limit the WiFi functionality. Does anyone else agree this seems a bit to low?

With 40 it has much more coverage.

### DIFF
--- a/lib/lib80211/regdomain.xml
+++ b/lib/lib80211/regdomain.xml
@@ -47,42 +47,42 @@
   <netband mode="11b">
     <band>
       <freqband ref="F1_2412_2462"/>
-      <maxpower>30</maxpower>
+      <maxpower>40</maxpower>
       <flags>IEEE80211_CHAN_B</flags>
     </band>
   </netband>
   <netband mode="11g">
     <band>
       <freqband ref="F1_2412_2462"/>
-      <maxpower>30</maxpower>
+      <maxpower>40</maxpower>
       <flags>IEEE80211_CHAN_G</flags>
     </band>
   </netband>
   <netband mode="11a">
     <band>
       <freqband ref="F1_5180_5240"/>
-      <maxpower>17</maxpower>
+      <maxpower>40</maxpower>
     </band>
     <band>
       <freqband ref="F1_5745_5805"/>
-      <maxpower>23</maxpower>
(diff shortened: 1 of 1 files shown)


JonathanDLee24:
17 percent of max power utilization to me seems to really limit the WiFi functionality. Does anyone else agree this seems a bit to low?

lwhsu:
cc @bzfbd @erikarn

bsdimp:
It would be useful to find another source of truth for these values to cross che k against.

jlduran:
The current values seem to be in line with:
https://www.ecfr.gov/current/title-47/chapter-I/subchapter-A/part-15/subpart-E/section-15.407
There could be a bug somewhere else in the calculations, but I don't think we should change the current values.

bzfbd:
Cc: @ccfbsd

JonathanDLee24:
Thanks everyone, I had so many issues with very low signal in pfsense with the wifi card. Not many people use the wifi card they like the external AP because of the issues that they have. I couldn't give up on it.

bsdimp:
I think this should be closed. If that's in error please reopen with an explanation. Thanks!

JonathanDLee24:
No close this I was hoping to get some more visibility on the need for WiFi improvements. I am running this now on pfSense 23.05.01 and the wifi is working fine.

bsdimp:
Per request reopened. I'll want bz's 2 cents at least though before it's ready

bzfbd:
> Per request reopened. I'll want bz's 2 cents at least though before it's ready

I thought he said "No, close this." without the punctuation.

PR #881, opened 2023-10-31
adduser(8): support creation of ZFS dataset

jgrafton:
On systems utilizing ZFS, default to creating a ZFS dataset for a new user's home directory if the parent directory resides on a ZFS dataset. Add a flag that disables this behavior if the administrator explicitly does not want it.

If run during installation from within a chroot, set mountpoint to legacy after dataset creation and mount directly into the chroot.  Then umount and reset the mountpoint to inherit from parent. 

Also support ZFS default encryption on user's home directory.

Requested here: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=263234

### DIFF
--- a/usr.sbin/adduser/adduser.8
+++ b/usr.sbin/adduser/adduser.8
@@ -24,7 +24,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd September 15, 2012
+.Dd January 19, 2024
 .Dt ADDUSER 8
 .Os
 .Sh NAME
@@ -32,7 +32,7 @@
 .Nd command for adding new users
 .Sh SYNOPSIS
 .Nm
-.Op Fl CDENShq
+.Op Fl CDENSZhq
 .Op Fl G Ar groups
 .Op Fl L Ar login_class
 .Op Fl M Ar mode
@@ -52,6 +52,10 @@ utility is a shell script, implemented around the
 command, for adding new users.
 It creates passwd/group entries, a home directory,
 copies dotfiles and sends the new user a welcome message.
+On systems where the parent of home directory is a ZFS dataset,
+.Nm
+will create the home directory as a ZFS dataset by default,
+unless the system administrator specified otherwise.
 It supports two modes of operation.
 It may be used interactively
 at the command line to add one user at a time, or it may be directed
@@ -295,6 +299,8 @@ In addition, it will be available for inclusion in the message file in the
 .Va randompass
 variable.
 .El
+.It Fl Z
+Do not attempt to create ZFS home dataset.
 .El
 .Sh FORMAT
 When the
--- a/usr.sbin/adduser/adduser.conf.5
+++ b/usr.sbin/adduser/adduser.conf.5
@@ -171,6 +171,8 @@ The default information to be held in the GECOS field of
 .It Va uidstart
 The default user ID setting.
 This must be a number above 1000 and fewer than 65534.
+.It Va Zflag
+Do not attempt to create ZFS home dataset.
 .El
 .Sh EXAMPLES
 The following is an example
--- a/usr.sbin/adduser/adduser.sh
+++ b/usr.sbin/adduser/adduser.sh
@@ -84,6 +84,7 @@ show_usage() {
 	echo "  -L		login class of the user"
 	echo "  -M		file permission for home directory"
 	echo "  -N		do not read configuration file"
+	echo "  -Z		do not attempt to create ZFS home dataset"
 	echo "  -S		a nonexistent shell is not an error"
 	echo "  -d		home directory"
 	echo "  -f		file from which input will be received"
@@ -283,6 +284,22 @@ add_user() {
 		;;
 	esac
 
+	# create ZFS dataset before home directory is created with pw
+	if [ "${Zcreate}" = "yes" ]; then
+		if [ "${Zencrypt}" = "yes" ]; then
+			echo "Enter encryption keyphrase for ZFS dataset (${zhome}):"
+		fi
+		if [ -n "$BSDINSTALL_CHROOT" ]; then
+			create_zfs_chrooted_dataset
+		else
+			create_zfs_dataset
+			if [ "$?" -ne 0 ]; then
+				err "There was an error adding user ($username)."
+				return 1
+			fi
+		fi
+	fi
+
 	_pwcmd="$_upasswd ${PWCMD} useradd $_uid $_name $_group $_grouplist $_comment"
 	_pwcmd="$_pwcmd $_shell $_class $_home $_dotdir $_passwdmethod $_passwd"
 	_pwcmd="$_pwcmd $_expire $_pwexpire"
@@ -303,6 +320,14 @@ add_user() {
 			info "Account ($username) is locked."
 		else
 			info "Account ($username) could NOT be locked."
+		fi
+	fi
+
+	# give newly created user permissions to their home zfs dataset
+	if [ "${Zcreate}" = "yes" ]; then
+		set_zfs_perms
+		if [ -n "$BSDINSTALL_CHROOT" ]; then
+			umount_legacy_zfs
 		fi
 	fi
 
@@ -479,6 +504,26 @@ get_homeperm() {
 	fi
 }
 
+# get_zfs_home
+#	Determine if homeprefix is located on a ZFS filesystem and if
+#	so, enable ZFS home dataset creation.
+#
+get_zfs_home() {
+	# check if zfs kernel module is loaded before attempting to run zfs to 
+	# prevent loading the kernel module on systems that don't use ZFS
+	if ! "$KLDSTATCMD" -q -m zfs; then
+		Zcreate="no"
+		return
+	fi
+	zfs_homeprefix=`${ZFSCMD} list -Ho name "${homeprefix}" 2>/dev/null`
+	if [ "$?" -ne 0 ]; then
+		Zcreate="no"
+	elif [ -z "${zfs_homeprefix}" ]; then
+		Zcreate="no"
+	fi
+	zhome="${zfs_homeprefix}/${username}"
+}
+
 # get_uid
 #	Reads a numeric userid in an interactive or batch session. Automatically
 #	allocates one if it is not specified.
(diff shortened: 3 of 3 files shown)


usr.sbin/adduser/adduser.8
```diff
@@ -52,6 +53,8 @@ utility is a shell script, implemented around the
 command, for adding new users.
 It creates passwd/group entries, a home directory,
 copies dotfiles and sends the new user a welcome message.
+It also supports creating a ZFS dataset for the user's
```
delphij, line 56:
```
On systems where the parent of home directory is a dataset,
.Nm
will create the home directory as a ZFS dataset by default,
unless the system administrator specified otherwise.
```

usr.sbin/adduser/adduser.8
```diff
@@ -295,6 +298,19 @@ In addition, it will be available for inclusion in the message file in the
 .Va randompass
 variable.
 .El
+.It Fl z Ar dataset
+Create home directory on ZFS 
+.Ar dataset .
+.It Fl Z
+Set ZFS creation flag to yes.  The
```
delphij, line 305: The default behavior should be to create ZFS dataset unless 1) the `homeprefix` is not a ZFS, or 2) administrator explicitly says no by specifying a command line option, or by an option set in `adduser.conf(5)`.

usr.sbin/adduser/adduser.conf.5
```diff
@@ -171,6 +171,14 @@ The default information to be held in the GECOS field of
 .It Va uidstart
 The default user ID setting.
 This must be a number above 1000 and fewer than 65534.
+.It Va Zflag
+Turn on ZFS support by default.
```
delphij, line 175: A negative option (inhibit use of ZFS) should be provided instead.

usr.sbin/adduser/adduser.conf.5
```diff
@@ -171,6 +171,14 @@ The default information to be held in the GECOS field of
...
+.It Va zfsopt
+Options passed into the 
+.Xr zfs-create 8
```
delphij, line 178: IMHO zfs create option should not be passed.  These datasets should inherit attributes from an upper level, especially compression, atime, etc.; encryption would be the only legitimate option to be passed.

usr.sbin/adduser/adduser.sh
```diff
@@ -197,11 +199,14 @@ save_config() {
 	echo "defaultgroups=$ugroups"	>> ${ADDUSERCONF}
 	echo "passwdtype=$passwdtype" 	>> ${ADDUSERCONF}
 	echo "homeprefix=$homeprefix" 	>> ${ADDUSERCONF}
+	echo "zfs_homeprefix=$zfs_homeprefix" 	>> ${ADDUSERCONF}
```
delphij, line 202: This should always be derived from `homeprefix` instead of passed from command line.

usr.sbin/adduser/adduser.sh
```diff
@@ -283,6 +288,11 @@ add_user() {
 		;;
 	esac
 
+	# create ZFS dataset if Zflag is set before home directory is created with pw
+	if [ -n "$Zflag" ]; then
```
delphij, line 292: This should be negative (opt-out) instead of opt-in.

Probably:

```
if [ -n "$(zfs list -o name "${homeprefix}" 2>/dev/null)" ]; then
```

(the output _may_ be used as `zfs_homeprefix` too).

usr.sbin/adduser/adduser.sh
```diff
@@ -306,6 +316,11 @@ add_user() {
 		fi
 	fi
 
+	# give newly created user permissions to their home zfs dataset
+	if [ -n "$Zflag" ]; then
+		set_zfs_perms
```
delphij, line 321: This should happen if the user directory is a ZFS dataset (maybe use a similar command like `zfs list` above).

usr.sbin/adduser/adduser.sh
```diff
@@ -890,6 +1020,7 @@ defaultLgroup=
 defaultgroups=
 defaultshell="${DEFAULTSHELL}"
 defaultHomePerm=
+zfsopt="-o compression=off -o atime=on"
```
delphij, line 1023: There should be *NO* default for creation options.  They should almost always be inherited from parent.

usr.sbin/adduser/adduser.sh
```diff
@@ -613,6 +665,47 @@ get_password() {
...
+	else
+		zsupported="yes"
+	fi
+}
+
+# create_zfs_dataset
+#   Create ZFS dataset owned by the user that was just added.
+#
+create_zfs_dataset() {
+	${ZFSCMD} create -p -o "mountpoint=${uhome}" ${zfsopt} "${zhome}"
```
delphij, line 687: DO NOT set mountpoint, unless it's absolutely necessary.  This would complicate replication to a different location (because ZFS would copy mountpoint to a foreign ZFS dataset).

delphij (changes requested):

usr.sbin/adduser/adduser.8
jgrafton: Thanks for the detailed review, @delphij.  My thought was to allow the administrator to specify both the zfs dataset and home directory separately and map the home directory onto the dataset by setting the mountpoint.  What you're saying is to simplify it by creating a zfs dataset only if the homeprefix sits on an existing dataset and provide an opt-out flag if the admin doesn't want zfs datasets?

usr.sbin/adduser/adduser.sh
jgrafton: In that case, the administrator would be required to pass in a homeprefix that is already mapped to a zfs dataset.  In your opinion, should there be an option where the script is able to create a dataset that's not mapped onto an existing location?

jgrafton:
I simplified the code with your suggestions @delphij.  This revision just adds a dataset if the parent resides on a dataset like you mentioned.  I'll get to work on adding encryption in a later revision.  Thanks again!

jgrafton:
Discovered bug during installation in chrooted environment.  The skeleton data for the user's home directory is copied into the underlying user directory instead of the mounted dataset.  Marking this draft until I resolve the issue.

usr.sbin/adduser/adduser.sh
```diff
@@ -202,6 +203,8 @@ save_config() {
 	echo "msgfile=$msgfile"		>> ${ADDUSERCONF}
 	echo "disableflag=$disableflag" >> ${ADDUSERCONF}
 	echo "uidstart=$uidstart"       >> ${ADDUSERCONF}
+	echo "Zflag=$Zflag"       >> ${ADDUSERCONF}
```
kevans91, line 206: I'd maybe call this `zfscreate=no`, just to be a little more obvious from the name

usr.sbin/adduser/adduser.sh
```diff
@@ -202,6 +203,8 @@ save_config() {
...
+	echo "Zencrypt=$Zencrypt"       >> ${ADDUSERCONF}
```
kevans91, line 207: I don't see a manpage addition for `Zencrypt`, but with a more descriptive name for `Zflag` you could make that instead a tri-state: `zfscreate=yes|no|encrypted`

usr.sbin/adduser/adduser.sh
```diff
@@ -479,6 +506,20 @@ get_homeperm() {
 	fi
 }
 
+# get_zfs_home
+#	Determine if homeprefix is located on a ZFS filesystem and if
+#	so, enable ZFS home dataset creation.
+#
+get_zfs_home() {
```
kevans91, line 513: This seems like it would override a configured `Zflag=no`, but I'm also not sure I understand the logic; `zfs list` would exit 0 if it found a prefix, and `-z "${zfs_homeprefix}` would flip Zflag on if we didn't detect a prefix.

There's an additional caveat that using `zfs` will trigger a load of zfs.ko on purely ufs systems; I'd recommend bailing out before calling it if `kldstat -q -m zfs` indicates that zfs isn't there

usr.sbin/adduser/adduser.sh
```diff
@@ -613,6 +654,84 @@ get_password() {
...
+		echo -n "$_prompt"
+		read _input
+
+		[ -z "$_input" ] && _input=Zencrypt
+		case $_input in
+		[Nn][Oo]|[Nn])
+			Zencrypt=
+			break
+			;;
+		[Yy][Ee][Ss]|[Yy][Ee]|[Yy])
```
kevans91, line 673: I think accepting "ye" is a little nonstandard; this should probably just be a case-insensitive "yes" or case-insensitive "y"

usr.sbin/adduser/adduser.sh
```diff
@@ -613,6 +654,84 @@ get_password() {
...
+#	Read ZFS options for the user's home dataset.
+#
+get_zfs_encryption() {
+	_input=
+	_prompt="Enable ZFS encryption? (yes/no) [no]: "
+	while : ; do
+		echo -n "$_prompt"
+		read _input
+
+		[ -z "$_input" ] && _input=Zencrypt
```
kevans91, line 667: The prompt describes "no" as the default, but this seems to be trying to use the current value of `Zencrypt` as the default? (But it should have been `$Zencrypt`... this should probably be `no` to match the prompt, though)

usr.sbin/adduser/adduser.sh
```diff
@@ -613,6 +654,84 @@ get_password() {
...
+		zfsopt="-o encryption=on -o keylocation=prompt -o keyformat=passphrase"
+	fi
+}
+
+# create_zfs_chrooted_dataset
+#   Create ZFS dataset owned by the user that was just added within a bsdinstall chroot
+#
+create_zfs_chrooted_dataset() {
+	${ZFSCMD} create -u ${zfsopt} "${zhome}"
+	if [ "$?" -ne 0 ]; then
```
kevans91, line 694: More concisely written as:

```
if ! ${ZFSCMD} create ...
```

usr.sbin/adduser/adduser.sh
```diff
@@ -613,6 +654,84 @@ get_password() {
...
+umount_legacy_zfs() {
+	${UMOUNTCMD} "${uhome}"
+	${ZFSCMD} inherit mountpoint "${zhome}"
+}
+
+# create_zfs_dataset
+#   Create ZFS dataset owned by the user that was just added.
+#
+create_zfs_dataset() {
+	${ZFSCMD} create ${zfsopt} "${zhome}"
```
kevans91, line 715: Ditto above; you're not using the exit code otherwise, might as well just test it directly

usr.sbin/adduser/adduser.sh
```diff
@@ -613,6 +654,84 @@ get_password() {
...
+	else
+		info "Successfully created ZFS dataset (${zhome})."
+	fi
+}
+
+# set_zfs_perms
+#   Give new user ownership of newly created zfs dataset.
+#
+set_zfs_perms() {
+	${ZFSCMD} allow "${username}" create,destroy,mount,snapshot "${zhome}"
```
kevans91, line 728: Ditto above w/ exit code

usr.sbin/adduser/adduser.sh
kevans91: Oh, right, Zflag means "no zfs dataset"... the name is definitely getting me here.

kevans91 (changes requested):

usr.sbin/adduser/adduser.sh
jgrafton: That's a bug, fixed.

usr.sbin/adduser/adduser.sh
jgrafton: I agree, it's a bit strange.  However, I followed the user input convention already used in the script.

usr.sbin/adduser/adduser.sh
jgrafton: Good call.  I changed both Zencrypt and Zcreate to yes/no bools to remove confusion.

usr.sbin/adduser/adduser.sh
jgrafton: You're right, Zflag is super confusing!  I set Zcreate and Zencrypt to yes/no bools but didn't implement the tri-state since the two booleans seem to "flow" better in the script.

usr.sbin/adduser/adduser.sh
kevans91: Huh, indeed- funky

kevans91 (approved):
I think this LGTM- thanks!

jgrafton:
> I think this LGTM- thanks!

Thanks Kyle!

igalic:
@kevans91 and go ahead and commit this?

usr.sbin/adduser/adduser.sh
```diff
@@ -283,6 +284,22 @@ add_user() {
 		;;
 	esac
 
+	# create ZFS dataset before home directory is created with pw
+	if [ "${Zcreate}" = "yes" ]; then
+		if [ "${Zencrypt}" = "yes" ]; then
+			echo "Enter encryption keyphrase for ZFS dataset (${zhome}):"
```
bsdimp, line 290: This doesn't look like it is prompting for anything or passing the results along, why not?

bsdimp:
Just a nit an a quick question

bsdimp:
merged. Thanks!

PR #882, opened 2023-11-01
Typo

ElyesH:


ElyesH:
ping

L4STADAM (approved):

bsdimp:
Looks like many of these have been pushed.
Since the remainder do contain user-visible changes, I'm pushing them.
However, there's many non-visble changes (eg just to comments) that I'm only merging because I'm doing the others.
We've recently updated our policies to avoid most typo changes that aren't user-visible.
So I'm landing the rest.
c384235c1633

PR #886, opened 2023-11-03
Remove return statement at the end of void functions

ElyesH:


mhorne (approved):

emaste:
FYI we have a `Pull request:` trailer that we typically use to link back to the PR, and also a `merged` label that I've now added.

PR #890, opened 2023-11-07
Remove return statement at the end of void functions

ElyesH:


bsdimp:
Kirk should review this.
Looks fine to my eye

igalic:
@ElyesH please do a rebase onto main, rather than merging main into your branch

bsdimp:
OK. There's little value provided by these changes. It falls under the don't submit wide-ranging cleanups rule that we've started enforcing. IF you are working in one of these areas, and want to submit an appropriately scoped patch based on that, we'll consider it then. But there's too many different things touched, and the value added by this cleanup is relatively low.

PR #892, opened 2023-11-08
tzset.3: Remove references to posixrules

jlduran:
Partial references to the file /usr/share/zoneinfo/posixrules are removed.  Other man pages under contrib still reference to this file, gone in 783c318fd1181d46554c31a5039db10e7f5eef49.

---

I am actually inclined to partially revert 783c318fd1181d46554c31a5039db10e7f5eef49, and bring back the file. However, I'm not familiar with the decisions taken in the first place.  I'm just trying to debug a [failing i386 test](https://ci.freebsd.org/view/all/job/FreeBSD-stable-14-i386-test/153/testReport/junit/usr.bin.file/file_test/contrib_file_tests/) (that may not be important at all)

### DIFF
--- a/lib/libc/gen/tzset.3
+++ b/lib/libc/gen/tzset.3
@@ -30,7 +30,7 @@
 .\"
 .\"	@(#)tzset.3	8.2 (Berkeley) 11/17/93
 .\"
-.Dd March 6, 2023
+.Dd November 8, 2023
 .Dt TZSET 3
 .Os
 .Sh NAME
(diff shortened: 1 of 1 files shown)


bsdimp:
I wonder why des thinks that this is unnecessary... 
The posix rules are a standard thing, no?

jlduran:
@bsdimp Other BSDs still keep this file (may not be relevant in this case), I don't know the reasons behind its removal. I was planning on digging up a little more (maybe @dag-erling could elucidate here a little bit?)

bsdimp:
I'm going to close this for now. This isn't ripe for committing just yet. We can reopen or create a new one once the discussions about why they were removed happen.

PR #893, opened 2023-11-09
{cut,diff,head,join,patch,xargs}: use getline() instead of fgetln()

arp242:
This replaces various uses of fgetln() with getline(). The main reason for this is portability, making things easier for people who want to compile these tools on non-FreeBSD systems.

I appreciate that's probably not the top concern for FreeBSD base tools, but fgetln() is impossible to port to most platforms, as concurrent access is essentially impossible to implement fully correct without the line buffer on the FILE struct. Other than this, many generic FreeBSD tools compile fairly cleanly on Linux with a few small changes.

Most uses of fgetln() pre-date getline() support (added in 2009 with 69099ba2ec8b), and there's been some previous patches (ee3ca711a898 8c98e6b1a7f3 1a2a4fc8ce1b) for other tools.

Obtained from:	https://github.com/dcantrell/bsdutils and
              	https://github.com/chimera-linux/chimerautils
Signed-off-by: Martin Tournoij <martin@arp242.net>

### DIFF
--- a/usr.bin/cut/cut.c
+++ b/usr.bin/cut/cut.c
@@ -275,14 +275,15 @@ b_cut(FILE *fp, const char *fname __unused)
 static int
 b_n_cut(FILE *fp, const char *fname)
 {
-	size_t col, i, lbuflen;
-	char *lbuf;
+	size_t col, i, bufsize = 0;
+	ssize_t lbuflen;
+	char *lbuf = NULL;
 	int canwrite, clen, warned;
 	mbstate_t mbs;
 
 	memset(&mbs, 0, sizeof(mbs));
 	warned = 0;
-	while ((lbuf = fgetln(fp, &lbuflen)) != NULL) {
+	while ((lbuflen = getline(&lbuf, &bufsize, fp)) >= 0) {
 		for (col = 0; lbuflen > 0; col += clen) {
 			if ((clen = mbrlen(lbuf, lbuflen, &mbs)) < 0) {
 				if (!warned) {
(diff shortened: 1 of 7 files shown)


igalic (changes requested):
can you please split this into one commit for each sys system, one for cut, diff, head, join, patch, xargs

and add to their commit message a reference to this pull request:

```
Pull Request: https://github.com/freebsd/freebsd-src/pull/893
```

arp242:
Sure – I changed it!

bsdimp:
So these changes look good.

I'm none-the-less nervous about them. What is the useage of the altenrative bsd tools that you pointed me to? Have you run the kyua tests on these programs before/after to make sure that there's no subtle regressions?

arp242:
> Have you run the kyua tests on these programs before/after to make sure that there's no subtle regressions?

I had assumed the CI would run this, but I gather it doesn't?

I've had great difficulty actually running the atf/kyua tests; it's been a while and I don't quite recall if I managed to get it working – I just remember a great deal of frustration.

bsdimp:
> > Have you run the kyua tests on these programs before/after to make sure that there's no subtle regressions?
> 
> I had assumed the CI would run this, but I gather it doesn't?
> 
> I've had great difficulty actually running the atf/kyua tests; it's been a while and I don't quite recall if I managed to get it working – I just remember a great deal of frustration.

(sorry missed this) I'll see if I can run kyua for these tests to see if there's no change.

bsdimp:
ci run results: all specific tests in /usr/tests/usr.sbin/{head,cut,diff,patch,join,xargs} pass

bsdimp:
5fbdcd65fe5c pushed

PR #896, opened 2023-11-13
sort OptionalObsoleteFiles blocks

brooksdavis:
OptionalObsoleteFiles.inc was almost, but not quite sorted which made it hard to figure out where to put things. There seems to have been a mix of more sorting errors, things that didn't move when the option controlling them changed, and odd choices for locations.

 I've broken this up into a set of commits moving one or two blocks at a time in hopes that if people have merge difficulty they will be able to merge individual changes where it's easy to see what the conflict is.


bsdimp:
A quick look suggests this is sane.
Do we document the desired order at the top?

brooksdavis:
> A quick look suggests this is sane. Do we document the desired order at the top?

I've added a comment to document the order (it's unfortunately awkward to enforce, probably requiring awk

tools/build/mk/OptionalObsoleteFiles.inc
```diff
@@ -1659,16 +1399,167 @@ OLD_FILES+=usr/share/dict/web2a
...
-OLD_FILES+=etc/dma/dma.conf
-OLD_DIRS+=etc/dma
-OLD_FILES+=usr/libexec/dma
-OLD_FILES+=usr/libexec/dma-mbox-create
-OLD_FILES+=usr/share/man/man8/dma.8.gz
-OLD_FILES+=usr/share/examples/dma/auth.conf
-OLD_FILES+=usr/share/examples/dma/mailer.conf
-OLD_DIRS+=usr/share/examples/dma
+
+.if ${MK_DMAGENT} == no
```
emaste: what happened here?

netchild (approved):

netchild (approved):

netchild (approved):

netchild (approved):

tools/build/mk/OptionalObsoleteFiles.inc
netchild: The old file doesn't have superfluous whitespace, and the new one doesn't seem to have either...

emaste (approved):

PR #898, opened 2023-11-14
${CFLAGS:N-flto} -> ${CFLAGS:N-flto*}

Hello71:
same reason as the original https://reviews.freebsd.org/D9659: -flto=<N>, -flto=full, and -flto=thin also produce the GIMPLE/bitcode which is not supported by genassym, so filter those out as well.


igalic:
cc @emaste

bsdimp:
lgtm. Can't hurt, might help.

PR #899, opened 2023-11-17
Rnda 271 Add printf code to stubbed functions

kbiczk-mobica:
Added printf code to stubbed functions


bsdimp:
This pull request looks like it got messed up. I can't find what's what from it. I can't find the code described in the description. It seems like a lot of extra linuxkpi files from OpenBSD, which while interesting on their own, aren't described in these commits.

I'm closing this pull request. Please sort out the bits you'd like to see in the tree and resubmit.

PR #902, opened 2023-11-19
Fix typos and add translation of tcsh signal names in greek

chrislongros:
My first FreeBSD src commit.


asomers:
Χαϊρε Χρίστε!  Thank you for your contribution.  But since FreeBSD is a downstream consumer of tcsh, I suggest you open a PR upstream instead.  You can do that here: https://github.com/tcsh-org/tcsh .

chrislongros:
> Χαϊρε Χρίστε! Thank you for your contribution. But since FreeBSD is a downstream consumer of tcsh, I suggest you open a PR upstream instead. You can do that here: https://github.com/tcsh-org/tcsh .

Γεια σου Alan. Thank you for your response! :)

PR #904, opened 2023-11-19
Fix typo of tcsh nls set 9 in greek

chrislongros:


igalic:
given that this is in Contrib, it needs to go upstream (first):  https://www.tcsh.org/ -> https://bugs.astron.com/my_view_page.php

chrislongros:
> given that this is in Contrib, it needs to go upstream (first): https://www.tcsh.org/ -> https://bugs.astron.com/my_view_page.php

Should I open a pull request here: https://github.com/tcsh-org/tcsh ?

igalic:
unfortunately, they have pull requests disabled, and issues seem to be ignored, so i would try https://bugs.astron.com/my_view_page.php

bsdimp:
So how does one submit patches upstream for this? If there is no effective upstream anymore, we make the changes locally...

asomers:
I don't know what @igalic is looking at.  Pull requests are definitely not disabled on tcsh's github page.

chrislongros:
> I don't know what @igalic is looking at. Pull requests are definitely not disabled on tcsh's github page.

I agree. I submitted my pull requests yesterday.

igalic:
yeah, sorry about that.

someone in GitHub's design team thought it was a good idea to hide pull requests on mobile, and I keep forgetting that.

PR #908, opened 2023-11-21
libsys

brooksdavis:
This is a partial reimplementation of https://reviews.freebsd.org/D14609 which aims to move the concrete implementations of the system call layer from libc to libsys to facilitate system call tracing, exploration of alternative system call interfaces, and perhaps hardening of the system to prevent libraries other than libsys from making system calls.

The implementation largely follows the linked review including linking libsys as a filter to libc (I've lost track of the record of why we've chosen that approach). It does not yet include the assembly refactoring to turn the libc stub versions into calls to abort2. I have moved auxv handling over to libsys as discussed in the linked review, but I'm of two minds about that because it adds several dependencies on string functions.

Some of the commits may seem odd, but I've tried hard to avoid commits that both move and change a non-trivial file as that breaks git blame and the history of many of these pieces is quite important. I've also tried to keep changes I'm less certain are a good idea in separate commits so as to make it easer to drop them later.

With this PR I'm looking to get:
 - reviews of the general idea
 - reactions to the structure of the changes both in terms of the final source tree and the commits
 - suggestions for things I need to fix or test


emaste (commented):
This patch LGTM

lib/libc/gen/getpagesize.c
```diff
@@ -29,45 +29,24 @@
...
 getpagesize(void)
 {
-	int mib[2];
-	static int value;
-	size_t size;
-	int error;
-
-	if (value != 0)
-		return (value);
+	int error, value;
```
emaste, line 45: init value to zero now that it is not in bss?

emaste:
s/git/it/ in `git has a much larger than typical dependency footprint``

brooksdavis:
> It seems I can't leave a review for just one patch in the series. I wanted to say that `getpagesizes(3): drop support for kernels before 9.0` LGTM and IMO could be committed by itself anytime.

I plan to pull a few of the first ones out for review in phab, but had promised someone I'd get the first cut of the PR out today so we can start talking about the shape.

lib/libc/gen/getpagesize.c
brooksdavis: Not necessary I think due to the `error != 0` fallback (which in practice is dead code)?

lib/libc/gen/getpagesize.c
```diff
@@ -29,45 +29,24 @@
  * SUCH DAMAGE.
  */
 
-#include <sys/cdefs.h>
-__SCCSID("@(#)getpagesize.c	8.1 (Berkeley) 6/4/93");
```
qemu-bsd-user, line 33: You could commit this anytime...

lib/libc/gen/getpagesizes.c
```diff
@@ -27,11 +27,8 @@
  */
 
 #include <sys/param.h>
-#include <sys/mman.h>
-#include <sys/sysctl.h>
-
-#include <errno.h>
-#include <link.h>
+#include <sys/auxv.h>
+#include <sys/errno.h>
```
qemu-bsd-user, line 31: You could commit this any time (I'm commenting early in a patch if I think I Can do that since I Don't see a way to talk about a commit otherwise.

lib/libsys/Symbol.map
```diff
@@ -1043,11 +1043,13 @@ FBSDprivate_1.0 {
...
-	__libc_sigwait;
 	_cpuset_getdomain;
 	__sys_cpuset_getdomain;
 	_cpuset_setdomain;
 	__sys_cpuset_setdomain;
 	rpctls_syscall;
+
+	__libc_interposing_slot;
+	__libc_sigwait;
+	__set_error_selector;
```
qemu-bsd-user, line 440: why the reordering?

qemu-bsd-user:
So I've looked at the first dozen or so changes as well as trying to suss out the big picture.
I see that you've mentioned you're going to pull some of these into phabs before this moves forward good.
So we'd have libc.so and a new dependency libsys.so. Will that work with old binaries (eg, /bin/ls from FreeBSD 14 on a FreeBSD system will have been compiled with only a libc depend, but will need to pull in both libc and libsys on 15. I imagine it will, but if it can't then I think we need a longer conversation. And how does this relate to the 'filter' you talked about... and what about the versioned symbols? IIRC it should be not a problem, but I wanna make sure.

so looks good, have some questions, but unless there's a technical roadblock I'm not seeing, this will be good.

Also this is bsdimp... I just noticed I was logged into qemu-bsd-user.

lib/libc/gen/getpagesize.c
emaste: Oh yes I'm not sure how I misread it.

lib/libsys/Symbol.map
brooksdavis: In later commits all the `_foo` and `__sys_foo` syscall symbols move to a generate file. There should probably be a preparatory commit that splits the block into "standard syscall stuff" and "everything else".

brooksdavis:
The filter linkage (libc is linked with `--auxiliary libsys.so`) means that any libc symbol present in libsys is preempted by the libsys version when ever libc is loaded (the lld docs suck on this, the binutils ones are more verbose). This looks a bit weird in gdb. When you set a breakpoint on one of these symbols you actually set it on both of them and if you would have created breakpoint 4, you also get 4.1 and 4.2[0]. I've lost track of why we were heading this particularly direction in 2019 as opposed to just linking in libsys or adding it to the linker script that is /usr/lib/libc.so and need to recover that reasoning.

One of the testing things we'll want to do is fairly comprehensively test some quite old FreeBSD binaries. Perhaps extracting /usr/tests from some old dist files and running tests with libc before and after the changes.

[0] gdb's `info breakpoint` is a bit weak here, it shows the files and line numbers for both, but since they are the same that's not helpful. It should show you which file each one was loaded from since the information required is knowable and can be seen with `info proc map`.

kostikbel (commented):
Now that you mention it.  Would it make sense to _not_ aim at ABI compat of libsys with libc.  Instead, you could make libsys export errno-less interface (I am not sure what would it be, might be functions would take *errorp explicitly).  Then you can keep libsys pure syscall wrappers.

Libc wrappers around syscalls then store returned errors into errno, and do whatever else we want.  Of course libc then cannot be a filter for libsys.

May be something like this:
`int open(libsys_res_t *res, const char *path, <whatever>);`
and `typedef struct __tag_libsys_res {
    union __just_tag {
         quad_t res;
         int error;
   };
} libsys_res_t;`
Return value is 0 if PSL_C is clear, then res is the valid union member, and -1 if PSL_C is set.  In the later case, error is errno-compatible error.

lib/libsys/__getosreldate.c
```diff
@@ -44,23 +44,10 @@ int
 __getosreldate(void)
 {
 	static int osreldate;
```
dpgao, line 46: Can this be made non-`static` just like in 9c820a3cf259cc857931c999f3f453b4a2045a5c?

clausecker:
I like this idea, though maybe there'd be a way to get rid of `errno` and move that part to the libc.  Programs just wishing to access system calls should not have to deal with this bit of poor libc design.

brooksdavis:
I'm pondering the proposal of an alternative interface that doesn't do the `errorno` dance and I somewhat like it, but it's more fraught than you might think at first glance.  For the sake of the discussion let's call the new interface `__nsys_foo`.

The relatively straightforward issue is that we need to shift the correct number of arguments and that might include arguments on the stack for anything with more than 5 arguments. That's not too bad, but we'd need appropriate syscall stubs for each case rather than `__sys_foo` being the same in all cases.

32-bit non-i386 compat is more complex. Let's consider the `preadv(2)` system call. The public interface is:
```
ssize_t preadv(int fd, struct iovec *iovp, u_int iovcnt, off_t offset);
```
and the new interface would be
```
int __nsys_preadv(int *resp, int fd, struct iovec *iovp, u_int iovcnt, off_t offset);
```
On 32-bit architectures that aren't i386 we get this:
```
ssize_t preadv(int fd, struct iovec *iovp, u_int iovcnt, int _pad, off_t offset);
int __nsys_preadv(int *resp, int fd, struct iovec *iovp, u_int iovcnt, off_t offset);
```
**Note that `preadv` has an implied `_pad` argument and `__nsys_preadv` does not.**
Normally the compiler hides these in the calling convention and in `struct preadv_args`, but if we need to shift the arguments in assembly we need to be aware that the pad won't be present in the `__nsys_preadv` arguments and only shift up to that point (let's assume for the moment that we don't add any new syscalls with multiple always-64-bit arguments). I think this almost certainly requires assembly to be by makesyscalls.lua.  Not impossible, but quite a deviation from the current SYS.h

The reverse problem also exists (for example `sendfile(2)`'s third argument is an off_t and thus will get padded for `__nsys_sendfile`, but doesn't require padding in the actual system call so we'd have to know it was there and start shifting by two registers.

I'm also not entirely sure what to do with `syscall(2)`. I guess you could blindly adjust the stack as though it took 10 arguments, but that seems a bit sketchy (things would be better on CheriBSD where I'd be able to observe how much I'm copying).  There's then a question if you'd want to bypass the kernel implementation entirely (this is apparently what linux does).

I don't think any of these are insurmountable, but they aren't entirely trivial and would require quite a lot of generated MD assembly and probably need to wait for some refactoring in makesyscalls.lua.

kostikbel:
Additional note is that external _public_ interface from libsys can be made different from the interface glued into libc. We can provide thread-safe interface for consumers, and do whatever hacks required to serve errno (and, if wanted, optimize to avoid creating and new call frame before kernel entry, but this is optional).

FYI, there is a libthr use of syscalls interface that tries to do that, see _umtx_op_err().  That said, one more place where syscall is done is signal trampoline (sigreturn(2)).

The only goal of libsys I see is to provide single place were syscalls are issued, we are not limited in how much interfaces we provide for outside.

lib/libsys/__getosreldate.c
brooksdavis: `__getosreldate` is used in realpath() which makes a performance enhancing syscall based on it so I decided to leave the optimization in place.

brooksdavis:
I think what I've got here is basically ready to land.  I've elected to move the ELF aux args and the interposing table to libsys. I believe this is necessary because otherwise the interposing table gets initialized with the libc versions of syscall stubs. auxv support then fallows because otherwise libsys has undefined symbols for auxv consuming entries the table. It's also part of the system interface so from that perspective I think it makes sense to put it in libsys.

kostikbel:
> I think what I've got here is basically ready to land. I've elected to move the ELF aux args and the interposing table to libsys. I believe this is necessary because otherwise the interposing table gets initialized with the libc versions of syscall stubs. auxv support then fallows because otherwise libsys has undefined symbols for auxv consuming entries the table. It's also part of the system interface so from that perspective I think it makes sense to put it in libsys.

Current interposing table is the mix of syscalls and libc interfaces.  Did you moved every referenced symbol to libsys?
This is not desirable, IMO.

brooksdavis:
> > I think what I've got here is basically ready to land. I've elected to move the ELF aux args and the interposing table to libsys. I believe this is necessary because otherwise the interposing table gets initialized with the libc versions of syscall stubs. auxv support then fallows because otherwise libsys has undefined symbols for auxv consuming entries the table. It's also part of the system interface so from that perspective I think it makes sense to put it in libsys.
> 
> Current interposing table is the mix of syscalls and libc interfaces. Did you moved every referenced symbol to libsys? This is not desirable, IMO.

I've reworked this to split syscalls and purely libc interfaces into two separate tables. I've retained the INTERPO_xxx enum with the same values to preserve the interface between libthr and libc. In due course we could switch syscall entries to use `__libsys_interposing_slot()` directly.

brooksdavis:
I've requested an exp-run: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=276391

lib/libthr/thread/thr_syscalls.c
```diff
@@ -641,6 +641,20 @@ __thr_interpose_libc(void)
...
+	    (interpos_func_t)__thr_##name;
+	SLOT(system);
+	SLOT(tcdrain);
+	SLOT(spinlock);
+	SLOT(spinunlock);
+	SLOT(map_stacks_exec);
+#undef SLOT
+
+	/*
+	 * XXX: should be __libsys_interposing_slot once enough time has
```
kostikbel, line 655: I do not think we need to wait to start using __libsys_interposing_slot() in libthr.  We do require that rtld/libc/libthr where updated atomically.

kostikbel (commented):
_once is really a private libc stuff, not related to libsys.  Can we dup the implementation, by e.g. compiling it both into libc and libsys, and avoiding libc referencing the libsys symbol?

This is following the libsys purpose of being replaceable.

kostikbel:
So errno/.cerror/errno_location are moving to libsys.  I did hoped that at least external libsys interface would be thread-safe.

It seems you did not handle umtx_op_err() from libthr. There is no reason to leave it in libthr, and more, it would contradict the purpose of libsys.

lib/libthr/thread/thr_syscalls.c
brooksdavis: OK, as long as we install libc before libthr (as noted in https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=197366#c8) we can do that. I think I need to keep there fallback in `__libc_interposing_slot` though.

brooksdavis:
> _once is really a private libc stuff, not related to libsys. Can we dup the implementation, by e.g. compiling it both into libc and libsys, and avoiding libc referencing the libsys symbol?

It looks like that would be straightforward. Just need to duplicate the implementation under another filename and not export the symbol.

lib/libthr/thread/thr_umtx.c
```diff
@@ -30,15 +30,13 @@
 #include "thr_private.h"
 #include "thr_umtx.h"
 
-#ifndef HAS__UMTX_OP_ERR
-int _umtx_op_err(void *obj, int op, u_long val, void *uaddr, void *uaddr2)
+int __noinline
+_umtx_op_err(void *obj, int op, u_long val, void *uaddr, void *uaddr2)
```
kostikbel, line 34: I suggest to move the fake implementation for other arches to libsys as well.  Then make the symbol MI.

brooksdavis: Do you think I should filter the symbol as done now or require that libthr link libsys for it? If we're going to filter then I think my current approach is cleanest as we have to have the symbol in libthr in order to filter it.

kostikbel: IMO it is better to link libthr to libsys for _ump_op_err, and put the symbol in the private namespace.  Only libthr is the consumer of the symbol, not libc, and we do not want to export it for third-parties.

kostikbel:
I do not have any more comments.

brooksdavis:
I've run into a problem with the current patch that I don't know how to debug. The exp-run (https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=276391) turned up some failures a couple R packages. It turns out that all this reshuffling is causing a problem with resolving thr_wake in the check_cancel() function in thr_sig.c. /usr/tests/lib/libthr/h_cancel demonstrates the issue. Any help debugging this would be appreciated.

markjdb:
The problem is infinite recursion in rtld. Something calls `__thr_cond_wait()`, which internally calls `thr_wake()`. The implementation is supplied by the filtee, libsys.so.7. At some point, while resolving `thr_wake`, rtld's `load_filtees()` is called, and this results in a longjmp back to the sigsetjmp() call in `_rtld_bind()` in order to upgrade the bind lock from a read lock to a write lock.

When upgrading the lock, `_thr_rtld_wlock_acquire()` is called, which triggers a symbol lookup, which triggers a `load_filtees()` call with the read lock held, which jumps back to `_rtld_bind()`, which tries to upgrade the bind lock...

I suppose that `_rtld_bind()` should unconditionally acquire the write lock if we're trying to resolve a call to `_thr_rtld_wlock_acquire()`, but I don't know how best to implement that.

However, why does load_filtees() try to upgrade the lock even when `obj->filtees_loaded` is already true? If I push the call to `lock_restart_for_upgrade()` inside the if-statement, the infinite recursion goes away, and the package build completes successfully. But I'm not sure if this is correct.

kostikbel:
> The problem is infinite recursion in rtld. Something calls `__thr_cond_wait()`, which internally calls `thr_wake()`. The implementation is supplied by the filtee, libsys.so.7. At some point, while resolving `thr_wake`, rtld's `load_filtees()` is called, and this results in a longjmp back to the sigsetjmp() call in `_rtld_bind()` in order to upgrade the bind lock from a read lock to a write lock.
> 
> When upgrading the lock, `_thr_rtld_wlock_acquire()` is called, which triggers a lazy symbol binding, which triggers a `load_filtees()` call with the read lock held, which jumps back to `_rtld_bind()`, which tries to upgrade the bind lock...
> 
> I suppose that `_rtld_bind()` should unconditionally acquire the write lock if we're trying to resolve a call to `_thr_rtld_wlock_acquire()`, but I don't know how best to implement that.
> 
> However, why does load_filtees() try to upgrade the lock even when `obj->filtees_loaded` is already true? If I push the call to `lock_restart_for_upgrade()` inside the if-statement, the infinite recursion goes away, and the package build completes successfully. But I'm not sure if this is correct.

This (moving the should be fine by itself.  It was simply not exercised under such circumstances.  Please commit the change.

But I am still wondering what symbol bind recursion you are referring to.  Could you please elaborate?  libthr forces early relocations before the rtld lock implementation is replaced by the umtx/thread aware, see the start of _thr_rtld_init().  Perhaps we need to add there some symbols that cause the recursion with libsys.

But

markjdb:
> But I am still wondering what symbol bind recursion you are referring to. Could you please elaborate? libthr forces early relocations before the rtld lock implementation is replaced by the umtx/thread aware, see the start of _thr_rtld_init(). Perhaps we need to add there some symbols that cause the recursion with libsys.

Sorry, my description is not quite right. It's `_thr_ast()` that seems to trigger the re-entry into `_rtld_bind()`. `_thr_rtld_lock_release()` calls it via `THR_CRITICAL_LEAVE()`.

I believe the recursion is: _rtld_bind() -> load_filtees() -> longjmp -> _rtld_bind() -> lock_upgrade() -> _thr_rtld_lock_release() -> _thr_ast() -> _rtld_bind() -> load_filtees() ...

kostikbel:
Please try this
```
diff --git a/lib/libthr/thread/thr_rtld.c b/lib/libthr/thread/thr_rtld.c
index 9026abf941e2..056a361db2e9 100644
--- a/lib/libthr/thread/thr_rtld.c
+++ b/lib/libthr/thread/thr_rtld.c
@@ -238,6 +238,7 @@ _thr_rtld_init(void)
 
 	mprotect(NULL, 0, 0);
 	_rtld_get_stack_prot();
+	_thr_ast(curthread);
 
 	li.rtli_version = RTLI_VERSION;
 	li.lock_create  = _thr_rtld_lock_create;
```

brooksdavis:
> Please try this
> 
> ```
> diff --git a/lib/libthr/thread/thr_rtld.c b/lib/libthr/thread/thr_rtld.c
> index 9026abf941e2..056a361db2e9 100644
> --- a/lib/libthr/thread/thr_rtld.c
> +++ b/lib/libthr/thread/thr_rtld.c
> @@ -238,6 +238,7 @@ _thr_rtld_init(void)
>  
>  	mprotect(NULL, 0, 0);
>  	_rtld_get_stack_prot();
> +	_thr_ast(curthread);
>  
>  	li.rtli_version = RTLI_VERSION;
>  	li.lock_create  = _thr_rtld_lock_create;
> ```

No change in my tests.

markjdb:
> Please try this

Sorry again, the problem is with the thr_wake() symbol, not _thr_ast(). I misunderstood because gdb was hiding a frame, because _thr_ast() (via check_cancel()) calls thr_wake() with a tail call.

If _thr_rtld_init() instead calls thr_wake(curthread->tid), then the crash is gone. I think such a call should be harmless?

kostikbel:
> > Please try this
> 
> Sorry again, the problem is with the thr_wake() symbol, not _thr_ast(). I misunderstood because gdb was hiding a frame, because _thr_ast() (via check_cancel()) calls thr_wake() with a tail call.
> 
> If _thr_rtld_init() instead calls thr_wake(curthread->tid), then the crash is gone. I think such a call should be harmless?

You may do thr_wake(-1) which is even less harmless.

markjdb:
> > > Please try this
> > 
> > 
> > Sorry again, the problem is with the thr_wake() symbol, not _thr_ast(). I misunderstood because gdb was hiding a frame, because _thr_ast() (via check_cancel()) calls thr_wake() with a tail call.
> > If _thr_rtld_init() instead calls thr_wake(curthread->tid), then the crash is gone. I think such a call should be harmless?
> 
> You may do thr_wake(-1) which is even less harmless.

Indeed, thanks. Shall I go ahead and commit that change?

```
commit 84ca126c999468bb4a658a1a168f9fec55df1448
Author: Mark Johnston <markj@FreeBSD.org>
Date:   Fri Feb 2 09:23:53 2024 -0500

    libthr: Force the thr_wake() symbol to be resolved during initialization
    
    Otherwise the lock upgrade performed by rtld's load_filtees() can result
    in infinite recursion, wherein:
    1. _rtld_bind() acquires the bind read lock,
    2. the source DSO's filtees haven't been loaded yet, so the lock upgrade
       in load_filtees() cause rtld to jump to _rtld_bind() and release the
       bind lock,
    3. _thr_rtld_lock_release() calls _thr_ast(), which calls thr_wake(),
       which hasn't been resolved yet,
    4. _rtld_bind() acquires the bind read lock in order to resolve
       thr_wake(),
    5. ...
    
    Reported by:    brooks
    Reviewed by:    kib
    Pull Request:   https://github.com/freebsd/freebsd-src/pull/908
    MFC after:      1 week

diff --git a/lib/libthr/thread/thr_rtld.c b/lib/libthr/thread/thr_rtld.c
index 9026abf941e2..767b4735e1fa 100644
--- a/lib/libthr/thread/thr_rtld.c
+++ b/lib/libthr/thread/thr_rtld.c
@@ -238,6 +238,7 @@ _thr_rtld_init(void)
 
        mprotect(NULL, 0, 0);
        _rtld_get_stack_prot();
+       thr_wake(-1);
 
        li.rtli_version = RTLI_VERSION;
        li.lock_create  = _thr_rtld_lock_create;
```

kostikbel:
Please go ahead.  The change to load_filtees() to avoid unneeded lock upgrade should be committed too.

markjdb:
> Please go ahead. The change to load_filtees() to avoid unneeded lock upgrade should be committed too.

Thank you. Both are committed now.

bsdimp:
Now that this has landed, should we close this?

brooksdavis:
This as supposed to have autoclosed with the merge of 8d2e8f720516e88a896390763ff372a6e5a2550f, but I pushed in the wrong place and lost a race with the mirror.

PR #911, opened 2023-11-29
ncal(1): print the correct option name when -B is specified twice

daniel-tameling:
When using the -B option twice, ncal currently prints the wrong option name, i.e. it warns that -A was specified twice:

```
$ ncal -B1 -B1
ncal: Double -A specified
```


bsdimp:
lgtm

PR #914, opened 2023-11-30
sound(4): Fix OSS API requests for more than 8 channels.

0EVSG:
Audio devices with more than 8 channels need bitperfect mode to operate, the vchan processing chain is limited to 8 channels. For these devices, let applications properly select a certain number of channels supported by the driver, instead of mapping the request to a vchan format.

Please note that with this patch
 - execution path only changes for device drivers that set the bitperfect flag
 - in any case the requested number of channels is limited to what the `AFMT_` audio format supports
 - further down the number of channels has to match one of the configurations of the driver anyway

`snd_uaudio(4)` is currently the only candidate that sets the bitperfect flag (for more than 8 channels), and could be affected by this issue. But it doesn't offer yet to select different device configurations depending on the number of channels requested.

My motivation is to bring in improvements of `snd_hdspe(4)` where this problem prevents applications from properly selecting a specific number of channels. See [here](https://github.com/0EVSG/snd_hdspe_alt).

### DIFF
--- a/sys/dev/sound/pcm/dsp.c
+++ b/sys/dev/sound/pcm/dsp.c
@@ -1502,24 +1502,31 @@ dsp_ioctl(struct cdev *i_dev, u_long cmd, caddr_t arg, int mode,
 
     	case SOUND_PCM_WRITE_CHANNELS:
 /*	case SNDCTL_DSP_CHANNELS: ( == SOUND_PCM_WRITE_CHANNELS) */
-		if (*arg_i < 0) {
+		if (*arg_i < 0 || *arg_i > AFMT_CHANNEL_MAX) {
 			*arg_i = 0;
 			ret = EINVAL;
 			break;
 		}
 		if (*arg_i != 0) {
-			struct pcmchan_matrix *m;
-			uint32_t ext;
+			uint32_t ext = 0;
 
 			tmp = 0;
-			if (*arg_i > SND_CHN_MAX)
-				*arg_i = SND_CHN_MAX;
+			/*
+			 * Map channel number to surround sound formats.
+			 * Devices that need bitperfect mode to operate
+			 * (e.g. more than SND_CHN_MAX channels) are not
+			 * subject to any mapping.
+			 */
+			if (!(dsp_get_flags(i_dev) & SD_F_BITPERFECT)) {
+				struct pcmchan_matrix *m;
 
-			m = feeder_matrix_default_channel_map(*arg_i);
-			if (m != NULL)
-				ext = m->ext;
-			else
-				ext = 0;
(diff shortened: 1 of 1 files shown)


sys/dev/sound/pcm/dsp.c
```diff
@@ -1502,24 +1502,30 @@ dsp_ioctl(struct cdev *i_dev, u_long cmd, caddr_t arg, int mode,
...
-			struct pcmchan_matrix *m;
-			uint32_t ext;
+			uint32_t ext = 0;
 
 			tmp = 0;
-			if (*arg_i > SND_CHN_MAX)
-				*arg_i = SND_CHN_MAX;
+			/*
+			 * Map channel number to surround sound formats.
+			 * Devices with more than 8 channels need bitperfect
```
emaste, line 1516: more than `SND_CHN_MAX` channels?

(I've been wondering what would be required to increase this to support USB interfaces with > 8 channels.)

0EVSG: @emaste, if I misread your comment, please elaborate - I hope it's clear why I'd want to avoid the limit of `SND_CHN_MAX` (== 8) in the bitperfect case?

Also, if you happen to know any committer interested in the sound infrastructure, please tell me. HPS' contributions are dearly missed in that department.

emaste: I think this change is correct, I just mean that the comment should probably refer to `SND_CHN_MAX` (or e.g. `Devices with more than SND_CHN_MAX (8) channels`). If hypothetically we increased `SND_CHN_MAX` to 16 in the future it would invalidate the comment as written.

> This limit is needed for some of the feeders, matrix, equalizer and volume AFAICT.

Yes - perhaps increasing it beyond 8 isn't practical, but I'm sure it's not impossible. It may be that what's needed here is to have some tooling / opinionated setup for virtual_oss, so that there's a straightforward and standard approach. In any case, I'm just musing here, not arguing against this proposed patch.

There are a couple of developers who might take an interest in the sound infrastructure, but nothing concrete yet. In the interim I will try to review and shepherd in this change.

0EVSG: Ok, got it, thanks for the clarification - and your efforts in general.

The updated comment should describe the situation more accurately now.

0EVSG:
Ping! Are there still any open issues with this patch?
@bukinr kindly helps me to get my [snd_hdspe improvements](https://reviews.freebsd.org/D43393) into the tree. Having this patch merged would eliminate some ugly workarounds in a follow-up snd_hdspe commit.

emaste:
Merged in 61c831679615

PR #915, opened 2023-12-01
Use iso646.h and varargs.h from Clang

mrpippy:
These files are functionally identical to the FreeBSD versions.


brooksdavis:
Historically, FreeBSD has provided its own header files for C interfaces. I'm not sure we want to be adding dependencies on the compiler we're reasonably likely to remove at some point. (This isn't "no", but I think there should be a broader discussion.)

mrpippy:
The motivation for this is https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=274542, these files seem like the easiest place to start.
I could leave the FreeBSD header files in-place, although would they be used by anything? Clang would use its own, and GCC from ports appears to include its own independent `iso646.h` and `varargs.h`.
It seems confusing to have header files in the "more obvious" place (`/usr/include`) that aren't actually being used.

(Also, is this the right place for a discussion?)

kevans91:
> (Also, is this the right place for a discussion?)

Starting it on a mailing list would get more eyeballs; freebsd-toolchain@ is probably the closest fit, though the question may be sufficiently philosophical to be good for freebsd-arch@.

bsdimp:
This sort of change needs to be socialized. I'd recommend the toolchain mailing list. i'd rather not do this piecemeal without consensus from the toolchain maintainers. So I'm going to close this. Please do resubmit when this is more mature and/or there's been a discussion and some kind of consensus reached.

PR #919, opened 2023-12-09
committers-ports.dot: rm, not nrm, for Ruslan Makhmatkhanov

grahamperrin:
Fixes: 
975be202db2f Add myself.


grahamperrin:
@mexicarne 

Consistent with `rm` at, for example: 

- <https://docs.freebsd.org/en/articles/contributors/#staff-committers>
- <https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=275425>.

mexicarne:
> @mexicarne
> 
> Consistent with `rm` at, for example:

I believe it was a newline character (\n) before the username)
See other records around. I think this part should be reverted back.

grahamperrin:
> I believe it was a newline character (\n) before the username)

True! 

My mistake. Your email address is correct in the PDF under <https://wiki.freebsd.org/Developers#Group_Memberships>. 

image

PR #922, opened 2023-12-11
ssl/d1_lib.c typos

johnson6841:
Name: Zhan Wei Wu
Email: david5061@gapp.nthu.edu.tw
Line62: over fill -> overfill
Line494: In practice it -> In practice, it
Lint 572:  We allow data remaining -> We allow data to remain
Line 610: require server side -> require a server side
This is from the Advanced UNIX Programming Course (Fall23) at NTHU.


bsdimp:
Since this is for an upstream project, I'm going to have to close it since we can't do anything with it now. I'm terribly sorry.

PR #924, opened 2023-12-11
daemon: many small fixes 

ngortheone:
Fixed narrowing conversions reported by clang-tidy:
- strtol returns long (lines 212, 241)
- read returns ssize_t (line 596)
- kevent.data explicitly cast to int before passing into strerror

Fixed typo in a comment.

### DIFF
--- a/usr.sbin/daemon/daemon.8
+++ b/usr.sbin/daemon/daemon.8
@@ -24,7 +24,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd September 18, 2023
+.Dd December 15, 2023
 .Dt DAEMON 8
 .Os
 .Sh NAME
@@ -151,6 +151,7 @@ option is used or not.
 .It Fl R , Fl -restart-delay Ar restart_delay_seconds
 Supervise and restart the program after the specified delay
 if it has been terminated.
+Valid values are 1-31536000 (up to 1 year).
 .It Fl r , Fl -restart
 Supervise and restart the program after a one-second delay if it has
 been terminated.
--- a/usr.sbin/daemon/daemon.c
+++ b/usr.sbin/daemon/daemon.c
@@ -66,6 +66,8 @@ enum daemon_mode {
 };
 
 struct daemon_state {
+	unsigned char buf[LBUF_SIZE];
+	size_t pos;
 	int pipe_fd[2];
 	char **argv;
 	const char *child_pidfile;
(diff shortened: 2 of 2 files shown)


ngortheone:
@kevans91 @bsdimp pls review

usr.sbin/daemon/daemon.c
```diff
@@ -588,7 +588,7 @@ listen_child(int fd, struct daemon_state *state)
 {
 	static unsigned char buf[LBUF_SIZE];
 	static size_t bytes_read = 0;
-	int rv;
+	ssize_t rv;
```
kevans91, line 591: This is good

usr.sbin/daemon/daemon.c
```diff
@@ -65,6 +65,8 @@ enum daemon_mode {
 };
 
 struct daemon_state {
+	long restart_delay;
```
kevans91, line 68: I don't think I'd bother with widening these, to be honest; `restart_delay` is in seconds and INT_MAX is a lot of seconds to wait, probably not reasonable. `stdmask` is 0, 1, 2, or 1 | 2.

IMO w should instead would convert those two `strtol` calls to `strtonum` with a range of [0, INT_MAX], cast the result to (int) and call it a day.

kevans91 (changes requested):

igalic (approved):
this reads quite sensible to me

usr.sbin/daemon/daemon.8
```diff
@@ -150,7 +150,7 @@ regardless of whether the
 option is used or not.
 .It Fl R , Fl -restart-delay Ar restart_delay_seconds
 Supervise and restart the program after the specified delay
-if it has been terminated.
+if it has been terminated. Valid values are 1-31536000 (up to 1 year)
```
kevans91, line 153: New sentences start on new lines

usr.sbin/daemon/daemon.c
```diff
@@ -49,12 +48,14 @@
...
 #include <unistd.h>
 #include <string.h>
-#include <strings.h>
 #define SYSLOG_NAMES
 #include <syslog.h>
 #include <time.h>
 #include <assert.h>
 
+/* 1 year in seconds */
+#define MAX_RESTART_DELAY 31536000
```
kevans91, line 57: I'd prefer spelling this out as `60 * 60 * 24 * 365` to make it also obvious that it's correct in the definition itself

usr.sbin/daemon/daemon.c
```diff
@@ -580,44 +582,44 @@ restrict_process(const char *user)
...
-	int rv;
+	ssize_t rv;
+	unsigned char *cp;
 
 	assert(state != NULL);
-	assert(bytes_read < LBUF_SIZE - 1);
+	assert(state->pos < LBUF_SIZE - 1);
 
-	rv = read(fd, buf + bytes_read, LBUF_SIZE - bytes_read - 1);
+	rv = read(state->pipe_fd[0], state->buf + state->pos, LBUF_SIZE - state->pos - 1);
```
kevans91, line 595: It's almost tempting to say we should, up by the `daemon_state` definition:

```
#define pipe_rd pipe_fd[0]
#define pipe_wr pipe_fd[1]
```
Then use `state->pipe_rd` and `state->pipe_wr` respectively, just to be a little neater

ngortheone: I dont like marcos, instead I simply deconstructed pipe_fd[2] into 2 individual fields.

usr.sbin/daemon/daemon.c
```diff
@@ -371,12 +377,12 @@ daemon_eventloop(struct daemon_state *state)
 	 */
 	(void)madvise(NULL, 0, MADV_PROTECT);
 
-	if (pipe(state->pipe_fd)) {
+	if (pipe(&state->pipe_rd)) {
```
ngortheone, line 380: This may raise some eyebrows. We can always have a local int[2] to make is safer

kevans91: Yeah, no, that's just way too ick

ngortheone: fixed with local variable

kevans91 (approved):
I think I like that a lot more, thanks!

cperciva:
LGTM

kevans91:
I botched the commit metadata. Landed as:

- a6f795cc89c (daemon: fix clang-tidy warnings)
- 6ac7c9f06ae (daemon: remove redundant parameter from listen_child())
- 24fd3e969eb (daemon: move buffer into daemon_state)
- 5745a584141 (daemon: replace memchr with memrchr)
- 407e3790c3a (daemon: separate pipe_fd[2] into pipe_rd and pipe_wr)

PR #925, opened 2023-12-11
Optimize ring buffer increments

AtariDreams:
These are all in ring buffers. The most common optimization is knowing that an index only increases by 1, and is modulus by some non-power of 2 number, we can check if it gets there before resetting the number to 0, this is better for platforms like ARM where there is no builtin modulus or division.

### DIFF
--- a/sys/dev/hptmv/ioctl.c
+++ b/sys/dev/hptmv/ioctl.c
@@ -87,7 +87,9 @@ static int
 event_queue_add(PHPT_EVENT pEvent)
 {
 	int p;
-	p = (event_queue_tail + 1) % MAX_EVENTS;
+	p = (event_queue_tail + 1);
+	if (p == MAX_EVENTS)
+		p = 0;
 	if (p==event_queue_head)
 	{
 		return -1;
@@ -102,9 +104,9 @@ event_queue_remove(PHPT_EVENT pEvent)
 {
 	if (event_queue_head != event_queue_tail)
 	{
-		*pEvent = hpt_event_queue[event_queue_head];
-		event_queue_head++;
-		event_queue_head %= MAX_EVENTS;
+		*pEvent = hpt_event_queue[event_queue_head++];
+		if (event_queue_head == MAX_EVENTS)
+			event_queue_head = 0;
 		return 0;
 	}
 	return -1;
--- a/sys/dev/sound/pci/maestro3.c
+++ b/sys/dev/sound/pci/maestro3.c
@@ -1160,7 +1160,7 @@ m3_handle_channel_intr:
 		pch = &sc->pch[i];
 		if (pch->active) {
 			pch->ptr = m3_pchan_getptr_internal(pch);
-			delta = pch->bufsize + pch->ptr - pch->prevptr;
+			delta = pch->ptr - pch->prevptr;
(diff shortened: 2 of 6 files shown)


sys/dev/hptmv/ioctl.c
```diff
@@ -87,7 +87,9 @@ static int
 event_queue_add(PHPT_EVENT pEvent)
 {
 	int p;
-	p = (event_queue_tail + 1) % MAX_EVENTS;
+	p = (event_queue_tail + 1);
+	if (p == MAX_EVENTS)
+		p = 0;
```
rilysh, line 92: Modulo doesn't necessarily yield zero, it can introduce some weird bugs, but I'm not completely sure.

sys/dev/sound/pci/maestro3.c
```diff
@@ -1160,7 +1160,7 @@ m3_intr(void *p)
 		pch = &sc->pch[i];
 		if (pch->active) {
 			pch->ptr = m3_pchan_getptr_internal(pch);
-			delta = pch->bufsize + pch->ptr - pch->prevptr;
+			delta = pch->ptr - pch->prevptr;
```
rilysh, line 1163: It should be (bufsize + ptr) - prevptr, but you removed bufsize. Have you tested this code locally and noticed that bufsize calculation wasn't necessary?

tools/tools/vt/mkkfont/mkkfont.c
```diff
@@ -96,7 +96,7 @@ print_mappings(struct font_header *fh, int map_index)
 		printf("{ 0x%04x, 0x%04x, 0x%02x },",
 		    be32toh(fm.vfm_src), be16toh(fm.vfm_dst),
 		    be16toh(fm.vfm_len));
-		col = (col + 1) % 2;
+		col ^= 1;
```
rilysh, line 99: This isn't equivalent. It will *only work* if col value is 1 or 0, anything else just will not. I don't think there would be a significant performance change with it.

rilysh:
@AtariDreams Most of these seem "premature optimizations". Have you tried these changes locally? Some of the changes seem to break on edge cases.

jrtc27 (changes requested):
Every change must be adequately justified. Some are totally unjustified and are churn that achieve no purpose whatsoever (for example, a[b++] and a[b]; b++ are entirely equivalent and modern compilers will optimise the two identically, and compilers know how to do % 2 and / 2 efficiently). Each change needs to be its own PR, with a proper, elaborated commit message and real evidence that it matters if you're going to claim it's in the name of performance. You can't just do a bunch of unrelated changes like this across multiple drivers and tools in one commit just because they all involve circular buffers.

sys/dev/hptmv/ioctl.c
bsdimp: This one is likely bad. The old code is perfectly fine.

bsdimp:
This review is too wide ranging. Please open one pull request per driver. Please indicate how they were tested. And even if it's tested, I'm not sure any of these are worth the risk bugs being introduced by subtly different semantics used. I'm closing it, but do resubmit the ones you can test on hardware. Maybe a call for testing on current@ and/or hackers@ would help find people with this hardware.

PR #926, opened 2023-12-12
exec(3) manpage: corect typo on 'ellipses'

razetime:
This is from the Advanced UNIX Programming Course (Fall'23) at NTHU.


jlduran:
I believe in this case, it is referring to the plural form of ellipsis, ellipses.

PR #927, opened 2023-12-13
Cirrus CI: Update pkg by standard means

jlduran:
pkg version 1.20.9 has been MFHd.

---

cc/ @markjdb @brooksdavis @bapt @emaste 

Granted, we are now back to using HTTP ;-)

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -93,10 +93,8 @@ task:
   install_script:
   - sh .cirrus-ci/pkg-install.sh ${TOOLCHAIN_PKG} git-lite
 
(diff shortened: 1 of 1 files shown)


emaste:
I don't think we need to update pkg explicitly at all -- https://cirrus-ci.com/task/5240449338703872 is https://github.com/freebsd/freebsd-src/commit/92aa071dc711d465c41e691bc2f836ab10055483, and has
```
pkg --version
1.20.9
```

jlduran:
Perfect! Thank you!

PR #934, opened 2023-12-15
sh(1): Fix typo

ChienPei:


igalic (approved):

bsdimp:
This has landed.

PR #935, opened 2023-12-15
sh(1): Fix typo

kevin1010607:


rilysh (approved):

bsdimp:
This has landed.

PR #936, opened 2023-12-15
sh(1): Fix typo

CHIUWEINING:


igalic (approved):
it's funny that we've extended this, but don't have the typo fixed https://github.com/NetBSD/src/blob/trunk/bin/sh/mail.c#L71

bsdimp:
This has landed.

PR #937, opened 2023-12-16
Update LOCKS and MAINTAINERS for Git

fel1x-developer:


bsdimp:
This has landed.

PR #938, opened 2023-12-16
Remove SVN related Git attributes

fel1x-developer:


qemu-bsd-user:
This looks good to my eye. I don't know why the first one is even there. The second is 100% safe to
remove after stable/12 is no longer supported... in 17 days... It likely is fine now, though, to be honest.

bsdimp:
Sorry, the comment from qemu-bsd-user was really from me... I was logged into that for other reasons.

gmshake (approved):
LGTM.

PR #941, opened 2023-12-18
bin/pax/pax.c: typo

johnson6841:
Name: Zhan Wei Wu
Email: david5061@gapp.nthu.edu.tw
Line72: same as uflag except inode -> same as uflag except for inode
This is from the Advanced UNIX Programming Course (Fall23) at NTHU.


bsdimp:
This is good, but in the future, please consider

1. Adding a one-line summary. In this case 'pax: Add missing word 'for' in comment'
2. Using the 'Event:' tag rather than just 'from the Advance....' line: 'Event: Advanced...'

PR #943, opened 2023-12-19
diff(1): fix a typo

YiPrograms:
On the manpage of diff(1), "when" is mistyped to "wen".

This is from the Advanced UNIX Programming Course (Fall'23) at NTHU.


gmshake (approved):
LGTM.

bsdimp:
This has landed.

PR #945, opened 2023-12-19
bhyve/audio.c: avoid re-calculating the length of dev_name

rilysh:
In the function audio_init(), strlen() is being called two times, first to get the length of dev_name and second to use in memcpy(). Creating a local variable and keeping the length avoids this re-calculation.


usr.sbin/bhyve/audio.c
```diff
@@ -80,15 +80,17 @@ audio_init(const char *dev_name, uint8_t dir)
 #endif
 	};
 #endif
+	size_t sz;
```
gmshake, line 83: I'd prefer nlen to sz.

gmshake (approved):
Looks good to me.

usr.sbin/bhyve/audio.c
rilysh: Thanks! I've updated the name.

bsdimp (approved):
This is ready to land.

bsdimp:

PR #946, opened 2023-12-19
Prefer aligned_alloc over memalign

AtariDreams:
It makes the code cleaner and more portable.

### DIFF
--- a/lib/libc/gen/valloc.c
+++ b/lib/libc/gen/valloc.c
@@ -35,10 +35,5 @@
 void *
 valloc(size_t i)
 {
-	void	*ret;
-
-	if (posix_memalign(&ret, getpagesize(), i) != 0)
-		ret = NULL;
-
-	return ret;
+	return alloc_align(getpagesize(), i);
 }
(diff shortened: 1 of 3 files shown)


rilysh (commented):
`aligned_alloc()` was introduced in ISO C11, and it superseded older `posix_memalign()`. So it does the opposite, it removes the portability of `posix_memalign()` (e.g. where ISO C11 isn't supported yet).

Besides there are no differences between these functions except readability. Not sure if it's a good idea to mix-up different standards.

bsdimp:
So, I'm on the fence on this one... but I think I'm net positive.
On the one hand, it seems like needless churn, plus it missed nvmecontrol.
On the other, it's small, and helps us future proof better. Downstream users can 'fix' this with an inline function in cdefs.h if they don't have C11 env. We do have that env and can use it.

So, please do nvmecontrol and make this one commit per directory and force push the results.

bsdimp:
OK. I think that changes that don't fix a problem are being deprioritized, so I'm closing this.

PR #947, opened 2023-12-22
bhyveload/bhyveload.c: avoid re-calculating path length and check if malloc(...) fails

rilysh:
1. In cb_open(...) function, strlen(3) was used two times, to retrieve
the length of "path". However, we can avoid calling strlen(3) for the
second time by keeping the calculated length in a variable.

2. In that same function malloc(3) is used, but there's no error
handling for it. Now if malloc(3) fails, it won't be able to
allocate memory space, and the function right after malloc(3),
stat(2) may fail as well. As stat(2) does check if the function
failed or succeeded, it has a free(3) call to cf, but cf allocation
did fail, so as a result, free(3) will try to _free_ an invalid
address, which will create a segfault.


kevans91:
> 1. In cb_open(...) function, strlen(3) was used two times, to retrieve
>    the length of "path". However, we can avoid calling strlen(3) for the
>    second time by keeping the calculated length in a variable.

GCC at least (I bet clang as well) may (probably will) elide the second call to strlen() anyways, since it's a pure function.

> 2. In that same function malloc(3) is used, but there's no error
>    handling for it.

This is correct.

>    Now if malloc(3) fails, it won't be able to
>    allocate memory space, and the function right after malloc(3),
>    stat(2) may fail as well. As stat(2) does check if the function
>    failed or succeeded, it has a free(3) call to cf, but cf allocation
>    did fail, so as a result, free(3) will try to _free_ an invalid
>    address, which will create a segfault.

This is not correct. We would have crashed from deref'ing a NULL ptr in the stat call anyways, and `free(NULL);` is defined to be acceptable in any case.

rilysh:
Thanks for the review!

> GCC at least (I bet clang as well) may (probably will) elide the second call to strlen() anyways, since it's a pure function.

Indeed, both GCC and Clang do the job pretty well, and omit the second call, but the situation seems dependent on how long and complex the function is. I tested a small demo on here: https://godbolt.org/z/vzxnbnnfn and both GCC and Clang do omit the call but if there's another function (that's in use in the program) in between the calls, both compilers don't omit the second call.

> This is not correct. We would have crashed from deref'ing a NULL ptr in the stat call anyways, and free(NULL); is defined to be acceptable in any case.

You're right. I certainly didn't know that free(NULL); was acceptable until I saw it in [ISO paper](https://www.open-std.org/JTC1/SC22/wg14/www/docs/n1124.pdf).

igalic:
> stat will probably fail, we'll either crash or free(NULL) and bubble up the wrong error.

so we need a better commit message then

usr.sbin/bhyveload/bhyveload.c
```diff
@@ -160,15 +160,22 @@ cb_open(void *arg __unused, const char *filename, void **hp)
...
+	size_t nlen;
 
 	if (!host_base)
 		return (ENOENT);
 
 	strlcpy(path, host_base, PATH_MAX);
-	if (path[strlen(path) - 1] == '/')
-		path[strlen(path) - 1] = 0;
+
+	nlen = strlen(path) - 1;
```
bsdimp, line 170: I'd prefer this be more like: len = strlen(path); if (path[len - 1] == '/') path[len-1]='\0';

kevans91:
> Thanks for the review!
> 
> > GCC at least (I bet clang as well) may (probably will) elide the second call to strlen() anyways, since it's a pure function.
> 
> Indeed, both GCC and Clang do the job pretty well, and omit the second call, but the situation seems dependent on how long and complex the function is. I tested a small demo on here: https://godbolt.org/z/vzxnbnnfn and both GCC and Clang do omit the call but if there's another function (that's in use in the program) in between the calls, both compilers don't omit the second call.

Right, the question for the compiler is whether there's a chance some called function may have some side effect that would change the value. IIRC if you only have pure functions in the sequence, it'll happily elide away but anything non-pure is fair game (unless, perhaps, it's defined in the same TU and doesn't call any impure functions itself?) Builtins they should both understand the ramifications of, so something like `sizeof` wouldn't be an issue.

rilysh:
> IIRC if you only have pure functions in the sequence, it'll happily elide away but anything non-pure is fair game (unless, perhaps, it's defined in the same TU and doesn't call any impure functions itself?)

Yeah, exactly. This issue only comes whenever in the middle of a (pure function) sequence an impure function call happens. Compilers ignore it and call that same function (if used) even if the impure function was unused in that context.

usr.sbin/bhyveload/bhyveload.c
rilysh: Thanks! I've updated it.

bsdimp:
The change itself looks good to me.
It should be two commits: 1 for the slight optimization (which is borderline interesting, to be honest) 1 for the additional malloc check. With good commit messages. Rebase to tip of main. Catch me on discord sometime if you need help with any of this.

usr.sbin/bhyveload/bhyveload.c
```diff
@@ -160,15 +160,22 @@ cb_open(void *arg __unused, const char *filename, void **hp)
...
+
+	len = strlen(path);
+	if (path[len - 1] == '/')
+		path[len - 1] = 0;
+
 	strlcat(path, filename, PATH_MAX);
 	cf = malloc(sizeof(struct cb_file));
+	if (cf == NULL)
+		return (ENOMEM);
+
```
bsdimp, line 178: this code is already in main.

usr.sbin/bhyveload/bhyveload.c
```diff
@@ -160,15 +160,22 @@ cb_open(void *arg __unused, const char *filename, void **hp)
 {
 	struct cb_file *cf;
 	char path[PATH_MAX];
+	size_t len;
 
 	if (!host_base)
 		return (ENOENT);
 
 	strlcpy(path, host_base, PATH_MAX);
```
bsdimp, line 168: this code doesn't seem to be...

bsdimp:
looking at this change against what's in main made me change my mind. This needs to be rebased and reworked. Tagging as such, since it's not ready.

rilysh:
I'm currently busy with some stuff... unsure when I will be free, for now I will close this merge request and will see later if it's needed anymore

PR #948, opened 2023-12-23
Use C99 flexible array

ElyesH:
tests/sys/fs: Use C99 flexible array

### DIFF
--- a/share/examples/tests/tests/atf/Makefile
+++ b/share/examples/tests/tests/atf/Makefile
@@ -46,7 +46,3 @@ file1:
 	@echo "File 1" > ${.TARGET}
 
 .include <bsd.test.mk>
-
-.if ${COMPILER_TYPE} == "clang" && ${COMPILER_VERSION} >= 180000
-CWARNFLAGS.printf_test.c+=	-Wno-format-truncation
-.endif
--- a/share/examples/tests/tests/plain/printf_test.c
+++ b/share/examples/tests/tests/plain/printf_test.c
@@ -62,7 +62,7 @@ snprintf__two_formatters(void)
 static void
 snprintf__overflow(void)
 {
-	char buffer[10];
+	char buffer[];
 
 	if (snprintf(buffer, sizeof(buffer), "0123456789abcdef") != 16)
 		errx(EXIT_FAILURE, "snprintf did not return the expected "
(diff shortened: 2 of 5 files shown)


share/examples/tests/tests/atf/Makefile
```diff
@@ -46,7 +46,3 @@ file1:
 	@echo "File 1" > ${.TARGET}
 
 .include <bsd.test.mk>
-
-.if ${COMPILER_TYPE} == "clang" && ${COMPILER_VERSION} >= 180000
-CWARNFLAGS.printf_test.c+=	-Wno-format-truncation
```
igalic, line 51: your code doesn't fix the printf truncation.
the examples that trigger printf truncation warnings do so because they test for 69c8025a498450b6bc4cc9e97ba779ed1c2e7f4b

igalic (changes requested):
please revert the parts that remove the silencing of the printf warnings

igalic:
cc @DimitryAndric

share/examples/tests/tests/plain/printf_test.c
```diff
@@ -62,7 +62,7 @@ snprintf__two_formatters(void)
 static void
 snprintf__overflow(void)
 {
-	char buffer[10];
+	char buffer[];
```
DimitryAndric, line 65: This is rather nonsensical. You have to allocate _some_ sort of buffer to print to, you cannot just specify a flexible array. Since this particular example is meant to demonstrate a `snprintf` truncation, the buffer size should stay as it is. Of course, this results in a compiler warning, but in this very specific case it can be completely and safely ignored.

tests/sys/fs/fusefs/cache.cc
```diff
@@ -142,7 +142,7 @@ TEST_P(Cache, truncate_by_surprise_invalidates_cache)
 	uint64_t attr_valid, entry_valid;
 	int fd;
 	ssize_t bufsize = strlen(CONTENTS);
-	uint8_t buf[bufsize];
+	uint8_t buf[];
```
DimitryAndric, line 145: This is completely wrong. From the previous line it seems that the buffer size is dynamic, but it is not really the case, since `CONTENTS` is defined on line 140 as a fixed string. It would be better to either use the compile-time size of `CONTENTS` (it would have to be redeclared as a `const char []` instead of a `const char *`), or use `strdup()` to make a dynamically allocated copy of the string. In the latter case, it would also have to be freed, of course.

PR #949, opened 2023-12-24
camcontrol(8): Fix typos

CY-chang:
On line 748, "bigger than" is mistyped as "bigger then", and on line 765, "more than" is mistyped as "more then".


bsdimp:

PR #951, opened 2023-12-24
gvinum(8): Fix a typo

CY-chang:
In the manual page of gvinum(8), "go up from" is mistyped as "go up form" on line 354.


bsdimp:

PR #956, opened 2023-12-27
periodic/daily/801.trim-zfs: Add a daily zfs trim script

llfw:
As mentioned in zpoolprops(7), on some SSDs, it may not be desirable to use ZFS autotrim because a large number of trim requests can degrade disk performance; instead, the pool should be manually trimmed at regular intervals.

Add a new daily periodic script for this purpose, 801.trim-zfs.  If enabled (daily_trim_zfs_enable=YES; the default is NO), it will run a 'zpool trim' operation on all online pools, or on the pools listed in 'daily_trim_zfs_pools'.

The trim is not started if the pool is degraded (which matches the behaviour of the existing 800.scrub-zfs script) or if a trim is already running on that pool.  Having autotrim enabled does not inhibit the periodic trim; it's sometimes to desirable to run periodic trims even with autotrim enabled, because autotrim can elide trims for very small regions.

PR:	275965

### DIFF
--- a/usr.sbin/periodic/etc/daily/801.trim-zfs
+++ b/usr.sbin/periodic/etc/daily/801.trim-zfs
@@ -1,0 +1,59 @@
+#!/bin/sh
+#
+#
+
+# If there is a global system configuration file, suck it in.
+#
+
+if [ -r /etc/defaults/periodic.conf ]
+then
+    . /etc/defaults/periodic.conf
+    source_periodic_confs
+fi
+
+case "$daily_trim_zfs_enable" in
+    [Yy][Ee][Ss])
+	echo
+	echo 'Trimming of zfs pools:'
+
+	if [ -z "${daily_trim_zfs_pools}" ]; then
+		daily_trim_zfs_pools="$(zpool list -H -o name)"
+	fi
+
+	rc=0
+	for pool in ${daily_trim_zfs_pools}; do
+		# sanity check
+		_status=$(zpool list -Hohealth "${pool}" 2> /dev/null)
+		if [ $? -ne 0 ]; then
+			rc=2
+			echo "   WARNING: pool '${pool}' specified in"
+			echo "            '/etc/periodic.conf:daily_trim_zfs_pools'"
+			echo "            does not exist"
+			continue
+		fi
+		case ${_status} in
+		FAULTED)
+			rc=3
+			echo "Skipping faulted pool: ${pool}"
+			continue ;;
+		UNAVAIL)
+			rc=4
+			echo "Skipping unavailable pool: ${pool}"
+			continue ;;
+		esac
+
+		if ! zpool status "${pool}" | grep -q '(trimming)'; then
+			echo "    starting trim of pool '${pool}'"
+			zpool trim ${daily_zfs_trim_flags} "${pool}"
+		else
+			echo "    trim of pool '${pool}' already in progress, skipping"
+		fi
+	done
+	;;
+
+    *)
+	rc=0
+	;;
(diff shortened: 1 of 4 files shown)


usr.sbin/periodic/etc/daily/801.trim-zfs
```diff
@@ -0,0 +1,60 @@
...
+		UNAVAIL)
+			rc=4
+			echo "Skipping unavailable pool: ${pool}"
+			continue ;;
+		esac
+
+		if ! zpool status "${pool}" | grep -q '(trimming)'; then
+			echo "    starting trim of pool '${pool}'"
+			zpool trim "${pool}"
+			[ $rc -eq 0 ] && rc=1
```
igalic, line 48: I don't understand this line

igalic (commented):
looks good, overall, at first glance.

usr.sbin/periodic/etc/daily/801.trim-zfs
lexiwinter: this is to indicate informational output from the script (exit 1) but only if we didn't already encounter a more serious error (exit 3 or 4).  i copied this from 800.scrub-zfs, but it seems to be the method that most periodic scripts use to do this (or at least roughly similar).

usr.sbin/periodic/etc/daily/801.trim-zfs
```diff
@@ -0,0 +1,60 @@
...
+			continue ;;
+		UNAVAIL)
+			rc=4
+			echo "Skipping unavailable pool: ${pool}"
+			continue ;;
+		esac
+
+		if ! zpool status "${pool}" | grep -q '(trimming)'; then
+			echo "    starting trim of pool '${pool}'"
+			zpool trim "${pool}"
```
bsdimp, line 47: Generally, we have knobs to control how the trim happens. For example `130.clean-msgs` has a parameter `daily_clean_msgs_days` which controls the number of days. Other things have flags that can be passed to this program or that.

There's only two options to zpool trim that might be relevant. One is `-d/--secure`  which would allow secure erase of the trimmed data. I'm not sure we even support that. The other is `-r/--rate` which specifies the maximum rate. That we support. Based on that, I'd add a `daily_trim_zfs_rate` variable and pass it -r to the zpool trim. You an see what `130.clean-msgs` does for an example. Had there been more options, I'd opt for the more generic `daily_trim_zfs_flags` to match the rc.d pattern, defaulting that to nothing and unconditionally passing that in.

bsdimp:
Generally I think this is a good idea, and the implementation is pretty close: just one item i'd be tempted to add.

usr.sbin/periodic/etc/daily/801.trim-zfs
llfw: i'm inclined to do this with `daily_trim_zfs_flags` anyway, even if there's only a single useful option right now, since this means less work for users later if more flags are added - does that seem reasonable?

llfw:
sorry, i accidentally pushed a commit to this PR that was meant for a different branch, i think i've successfully un-done that.

llfw:
i've added `daily_zfs_trim_flags`.

i also modified the script to *not* exit with rc 1 if a trim was started, because i think this is not "notable information" as described in periodic(8).  it will still show the output, it just won't be included in the periodic mail if `daily_show_success=NO`.

rebased on current main at the same time.

bsdimp:
I pushed something bad trying to autoclose this

bsdimp:
and now that it's closed, things work better.

PR #957, opened 2023-12-27
libkrb5: avoid crash if MD4 is not available

lexiwinter:
ARCFOUR_string_to_key uses MD4 to create a digest of the provided password.  It does not check for an error return from EVP_DigestInit_ex, so if MD4 is not available in libcrypto, it will crash when attempting to use the invalid EVP object.

This manifests as a crash in kadmind (or kadmin -l) when adding a new key if the "default_keys" setting includes arcfour-hmac-md5 (which it does by default) and MD4 is not available.

It now returns an error instead of crashing.

PR:	275915

### DIFF
--- a/crypto/heimdal/lib/krb5/salt-arcfour.c
+++ b/crypto/heimdal/lib/krb5/salt-arcfour.c
@@ -53,7 +53,11 @@ ARCFOUR_string_to_key(krb5_context context,
 	goto out;
     }
 
-    EVP_DigestInit_ex(m, EVP_md4(), NULL);
+    if (EVP_DigestInit_ex(m, EVP_md4(), NULL) == 0) {
+	ret = EOPNOTSUPP;
+	krb5_set_error_message (context, ret, N_("Cannot create digest", ""));
+	goto out;
+    }
 
     ret = wind_utf8ucs2_length(password.data, &len);
     if (ret) {


igalic:
this should be submitted upstream first, https://github.com/heimdal/heimdal/blob/master/lib/krb5/salt-arcfour.c

lexiwinter:
upstream isn't affected by this, because they ship a compatibility shim called libhcrypto that provides its own implementation of MD4 (and several other algorithms) for OpenSSL 3.0.  FreeBSD has removed this library and connects Heimdal directly to OpenSSL, which is why the bug appears here.

i can still submit the code change upstream, but since it doesn't fix an actual bug for them, is it worth doing that?

igalic:
let's cc @cschuber for a professional opinion

lexiwinter:
actually, the situation is a bit more complicated than i described it: upstream has a configure option to disable the hcrypto fallback, but it doesn't work (because of a build failure).  so this could perhaps theoretically affect upstream as well, but i wasn't able to actually reproduce the issue.

i'm thinking it might make sense to submit the patch anyway, since the code is wrong (missing error checking) regardless of whether it causes an actual bug.  but i'll wait to see what Cy says.

lexiwinter:
i've converted this PR to a draft until we work out what to do with it.  i also mailed -arch@ about the status of Heimdal, which might affect it.

lexiwinter:
Cy replied [on the list](https://lists.freebsd.org/archives/freebsd-arch/2023-December/000580.html).  in short, the last Heimdal merged was abandoned in favour of switching base to use MIT Kerberos.  this makes it unlikely (i would assume) that anything from upstream will be merged before then except security fixes, making this an "inactive" upstream per [CONTRIBUTING.md](CONTRIBUTING.md).

llfw:
since there's been no activity here i converted it from draft to review requested.

bsdimp:
I think cy@ has landed his changes. What's the status of this?

cschuber:
Upstream doesn't have the patch either. We can merge this until MIT replaces Heimdal

cschuber:
I believe we fixed this in stable/14 with c7db2e15e4045e1daba939bb151fc5878f791c7b and 17e941a0c88cac2d8cd28d6614448adbd65d4b72. Can you try these, please?

llfw:
i don't have a stable/14 machine to test but it seems to be fixed in main as of 7aa4e4eb36171d349c7f3948e16c09ff7540c241, thanks.

cschuber:
It was fixed in main early January. The original patch for PR/272835 (https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=272835) was tested mid December.

PR #958, opened 2023-12-27
nfsstat: update option strings in docs

lexiwinter:
Add the missing -q option to the nfsstat(1) manpage SYNOPSIS (it is already documented in DESCRIPTION), and add the missing -E and -q options to the built-in usage output.

PR: 275912

### DIFF
--- a/usr.bin/nfsstat/nfsstat.1
+++ b/usr.bin/nfsstat/nfsstat.1
@@ -25,7 +25,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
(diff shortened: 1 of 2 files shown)


asomers (approved):
LGTM.  But you should add something like "MFC after: 2 weeks" to your commit message.  Also, though it's beyond the scope of this PR, I notice that the -q flag is ignored if neither the -e nor -E flag is used.  That's inconsistent.

lexiwinter:
i added MFC after: 2 weeks.  does that mean i should come back and submit new PRs for the MFCs, or will someone else handle that?

asomers:
Whoever commits the PR will handle the MFC.  You don't have commit access, right?

asomers:
Fixed by 30ce26cacfd6 .

PR #955, opened 2023-12-26
Fix typo (triple S)

jlduran:

PR #960, opened 2023-12-28
CONTRIBUTING.md: fix typo from commmit to commit

rilysh:
At line 44, commit is spelled "commmit" (one extra "m"). It should be spelled "commit".


igalic (approved):
👍 (and now I wonder how many more of those we have. see #954 and #955)

emaste:
committed in 5f71f9636efa, thanks

PR #961, opened 2023-12-28
nfsstat: make -q work without -e/-E

lexiwinter:
cc @asomers


asomers (approved):
LGTM.

lexiwinter:
should i amend this with an MFC after (like the other nfsstat change)?  it's a pretty minor fix that should be MFCable without problems, but i'm not sure what the criteria are for MFCs.

lexiwinter:
i took the initiative and added the MFC after tag to the commit.

emaste:
@asomers do you want to land this? I can add it to my queue if you like

asomers:
I'll take care of it, @emaste

asomers:
Fixed by f79b200b6aef

PR #962, opened 2023-12-28
amd64 & i386: enable VIMAGE in MINIMAL

igalic:
VNET(9) is very useful, and is not loadable.
Enable it in MINIMAL.

### DIFF
--- a/sys/amd64/conf/MINIMAL
+++ b/sys/amd64/conf/MINIMAL
@@ -36,6 +36,7 @@ makeoptions	WITH_CTF=1		# Run ctfconvert(1) for DTrace support
 options 	SCHED_ULE		# ULE scheduler
 options 	NUMA			# Non-Uniform Memory Architecture support
 options 	PREEMPTION		# Enable kernel thread preemption
+options 	VIMAGE			# Subsystem virtualization, e.g. VNET
 options 	INET			# InterNETworking
 options 	INET6			# IPv6 communications protocols
 options 	TCP_OFFLOAD		# TCP offload
--- a/sys/i386/conf/MINIMAL
+++ b/sys/i386/conf/MINIMAL
@@ -41,6 +41,7 @@ makeoptions	WITH_CTF=1		# Run ctfconvert(1) for DTrace support
 
 options 	SCHED_ULE		# ULE scheduler
 options 	PREEMPTION		# Enable kernel thread preemption
+options 	VIMAGE			# Subsystem virtualization, e.g. VNET
 options 	INET			# InterNETworking
 options 	INET6			# IPv6 communications protocols
 options 	TCP_OFFLOAD		# TCP offload


gmshake:
IMO VNET is stable enough, so why is it not included in MINIMAL ? Is it a historical reason?

kprovost:
It's worth keeping some kernel configurations around that do not set VIMAGE, if only so we can ensure that keeps building.

I'm not familiar enough with the different configs to say that must be MINIMAL, but we ought to keep at least some.

igalic:
@kprovost we could follow in the footsteps of 38843fe0f27cb93dc50a3b0d7ba9a958c0581de7 and extend MINIMALUP.

What'd you think?

igalic:
> IMO VNET is stable enough, so why is it not included in MINIMAL ? Is it a historical reason?

i honestly think not enough people are using MINIMAL.
Maybe that's because booting from it is unnecessarily hard. funnily enough, the current amd64 GENERIC-MMCCAM is based on MINIMAL, not GENERIC. #798 corrects that, while also adding MINIMAL-MMCCAM, because it's actually useful. and would be more useful with this PR merged

bsdimp:
MINIMAL works on i386 and amd64. At least it's what drove the devmatch work and in that environment, as long as you load your root device and filesystem in the loader the rest are loaded.  Loader integration is needed here, but even so, I've had enough bug reports to know that several people are using it.

The story is somewhat different on arm64 where many devices might be needed to get things setup. There's some loader integration for FDT devices, but I don't know how far Manu got here.  If you have that all loaded in the loader, then it's the same as amd64. On arm64 it's an after thought on my part that Manu improved for the generations of hardware that existed 3 years ago. Beyond that, it's poorly tested and lightly used.

MINIMAL isn't a finished feature. That's one reason why VNET wasn't in there: I DGAS about VNET on my laptop. The reason is no stronger than that. I don't run multiple jails where VNET matters on my laptops. I honestly don't know how important VNET is here, but I suspect that it's not terrible to include it since the performance hit is only a few percent.

kprovost:
> @kprovost we could follow in the footsteps of [38843fe](https://github.com/freebsd/freebsd-src/commit/38843fe0f27cb93dc50a3b0d7ba9a958c0581de7) and extend MINIMALUP.
> 
> What'd you think?

Possibly. It's a bit odd to have it present in MINIMAL but not MINIMALUP though. Again, I don't have strong feelings about which config shouldn't include VIMAGE, other than "there should be at least one". There are downstream users who disable VIMAGE and we ought to make sure we don't break it.

igalic:
I'll try one more argument, re @bsdimp's "I DGAS about VNET on my laptop":
in its inception, VIMAGE was proposed for more than just network virtualization, and here's another step to try to take it into that direction that I hope gets some more attention: [⚙️ D15865 Provide process space virtualisation functionality for jails](https://reviews.freebsd.org/D15865).

for that reason, I really want to make sure VIMAGE is included in MINIMAL.

gmshake:
> I'll try one more argument, re @bsdimp's "I DGAS about VNET on my laptop": in its inception, VIMAGE was proposed for more than just network virtualization, and here's another step to try to take it into that direction that I hope gets some more attention: [⚙️ D15865 Provide process space virtualisation functionality for jails](https://reviews.freebsd.org/D15865).

That's an interesting feature ;)

> 
> for that reason, I really want to make sure VIMAGE is included in MINIMAL.
> 
> there's also this, from the file itself:
> 
> https://github.com/freebsd/freebsd-src/blob/53df7e58cca7e6726b44f447f10ddc3451cfa851/sys/amd64/conf/MINIMAL#L4-L6

igalic:
i just found another surprise: MINIMAL has no `device cd`
but, really, that only bit me because #798 .

PR #964, opened 2023-12-29
nanobsd: Bump default target media size

jlduran:
Otherwise the instructions in the article [^1] will fail due to a lack of space.

[^1]: https://docs.freebsd.org/en/articles/nanobsd/

Related to:	998abf5a1274 ("nanobsd: Bump rescue size to 8GB")

### DIFF
--- a/tools/tools/nanobsd/defaults.sh
+++ b/tools/tools/nanobsd/defaults.sh
@@ -105,7 +105,7 @@ NANO_NEWFS="-b 4096 -f 512 -i 8192 -U"
 NANO_DRIVE=ada0
 
 # Target media size in 512 bytes sectors
-NANO_MEDIASIZE=2000000
+NANO_MEDIASIZE=16000000
 
 # Number of code images on media (1 or 2)
 NANO_IMAGES=2


bsdimp:
Why 16g?

bsdimp:
This may be due to much larger debug files being installed by default now too.. and maybe i should have looked more closely at rescue when it went in

bsdimp:
Ah, I think I see it.
We needed to add 
```
# We do not need these for rescue
WITHOUT_LIB32=true
WITHOUT_TESTS=true
WITHOUT_DEBUG_FILES=t
WITHOUT_KERNEL_SYMBOLS=t
```
to `CONF_WORLD` in the rescue case. Maybe we should have them there by default? I wonder if that's enough to keep the current smallish sizes...

Also, I need to find some time to go through the nanobsd handbook article. It's a bit dated in some of its advice...

jlduran:
Will try those suggestions!... Currently also cleaning up the dhcpd example (building), the venerable PC Engines I believe is out of production, so won't touch that one, plus other typos/style fixes should be submitted shortly... (no shellcheck).

Not to be interpreted as a sabotage, but in the article, it would be great if we could add a section on how to build a NanoBSD-ish image using poudriere, a very basic example, like the dhcpd one. Let me know if this is of any interest (I currently build my home router and raspberry pi this way).

bsdimp:
Yea. Poudriere is an interesting angle. It's likely better, though, to come up with a more complex example for it. It's quite useful for when you have several pkgs rather than just a few. At the low end, Poudriere offers little over nano. When the build is more complex then Poudriere's dependency solver and clean build env really shines. I've been thinking for a long time, though. Nanobsd.sh is better at subsetting... if poudriere could better subset we might move on entirely from nano... id also be happier if poudriere were more modular and could run w/o root which nano can do.. so maybe an article comparing and contrasting the would be good too

bsdimp:
i'd also like to see nanobsd become 100% root-free out of the box...
But that's clearly beyond the scope of this PR.
https://reviews.freebsd.org/D43232
is what I did to get rescure building again btw. I kinda think we need that in defaults.sh, but we also need to document it for people that want to override (instead of augment). Most conf files today do a straight override.

jlduran:
Closing this pull request.
It will be re-submitted with the changes proposed, together with an overall cleanup.

PR #965, opened 2023-12-29
sctp: update manpages

llfw:
Add missing documentation for sctp_sendv(3) and sctp_recvv(3).

Add a note that sctp_send(3), sctp_sendx(3) and sctp_recvmsg(3) are deprecated by RFC 6458.

Add a STANDARDS section to all functions specified in RFC 6458 to indicate their standards conformance.

---

this should probably be reviewed by someone who knows more about SCTP than i do.

### DIFF
--- a/lib/libc/net/sctp_recvmsg.3
+++ b/lib/libc/net/sctp_recvmsg.3
@@ -25,11 +25,12 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd April 23, 2015
+.Dd December 29, 2023
 .Dt SCTP_RECVMSG 3
 .Os
 .Sh NAME
-.Nm sctp_recvmsg
+.Nm sctp_recvmsg ,
+.Nm sctp_recvv
 .Nd receive a message from an SCTP socket
 .Sh LIBRARY
 .Lb libc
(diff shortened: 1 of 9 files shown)


igalic:
cc @kprovost @tuexen

lib/libc/net/sctp_send.3
```diff
@@ -48,11 +49,25 @@
...
+.Fa "int flags"
+.Fc
 .Sh DESCRIPTION
 The
 .Fn sctp_send
 system call
 is used to transmit a message to another SCTP endpoint.
+The
+.Fn sctp_sendx
+function is used to transmit a message to multiple endpoint addresses.
```
tuexen, line 65: Maybe something like
When issuing an implicit association setup, the
.Fn sctp_sendx
function allows to specify multiple addresses of the peer, which can then be used for the association setup.
This is equivalent to using
.Xr sctp_connectx 3 .

tuexen:
Looking at, for example [IP6(4)](https://man.freebsd.org/cgi/man.cgi?query=ip6&apropos=0&sektion=0&manpath=FreeBSD+14.0-RELEASE+and+Ports&arch=default&format=html#end), it might be useful to add references to RFC 6458 in the `SEE ALSO` section to the man pages  where you added a `STANDARDS` section.

lib/libc/net/sctp_recvmsg.3
```diff
@@ -42,22 +43,28 @@
...
 .Fn sctp_recvmsg
-system call
-is used to receive a message from another SCTP endpoint.
-The
-.Fn sctp_recvmsg
-call is used by one-to-one (SOCK_STREAM) type sockets after a
-successful
+and
+.Fn sctp_recvv
+functions are used to receive a message from another SCTP endpoint.  They are
```
bsdbcr, line 57: You need to have a line break after the sentence stop here.

lib/libc/net/sctp_recvmsg.3
```diff
@@ -231,6 +238,158 @@ The
 .Fa sinfo->info_timetolive
 field is not used by
 .Fn sctp_recvmsg .
+.Pp
+The
+.Fn sctp_recvv
+function works as
+.Fn sctp_recvmsg
+with two differences.  Firstly, the receive buffer is passed as an array
```
bsdbcr, line 246: Another line break is needed here when the first sentence ends.

bsdbcr:
The man page changes need a line break after a sentence stop. Running "mandoc -Tlint" and the textproc/igor port find these sort of issues.

lib/libc/net/sctp_send.3
llfw: i replaced the text with this, which i think is still accurate, but is slightly shorter to fit the style of the paragraph:

```
The
.Fn sctp_sendx
function is used to specify multiple peer addresses when creating an implicit
association, as in
.Xr sctp_connectx 3 .
```

llfw:
- fixed lint errors
- added .Rs for RFC 6458
- changed the description of sctp_sendx
- rebased on current main (although there were no conflicts here)

bsdbcr:
The man page changes look good now, thank you! 👍

PR #970, opened 2023-12-31
bintrans(1): support 'URL-safe' base64 variant

llfw:
Add a new algorithm, b64urlencode/b64urldecode, which supports the URL-safe variant of Base64 defined in RFC 4648 §5.  This replaces the '+' and '/' characters with '-' and '_'.

Add a -U flag to base64(1) to enable this algorithm.

b64encode(1)/b64decode(1) do not support this algorithm for now, since this would require defining a new file format.

PR:	152229

---

i'm submitting this as a draft because there's a couple of things i'm unsure about:

- to support URL-safe base64, i had to import `b64_{pton,ntop}` from libresolv into bintrans.  i think it would be useful to have this version available as a libc function (like a generic `bintrans_{encode,decode}()`), but i'm not sure if this is worth the text or namespace bloat.
- i've hidden the 'b64urlencode' and 'b64urldecode' algorithms behind `#ifdef notyet` because it would require changing the file header from `begin-base64` to something else.  i'm not sure if this is worth it.
- the bintrans code is a bit odd with several different `main()` functions for its various algorithms - i'd like to clean this up, but i think this should be a separate PR.

### DIFF
--- a/usr.bin/bintrans/Makefile
+++ b/usr.bin/bintrans/Makefile
@@ -2,7 +2,7 @@
 .include <src.opts.mk>
 
 PROG=	bintrans
-SRCS=	bintrans.c uuencode.c uudecode.c qp.c
+SRCS=	bintrans.c uuencode.c uudecode.c qp.c base64.c
 MAN=	bintrans.1 uuencode.format.5
 LINKS+=	${BINDIR}/bintrans ${BINDIR}/uuencode
 LINKS+=	${BINDIR}/bintrans ${BINDIR}/b64encode
--- a/usr.bin/bintrans/base64.c
+++ b/usr.bin/bintrans/base64.c
@@ -1,0 +1,337 @@
+/*
+ * Copyright (c) 1996, 1998 by Internet Software Consortium.
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND INTERNET SOFTWARE CONSORTIUM DISCLAIMS
+ * ALL WARRANTIES WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES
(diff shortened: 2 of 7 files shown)


pstef:
> to support URL-safe base64, i had to import b64_{pton,ntop} from libresolv into bintrans. i think it would be useful to have this version available as a libc function (like a generic bintrans_{encode,decode}()), but i'm not sure if this is worth the text or namespace bloat.

I don't see why, resolv.h provides b64_pton and _ntop and I can see you've replaced calls to those with calls to the copies you introduced. I can't see what the copies do better. Currently the only improvement I'd like to see there is described but not implemented in https://reviews.freebsd.org/D35761.

> the bintrans code is a bit odd with several different main() functions for its various algorithms - i'd like to clean this up, but i think this should be a separate PR.

This is because I modularized what was a couple of separate programs and added a couple more on top of it, mostly in https://reviews.freebsd.org/D32943, https://reviews.freebsd.org/D32945, and bapt added a quoted-printable coder in https://reviews.freebsd.org/D36314. But if you can see a way to improve this, feel free to let me know when the PR is ready.

lexiwinter:
> I don't see why

the imported versions take an encoding_t argument which selects between BASE64 and BASE64URL.  this is required for the way i've implemented this, which allows both encoding and decoding this variant.  the alternative would be to teach the existing b64_pton() to silently accept the variant (i believe this was the route originally proposed in the PR) but i don't like this, because 1) it means we can't generate this encoding, only decode it, and 2) it means every user of b64_pton() would now accept the new encoding, which might not be desirable.

> This is because I modularized what was a couple of separate programs and added a couple more on top of it,

i assumed it was something like that.  i didn't mean to impinge your code :-) - in fact the multiple mains itself seems fine given that the various algorithms all take different arguments, but i found the call flow was a bit tricky to unpick, and i think it could be simplified a bit.

bsdimp:
This looks OK, but has a lot of #ifdef notyet still. I'm going to close this. Once it's finished, please resubmit this (or reopen if github will allow it). I think it's a good addition, but we strongly discourage 'draft' pull requests.

PR #972, opened 2023-12-31
umount: Add -d option to detach md devices

ricardobranco777:
Add a `-d` option to umount(8) to detach a memory device.  A similar option exists in Linux's umount(8).

Follow-up to https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=114468

### DIFF
--- a/sbin/umount/umount.8
+++ b/sbin/umount/umount.8
@@ -25,7 +25,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd June 19, 2020
+.Dd January 16, 2024
 .Dt UMOUNT 8
 .Os
 .Sh NAME
@@ -33,7 +33,7 @@
 .Nd unmount file systems
 .Sh SYNOPSIS
 .Nm
-.Op Fl fNnv
+.Op Fl dfNnv
 .Ar special ... | node ... | fsid ...
 .Nm
 .Fl a | A
@@ -71,6 +71,11 @@ except for those mounted at
 .Pa /
 or
 .Pa /dev .
+.It Fl d
+If the filesystem is mounted on an
+.Xr md 4
+device (a memory disk), detach it after
+.Xr unmount 2 .
 .It Fl F Ar fstab
 Specify the
 .Pa fstab
@@ -182,7 +187,8 @@ file system table
 .Xr unmount 2 ,
 .Xr fstab 5 ,
 .Xr autounmountd 8 ,
-.Xr mount 8
+.Xr mount 8 ,
+.Xr mdconfig 8
 .Sh HISTORY
 A
 .Nm
--- a/sbin/umount/umount.c
+++ b/sbin/umount/umount.c
@@ -50,10 +50,20 @@
 
 #include "mounttab.h"
 
+/* used by md_detach() */
+#include <sys/ioctl.h>
+#include <sys/mdioctl.h>
+#include <fcntl.h>
+#include <paths.h>
+
+#define DEV_MD     _PATH_DEV   MD_NAME
+#define DEV_MDCTL   _PATH_DEV  MDCTL_NAME
+
 typedef enum { FIND, REMOVE, CHECKUNIQUE } dowhat;
 
 static struct addrinfo *nfshost_ai = NULL;
-static int	fflag, vflag;
+static int	dflag, fflag, vflag;
+static int	all;
 static char	*nfshost;
 
 struct statfs *checkmntlist(char *);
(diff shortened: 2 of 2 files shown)


sbin/umount/umount.c
```diff
@@ -455,6 +471,14 @@ umountfs(struct statfs *sfs)
 		clnt_destroy(clp);
 	}
 	free(orignfsdirname);
+
+	if (dflag) {
+		if (md_detach(sfs->f_mntfromname) && !multiple)
+			return -1;
```
rilysh, line 477: I should be `return (-1);` to keep consistently with rest of the source.

sbin/umount/umount.c
```diff
@@ -139,6 +153,8 @@ main(int argc, char *argv[])
 			errx(1, "%s: %s", nfshost, gai_strerror(error));
 	}
 
+	multiple = all;
```
igalic, line 156: why do we need a new name for this?

sbin/umount/umount.c
```diff
@@ -455,6 +471,14 @@ umountfs(struct statfs *sfs)
 		clnt_destroy(clp);
 	}
 	free(orignfsdirname);
+
+	if (dflag) {
+		if (md_detach(sfs->f_mntfromname) && !multiple)
+			return (-1);
+		else if (vflag)
```
igalic, line 478: why is this an `else if` and not just an `if`?

sbin/umount/umount.c
```diff
@@ -637,7 +661,47 @@ usage(void)
...
 }
+
+int
+md_detach(const char *device)
+{
+	struct md_ioctl mdio;
+	char *p;
+	int fd;
+
+	bzero(&mdio, sizeof(mdio));
```
igalic, line 676: please use memset here

sbin/umount/umount.c
```diff
@@ -637,7 +661,47 @@ usage(void)
...
+
+	mdio.md_version = MDIOVERSION;
+	mdio.md_options = fflag ? MD_FORCE : 0;
+
+	if (strncmp(device, DEV_MD, sizeof(DEV_MD) - 1)) {
+		if (!multiple)
```
igalic, line 682: We checked this before entering the function

sbin/umount/umount.c
ricardobranco777: > why do we need a new name for this?

I did it mainly to avoid making `all` a global variable, but I agree that it's a much better option.

sbin/umount/umount.c
ricardobranco777: Fixed.

sbin/umount/umount.c
ricardobranco777: > why is this an `else if` and not just an `if`?

Fixed the whole block logic.

sbin/umount/umount.8
```diff
@@ -71,6 +71,9 @@ except for those mounted at
 .Pa /
 or
 .Pa /dev .
+.It Fl d
+If the filesystem is mounted on a memory disk (see
+.Xr md 4) , detach it after unmount(2).
```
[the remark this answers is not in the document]
ricardobranco777: Fixed.

sbin/umount/umount.c
```diff
@@ -637,7 +661,47 @@ usage(void)
...
+	    "usage: umount [-dfNnv] special ... | node ... | fsid ...",
+	    "       umount -a | -A [-F fstab] [-dfnv] [-h host] [-t type]");
 	exit(1);
 }
+
+int
+md_detach(const char *device)
+{
+	struct md_ioctl mdio;
+	char *p;
```
rilysh, line 673: Perhaps it's better to name the variable name `eptr` (or something similar) instead of just `p`.
As that argument is named `endptr`, it's easier to grasp (by that name).

ricardobranco777: Fixed.

sbin/umount/umount.c
igalic: sorry, i meant for multiple

ricardobranco777: > sorry, i meant for multiple

In this case we need the check to avoid a warning when using `-a`.

sbin/umount/umount.c
```diff
@@ -455,6 +469,16 @@ umountfs(struct statfs *sfs)
...
 	}
 	free(orignfsdirname);
+
+	if (dflag) {
+		if (md_detach(sfs->f_mntfromname) == 0) {
+			if (vflag)
+				(void)printf("%s: detached\n",
+				    sfs->f_mntfromname);
+		} else if (!all)
+			return (-1);
```
bsdimp, line 479: so this means umount will return a failure code if -d is specified on non /dev/md backed devices. Is this what you want?

ricardobranco777: > so this means umount will return a failure code if -d is specified on non /dev/md backed devices. Is this what you want?

Yes, but only if the `-a` wasn't specified for obvious reasons.

sbin/umount/umount.c
```diff
@@ -70,24 +80,28 @@ int	 checkname (char *, char **);
 int	 umountfs(struct statfs *sfs);
 void	 usage (void);
 int	 xdr_dir (XDR *, char *);
+int	 md_detach(const char *);
```
bsdimp, line 83: Since this is only used in umount.c, it should be static... but all the other functions above it should be too, so consistency wins out over standard practice. Just  fyi for next time.

bsdimp:
OK. I think this is ready. The style script spotted a non-static function used only in umount... but it is consistent with the others, so I'm going to let the consistency rule trump the static rule.

PR #976, opened 2024-01-01
procfs: Add self & exe symlinks like NetBSD does

ricardobranco777:
This simple patch adds `self` and `exe` symlinks to /proc like NetBSD does.  It would simplify a lot of patches in ports.


bsdimp:
This is good...  Some of your commits, this one included, are only the subject. A few words about why in the 'body' of the commit message is usually a good thing too. I've taken care of writing a couple, but I thought I'd mention it for future reference.

igalic:
I was convinced this was already committed…

PR #977, opened 2024-01-01
Prefer atomic_compare_exchange_weak_explicit in at_quick_exit

AtariDreams:
It's often more efficient on some architectures, and the worst case scenario is that h-next is assigned the same value it has before, which is rare, and doesn't change anything. In addition, release/acquire is fine here too.

### DIFF
--- a/lib/libc/stdlib/quick_exit.c
+++ b/lib/libc/stdlib/quick_exit.c
@@ -54,7 +54,7 @@ at_quick_exit(void (*func)(void))
 		return (-1);
 	}
 	h->cleanup = func;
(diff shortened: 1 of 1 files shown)


rilysh (commented):
Weaker version of atomic_exchange_* has weaker guarantee than the strong one, and also allowed to fail spuriously even if the comparison was successful.

I _think_ it's mostly negligible here. It can improves the performance slightly, but also can introduce some rare bugs (I'm not sure about it). IMO weaker version is absolutely okay, if it's otherwise.

bsdimp:
There's too little benefit for the risk here. There's no current bug, and I don't want to risk introducing a new bug.

PR #980, opened 2024-01-02
Fmake CI test run

bsdimp:
Bootstrapping from fmake hasn't been relevant since FreeBSD 9 when we switched to bmake. We can safely delete the workarounds we put into the tree for that transition now. We haven't supported building -current from FreeBSD 8.x (or 9.x with fmake as the default make) since FreeBSD 11ish. It's well past time to remove it.

I opened up this commit to run github's CI stuff, though it works on MacOS already...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,15 +137,11 @@ __DO_WORLDS=no
 __DO_WORLDS?=yes
 __DO_KERNELS?=yes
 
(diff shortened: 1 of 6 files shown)


brooksdavis (approved):
Thanks for cleaning up these bits.

bsdimp:
thanks brooks. Landed.

PR #981, opened 2024-01-02
release: Add missing /var/crash ZFS dataset

jlduran:
This matches the layout from bsdinstall.

Fixes:	89585511cc05 ("release: Add support for creating ZFS-based VM images")


markjdb:
Thanks, not sure how I missed that. :(

Pushed as adc215de9cabadb89eb4a1c99e047906eacaa58a.

PR #982, opened 2024-01-02
cd9660: Add support for mask,dirmask,uid,gid options

ricardobranco777:
This patch adds arbitrary mask, dirmask, uid & gid support to ISO9660 ala MSDOSFS.

To test:

```
cd /usr/src/sbin/mount_cd9660
make clean obj depend
make
sudo make install
cd /usr/src/sys/modules/cd9660
make clean obj depend
make
sudo make install
sudo kldunload cd9660
sudo mdconfig -a -t vnode -f /path/to/iso.iso
sudo mount_cd9660 -m640 -M750 -Uricardo -Goperator /dev/md0 /mnt
```

### DIFF
--- a/sbin/mount_cd9660/mount_cd9660.8
+++ b/sbin/mount_cd9660/mount_cd9660.8
@@ -29,7 +29,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd August 11, 2018
+.Dd January 2, 2024
 .Dt MOUNT_CD9660 8
 .Os
 .Sh NAME
@@ -39,8 +39,12 @@
 .Nm
 .Op Fl begjrv
 .Op Fl C Ar charset
+.Op Fl G Ar gid
+.Op Fl m Ar mask
+.Op Fl M Ar mask
 .Op Fl o Ar options
 .Op Fl s Ar startsector
+.Op Fl U Ar uid
 .Ar special node
 .Sh DESCRIPTION
 The
@@ -66,6 +70,37 @@ Do not strip version numbers on files.
 only the last one will be listed.)
 In either case, files may be opened without explicitly stating a
 version number.
+.It Fl G Ar group
+Set the group of the files in the file system to
+.Ar group .
+The default gid on non-Rockridge volumes is zero.
+.It Fl U Ar user
+Set the owner of the files in the file system to
+.Ar user .
+The default uid on non-Rockridge volumes is zero.
+.It Fl m Ar mask
+Specify the maximum file permissions for files
+in the file system.
+(For example, a
+.Ar mask
+of
+.Li 755
+specifies that, by default, the owner should have
+read, write, and execute permissions for files, but
+others should only have read and execute permissions).
+See
+.Xr chmod 1
+for more information about octal file modes.
+Only the nine low-order bits of
+.Ar mask
+are used.
+The default
+.Ar mask
+on non-Rockridge volumes is 755.
+.It Fl M Ar mask
+Specify the maximum file permissions for directories
+in the file system.
+See the previous option's description for details.
 .It Fl j
 Do not use any Joliet extensions included in the file system.
 .It Fl o
--- a/sbin/mount_cd9660/mount_cd9660.c
+++ b/sbin/mount_cd9660/mount_cd9660.c
@@ -44,8 +44,11 @@
 
 #include <arpa/inet.h>
 
+#include <ctype.h>
 #include <err.h>
 #include <errno.h>
+#include <grp.h>
+#include <pwd.h>
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
@@ -60,6 +63,9 @@ static struct mntopt mopts[] = {
 	MOPT_END
 };
 
+static gid_t	a_gid(const char *);
+static uid_t	a_uid(const char *);
+static mode_t	a_mask(const char *);
 static int	get_ssector(const char *dev);
 static int	set_charset(struct iovec **, int *iovlen, const char *);
 void	usage(void);
@@ -80,7 +86,7 @@ main(int argc, char **argv)
 	mntflags = verbose = 0;
 	ssector = -1;
 
-	while ((ch = getopt(argc, argv, "begjo:rs:vC:")) != -1)
+	while ((ch = getopt(argc, argv, "begG:jm:M:o:rs:U:vC:")) != -1)
 		switch (ch) {
 		case 'b':
 			build_iovec(&iov, &iovlen, "brokenjoliet", NULL, (size_t)-1);
@@ -90,6 +96,15 @@ main(int argc, char **argv)
 			break;
 		case 'g':
 			build_iovec(&iov, &iovlen, "gens", NULL, (size_t)-1);
+			break;
+		case 'G':
+		        build_iovec_argf(&iov, &iovlen, "gid", "%d", a_gid(optarg));
+			break;
+		case 'm':
+			build_iovec_argf(&iov, &iovlen, "mask", "%u", a_mask(optarg));
+			break;
+		case 'M':
+			build_iovec_argf(&iov, &iovlen, "dirmask", "%u", a_mask(optarg));
 			break;
 		case 'j':
 			build_iovec(&iov, &iovlen, "nojoliet", NULL, (size_t)-1);
@@ -109,6 +124,9 @@ main(int argc, char **argv)
 			break;
 		case 's':
 			ssector = atoi(optarg);
+			break;
+		case 'U':
+		        build_iovec_argf(&iov, &iovlen, "uid", "%d", a_uid(optarg));
 			break;
 		case 'v':
 			verbose++;
@@ -173,8 +191,8 @@ void
 usage(void)
 {
 	(void)fprintf(stderr,
-"usage: mount_cd9660 [-begjrv] [-C charset] [-o options] [-s startsector]\n"
-"                    special node\n");
+"usage: mount_cd9660 [-begjrv] [-C charset] [-G gid] [-m mask] [-M mask]\n"
+"                    [-o options] [-U uid] [-s startsector] special node\n");
 	exit(EX_USAGE);
 }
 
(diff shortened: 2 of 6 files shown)


sbin/mount_cd9660/mount_cd9660.8
```diff
@@ -66,6 +70,37 @@ Do not strip version numbers on files.
...
+.It Fl m Ar mask
+Specify the maximum file permissions for files
+in the file system.
+(For example, a
+.Ar mask
+of
+.Li 755
+specifies that, by default, the owner should have
+read, write, and execute permissions for files, but
+others should only have read and execute permissions.
```
jlduran, line 90: Missing closing parenthesis.

sbin/mount_cd9660/mount_cd9660.8
jlduran, line 1: Missing date bump.

jlduran (commented):
Nice! Just a couple of minor issues with the man page.

sbin/mount_cd9660/mount_cd9660.8
ricardobranco777: > Missing closing parenthesis.

Good catch.  I copied this verbatim from the `mount_msdosfs` manpage that has the same error.

sbin/mount_cd9660/mount_cd9660.8
ricardobranco777: Fixed.

sbin/mount_cd9660/mount_cd9660.c
```diff
@@ -254,3 +272,58 @@ set_charset(struct iovec **iov, int *iovlen, const char *localcs)
...
+	return (uid);
+}
+
+static mode_t
+a_mask(const char *s)
+{
+	int done, rv;
+	char *ep;
+
+	done = 0;
```
bsdjhb, line 320: You might be able to simplify this slightly as:

```
     rv = strtol(optarg, &ep, 8);
     if (ep == optarg || rv < 0 || *ep != '\0')
         errx(...)
```

That is, relying on strtol to return optarg == ep if the first character is invalid and not needing `done` or a default value for `rv`.

Note that FreeBSD's style prefers explicit tests for things like `foo != NULL` or `bar != '\0'` rather than relying on implicit tests like `if (foo)`.

sys/fs/cd9660/cd9660_node.c
```diff
@@ -146,8 +146,22 @@ cd9660_defattr(struct iso_directory_record *isodir, struct iso_node *inop,
 	}
 	if (!ap) {
 		inop->inode.iso_mode |= S_IRUSR|S_IXUSR|S_IRGRP|S_IXGRP|S_IROTH|S_IXOTH;
-		inop->inode.iso_uid = (uid_t)0;
-		inop->inode.iso_gid = (gid_t)0;
+		/*
```
bsdjhb, line 149: The changes in this function should be a separate commit with its own log message.

sys/fs/cd9660/cd9660_node.c
```diff
@@ -146,8 +146,22 @@ cd9660_defattr(struct iso_directory_record *isodir, struct iso_node *inop,
...
+		 * Set some "sane" permissions for non-Rockridge volumes
+		 */
+		switch (inop->inode.iso_mode & S_IFMT) {
+		case S_IFREG:
+			/* Turn off executable bits for regular files */
+			inop->inode.iso_mode &= ~(VEXEC | VEXEC >> 3 | VEXEC >> 6);
```
bsdjhb, line 155: I would suggest using `S_IX*` instead of `VEXEC` with shifts.  However, I'm not sure if makes sense to turn off executable permissions on all files as a non-rock-ridge image may contain executables.

sys/fs/cd9660/cd9660_node.c
```diff
@@ -146,8 +146,22 @@ cd9660_defattr(struct iso_directory_record *isodir, struct iso_node *inop,
...
+			/* FALLTHROUGH */
+		case S_IFDIR:
+			/* Turn on write permissions (owner) */
+			inop->inode.iso_mode |= VWRITE;
```
bsdjhb, line 159: Should be `S_IWUSR` instead of `VWRITE`, but also cd9660 in FreeBSD doesn't support read/write mounts so I'm not sure why it makes sense to add this permission for regular files and directories?

sys/fs/cd9660/cd9660_vfsops.c
```diff
@@ -394,6 +403,23 @@ iso_mountfs(struct vnode *devvp, struct mount *mp)
 	vfs_flagopt(mp->mnt_optnew, "nojoliet", &isomp->im_flags, ISOFSMNT_NOJOLIET);
 	vfs_flagopt(mp->mnt_optnew, "kiconv", &isomp->im_flags, ISOFSMNT_KICONV);
 
+	if (1 == vfs_scanopt(mp->mnt_optnew, "uid", "%d", &v)) {
```
bsdjhb, line 406: Normal FreeBSD style here would be to use `... == 1` rather than `1 == ...`

sys/fs/cd9660/cd9660_vnops.c
```diff
@@ -140,8 +144,27 @@ cd9660_access(struct vop_access_args *ap)
 		}
 	}
 
-	return (vaccess(vp->v_type, ip->inode.iso_mode, ip->inode.iso_uid,
-	    ip->inode.iso_gid, ap->a_accmode, ap->a_cred));
+	file_mode = ip->inode.iso_mode;
+	switch (vp->v_type) {
+	case VDIR:
+		if (ip->i_mnt->im_flags & ISOFSMNT_DMASK)
+			file_mode = ip->i_mnt->im_dmask;
```
bsdjhb, line 151: Should this be `file_mode &= ip->i_mnt->im_dmask` instead of an assignment?  msdosfs uses the &

sys/fs/cd9660/cd9660_vnops.c
```diff
@@ -140,8 +144,27 @@ cd9660_access(struct vop_access_args *ap)
...
+		break;
+	case VREG:
```
bsdjhb, line 153: FWIW, msdosfs applies the non-directory mask to all non-`VDIR` vnodes, not just `VREG`.

sys/fs/cd9660/cd9660_vnops.c
```diff
@@ -169,9 +192,26 @@ cd9660_getattr(struct vop_getattr_args *ap)
 	vap->va_fileid	= ip->i_number;
 
 	vap->va_mode	= ip->inode.iso_mode;
+	switch (vp->v_type) {
+	case VDIR:
+		if (ip->i_mnt->im_flags & ISOFSMNT_DMASK)
```
bsdjhb, line 197: Hmm, these also assign the mask instead of anding the mask.  In the case of an image with rock-ridge attributes this can result in increased permissions which doesn't match the description in the manpage changes.

sbin/mount_cd9660/mount_cd9660.c
ricardobranco777: I copied this verbatim from `mount_msdosfs.c`:

https://github.com/freebsd/freebsd-src/blob/main/sbin/mount_msdosfs/mount_msdosfs.c#L277

sys/fs/cd9660/cd9660_node.c
ricardobranco777: I agree with the `S_IX*` part but turning off the executable bits means a sane default for me as I wouldn't expect executables from non-Rock-Ridge.

sys/fs/cd9660/cd9660_node.c
ricardobranco777: Good catch.  I think I enabled this because when copying with the default `cp` alias (`alias cp='cp -ip'`) I ended up with non-writable files.  Will fix.

sys/fs/cd9660/cd9660_vnops.c
ricardobranco777: This part is enforcing the dmask passed as option.

sys/fs/cd9660/cd9660_vnops.c
ricardobranco777: Because this is about enforcing the fmask.  Will check again the whole thing in the manpage changes.

sys/fs/cd9660/cd9660_node.c
ricardobranco777: Fixed.

sys/fs/cd9660/cd9660_node.c
ricardobranco777: I removed the whole change of this block.

sys/fs/cd9660/cd9660_node.c
ricardobranco777: Fixed.

sys/fs/cd9660/cd9660_vfsops.c
ricardobranco777: Fixed.

sys/fs/cd9660/cd9660_vnops.c
ricardobranco777: Fixed.

sys/fs/cd9660/cd9660_vnops.c
ricardobranco777: Fixed.

sys/fs/cd9660/cd9660_vnops.c
ricardobranco777: Fixed.

sbin/mount_cd9660/mount_cd9660.c
bsdjhb: That's fair.

sbin/mount_cd9660/mount_cd9660.8
```diff
@@ -73,11 +73,11 @@ version number.
 .It Fl G Ar group
 Set the group of the files in the file system to
 .Ar group .
-The default gid on non-Rockridge volumes is zero.
+The default gid on non-Rockridge volumes is zero is the group that mounted the filesystem.
```
bsdjhb, line 76: s/is zero is/ here as well

sys/fs/cd9660/cd9660_vnops.c
```diff
@@ -140,8 +144,25 @@ cd9660_access(struct vop_access_args *ap)
 		}
 	}
 
-	return (vaccess(vp->v_type, ip->inode.iso_mode, ip->inode.iso_uid,
-	    ip->inode.iso_gid, ap->a_accmode, ap->a_cred));
+	file_mode = ip->inode.iso_mode;
+	switch (vp->v_type) {
```
bsdjhb, line 148: If you wanted, you could perhaps simplify the `switch` down to an if.

I looked at msdosfs a bit more and what it does is to set default values of its equivalent of `im_dmask` and `im_fmask` so that it can use them unconditionally without needing flags.  You could choose to do that here as well by initializing `im_dmask` and `im_fmask` to `ACCESSPERMS` in `cd9660_mount` before you check for the options.  You could them remove the `ISOFSMNT_DMASK` and `ISOFSMNT_FMASK` flags and replace this switch with something like `file_mode &= (vp->v_type == VDIR ? ip->i_mnt->im_dmask : ip->i_mnt->im_fmask;` both here and in the `VOP_GETATTR`.

What you have now is also correct, but making that change might be more concise.

bsdjhb (commented):
I think the first commit looks good (perhaps modulo some simplifications I suggested).  The second change is probably ok, but probably needs a second opinion.

sbin/mount_cd9660/mount_cd9660.8
ricardobranco777: Fixed.

sys/fs/cd9660/cd9660_vnops.c
ricardobranco777: Brilliant!  I followed your suggestions and successfully tested a RR & non-RR image with & without the introduced options and it seems to work.

bsdjhb:
I've merged the first commit.  Will try to rustle up some other reviewers for the policy change in the second commit.

ricardobranco777:
> I've merged the first commit. Will try to rustle up some other reviewers for the policy change in the second commit.

Thank you!

PR #983, opened 2024-01-02
quot: Use getopt(3) and show usage() if no arguments

ricardobranco777:
quot: Use getopt(3) and show usage() if no arguments


usr.sbin/quot/quot.c
```diff
@@ -575,42 +575,47 @@ main(int argc, char *argv[])
 	struct statfs *mp;
 	struct fstab *fs;
 	int cnt;
+	int ch;
+
+	if (argc == 1)
+		usage();
```
bsdimp, line 581: should this be before or after we parse the args? There's a few 'no sense' combinations that we should likely prohibit and warn about.

like does it make sense to specify -a and also args? 

quot is quite old and also not a lot used, so hasn't had the updating for these sorts of things that other programs have had.

bsdimp:
I generally like this. I made a suggestion for maybe going a smidge farther, but apart from asking if the test for no args is in the right place, I like this change.

usr.sbin/quot/quot.c
ricardobranco777: Hopefully fixed.

PR #984, opened 2024-01-02
Disable capsicum on md5 et al

ricardobranco777:
Quick fix for https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=276043

### DIFF
--- a/sbin/md5/Makefile
+++ b/sbin/md5/Makefile
@@ -63,7 +63,7 @@ LIBADD=	md
 # available when building for Linux/MacOS or older FreeBSD hosts.
 # We need to bootstrap md5 when building on Linux since the md5sum command there
 # produces different output.
-CFLAGS+=-DHAVE_CAPSICUM
+#CFLAGS+=-DHAVE_CAPSICUM
 .endif
 
 .include <src.opts.mk>


markjdb:
I don't think this is the right move. The problem is specific to the cpuinfo file. md5, wc, etc., are not broken in general, it's just that cpuinfo behaves different when read from a capsicum sandbox. Setting CTFLAG_CAPRD on the hw.model sysctl (like we already do for the hw.machine sysctl) should fix the problem.

ricardobranco777:
> I don't think this is the right move. The problem is specific to the cpuinfo file. md5, wc, etc., are not broken in general, it's just that cpuinfo behaves different when read from a capsicum sandbox. Setting CTFLAG_CAPRD on the hw.model sysctl (like we already do for the hw.machine sysctl) should fix the problem.

Where is that done?  Couldn't find anything with CTFLAG_CAPRD.

markjdb:
> > I don't think this is the right move. The problem is specific to the cpuinfo file. md5, wc, etc., are not broken in general, it's just that cpuinfo behaves different when read from a capsicum sandbox. Setting CTFLAG_CAPRD on the hw.model sysctl (like we already do for the hw.machine sysctl) should fix the problem.
> 
> Where is that done? Couldn't find anything with CTFLAG_CAPRD.

I meant to write CTLFLAG_CAPRD. A proposed patch is here: https://reviews.freebsd.org/D43281

ricardobranco777:
> > > I don't think this is the right move. The problem is specific to the cpuinfo file. md5, wc, etc., are not broken in general, it's just that cpuinfo behaves different when read from a capsicum sandbox. Setting CTFLAG_CAPRD on the hw.model sysctl (like we already do for the hw.machine sysctl) should fix the problem.
> > 
> > 
> > Where is that done? Couldn't find anything with CTFLAG_CAPRD.
> 
> I meant to write CTLFLAG_CAPRD. A proposed patch is here: https://reviews.freebsd.org/D43281

Thanks.

But the md5 code needs a fix anyway for the `XXX` block.  Why only the last argument?

markjdb:
> > > > I don't think this is the right move. The problem is specific to the cpuinfo file. md5, wc, etc., are not broken in general, it's just that cpuinfo behaves different when read from a capsicum sandbox. Setting CTFLAG_CAPRD on the hw.model sysctl (like we already do for the hw.machine sysctl) should fix the problem.
> > > 
> > > 
> > > Where is that done? Couldn't find anything with CTFLAG_CAPRD.
> > 
> > 
> > I meant to write CTLFLAG_CAPRD. A proposed patch is here: https://reviews.freebsd.org/D43281
> 
> Thanks.
> 
> But the md5 code needs a fix anyway in the `XXX` block. Why only the last argument?

It's because md5 was sandboxed before cap_fileargs(3) was introduced. That library lets one access an arbitrary list of files from within a capability mode sandbox. md5 should be converted to use cap_fileargs(3). wc(1) already does: see the uses of fileargs_*. If you'd like to work on this, please cc me on the PR and I'll help get it landed.

PR #985, opened 2024-01-02
wc: Do not use st_size if it equals zero

ricardobranco777:
Fix for wc(1) not being able to `wc -c` a file in pseudo-filesystems that advertize a zero size value.


rilysh (approved):

igalic:
is there a bug in bugzilla?

ricardobranco777:
> is there a bug in bugzilla?

Forgot to open https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=276093

PR #986, opened 2024-01-03
bin/ed/ed(1): typo fix

wipbeet:
Adding a missing verb "is" on line 723 :
"If the g (global) suffix is given, then every match to be replaced." should be "If the g (global) suffix is given, then every match is to be replaced."

Changing is to are on line 835 :
"Any previous contents of file is lost without warning." changed to "Any previous contents of file are lost without warning."

Event: This is from the Advanced UNIX Programming Course (Fall’23) at NTHU.

### DIFF
--- a/bin/ed/ed.1
+++ b/bin/ed/ed.1
@@ -720,7 +720,7 @@ with
 By default, only the first match in each line is replaced.
 If the
 .Em g
-(global) suffix is given, then every match to be replaced.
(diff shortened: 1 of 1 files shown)


gmshake (approved):
LGTM.

lwhsu (approved):

lwhsu:
Merged as 84886bf93ae96ee4d19cd7f31696ba368cabddae

PR #988, opened 2024-01-03
md5: Enter capability mode earlier

ricardobranco777:
Fix for https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=276043

### DIFF
--- a/sbin/md5/Makefile
+++ b/sbin/md5/Makefile
@@ -58,15 +58,18 @@ MLINKS=	md5.1 md5sum.1 \
 
 LIBADD=	md
 
-.ifndef(BOOTSTRAPPING)
+.include <src.opts.mk>
+
+.if ${MK_CASPER} != "no" && !defined(RESCUE) && !defined(BOOTSTRAPPING)
 # Avoid depending on capsicum during bootstrap. caph_limit_stdout() is not
 # available when building for Linux/MacOS or older FreeBSD hosts.
 # We need to bootstrap md5 when building on Linux since the md5sum command there
 # produces different output.
 CFLAGS+=-DHAVE_CAPSICUM
+CFLAGS+=-DWITH_CASPER
+LIBADD+=	casper
+LIBADD+=	cap_fileargs
 .endif
-
-.include <src.opts.mk>
 
 HAS_TESTS=
 SUBDIR.${MK_TESTS}+= tests
--- a/sbin/md5/md5.c
+++ b/sbin/md5/md5.c
@@ -46,6 +46,8 @@
 #ifdef HAVE_CAPSICUM
 #include <sys/capsicum.h>
 #include <capsicum_helpers.h>
+#include <libcasper.h>
+#include <casper/cap_fileargs.h>
 #endif
 
 /*
@@ -310,6 +312,7 @@ gnu_check(const char *checksumsfile)
 	const char	*digestname;
 	size_t	digestnamelen;
 	size_t	hashstrlen;
+	struct stat st;
 
 	if (strcmp(checksumsfile, "-") == 0)
 		inp = stdin;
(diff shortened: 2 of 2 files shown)


sbin/md5/md5.c
```diff
@@ -313,7 +319,11 @@ gnu_check(const char *checksumsfile)
 
 	if (strcmp(checksumsfile, "-") == 0)
 		inp = stdin;
+#ifdef HAVE_CAPSICUM
+	else if ((inp = fileargs_fopen(fa, checksumsfile, "r")) == NULL)
```
markjdb, line 323: I think there's no need to have HAVE_CAPSICUM anymore. If WITH_CASPER is not defined, the fileargs_* functions will automatically provide non-casper implementations. See cap_filesargs.h.

ricardobranco777: Fixed.

sbin/md5/md5.c
```diff
@@ -610,20 +622,11 @@ main(int argc, char *argv[])
...
-			 * earlier.
-			 */
-			if (*(argv + 1) == NULL) {
 #ifdef HAVE_CAPSICUM
-				cap_rights_init(&rights, CAP_READ, CAP_FSTAT);
-				if (caph_rights_limit(fileno(f), &rights) < 0 ||
-				    caph_enter() < 0)
-					err(1, "capsicum");
+			if (caph_rights_limit(fileno(f), &rights) < 0 ||
+			    caph_enter() < 0)
```
markjdb, line 627: I don't think we should be calling caph_enter() at all here. Now you call caph_enter_casper() before processing input, that's all that's needed.

ricardobranco777: Makes sense. Hopefully fixed now.

ricardobranco777:
This PR fails with GNU `-c` option.

ricardobranco777:
`make check` is now passing...

Had to get rid of the text vs binary mode stuff to get sane pathnames to pass to `fileargs_init(3)`.  The manpage states that _this implementation does not differentiate between binary and text mode_ anyway.

sbin/md5/md5.c
ricardobranco777: > I think there's no need to have HAVE_CAPSICUM anymore. If WITH_CASPER is not defined, the fileargs_* functions will automatically provide non-casper implementations. See cap_filesargs.h.

I think we need HAVE_CAPSICUM because otherwise /usr/src/rescue fails to compile:

```
ld: error: undefined symbol: fileargs_init
>>> referenced by _$$hide$$ md5.lo md5.c
>>>               md5.lo:(_$$hide$$ md5.lo main)

ld: error: undefined symbol: fileargs_fopen
>>> referenced by _$$hide$$ md5.lo md5.c
>>>               md5.lo:(_$$hide$$ md5.lo main)

ld: error: undefined symbol: fileargs_free
>>> referenced by _$$hide$$ md5.lo md5.c
>>>               md5.lo:(_$$hide$$ md5.lo main)
cc: error: linker command failed with exit code 1 (use -v to see invocation)
*** Error code 1
```

ricardobranco777: Fixed.

sbin/md5/md5.c
```diff
@@ -357,6 +361,15 @@ gnu_check(const char *checksumsfile)
...
 			errx(1, "malloc failed");
+
+		if (*filename == '*' ||
+		    *filename == ' ' ||
+		    *filename == 'U' ||
+		    *filename == '^') {
+			if (lstat(filename, &st) != 0)
+				filename++;
+		}
+
```
markjdb, line 372: Why did this block move? Note that fileargs_lstat() is available.

ricardobranco777: > Why did this block move? Note that fileargs_lstat() is available.

I had to move it to make exactly one call to fileargs_init() where I need to pass argc & argv.  The latter is recreated after the call to gnu_check() which returns filenames with those one char prefixes.

With this move we're making gnu_check() return a list of sane filenames that we can use.  It's also the place where this block of code makes sense.

I didn't need fileargs_lstat() and it works on symlinks as it is.  It passed the tests.  Do you think I have to change the call?

sbin/md5/Makefile
```diff
@@ -68,6 +68,12 @@ CFLAGS+=-DHAVE_CAPSICUM
```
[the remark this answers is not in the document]
ricardobranco777: > We also need to disable casper when `BOOTSTRAPPING` is set. Note that all of the CI jobs are currently failing.

Hopefully fixed now.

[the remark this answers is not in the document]
ricardobranco777: Will rebase from latest main.

markjdb:
Merged, thank you.

dag-erling:
I should have been consulted on this PR and it should not have been committed as is. Now I'm going to have to figure out how to undo the damage...

PR #990, opened 2024-01-03
tail: Do not trust st_size if it equals zero

ricardobranco777:
Fix for `tail` not being able to operate on files residing in pseudo-filesystems that advertize a zero size value.

To reproduce: `tail -1 /compat/linux/proc/mounts`

Bug: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=276107


delphij (approved):

delphij:
LGTM; pushed to `main`.

emaste:
1fb3caee72241b9b4dacbfb0109c972a86d4401f

emaste:
(oops, clicked the wrong button and reopened by accident)

PR #999, opened 2024-01-05
/bin/pax/file_subs.c: typo

mnu1011:
Line719: has -> have
(The auxiliary verb should be followed by the base form of the verb)

### DIFF
--- a/bin/pax/file_subs.c
+++ b/bin/pax/file_subs.c
@@ -716,7 +716,7 @@ set_pmode(char *fnm, mode_t mode)
  *	block boundaries significantly reduces the overhead when copying files
  *	that are NOT very sparse. This overhead (when compared to a write) is
  *	almost below the measurement resolution on many systems. Without it,
- *	files with holes cannot be safely copied. It does has a side effect as
+ *	files with holes cannot be safely copied. It does have a side effect as
  *	it can put holes into files that did not have them before, but that is
  *	not a problem since the file contents are unchanged (in fact it saves
  *	file space). (Except on paging files for diskless clients. But since we


lwhsu:
Please note that your name is different from the `From:` (`Author:` in git) line, as you can see in https://patch-diff.githubusercontent.com/raw/freebsd/freebsd-src/pull/999.patch , note the space between in your last and first name.  You don't have to specify them again in the pull request comment as what in commit object is sufficient and more import, as they will be recorded forever.

For a pull request, we don't need "merge commit" (the one with "Merge branch ...") and the duplicate commits (see above link, it contains two commits with the same change.) Please squash them and do a force push to update this PR.  Also, you don't need to attach the patch file, as the commit itself is what we only need, it just takes unnecessary resource.

Finally, please change the last line:

```
```

to

```
Event:    Advanced UNIX Programming Course (Fall’23) at NTHU.
```

lwhsu:
> Thank you for taking the time to respond. I have a couple of questions:
> 
>     1. when you mentioned that my name and the 'From: (Author: in git) line' are different, are you referring to the distinction between mnu1011 and HUNG-CHI CHANG? If so, it's because of Spec requirements, so I configured gitconfig to use my passport name. Otherwise, I prefer to use the name mnu1011.
> 
>     2. In pull request comments and commits, is there no need to explicitly mention Name and Email? Because commit metadata already includes this information?

If that is the case, no need to mention the name and email in the commit message and we can accept your commit metadata.

lwhsu (approved):

lwhsu:
Merged as 711a74e19ffd7bcd116744a333c5c22835c17307

PR #1000, opened 2024-01-05
share/doc/IPv6/IMPLEMENTATION typo

Koios1143:
Line `710`, using `do` might be better to fit the grammar rather than `does`.
Line `1484`, delete a `s` after `need` might be better fit `routers`.

Name: Ho-Kun Lin
Email: s110062126@m110.nthu.edu.tw

### DIFF
--- a/share/doc/IPv6/IMPLEMENTATION
+++ b/share/doc/IPv6/IMPLEMENTATION
@@ -707,7 +707,7 @@ node-local multicast group ff01::1.
 1.4.2 Stateless address autoconfiguration on hosts
 
 In IPv6 specification, nodes are separated into two categories:
-routers and hosts.  Routers forward packets addressed to others, hosts does
+routers and hosts.  Routers forward packets addressed to others, hosts do
 not forward the packets.  net.inet6.ip6.forwarding defines whether this
 node is a router or a host (router if it is 1, host if it is 0).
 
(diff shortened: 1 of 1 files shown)


lwhsu:
Please check the instruction from TA. You don't have to provide your name and email in the comment but you have to provide those in the commit metadata.

As you can see in https://patch-diff.githubusercontent.com/raw/freebsd/freebsd-src/pull/1000.patch , the `From:` (`Author:` in git) line is generated by github, which is not encouraged.

Also, please use:

```
```

At the bottom of the commit message.

lwhsu:
You don't have to close this. Just update the commit in your local branch and do a force push is sufficient.

Koios1143:
So if there's any wrong with the commit, even after opening a pull request, I can still update that branch, and then the pull request will apply the new commit, not the old one. Is that right?

lwhsu:
> So if there's any wrong with the commit, even after opening a pull request, I can still update that branch, and then the pull request will apply the new commit, not the old one. Is that right?

Exactly. That's the whole point of the PR and the review process.  Those are the key words you might want to check out: `git commit --amend`, `git rebase -i`, `git push -f`.

Koios1143:
Got it!
Appreciate for kind reply with those details and keywords, I'll do it next time when I open a pull request.
It's a nice experience of the first pull request 😌

PR #1003, opened 2024-01-05
share/doc/IPv6/IMPLEMENTATION typo

Koios1143:
Line `710`, using `do` might be better to fit the grammar rather than `does`.
Line `1484`, delete a `s` after `need` might be better fit `routers`.


lwhsu:
Merged as dcde2454bc52f5d959a4e65692350773044d88c6

PR #1005, opened 2024-01-06
units.c: decrease strlen() calls and use else if in lookupunit() function

rilysh:
1. A lot of times strlen() was used to find the length of a same string. Sometimes compilers can omit these accordingly, but calls inside complex loops are often missed. We can decrease these calls by keeping the length in a variable.

2. In lookupunit() function, two if statements are being used. If the first one fails it will go to the second one, which is okay but if the first one succeeds, it will still go to the second but it will fail (for the second time). As the unit is a const and the same length was used we can use an else if, so it wouldn't be called for the second time if the first branch succeeds.

### DIFF
--- a/usr.bin/units/units.c
+++ b/usr.bin/units/units.c
@@ -117,7 +117,8 @@ readunits(const char *userfile)
 {
 	FILE *unitfile;
 	char line[512], *lineptr;
-	int len, linenum, i;
+	int linenum, i;
+	size_t len0, len1;
 	cap_rights_t unitfilerights;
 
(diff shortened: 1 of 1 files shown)


bsdimp:
I think this change is correct, but I'm not 100% sure.

bsdimp:
(1) seems too churny to not fix a specific problem. While that was oK when you submitted this, we've recently updated our policy to disallow it when there's not a problem that can be articulated.
(2) seems OK. You can articulate a change for it.

so, can you resubmit with (1) removed?

PR #1006, opened 2024-01-06
exit.3: add the comma after an empty space

rilysh:
exit(3) man page shows __cxa_atexit(3,) instead of __cxa_atexit(3), in a particular section. It seems the comma gets inside the parenthesis and with an extra space, it can be viewed as expected.

### DIFF
--- a/lib/libc/stdlib/exit.3
+++ b/lib/libc/stdlib/exit.3
@@ -78,7 +78,7 @@ The
 implementation of the
 .Fn _Exit
 function does not call destructors registered with
-.Xr __cxa_atexit 3,
+.Xr __cxa_atexit 3 ,
 does not flush buffers, and does not close streams.
 .Pp
 Both functions make the low-order eight bits of the


lib/libc/stdlib/exit.3
```diff
@@ -78,7 +78,7 @@ The
 implementation of the
 .Fn _Exit
 function does not call destructors registered with
-.Xr __cxa_atexit 3,
+.Xr __cxa_atexit 3 ,
```
igalic, line 81: this man page doesn't exist tho

(at least i can't find it on https://man.freebsd.org/ or https://man-dev.freebsd.org/)

igalic (commented):
this might not be a valid `.Xr` to begin with.

lib/libc/stdlib/exit.3
rilysh: This exists at: https://man.freebsd.org/cgi/man.cgi?query=exit&apropos=0&sektion=3&manpath=FreeBSD+15.0-CURRENT&arch=default&format=html

It's a libc function.

igalic: I mean the man page explaining `__cxa_atexit`

rilysh: I believe __cxa_atexit() is an internal function and not something that can be used by user-space programs.

But I'm not sure why there's no man page.

bsdimp:
This change is correct. We sometimes reference manual entries that don't exist to indicate where to find the code...  In this case, it makes sense to document exit's behavior...  This is, however, an internal interface that real users shouldn't be calling.

I think we should land this since it's an actual error and produces an change in the output.

bsdimp:
2a3d453049ae6da2b0071185827d6b140fbeea62

PR #1016, opened 2024-01-08
bhyve/acpi.c: return ENOMEM instead of EFAULT and call free() after being used

rilysh:
1. In basl_load() function, when allocation fails, it returns an EFAULT instead of ENOMEM. An EFAULT can mislead in some scenarios, whereas an ENOMEM for an allocation function makes much more sense.

2. Call free() on addr, as it's not being used anymore after the basl_table_append_bytes() function.


igalic (approved):
👍 this seems sensible

markjdb:
This seems ok. There are other leaks, e.g., if the BASL_EXEC bodies fail, we will leak `addr`, but that's going to be a fatal error anyway. This patch would at least placate static analyzers.

PR #1019, opened 2024-01-09
man: fix redirect for locales and file arguments

mohd-akram:
Use the locale-specific path for localized pages and the appropriate parent directory for file arguments when handling redirects.

### DIFF
--- a/usr.bin/man/man.sh
+++ b/usr.bin/man/man.sh
@@ -297,7 +297,7 @@ manpath_warnings() {
 	fi
 }
 
-# Usage: man_check_for_so page path
+# Usage: man_check_for_so path
 # Returns: True if able to resolve the file, false if it ended in tears.
 # Detects the presence of the .so directive and causes the file to be
 # redirected to another source file.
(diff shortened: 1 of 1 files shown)


bsdimp:
This change looks like it should be good. How can I test, though, that it fixes the bug you are chasing? And what's the current misbehavior look like?

mohd-akram:
This is how I tested:

Setup:

```shell
cd /tmp
mkdir -p man/man1/amd64
mkdir -p man/fr.UTF-8/man1
echo hello >man/man1/page.1
echo salut >man/fr.UTF-8/man1/page.1
echo '.so man1/page.1' >man/man1/regular.1
echo '.so man1/page.1' >man/man1/amd64/machine.1
echo '.so man1/page.1' >man/fr.UTF-8/man1/locale.1
```

For testing the local version `alias man=~/freebsd-src/usr.bin/man/man.sh` and `unalias man` afterwards.

Tests:

```shell
export MANPAGER=cat
(cd man/man1/ && man ./regular.1) #should show hello
(cd man/man1/amd64 && man ./machine.1) # should show hello
LANG=fr_FR.UTF-8 MANPATH=man man locale # should show salut
```

PR #1020, opened 2024-01-10
amd64 MINIMAL config: Add back FFS

ricardobranco777:
Without this option (that is present in i386 MINIMAL), the kernel cannot boot into UFS.

Fixes https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=276240

### DIFF
--- a/sys/amd64/conf/MINIMAL
+++ b/sys/amd64/conf/MINIMAL
@@ -40,6 +40,7 @@ options 	INET			# InterNETworking
 options 	INET6			# IPv6 communications protocols
 options 	TCP_OFFLOAD		# TCP offload
 options 	SCTP_SUPPORT		# Allow kldload of SCTP
+options 	FFS			# Berkeley Fast Filesystem
 options 	SOFTUPDATES		# Enable FFS soft updates support
 options 	UFS_ACL			# Support for access control lists
 options 	UFS_DIRHASH		# Improve performance on big directories


igalic:
and that's not something that can be loaded as module in [`loader.conf(5)`](https://man.freebsd.org/loader.conf(5)) ?

ricardobranco777:
> and that's not something that can be loaded as module in [`loader.conf(5)`](https://man.freebsd.org/loader.conf(5)) ?

Maybe, but I don't see it documented in the file and minimal kernels are supposed to just work (at least boot).  Instead, I was stuck with a rootmount shell with no option to load the module.

bsdjhb:
FFS was intentionally removed from the amd64 MINIMAL config in 0c1c5e36ebbb7e30a5e57a044a20ea2c4f7cf793 (a change that was just not mirrored over to i386).  The intention is that you would use `load ufs` from the loader or `ufs_enable=YES` in loader.conf.

bsdimp:
We do need to document MINIMAL better....

qemu-bsd-user:
Maybe we should create a GENERIC.9 and a MINIMAL.9 for auxiliary information you need to know about the config.

ricardobranco777:
Are _all_ the UFS goodies available when compiled as a module?

ricardobranco777:
I still think that with UFS should work just like with ZFS with minimal kernel.  I didn't need `zfs_load=YES` at all.

igalic:
here's my `/boot/loader.conf`, as created by `bsdinstall`, because it knew I'm going to use ZFS on root, but assumed I'm using the GENERIC kernel:

```sh
security.bsd.allow_destructive_dtrace=0
kern.geom.label.disk_ident.enable="0"
kern.geom.label.gptid.enable="0"
cryptodev_load="YES"
zfs_load="YES"
```

bsdimp:
> Are _all_ the UFS goodies available when compiled as a module?

As far as I know.  If not, let me know.

I'm going to close this since it's not something we're going to do unless we learn new information to change our minds (like you tried it and it failed in some way).

PR #1022, opened 2024-01-10
Refactor strchr(p, 0) -> p + strlen(p)

AtariDreams:
They are both equivalent. However, given that strchr actually in fact just special cases 0, it is better to just make the intention clearer anyway by conveying in the code itself that this is what is happening.

### DIFF
--- a/usr.bin/patch/pch.c
+++ b/usr.bin/patch/pch.c
@@ -1478,9 +1478,8 @@ do_ed_script(void)
 			if (*t == 's') {
 				for (;;) {
 					continuation = 0;
-					t = strchr(buf, '\0') - 1;
-					while (--t >= buf && *t == '\\')
-						continuation = !continuation;
+					for (t = buf; *t && *t == '\\'; t++)
+						continuation ^= 1;
 					if (!continuation ||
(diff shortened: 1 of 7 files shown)


usr.bin/patch/pch.c
```diff
@@ -1478,9 +1478,8 @@ do_ed_script(void)
 			if (*t == 's') {
 				for (;;) {
 					continuation = 0;
-					t = strchr(buf, '\0') - 1;
-					while (--t >= buf && *t == '\\')
-						continuation = !continuation;
+					for (t = buf; *t && *t == '\\'; t++)
+						continuation ^= 1;
```
rilysh, line 1482: Why `continuation ^= 1`? This makes understanding the logic even harder. IMO the previous one is better here.

igalic (changes requested):
a lot of this feels like code churn, and it's doing way more than the commit message says

it's also not split one commit per subsystem. that's the absolute bare minimum for this PR to be considered

bsdimp:
Since this doesn't solve an actual problem, make things demonstrably faster, etc. I'm going to close it. IT's too churny and only introduces risk of regression.

PR #1023, opened 2024-01-11
traceroute6: remove -l flag

llfw:
The -l flag was used to tell traceroute6(8) to show both hostname and address for each hop.  However, traceroute(8) already does this by default, and there's no reason for traceroute6 to behave differently.

Make this the default behaviour, and accept -l for backward compatibility as a no-op flag.


llfw:
someone suggested this behaviour might be because IPv6 addresses are longer, however in pratice, there's basically no difference:

```
[1? 7!] lexi@hemlock ~
% traceroute6 -Il bbc.co.uk
traceroute6: Warning: bbc.co.uk has multiple addresses; using 2a04:4e42:400::81
traceroute6 to bbc.co.uk (2a04:4e42:400::81) from 2001:8b0:aab5:106::12, 64 hops max, 20 byte packets
 1  VLAN106.CSW1.EDEN.LE-FAY.ORG (2001:8b0:aab5:106::1)  0.292 ms  0.225 ms  0.200 ms
 2  SFP1.CR1.EDEN.LE-FAY.ORG (2001:8b0:aab5:fffd::2)  0.263 ms  0.217 ms  0.191 ms
 3  y.witless.thn.aa.net.uk (2001:8b0:0:53::134)  6.176 ms  6.237 ms  6.462 ms
 4  o.aimless.tch.aa.net.uk (2001:8b0:0:53::105)  5.795 ms  6.751 ms  6.528 ms
 5  2001:7f8:4::d361:1 (2001:7f8:4::d361:1)  6.680 ms  7.126 ms  6.844 ms
 6  2a04:4e42:400::81 (2a04:4e42:400::81)  6.636 ms  6.884 ms  6.956 ms

[8!] lexi@hemlock ~
% traceroute -I bbc.co.uk
traceroute: Warning: bbc.co.uk has multiple addresses; using 151.101.0.81
traceroute to bbc.co.uk (151.101.0.81), 64 hops max, 48 byte packets
 1  VLAN106.CSW2.EDEN.LE-FAY.ORG (10.1.6.1)  0.268 ms  0.181 ms  0.155 ms
 2  SFP1.CR1.EDEN.LE-FAY.ORG (10.255.0.6)  0.191 ms  0.176 ms  0.176 ms
 3  y.witless.thn.aa.net.uk (90.155.53.134)  5.985 ms  5.591 ms  6.007 ms
 4  o.aimless.tch.aa.net.uk (90.155.53.105)  6.103 ms  5.742 ms  5.804 ms
 5  195.66.225.91 (195.66.225.91)  6.257 ms  6.631 ms  6.391 ms
 6  151.101.0.81 (151.101.0.81)  6.179 ms  6.237 ms  6.299 ms
```

the only long line of output from traceroute6 is because it prints the source address by default.

jlduran:
Nice catch!
I would have attempted to fix it first. Take a look at [netstat(1)](https://github.com/freebsd/freebsd-src/blob/main/usr.bin/systat/netstat.c)'s same (and working) `inetname()` implementation:
https://github.com/freebsd/freebsd-src/blob/ed1a88a3116a59b4fd37912099a575b4c8f559dc/usr.bin/systat/netstat.c#L561-L603
Given the time, and the fact that nobody noticed it, removing it could also be another option, although, more cleanup is needed.

jlduran:
Sorry for the noise, originally I had misinterpreted the pull request.
I like the idea of unifying the output of The KAME project network utilities with the original BSD ones, or even better, unifying common code.

llfw:
i assume the eventual plan is to merge traceroute6 into traceroute, as was done with ping, but that's obviously a much larger project :-)  in the mean time, removing behaviour differences seems useful to reduce the amount of future work needed for that.

bsdimp:
This looks sane to me as well.
I'll see if it still applies.
I'll also add a date bump to the man page (hmmm, maybe we should script that)

jlduran:
There is also a mention to `traceroute6 -l` in `share/examples/IPv6/USAGE`, this is a very outdated document (historical?). Perhaps it should be updated as well .

bsdimp:
> There is also a mention to `traceroute6 -l` in `share/examples/IPv6/USAGE`, this is a very outdated document (historical?). Perhaps it should be updated as well .

The document hasn't meaningfully changed since 2002. I think updating it for this might be misguided effort. Maybe we should just retire it, though the network folks can say better than I.

llfw:
that document spends a lot of time talking about a `prefix` command that doesn't seem to exist anymore.  rather than updating it, it might be more sensible to remove it in favour of the Handbook, which already explains most of this and is more up-to-date.

llfw:
i pushed a rebase (no conflicts) and the .Dd update (this would be a handy thing to check in a git-commit hook, if we have any of those?).

bsdimp:
Pushed in as 0a49be722fe0

PR #1028, opened 2024-01-11
fix - fix typo in cpuset.1

chiajungchang2003:
In man page cpuset(1), policy should be policies at line 147 and "prefer policy" should be "preferred policy" at line 154
- Event: Advanced UNIX programming course (Fall'23) at NTHU


lwhsu:
Please include `Event: Advanced UNIX programming course (Fall'23) at NTHU` in the last line of your comment message. You can update your git config and do a git commit --amend and push -f to update your branch and the pull request.

lwhsu:
The original texts are correct. The first "policy" should be single, we may add "the" before it  but it's fine. The second "prefer" is the name of the policy.

PR #1043, opened 2024-01-12
lib/libunbound/config.h: Modified comments

fffelix-huang:
name: Ting-Hsuan Huang
email: felixhuang07@gmail.com
Event: from the Advanced UNIX Programming Course (Fall’23) at NTHU.

Most of the lines use `Define to 1 if you ...` instead of `Define if you`, therefore we have modified it to be consistent with other lines.

### DIFF
--- a/lib/libunbound/config.h
+++ b/lib/libunbound/config.h
@@ -484,7 +484,7 @@
 /* Define to 1 if you have the <poll.h> header file. */
 #define HAVE_POLL_H 1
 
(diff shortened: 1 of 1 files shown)


igalic:
(lib)unbound is an upstream project.
config.h looks to be generated from this file, https://github.com/NLnetLabs/unbound/blob/master/config.h.in

lwhsu:
This is a generated file so I don't think we should change it directly.

PR #1057, opened 2024-01-13
service(8): direct user to rc(8) for a list of valid commands

llfw:
New users may refer to service(8) to discover how to manage services,
but this manpage does not explain which commands are permitted besides
start/stop (for example, 'enable').  Add a paragraph that directs the
reader to rc(8) to discover this.

MFC after: 2 weeks

### DIFF
--- a/usr.sbin/service/service.8
+++ b/usr.sbin/service/service.8
@@ -22,7 +22,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd November 9, 2021
+.Dd January 23, 2024
 .Dt SERVICE 8
 .Os
 .Sh NAME
@@ -63,6 +63,13 @@ see
 .Pc .
 It can also be used to list
 the scripts using various criteria.
+.Pp
+The set of permissible values for
+.Ar command
+depends on the particular rc.d script being invoked.
+For a list of standard commands which are supported by most rc.d
+scripts, see
+.Xr rc 8 .
 .Pp
 The options are as follows:
 .Bl -tag -width F1
@@ -129,13 +136,44 @@ option is used, the corresponding variable is set accordingly.
 .Sh EXIT STATUS
 .Ex -std
 .Sh EXAMPLES
-The following are examples of typical usage of the
-.Nm
-command:
+These are some examples of the most common service commands.
+For a full list of commands available in most rc.d scripts, see
+.Xr rc 8 .
+.Pp
+Enable a service, then start it:
+.Bd -literal -offset -indent
+service sshd enable
+service sshd start
+.Ed
+.Pp
+Stop a service, then disable it:
+.Bd -literal -offset -indent
+service sshd stop
+service sshd disable
+.Ed
+.Pp
+Start a service which is not enabled:
+.Bd -literal -offset -indent
+service sshd onestart
+.Ed
+.Pp
+Report the status of a service:
 .Bd -literal -offset -ident
 service named status
-service -j dns named status
+.Ed
+.Pp
+Restart a service running in a jail:
+.Bd -literal -offset -indent
+service -j dns named restart
+.Ed
+.Pp
+Start a service with a specific environment variable set:
+.Bd -literal -offset -indent
 service -E LC_ALL=C.UTF-8 named start
+.Ed
+.Pp
+Report a verbose listing of all available services:
+.Bd -literal -offset -indent
 service -rv
 .Ed
 .Pp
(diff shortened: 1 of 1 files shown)


concussious:
Is it appropriate for me to say that it'd be very nice to mention sysrc(8) as well here? My experiences using service(8) are immediately preceeded by sysrc(8) at least 1/3rd of the time.

igalic:
maybe just outright mentioning the most common ones would also help

instead of `sysrc foo_enable=YES` i usually use `service foo enable`. unlike sysrc, this catches my typos.

concussious:
+1 for adding an `service __ enable` example.

I didn't know this was possible and is much better than what I was doing with sysrc, thank you.

llfw:
i also didn't realise `service X enable` was possible until someone on IRC mentioned that it's not documented.  i'll add a couple of examples to the PRs.

usr.sbin/service/service.8
```diff
@@ -161,10 +195,11 @@ _service () {
 complete -F _service service
 .Ed
 .Sh SEE ALSO
-.Xr bash 1 Pq Pa ports/shells/bash ,
 .Xr rc.conf 5 ,
 .Xr rc 8 ,
-.Xr rcorder 8
+.Xr sysrc 8 ,
+.Xr rcorder 8 ,
+.Xr bash 1 Pq Pa ports/shells/bash .
```
igalic, line 202: why is this at the end now?

usr.sbin/service/service.8
igalic, line 1: this file now definitely needs a `.Dd` bump

igalic (changes requested):

usr.sbin/service/service.8
llfw: because shells/bash isn't even in the base system, so it's the least useful Xr in the manpage and should be listed last.  to be honest, i'm not even sure why it's listed at all; if it's because of the completion example, should `csh(1)` be listed too?

usr.sbin/service/service.8
llfw: bumped Dd

usr.sbin/service/service.8
igalic: i dunno, i would remove ax it.

llfw: ok, removed!

usr.sbin/service/service.8
```diff
@@ -129,13 +136,40 @@ option is used, the corresponding variable is set accordingly.
...
-The following are examples of typical usage of the
-.Nm
-command:
+Enable a service, then start it:
+.Bd -literal -offset -indent
+service sshd enable
+service sshd start
+.Ed
+.Pp
+Disable a service, then stop it:
```
igalic, line 145: Unless you wanna show off `onestop` below, you should reverse the phrasing here.

igalic (commented):
minor nit, otherwise we're good to go

usr.sbin/service/service.8
llfw: are you sure?  usually `stop` would come before `disable`, because otherwise stop won't work.

or do you mean we should show an example of `disable` followed by `onestop`?

llfw: never mind, i see what you mean.  now fixed.

igalic (approved):
👍 thank you so much for this

llfw:
fixed a puncuation error introduced in the `SEE ALSO` section.

igalic (approved):
👍🏻

debdrup:
I'll have a look-see at this one :)

DanielO:
This looks good to me, although perhaps mention extracommands (but rc(8) covers this so I am ambivalent)

debdrup:
I like the idea of this, but have some feedback that's also covered by Daniel above.

> This looks good to me, although perhaps mention extracommands (but rc(8) covers this so I am ambivalent)

Yes, this is a good point - I'd much rather keep the initial paragraph, **and** make explicit reference to rc(8) for all subcommands available.
I think it also makes sense to add a note that makes it clear that the examples given are only some of the most commonly used ones.

Otherwise, it looks good to me.

llfw:
i'm inclined not to mention `extracommands` because this isn't really specific to `service(8)`, it's part of `rc(8)` itself, and i don't think we want `service(8)` to duplicate the entire rc-system documentation.

as a separate project, it might be nice if there were a single manual page called something like rc(5) that explained how to invoke an rc script, since this information is a bit hidden in the middle of `rc(8)` and that manpage in general is aimed more at people implementing scripts than people using them.

i will add some extra wording to the examples section to mention that other commands exist, though.

llfw:
added a paragraph to EXAMPLES mentioning there are more commands in rc(8).  that duplicates the first paragraph, but i think that's okay, since users might page straight to the examples section.

debdrup:
Thank you for the work, it has become merged ;)

ykla:
see also <https://github.com/freebsd/freebsd-src/commit/da45b4da83c6a68c3bc3a1682ec77d44141164d2>

PR #1058, opened 2024-01-14
wcsstr does not take restricted pointers

AtariDreams:
wcsstr is not supposed to have restrict qualifiers. No other library has it, and the C standard says it does not have them either.


bsdimp:
`POSIX` begs to differ
```
wchar_t *wcsstr(const wchar_t *restrict ws1,
       const wchar_t *restrict ws2);
```
From POSIX.1 ,Issue 6 and Issue 7. Both aligned with C99.
The draft for issue 8 also has restrict.

Although the C99 and C11 standards do not have the restrict qualifier for this function. C2x does something weird with generics I don't quite understand, but that's OK.

So, since POSIX has it, I'm just going to close this pull request. We do not offer a mode that's C compliant without POSIX.

PR #1059, opened 2024-01-14
fdwrite.c: initialize pointers to NULL and a few other cleanups

rilysh:
1. Both trackbuf and vrfybuf are initialized to zero (NULL). While it's okay to initialize pointers to zero, to keep consistency, as they're explicitly pointers, it's better to just use NULL ((void *)0) instead of 0 (both are equivalent to the compilers).
2. Call free() for both trackbuf and vrfybuf after their job has been done.
3. Remove the register keyword. Compilers generally ignore this keyword (except for very very old compilers and CPUs).
4. Remove the ctype.h header. It's not being used anywhere in the file.

### DIFF
--- a/usr.sbin/fdwrite/fdwrite.c
+++ b/usr.sbin/fdwrite/fdwrite.c
@@ -10,7 +10,6 @@
  *
  */
 
-#include <ctype.h>
 #include <err.h>
 #include <fcntl.h>
 #include <paths.h>
@@ -26,7 +25,7 @@ format_track(int fd, int cyl, int secs, int head, int rate,
      int gaplen, int secsize, int fill, int interleave)
 {
     struct fd_formb f;
-    register int i,j;
+    int i,j;
     int il[100];
 
     memset(il,0,sizeof il);
@@ -68,7 +67,7 @@ main(int argc, char **argv)
     int bpt, verbose=1, nbytes=0, track;
     int interactive = 1;
     const char *device= "/dev/fd0";
-    char *trackbuf = 0,*vrfybuf = 0;
+    char *trackbuf = NULL,*vrfybuf = NULL;
     struct fd_type fdt;
     FILE *tty;
 
@@ -192,5 +191,8 @@ main(int argc, char **argv)
     }
     if(verbose)
 	printf("%d bytes on %d flopp%s\n",nbytes,fdn,fdn==1?"y":"ies");
+
+    free(trackbuf);
+    free(vrfybuf);
     exit(0);
 }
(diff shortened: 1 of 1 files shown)


usr.sbin/fdwrite/fdwrite.c
```diff
@@ -26,7 +25,7 @@ format_track(int fd, int cyl, int secs, int head, int rate,
      int gaplen, int secsize, int fill, int interleave)
 {
     struct fd_formb f;
-    register int i,j;
+    int i,j;
     int il[100];
 
     memset(il,0,sizeof il);
```
igalic, line 31: while you're already here, why not format this according to `style(9)`.

igalic (approved):
looks sensible 👍

usr.sbin/fdwrite/fdwrite.c
rilysh: I intentionally left it behind as `style(9)` also says to indent the first level with 8 characters (tabs) and running `indent(1)` drastically changes the indentation (which I wasn't sure about whether should I apply or not)...

I think for now I'll apply it for these (CI) warnings only. If full indentation is acceptable, then I will add that too.
CI Link: https://github.com/freebsd/freebsd-src/actions/runs/7517809483/job/20464384723?pr=1059

bsdimp:
We have plans to remove floppy support soon.  Not sure it is worth the effort

rilysh:
> We have plans to remove floppy support soon. Not sure it is worth the effort

Whoops... I didn't know that, is there a current plan on which version of FreeBSD release floppy support will be stripped away? (e.g. the version, 14.0)

igalic:
14.0 has sailed, so even tho i haven't seen it on the plan for 15.0, the fact that we're getting rid of 32 bit platforms (in the kernel) kind of hints at other things that might fall to the ax

bsdimp:
> 14.0 has sailed, so even tho i haven't seen it on the plan for 15.0, the fact that we're getting rid of 32 bit platforms (in the kernel) kind of hints at other things that might fall to the ax

We've explicitly had fd on the list for a while. It won't be in 15. If we'd realized sooner, we'd have trimmed it for 14, but the trim list for that was already fairly large.

bsdimp:
I went ahead and pushed this...  Though in the future, a pull request like this would run into our new policy that say a change has to fix a real problem.

bsdimp:
damn, I gotta update my landing scripts to properly credit... I just noticed @igalic approved, which should have added a reviewed by line.

PR #1060, opened 2024-01-14
find: Allow '/' to be used with -perm for GNU compatibility

ricardobranco777:
Since 2005 GNU find(1) no longer allows '+' to be used with `-perm`, forcing the use of a `/` instead.

This PR adds support for this character as synonym for '+'.

Fixes https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=276326

### DIFF
--- a/usr.bin/find/find.1
+++ b/usr.bin/find/find.1
@@ -28,7 +28,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd December 22, 2023
+.Dd January 15, 2024
 .Dt FIND 1
 .Os
 .Sh NAME
@@ -757,7 +757,7 @@ Slashes
 .Pq Dq Li /
 are treated as normal characters and do not have to be
 matched explicitly.
-.It Ic -perm Oo Cm - Ns | Ns Cm + Oc Ns Ar mode
+.It Ic -perm Oo Cm - Ns | Ns Cm + Ns | Ns Cm / Oc Ns Ar mode
 The
 .Ar mode
 may be either symbolic (see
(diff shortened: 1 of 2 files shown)


igalic:
I feel like this would require a documentation update

ricardobranco777:
> I feel like this would require a documentation update

Fixed.

bsdimp:
Can you rebase to a newer main and do a `git push --force-with-lease` so that the style checker also runs.

To my eyeballs, though, this looks good.

ricardobranco777:
> --force-with-lease

Done.

usr.bin/find/find.1
igalic, line 1: this gilt will need a  .Dd bump

igalic (changes requested):
one final request
looks fine otherwise

igalic (changes requested):
one final request

qemu-bsd-user:
Once the .Dd bump happens, I'll commit. It all looks good otherwise.
I also need to look at why my style checker didn't run with the forced push, but that's not the fault of the submitter...

usr.bin/find/find.1
ricardobranco777: Fixed. Thanks!

ricardobranco777:
> Once the .Dd bump happens, I'll commit. It all looks good otherwise. I also need to look at why my style checker didn't run with the forced push, but that's not the fault of the submitter...

Fixed.

jlduran:
> I also need to look at why my style checker didn't run with the forced push

@bsdimp disguised as @qemu-bsd-user, maybe it is missing the `synchronize` activity type?
Ref: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request, yes it is the only activity type not in the past tense.

emaste:
I find `preceded by a plus (“+”) or a slash (“/”) (for GNU compatibility),` slightly awkward (although this entire `-perm` paragraph is already awkward).

I think `/` is not accepted by find on other BSDs or macOS, so we likely still prefer `+`? If the intent is to allow `/` for GNU compatibility but we feel that `+` is preferred I would probably leave the existing text intact, but add a sentence at the end along the lines of `For compatibility with GNU find a slash ("/") is also accepted in place of a plus ("+")`.

In any case the code change LGTM (and thanks for improving GNU compatibility).

bsdimp:
Merged. I hope it's OK, but I added a blurb in the STANDARDS section about this extension.

PR #1061, opened 2024-01-15
document most commong KERNCONFs

igalic:
document GENERIC and MINIMAL and their relatives.
How they are enabled, how they are used.

-----

Following the discussion in #1020 , I would like to start documenting our most common KERNCONFs and how they are used.

### DIFF
--- a/share/man/man9/MINIMAL.9
+++ b/share/man/man9/MINIMAL.9
@@ -1,0 +1,35 @@
+.\"
+.\" SPDX-License-Identifier: BSD-2-Clause
+.\"
+.\" Copyright (c) 2024 Mina Galić
+.\"
+.\" This manual page was written by Mina Galić <FreeBSD@igalic.co>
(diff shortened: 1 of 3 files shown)


share/man/man9/MINIMAL.9
```diff
@@ -0,0 +1,35 @@
...
+The
+.Va MINIMAL
+kernel is the minimal most kernel configuration that results in a bootable
+system on Tier 1 architectures.
+Additionally, the
+.Va MINIMALUP
+kernel is provided as a compile target for testing configurations without
+.Xr SMP 4 ,
+and without
+.Xr VNET 9
```
bsdimp, line 28: VNET is now in MINIMAL

bsdimp:
I think it would be better to go in a slightly different direction.
Let's create a kernconf(9) that talks about the common config files, and then cross references config(5).

PR #1062, opened 2024-01-15
man: avoid unportable use of utilities

mohd-akram:
These flags are not documented.


bsdimp:
This change is good. I wrote a longer commit message explaining what was unportable and what changes were made. Please try to do that for future commits. Everything else about the commit was correct.

bsdimp:
> Thanks, noted.

Thanks for finding these and fixing the problems.

PR #1064, opened 2024-01-15
GitHub: style: Add the synchronize activity type

jlduran:
@bsdimp Here is what I believe is the missing piece for the style check to re-run upon `git push --force-with-lease` et al.


bsdimp:
Love it.

bsdimp:
Woot! Thanks. Landed.

PR #1066, opened 2024-01-17
Fixes to make bootstrapping from Linux/musl work

valpackett:
Here's a bunch of fixes that allowed me to build FreeBSD under @chimera-linux which uses musl libc and a port of our own core programs. The more complex changes are definitely subject to improvements, suggestions/bikeshedding welcome!

+also required: https://github.com/openzfs/zfs/pull/15780

### DIFF
--- a/contrib/mandoc/config.h
+++ b/contrib/mandoc/config.h
@@ -13,7 +13,7 @@
 #define HAVE_ENDIAN 0
 #define HAVE_ERR 1
 #define HAVE_FTS 1
-#if defined(__GLIBC__) || defined(__APPLE__)
+#if defined(__linux__) || defined(__APPLE__)
 #define HAVE_FTS_COMPARE_CONST 0
 #else
 #define HAVE_FTS_COMPARE_CONST 1
--- a/kerberos5/include/config.h
+++ b/kerberos5/include/config.h
@@ -1,3 +1,5 @@
+#include <resolv.h>
+
 /* include/config.h.  Generated from config.h.in by configure.  */
 /* include/config.h.in.  Generated from configure.ac by autoheader.  */
 
@@ -694,7 +696,9 @@
 #define HAVE_RES_NDESTROY 1
 
 /* Define to 1 if you have the `res_nsearch' function. */
+#if __RES >= 19991006
 #define HAVE_RES_NSEARCH 1
+#endif
 
 /* Define to 1 if you have the `res_search' function. */
 #define HAVE_RES_SEARCH 1
--- a/lib/libpmc/pmu-events/jevents.c
+++ b/lib/libpmc/pmu-events/jevents.c
@@ -1355,7 +1355,7 @@ err_out:
 #include <fts.h>
 
 static int
-#if defined(__GLIBC__) || defined(__APPLE__)
+#if defined(__linux__) || defined(__APPLE__)
 fts_compare(const FTSENT **a, const FTSENT **b)
 #else
 fts_compare(const FTSENT * const *a, const FTSENT * const *b)
--- a/tools/build/cross-build/include/common/sys/_types.h
+++ b/tools/build/cross-build/include/common/sys/_types.h
@@ -45,3 +45,6 @@
  * Neither GLibc nor macOS define __va_list but many FreeBSD headers require it.
  */
 typedef __builtin_va_list __va_list;
+
+/* Needed for opensolaris compat. */
+typedef	__int64_t	off64_t;
--- a/tools/build/cross-build/include/mac/sys/_types.h
+++ b/tools/build/cross-build/include/mac/sys/_types.h
@@ -41,5 +41,3 @@
  * __darwin_ct_rune_t exists.
  */
 typedef __darwin_ct_rune_t __ct_rune_t;
-/* Needed for opensolaris compat. */
-typedef	__int64_t	off64_t;
--- a/tools/build/cross-build/progname.c
+++ b/tools/build/cross-build/progname.c
@@ -36,7 +36,6 @@
 #include <sys/cdefs.h>
 #include <stdlib.h>
 
-#ifdef __GLIBC__
 extern const char *__progname;
 const char *
 getprogname(void)
@@ -48,4 +47,3 @@ setprogname(const char *progname)
 {
 	__progname = progname;
 }
-#endif /* __GLIBC__ */
--- a/tools/build/mk/Makefile.boot
+++ b/tools/build/mk/Makefile.boot
@@ -46,6 +46,10 @@ CFLAGS+=	-I${SRCTOP}/tools/build/cross-build/include/linux
 CFLAGS+=	-D_GNU_SOURCE=1
 # Needed for sem_init, etc. on Linux (used by usr.bin/sort)
 LDADD+=	-pthread
+.if exists(/usr/lib/libfts.so) || exists(/usr/lib/libfts.a) || exists(/lib/libfts.so) || exists(/lib/libfts.a)
+# Needed for fts_open, etc. on musl (used by usr.bin/grep)
+LDADD+=	-lfts
+.endif
 
 .elif ${.MAKE.OS} == "Darwin"
 CFLAGS+=	-D_DARWIN_C_SOURCE=1
--- a/tools/build/mk/Makefile.boot
+++ b/tools/build/mk/Makefile.boot
@@ -41,10 +41,6 @@ CWARNFLAGS.clang+=-Wno-typedef-redefinition
 CWARNFLAGS+=	-Wno-system-headers
 CWARNFLAGS.clang+=-Werror=incompatible-pointer-types-discards-qualifiers
 
-# b64_pton and b64_ntop is in libresolv on MacOS and Linux:
-# TODO: only needed for uuencode and uudecode
-LDADD+=-lresolv
-
 .if ${.MAKE.OS} == "Linux"
 CFLAGS+=	-I${SRCTOP}/tools/build/cross-build/include/linux
 CFLAGS+=	-D_GNU_SOURCE=1
--- a/usr.bin/grep/grep.c
+++ b/usr.bin/grep/grep.c
@@ -62,9 +62,8 @@ const char	*errstr[] = {
 /* 8*/	"%s (BSD grep, GNU compatible) %s\n",
 };
 
-/* Flags passed to regcomp() and regexec() */
+/* Flags passed to regcomp() */
 int		 cflags = REG_NOSUB | REG_NEWLINE;
-int		 eflags = REG_STARTEND;
 
 bool		 matchall;
 
--- a/usr.bin/localedef/bootstrap/ctype.h
+++ b/usr.bin/localedef/bootstrap/ctype.h
@@ -44,13 +44,23 @@
  * However, only do so if they are not already define as a macro (as is the
  * case on Linux).
  */
-#ifndef isxdigit
 __BEGIN_DECLS
+#ifndef isalpha
 int	isalpha(int);
-int	isascii(int);
-int	isdigit(int);
-int	isgraph(int);
-int	islower(int);
-int	isxdigit(int);
-__END_DECLS
 #endif
+#ifndef isascii
+int	isascii(int);
+#endif
+#ifndef isdigit
+int	isdigit(int);
+#endif
+#ifndef isgraph
+int	isgraph(int);
+#endif
+#ifndef islower
+int	islower(int);
+#endif
+#ifndef isxdigit
+int	isxdigit(int);
+#endif
+__END_DECLS
--- a/usr.bin/sed/process.c
+++ b/usr.bin/sed/process.c
@@ -68,8 +68,7 @@ static inline int	 applies(struct s_command *);
 static void		 do_tr(struct s_tr *);
 static void		 flush_appends(void);
 static void		 lputs(char *, size_t);
-static int		 regexec_e(regex_t *, const char *, int, int, size_t,
-			     size_t);
+static int		 regexec_e(regex_t *, char *, int, int, size_t, size_t);
 static void		 regsub(SPACE *, char *, char *);
 static int		 substitute(struct s_command *);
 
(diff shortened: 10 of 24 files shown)


bsdimp:
This looks quite interesting... i like parts and have concerns about parts. Ill go through it with a fine toothed comb.

emaste (commented):
style(9) disagrees with this change, `(<sys/param.h> includes <sys/types.h>; do not include both.)` so we'll need to address that somehow -- possibly by amending style to include something like `do not include both, unless required for portability`.

(gh code review does not make it clear the code review comment is going to be left on the general conversation, not the specific file or commit -- this is with respect to 032c239e83ef2917921153ecb5daa6ab1abd4056 sys/capsicum.h)

kerberos5/include/config.h
```diff
@@ -694,7 +696,9 @@
 #define HAVE_RES_NDESTROY 1
 
 /* Define to 1 if you have the `res_nsearch' function. */
+#if __RES >= 19991006
```
emaste, line 699: cc @cschuber - is there any automation involved in the generation of config.h on krb5 updates?

emaste:
jevents.c should coordinate with upstream

tools/build/cross-build/include/common/langinfo.h
```diff
@@ -0,0 +1,46 @@
...
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include_next <langinfo.h>
+
+#ifndef ALTMON_1
```
jrtc27, line 33: https://github.com/freebsd/freebsd-src/commit/bd234c0d4c8256db7e5a1fdda9ef311c9e0080e4 should have made these no longer required (they aren't present on macOS yet things build just fine today). Are you sure you need them, and if so, why?

bsdimp:
> style(9) disagrees with this change, `(<sys/param.h> includes <sys/types.h>; do not include both.)` so we'll need to address that somehow -- possibly by amending style to include somethin


We need to document nonstandard usage.

I have a lot of other comments as well. Will do them after bowling.

bsdimp (changes requested):
Please don't land anything else from this PR. I have several comments pending i wanted to think through.

tools/build/cross-build/include/common/langinfo.h
valpackett: I've definitely started working on this before pulling that commit in. Works fine without the file now, I'll remove it.

bsdimp:
```
diff --git a/tools/build/cross-build/include/linux/sys/types.h b/tools/build/cross-build/include/linux/sys/types.h
index 770b3057f8a8..e72e1419bccd 100644
--- a/tools/build/cross-build/include/linux/sys/types.h
+++ b/tools/build/cross-build/include/linux/sys/types.h
@@ -35,6 +35,11 @@
  */
 #pragma once
 
+/*
+ * FreeBSD's types.h always includes sys/cdefs.h, but Linux headers often don't.
+ */
+#include <sys/cdefs.h>
+
 #include_next <sys/types.h>
 
 #ifndef __size_t
```
I think will fix the issue...

contrib/mandoc/config.h
```diff
@@ -13,7 +13,7 @@
 #define HAVE_ENDIAN 0
 #define HAVE_ERR 1
 #define HAVE_FTS 1
-#ifdef __GLIBC__
+#ifdef __linux__
```
bsdimp, line 16: This likely should be upstraemed, but it changes a local change, so it's OK.Both the changes need to go up.... :)

include/nl_types.h
```diff
@@ -34,6 +34,7 @@
 #ifndef _NL_TYPES_H_
 #define _NL_TYPES_H_
 
+#include <sys/cdefs.h>
```
bsdimp, line 37: I commented on this.

kerberos5/include/config.h
bsdimp: This likely needs a comment, but this doesn't feel like the right place to solve this...

lib/libpmc/pmu-events/jevents.c
```diff
@@ -1355,7 +1355,7 @@ int main(int argc, char *argv[])
 #include <fts.h>
 
 static int
-#ifdef __GLIBC__
+#ifdef __linux__
```
bsdimp, line 1358: this looks good. __GLIBC__ vs __linux__ seems sane.

sys/sys/capsicum.h
```diff
@@ -40,6 +40,7 @@
 #define	_SYS_CAPSICUM_H_
 
 #include <sys/param.h>
+#include <sys/types.h>
```
bsdimp, line 43: This should likely just be added to `./tools/build/cross-build/include/linux/sys/param.h`

usr.bin/grep/grep.c
```diff
@@ -62,9 +62,8 @@ const char	*errstr[] = {
 /* 8*/	"%s (BSD grep, GNU compatible) %s\n",
 };
 
-/* Flags passed to regcomp() and regexec() */
+/* Flags passed to regcomp() */
 int		 cflags = REG_NOSUB | REG_NEWLINE;
-int		 eflags = REG_STARTEND;
```
bsdimp, line 67: This is a very odd change.... It should be #ifdef linux or something, no?

bsdimp (commented):
Mostly I'm wondering why the sed/grep changes aren't conditional.

bsdimp:
Also, if the changes to use cross-tools includes work... then maybe we can eliminate some of the added ctypes.h includes.

usr.bin/grep/grep.c
valpackett: There's `#ifdef REG_STARTEND` but inside of `util.c`. There it's choosing between using the flag if it exists, and a fallback manual implementation in the `#else`. With that, the global default for the flags doesn't quite make sense.

tools/build/cross-build/include/common/pwd.h
```diff
@@ -51,6 +51,7 @@ int uid_from_user(const char *name, uid_t *uid);
 const char *user_from_uid(uid_t uid, int noname);
 
 #ifdef __linux__
+#include <sys/cdefs.h>
```
arichardson, line 54: I guess this is just for the __unused? Could add a cast to void inside the function body instead?

bsdimp:
Ping? What's up with this? Sounded like you were preparing an updated, but it doesn't look like one's been pushed.
I don't have this environment, so I can't easily mine this for changes and commit piecemeal.
Can you rebase and update the PR please? Despite all the fiddly comments, I think this would be good to have.
If the upstream bits are getting in the way, maybe just resubmit those w/o them, we'll get that in and then see what
to do about upstream?
Thanks!

valpackett:
Oop… yeah, sorry, was so busy with all the everything again. Thanks for the ping. Rebased, removed langinfo, did the cdefs-via-types thing instead of touching capsicum.h.

BTW, feel free to do more cherry-picking-and-pushing of commits you'd consider uncontroversial, the PR getting smaller always makes it better to work with (:

tools/build/cross-build/include/common/sys/_types.h
```diff
@@ -45,3 +45,6 @@
  * Neither GLibc nor macOS define __va_list but many FreeBSD headers require it.
  */
 typedef __builtin_va_list __va_list;
+
+/* Needed for opensolaris compat. */
+typedef	__int64_t	off64_t;
```
bsdimp, line 50: this is fine now, but may cause some grief in the future when I land our better compat with open solaris off64_t stuff

tools/build/cross-build/include/mac/sys/_types.h
```diff
@@ -41,5 +41,3 @@
  * __darwin_ct_rune_t exists.
  */
 typedef __darwin_ct_rune_t __ct_rune_t;
-/* Needed for opensolaris compat. */
-typedef	__int64_t	off64_t;
```
bsdimp, line 45: I don't understand this one...

tools/build/cross-build/progname.c
```diff
@@ -36,7 +36,6 @@
 #include <sys/cdefs.h>
 #include <stdlib.h>
 
-#ifdef __GLIBC__
```
bsdimp, line 39: or this

usr.bin/localedef/bootstrap/ctype.h
```diff
@@ -44,13 +44,23 @@
  * However, only do so if they are not already define as a macro (as is the
  * case on Linux).
  */
-#ifndef isxdigit
 __BEGIN_DECLS
+#ifndef isalpha
 int	isalpha(int);
```
bsdimp, line 49: quick question here: there's two ways to fix this: one is how you've done (and is likely fine for bootstrap), the other is to make these int (isalpha)(int); to prevent the macro expansion. Is there a reason to prefer one over the other?

bsdimp (commented):
just a few notes... I'll see what I can land from this and we'll iterate.

usr.bin/sed/process.c
```diff
@@ -672,11 +671,23 @@ regexec_e(regex_t *preg, const char *string, int eflags, int nomatch,
 		defpreg = preg;
 
 	/* Set anchors */
+#ifndef REG_STARTEND /* in cross-build tools */
```
arichardson, line 674: This looks different from the wrapper in grep.c (the former does not have a loop). Would it make sense to pull this out into a shared helper?

tools/build/cross-build/include/mac/sys/_types.h
jrtc27: Moved to common given it’s needed for musl, it seems

tools/build/cross-build/progname.c
jrtc27: I guess because tools/build/Makefile only builds it on Linux, so it was previously a stupid ifdef, and now is unhelpful for musl.

tools/build/mk/Makefile.boot
```diff
@@ -41,15 +41,15 @@ CWARNFLAGS.clang+=-Wno-typedef-redefinition
...
-# b64_pton and b64_ntop is in libresolv on MacOS and Linux:
-# TODO: only needed for uuencode and uudecode
-LDADD+=-lresolv
-
 .if ${.MAKE.OS} == "Linux"
 CFLAGS+=	-I${SRCTOP}/tools/build/cross-build/include/linux
 CFLAGS+=	-D_GNU_SOURCE=1
 # Needed for sem_init, etc. on Linux (used by usr.bin/sort)
 LDADD+=	-pthread
+.if exists(/usr/lib/libfts.so) || exists(/usr/lib/libfts.a) || exists(/lib/libfts.so) || exists(/lib/libfts.a)
```
jrtc27, line 49: This feels dodgy

usr.bin/localedef/bootstrap/ctype.h
jrtc27: I guess you shouldn't have a prototype for something that's implemented as a macro, since it doesn't really exist

jrtc27 (commented):
My general comment is that there's very little by way of explanation and justification, with a lot of unrelated changes being lumped into a single commit that has no body in the message. Each change should document *why* it's needed, and be limited to one thing. As it stands for a lot of these I look at the diff and can only really speculate on why they're needed.

bsdimp:
> @bsdimp Please liaise with me before landing any of this, I would like to ensure I review it first

Will do. The changes themselves likely are OK, but I share your concerns about justification and regressions, though the cross build github jobs work. The changes also aren't chunked up quite right, and need to be regroups a little. I'll post phab reviews and CC you.  I need to find a musl build env...

tools/build/cross-build/include/mac/sys/_types.h
valpackett: Exactly

tools/build/cross-build/progname.c
arichardson: I feel like this file might have been compiled for macos in the past so the ifdef was there to only use it for Linux. Dropping it looks good.

contrib/mandoc/config.h
```diff
@@ -13,7 +13,7 @@
 #define HAVE_ENDIAN 0
 #define HAVE_ERR 1
 #define HAVE_FTS 1
-#if defined(__GLIBC__) || defined(__APPLE__)
+#if defined(__linux__) || defined(__APPLE__)
```
bsdimp, line 16: Is this really #ifndef __FreeBSD__? Or does NetBSD/OpenBSD also have this new const interface?

bsdimp:
OK. To make progress on this, I've setup a Alpine Linux VM. I've started mining this PR for bits of code that can be committed. I've also added a few comments for some weird things that I've noticed. I'm testing this now. https://reviews.freebsd.org/D45349 is the first of 7 reviews that is the result. These are all uninteresting bits. I added Val as a co-author on almost all of them. I could swap author and co-author on them though. I wrote either extra comments, or a commit message for these, so I think it's only fair to share the credit and I'm agnostic on how. The co-authored-by stuff was the path of least resistance for me, but let's chat if that bothers anybody. It also slices up what's here a bit differently, but not that differently. I hope Co-authored-by: is the right git trailer to use for github to give co-credit. If not, I'll change them to whatever does.

I added @jrtc27 to the reviews as well.  My hope is that we can commit those reviews, close this review out and then do the rest of this pull request with new, smaller pull requests. The size of this request made it linger far too long and I'd like to try to get things in faster, which means I think that future pull requests should be smaller and more easy to manage. There's just too many moving parts otherwise.

bsdimp:
OK. All three builds succeeded in the musl branch I pushed. Though on alpine Linux 'success' was defined to be 'made it as far as the hack and slash version I did to validate what I cleaned up and put into the musl branch'. The stat stuff looks ready too, I overlooked it earlier and I'm out of time today. The base64 stuff, the resolver stuff, the pthreads stuff and the regexp stuff I don't pretend to grok all remain, each of which may need some tweaking and I'd suggest should be a separate future pull request in the future to keep things from getting bogged down.

bsdimp:
so, it's time to close this one. I've mined all the best bits out. Maybe there's more to get (like the stat.h patch), but I'm going to close this and we'll start on the other stuff from that base.

PR #1068, opened 2024-01-17
kerberos: Fix numerous segfaults when using weak crypto

cschuber:
NOTE: This PR's only purpose is to test this proposed commit against Linux and MacOS. Do not commit the PR.

Weak crypto is provided by the openssl legacy provider which is not load by default. Load the legacy providers as needed.

When the legacy provider is loaded into the default context the default provider will no longer be automatically loaded. Without the default provider the various kerberos applicaions and functions will abort().

This is the second attempt at this patch. Instead of linking secure/lib/libcrypto at build time we now link it at runtime, avoiding buildworld failures under Linux and MacOS. This is because TARGET_ENDIANNESS is undefined at pre-build time.

PR:			272835
MFC after:		3 days
Tested by:		netchild, Joerg Pulz <Joerg.Pulz@frm2.tum.de>

### DIFF
--- a/crypto/heimdal/lib/kadm5/create_s.c
+++ b/crypto/heimdal/lib/kadm5/create_s.c
@@ -169,6 +169,10 @@ kadm5_s_create_principal(void *server_handle,
     ent.entry.keys.len = 0;
     ent.entry.keys.val = NULL;
 
+    ret = fbsd_ossl_provider_load();
+    if (ret)
+	goto out;
+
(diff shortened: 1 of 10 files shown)


cschuber:
NOTE: This PR's only purpose is to test this proposed commit against Linux and MacOS builds. Do not commit the PR.

cschuber:
Cross-builds were successful.

PR #1069, opened 2024-01-17
add testcase for /bin/hostname

LeeLin2602:


igalic (changes requested):
this looks okay to my eye, except for one line

bin/hostname/Makefile
```diff
@@ -1,5 +1,11 @@
+#	@(#)Makefile	8.1 (Berkeley) 5/31/93
```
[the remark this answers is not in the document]
LeeLin2602: It was removed in [90aea514c6249118e880d75972d063362f4bf492](https://github.com/freebsd/freebsd-src/commit/90aea514c6249118e880d75972d063362f4bf492). I modified the source code from FreeBSD 14.0 Release hence it does still exist in my code. I will delete it later. Thank you for reminder.

bin/hostname/tests/hostname_test.sh
```diff
@@ -0,0 +1,51 @@
...
+recycle(){
+  jail -f "$root_dir/jail.conf" -r test-hostname-jail
+  rm "$root_dir/jail.conf"
+}
+
+basic_head()
+{
+  atf_set "descr" "main test"
+}
+
```
jlduran, line 36: I usually see —as documented in `atf-sh(3)`— that `head`, `body`, and `cleanup` (optionally), don't get this space between subroutines.

bin/hostname/tests/hostname_test.sh
```diff
@@ -0,0 +1,51 @@
...
+#
+
+#
+# These tests need to run in a multibyte locale with non-localized
+# error messages.
+#
+export LC_CTYPE=C.UTF-8
+export LC_MESSAGES=C
+
+root_dir='/usr/tests/bin/hostname'
```
jlduran, line 14: Maybe `root_dir="$(atf_get_srcdir)"`?
What if I do not run the tests under `/usr/tests`?

jlduran (commented):
All minor (and largely pedantic) nits. Thank you!

bin/hostname/tests/hostname_test.sh
jlduran: Maybe even under `.`? (Then temporary directory created by kyua).

LeeLin2602: I think `root_dir="$(atf_get_srcdir)"` is a little bit better for such case that interrupting causes the files not deleted?

jlduran: > I think `root_dir="$(atf_get_srcdir)"` is a little bit better for such case that interrupting causes the files not deleted?

That is the main reason I also suggested adding a `basic_cleanup()` with `recycle()` inside it (and not in `basic_body()`), so that the file is always removed, even when the test fails.

bin/hostname/tests/hostname_test.sh
```diff
@@ -0,0 +1,51 @@
+#
```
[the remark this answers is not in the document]
LeeLin2602: I don't see any other files doing so, could you please list one?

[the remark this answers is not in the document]
jlduran: > I don't see any other files doing so, could you please list one?

Too many files, just a few that I glanced over recently:
tests/sys/netinet6/*.sh
usr.sbin/makefs/tests/makefs_zfs_tests.sh

Non-test files:
usr.sbin/freebsd-update/freebsd-update.sh

Try:
```
grep "^#-$" **/*.sh
```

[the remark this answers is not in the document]
LeeLin2602: Thank you for information. I am just curious.

bin/hostname/tests/hostname_test.sh
```diff
@@ -0,0 +1,51 @@
...
+init(){
+  echo "$test_jail_conf" > "$root_dir/jail.conf"
+  jail -f "$root_dir/jail.conf" -c test-hostname-jail
+}
+
+recycle(){
+  jail -f "$root_dir/jail.conf" -r test-hostname-jail
+  rm "$root_dir/jail.conf"
+}
+
```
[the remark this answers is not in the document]
jlduran: Missing:
```
atf_test_case basic cleanup
```
on top

bin/hostname/tests/hostname_test.sh
jlduran, line 1: Sorry to nag over this minor detail, what I meant was:
```sh
atf_test_case basic cleanup
basic_head()
{
...
}
basic_body()
{
...
}
basic_cleanup()
{
...
}

atf_init_test_cases()
{
	atf_add_test_case basic
}
```
Per example in [`atf-sh(3)`](https://man.freebsd.org/cgi/man.cgi?query=atf-sh&apropos=0&sektion=3&arch=default&format=html), saying "Shell-based test programs always follow this template".

Thank you for the contribution, it should be fine anyway.

LeeLin2602: Please check again. Thank you.

jlduran: Looking good! Does this test need root permissions? If so, you should add:
```
atf_set require.user root
```
to the `basic_head()` subroutine (Per `atf-test-case(4)`).

bin/hostname/tests/hostname_test.sh
```diff
@@ -0,0 +1,57 @@
...
+test_jail_conf='test-hostname-jail {
+    host.hostname = "test-hostname";
+    path = "/";
+    persist;
+}'
+
+init()
+{
+    echo "$test_jail_conf" > "$root_dir/jail.conf"
```
lwhsu, line 24: For this and the following ones, shell variables should be quoted by braces. i.e., `${test_jail_conf}` here.

bin/hostname/tests/hostname_test.sh
```diff
@@ -0,0 +1,57 @@
...
+    jail -f "$root_dir/jail.conf" -c test-hostname-jail
```
lwhsu, line 25: I suggest extracting `test-hostname-jail` to an variable, as it's used multiple times in this file.

bin/hostname/tests/hostname_test.sh
lwhsu: And since the test_jail_conf variable is only used once. Perhaps we can do something like this:

```
cat > "${root_dir}/jail.conf" <<EOF
test-hostname-jail {
    host.hostname = "test-hostname";
    path = "/";
    persist;
}
EOF
```

bin/hostname/tests/hostname_test.sh
```diff
@@ -0,0 +1,57 @@
...
+    atf_set "descr" "main test"
+}
+basic_body()
+{
+    init
+    jexec test-hostname-jail "hostname"
+    result=$(jexec test-hostname-jail "hostname")
+    if [ "$result" != "test-hostname" ]; then
+        atf_fail "Hostname does not match: $result" 
+    fi
```
lwhsu, line 47: We can use `atf_check_equal "test-hostname" "${result}"` to replace these 3 lines.

bin/hostname/tests/hostname_test.sh
```diff
@@ -0,0 +1,57 @@
...
+{
+    jail -f "$root_dir/jail.conf" -r test-hostname-jail
+    rm "$root_dir/jail.conf"
+}
+
+atf_test_case basic cleanup
+basic_head()
+{
+    atf_set require.user root
+    atf_set "descr" "main test"
```
lwhsu, line 38: We need a better description here.

lwhsu (changes requested):

lwhsu:
The basic test is nice. If possible, I would suggest adding more tests to cover the options for hostname(1), and maybe even test set the hostname.

bin/hostname/tests/hostname_test.sh
```diff
@@ -0,0 +1,57 @@
+#-
+# Copyright (c) 2023 Klara, Inc.
```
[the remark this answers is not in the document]
igalic: I figured @LeeLin2602 was a new Klara hire 😅

bin/hostname/tests/hostname_test.sh
LeeLin2602: Personally I don't like this way to code. I would break indent rule usually.

lwhsu:
Merged as 73974a78f7f60acdbb26cdd5325c21c0bb1d2d4d

PR #1070, opened 2024-01-18
netlink.h: s/typeof/__typeof

llfw:
typeof() does not exist in -std=c99 mode and the relevant #define is only for _KERNEL, so use __typeof here instead.

MFC after:	2 weeks


igalic:
I would've thought cdefs.h aliases that, but it doesn't.

bsdjhb (approved):

bsdjhb:
Merged in 4261507a5e3e73ef6f9535935d22785056954b19, thanks!

PR #1071, opened 2024-01-18
bsdinstall/distfetch.c: check environment variables before downloading and handle memory allocation errors

rilysh:
1. Currently, distfetch checks environment variables existence when it will use them or in a case (in chdir()) it doesn't check at all. As they are necessary to set before doing anything with it, check them, if they set or not, before proceeding any further. This also avoids extra cleaning when that environment variable isn't set.
2. Handle memory allocation error in malloc(PATH_MAX) and replace (sizeof const char *) with (sizeof char *). Both are similar and const doesn't have a size.
3. Indent the error message a bit in chdir().

### DIFF
--- a/usr.sbin/bsdinstall/distfetch/distfetch.c
+++ b/usr.sbin/bsdinstall/distfetch/distfetch.c
@@ -79,14 +79,14 @@ main(void)
 	if (bsddialog_init() == BSDDIALOG_ERROR) {
 		free(diststring);
 		errx(EXIT_FAILURE, "Error libbsddialog: %s\n",
-		     bsddialog_geterror());
+		    bsddialog_geterror());
 	}
 	bsddialog_initconf(&conf);
 	bsddialog_backtitle(&conf, OSNAME " Installer");
 
 	for (i = 0; i < ndists; i++) {
-	        if ((urls[i] = malloc(PATH_MAX)) == NULL) {
-		        free(urls);
+		if ((urls[i] = malloc(PATH_MAX)) == NULL) {
+			free(urls);
 			free(diststring);
 			bsddialog_end();
 			errx(EXIT_FAILURE, "Error: distfetch URLs out of memory!");
--- a/usr.sbin/bsdinstall/distfetch/distfetch.c
+++ b/usr.sbin/bsdinstall/distfetch/distfetch.c
@@ -46,7 +46,7 @@ static int fetch_files(int nfiles, char **urls);
 int
 main(void)
 {
-	char *diststring;
+	char *diststring, *dists, *distdir, *distsite;
 	char **urls;
 	int i;
 	int ndists = 0;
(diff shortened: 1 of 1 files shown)


igalic (commented):
this looks sensible at first blush

usr.sbin/bsdinstall/distfetch/distfetch.c
```diff
@@ -46,47 +46,60 @@ static int fetch_files(int nfiles, char **urls);
...
-		    bsddialog_geterror());
+		     bsddialog_geterror());
 	}
 	bsddialog_initconf(&conf);
 	bsddialog_backtitle(&conf, OSNAME " Installer");
 
 	for (i = 0; i < ndists; i++) {
-		urls[i] = malloc(PATH_MAX);
+	        if ((urls[i] = malloc(PATH_MAX)) == NULL) {
+		        free(urls);
```
jrtc27, line 89: If you want to be complete you need to free urls[j] for 0 <= j < i. But there's really no point in freeing things before you call errx, the process is going to exit...

usr.sbin/bsdinstall/distfetch/distfetch.c
```diff
@@ -46,47 +46,60 @@ static int fetch_files(int nfiles, char **urls);
...
+	if ((urls = calloc(ndists, sizeof(char *))) == NULL) {
 		free(diststring);
 		errx(EXIT_FAILURE, "Error: distfetch URLs out of memory!");
 	}
 
 	if (bsddialog_init() == BSDDIALOG_ERROR) {
 		free(diststring);
 		errx(EXIT_FAILURE, "Error libbsddialog: %s\n",
-		    bsddialog_geterror());
+		     bsddialog_geterror());
```
jrtc27, line 82: The existing style was correct; see https://man.freebsd.org/style(9). Block-level indentation is 1 tab, continuation lines are an additional 4 spaces.

usr.sbin/bsdinstall/distfetch/distfetch.c
jrtc27, line 89:
```suggestion
		if ((urls[i] = malloc(PATH_MAX)) == NULL) {
			free(urls);
```
Tabs not spaces

rilysh: It's possible but it just going to add more mess and as you said, it's going to die, so at that point it will hold very little value to it.

There's no explicit reason...I just felt like to free the parent urls variable, as diststring also going to be free anyway.

usr.sbin/bsdinstall/distfetch/distfetch.c
rilysh: Thanks! I've updated it.

usr.sbin/bsdinstall/distfetch/distfetch.c
rilysh: Sorry about it, somehow Emacs just added spaces instead of tabs, I'm not quite sure why, but I think I've fixed it.
Thanks for the review!

igalic:
looks like the style checker is just erroring out

either way, you can squash those two commits

usr.sbin/bsdinstall/distfetch/distfetch.c
```diff
@@ -78,15 +85,21 @@ main(void)
 	bsddialog_backtitle(&conf, OSNAME " Installer");
 
 	for (i = 0; i < ndists; i++) {
```
bsdimp, line 87: I'd be tempted to have a urls_paths = malloc(ndists * PATH_MAX); and then set the urls[i] to the right offset. That would simplify the error handling.
But like you said... We're exiting soon anyway...

bsdimp:
landed.

bsdimp:
I backed this out after pushing. I'd somehow gotten the impression that @jrtc27 's objections had been addressed, but I was mistaken. Also looking at it critically, it fails our newer guidance on relevance and fixing a known problem, so I'll leave it closed.

PR #1072, opened 2024-01-19
A set of changes to manual pages that were identified to be out of alignment

chrisdavidson:
These are a set of changes to manual pages to update out-dated information.

There are a few categories in this pull request that fall under these categories:
1. Misalignment of man pages sections 2->3
2. Removal of commented ouot .Xr references to clean up
3. Transition of historical (no longer valid) references, that still hold significant historical reference
4. Removal of references that are no longer valid


rilysh:
That's a lot of changes to review at once, I think you should split them in single commits.
e.g. 
```
errono.3: fixed abc and changed xyz
example.1: changed abc
libexample.2: changed abc
```

contrib/sendmail/vacation/vacation.1
```diff
@@ -206,7 +206,7 @@ reads the first line from the standard input for a
 UNIX
 ``From''
 line to determine the sender.
-Sendmail(8)
+.Xr sendmail(8)
```
bsdimp, line 209: .Xr sendmail 8

bsdimp:
Many of the items you changed in contrib are unused.

And some of the ones we do use should be upstreamed first, if possible.   I'll look at the list....

And I'm thinking the total size is on the edge of being too big.

I like what you've done here on the whole.

contrib/sendmail/vacation/vacation.1
chrisdavidson: Okay, I can give that a try. Would it be preferred a pull request PER set of diffs or to have multiple diffs in one pull request?

Appreciate the input as this is first pull request and was trying to balance the set of changes.

chrisdavidson:
thanks for the input!

I saw the issue, with the vacation.1, i would be more than happy to fix real quick and upload..

Would you advice a distinct pull request to address this issue? or to just add to this pull request?

Any guidance would be greatly appreciated.

usr.sbin/bsnmpd/modules/snmp_wlan/snmp_wlan.3
```diff
@@ -64,7 +64,7 @@ Any optional parameters may be set via the
 and
 .Va wlanIfaceLocalAddress
 columns.
-To finally create the interface in the system, a SET with value of active(1) to
+To finally create the interface in the system, a SET with value of active = 1 to
```
yuripv, line 67: I don't think this is correct, it's snmp enum name and its value in parentheses.

usr.sbin/fdwrite/fdwrite.1
```diff
@@ -99,7 +99,7 @@ The
 .Nm
 utility was written while waiting for ``make world'' to complete.
 Some of the code was taken from
-.Xr fdformat 8 .
+.Xrfdformat 8 .
```
yuripv, line 102: This looks bogus.

usr.sbin/ntp/doc/ntp.conf.5
```diff
@@ -824,7 +824,7 @@ with one server and no authentication with another might not be wise.
 .Ss Key Management
 The cryptographic values used by the Autokey protocol are
 incorporated as a set of files generated by the
-.Xr ntp\-keygen 1ntpkeygenmdoc
+.Xr ntp-keygen 1ntpkeygenmdoc
```
yuripv, line 827:
```suggestion
.Xr ntp-keygen 8
```

usr.sbin/ntp/doc/ntp.conf.5
```diff
@@ -1628,7 +1628,7 @@ This restriction is not enabled by default,
 to maintain backward compatibility.
 Expect
 .Cm noepeer
-to become the default in ntp\-4.4.
+to become the default in ntp-4.4.
```
yuripv, line 1631: Make sure `mandoc -Tlint` doesn't whine now.

yuripv (changes requested):
Some nitpicking (just a quick look through the changes).

usr.sbin/bsnmpd/modules/snmp_wlan/snmp_wlan.3
igalic: In that case it needs to be put into syntax that guards it from being turned into a man page link. such as a definition line, or a function call

yuripv: > In that case it needs to be put into syntax that guards it from being turned into a man page link. such as a definition line, or a function call

There is no macro before "active(1)", where/how exactly is it turned to a link?

igalic: automatically.

yuripv: > automatically.

I have no idea what you mean by that, but wherever it is done should be fixed instead, this is a proper mdoc man page, so only explicit link macro should be turned into a link.

igalic: I agree.
But sometimes we do it for… reasons
see https://github.com/freebsd/freebsd-src/blob/47a6fb9d5a2ebec12114a604053ffbd2929f0021/share/man/man9/domain.9#L122-L152

yuripv: > I agree. But sometimes we do it for… reasons see
> 
> https://github.com/freebsd/freebsd-src/blob/47a6fb9d5a2ebec12114a604053ffbd2929f0021/share/man/man9/domain.9#L122-L152

That's different, literal block output in your example is to keep formatting.  And in this case there's nothing looking like a macro so there's no need to escape or otherwise obfuscate the text.

yuripv: Just checked the `mandoc -Thtml` output, it correctly does NOT create a link here:
```
<dl class="Bl-tag">
  <dt id="wlanInterfaceTable"><var class="Va">wlanInterfaceTable</var></dt>
  <dd>The table is used for creation and deletion of virtual wireless
      interfaces. To add a new interface, a SET should be executed on the
      <var class="Va">wlanIfaceName</var> column with value the desired name of
      the interface. Next the parent interface must be set via
      <var class="Va">wlanParentIfName</var> column. Any optional parameters may
      be set via the <var class="Va">wlanIfaceOperatingMode</var>,
      <var class="Va">wlanIfaceFlags</var>, <var class="Va">wlanIfaceBssid</var>
      and <var class="Va">wlanIfaceLocalAddress</var> columns. To finally create
      the interface in the system, a SET with value of active(1) to
      <var class="Va">wlanIfaceStatus</var> column should be executed. To
      destroy a wireless interface a SET with value of destroy(6) to the
      relevant <var class="Va">wlanIfaceStatus</var> column should be
    executed.</dd>
```

lib/geom/part/gpart.8
```diff
@@ -219,7 +219,7 @@ MBR bootcode contains Volume Serial Number by default, and
 .Nm
 tries to preserve it when installing new bootstrap code.
 This option skips preservation to help with some versions of
-.Xr boot0cfg 8
```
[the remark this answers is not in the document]
chrisdavidson: fixed in next commit

lib/geom/part/gpart.8
```diff
@@ -1503,7 +1503,7 @@ and
 .Ed
 .Sh SEE ALSO
 .Xr geom 4 ,
-.Xr boot0cfg 8 ,
```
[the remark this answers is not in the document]
chrisdavidson: fixed in next commit.

usr.sbin/fdwrite/fdwrite.1
chrisdavidson: fixed and commited, ready to update

lib/geom/virstor/gvirstor.8
```diff
@@ -276,7 +276,7 @@ has in its operation: that the size of the virtual storage device will not
 change once it is set, and that the sizes of individual physical storage
 components will always remain constant during their existence.
 For alternative ways to implement virtual or resizable file systems see
-.Xr zfs 1M ,
+.Xr zfs 8M ,
```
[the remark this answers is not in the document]
chrisdavidson: fixed

chrisdavidson:
thank you everyone for the help and guidance on this pull request. With some help, I got the commits broken down.. there may be some commits that are experiencing anomalies and will diligently work through them.

Any questions, please do let me know.

rilysh (commented):
Hmm, a do..while loop, executes the content first (that's inside the loop) and then sees the condition in the while section, whereas only a while loop doesn't.

I'm not sure if someone has commented already on some issue (if yes, you can ignore my comment)...

chrisdavidson:
@rilysh the comment you made, i think this may have gotten sucked into my rebase and not aligned to me, but looks like me do to rebase.

chrisdavidson:
After careful consideration, I came to the conclusion that I should re-evaluate how this pull request was done. I am going to close this pull request out and work on a path forward that is more conducive to all parties involved. Thank you everyone for your guidance and updates in this pull request.

PR #1076, opened 2024-01-20
bsdiff: Use mmap instead of malloc

ricardobranco777:
Use mmap instead of malloc in bsdiff/bspatch.

Debian adds a [patch](https://sources.debian.org/patches/bsdiff/4.3-23/) only to bsdiff to use mmap.  Also [Chromium's fork](https://github.com/cgmb/bsdiff-chromium) uses mmap.

I tested it and it works.  Haven't done any regression or performance tests though.

### DIFF
--- a/usr.bin/bsdiff/bsdiff/bsdiff.c
+++ b/usr.bin/bsdiff/bsdiff/bsdiff.c
@@ -38,6 +38,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <sys/mman.h>
 
 #ifndef O_BINARY
 #define O_BINARY 0
@@ -115,7 +116,7 @@ int main(int argc,char *argv[])
 {
 	int fd;
 	u_char *old,*new;
-	off_t oldsize,newsize;
+	off_t oldsize,newsize,xnewsize;
 	saidx_t *I;
 	off_t scan,pos,len;
 	off_t lastscan,lastpos,lastoffset;
@@ -147,10 +148,9 @@ int main(int argc,char *argv[])
 		err(1, "%s", argv[1]);
 	}
 
-	if (((old=malloc(oldsize+1))==NULL) ||
-		(lseek(fd,0,SEEK_SET)!=0) ||
-		(read(fd,old,oldsize)!=oldsize) ||
-		(close(fd)==-1)) err(1,"%s",argv[1]);
+	old = mmap(NULL, oldsize+1, PROT_READ, MAP_SHARED, fd, 0);
+	if (old == MAP_FAILED || close(fd) == -1)
+		err(1, "%s", argv[1]);
 
 	if(((I=malloc((oldsize+1)*sizeof(saidx_t)))==NULL)) err(1,NULL);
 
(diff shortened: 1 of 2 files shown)


igalic:
unrelated to this PR: good grief, the style on these two files

rilysh (approved):
Looks good!

bsdimp:
Please attend to the style checker reported problems. the style checker found several, and can be run interactively as `tools/build/checkstyle9.pl main..` I'm working to get those reports more inline...

bsdimp:
So the code itself does match the rather non-FreeBSD-ish style that these are written in.

It assumes that these files are mmapable and seekable, but that's no different in practice than the malloc+lseek+read that it replaces (so there's no need to do that as a fallback). I'd like Colin to give it a onceover though. @cperciva what do you think?

bsdimp (approved):
Despite my style comment, I've changed my mind: I think this is ready to go.

cperciva:
LGTM.  Sorry about the horrible code style... I wrote this in 2003 with a very small terminal and adopted a personal coding style of "fit as much as possible onto the screen".

igalic:
> LGTM. Sorry about the horrible code style... I wrote this in 2003 with a very small terminal and adopted a personal coding style of "fit as much as possible onto the screen".

should we… reformat it?

cperciva:
> should we… reformat it?

I've been meaning to do that for about 20 years and never got around to it.  I certainly won't object if someone wants to make it stylish.

usr.bin/bsdiff/bspatch/bspatch.c
```diff
@@ -151,7 +152,7 @@ int main(int argc, char *argv[])
 	if (cap_enter() < 0)
 		err(1, "failed to enter security sandbox");
 
-	cap_rights_init(&rights_ro, CAP_READ, CAP_FSTAT, CAP_SEEK);
+	cap_rights_init(&rights_ro, CAP_READ, CAP_FSTAT, CAP_SEEK, CAP_MMAP_R);
```
bsdjhb, line 155: I think a more faithful use of capsicum here would be to add a new rights_mmap_r that only includes `CAP_SEEK` and `CAP_MMAP_R` leaving `rights_ro` unchanged and using the new permission for `oldfd`.  Perhaps that doesn't matter in this case though?

usr.bin/bsdiff/bsdiff/bsdiff.c
```diff
@@ -147,10 +148,9 @@ int main(int argc,char *argv[])
 		err(1, "%s", argv[1]);
 	}
 
-	if (((old=malloc(oldsize+1))==NULL) ||
-		(lseek(fd,0,SEEK_SET)!=0) ||
-		(read(fd,old,oldsize)!=oldsize) ||
-		(close(fd)==-1)) err(1,"%s",argv[1]);
+	old = mmap(NULL, oldsize+1, PROT_READ, MAP_SHARED, fd, 0);
+	if (old == MAP_FAILED || close(fd) == -1)
+		err(1, "%s", argv[1]);
```
bsdjhb, line 154: Here and elsewhere we no longer close the file descriptor.  I think you could just close them after mmap.

usr.bin/bsdiff/bspatch/bspatch.c
ricardobranco777: > I think a more faithful use of capsicum here would be to add a new rights_mmap_r that only includes `CAP_SEEK` and `CAP_MMAP_R` leaving `rights_ro` unchanged and using the new permission for `oldfd`. Perhaps that doesn't matter in this case though?

We're only mmap'ing the input file, not the patch file.

bsdimp:
landed

bsdimp:
Note: I missed rylish's review, my apologies.

PR #1081, opened 2024-01-23
sysctl.8: fix format typo

ihoro:
Before:
before

After:
after


kprovost:
Merged in https://cgit.freebsd.org/src/commit/?id=380b7eb309478e1aac1609a3dfe1c28436add1c4

PR #1082, opened 2024-01-23
.gitignore: add sys/*/compile

llfw:


igalic (approved):
👍

emaste:
Commit message should explain not just "what" but also "why". What produces sys/*/compile files that need to be ignored?

llfw:
improved the commit message, no functional changes

jrtc27:
> Files in sys/*/compile are created when compiling the kernel with
> config(8).  They are never source files and should never be committed to
> source control, so list this entire directory in .gitignore.

Do we actually support people using config(8) directly to build kernels? I'm inclined to say anything other than a buildkernel in a tree that's had kernel-toolchain (whether directly or via buildworld / kernel-toolchains / etc) run is unsupported. For one you won't have any tools bootstrapped that you might need.

emaste:
I think I haven't used the config-method in a decade or two, and it may well be unsupported, but I wouldn't object to having entries in .gitignore

bsdimp:
I use the config method from time to time. It still works, is still useful and i like this change.

bsdimp:
Tiny tweak to the commit message. 93a25ac9f4c0
Not sure how to credit @emaste or @jrtc27 on this since they didn't explicitly say OK (though the thumbs up might be?)

PR #1083, opened 2024-01-23
glabel: Add support for detecting Linux swap

ricardobranco777:
This will benefit dual-boot installations and also live-CD's.

How I tested it:

1. Compile kernel without GEOM_LABEL
1. `cd /sys/modules/geom/geom_label`
1. `make`
1. `sudo make install`
1. Reboot
1. Attach virtual disk to VM.
1. `echo -e 'n\np\n1\n\n\nt\n82\nw' | sudo chroot /compat/linux fdisk /dev/vtbd1` to create a Linux swap partition (type 82)
1. `sudo chroot /compat/linux mkswap -L MYSWAP /dev/vtbd1s1`
1. `sudo sh -c 'echo geom_label_load=\"YES\" >> /boot/loader.conf'`
1. Reboot
1. `sudo gnop create -v -o 4096 /dev/swaplinux/MYSWAP`
1. `sudo swapon /dev/swaplinux/MYSWAP.nop`
1. `swapinfo | grep MYSWAP`

Cleanup:
1. `sudo swapoff /dev/swaplinux/MYSWAP.nop`
1. `sudo gnop destroy /dev/swaplinux/MYSWAP.nop`

For testing the mount patch that does the gnop dance automatically:
1. `cd /usr/src/sbin/swapon`
1. `make`
1. `sudo make install`
1. `sudo swapon /dev/swaplinux/MYSWAP`
1. `swapinfo |grep MYSWAP`
1. `sudo swapoff /dev/swaplinux/MYSWAP`

The commands for using gnop can be automated with this patch to swapon: https://github.com/freebsd/freebsd-src/pull/1084

### DIFF
--- a/sys/geom/label/g_label_swaplinux.c
+++ b/sys/geom/label/g_label_swaplinux.c
@@ -1,0 +1,91 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <sys/param.h>
+#include <sys/systm.h>
+#include <sys/kernel.h>
+#include <sys/malloc.h>
+
+#include <geom/geom.h>
+#include <geom/geom_dbg.h>
+#include <geom/label/g_label.h>
+
+/*
+ * Taken from
+ * https://github.com/util-linux/util-linux/blob/master/include/swapheader.h
+ */
+
+#define SWAP_VERSION 1
+#define SWAP_UUID_LENGTH 16
+#define SWAP_LABEL_LENGTH 16
+#define SWAP_SIGNATURE "SWAPSPACE2"
+#define SWAP_SIGNATURE_SZ (sizeof(SWAP_SIGNATURE) - 1)
+
+struct swap_header_v1_2 {
+	char	      bootbits[1024];    /* Space for disklabel etc. */
+	uint32_t      version;
+	uint32_t      last_page;
+	uint32_t      nr_badpages;
+	unsigned char uuid[SWAP_UUID_LENGTH];
+	char	      volume_name[SWAP_LABEL_LENGTH];
+	uint32_t      padding[117];
+	uint32_t      badpages[1];
+};
+
+typedef union {
+	struct swap_header_v1_2	header;
+	struct {
+		uint8_t reserved[4096 - SWAP_SIGNATURE_SZ];
+		char	signature[SWAP_SIGNATURE_SZ];
+	} tail;
+} swhdr_t;
+
+#define sw_version	header.version
+#define sw_volume_name	header.volume_name
+#define sw_signature	tail.signature
+
+static void
+g_label_swaplinux_taste(struct g_consumer *cp, char *label, size_t size)
+{
+	struct g_provider *pp;
+	swhdr_t *hdr;
+
+	g_topology_assert_not();
+	pp = cp->provider;
+	label[0] = '\0';
+
+	KASSERT(pp->sectorsize != 0, ("Tasting a disk with 0 sectorsize"));
+	if ((4096 % pp->sectorsize) != 0)
+		return;
+
+	hdr = (swhdr_t *)g_read_data(cp, 0, 4096, NULL);
+	if (hdr == NULL)
+		return;
+
+	/* Check version and magic string */
+	if (hdr->sw_version == SWAP_VERSION &&
+	    !memcmp(hdr->sw_signature, SWAP_SIGNATURE, SWAP_SIGNATURE_SZ))
+		G_LABEL_DEBUG(1, "linux swap detected on %s.", pp->name);
+	else
+		goto exit_free;
+
+	/* Check for volume label */
+	if (hdr->sw_volume_name[0] == '\0')
+		goto exit_free;
+
+	/* Terminate label */
+	hdr->sw_volume_name[sizeof(hdr->sw_volume_name) - 1] = '\0';
+	strlcpy(label, hdr->sw_volume_name, size);
+
+exit_free:
+	g_free(hdr);
+}
+
+struct g_label_desc g_label_swaplinux = {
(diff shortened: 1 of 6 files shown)


sys/geom/label/g_label_swaplinux.c
```diff
@@ -0,0 +1,111 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2024 Ricardo Branco
+ * All rights reserved.
```
igalic, line 5: we've been trying our hardest to get rid of this line

igalic (commented):
they only comment i have do far, I haven't had  time / brain to read / comprehend the rest

sys/geom/label/g_label_swaplinux.c
ricardobranco777: > we've been trying our hardest to get rid of this line

I copied it from the other files in the folder.

igalic: ordinarily, I  would say something like: "yeah, but that probably means we weren't able to track down the original author and ask them if we can remove that line"

but @pjd is still a very active committer, so, 🤷🏻‍♀️

bsdimp: It takes a concerted effort to find all the people with all rights reseved lines, ask them then remove. It gets more complicated when there are multiple copyright holders. And it's not the #1 priority

ricardobranco777: Hopefully fixed now.

ricardobranco777:
> What is the purpose of the labeling? 

To detect labelled Linux swap partitions that we can reuse. 

> Note that you cannot use Linux swap partition for FreeBSD swap/core dump as is because this would potentially damage the Linux' swap header, unless I miss something that magically prevents the corruption.

I tested it on another VM attaching the same qcow2 and I this command ran fine:

```
# file -sbL /dev/vdb1
Linux swap file, 4k page size, little endian, version 1, size 261887 pages, 0 bad pages, LABEL=MYSWAP, UUID=5008f442-a13f-4284-9079-0c7c2fded917
```

But anyway `swapon` calls `mkswap` if needed:
https://github.com/util-linux/util-linux/blob/master/sys-utils/swapon.c#L330

> If I am right, you need to chomp the first page from the volume (by some intermediate geom).

How can I do that?  And how can I add the missing sysctl?

kostikbel:
> > What is the purpose of the labeling?
> 
> To detect labelled Linux swap partitions that we can reuse.
> 
> > Note that you cannot use Linux swap partition for FreeBSD swap/core dump as is because this would potentially damage the Linux' swap header, unless I miss something that magically prevents the corruption.
> 
> I didn't notice any corruption but I can stress test it further by filling the swap and checking the headers.
Perhaps.  Did you see the swapping at all?

> 
> `swapon` on Linux calls `mkswap` if needed anyway: https://github.com/util-linux/util-linux/blob/master/sys-utils/swapon.c#L330

Yes, but damaging the header would make this label not appear on next taste, unless Linux boot repaired the damage.

> 
> > If I am right, you need to chomp the first page from the volume (by some intermediate geom).
> 
> Maybe you're right as the size of the partition shown by `fdisk` is the same as the Capacity shown by `swapinfo`.
> 
> How can I fix that?
This is not completely trivial, you need to create an intermediate provider that does the translation for size and block number for io.  I do not have a good example other than gpart.

ricardobranco777:
> This is not completely trivial, you need to create an intermediate provider that does the translation for size and block number for io. I do not have a good example other than gpart.

This can be done with gnop(8) to skip the first 4096 bytes like this:

`sudo gnop create -v -o 4096 /dev/swaplinux/MYSWAP`
`sudo swapon /dev/swaplinux/MYSWAP.nop`

ricardobranco777:
Will submit a fix for swapon(8) that recognizes Linux swap to call gnop.

kostikbel (commented):
This is a weird thing.  Assume I as user created a gnop device and want to start the swapping on it.  First, your patch would then lay yet another gnop provider on top of the underlying provider on swapon.  On swapoff, the gnop provider is destroyed, and it could be mine instead of newly created.

To make my position clear: I have no objections at all against adding the new label type for the Linux swap partitions (your first patch).  Instead, I do not see a clean way to use that partition as the shared swap between Linux and FreeBSD.  I accept the idea of layering gnop provider on top of Linux swap label, but IMO it should be left to user and just documented somewhere, making the emphasis on the need to preserve the first page to keep Linux label intact.

ricardobranco777:
> This is a weird thing. Assume I as user created a gnop device and want to start the swapping on it. First, your patch would then lay yet another gnop provider on top of the underlying provider on swapon. On swapoff, the gnop provider is destroyed, and it could be mine instead of newly created.

I'm aware of it and I plan to fix this in another PR to swapon.

bsdimp:
Ping? What's the status of these updates?

bsdimp:
Re-reading everything, it's clear there's not consensus for this approach. Should that be worked out, please submit a new pull request.

ricardobranco777:
@kostikbel @bsdimp If I'm reading https://github.com/freebsd/freebsd-src/blob/main/sys/vm/swap_pager.c#L2432 right, FreeBSD already skips the first blocks to "avoid overwriting avoid overwriting any bsd label at the front of the partition".  Much like NetBSD does.

ricardobranco777:
I'm 96.5% sure FreeBSD skips the first page like NetBSD.

I opened a new version with the only difference being `s/4096/PAGE_SIZE` to accomodate for other architectures (on Linux, x86_64 & s390x use 4096, ppc64le uses 65536 & aarch64 uses 16384).

https://github.com/freebsd/freebsd-src/pull/1205

PR #1084, opened 2024-01-26
swapon: Do not overwrite Linux swap header

ricardobranco777:
The Linux swap partition uses a 4096-byte header that holds the label & UUID. You can see the structure [here](https://github.com/util-linux/util-linux/blob/master/include/swapheader.h). 

The problem is thus how to prevent FreeBSD from overwriting the header. It turns out there's a neat way to do just that with GEOM using the [gnop](https://man.freebsd.org/cgi/man.cgi?query=gnop&apropos=0&sektion=0&manpath=FreeBSD+14.0-RELEASE+and+Ports&arch=default&format=html) system tool:

```
sudo gnop create -v -o 4096 /dev/vtbd1s1
sudo swapon /dev/vtbd1s1.nop
...
sudo swapoff /dev/vtbd1s1.nop
sudo gnop destroy /dev/vtbd1s1.nop
```

This PR modifies **swapon** to automate this process, which also works with /etc/fstab.  This will benefit dual-boot installations and also live-CD's.

For automated detection of swap label by GEOM see this PR: https://github.com/freebsd/freebsd-src/pull/1083

Fixes: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=221935

### DIFF
--- a/sbin/swapon/Makefile
+++ b/sbin/swapon/Makefile
@@ -6,6 +6,8 @@ LINKS=	${BINDIR}/swapon ${BINDIR}/swapoff
 LINKS+=	${BINDIR}/swapon ${BINDIR}/swapctl
 MLINKS=	swapon.8 swapoff.8
 MLINKS+=swapon.8 swapctl.8
+SRCS=	swapon.c swaplinux.c
+HDRS+=	extern.h
 
 LIBADD=	util
 
--- a/sbin/swapon/extern.h
+++ b/sbin/swapon/extern.h
@@ -1,0 +1,5 @@
+/*-
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+int is_linux_swap(const char *);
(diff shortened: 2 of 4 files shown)


bsdjhb:
This seems sensible to me in light of the earlier discussion in #1083.

bsdimp:
landed. Thanks! this one looks to be uncontested.

bsdimp:
OK, I saw John's review but didn't read #1083 closely enough... I've backed this out in -current and will leave this closed. Pull requests aren't the place to hash out disagreements like this.

ricardobranco777:
> OK, I saw John's review but didn't read #1083 closely enough... I've backed this out in -current and will leave this closed. Pull requests aren't the place to hash out disagreements like this.

Do you know what was the failure in armv7?

bsdimp:
> Do you know what was the failure in armv7?

Alignment requirement increased by cast.

The feedback was that it would be better if this were opt in rather than automatic... I'm happy to revisit.

PR #1085, opened 2024-01-27
add noexcept to stdlib.h for C++11

llfw:
C++11 requires that certain functions in `<stdlib.h>`/`<cstdlib>` be marked `noexcept`.  To support this, add a new macro `__NOEXCEPT` to `<sys/cdefs.h>` and apply it those functions.

I also added `__NOEXCEPT_IF` to `<sys/cdefs.h>` for completeness, but it's currently not used.

Tested with buildworld.

### DIFF
--- a/include/stdlib.h
+++ b/include/stdlib.h
@@ -84,9 +84,9 @@ extern int __mb_cur_max;
 extern int ___mb_cur_max(void);
 #define	MB_CUR_MAX	((size_t)___mb_cur_max())
 
-_Noreturn void	 abort(void);
+_Noreturn void	 abort(void) __noexcept;
 int	 abs(int) __pure2;
-int	 atexit(void (* _Nonnull)(void));
+int	 atexit(void (* _Nonnull)(void)) __noexcept;
 double	 atof(const char *);
 int	 atoi(const char *);
 long	 atol(const char *);
@@ -154,7 +154,7 @@ unsigned long long
 	 strtoull(const char * __restrict, char ** __restrict, int);
 #endif /* __LONG_LONG_SUPPORTED */
 
-_Noreturn void	 _Exit(int);
+_Noreturn void	 _Exit(int) __noexcept;
 #endif /* __ISO_C_VISIBLE >= 1999 */
 
 /*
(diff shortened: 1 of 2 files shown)


brooksdavis:
Is there a reason why you used upper case? We've typically used lower case (e.g. for `__restrict` below).

bsdimp:
We did __restrict because that's natively supported by compilers always (the restrict key word is only live for new-enough standard configurations).  It seems like doing the same here would be good, regardless of the compiler's support unless
we have a pathological case where we'd want to use this with < C++11. I think that we should follow the pattern here as well.

though we might deviate here because there's a `noexcept` and `noexcept(expr)` forms in the underlying language. Still lower case seems warranted.

llfw:
changed both macros to lowercase

llfw:
i've just noticed an issue here which is that `exit()` should *not* be `noexcept`.  i'll submit a fix shortly, in the mean time don't commit.

llfw:
this in this commit, this is based on <https://www.eel.is/c++draft/cstdlib.syn#header:%3Ccstdlib%3E>.

bsdimp:
This looks good. I'm landing it assuming that the buildworld works.
This matches the list from the C++ standard and various standard tracking sites.
The style looks good, though there's an 'X' in the checks. That's a false positive in the checker because of the _Nonnull keyword.
The commits are good way to break up these changes.

llfw:
this is highly unlikely to cause any runtime problems but, fwiw, i did run this branch on my desktop for a few days with no issues (including a ports build).

bsdimp:
Yea, it passed the buildworld (I expected it would). when pushing changes to system header files, I like to verify a few more things than with other changes. I didn't think this would need a full exp run, though, because __noexcept is in implementation space and generally not user visible.

bsdimp:
c27a89971805

Note to self: check for blank lines in the commit messages after my script runs.

llfw:
partially reverted for `std::quick_exit` in d56a6f0516a748774673272f227c1862827b46fc, ref: https://lists.freebsd.org/archives/dev-commits-src-main/2024-February/021484.html

(mentioning for posterity in case someone comes across this closed PR later.)

PR #1087, opened 2024-01-28
Import monitor tool from DragonflyBSD

ricardobranco777:
- Import [monitor](https://leaf.dragonflybsd.org/cgi/web-man?command=monitor&section=ANY) tool from DragonflyBSD.
- Add support for NOTE_OPEN, NOTE_CLOSE, NOTE_CLOSE_WRITE & NOTE_READ supported by FreeBSD & NetBSD 10.
- Add support for NOTE_TRUNCATE supported by OpenBSD.

### DIFF
--- a/usr.bin/monitor/monitor.1
+++ b/usr.bin/monitor/monitor.1
@@ -1,0 +1,68 @@
+.\" Copyright (c) 2008 The DragonFly Project.  All rights reserved.
+.\"
+.\" This code is derived from software contributed to The DragonFly Project
+.\" by Matthew Dillon <dillon@backplane.com>
+.\"
+.\" Redistribution and use in source and binary forms, with or without
+.\" modification, are permitted provided that the following conditions
+.\" are met:
+.\"
+.\" 1. Redistributions of source code must retain the above copyright
+.\"    notice, this list of conditions and the following disclaimer.
+.\" 2. Redistributions in binary form must reproduce the above copyright
+.\"    notice, this list of conditions and the following disclaimer in
+.\"    the documentation and/or other materials provided with the
+.\"    distribution.
+.\" 3. Neither the name of The DragonFly Project nor the names of its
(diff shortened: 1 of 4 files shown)


jlduran:
I wasn't aware of this utility. Very useful, thank you!
`usage()` is missing the `-q` flag. Perhaps we should submit this minor fix upstream.
I don't know what's the policy for putting this code directly under usr.bin vs. under contrib.

jlduran:
Shouldn't `-q` and `-v` be mutually exclusive?

usr.bin/monitor/monitor.1
```diff
@@ -0,0 +1,69 @@
...
+.Nm
+utility monitors one or more files or directories and outputs the
+appropriate file name to stdout whenever a change occurs.
+Supported options are as follows:
+.Bl -tag -width indent
+.It Fl q
+Be more quiet.
+Specifying this option will cause only the file paths to be output.
+.It Fl v
+Be more verbose.
```
jlduran, line 54: Specifying this option will print an additional size column.
(Or something similar).

igalic:
does it make sense to use the vendor workflow for this? https://docs.freebsd.org/en/books/developers-handbook/book/#policies-contributed

bsdimp:
> does it make sense to use the vendor workflow for this? https://docs.freebsd.org/en/books/developers-handbook/book/#policies-contributed

It might. But that's more up to the person that lands this. Our vendor branches aren't a grwat fit with github pull requests and a bit of a steep hill for a novice to climb.

It would be useful to get the hash from the DF repo, though. That would help me quickly do the vendor import first. Ideally we'd be unchanged, but some upstream sources aren't responsive enough sometimes to do that universally. DF has been a mixed bag in the past so I wouldn't block on that.

ricardobranco777:
Added support for events supported by FreeBSD, NetBSD & OpenBSD.

PR #1088, opened 2024-01-29
chflags: Fix -f option

ricardobranco777:
There's a bug with the -f option when used with the -v option.

To reproduce:

```
$ chflags -vf uchg /etc/passwd
/etc/passwd
$ echo $?
0
```

Fixes https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=276723

### DIFF
--- a/bin/chflags/chflags.c
+++ b/bin/chflags/chflags.c
@@ -182,9 +182,11 @@ main(int argc, char *argv[])
 		if (newflags == p->fts_statp->st_flags)
 			continue;
 		if (chflagsat(AT_FDCWD, p->fts_accpath, newflags,
-		    atflag) == -1 && !fflag) {
-			warn("%s", p->fts_path);
-			rval = 1;
+		    atflag) == -1) {
+			if (!fflag) {
+				warn("%s", p->fts_path);
+				rval = 1;
+			}
 		} else if (vflag || siginfo) {
 			(void)printf("%s", p->fts_path);
 			if (vflag > 1 || siginfo)


bin/chflags/chflags.c
```diff
@@ -182,9 +182,11 @@ main(int argc, char *argv[])
 		if (newflags == p->fts_statp->st_flags)
 			continue;
 		if (chflagsat(AT_FDCWD, p->fts_accpath, newflags,
-		    atflag) == -1 && !fflag) {
-			warn("%s", p->fts_path);
-			rval = 1;
+		    atflag) == -1) {
+			if (!fflag) {
+				warn("%s", p->fts_path);
+				rval = 1;
```
bsdimp, line 188: Help me understand... If -f is set, the fflag is set, and we'll not set rval=1 and we'll still have the bad return code this bug is supposed to fix. What am I missing?

markjdb: I believe the notion is that if both -v and -f are specified, and the chflags operation fails, the file name shouldn't be printed. Though, the documentation doesn't really say one way or another what should happen here. This change isn't obviously right IMO.

ricardobranco777: > Help me understand... If -f is set, the fflag is set, and we'll not set rval=1 and we'll still have the bad return code this bug is supposed to fix. What am I missing?

Fixed.

kostikbel (commented):
Could you please mail the patch to me (kib@freebsd.org) in the git format-patch format, with the metadata filled.  Most important is the right author field.  Thank you.

kostikbel:
Oh wait I did not see the discussion after the submission.  I do think that the original patch was right and the new patch is not.  From the documentation in chflags(1) for -f:
`
     -f      Do not display a diagnostic message if chflags could not modify
             the flags for file, nor modify the exit status to reflect such
             failures.
`
So setting rval to 1 only when fflag was not set was IMO correct.

ricardobranco777:
> Oh wait I did not see the discussion after the submission. I do think that the original patch was right and the new patch is not. From the documentation in chflags(1) for -f: `-f Do not display a diagnostic message if chflags could not modify the flags for file, nor modify the exit status to reflect such failures.` So setting rval to 1 only when fflag was not set was IMO correct.

In this sense it follows `rm -f` logic but not `chmod -f`.  And the patch that landed `-f` & `-v` was inspired by chmod.

ricardobranco777:
> So chmod is also not following its own documentation, at least according to my reading. -f must not cause non-zero exit status. Either we fix the documentation, or code.

Reverted to the original patch.

kostikbel:
Ok, do we want the same change for chmod?

ricardobranco777:
> Ok, do we want the same change for chmod?

I'll send a patch later.

bsdimp:
I agree with kib. I'd not checked the docs.
-f is a FreeBSDism, it's not in the latest POSIX standard, so it should follow the docs.

bsdimp:
So I think it's almost ready to land. The commit message, however, should be expanded a bit. FreeBSD likes to have the 'why' of the commit explained after the 'subject' line. Something like: 'chflags is documented to not report problems and  exit 0 even if it encounters errors and the -f is set. Make the code match the docs'  or similar (the wording is a bit awkward, I'll admit, but should give the flavor of what I'm asking for). Thanks for being responsive to feedback, even if it was initially a little off base.

ricardobranco777:
> So I think it's almost ready to land. The commit message, however, should be expanded a bit. FreeBSD likes to have the 'why' of the commit explained after the 'subject' line. Something like: 'chflags is documented to not report problems and exit 0 even if it encounters errors and the -f is set. Make the code match the docs' or similar (the wording is a bit awkward, I'll admit, but should give the flavor of what I'm asking for). Thanks for being responsive to feedback, even if it was initially a little off base.... but I see kib has taken care of this, so for next time with chmod maybe?

The code already matches the doc I think. Will fix the commit message.

PR #1090, opened 2024-01-30
libkern/strcpy.c: remove unnecessary parenthesis

rilysh:
This is a trivial change, found this while I was looking for strcpy() implementation. Seems added a long time ago.
From
`(strcpy)(...)` 
to
`strcpy(...)`

### DIFF
--- a/sys/libkern/strcpy.c
+++ b/sys/libkern/strcpy.c
@@ -33,7 +33,7 @@
 #include <sys/libkern.h>
 
 char *
-(strcpy)(char * __restrict to, const char * __restrict from)
+strcpy(char * __restrict to, const char * __restrict from)
 {
 	char *save = to;
 


igalic (approved):

rilysh:
Oh wait, this is interesting, builds are failing because of a conflict of a function and that is strcpy() (I just changed). It seems like parenthesis were used so as not to conflict with other implementations (e.g. libc one). Though I didn't really read all the errors....

rilysh:
I don't think this function is particularly used... as there's also `#define strcpy(d, s)    __builtin_strcpy((d), (s))` which will use "compiler version" of strcpy() anyway.

I'll close this PR for now.

markjdb:
The parentheses are there intentionally. Note that other, similar functions (e.g., strlen()) are defined the same way. It's worth checking `git blame` output when making such changes.

The parentheses prevent macro expansion in the situation where the function name is also a preprocessor symbol. As you note, the problem is related to the fact that there is a `#define strcpy` in an included header. Without the parentheses, the preprocessor would expand `strcpy(char *to, ...)` to `__builtin_strcpy(char *to, ...)`, which won't compile. Wrapping the function name with parentheses is a hack to avoid this expansion.

A more interesting use of this hack is with `malloc()`, in sys/sys/malloc.h and kern_malloc.c.

rilysh:
Interesting, I didn't know that this can be used for functions as well, thank you for the explanation!

PR #1091, opened 2024-01-30
daily/223.backup-zfs: improve daily_backup_zfs_verbose behaviour

llfw:
- 223.backup-zfs would previously honour the daily_backup_zfs_verbose flag for zfs/zpool list, but not for the properties list.  fix it to show a diff for both of these if requested.

- if daily_backup_zfs_verbose was disabled, 223.backup-zfs would still set rc=1 if the backup files changed, which caused periodic(8) to send a useless email even if daily_show_success=NO was set.

  change this so that it only sets rc=1 if diff output is enabled, i.e. the output is actually useful to the admin.

---

every day, i get several emails like this:

```
From: Charlie Root <root@ilythia.eden.le-Fay.ORG>
Subject: ilythia.eden.le-fay.org daily run output
To: root@ilythia.eden.le-Fay.ORG
Date: Tue, 30 Jan 2024 03:13:37 +0000 (GMT)


Backup of ZFS information for all imported pools

-- End of daily output --
```

this PR fixes that, without removing functionality for people who want this information.

### DIFF
--- a/usr.sbin/periodic/etc/daily/223.backup-zfs
+++ b/usr.sbin/periodic/etc/daily/223.backup-zfs
@@ -13,6 +13,7 @@ then
 fi
 
 bak_dir=/var/backups
+rc=0
 
 rotate() {
 	base_name=$1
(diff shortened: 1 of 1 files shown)


bsdimp:
4b1d3a30914e

PR #1093, opened 2024-02-01
kldxref: fix maketempfile function's way of finding root dir

igalic:
Rather than assuming that the "root" is passed as directory and will be
marked by a trailing slash, we just assume that the directory, which has
been checked previously to be a directory, is a directory.

This fixes an inconsistency between `kldxref /boot/modules`, which tries
to create the temp file in `/boot/`, and `kldxref /boot/modules/`, which
tries to create it in `/boot/modules/` itself.

### DIFF
--- a/usr.sbin/kldxref/kldxref.c
+++ b/usr.sbin/kldxref/kldxref.c
@@ -717,12 +717,9 @@ read_kld(char *filename, char *kldname)
 static FILE *
 maketempfile(char *dest, const char *root)
 {
-	char *p;
-	int n, fd;
+	int fd;
 
-	p = strrchr(root, '/');
-	n = p != NULL ? p - root + 1 : 0;
-	if (snprintf(dest, MAXPATHLEN, "%.*slhint.XXXXXX", n, root) >=
+	if (snprintf(dest, MAXPATHLEN, "%s/lhint.XXXXXX", root) >=
 	    MAXPATHLEN) {
 		errno = ENAMETOOLONG;
 		return (NULL);


igalic:
cc @bsdimp @bsdjhb

usr.sbin/kldxref/kldxref.c
```diff
@@ -717,13 +717,13 @@ read_kld(char *filename, char *kldname)
...
+	int fd;
 
-	p = strrchr(root, '/');
-	n = p != NULL ? p - root + 1 : 0;
-	if (snprintf(dest, MAXPATHLEN, "%.*slhint.XXXXXX", n, root) >=
-	    MAXPATHLEN) {
+	if (stat(root, &stroot) != 0 || !S_ISDIR(stroot.st_mode))
+		return (NULL);
+	if (snprintf(dest, MAXPATHLEN, "%.*s/lhint.XXXXXX", (int)strlen(root),
+	    root) >= MAXPATHLEN) {
```
bsdjhb, line 726:
```suggestion
	if (snprintf(dest, MAXPATHLEN, "%s/lhint.XXXXXX", root) >= MAXPATHLEN) {
```

Subject to line wrapping.

bsdjhb (commented):
I'm not even sure we need the stat() call though.  This function is only called for a result from fts_read(3) where fts_info == FTS_D.  That is, we only pass directories to maketempfile() already.

igalic:
> I'm not even sure we need the stat() call though. This function is only called for a result from fts_read(3) where fts_info == FTS_D. That is, we only pass directories to maketempfile() already.

I'll remove that stat and resubmit

igalic:
@bsdjhb I think this is ready to go

bsdimp (approved):

bsdimp:
This passes my mini series of tests. There's no kyua tests, so I'll land it.
f4613af424cc

PR #1094, opened 2024-02-01
share/examples/IPv6/USAGE: remove

llfw:
This document dates from the KAME days and, among other things, references the 'prefix' command which has not existed for a long time. Since IPv6 configuration is now documented in the Handbook, remove this obsolete file.

---

cc: @bsdimp @jlduran since this came up in #1023


llfw:
to address a potential objection that this should be rewritten instead of removed: if someone wants to rewrite it to be current, they are free to do that, but in the mean time it gives wrong information which will confuse users, so it is better to remove it than leave it there.

bsdimp:
4339f1e667ff

PR #1097, opened 2024-02-03
Add unittests for some simple system calls

LeeLin2602:
kill(2), wait4(2): add some testcase including different signals and waitpid wnohang.


etc/mtree/BSD.tests.dist
```diff
@@ -933,6 +933,11 @@
         ..
         sys
         ..
+        syscalls
+            fork
+            ..
+            kill
```
emaste, line 939: missing a `..` here?

tests/sys/syscalls/fork/fork_test.c
```diff
@@ -0,0 +1,74 @@
...
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+#include <atf-c.h>
+#include <sys/wait.h>
+#include <sys/syscall.h>
+
+static int
+simple_test(int expected)
```
emaste, line 33: arg probably better named `exit_status` - we're testing fork() with the child providing different return values (and, if we're going to test two values we should probably check 0 and some non-zero value)

LeeLin2602: done

etc/mtree/BSD.tests.dist
LeeLin2602: done

bsdimp:
These tests are a bit too basic. The don't test things that aren't already tested extensively elsewhere, and we don't test that various aspects of fork work (like inherited file descriptors, etc).

LeeLin2602:
Hi, thank you for your comments.
But I would like to declare that this is the very first one step for our plan.
As you mentioned, this has been well tested in someplace else, however, this is the problem.
For a beginner, we cannot find the exact test file path quickly. And when we want to add more tests, we also have no idea where to append.
Therefore, this is our goal:
We are trying to test all the syscalls and would like to take it step by step.
So we finished the test for the most simple syscall. Hence, this is our very first step.
I hope you could understand it.
Thank you for your replying again.

PR #1098, opened 2024-02-03
Add support for 64-bit machines with 32-bit UEFI implementations

VexedUXR:
Some machines (mostly baytrail notebooks) have 64-bit capable cpus but are stuck on
32-bit uefi firmware, with so CSM.

Add support for them by making the amd64 loader compile itself again with DO32=1,
(see the amd64 Makefile.inc) all of the extra ia32 specific code is in loader/arch/amd64/ia32.
The loader can be disabled using MK_LOADER_IA32.

Tested on qemu and the Acer one 10 s1002
Also cross-compiled to armv6 to make sure nothing broke.

### DIFF
--- a/stand/defs.mk
+++ b/stand/defs.mk
@@ -40,7 +40,14 @@ WARNS?=		1
 BOOTSRC=	${SRCTOP}/stand
 EFISRC=		${BOOTSRC}/efi
 EFIINC=		${EFISRC}/include
+# For amd64, there's a bit of mixed bag. Some of the tree (i386, lib*32) is
+# built 32-bit and some 64-bit (lib*, efi). Centralize all the 32-bit magic here
+# and activate it when DO32 is explicitly defined to be 1.
+.if ${MACHINE_ARCH} == "amd64" && ${DO32:U0} == 1
+EFIINCMD=	${EFIINC}/i386
+.else
 EFIINCMD=	${EFIINC}/${MACHINE}
+.endif
 FDTSRC=		${BOOTSRC}/fdt
 FICLSRC=	${BOOTSRC}/ficl
 LDRSRC=		${BOOTSRC}/common
@@ -130,9 +137,6 @@ CFLAGS+=	-m32 -mcpu=powerpc -mbig-endian
 CFLAGS+=	-m32 -mcpu=powerpc -mlittle-endian
 .endif
 
-# For amd64, there's a bit of mixed bag. Some of the tree (i386, lib*32) is
-# build 32-bit and some 64-bit (lib*, efi). Centralize all the 32-bit magic here
-# and activate it when DO32 is explicitly defined to be 1.
 .if ${MACHINE_ARCH} == "amd64" && ${DO32:U0} == 1
 CFLAGS+=	-m32
 # LD_FLAGS is passed directly to ${LD}, not via ${CC}:
--- a/stand/efi/libefi32/Makefile
+++ b/stand/efi/libefi32/Makefile
@@ -1,0 +1,4 @@
+DO32=1
+
+.PATH:	${.CURDIR}/../libefi
+.include "${.CURDIR}/../libefi/Makefile"
--- a/stand/efi/loader/Makefile
+++ b/stand/efi/loader/Makefile
@@ -6,7 +6,13 @@ LOADER_EXT2FS_SUPPORT?=	no
 
 .include <bsd.init.mk>
 
+.if ${MACHINE} == "amd64" && ${DO32:U0} == 1
+__arch=	i386
+LOADER?=	loader_ia32
+.else
+__arch=	${MACHINE}
 LOADER?=	loader_${LOADER_INTERP}
+.endif
 PROG=		${LOADER}.sym
 INTERNALPROG=
 WARNS?=		3
@@ -52,13 +58,13 @@ CFLAGS.gfx_fb.c += -DHAVE_MEMCPY -I${SRCTOP}/sys/contrib/zlib
 CWARNFLAGS.main.c+=	-Wno-format
 
 .PATH: ${.CURDIR}/../loader
-.PATH: ${.CURDIR}/../loader/arch/${MACHINE}
-.include "${.CURDIR}/../loader/arch/${MACHINE}/Makefile.inc"
+.PATH: ${.CURDIR}/../loader/arch/${__arch}
+.include "${.CURDIR}/../loader/arch/${__arch}/Makefile.inc"
 
 CFLAGS+=	-I${.CURDIR}
-CFLAGS+=	-I${.CURDIR}/arch/${MACHINE}
+CFLAGS+=	-I${.CURDIR}/arch/${__arch}
 CFLAGS+=	-I${EFISRC}/include
-CFLAGS+=	-I${EFISRC}/include/${MACHINE}
+CFLAGS+=	-I${EFISRC}/include/${__arch}
 CFLAGS+=	-I${SYSDIR}/contrib/dev/acpica/include
 CFLAGS+=	-I${BOOTSRC}/i386/libi386
 CFLAGS+=	-DEFI
@@ -95,11 +101,11 @@ CLEANFILES+=	8x16.c
 FILES+=	${LOADER}.efi
 FILESMODE_${LOADER}.efi=	${BINMODE}
 
-.if ${LOADER_INTERP} == ${LOADER_DEFAULT_INTERP}
+.if ${LOADER_INTERP} == ${LOADER_DEFAULT_INTERP} && ${__arch} != "i386"
 LINKS+=		${BINDIR}/${LOADER}.efi ${BINDIR}/loader.efi
 .endif
 
-LDSCRIPT=	${.CURDIR}/../loader/arch/${MACHINE}/${MACHINE}.ldscript
+LDSCRIPT=	${.CURDIR}/../loader/arch/${__arch}/${__arch}.ldscript
 LDFLAGS+=	-Wl,-T${LDSCRIPT},-Bsymbolic,-znotext -pie
 .if ${LINKER_TYPE} == "bfd" && ${LINKER_VERSION} >= 23400
 LDFLAGS+=	-Wl,--no-dynamic-linker
@@ -125,8 +131,14 @@ ${LOADER}.efi: ${PROG}
 		--output-target=${EFI_TARGET} ${.ALLSRC} ${.TARGET}
 
 LIBEFI=		${BOOTOBJ}/efi/libefi/libefi.a
+LIBEFI32=	${BOOTOBJ}/efi/libefi32/libefi.a
 
+.if ${__arch} == "i386"
+DPADD=		${LDR_INTERP32} ${LIBEFI32} ${LIBSA32} ${LDSCRIPT}
+LDADD=		${LDR_INTERP32} ${LIBEFI32} ${LIBSA32}
+.else
 DPADD=		${LDR_INTERP} ${LIBEFI} ${LIBSAFDT} ${LIBEFI_FDT} ${LIBSA} ${LDSCRIPT}
 LDADD=		${LDR_INTERP} ${LIBEFI} ${LIBSAFDT} ${LIBEFI_FDT} ${LIBSA}
+.endif
 
 .include <bsd.prog.mk>
--- a/stand/efi/loader/bootinfo.c
+++ b/stand/efi/loader/bootinfo.c
@@ -447,8 +447,14 @@ bi_load(char *args, vm_offset_t *modulep, vm_offset_t *kernendp, bool exit_bs)
 	module = *modulep;
 	file_addmetadata(kfp, MODINFOMD_MODULEP, sizeof(module), &module);
 #endif
-#if defined(EFI) && !defined(__i386__)
+#ifdef EFI
+#ifndef __i386__
 	file_addmetadata(kfp, MODINFOMD_FW_HANDLE, sizeof(ST), &ST);
+#endif
+#if defined(__amd64__) || defined(__i386__)
+	file_addmetadata(kfp, MODINFOMD_EFI_ARCH, sizeof(MACHINE_ARCH),
+	    MACHINE_ARCH);
+#endif
 #endif
 #ifdef LOADER_GELI_SUPPORT
 	geli_export_key_metadata(kfp);
--- a/stand/efi/loader/bootinfo.c
+++ b/stand/efi/loader/bootinfo.c
@@ -185,7 +185,7 @@ bi_load_efi_data(struct preloaded_file *kfp, bool exit_bs)
 	struct efi_map_header *efihdr;
 	bool do_vmap;
 
-#if defined(__amd64__) || defined(__aarch64__)
+#if defined(__amd64__) || defined(__aarch64__) || defined(__i386__)
 	struct efi_fb efifb;
 
 	efifb.fb_addr = gfx_state.tg_fb.fb_addr;
@@ -339,7 +339,16 @@ bi_load(char *args, vm_offset_t *modulep, vm_offset_t *kernendp, bool exit_bs)
 	vm_offset_t size;
 	char *rootdevname;
 	int howto;
+#ifdef __i386__
+	/*
+	 * The 32-bit UEFI loader is used to
+	 * boot the 64-bit kernel on machines
+	 * that support it.
+	 */
+	bool is64 = true;
+#else
 	bool is64 = sizeof(long) == 8;
+#endif
 #if defined(LOADER_FDT_SUPPORT)
 	vm_offset_t dtbp;
 	int dtb_size;
@@ -438,7 +447,7 @@ bi_load(char *args, vm_offset_t *modulep, vm_offset_t *kernendp, bool exit_bs)
 	module = *modulep;
 	file_addmetadata(kfp, MODINFOMD_MODULEP, sizeof(module), &module);
 #endif
-#ifdef EFI
+#if defined(EFI) && !defined(__i386__)
 	file_addmetadata(kfp, MODINFOMD_FW_HANDLE, sizeof(ST), &ST);
 #endif
 #ifdef LOADER_GELI_SUPPORT
--- a/stand/efi/loader/copy.c
+++ b/stand/efi/loader/copy.c
@@ -182,13 +182,13 @@ out:
 #endif
 
 #if defined(__aarch64__) || defined(__amd64__) || defined(__arm__) || \
-    defined(__riscv)
+    defined(__riscv) || defined(__i386__)
 #define	EFI_STAGING_2M_ALIGN	1
 #else
 #define	EFI_STAGING_2M_ALIGN	0
 #endif
 
-#if defined(__amd64__)
+#if defined(__amd64__) || defined(__i386__)
 #define	EFI_STAGING_SLOP	M(8)
 #else
 #define	EFI_STAGING_SLOP	0
@@ -209,7 +209,7 @@ efi_copy_free(void)
 	stage_offset = 0;
 }
 
-#ifdef __amd64__
+#if defined(__amd64__) || defined(__i386__)
 int copy_staging = COPY_STAGING_AUTO;
 
 static int
@@ -281,7 +281,7 @@ command_staging_slop(int argc, char *argv[])
 COMMAND_SET(staging_slop, "staging_slop", "set staging slop",
     command_staging_slop);
 
-#if defined(__amd64__)
+#if defined(__amd64__) || defined(__i386__)
 /*
  * The staging area must reside in the first 1GB or 4GB physical
  * memory: see elf64_exec() in
@@ -320,7 +320,8 @@ efi_copy_init(void)
 	 */
 	if (running_on_hyperv())
 		efi_verify_staging_size(&nr_pages);
-
+#endif
+#if defined(__amd64__) || defined(__i386__)
 	staging = get_staging_max();
 #endif
 	status = BS->AllocatePages(EFI_ALLOC_METHOD, EfiLoaderCode,
@@ -380,9 +381,10 @@ efi_check_space(vm_offset_t end)
 	end += staging_slop;
 
 	nr_pages = EFI_SIZE_TO_PAGES(end - staging_end);
-#if defined(__amd64__)
+#if defined(__amd64__) || defined(__i386__)
 	/*
-	 * amd64 needs all memory to be allocated under the 1G or 4G boundary.
+	 * The amd64 kernel needs all memory to be allocated under the 1G or
+	 * 4G boundary.
 	 */
 	if (end > get_staging_max())
 		goto before_staging;
@@ -427,7 +429,7 @@ expand:
 #if EFI_STAGING_2M_ALIGN
 	nr_pages += M(2) / EFI_PAGE_SIZE;
 #endif
-#if defined(__amd64__)
+#if defined(__amd64__) || defined(__i386__)
 	new_base = get_staging_max();
 #endif
 	status = BS->AllocatePages(EFI_ALLOC_METHOD, EfiLoaderCode,
--- a/stand/efi/loader/framebuffer.c
+++ b/stand/efi/loader/framebuffer.c
@@ -149,7 +149,16 @@ efifb_from_gop(struct efi_fb *efifb, EFI_GRAPHICS_OUTPUT_PROTOCOL_MODE *mode,
 {
 	int result;
 
-	efifb->fb_addr = mode->FrameBufferBase;
+	/*
+	 * The Asus EEEPC 1025C, and possibly others,
+	 * require the address to be masked.
+	 */
+	efifb->fb_addr =
+#ifdef __i386__
+	    mode->FrameBufferBase & 0xffffffff;
+#else
+	    mode->FrameBufferBase;
+#endif
 	efifb->fb_size = mode->FrameBufferSize;
 	efifb->fb_height = info->VerticalResolution;
 	efifb->fb_width = info->HorizontalResolution;
@@ -555,6 +564,7 @@ efi_has_gop(void)
 int
 efi_find_framebuffer(teken_gfx_t *gfx_state)
 {
+	EFI_PHYSICAL_ADDRESS ptr;
 	EFI_HANDLE *hlist;
 	UINTN nhandles, i, hsize;
 	struct efi_fb efifb;
@@ -651,16 +661,15 @@ efi_find_framebuffer(teken_gfx_t *gfx_state)
 	    efifb.fb_mask_blue | efifb.fb_mask_reserved);
 
 	if (gfx_state->tg_shadow_fb != NULL)
-		BS->FreePages((EFI_PHYSICAL_ADDRESS)gfx_state->tg_shadow_fb,
+		BS->FreePages((uintptr_t)gfx_state->tg_shadow_fb,
 		    gfx_state->tg_shadow_sz);
 	gfx_state->tg_shadow_sz =
 	    EFI_SIZE_TO_PAGES(efifb.fb_height * efifb.fb_width *
 	    sizeof(EFI_GRAPHICS_OUTPUT_BLT_PIXEL));
 	status = BS->AllocatePages(AllocateAnyPages, EfiLoaderData,
-	    gfx_state->tg_shadow_sz,
-	    (EFI_PHYSICAL_ADDRESS *)&gfx_state->tg_shadow_fb);
-	if (status != EFI_SUCCESS)
-		gfx_state->tg_shadow_fb = NULL;
+	    gfx_state->tg_shadow_sz, &ptr);
+	gfx_state->tg_shadow_fb = status == EFI_SUCCESS ?
+	    (uint32_t *)(uintptr_t)ptr : NULL;
 
 	return (0);
 }
--- a/stand/efi/loader/framebuffer.c
+++ b/stand/efi/loader/framebuffer.c
@@ -656,7 +656,7 @@ efi_find_framebuffer(teken_gfx_t *gfx_state)
 	gfx_state->tg_shadow_sz =
 	    EFI_SIZE_TO_PAGES(efifb.fb_height * efifb.fb_width *
 	    sizeof(EFI_GRAPHICS_OUTPUT_BLT_PIXEL));
-	status = BS->AllocatePages(AllocateMaxAddress, EfiLoaderData,
+	status = BS->AllocatePages(AllocateAnyPages, EfiLoaderData,
 	    gfx_state->tg_shadow_sz,
 	    (EFI_PHYSICAL_ADDRESS *)&gfx_state->tg_shadow_fb);
 	if (status != EFI_SUCCESS)
--- a/stand/efi/loader/main.c
+++ b/stand/efi/loader/main.c
@@ -964,7 +964,7 @@ main(int argc, CHAR16 *argv[])
 	archsw.arch_getdev = efi_getdev;
 	archsw.arch_copyin = efi_copyin;
 	archsw.arch_copyout = efi_copyout;
-#ifdef __amd64__
+#if defined(__amd64__) || defined(__i386__)
 	archsw.arch_hypervisor = x86_hypervisor;
 #endif
 	archsw.arch_readin = efi_readin;
--- a/stand/efi/loader/main.c
+++ b/stand/efi/loader/main.c
@@ -45,6 +45,7 @@
 #include <disk.h>
 #include <dev_net.h>
 #include <net.h>
+#include <inttypes.h>
 
 #include <efi.h>
 #include <efilib.h>
@@ -920,7 +921,7 @@ acpi_detect(void)
 		if ((rsdp = efi_get_table(&acpi)) == NULL)
 			return;
 
-	sprintf(buf, "0x%016llx", (unsigned long long)rsdp);
+	sprintf(buf, "0x%016"PRIxPTR, (uintptr_t)rsdp);
 	setenv("acpi.rsdp", buf, 1);
 	revision = rsdp->Revision;
 	if (revision == 0)
--- a/tools/boot/install-boot.sh
+++ b/tools/boot/install-boot.sh
@@ -42,11 +42,10 @@ get_uefi_bootname() {
 }
 
 make_esp_file() {
-    local file sizekb loader device stagedir fatbits efibootname
+    local file sizekb device stagedir fatbits efibootname
 
     file=$1
     sizekb=$2
-    loader=$3
 
     if [ "$sizekb" -ge "$fat32min" ]; then
         fatbits=32
@@ -58,8 +57,25 @@ make_esp_file() {
 
     stagedir=$(mktemp -d /tmp/stand-test.XXXXXX)
     mkdir -p "${stagedir}/EFI/BOOT"
-    efibootname=$(get_uefi_bootname)
-    cp "${loader}" "${stagedir}/EFI/BOOT/${efibootname}.efi"
+
+    # Allow multiple files to be copied.
+    # We do this in pairs, e.g:
+    # make_esp_file ... loader1.efi bootx64 loader2.efi bootia32
+    #
+    # If the second argument is left out,
+    # determine it automatically.
+    shift; shift # Skip $file and $sizekb
+    while [ ! -z $1 ]; do
+        if [ ! -z $2 ]; then
+            efibootname=$2
+        else
+            efibootname=$(get_uefi_bootname)
+        fi
+        cp "$1" "${stagedir}/EFI/BOOT/${efibootname}.efi"
+
+        shift; shift || : # Ignore failure to shift
+    done
+
     makefs -t msdos \
 	-o fat_type=${fatbits} \
 	-o sectors_per_cluster=1 \
@@ -70,13 +86,20 @@ make_esp_file() {
 }
 
 make_esp_device() {
-    local dev file mntpt fstype efibootname kbfree loadersize efibootfile
+    local dev file dst mntpt fstype efibootname kbfree loadersize efibootfile
     local isboot1 existingbootentryloaderfile bootorder bootentry
 
     # ESP device node
     dev=$1
     file=$2
+    # Allow caller to override the default
+    if [ ! -z $3 ]; then
+	    efibootname=$3
+    else
+	    efibootname=$(get_uefi_bootname)
+    fi
 
+    dst=$(basename ${file%.efi})
     mntpt=$(mktemp -d /tmp/stand-test.XXXXXX)
 
     # See if we're using an existing (formatted) ESP
@@ -93,7 +116,6 @@ make_esp_device() {
 
     echo "Mounted ESP ${dev} on ${mntpt}"
 
-    efibootname=$(get_uefi_bootname)
     kbfree=$(df -k "${mntpt}" | tail -1 | cut -w -f 4)
     loadersize=$(stat -f %z "${file}")
     loadersize=$((loadersize / 1024))
@@ -114,7 +136,7 @@ make_esp_device() {
         fi
     fi
 
-    if [ ! -f "${mntpt}/EFI/freebsd/loader.efi" ] && [ "$kbfree" -lt "$loadersize" ]; then
+    if [ ! -f "${mntpt}/EFI/freebsd/${dst}.efi" ] && [ "$kbfree" -lt "$loadersize" ]; then
         umount "${mntpt}"
 	rmdir "${mntpt}"
         echo "Failed to update the EFI System Partition ${dev}"
@@ -126,24 +148,26 @@ make_esp_device() {
     mkdir -p "${mntpt}/EFI/freebsd"
 
     # Keep a copy of the existing loader.efi in case there's a problem with the new one
-    if [ -f "${mntpt}/EFI/freebsd/loader.efi" ] && [ "$kbfree" -gt "$((loadersize * 2))" ]; then
-        cp "${mntpt}/EFI/freebsd/loader.efi" "${mntpt}/EFI/freebsd/loader-old.efi"
+    if [ -f "${mntpt}/EFI/freebsd/${dst}.efi" ] && [ "$kbfree" -gt "$((loadersize * 2))" ]; then
+        cp "${mntpt}/EFI/freebsd/${dst}.efi" "${mntpt}/EFI/freebsd/${dst}-old.efi"
     fi
 
     echo "Copying loader to /EFI/freebsd on ESP"
-    cp "${file}" "${mntpt}/EFI/freebsd/loader.efi"
+    cp "${file}" "${mntpt}/EFI/freebsd/${dst}.efi"
 
-    if [ -n "${updatesystem}" ]; then
-        existingbootentryloaderfile=$(efibootmgr -v | grep "${mntpt}//EFI/freebsd/loader.efi")
+    # efibootmgr won't work on systems with ia32 UEFI firmware
+    # since we only use it to boot the 64-bit kernel
+    if [ -n "${updatesystem}" ] && [ ${efibootname} != "bootia32" ]; then
+        existingbootentryloaderfile=$(efibootmgr -v | grep "${mntpt}//EFI/freebsd/${dst}.efi")
 
         if [ -z "$existingbootentryloaderfile" ]; then
             # Try again without the double forward-slash in the path
-            existingbootentryloaderfile=$(efibootmgr -v | grep "${mntpt}/EFI/freebsd/loader.efi")
+            existingbootentryloaderfile=$(efibootmgr -v | grep "${mntpt}/EFI/freebsd/${dst}.efi")
         fi
 
         if [ -z "$existingbootentryloaderfile" ]; then
             echo "Creating UEFI boot entry for FreeBSD"
-            efibootmgr --create --label FreeBSD --loader "${mntpt}/EFI/freebsd/loader.efi" > /dev/null
+            efibootmgr --create --label FreeBSD --loader "${mntpt}/EFI/freebsd/${dst}.efi" > /dev/null
             if [ $? -ne 0 ]; then
                 die "Failed to create new boot entry"
             fi
--- a/usr.sbin/bsdinstall/scripts/bootconfig
+++ b/usr.sbin/bsdinstall/scripts/bootconfig
@@ -109,32 +109,40 @@ if [ -n "$(awk '{if ($2=="/boot/efi") printf("%s\n",$1);}' $PATH_FSTAB)" ]; then
     #	    i386)	ARCHBOOTNAME=ia32 ;; # no support for this in i386 kernels, rare machines
 	    *)		die "Unsupported arch $(uname -m) for UEFI install"
 	esac
+
+	if [ `sysctl -n machdep.efi_arch` == i386 ]; then
+		ARCHBOOTNAME=ia32
+		file=loader_ia32.efi
+	else
+		file=loader.efi
+	fi
+
 	BOOTDIR="/efi/boot"
 	BOOTNAME="${BOOTDIR}/boot${ARCHBOOTNAME}.efi"
 	FREEBSD_BOOTDIR="/efi/freebsd"
-	FREEBSD_BOOTNAME="${FREEBSD_BOOTDIR}/loader.efi"
+	FREEBSD_BOOTNAME="${FREEBSD_BOOTDIR}/${file}"
 	mntpt="$BSDINSTALL_CHROOT/boot/efi"
 
-	f_dprintf "Installing loader.efi onto ESP"
+	f_dprintf "Installing ${file} onto ESP"
 	mkdir -p "${mntpt}/${FREEBSD_BOOTDIR}" "${mntpt}/${BOOTDIR}"
-	cp "$BSDINSTALL_CHROOT/boot/loader.efi" "${mntpt}/${FREEBSD_BOOTNAME}"
+	cp "$BSDINSTALL_CHROOT/boot/${file}" "${mntpt}/${FREEBSD_BOOTNAME}"
 
 	#
-	# The following shouldn't be necessary. UEFI defines a way to
-	# specifically select what to boot (which we do via
-	# efibootmgr). However, virtual environments often times lack
-	# support for the NV variables efibootmgr sets. In addition,
-	# some UEFI implementations have features that interfere with
-	# the setting of these variables. To combat that, we install the
-	# default removable media boot file as a fallback if it doesn't
-	# exist. We don't install it all the time since that can
-	# interfere with other installations on the drive (like rEFInd).
+	# UEFI defines a way to specifically select what to boot
+	# (which we do via efibootmgr). However, if we booted from an ia32
+	# UEFI environment, we wouldn't have access to efirt. In addition,
+	# virtual environments often times lack support for the NV variables
+	# efibootmgr sets, and some UEFI implementations have features that
+	# interfere with the setting of these variables. To combat that, we
+	# install the default removable media boot file if it doesn't exist.
+	# We don't install it all the time since that can interfere with other
+	# installations on the drive (like rEFInd).
 	#
 	if [ ! -f "${mntpt}/${BOOTNAME}" ]; then
-		cp "$BSDINSTALL_CHROOT/boot/loader.efi" "${mntpt}/${BOOTNAME}"
+		cp "$BSDINSTALL_CHROOT/boot/${file}" "${mntpt}/${BOOTNAME}"
 	fi
 
-	if [ "$BSDINSTALL_CONFIGCURRENT" ]; then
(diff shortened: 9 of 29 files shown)


VexedUXR:
Cant reproduce build errors locally.
Maybe this can be fixed by using __aligned to force padding?

VexedUXR:
Clang seems to ignore -Wint-to-pointer-cast and -Wpointer-to-int-cast when -m32 is specified.
Can reproduce errors with gcc toolchain.

bsdimp:
This makes it easier for me to look at this and to have a conversation. I got too busy last time and naybe this way we'll keep momentum.

VexedUXR:
Fixed.
Everything looks fine now

VexedUXR:
Fixed rsdp cast

bsdimp:
OK. I've ordered an inexpensive laptop to see if I can use it to test this.
Generally I like this, but it's all one commit
and I've left a few comments, but I'm thinking about others
Some of this I think that I can land quickly, but other bits need a little refinement. If things are broken down better I can do that. I'll tag a couple of things first.

share/man/man5/src.conf.5
```diff
@@ -1,5 +1,5 @@
 .\" DO NOT EDIT-- this file is @generated by tools/build/options/makeman.
-.Dd November 13, 2023
+.Dd February 3, 2024
```
bsdimp, line 2: I'd omit this generated file from the review. Other people's changes will just get in the way.

stand/defs.mk
```diff
@@ -38,7 +38,11 @@ WARNS?=		1
 BOOTSRC=	${SRCTOP}/stand
 EFISRC=		${BOOTSRC}/efi
 EFIINC=		${EFISRC}/include
+.if ${MACHINE_ARCH} == "amd64" && ${DO32:U0} == 1
```
bsdimp, line 41: A quick comment here about WHY might be good.

stand/efi/libefi32/Makefile
```diff
@@ -0,0 +1,5 @@
+
+DO32=1
+
+.PATH:	${.CURDIR}/../libefi
+.include "${.CURDIR}/../libefi/Makefile"
```
bsdimp, line 5: This and the efi/Makefile change should be bundled together.

stand/efi/loader/Makefile
```diff
@@ -81,7 +81,7 @@ CFLAGS+=	-DEFI_STAGING_SIZE=${EFI_STAGING_SIZE}
 CFLAGS+= -DEFI_SECUREBOOT
 .endif
 
-NEWVERSWHAT=	"EFI loader" ${MACHINE}
+NEWVERSWHAT?=	"EFI loader" ${MACHINE}
```
bsdimp, line 84: this one chunk can be done in a commit with a good commit message about allowing it to be overriden. It's otherwise ready to land.

stand/efi/loader/arch/amd64/ia32/amd64_tramp32.S
```diff
@@ -0,0 +1,97 @@
+/*-
+ * Copyright (c) 2013 The FreeBSD Foundation
```
bsdimp, line 2: where did this come from? The old 32-bit uefi stuff I deleted?

stand/efi/loader/bootinfo.c
```diff
@@ -186,7 +186,7 @@ bi_load_efi_data(struct preloaded_file *kfp, bool exit_bs)
 	struct efi_map_header *efihdr;
 	bool do_vmap;
 
-#if defined(__amd64__) || defined(__aarch64__)
+#if defined(__amd64__) || defined(__aarch64__) || defined(__uefi64__)
```
bsdimp, line 189: The __uefi64__ bothers me a bit. What does it really mean?

usr.sbin/bsdinstall/scripts/bootconfig
```diff
@@ -132,6 +132,9 @@ if [ -n "$(awk '{if ($2=="/boot/efi") printf("%s\n",$1);}' $PATH_FSTAB)" ]; then
 	#
 	if [ ! -f "${mntpt}/${BOOTNAME}" ]; then
 		cp "$BSDINSTALL_CHROOT/boot/loader.efi" "${mntpt}/${BOOTNAME}"
+		if [ -e "$BSDINSTALL_CHROOT/boot/loader-ia32.efi" ]; then
+			cp "$BSDINSTALL_CHROOT/boot/loader-ia32.efi" "${mntpt}/${BOOTDIR}/bootia32.efi"
+		fi
```
bsdimp, line 137: this could also be done in a separate commit.

bsdimp (changes requested):
Let's get started at breaking this down into smaller chunks, but I think we can do that as a series of commits. Remember to `git push --force-with-lease` to update the changes and to rebase forward please. Thanks!

stand/defs.mk
VexedUXR: Because libefi uses it. Or do you mean a comment in the actual makefile?

usr.sbin/bsdinstall/scripts/bootconfig
VexedUXR: Should it be bundled with the bsdinstall change?

VexedUXR: install-boot change**

stand/defs.mk
VexedUXR: Would it be better to move the comment explaining DO32 up here?

bsdimp:
I'll take a look at this.
IIRC, I can use qemu-system-x86_64 to test this if I load the i386 EDKII firmware, right?
Thanks for the updates.

VexedUXR:
> IIRC, I can use qemu-system-x86_64 to test this if I load the i386 EDKII firmware, right?

Yea, thats how I did it

VexedUXR:
Just a little cleanup: Use NULL instead of 0.
Also, put the 32-bit compatibility fixes in their own commit.

stand/efi/loader/bootinfo.c
bsdimp: Shouldn't __i386__ be defined in this case?  I think that will suffice if it is... If we need more nuance in the future, we can go there then...  But since i386 is on its way out of FreeBSD and the number of i386 platforms that will start using it is nil,  I don't think we'll ever need anything else.

VexedUXR: Yea that makes sense, it would be the same with \_\_i386__ and if support for i386 is getting dropped, we can safely use it.

VexedUXR:
> I'll try to get to this before we have the 14.1 freeze.

Great :)

VexedUXR:
I've changed all of the `__uefi64__` to `__i386__` and fixed the merge conflicts, I also updated the linker script stuff in accordance with 5b3b9a5. I added my name to the copyright notice at the top of the amd64_tramp32.S file since I rewrote most of it, let me know if you dont what that.

Also, thers a typo at https://github.com/freebsd/freebsd-src/blob/main/stand/liblua/lutils.h#L37 and I have no idea why it causes the 32-bit loader to crash, but not the 64-bit one, probably some linker weirdness.

bsdimp:
I thought I'd fixed that typo... It should make it not compile at all..

VexedUXR:
Wouldn't it just make a set called `_LUA_COMPILE_SET` instead of `Xlua_compile_set`

As long as the set isn't accessed outside the macros, there shouldn't be any problems
Of course in our case, it is causing problems since we `objcopy -j set_Xlua_compile_set`

VexedUXR:
I was actually interested in finding out why it only crashed on the 32-bit loader, turns out its fairly straightforward.

With the 32-bit loader, the [addend is expected to be present](https://github.com/freebsd/freebsd-src/blob/main/stand/common/self_reloc.c#L115) in the `set_Xlua_compile_set` section, which doesn't exist in this case. The address this section is expected to start at is just full of uninitialized data, so when `LUA_FOREACH_SET` is used we end up jumping to some garbage and crashing. With the 64-bit loader [we get the addend from the relocation entry](https://github.com/freebsd/freebsd-src/blob/main/stand/common/self_reloc.c#L112), so thats not a problem.

Ah, also, while looking into this I noticed that the i386 script I re-added wasn't up to date with d024bc7, I'll check the rest of the script and push a fix soon.

bsdimp:
This looks closish to being good to go. I'll test it out.

stand/efi/loader/framebuffer.c
```diff
@@ -650,17 +651,22 @@ efi_find_framebuffer(teken_gfx_t *gfx_state)
 	gfx_state->tg_fb.fb_bpp = fls(efifb.fb_mask_red | efifb.fb_mask_green |
 	    efifb.fb_mask_blue | efifb.fb_mask_reserved);
 
-	if (gfx_state->tg_shadow_fb != NULL)
-		BS->FreePages((EFI_PHYSICAL_ADDRESS)gfx_state->tg_shadow_fb,
+	/*
+	 * EFI_PHYSICAL_ADDRESS is 8 bytes wide on 32-bit systems too,
```
emaste, line 655: be explicit about what the "too" is referring to -- e.g. `on both 32- and 64-bit systems`

stand/efi/loader/Makefile
```diff
@@ -97,10 +97,10 @@ FILES+=	${LOADER}.efi
 FILESMODE_${LOADER}.efi=	${BINMODE}
 
 .if ${LOADER_INTERP} == ${LOADER_DEFAULT_INTERP}
-LINKS+=		${BINDIR}/${LOADER}.efi ${BINDIR}/loader.efi
+LINKS?=		${BINDIR}/${LOADER}.efi ${BINDIR}/loader.efi
```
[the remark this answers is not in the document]
VexedUXR: I added a comment explaining `LINKS`, `LDSCRIPT`, `DPADD` and `LDADD`. Not sure if I should include `NEWVERSWHAT` since its kind of its own thing.

stand/efi/loader/framebuffer.c
```diff
@@ -650,17 +651,22 @@ efi_find_framebuffer(teken_gfx_t *gfx_state)
...
 	    efifb.fb_mask_blue | efifb.fb_mask_reserved);
 
-	if (gfx_state->tg_shadow_fb != NULL)
-		BS->FreePages((EFI_PHYSICAL_ADDRESS)gfx_state->tg_shadow_fb,
+	/*
+	 * EFI_PHYSICAL_ADDRESS is 8 bytes wide on both 32 and 64-bit
+	 * systems, so we cant just use a pointer.
+	 */
+	tg_shadow_fb = (uintptr_t)gfx_state->tg_shadow_fb;
+	if (tg_shadow_fb)
```
bsdimp, line 659: This should be if (tg_shadow_fb != 0) since it's a number, not a boolean.

stand/efi/loader/framebuffer.c
```diff
@@ -650,17 +651,22 @@ efi_find_framebuffer(teken_gfx_t *gfx_state)
...
 	    EFI_SIZE_TO_PAGES(efifb.fb_height * efifb.fb_width *
 	    sizeof(EFI_GRAPHICS_OUTPUT_BLT_PIXEL));
 	status = BS->AllocatePages(AllocateMaxAddress, EfiLoaderData,
 	    gfx_state->tg_shadow_sz,
-	    (EFI_PHYSICAL_ADDRESS *)&gfx_state->tg_shadow_fb);
-	if (status != EFI_SUCCESS)
-		gfx_state->tg_shadow_fb = NULL;
+	    &tg_shadow_fb);
+	gfx_state->tg_shadow_fb = status == EFI_SUCCESS ?
+	    (uint32_t *)(uintptr_t)tg_shadow_fb : NULL;
```
bsdimp, line 669: If this is above 4G, I think this truncates in the using the 32-bit loader, no?

stand/efi/loader/arch/amd64/ia32/elf64_freebsd32.c
```diff
@@ -0,0 +1,282 @@
...
+	}
+	trampcode = ptr64;
+	bzero((void *)(uintptr_t)trampcode, EFI_PAGE_SIZE);
+	trampstack = trampcode + EFI_PAGE_SIZE - 8;
+	bcopy((void *)&amd64_tramp, (void *)(uintptr_t)trampcode,
+		amd64_tramp_size);
+	trampoline = (void *)(uintptr_t)trampcode;
+
+	ptr64 = copy_staging == COPY_STAGING_ENABLE ?
+	    0x0000000040000000 /* 1G */ :
```
bsdimp, line 145: This should be styled as `0x0000000040000000 : /* 1G */`

stand/efi/loader/arch/amd64/ia32/elf64_freebsd32.c
```diff
@@ -0,0 +1,282 @@
...
+	    0x0000000100000000; /* 4G */;
```
bsdimp, line 146: Delete the ';' here at the end to fix the style issue.

stand/efi/loader/arch/amd64/ia32/elf64_freebsd32.c
```diff
@@ -0,0 +1,282 @@
...
+	if (copy_auto)
+		copy_staging = fp->f_kernphys_relocatable ?
+		    COPY_STAGING_DISABLE : COPY_STAGING_ENABLE;
+
+	if ((md = file_findmetadata(fp, MODINFOMD_ELFHDR)) == NULL)
+		return (EFTYPE);
+	ehdr = (Elf_Ehdr *)&(md->md_data);
+
+	ptr64 = copy_staging == COPY_STAGING_ENABLE ?
+	    0x0000000040000000 /* 1G */ :
```
bsdimp, line 127: Change to `0x0000000040000000 : /* 1G */ `

stand/efi/loader/arch/amd64/ia32/elf64_freebsd32.c
```diff
@@ -0,0 +1,282 @@
...
+	    0x0000000100000000; /* 4G */;
```
bsdimp, line 128: delete the ; at the end

stand/efi/loader/arch/amd64/ia32/elf64_freebsd32.c
```diff
@@ -0,0 +1,282 @@
...
+	}
+	GDT = (gdt_t *)(uintptr_t)ptr64;
+	bzero(GDT, EFI_PAGE_SIZE);
+	GDT[1] = GDT_P | GDT_E | GDT_S | GDT_RW | GDT_L; /* CS */
+	gdtr = (struct GDTR *)&GDT[2];
+	gdtr->ptr = (uintptr_t)GDT;
+	gdtr->size = 2 * sizeof(uint64_t) - 1;
+
+	if (copy_staging == COPY_STAGING_ENABLE) {
+		ptr64 = 0x0000000040000000; /* 1G */
```
bsdimp, line 164: Picking a random spot... Maybe we should have constants for 1G and 4G...  I think existing code might be like that, though, and should be fixed too.

stand/efi/loader/bootinfo.c
```diff
@@ -433,7 +438,7 @@ bi_load(char *args, vm_offset_t *modulep, vm_offset_t *kernendp, bool exit_bs)
 	module = *modulep;
 	file_addmetadata(kfp, MODINFOMD_MODULEP, sizeof(module), &module);
 #endif
-#ifdef EFI
+#if defined(EFI) && !defined(__i386__)
```
bsdimp, line 441: Why omit this on x86?

stand/efi/loader/bootinfo.c
```diff
@@ -339,7 +339,12 @@ bi_load(char *args, vm_offset_t *modulep, vm_offset_t *kernendp, bool exit_bs)
 	vm_offset_t size;
 	char *rootdevname;
 	int howto;
-	bool is64 = sizeof(long) == 8;
+	bool is64 =
+#ifdef __i386__
+	true;
+#else
+	sizeof(long) == 8;
+#endif
```
bsdimp, line 347: Isn't the above always true now?

stand/efi/loader/copy.c
```diff
@@ -380,7 +381,7 @@ efi_check_space(vm_offset_t end)
 	end += staging_slop;
 
 	nr_pages = EFI_SIZE_TO_PAGES(end - staging_end);
-#if defined(__amd64__)
+#if defined(__amd64__) || defined(__i386__)
 	/*
 	 * amd64 needs all memory to be allocated under the 1G or 4G boundary.
```
bsdimp, line 386: This coment needs to be adjusted

stand/efi/loader/efi_main.c
```diff
@@ -87,6 +87,15 @@ efi_main(EFI_HANDLE image_handle, EFI_SYSTEM_TABLE *system_table)
 	BS = ST->BootServices;
 	RS = ST->RuntimeServices;
 
+#ifdef __i386__
+	/* Check for long mode */
+	asm("cpuid" : "=d"(status) : "a"(0x80000001) : "%ebx", "%ecx");
+	if (!(status & 0x20000000)) {
+		ST->ConOut->OutputString(ST->ConOut, (CHAR16 *)L"No long mode.\r\n");
```
bsdimp, line 94: I'd wrap the message, and expaned the message to be 'No long mode support on this CPU.' Maybe with a 'Unable to proceed'

stand/efi/loader/framebuffer.c
```diff
@@ -650,17 +651,22 @@ efi_find_framebuffer(teken_gfx_t *gfx_state)
 	gfx_state->tg_fb.fb_bpp = fls(efifb.fb_mask_red | efifb.fb_mask_green |
 	    efifb.fb_mask_blue | efifb.fb_mask_reserved);
 
-	if (gfx_state->tg_shadow_fb != NULL)
-		BS->FreePages((EFI_PHYSICAL_ADDRESS)gfx_state->tg_shadow_fb,
+	/*
+	 * EFI_PHYSICAL_ADDRESS is 8 bytes wide on both 32 and 64-bit
+	 * systems, so we cant just use a pointer.
+	 */
+	tg_shadow_fb = (uintptr_t)gfx_state->tg_shadow_fb;
```
bsdimp, line 658: Why cast through a uintptr_t when EFI_PHYSICAL_ADDRESS is always 8 bytes? It seems weird.

stand/efi/loader/arch/amd64/ia32/elf64_freebsd32.c
```diff
@@ -0,0 +1,282 @@
...
+static gdt_t *GDT;
+
+extern EFI_PHYSICAL_ADDRESS staging;
+
+static void (*trampoline)(uint32_t stack, void *copy_finish, uint32_t kernend,
+    uint32_t modulep, uint64_t *pagetable, struct GDTR *gdtr, uint64_t entry);
+
+extern uintptr_t amd64_tramp;
+extern uint32_t amd64_tramp_size;
+
```
bsdimp, line 95: all these externs should live in a .h file

bsdimp (changes requested):
Found a few more changes that I think we need still.. This is much improved over the first pass...  Just had some size questions...

bsdimp:
OK. Another round of comments. Apart from the style nits that the style checker highlighted (which I gave advice on), I have worries about uintptr_t only being 4 bytes and whether or not that's a problem.

stand/efi/loader/framebuffer.c
VexedUXR: If `AllocatePages` was called with `AllocateAnyPages`, then that would be true yes. However, to my surprise, its called with `AllocateMaxAddress` which means that its treats the contents of `tg_shadow_fb` as the highest address it could give us. Since `gfx_state->tg_shadow_fb` starts out as NULL, the `AllocatePages` call fails every time...
This looks like it has been broken since 221376db0cb77035e325baf8b54d6224925b41f8

The following fixes it and insures that `ptr` is below 4G:
```
EFI_PHYSICAL_ADDRESS ptr = UINT32_MAX;

...

if (gfx_state->tg_shadow_fb != NULL)
	BS->FreePages((EFI_PHYSICAL_ADDRESS)gfx_state->tg_shadow_fb,
	    gfx_state->tg_shadow_sz);
gfx_state->tg_shadow_sz =
    EFI_SIZE_TO_PAGES(efifb.fb_height * efifb.fb_width *
    sizeof(EFI_GRAPHICS_OUTPUT_BLT_PIXEL));

/*
 * EFI_PHYSICAL_ADDRESS is 8 bytes wide on both 32 and 64-bit
 * systems, so we cant just use a pointer.
 */
status = BS->AllocatePages(AllocateMaxAddress, EfiLoaderData,
    gfx_state->tg_shadow_sz, &ptr);
gfx_state->tg_shadow_fb = status == EFI_SUCCESS ?
    (uint32_t *)(uintptr_t)ptr : NULL;

return (0);
```

It looks like what you're describing also happens with `heap` in `efi_main.c`.

Nice catch

stand/efi/loader/bootinfo.c
VexedUXR: Some of its members are `UINTN`s, which is 8 bytes on amd64 and 4 bytes on x86. Even if the kernel was modified to detect this, all the efirt functions are required to be run in protected mode.

stand/efi/loader/bootinfo.c
VexedUXR: Yes, that's the point, to force the alignment that the 64-bit kernel would expect.

stand/efi/loader/copy.c
VexedUXR: I interpreted that as "the amd64 kernel needs all memory ...", as the above comments were also taking about the kernel.

stand/efi/loader/efi_main.c
VexedUXR: Not sure what you mean by wrap, but I've changed it to: `This CPU doesn't support long mode.\r\nUnable to proceed.\r\n`

stand/efi/loader/arch/amd64/ia32/elf64_freebsd32.c
VexedUXR: Agreed, but the other elf64_freebsd files also do this.
These should maybe go with G(X) and M(X), wherever they go.

VexedUXR:
Ok, I've addressed the comments but I don't think I'll push yet as it looks like there's still some discussion to be had.

stand/efi/loader/bootinfo.c
VexedUXR: Oh, I think I missed your point here. You meant its always true as in we are always loading a 64-bit kernel?

I thought 32-bit arm was still supported, if not then yea we can just remove the checks.

hhartzer:
This is very cool! Although, I guess this didn't make it into 14.1-BETA 1?

I have a laptop that I can test this with.

VexedUXR:
> I have a laptop that I can test this with.

Feel free :), you should be able to just build an img and run it.

hhartzer:
I'll try! Looks like it applies cleanly to 14.1 but not 14.0.

VexedUXR:
Just a heads up, you may need to fix https://github.com/freebsd/freebsd-src/blob/main/stand/liblua/lutils.h#L37
It should be "COMPILE".

hhartzer:
Interesting! Should that be fixed in this PR, or another?

VexedUXR:
Yea its out of scope for this one.

Though, its a pretty trivial fix and @bsdimp already knows about this, (he probably just forgot, can't blame him, I did too until yesterday). So I never really bothered making a new PR.

bsdimp:
Oh, that one. I should do it and try to get it into BETA3.

stand/efi/loader/framebuffer.c
VexedUXR: Hmm, doing some reading, the UEFI spec in section 2.3.2.1 says:
>All descriptors are set to their 4GiB limits so that all of memory is accessible from all segments.

I guess that implies that, within the memory controlled by UEFI at least, `AllocatePages` can't actually give us an address over 2^30?

Though it also says in section 2.3.2:
>Paging mode may be enabled. If paging mode is enabled, PAE (Physical Address Extensions)
mode is recommended.

So IMO it's a safe bet to do this anyways to avoid overflows on any weird implementations.

hhartzer:
I took 14.1-release and applied this patch to it, plus the typo fix. I made a disc1.iso image.

It books in UEFI mode to the bootloader. Boot loader works, then it loads the kernels.

There's immediately a crash and a reboot.

I captured a video and it seems to  show the EFI framebuffer information and then reboots immediately.

I did disable `COMPAT_FREEBSD32` in this build. Might that matter in this instance?

This laptop is using a modified BIOS for 64 bit support (albeit, via BIOS installs).

The installer does boot in BIOS mode. I'm assuming if the installer doesn't work with 32 bit UEFI, I shouldn't try installing it that way yet.

This makes me wonder if your patch should work for the unmodified BIOS, which does not support 64 bit, to use 32 bit UEFI and then install in 64 bit mode? I really have no idea.

Thanks!

bsdimp:
typo commit pushed.

VexedUXR:
>I captured a video and it seems to show the EFI framebuffer information and then reboots immediately.

Hm, that happens before the kernel handoff, but a a lot of things happen in between. Could you add an `asm("hlt");` or a `while(1);` right above where the trampoline is called? It's at line 270 in `stand/efi/loader/arch/amd64/ia32/elf64_freebsd32.c`.
Not sure how else to verify where the crash is happening... But this should narrow it down a bit.

>This laptop is using a modified BIOS for 64 bit support

Interesting, what does this modification do?
By add 64-bit support do you mean add CSM?

Also, how much ram does this machine have? Might be something related to what Warner mentioned.

hhartzer:
4GB of RAM. It's a Atom N2600. The BIOS locks out 64 bit functionality by default, for some reason.

I think you might be right about CSM. This is the BIOS in question (although not 100% certain that's the same as the one I'm using now, but I assume functionally identical if not identical).

https://www.bios-mods.com/forum/showthread.php?pid=69563#pid69563

Is there a way I can do a faster recompile after that change?

Thanks!

VexedUXR:
>Is there a way I can do a faster recompile after that change?

`make -C stand efi`

The loader should be at `/usr/obj/.../amd64.amd64/stand/efi/loader_lua/ia32/loader_lua-ia32.efi`
Then you can manually copy it to the efi partition, replacing `bootia32.efi`

hhartzer:
Thank you! That was helpful.

Unfortunately, I'm not seeing more than I did before, although this is easier to read than a frame in the video.

I'll try to type this out as best as I can.

```
Loading kernel...
(...)
can't find '/boot/entropy'
can't find '/etc/hostid'
staging 0xbb4000000 (not copying) tramp 0xbf3c7000 PT4 0xbb34d000 GDT 0xbf3b1000
Start @ 0xffffffff8037c000 ...
EFI framebuffer information:
addr, size 0xffffffffbf800000, 0x7b0000
dimensions 800 x 600
stride     800
masks      0x00ff0000, 0x0000ff00, 0x000000ff, 0xff000000
```

VexedUXR:
>Unfortunately, I'm not seeing more than I did before

There's some good info here :)

That framebuffer address is definitely wrong. Does your laptop use UGA by any chance?
What make and model is it?

hhartzer:
Great!

I'm not familiar with UGA. How would I determine that?

It's a Asus EEEPC 1025C. Resolution should be 1024x600. When the boot loader loads, it looks like 800x600.

VexedUXR:
Escape the loader prompt and type `show efi-version`. AFAIK UGA was only used on EFI 1.x

Also, do you have discord? It would be easier to talk and troubleshoot on there rather than relying on email notifications.

hhartzer:
`show efi-version` says 2.00.

I'll see if I can get a Discord account together.

VexedUXR:
Update: After some troubleshooting, we managed to find the bug. Looks like the bios was setting the higher half of the fb address (my best guess is some sign-extension fun). Fixed by `& 0xffffffff`ing the fb address, only on the i386 build.

@bsdimp, apart from some review comments above, I have the commits ready. The main thing being about what the UEFI spec says about the max available address on ia32, I'm not sure what to make of it.

Thanks again @hhartzer!

bsdimp:
I'll review where we are with this.

stand/loader.mk
```diff
@@ -167,13 +167,18 @@ CFLAGS+=	-DELF_VERBOSE
 .error Define HELP_FILENAME before including loader.mk
 .endif
 
+CFLAGS+=	-DHELP_FILENAME=\"${HELP_FILENAME}\"
+
+# If we are making the ia32 loader, we dont need to make another
+# help file.
```
jrtc27, line 173: What's wrong with just setting INSTALL_LOADER_HELP_FILE in the including Makefile?

tools/boot/install-boot.sh
```diff
@@ -60,6 +60,10 @@ make_esp_file() {
     mkdir -p "${stagedir}/EFI/BOOT"
     efibootname=$(get_uefi_bootname)
     cp "${loader}" "${stagedir}/EFI/BOOT/${efibootname}.efi"
+    loader=$(echo ${loader} | sed s,.efi,-ia32.efi,)
```
jrtc27, line 63: . isn't escaped, you probably should anchor it, and all of this can be just done in a plain shell expansion with `${loader%.efi}-ia32.efi`

usr.sbin/bsdinstall/scripts/bootconfig
```diff
@@ -132,6 +132,9 @@ if [ -n "$(awk '{if ($2=="/boot/efi") printf("%s\n",$1);}' $PATH_FSTAB)" ]; then
 	#
 	if [ ! -f "${mntpt}/${BOOTNAME}" ]; then
 		cp "$BSDINSTALL_CHROOT/boot/loader.efi" "${mntpt}/${BOOTNAME}"
+		if [ -e "$BSDINSTALL_CHROOT/boot/loader-ia32.efi" ]; then
+			cp "$BSDINSTALL_CHROOT/boot/loader-ia32.efi" "${mntpt}/${BOOTDIR}/bootia32.efi"
```
jrtc27, line 136: Should /efi/freebsd/loader.efi not be the thing your system needs to boot? Or at least have a separate /efi/freebsd/loader(-ia)32.efi? Otherwise you're only using the removable media boot file, which is bad practice.

stand/efi/loader/arch/amd64/Makefile.inc
```diff
@@ -1,15 +1,48 @@
...
-	spinconsole.c
-
 CFLAGS+=	-fPIC
 LDFLAGS+=	-Wl,-znocombreloc
+
+.if ${MK_LOADER_IA32} != "no"
+# Make the 32-bit loader before we start
+.BEGIN: loader32
+loader32: .PHONY
+	${MAKE} -C ${.CURDIR} MAKEOBJDIR=${.OBJDIR}/ia32 DO32=1 ${.TARGETS}
```
jrtc27, line 46: Why can't we just have a Makefile in the ia32 subdir, use normal SUBDIR+= and have it set DO32 automatically? Explicit recursive make combined with .PHONY is a bit gross.

stand/efi/loader/arch/amd64/ia32/elf64_freebsd32.c
```diff
@@ -0,0 +1,282 @@
...
+	struct file_metadata	*md;
+	struct GDTR	*gdtr;
+	Elf_Ehdr 		*ehdr;
+	vm_offset_t		modulep, kernend, trampcode, trampstack;
+	int			err, i;
+	bool			copy_auto;
+
+	/*
+	 * EFI_PHYSICAL_ADDRESS is 8 bytes wide on both 32 and 64-bit
+	 * systems, so we cant just use a pointer.
```
jrtc27, line 113: "can't"

jrtc27: But I don't get the point of this comment. AllocatePages is defined as taking an `EFI_PHYSICAL_ADDRESS *` so of course that's what we should use?

stand/efi/loader/arch/amd64/ia32/elf64_freebsd32.c
```diff
@@ -0,0 +1,282 @@
...
+	    efi_copy_finish : efi_copy_finish_nop, kernend, modulep,
+	    PT4, gdtr, ehdr->e_entry);
+
+	panic("exec returned");
+}
+
+static int
+elf64_obj_exec(struct preloaded_file *fp)
+{
+
```
jrtc27, line 280:
```suggestion
```

stand/efi/loader/framebuffer.c
VexedUXR: >I don't think I'll push yet as it looks like there's still some discussion to be had.

I'll try to separate the changes that are still being discussed and push after applying the changes you suggested.

usr.sbin/bsdinstall/scripts/bootconfig
VexedUXR: The problem here is that `efibootmgr` relies on `efirt`, but since the ia32 loader doesn't provide the kernel with the EFI system table, we can't use it.
I disabled it mainly because the efirt functions expect to be run in 32-bit protected mode.

jrtc27: I realise, but it's confusing to have a single FreeBSD-specific file that's not the right thing, plus if something tramples on the removable media path then you've got no copy on your disk to manually load.

stand/efi/loader/arch/amd64/Makefile.inc
VexedUXR: Do you mean a makefile in each `loader_*`?

If so, it would probably look something like this:
`loader_lua/Makefile`
```make
LOADER_INTERP=lua
INSTALL_LOADER_HELP_FILE=no

.if ${MK_LOADER_IA32:Uyes} != "no"
SUBDIR+=        ia32
.endif
  
.include "../loader/Makefile"
```
`loader_lua/ia32/Makefile`
```make
DO32=1
LOADER_INTERP=lua
INSTALL_LOADER_HELP_FILE=no
NEWVERSWHAT="EFI loader" amd64-ia32

.include "../../loader/Makefile"
```
This is an improvement, but the problem here is that loader's makefile expects `${.CURDIR}/../loader` to work and I'm not sure how to get around that properly.

stand/efi/loader/arch/amd64/ia32/elf64_freebsd32.c
VexedUXR: That's fair enough. I added this here after fixing the issue with `framebuffer.c`, should that comment also be removed?

usr.sbin/bsdinstall/scripts/bootconfig
VexedUXR: That makes sense

VexedUXR:
Pushed the changes that aren't being discussed / that I don't have questions about.
Also rebased to eb690a05

bsdimp:
OK. I landed a little bit of this PR, and updated the PR to rebase it and resolve the conflicts that I created due to minor tweaking I did. Hope that doesn't mess up any other WIP around this.

stand/efi/loader/arch/amd64/Makefile.inc
VexedUXR: I found another way to use `SUBDIR` instead of calling make explicitly.
```
$ cd stand/efi
$ ls
... loader-ia32 loader loader_lua ...
```
```
$ cat loader-ia32/Makefile 

.PATH: ${.CURDIR}/../loader/arch/amd64/ia32

# LOADER_INTERP should be exported by
# the parent.
.if empty(LOADER_INTERP)
.error Expected LOADER_INTERP to be defined.
.endif

DO32=1
NEWVERSWHAT="EFI loader" amd64-ia32
INSTALL_LOADER_HELP_FILE=no

.include "../loader/Makefile"
```
```
$ cat loader/arch/amd64/Makefile.inc 

.PATH:  ${BOOTSRC}/i386/libi386
SRCS+=  nullconsole.c \
        comconsole.c \
        spinconsole.c

CFLAGS+=        -fPIC
LDFLAGS+=       -Wl,-znocombreloc

.if ${DO32:U0} == 0
SRCS+=  amd64_tramp.S \
        start.S \
        elf64_freebsd.c \
        trap.c \
        multiboot2.c \
        exc.S

MAKEOBJDIR:=    ${.OBJDIR}/ia32

.export MAKEOBJDIR
.export LOADER_INTERP

SUBDIR.${MK_LOADER_IA32}+=      ../loader-ia32

.elif ${MK_LOADER_IA32} == "yes"
LOADER:=        ${LOADER}-ia32

CFLAGS+=        -I${EFISRC}/include/i386

LDSCRIPT=       ${.CURDIR}/../loader/arch/amd64/ia32/i386.ldscript
.if ${LOADER_INTERP} == ${LOADER_DEFAULT_INTERP}
LINKS=  ${BINDIR}/${LOADER}.efi ${BINDIR}/loader-ia32.efi
.endif

SRCS+=  amd64_tramp32.S \
        start32.S \
        elf64_freebsd32.c

LIBEFI32=       ${BOOTOBJ}/efi/libefi32/libefi.a
EFI_TARGET=     efi-app-ia32

DPADD=  ${LDR_INTERP32} ${LIBEFI32} ${LIBSA32} ${LDSCRIPT}
LDADD=  ${LDR_INTERP32} ${LIBEFI32} ${LIBSA32}
.endif
```
Unlike the other solution I gave this works, but there's that ugly `MAKEOBJDIR` hack.
Could you elaborate on your initial suggestion? I'm not sure how else to do this.

VexedUXR:
Okay, a lot of changes this time. Hopefully I don't forget anything.
Instead of scattering everything across a bunch of review comments, I'll just put everything in one place.

### We can't get addresses over 4G from UEFI
The UEFI spec requires all memory to be identity mapped if paging is enabled. It also allows PAE, but that doesn't change the size of the virtual address space. So we don't have to worry about getting addresses over 4G.
### `elf64_freebsd32.c` cleanup
Given that last point, we can simplify this file and imo make it more readable. There was no functional change here just some refactoring. There are still some externs in the file though, same as the original `elf64_freebsd.c`. It would be good to move those but I don't know what header those externs would go in. I also copied the G(x) and M(x) macros from `copy.c` to use them here, that should also be in a header but I'm not sure which one. (Maybe save this for another PR?)
### Using `SUBDIR` instead of explicitly calling make
This one took me a few tries, as seen [here](https://github.com/freebsd/freebsd-src/pull/1098#discussion_r1604341384). I ended up just creating a new makefile for each loader, e.g `loader_lua-ia32`. I was reluctant to do it this way in the beginning since it involved a lot of duplicated makefiles, but I feel like every other way I found to do this either had some caveat or was close to working but would fail on `make install` (also they were all pretty janky). I did actually find one other way that I didn't post here, but it involved messing with one of `auto.obj.mk`'s internal variables so I feel like that was a no-go.
### Cleanup/Fixing the copying of the ia32 loader in the scripts
This worked* before, but it would just jankily copy the ia32 loader along with the 64-bit one if it found it. In order to make this better, I needed a way to figure out which one we needed. I was using `efibootmgr` then checking if it failed to figure out if we booted from the ia32 loader. This had two problems: 1. It looked pretty cryptic, so I probably needed to comment what was happening each time I used it. 2: One could simply not have `efirt` loaded and it would give a false positive. In order to get around this I added a new sysctl, `machdep.efi_arch` which would report the UEFI architecture we booted from using metadata given by the loader. This made it much simpler to detect what we booted from (though there are still some weird cases, which I'll get onto now). This works just fine in the bsdinstall script, we simply check the sysctl then copy the required loader. In `install-boot.sh` however, it gets a little more complicated. We try to figure out which loader we need by checking a few things. If `updatesystem` is set (which implies that we are booting on the same system the script is running on) and the `machdep.efi_arch` sysctl is `i386`, we can safely assume that we only need the ia32 loader. We also check for an existing loader, if there is one, we use `file` to figure out weather it's the ia32 one or not then copy the new one accordingly. However, if we can't figure out which one we need, we just copy both. The 64-bit one goes to `loader.efi` and the 32-bit one goes to `loader-ia32.efi`.

That turned out longer than expected, hopefully it wasn't too long.
Hopefully I can catch a Cirrus CI run :)
Thanks.

VexedUXR:
Kernel build failures seem unrelated (it doesn't even get to `buildkernel`).
Can confirm that my local `buildworld` and `buildkernel` work.

stand/efi/loader/arch/amd64/Makefile.inc
```diff
@@ -1,15 +1,40 @@
...
 	elf64_freebsd.c \
 	trap.c \
 	multiboot2.c \
 	exc.S
 
-.PATH:	${BOOTSRC}/i386/libi386
-SRCS+=	nullconsole.c \
-	comconsole.c \
-	spinconsole.c
+SUBDIR.${MK_LOADER_IA32}+=	../loader_${LOADER_INTERP}-ia32
```
jrtc27, line 18: Why isn't this up in stand/efi/Makefile? That would be the normal way to do this

stand/efi/libefi32/Makefile
```diff
@@ -0,0 +1,5 @@
+
```
bsdimp, line 1: extra blank line

stand/efi/loader/arch/amd64/Makefile.inc
bsdimp: because different architectures have different consoles.
comconsole.c is x86 only, for example, and is a mistake from the past that bypasses UEFI entirely to allow better configurability of the serial console than UEFI can do w/o us adding ACPI parsing to it (so it's the least bad mistake we can likely make given the x86 expectations from the users).

stand/efi/loader/arch/amd64/Makefile.inc
```diff
@@ -1,15 +1,40 @@
...
+
+SRCS+=	amd64_tramp32.S \
+	start32.S \
+	elf64_freebsd32.c
+
+LIBEFI32=	${BOOTOBJ}/efi/libefi32/libefi.a
+
+DPADD=	${LDR_INTERP32} ${LIBEFI32} ${LIBSA32} ${LDSCRIPT}
+LDADD=	${LDR_INTERP32} ${LIBEFI32} ${LIBSA32}
+.endif
```
bsdimp, line 40: This seems alll wrong, though. It seems like we should use stnad/efi/loader/arch/i386/Makefile.inc when building the ia32 version.

stand/efi/loader/arch/amd64/Makefile.inc
jrtc27: No I mean the SUBDIR bit; that's where loader_lua etc get recursed into from, so should also be where loader_lua-ia32 etc are recursed into?

VexedUXR: > Why isn't this up in stand/efi/Makefile? That would be the normal way to do this

I thought this would be more convenient since we could just use `LOADER_INTERP`, but now I realize that `SUBDIR.yes.yes` is a thing so we can do `SUBDIR.${MK_FORTH}.${MK_LOADER_IA32}`. Will update.

> Oh that. Yes. In fact, I think we likely just want loader_ia32.efi and not all the variations since it's a special thing.

Hm, so would that only build the lua loader?

stand/efi/libefi32/Makefile
VexedUXR: Isn't a blank line left at the top of the other Makefiles? (At least, the ones that don't start with comments)
If that's not an enforced style, then yeah I'll remove it.

stand/efi/loader/arch/amd64/Makefile.inc
VexedUXR: Yes that would make more sense now that you mention it.
When I first wrote this patch, the arch/i386 dir was being used for the half-completed 32-bit loader.

VexedUXR:
Moved everything to `arch/i386`.
Also made a single `loader_ia32` instead of all the variations (It just uses `LOADER_DEFAULT_INTERP`). Let me know if I misunderstood what you meant though.

I'm not requesting a review this time since I'm not sure it it would mark [this](https://github.com/freebsd/freebsd-src/pull/1098#discussion_r1624870992) as resolved.

VexedUXR:
Some small cleanup. [Diff](https://github.com/freebsd/freebsd-src/compare/e3c83d812d2cbc966a85fb8a48b25eb957db203a..2fd75f4de6bb94b81d9c7bf6f090691d4f389762)
Usually I'd keep these local until the next set of changes, but I figured these might get mentioned in a review.

LadySerenaKitty:
I am actively trying to install FreeBSD to such a machine, a Dell Venue 10 Pro (5055).  It has a rather awful firmware that can only run 32-bit EFI binaries but it has a 64-bit processor, and is an EFI Class 3 machine.
image

I had to take a video at 240fps to capture this.  After applying this MR, I manually added the `loader_ia32.efi` to my USB install stick.  I tried it with `FreeBSD-14.0-RELEASE-amd64-memstick.img` and `FreeBSD-14.1-RELEASE-amd64-memstick.img`.  It fails and then the machine just goes back to booting some other random OS.

stand/efi/libefi32/Makefile
VexedUXR: Ah, just saw e9ac41698b2f322d55ccf9da50a3596edb2c1800.
Guess I confused that for a style.

bsdimp:
I think this is as ready as it's going to get, and plan to land it absent objections...

tools/boot/install-boot.sh
```diff
@@ -59,7 +59,12 @@ make_esp_file() {
     stagedir=$(mktemp -d /tmp/stand-test.XXXXXX)
     mkdir -p "${stagedir}/EFI/BOOT"
     efibootname=$(get_uefi_bootname)
+
     cp "${loader}" "${stagedir}/EFI/BOOT/${efibootname}.efi"
+    loader="${loader%.efi}_ia32.efi"
```
jrtc27, line 64: I really don't like this magic... can we not just make it optionally take extra files as arguments and update the release scripts?

tools/boot/install-boot.sh
```diff
@@ -95,6 +100,27 @@ make_esp_device() {
 
     efibootname=$(get_uefi_bootname)
     kbfree=$(df -k "${mntpt}" | tail -1 | cut -w -f 4)
+
+    # For amd64 there are two efi loaders, one for 64-bit uefi and the other for 32-bit.
+    # If we can figure out which one we need, we copy it to /EFI/freebsd/loader.efi
+    # If we can't, then install the 64-bit one as loader.efi and the 32-bit one as loader_ia32.efi
+    if ([ -n "${updatesystem}" ] && [ `sysctl -n machdep.efi_arch` == i386 ]) || \
+       ([ -f "${mntpt}/EFI/freebsd/loader.efi" ] && (file "${mntpt}/EFI/freebsd/loader.efi" | grep -q '80386')) ; then
+        file="${file%.efi}_ia32.efi"
```
jrtc27, line 109: As above

usr.sbin/bsdinstall/scripts/bootconfig
```diff
@@ -117,21 +117,29 @@ if [ -n "$(awk '{if ($2=="/boot/efi") printf("%s\n",$1);}' $PATH_FSTAB)" ]; then
 
 	f_dprintf "Installing loader.efi onto ESP"
 	mkdir -p "${mntpt}/${FREEBSD_BOOTDIR}" "${mntpt}/${BOOTDIR}"
-	cp "$BSDINSTALL_CHROOT/boot/loader.efi" "${mntpt}/${FREEBSD_BOOTNAME}"
+
+	if [ -f "$BSDINSTALL_CHROOT/boot/loader_ia32.efi" ] && [ `sysctl -n machdep.efi_arch` == i386 ]; then
```
jrtc27, line 121: Why not hoist this up so you can just set the variables to the right thing in the first place?

stand/efi/loader/main.c
```diff
@@ -920,7 +921,7 @@ acpi_detect(void)
 		if ((rsdp = efi_get_table(&acpi)) == NULL)
 			return;
 
-	sprintf(buf, "0x%016llx", (unsigned long long)rsdp);
+	sprintf(buf, "0x%016"PRIxPTR, (uintptr_t)rsdp);
```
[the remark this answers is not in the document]
VexedUXR: We compile with -Werror, so casting a pointer to an integer of a different size should cause the build to fail.
iirc I only noticed this because the GCC build failed. Clang seemingly ignores -Wpointer-to-int-cast when the size of the integer is bigger than that of the pointer's, which is probably why it doesn't fail on 32-bit arm

stand/efi/loader/framebuffer.c
```diff
@@ -651,16 +661,16 @@ efi_find_framebuffer(teken_gfx_t *gfx_state)
...
 	if (gfx_state->tg_shadow_fb != NULL)
-		BS->FreePages((EFI_PHYSICAL_ADDRESS)gfx_state->tg_shadow_fb,
+		BS->FreePages((uintptr_t)gfx_state->tg_shadow_fb,
 		    gfx_state->tg_shadow_sz);
 	gfx_state->tg_shadow_sz =
 	    EFI_SIZE_TO_PAGES(efifb.fb_height * efifb.fb_width *
 	    sizeof(EFI_GRAPHICS_OUTPUT_BLT_PIXEL));
-	status = BS->AllocatePages(AllocateMaxAddress, EfiLoaderData,
-	    gfx_state->tg_shadow_sz,
-	    (EFI_PHYSICAL_ADDRESS *)&gfx_state->tg_shadow_fb);
```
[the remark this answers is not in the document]
VexedUXR: Yea

stand/efi/loader/framebuffer.c
```diff
@@ -149,7 +149,16 @@ efifb_from_gop(struct efi_fb *efifb, EFI_GRAPHICS_OUTPUT_PROTOCOL_MODE *mode,
 {
 	int result;
 
-	efifb->fb_addr = mode->FrameBufferBase;
+	efifb->fb_addr =
+#ifdef __i386__
+	/*
+	 * Some weird implementations give a bogus address
```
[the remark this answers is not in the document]
VexedUXR: It was reported above: https://github.com/freebsd/freebsd-src/pull/1098#issuecomment-2106035489
The machine is an Asus EEEPC 1025C.
I probably shouldn't have said implementation**s**, would specifying the machine here be better?

stand/efi/loader/efi_main.c
```diff
@@ -87,6 +87,17 @@ efi_main(EFI_HANDLE image_handle, EFI_SYSTEM_TABLE *system_table)
 	BS = ST->BootServices;
 	RS = ST->RuntimeServices;
 
+#ifdef __i386__
+	/* Check for long mode */
+	asm("cpuid" : "=d"(status) : "a"(0x80000001) : "%ebx", "%ecx");
```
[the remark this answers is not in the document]
VexedUXR: Hmm, what about some setup function that then calls efi_main. We could have it in loader/arch/i386

tools/boot/install-boot.sh
VexedUXR: Ah, I didn't think of that. That would be much better, I think.

tools/boot/install-boot.sh
VexedUXR: Maybe we should make it so the 64-bit one is always loader.efi and the 32-bit one is always loader_ia32.efi

Looking at this now, I think occasionally copying the 32-bit one to loader.efi would just cause confusion.

VexedUXR:
I moved the checking for long mode code into it's own file at `loader/i386/setup.c`. I was originally thinking about putting this code in `loader/i386/elf64_freebsd.c` instead of creating a new file for a single function, but I feel like it doesn't quite fit there. The setup function is called by _start.

For the scripts, make_esp_file can now take extra arguments in pairs. e.g `make_esp_file ... loader1.efi bootx64 loader2.efi bootia32`. The second argument is optional so we can still simply `make_esp_file ... loader.efi`.
make_esp_device takes an optional third argument that specifies efibootname. This is currently unused since, as far as I can tell, make_esp_device isn't used outside of install-boot. The extra argument allows us to do something like this in the future:
```
make_esp_device loader.efi bootx64
make_esp_device loader_ia32.efi bootia32
```

bsdimp:
OK. I'm finally landing this....  Likely should MFC to 14 too, but not 13, imho (though maybe?).
This does increase the technical debt a bit, which makes me nervous.

VexedUXR:
Thanks!

> This does increase the technical debt a bit, which makes me nervous.

I wouldn't mind working on some cleanup PRs when I find the time. First thing that comes to mind are those externs you mentioned a while back in most of the `elf*_freebsd.c` files.

This PR has come a long way. Thanks for the reviews everyone :)

bsdimp:
> Thanks!
> 
> > This does increase the technical debt a bit, which makes me nervous.
> 
> I wouldn't mind working on some cleanup PRs when I find the time. First thing that comes to mind are those externs you mentioned a while back in most of the `elf*_freebsd.c` files.

Please do. That's the area I'd thought could use the work. I didn't want to force this to be perfect to get it in. I welcome followups.

> This PR has come a long way. Thanks for the reviews everyone :)

Yes. Thanks!

PR #1099, opened 2024-02-03
libipsec: make const-correct

llfw:
- add const to the appropriate places in the libipsec public API and the relevant internal functions needed to support that.

- replace caddr_t with c_caddr_t in ipsec_dump_policy()

- update the ipsec_dump_policy manpage to use c_caddr_t (this manpage was already wrong as it had "char *" instead of caddr_t previously).

While here, update pfkeyv2.h to not cast away const in the PFKEY_*() macros.

This should not cause any ABI changes as the actual types have not changed.

---

this came up during my work on the traceroute merge (and will make that easier) but should fixed regardless of that.

### DIFF
--- a/lib/libipsec/ipsec_dump_policy.c
+++ b/lib/libipsec/ipsec_dump_policy.c
@@ -67,9 +67,9 @@ static char *set_address(char *, size_t, struct sockaddr *);
  * When delimiter == NULL, alternatively ' '(space) is applied.
  */
 char *
-ipsec_dump_policy(caddr_t policy, char *delimiter)
+ipsec_dump_policy(c_caddr_t policy, const char *delimiter)
 {
-	struct sadb_x_policy *xpl = (struct sadb_x_policy *)policy;
+	const struct sadb_x_policy *xpl = (const struct sadb_x_policy *)policy;
 	struct sadb_x_ipsecrequest *xisr;
 	size_t off, buflen;
 	char *buf;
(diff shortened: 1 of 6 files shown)


llfw:
can this be merged or does it need more review or changes first?

lib/libipsec/ipsec_dump_policy.c
```diff
@@ -67,9 +67,9 @@ static char *set_address(char *, size_t, struct sockaddr *);
  * When delimiter == NULL, alternatively ' '(space) is applied.
  */
 char *
-ipsec_dump_policy(caddr_t policy, char *delimiter)
+ipsec_dump_policy(c_caddr_t policy, char const *delimiter)
 {
-	struct sadb_x_policy *xpl = (struct sadb_x_policy *)policy;
+	struct sadb_x_policy const *xpl = (struct sadb_x_policy const *)policy;
```
bsdimp, line 72: why isn't this `const struct sadb_x_policy *xpl` here and elsewhere.
The project prefers `const type *foo;` rather than `type const *foo;`

llfw:
ok, i'll change that to put const in the other place.  (i mostly write C++ where this ordering is the more common one, so i do it that way out of habit.)

llfw:
replaced 'T const' with 'const T', also rebased on current main (no conflicts), tested by buildworld amd64.

bsdimp (approved):
These look good now.

bsdimp:
merged

PR #1100, opened 2024-02-03
move traceroute from contrib to base

llfw:
in preparation for the traceroute/traceroute6 merge, move traceroute from contrib into usr.sbin.  traceroute hasn't had a vendor update since 2002, while it's had many FreeBSD-specific commits since then, so it's basically not contrib anymore.

this introduces new files in base under the 4-clause BSD license.  i asked core@ about this and @emaste ok'd it:

> The 4-clause prohibition is on introducing new files, and if all we're
> really doing here is acknowledging that there is no viable upstream
> any longer and moving the source to reflect that then moving those
> files while keeping the existing licenses is fine.
>
> If an easy opportunity to replace or rewrite portions presents itself
> then this work could provide a good opportunity to do so -- but it's
> not required. One thing to be sure of though is that we don't end up
> increasing the scope of the 4-clause license during this work - e.g.,
> when moving functions between files or so.

since i'm creating source churn anyway, i've made a few cleanups to the source code at the same time, but this is mostly fairly minor nits; i was tempted to refactor the entire thing, but that can wait until later.


llfw:
note that style errors were already present in the original source.  i can fix these, but i'd rather get this merged and do any further cleanups during the merge.

bsdimp:
I'm thinking that maybe we move as one unit. Then see what clang-format does. Then other code cleanup? Comments?

llfw:
that's fine with me -- i've already done a lot more cleanup in the `traceroute-merge` branch so we could just drop the cleanups commit from here if that makes it easier to merge.  (i don't think the cleanups fix any actual bugs.)

bsdimp:
I landed this.
3f0b108924428ff91f2640de5f01df0d0bf45253

PR #1101, opened 2024-02-04
glabel: Remove support for old reiserfs

ricardobranco777:
The filesystem version supported in glabel is old (v2) and we don't even have a FUSE package in ports to use it at all.


bsdimp:
Yea, this is as close to no brainer yes as it gets... landed. Sorry for the delay and thank you for your contribution.

PR #1102, opened 2024-02-04
.profile: Don't bother checking for /home symlink

collinfunk:
Since FreeBSD 14.0, user directories are created directly under /home. This check should no longer be needed.


igalic:
this needs an MFC to 14

igalic (changes requested):
this (commit message) needs an MFC to 14 then

collinfunk:
@igalic My bad, amended and forced. Let me know if I missed anything.

emaste:
I'm not sure this needs a MFC to 14 in fact; the only impact if /home is not a symlink is that there's a minuscule time cost for the `[ "$PWD" != "$HOME" ]` test. I would just leave it in place in 14 in case the user's .profile ends up getting used on a system with symlink /home.

emaste:
Also,
This reverts commit 4cea05a273c875b5d5d4c41bfa6f2f0a60fa4a66.

collinfunk:
Should I just remove the MFC and add a note referencing that commit then? Also, the merge from freebsd:main was a mistake. I thought I was on a different branch. :(

emaste:
committed in fd1066bed67f642ed01f5dd62b7b6cfb0a45fd4c

PR #1103, opened 2024-02-05
axgbe: Various stability improvements

swhite2:
Note: this is a continuation of https://reviews.freebsd.org/D40725. First half containing only link and module compatibility changes went in with https://github.com/freebsd/freebsd-src/commit/445bed5c4d859575951361e432024396b938f863

- Hook in RSS glue.
- Default to "off" for the split header feature to ensure netmap compatibility.
- Change the PCS indirection register values based on hardware type (ported from Linux).
- Move tunable settings to sysctl_init() and set the defaults there. Ensure it's called at the right time by moving it back.
- Reset PHY RX data path when mailbox command times out (Ported from Linux).
- Check if VLAN HW tagging is enabled before assuming a VLAN tag is present in a descriptor.
- Disable the hardware filter since multicast traffic is dropped in promisc mode.
- Remove unnecessary return statement.
- Missing sfp_get_mux, causing a race between ports to read SFP(+) sideband signals.
- Validate and fix incorrectly initialized polarity/configuration registers.
- Remove unnecessary SFP reset.
- axgbe_isc_rxd_pkt_get has no error state, remove unnecessary big packet check.
- Enable RSF to prevent zero-length packets while in Netmap mode.
- DMA cache coherency update (ported from Linux).

### DIFF
--- a/sys/dev/axgbe/if_axgbe_pci.c
+++ b/sys/dev/axgbe/if_axgbe_pci.c
@@ -56,6 +56,14 @@
 #include "ifdi_if.h"
 #include "opt_inet.h"
 #include "opt_inet6.h"
+#include "opt_rss.h"
+
+#ifdef RSS
+
+#include <net/rss_config.h>
+#include <netinet/in_rss.h>
+
+#endif
 
 MALLOC_DEFINE(M_AXGBE, "axgbe", "axgbe data");
 
@@ -103,6 +111,7 @@ static bool axgbe_if_needs_restart(if_ctx_t, enum iflib_restart_event);
 #endif
 static void axgbe_set_counts(if_ctx_t);
 static void axgbe_init_iflib_softc_ctx(struct axgbe_if_softc *);
+static void axgbe_initialize_rss_mapping(struct xgbe_prv_data *);
 
 /* MII interface registered functions */
 static int axgbe_miibus_readreg(device_t, int, int);
@@ -277,11 +286,11 @@ axgbe_register(device_t dev)
 		 * No tunable found, generate one with default values
 		 * Note: only a reboot will reveal the new kenv
 		 */
-		error = kern_setenv("dev.ax.sph_enable", "1");
+		error = kern_setenv("dev.ax.sph_enable", "0");
 		if (error) {
 			printf("Error setting tunable, using default driver values\n");
 		}
-		axgbe_sph_enable = 1;
+		axgbe_sph_enable = 0;
 	}
 
 	if (!axgbe_sph_enable) {
@@ -389,6 +398,7 @@ axgbe_if_attach_pre(if_ctx_t ctx)
 	if_softc_ctx_t		scctx;
 	if_shared_ctx_t		sctx;
 	device_t		dev;
+	device_t		rdev;
 	unsigned int		ma_lo, ma_hi;
 	unsigned int		reg;
 	int			ret;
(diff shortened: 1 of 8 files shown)


tuexen:
Did you test this on a system with armv8, like an Overdrive 3000?

fichtner:
The patches were designed and implemented for operational problems on amd64 systems… so no.

sys/dev/axgbe/if_axgbe_pci.c
```diff
@@ -56,6 +56,14 @@
 #include "ifdi_if.h"
 #include "opt_inet.h"
 #include "opt_inet6.h"
+#include "opt_rss.h"
+
+#ifdef RSS
+
```
bsdimp, line 62: We don't normally have blank lines like this after the ifdef line, before the extra includes (or before the final endif.

sys/dev/axgbe/xgbe-dev.c
```diff
@@ -1451,7 +1451,8 @@ xgbe_dev_read(struct xgbe_channel *channel)
 
 	if (!err || !etlt) {
 		/* No error if err is 0 or etlt is 0 */
-		if (etlt == 0x09) {
+		if (etlt == 0x09 &&
+			(if_getcapenable(pdata->netdev) & IFCAP_VLAN_HWTAGGING)) {
```
bsdimp, line 1455: this indent should be 4 spaces.

sys/dev/axgbe/xgbe-phy-v2.c
```diff
@@ -2102,6 +2243,29 @@ xgbe_phy_pll_ctrl(struct xgbe_prv_data *pdata, bool enable)
...
+static void
+xgbe_phy_rx_reset(struct xgbe_prv_data *pdata)
+{
+	int reg;
+
+	reg = XMDIO_READ_BITS(pdata, MDIO_MMD_PCS, MDIO_PCS_DIGITAL_STAT,
+			     XGBE_PCS_PSEQ_STATE_MASK);
+
+	if (reg == XGBE_PCS_PSEQ_STATE_POWER_GOOD) {
+		/* Mailbox command timed out, reset of RX block is required.
```
bsdimp, line 2255: The /* should be on its own line.

bsdimp:
A few nits... this looks decent...

bsdimp:
I should be clear: if this is tested working on amd64, then fixing this few nits will suffice. new architecture support, while desirable, shouldn't gate it (though a known regression would, so I'll be doing a universe kernel build before landing). All the cross build tests, though, suggest it will be fine.

fichtner:
FWIW, this code has been running fine for at least a year now with customers on OPNsense with the relevant hardware on amd64. Part of the batched changes even for a couple of years ever since the driver was extended to 10G.

bsdimp:
> FWIW, this code has been running fine for at least a year now with customers on OPNsense with the relevant hardware on amd64. Part of the batched changes even for a couple of years ever since the driver was extended to 10G.

Excellent report. The only magic that needs an explanation is the root complex check...

swhite2:
@bsdimp https://github.com/freebsd/freebsd-src/pull/1103/commits/7a4ef9b75c6bcea892ba99ea1e81d45c4f3ef233 should be it. Thank you for taking the time to look at this.

bsdimp:
Merged. I went ahead and folded the second commit you made. Generally, we squash style commits that fix errors in the other commits in the series. Thanks for updating this driver. Sorry it's taken me so long to get to closure.

PR #1104, opened 2024-02-05
merge traceroute6 into traceroute

llfw:
this PR depends on #1099 (libipsec const-correctness) and #1100 (move traceroute into base).  depending on what exactly we decide to commit in #1100, it may end up needing a rebase.

it is not ready for merging (a significant number of features don't work yet), but i wanted to submit it as a draft PR to validate the basic approach, which is explained in the commit message "traceroute: support IPv6".

a browsable version of the branch is available here: https://github.com/llfw/freebsd-src/tree/traceroute-merge/usr.sbin/traceroute

(if you'd prefer i not do this, do let me know.)


bsdimp:
I think this is good to go once #1099 is fixed.

bsdimp:
> > I think this is good to go once #1099 is fixed.
> 
> don't merge this yet since it doesn't work, quite a few things are still broken in this version (that's why i left it as a draft).
> 
> i've been meaning to get this finished but i got side-tracked by some ARM64 and INET stuff...

OK. those more recent ones are up next for me...

bsdimp:
This is a good idea, but pull request drafts aren't super useful since there's no testing that this can trigger. Please reopen when you have something more concrete.

PR #1105, opened 2024-02-06
msgs: Use C99 bool instead of defining our own

collinfunk:
Use stdbool.h definitions instead of defining non-standard ones.


emaste:
Committed in 912b52a62c72ca9c96a396148a6b30ff9c0917e7

PR #1107, opened 2024-02-07
ctags: Use C99 bool instead of defining our own

collinfunk:
Use stdbool.h definitions instead of defining non-standard ones.

### DIFF
--- a/usr.bin/ctags/fortran.c
+++ b/usr.bin/ctags/fortran.c
@@ -41,14 +41,14 @@ static void takeprec(void);
 
 char *lbp;				/* line buffer pointer */
 
-int
+bool
 PF_funcs(void)
 {
 	bool	pfcnt;			/* pascal/fortran functions found */
 	char	*cp;
 	char	tok[MAXTOKEN];
 
-	for (pfcnt = NO;;) {
+	for (pfcnt = false;;) {
 		lineftell = ftell(inf);
 		if (!fgets(lbuf, sizeof(lbuf), inf))
 			return (pfcnt);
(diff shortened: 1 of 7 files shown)


collinfunk:
Two of the style(9) failures are from the weird GETC macro. The others are because I did not add previously missing parentheses around return values. Should I amend or leave it for a different patch?

bsdimp:
I'd match the existing style, so don't change the returns.
The GETC stuff is a weird macro, and it's just a false positive, so it's good.

bsdimp:
I think this is ready. I'll try to land it later today.

usr.bin/ctags/fortran.c
```diff
@@ -48,7 +48,7 @@ PF_funcs(void)
 	char	*cp;
 	char	tok[MAXTOKEN];
 
-	for (pfcnt = NO;;) {
+	for (pfcnt = false;;) {
```
markpeek, line 51: Should `int PF_funcs(void)` be changed to `bool PF_funcs(void)` due to these changes?

bsdimp: I think so because we're return pfcnt 3 lines below.

collinfunk: Thanks for checking, you are right. While fixing it I also noticed that "skip_key" in C.c and ctags.h should return bool. I'll amend those changes to the commit.

collinfunk:
That should be all of them hopefully...

markjdb:
Merged.

PR #1108, opened 2024-02-07
Lagg.4 - removed .Xr reference to /etc/rc.conf(5)

chrisdavidson:
lagg.4: manual page incorrectly .Xr to /etc/rc.conf(5) removed link and kept test of /etc/rc.conf

### DIFF
--- a/share/man/man4/lagg.4
+++ b/share/man/man4/lagg.4
@@ -214,7 +214,7 @@ The following example shows how to create an infiniband failover interface.
 .Ed
 .Pp
 Configure two ethernets for failover with static IP in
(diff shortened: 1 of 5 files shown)


bsdimp:
https://man.freebsd.org/cgi/man.cgi?query=rc.conf&sektion=5&manpath=FreeBSD+14.0-RELEASE

So either .Xr rc.conf 5 or .Pa /etc/rc.conf

chrisdavidson:
i am going to re-evaluate my git workflow process as there are some holes in it causing significant pain for the reviewers involved. Once I have resolved these I will resubmit these pull requests in a different fashion.

PR #1110, opened 2024-02-08
fusefs: only test for incoherency if FN_SIZECHANGE is set

etsal:
FUSE emits spurious incoherency warnings in writethrough mode. The warnings are triggered by setattr calls generated by vnode truncation turning the cached va_size vattr stale, causing comparisons with the fresh version provided by the server to fail. Only validate the vnode's va_size vattr if the FN_SIZECHANGE flag is set.

### DIFF
--- a/sys/fs/fuse/fuse_internal.c
+++ b/sys/fs/fuse/fuse_internal.c
@@ -269,10 +269,10 @@ fuse_internal_cache_attrs(struct vnode *vp, struct fuse_attr *attr,
 
 	if (vnode_isreg(vp) &&
 	    fvdat->cached_attrs.va_size != VNOVAL &&
+	    fvdat->flag & FN_SIZECHANGE &&
 	    attr->size != fvdat->cached_attrs.va_size)
 	{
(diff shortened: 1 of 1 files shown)


emaste:
CC @asomers

asomers (approved):
This LGTM.  We were correctly checking that flag in writeback mode, but not in writethrough mode.  Thanks for fixing it.

asomers:
BTW @etsal this looks like your first ever commit to src.  Are you being paid in any way for the work you do on FreeBSD?  If so, we should tag your commit message with "Sponsored by: Somebody".

etsal:
I am part of the RCS Lab at the University of Waterloo, so we could add a tag  "This is a part of the research work at RCSLab, University of Waterloo.", as previously done in 5e29272bf57e24a16f3a17f037f9a5e7cf25ed90 for a contribution made by a labmate of mine.

asomers:
Ok, I will so tag it.

asomers:
Fixed by 8758bf0a

PR #1111, opened 2024-02-08
cxgbe: Free memory if alloc_wrqe fails and return error

AtariDreams:
There does not seem to be any sense in returning 0 like nothing if an alloc fails. Free memory and return an error instead.

### DIFF
--- a/sys/dev/cxgbe/iw_cxgbe/cq.c
+++ b/sys/dev/cxgbe/iw_cxgbe/cq.c
@@ -88,6 +88,7 @@ static int destroy_cq(struct c4iw_rdev *rdev, struct t4_cq *cq,
 			  cq->memsize, cq->queue,
 			  dma_unmap_addr(cq, mapping));
 	c4iw_put_cqid(rdev, cq->cqid, uctx);
+	free_wrqe(wr);
 	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


sys/dev/cxgbe/iw_cxgbe/cq.c
```diff
@@ -132,9 +132,12 @@ create_cq(struct c4iw_rdev *rdev, struct t4_cq *cq,
 	wr_len = sizeof *res_wr + sizeof *res;
 
 	wr = alloc_wrqe(wr_len, &sc->sge.ctrlq[0]);
-	if (wr == NULL)
-		return (0);
-        res_wr = wrtod(wr);
+	if (!wr) {
```
markpeek, line 135: Not an expert on this code but it seems like there should be calls to `free_wrqe()` (and might need an `err5` label). Not seeing this in either `create_cq()` or `destroy_cq()`.

igalic (changes requested):
just commenting on the commit message:

the prefix here should be

> cxgbe: …

cq isn't really a unit that makes sense to anyone

sys/dev/cxgbe/iw_cxgbe/cq.c
AtariDreams: I'm aware. However I want to investigate and make sure this memory is being leaked because a double free is arguably worse.

sys/dev/cxgbe/iw_cxgbe/cq.c
```diff
@@ -132,9 +133,12 @@ create_cq(struct c4iw_rdev *rdev, struct t4_cq *cq,
...
```
bsdimp, line 136: This is a style(9) issue. The old check against NULL was the proper style.

bsdimp:
I'm thinking that Chelsio is almost an upstream and this should go in via np@ committing things rather than committing it here. 
@bsdjhb can you look and figure out if this bit of the code should be changed in the FreeBSD tree, or go in via chelsio first?

PR #1112, opened 2024-02-08
drm: Restore from save_set.fb instead of assigning to old_fb

AtariDreams:
We should also recover set->crtc->x  and set->crtc->y when we fail in fb_changed.


bsdimp:
Why?

bsdimp:
drm2 is used only on armv7, and is too niche to get reviewed here.
So I'm closing this. please post this to arm@ and see if you can get some interest there. If so, then I'll reopen.

PR #1113, opened 2024-02-11
which: Use size_t instead of ssize_t for pathlen

collinfunk:
The "pathlen" variable is the return value of strlen(3) and is then passed as an argument to malloc(3) and memcpy(3). The size_t type matches the prototype for these functions. The size_t type is unsigned so it can fit larger $PATH values than ssize_t. However, In practice ssize_t should be larger enough so this change is just for clarity.

### DIFF
--- a/usr.bin/which/which.c
+++ b/usr.bin/which/which.c
@@ -45,7 +45,7 @@ int
 main(int argc, char **argv)
 {
 	char *p, *path;
-	ssize_t pathlen;
+	size_t pathlen;
 	int opt, status;
 
 	status = EXIT_SUCCESS;


igalic (approved):
this seems sensible.
I'm surprised it's not been throwing a warning

collinfunk:
> this seems sensible. I'm surprised it's not been throwing a warning

I think it might have thrown -Wsign-compare if it started to be used for comparison. I assume that's probably turned off in most places because it would cause meaningless (mostly) spam with old code. I just noticed it while looking at something else and it annoyed me. :)

gmshake (approved):
LGTM.

markjdb:
Merged in 8268a31bcceb9ebe32d380cab792c89c5d897d15, thank you.

PR #1114, opened 2024-02-11
Updates to 3 manual pages: lagg.4, xlocale.3, buf.9

chrisdavidson:
Update manual page references and macros to align to mandoc syntax

xlocale.3: Comment out reference to atof_l(3), atoi_l(3), atol_l(3), atoll_l(3)
  These manual page references do not exist.
lagg.4: Change the reference for /etc/rc.conf from a reference link
  .Xr -> .Pa based on the context within the manual page it is used.
buf.9: Remove .Xr entries from the file
  The buf.9 manual page contains a commented out .Xr reference.
  The <filmmein> 9 entry is a placeholder and has been removed for
  clarity.

### DIFF
--- a/lib/libc/locale/xlocale.3
+++ b/lib/libc/locale/xlocale.3
@@ -166,10 +166,10 @@ along with the headers that expose them, is provided here:
 .Xr vsprintf_l 3 ,
 .Xr vsscanf_l 3
 .It In stdlib.h
-.Xr atof_l 3 ,
-.Xr atoi_l 3 ,
-.Xr atol_l 3 ,
-.Xr atoll_l 3 ,
+.\".Xr atof_l 3 ,
+.\".Xr atoi_l 3 ,
+.\".Xr atol_l 3 ,
+.\".Xr atoll_l 3 ,
 .Xr mblen_l 3 ,
 .Xr mbstowcs_l 3 ,
 .Xr mbtowc_l 3 ,
--- a/share/man/man4/lagg.4
+++ b/share/man/man4/lagg.4
@@ -214,7 +214,7 @@ The following example shows how to create an infiniband failover interface.
 .Ed
 .Pp
 Configure two ethernets for failover with static IP in
-.Xr /etc/rc.conf 5 :
+.Pa /etc/rc.conf :
 .Bd -literal -offset indent
 cloned_interfaces="lagg0"
 ifconfig_lagg0="laggproto failover laggport bge0 laggport bge1 \e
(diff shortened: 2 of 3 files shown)


rilysh (approved):
I think the man pages for `ato*` functions are needed to be added(?), but everything else looks good!

collinfunk:
I think that the links to atoi_l and friends should remain in xlocale(3). Some of the other ones on that page are also missing (e.g. strtol_l, wcstol_l). As long as the symbols still exist they should probably still be referenced there even if the man pages are incomplete.

chrisdavidson:
thank you both for the comments and the approval.

The <function name>_l situation, in the xlocale.3 (2 second paragraph in description) references there are _l versions of these C functions.. which then brings us to stdlib.h. (which this section lists those _l functions)

I was thinking, but was previously rejected, to just have the _l version of the functions points to the non _l version manual pages. 

This was based on the function calls look to be the same but the way the _l functions operate are different.

Which gets us to: Should we just have all _l functions point to the non _l manual pages or just create distinct manual pages for the _l functions.

my gut is telling me the former option (linked to non _l versions) but i will default to the experts.

mhorne (commented):
My suggestion for the `_l`-prefixed links is to retain them, but comment them out and add a note that the pages are missing/nonexistent. Unless you want to work on adding the missing pages, but I am not asking for that :)

The other changes look ready to me, but the commit messages themselves need to be fixed. They need a proper "subject line", not just a paragraph of text. See the guidelines: https://docs.freebsd.org/en/articles/committers-guide/#commit-log-message

With the commit message text fixed, I can merge those changes.

Edit: an example commit message for `share/man/man4/tcp_rack.4`:
```
commit 7701b993554321f305ef662a26238f375c4d2dba
Author: Gordon Bergling <gbe@FreeBSD.org>
Date:   Tue Mar 5 03:03:50 2024 +0100

    tcp_rack.4: Remove a mentioned paper
    
    The paper "T-RACKs: A Faster Recovery Mechanism for
    TCP in Data Center Networks" has nothing to do with
    our TCP RACK implementation, so remove it.
    
    Reported by:    tuexen
    MFC after:      3 days
```

chrisdavidson:
Thank you @mhorne for your review and recommendations. I will review the suggested guidance document and refine accordingly.

It would be best to amend these commits and NOT make a new one, correct? (still learning)

mhorne:
> Thank you @mhorne for your review and recommendations. I will review the suggested guidance document and refine accordingly.
> 
> It would be best to amend these commits and NOT make a new one, correct? (still learning)

Yes, amend them. After that, you will need to force push the branch.

chrisdavidson:
hi @mhorne i think this one is good to go now after all the suggested updates.

bsdimp:
Added go/nogo tag for @mhorne to commit or not...

mhorne:
Done. This can be closed.

bsdimp:
merged

PR #1116, opened 2024-02-13
cp: Remove useless ssize_t cast

collinfunk:
Both wcount and wresid are ssize_t so this cast is not needed. Just remove it so the code is easier to read.


markjdb:
Landed in 8b864151856fbba97fed16b1200248cf9f6f3ccf

PR #1117, opened 2024-02-13
Perfer sys/cdefs.h __unreachable over the builtin

collinfunk:
The __builtin_unreachable macro provided by Clang and GCC is a hint to the compiler used for optimization. The programs work fine even if the compiler doesn't support it. The sys/cdefs.h has had __unreachable for 9 years (commit 732b31de5d9244bd1cc98192e09ee1881e9f55e9). It expands to the builtin if it is available. In the rare case that it is unsupported it expands to a null statement so compilation does not fail.

### DIFF
--- a/bin/nproc/nproc.c
+++ b/bin/nproc/nproc.c
@@ -94,7 +94,7 @@ main(int argc, char *argv[])
 			break;
 		case OPT_VERSION:
 			version();
-			__builtin_unreachable();
+			__unreachable();
 		case OPT_HELP:
 			help();
 			exit(EXIT_SUCCESS);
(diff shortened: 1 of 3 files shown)


collinfunk:
@bsdimp @emaste ISO C 23 N2826 adds the function-like macro unreachable() to stddef.h. Can I submit a separate patch proposing that change? I'm not sure if it is something left up to the GCC/Clang people to include with their headers though...

igalic (changes requested):
this needs one commit per subsystem.
you're touching completely unrelated parts of the system here

collinfunk:
> this needs one commit per subsystem.

Oops, my bad. Split into three commits, one for each program.

igalic (approved):
👍🏻

bsdimp:
Landed this... go ahead and start to implement the c23 defined macro if you can do it w/o problems on older C compilers.

bsdimp:
OK, automation still bad, but I caught it this time.

PR #1118, opened 2024-02-13
linux: ignore setsockopt(IPV6_RECVERR)

llfw:
Under Linux, the socket options IP_RECVERR and IPV6_RECVERR are used to receive socket errors via a dedicated 'error queue' which can be retrieved via recvmsg().  FreeBSD does not support this functionality.

For IPv4, the sysctl compat.linux.ignore_ip_recverr can be set to 1 to silently ignore attempts to set IP_RECVERR and return success to the application, which is wrong, but is required for (among other things) a functional DNS client in recent versions of glibc.

Add support for ignoring IPV6_RECVERR, controlled by the same sysctl. This fixes DNS in Linux when using IPv6 resolvers.

### DIFF
--- a/sys/compat/linux/linux_socket.c
+++ b/sys/compat/linux/linux_socket.c
@@ -501,6 +501,11 @@ linux_to_bsd_ip6_sockopt(int opt)
 		    "unsupported IPv6 socket option IPV6_RECVFRAGSIZE (%d)",
 		    opt);
 		return (-2);
+	case LINUX_IPV6_RECVERR:
+		LINUX_RATELIMIT_MSG_OPT1(
+		    "unsupported IPv6 socket option IPV6_RECVERR (%d), you can not get extended reliability info in linux programs",
+		    opt);
+		return (-2);
 
 	/* unknown sockopts */
 	default:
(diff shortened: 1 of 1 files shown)


sys/compat/linux/linux_socket.c
```diff
@@ -2111,6 +2116,11 @@ linux_setsockopt(struct thread *td, struct linux_setsockopt_args *args)
 		name = linux_to_bsd_ip_sockopt(args->optname);
 		break;
 	case IPPROTO_IPV6:
+		if (args->optname == LINUX_IPV6_RECVERR &&
+		    linux_ignore_ip_recverr) {
+			/* XXX: as above, for the IPv6 case */
```
jlduran, line 2121: I would have used the same `XXX` comment `This is a hack to unbreak DNS resolution with glibc 2.30 and above.` to keep it symmetrical.

jlduran (approved):

llfw:
i should add: without this change, Linux emulation doesn't work for me at all, because my network only provides IPv6 nameservers, and glibc DNS resolution over IPv6 transport doesn't work without this change.

jlduran:
For my future reference: commit da6d8ae6d83ac199098ec97aa88f57a36c4acd6b.

llfw:
> Please incorporate @jlduran's feedback and then we'll land this

done!

llfw:
i forgot to note this in the commit message but this one should probably be MFC'd.

jlduran (approved):

bsdimp:
merged

PR #1120, opened 2024-02-14
wlan.4 - Removed the Aironet Communications kernel reference for 4500/4800 devices

chrisdavidson:
wlan.4 - This is also references wifi.4. The Aironet Communication module for 4500/4800 devices are no longer in FreeBSD. It was confirmed that the last time it was available is Release 13.2 - https://man.freebsd.org/cgi/man.cgi?query=an&apropos=0&sektion=4&manpath=FreeBSD+13.2-RELEASE&arch=default&format=html


emaste:
Thank you for your contribution. Pushed to main as bbd29c4394de.

PR #1121, opened 2024-02-15
math: Add long double constant definitions

collinfunk:
These constants are GNU libc extensions that are likely to be adopted by the next POSIX revision [1]. The definitions can be verified in a PARI-GP shell session:

* M_El: exp (1)
* M_LOG2El: log (exp (1)) / log (2)
* M_LOG10El: log (exp (1)) / log (10)
* M_LN2l: log (2)
* M_LN10l: log (10)
* M_PIl: Pi
* M_PI_2l: Pi / 2
* M_PI_4l: Pi / 4
* M_1_PIl: 1 / Pi
* M_2_PIl: 2 / Pi
* M_2_SQRTPIl: 2 / sqrt (Pi)
* M_SQRT2l: sqrt (2)
* M_SQRT1_2l: 1 / sqrt (2)

[1] https://austingroupbugs.net/view.php?id=828


bsdimp:
short term NAK on this (leave the PR open).
We need to be careful about namespace pollution and I'd like to start marking those glibc extensions that are likely going to be in POSIX issue8... At the very least it needs to be marked with #if __BSD_VISIBLE I'd think today.

collinfunk:
> short term NAK on this (leave the PR open). We need to be careful about namespace pollution and I'd like to start marking those glibc extensions that are likely going to be in POSIX issue8... At the very least it needs to be marked with #if __BSD_VISIBLE I'd think today.

No worries. I'll leave it open. These values are under '#if __BSD_VISIBLE || __XSI_VISIBLE' along with the non-long double constants. The diff on GitHub isn't large enough to show that so I figured I'd mention it.

bsdimp:
> No worries. I'll leave it open. These values are under '#if __BSD_VISIBLE || __XSI_VISIBLE' along with the non-long double constants. The diff on GitHub isn't large enough to show that so I figured I'd mention it.

Oh, then remove my NAK hold if others can verify this. Maybe add `if __BSD_VISIBLE || __XSI_VISIBLE >= 800` (the latter of which we don't actually set).

collinfunk:
> Oh, then remove my NAK hold if others can verify this. Maybe add `if __BSD_VISIBLE || __XSI_VISIBLE >= 800` (the latter of which we don't actually set).

Sure, that makes sense to me. I'll wait to see if any more people have comments before actually changing the commit though. I don't want to change the visibility of the existing double constants (no 'l' suffix). But these new ones being under `__XSI_VISIBLE >= 800` seems like a good idea to me. Thanks.

allanjude:
> > No worries. I'll leave it open. These values are under '#if __BSD_VISIBLE || __XSI_VISIBLE' along with the non-long double constants. The diff on GitHub isn't large enough to show that so I figured I'd mention it.
> 
> Oh, then remove my NAK hold if others can verify this. Maybe add `if __BSD_VISIBLE || __XSI_VISIBLE >= 800` (the latter of which we don't actually set).

I have verified that with expanding the context a few lines, it is behind the existing `#if __BSD_VISIBLE || __XSI_VISIBLE` check in the file

bsdimp:
I'll merge this, but Add the _XSI_VISISBLE change, though we have a long ways to go for full 800 support.

bsdimp:
merged with the change I requested.

collinfunk:
Sorry, I saw this a bit late. Thanks for making that change for me.

PR #1124, opened 2024-02-21
sys/amd64/conf: unify MINIMAL and GENERIC debug

llfw:
Add debugging features (WITNESS, INVARIANTS, etc.) to MINIMAL to match the options enabled in GENERIC.

Add a new MINIMAL-NODEBUG kernel which turns these features off in the same way GENERIC-NODEBUG does.

This reduces the differences between GENERIC and MINIMAL in a way that should be less confusing to users.

---

this only addresses amd64 for now.  other archs can have the equivalent changes made afterwards if this causes no issues.

### DIFF
--- a/sys/amd64/conf/GENERIC
+++ b/sys/amd64/conf/GENERIC
@@ -94,25 +94,7 @@ options 	RCTL			# Resource limits
 options 	KDB			# Enable kernel debugger support.
 options 	KDB_TRACE		# Print a stack trace for a panic.
 # For full debugger support use (turn off in stable branch):
-options 	BUF_TRACKING		# Track buffer history
-options 	DDB			# Support DDB.
-options 	FULL_BUF_TRACKING	# Track more buffer history
-options 	GDB			# Support remote GDB.
-options 	DEADLKRES		# Enable the deadlock resolver
-options 	INVARIANTS		# Enable calls of extra sanity checking
-options 	INVARIANT_SUPPORT	# Extra sanity checks of internal structures, required by INVARIANTS
-options 	QUEUE_MACRO_DEBUG_TRASH	# Trash queue(2) internal pointers on invalidation
-options 	WITNESS			# Enable checks to detect deadlocks and cycles
-options 	WITNESS_SKIPSPIN	# Don't run witness on spinlocks for speed
-options 	MALLOC_DEBUG_MAXZONES=8	# Separate malloc(9) zones
-options 	VERBOSE_SYSINIT=0	# Support debug.verbose_sysinit, off by default
-
-# Kernel Sanitizers
-#options 	COVERAGE		# Generic kernel coverage. Used by KCOV
-#options 	KCOV			# Kernel Coverage Sanitizer
-# Warning: KUBSAN can result in a kernel too large for loader to load
-#options 	KUBSAN			# Kernel Undefined Behavior Sanitizer
-#options 	KCSAN			# Kernel Concurrency Sanitizer
+include ../../conf/std.debug
 
 # Kernel dump features.
 options 	EKCD			# Support for encrypted kernel dumps
--- a/sys/amd64/conf/GENERIC
+++ b/sys/amd64/conf/GENERIC
@@ -94,7 +94,7 @@ options 	RCTL			# Resource limits
 options 	KDB			# Enable kernel debugger support.
 options 	KDB_TRACE		# Print a stack trace for a panic.
 # For full debugger support use (turn off in stable branch):
-include ../../conf/std.debug
+include "std.debug"
 
 # Kernel dump features.
 options 	EKCD			# Support for encrypted kernel dumps
--- a/sys/amd64/conf/MINIMAL
+++ b/sys/amd64/conf/MINIMAL
@@ -79,7 +79,7 @@ options 	INCLUDE_CONFIG_FILE	# Include this file in kernel
 options 	KDB			# Enable kernel debugger support.
 options 	KDB_TRACE		# Print a stack trace for a panic.
 # For full debugger support use (turn off in stable branch):
-include ../../conf/std.debug
+include "std.debug"
 
 # Make an SMP-capable kernel by default
 options 	SMP			# Symmetric MultiProcessor Kernel
--- a/sys/amd64/conf/MINIMAL
+++ b/sys/amd64/conf/MINIMAL
@@ -78,6 +78,8 @@ options 	INCLUDE_CONFIG_FILE	# Include this file in kernel
 # Debugging support.  Always need this:
 options 	KDB			# Enable kernel debugger support.
 options 	KDB_TRACE		# Print a stack trace for a panic.
+# For full debugger support use (turn off in stable branch):
+include ../../conf/std.debug
 
 # Make an SMP-capable kernel by default
 options 	SMP			# Symmetric MultiProcessor Kernel
--- a/sys/amd64/conf/MINIMAL-NODEBUG
+++ b/sys/amd64/conf/MINIMAL-NODEBUG
@@ -6,6 +6,6 @@
 #NO_UNIVERSE
 
 include MINIMAL
-include "../../conf/std.nodebug"
+include "std.nodebug"
 
 ident   MINIMAL-NODEBUG
--- a/sys/amd64/conf/MINIMAL-NODEBUG
+++ b/sys/amd64/conf/MINIMAL-NODEBUG
@@ -1,0 +1,11 @@
+#
+# MINIMAL-NODEBUG -- Non-debug MINIMAL kernel.
+#
+# This is the MINIMAL equivalent to GENERIC-NODEBUG.
+
+#NO_UNIVERSE
+
+include MINIMAL
+include "../../conf/std.nodebug"
+
+ident   MINIMAL-NODEBUG
--- a/sys/conf/std.debug
+++ b/sys/conf/std.debug
@@ -1,0 +1,17 @@
+#
+# std.debug -- Enable debug options for -CURRENT.
+#
+
+options 	BUF_TRACKING		# Track buffer history
+options 	DDB			# Support DDB.
+options 	FULL_BUF_TRACKING	# Track more buffer history
+options 	GDB			# Support remote GDB.
+options 	DEADLKRES		# Enable the deadlock resolver
+options 	INVARIANTS		# Enable calls of extra sanity checking
+options 	INVARIANT_SUPPORT	# Extra sanity checks of internal structures, required by INVARIANTS
+options 	QUEUE_MACRO_DEBUG_TRASH	# Trash queue(2) internal pointers on invalidation
+options 	WITNESS			# Enable checks to detect deadlocks and cycles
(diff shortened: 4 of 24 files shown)


sys/amd64/conf/MINIMAL
```diff
@@ -78,6 +78,26 @@ options 	INCLUDE_CONFIG_FILE	# Include this file in kernel
...
+options 	FULL_BUF_TRACKING	# Track more buffer history
+options 	GDB			# Support remote GDB.
+options 	DEADLKRES		# Enable the deadlock resolver
+options 	INVARIANTS		# Enable calls of extra sanity checking
+options 	INVARIANT_SUPPORT	# Extra sanity checks of internal structures, required by INVARIANTS
+options 	QUEUE_MACRO_DEBUG_TRASH	# Trash queue(2) internal pointers on invalidation
+options 	WITNESS			# Enable checks to detect deadlocks and cycles
+options 	WITNESS_SKIPSPIN	# Don't run witness on spinlocks for speed
+options 	MALLOC_DEBUG_MAXZONES=8	# Separate malloc(9) zones
+options 	VERBOSE_SYSINIT=0	# Support debug.verbose_sysinit, off by default
```
igalic, line 93: let's do a separate commit, and put all this stuff into `../../conf/std.debug`

igalic (changes requested):

sys/amd64/conf/MINIMAL-NODEBUG
```diff
@@ -0,0 +1,10 @@
+#
+# MINIMAL-NODEBUG -- Non-debug MINIMAL kernel.
+#
+# This is the MINIMAL equivalent to GENERIC-NODEBUG.
```
igalic, line 4: I think this should be annotated with

```
#NO_UNIVERSE
```

there's nothing we'll get out of compiling this during UNIVERSE

igalic (changes requested):

sys/amd64/conf/MINIMAL
llfw: done

igalic (approved):
i like it! :+1: 
Thank you very much for humouring my suggestion to add std.debug, this has been on my TODO list forever. This was a perfect opportunity.

bsdimp:
@markjdb should look at this too...

bsdimp:
Might want to do this for all architectures....

sys/amd64/conf/MINIMAL
```diff
@@ -78,6 +78,15 @@ options 	INCLUDE_CONFIG_FILE	# Include this file in kernel
...
 options 	KDB_TRACE		# Print a stack trace for a panic.
+# For full debugger support use (turn off in stable branch):
+include ../../conf/std.debug
+
+# Kernel Sanitizers
+#options 	COVERAGE		# Generic kernel coverage. Used by KCOV
+#options 	KCOV			# Kernel Coverage Sanitizer
+# Warning: KUBSAN can result in a kernel too large for loader to load
+#options 	KUBSAN			# Kernel Undefined Behavior Sanitizer
+#options 	KCSAN			# Kernel Concurrency Sanitizer
```
markjdb, line 89: Why add these lines? Why not KASAN and KMSAN too?

This looks like it belongs in a different commit.

sys/conf/std.debug
```diff
@@ -0,0 +1,16 @@
+#
+# std.nodebug -- Enable debug options for -CURRENT.
```
markjdb, line 2: This should be `std.debug`.

markjdb (commented):
I like this. It's a step we need towards adding a GENERIC-DEBUG to stable branches.

We should make the same change for at least arm64. i386 wouldn't hurt either.

sys/amd64/conf/MINIMAL
llfw: > Why add these lines? Why not KASAN and KMSAN too?

this is what's currently in `GENERIC`, i just copied it as-is.  ASAN and KMSAN are only referenced in their dedicated config files `GENERIC-K?SAN`.  curiously, `CSAN` is mentioned in both places

this might be a good time to make this more consistent; does it make sense to just document all the *SAN options in GENERIC/MINIMAL?

sys/conf/std.debug
llfw: fixed.

sys/amd64/conf/MINIMAL
markjdb: Ah, I see. Would it make more sense to have it in std.debug, so that they're not duplicated?

llfw: yes, but since users can't (or shouldn't) edit std.debug just to build a new kernel, i think this should come with a comment in GENERIC referencing the file and suggesting copying the lines from there to the appropriate kernel config -- does that seem reasonable?

llfw:
i'll add equivalent configs for i386 and arm64.  i could do other architectures at the same time but i can't easily test those (other than buildkernel) as i don't use them or have appropriate hardware.

sys/amd64/conf/MINIMAL
markjdb: Well, given that we already have GENERIC-KASAN etc., this stuff probably really belongs in sys/amd64/conf/NOTES.

markjdb:
> i'll add equivalent configs for i386 and arm64. i could do other architectures at the same time but i can't easily test those (other than buildkernel) as i don't use them or have appropriate hardware.

If you do the other architectures, I'll test them (in QEMU). As mentioned, I'd like to have GENERIC-DEBUG available on stable branches, and it'll be easiest if all platforms which have a GENERIC config can also have GENERIC-DEBUG.

If you still prefer not to handle the other platforms, I can handle them as a follow-up.

llfw:
i've added i386 and arm64.  this slightly changes the default debug options on all three platforms:

- amd64 gains ALT_BREAK_TO_DEBUGGER (previously only on arm64)
- i386 gains ALT_BREAK_TO_DEBUGGER (from arm64) and BUF_TRACKING, FULL_BUF_TRACKING and QUEUE_MACRO_DEBUG_TRASH (from amd64)
- arm64 gains BUF_TRACKING, FULL_BUF_TRACKING and QUEUE_MACRO_DEBUG_TRASH (from amd64)

basically, all 3 archs now have the set of all debug options which were previously enabled on any of them.  i think this is desirable and was probably overlooked by accident before.

i'll look at doing other archs too, but it'll take a while as building all these kernerls on my puny desktop takes a long time!

sys/amd64/conf/MINIMAL
llfw: i've moved these to NOTES for amd64.  other archs might need similar changes, i'll check those too.

sys/amd64/conf/GENERIC
```diff
@@ -94,25 +94,7 @@ options 	RCTL			# Resource limits
...
-options 	MALLOC_DEBUG_MAXZONES=8	# Separate malloc(9) zones
-options 	VERBOSE_SYSINIT=0	# Support debug.verbose_sysinit, off by default
-
-# Kernel Sanitizers
-#options 	COVERAGE		# Generic kernel coverage. Used by KCOV
-#options 	KCOV			# Kernel Coverage Sanitizer
-# Warning: KUBSAN can result in a kernel too large for loader to load
-#options 	KUBSAN			# Kernel Undefined Behavior Sanitizer
-#options 	KCSAN			# Kernel Concurrency Sanitizer
+include ../../conf/std.debug
```
bsdimp, line 97: I'd have thought that a simple 'std.debug' would work here. Does it not?

llfw: existing users of std.nodebug use the full (relative) path, but i'll test this and update all of them if it works (waiting for powerpc to build...)

bsdimp:
include std.debug
or
include std.nodebug
is going to work a lot better than the relative paths. People do crazy things with their config files and relative paths might not be what you think they are. But if you put std.debug and std.nodebug in sys/conf, you're golden.

sys/amd64/conf/GENERIC
bsdimp: the existing uses are wrong too then.

bsdimp:
I'd prep one commit before all of yours to go from ../../conf/std.nodebug to std.nodebug (with quotes) and fix yours.
Thanks for doing this... there's other things we might want to do this way that I think I now know how we might be able to do.
config(8) sucks, but isn't horrific (we've kept it 1 inch away from being horrific with various hacks over the years).

bsdimp:
Or if you like, I can fix nodebug and push that. Either way... I have one for arm and amd64 to test my assertions of wrongness :)

llfw:
i'm about to push more commits for powerpc, riscv, and one commit to fix all includes of "include ../../" in sys/*/conf.  after that i'm happy to squash or reorder commits as needed before committing to main.

bsdimp:
> i'm about to push more commits for powerpc, riscv, and one commit to fix all includes of "include ../../" in sys/*/conf. after that i'm happy to squash or reorder commits as needed before committing to main.

OK. I'll sit tight and comment more extensively on that push then.

llfw:
this should be everything.  i haven't tested the last `include` change but i'm about to re-build all the kernels to check that.

llfw:
also, let me know how you want me to squash/merge these prior to commit, i'm sure these don't need to be individual commits.  perhaps one to add std.debug, one to change the GENERIC/MINIMAL kernels, and one to fix all the includes.

llfw:
tested here in buildkernel with no issues, i am just installing qemu to see if i can use markjdb/bricoler to easily test all the new kernels in qemu without setting it all up by hand.

sys/conf/std.debug
```diff
@@ -0,0 +1,18 @@
...
+options 	DEADLKRES		# Enable the deadlock resolver
+options 	INVARIANTS		# Enable calls of extra sanity checking
+options 	INVARIANT_SUPPORT	# Extra sanity checks of internal structures, required by INVARIANTS
+options 	QUEUE_MACRO_DEBUG_TRASH	# Trash queue(2) internal pointers on invalidation
+options 	WITNESS			# Enable checks to detect deadlocks and cycles
+options 	WITNESS_SKIPSPIN	# Don't run witness on spinlocks for speed
+options 	MALLOC_DEBUG_MAXZONES=8	# Separate malloc(9) zones
+options 	VERBOSE_SYSINIT=0	# Support debug.verbose_sysinit, off by default
+options		ALT_BREAK_TO_DEBUGGER	# Enter debugger on keyboard escape sequence
+
```
bsdimp, line 18: Delete this last line...

bsdimp:
One nit... (the blank line) 
two commits seem best for this: (1) to fix the path thing. (2) the rest.

llfw:
tested on amd64.amd64, arm64.aarch64, riscv.riscv64, i386.386, GENERIC boots fine.  still pending armv6, armv7, powerpc, powerpc64, powerpc64le.  sorry, this is taking a while due how long it takes me to build a release.

llfw:
i've tested this on amd64, arm64, riscv and arm64.  testing a kernel on other architectures seems to be unreasonable complicated so i'm going to ignore those - anyone else is welcome to test before committing.

bsdimp:
landed

PR #1125, opened 2024-02-23
ACPI/IIC: Miscellaneous changes

VexedUXR:
Some changes that I feel are too small to split into multiple pull requests.

acpi_iicbus:
The underlying i2c driver expects the addresses it receives to need shifting, the opregion handler however receives the addresses already shifted, so shift the address to the left in the handler before sending it off.
See https://uefi.org/specs/ACPI/6.5/13_System_Mgmt_Bus_Interface_Specification.html#smbus-slave-addresses

ig4_iic:
Some DSDTs define non-existent devices, its a little disconnecting to have a device present on the DSDT then have it seemingly ignored by the system, so I added a warning for it (only prints if bootverbose is set).

The other two changes are fairly trivial, one updates an outdated link and the other replaces an incorrect use of sizeof with nitems.

Also, about the acpi_iicbus driver, what testing would be needed to get the address space handler enabled by default? It currently works just fine with my laptop's battery and i2c(8), will see if I can test it on other things.

Tested on the Acer one 10 s1002.

### DIFF
--- a/sys/dev/acpica/acpi_cmbat.c
+++ b/sys/dev/acpica/acpi_cmbat.c
@@ -344,7 +344,7 @@ acpi_cmbat_get_bix(void *arg)
     bix_buffer.Pointer = NULL;
     bix_buffer.Length = ACPI_ALLOCATE_BUFFER;
 
-    for (n = 0; n < sizeof(bobjs); n++) {
+    for (n = 0; n < nitems(bobjs); n++) {
 	as = AcpiEvaluateObject(h, bobjs[n].name, NULL, &bix_buffer);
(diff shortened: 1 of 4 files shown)


ehem (approved):
This looks like an outright bug and **should** be fixed.

bsdimp:
This looks good, so say I and @markjdb

bsdimp:
Merged. Thanks for your contribution. Sorry it took a little while to get things moving from PRs into the tree again...

VexedUXR:
> Sorry it took a little while to get things moving from PRs into the tree again...

All good :)

PR #1126, opened 2024-02-23
misc cleanups and some Phabricator commits

ehem:
Alas, not much of a theme, just some small patches for things which are overdue for getting in.

### DIFF
--- a/sys/arm/arm/pmu.c
+++ b/sys/arm/arm/pmu.c
@@ -46,6 +46,7 @@
 #include <sys/timetc.h>
 #include <sys/pmc.h>
 #include <sys/pmckern.h>
+#include <sys/proc.h>
 
 #include <machine/bus.h>
 #include <machine/cpu.h>
@@ -65,9 +66,6 @@ uint32_t ccnt_hi[MAXCPU];
 static int
 pmu_intr(void *arg)
 {
-#ifdef HWPMC_HOOKS
-	struct trapframe *tf;
-#endif
 	uint32_t r;
 #if defined(__arm__) && (__ARM_ARCH > 6)
 	u_int cpu;
@@ -87,10 +85,8 @@ pmu_intr(void *arg)
 
 #ifdef HWPMC_HOOKS
 	/* Only call into the HWPMC framework if we know there is work. */
-	if (r != 0 && pmc_intr) {
-		tf = arg;
-		(*pmc_intr)(tf);
-	}
+	if (r != 0 && pmc_intr)
+		(*pmc_intr)(curthread->td_intr_frame);
 #endif
 
 	return (FILTER_HANDLED);
--- a/sys/arm/include/atomic.h
+++ b/sys/arm/include/atomic.h
@@ -1052,6 +1052,12 @@ atomic_thread_fence_seq_cst(void)
 	dmb();
 }
 
+#define atomic_add_ptr			atomic_add_32
+#define atomic_add_acq_ptr		atomic_add_acq_32
+#define atomic_add_rel_ptr		atomic_add_rel_32
+#define atomic_subtract_ptr		atomic_subtract_32
+#define atomic_subtract_acq_ptr		atomic_subtract_acq_32
+#define atomic_subtract_rel_ptr		atomic_subtract_rel_32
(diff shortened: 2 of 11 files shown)


ehem:
First four are simply things for which Phabricator is ridiculous overkill.  Very simple, self-contained and not really worthy of major reviewer time.

Last 3 did go through Phabricator, but that doesn't really have a good way of getting things into FreeBSD's repository.  The last one the order of the first two arguments to `intr_add_handler()` was swapped for the stated reason.

bsdimp:
Normally I'd just close this due to scoping issues. However, several of these things can be landed due to reviewed in Phab, so I'll make an exception and land what's good and we'll see where we are.

ehem (commented):
Actually, this should be marked `Fixes: 4c72d075a57 ("LinuxKPI: const argument to irq_set_affinity_hint()")`.  4c72d075a57 _should have_ gone further and marked the argument `const`, but alas stopped outside the core.

ehem (commented):
Worth mentioning, this is in fact the _last_ use of this construct.  Bring this into the tree and [D37688](https://reviews.freebsd.org/D37688) can then go in.  That bit of cruft from i386 could //finally// be removed 20(?) years after it was considered legacy.

ehem:
Problem is these are rather _tiny_.  Seems overkill to use individual pull requests for several near-trivial fixes.

ehem:
Mostly rebasing forward as pull branches tend to do.  Did adjust the commit message of commit 3 as I suggested.

bsdimp:
OK. Thanks for winnowing down the original set to these, which were ready to go. I didn't see anything that needs changing.  It builds and boots for me as well. I know this has taken a while to land, so thank you for your contribution and patience.

bsdimp:
So this should autoclose as merged due to my force push. I'll give it a few minutes to catch up.

ehem:
The method you're using loses the author dates.  Those can be useful for figuring out how long it takes patches to get through various processes.  For instance that 3 of these patches had been in waiting in queues for 2 years.

bsdimp:
OK. The commit date is what most people want to be monotonic and we'll do a commit hook for. The author date may or may not be preserved, and I make no commitment to getting it right: it's one more thing on landing pull requests that slows down the process. I found an --ignore-date that I'll remove in my scripting, but if things get lost, they get lost. Phabricator doesn't even have this data, that I can see, and our tools go out of their way to reset the commit date to now after `git arc stage`. We're also evaluating other systems that may or may not preserve author dates, so the project really can't commit to preserving them. Again, I hate to be so grumpy about this, but it's already hard enough landing pull requests, and adding additional burden there makes no sense unless there's a clear, compelling reason for me to pay the per-pull-request tax. So I'll do the easy thing ( remove the --ignore-date), but you get what you get, and if there's some that land with it and some that land without due to some crazy difference, that's something that likely cannot be helped.

ehem:
> OK. The commit date is what most people want to be monotonic and we'll do a commit hook for. The author date may or may not be preserved, and I make no commitment to getting it right: it's one more thing on landing pull requests that slows down the process. I found an --ignore-date that I'll remove in my scripting, but if things get lost, they get lost. Phabricator doesn't even have this data, that I can see, and our tools go out of their way to reset the commit date to now after `git arc stage`. We're also evaluating other systems that may or may not preserve author dates, so the project really can't commit to preserving them. Again, I hate to be so grumpy about this, but it's already hard enough landing pull requests, and adding additional burden there makes no sense unless there's a clear, compelling reason for me to pay the per-pull-request tax. So I'll do the easy thing ( remove the --ignore-date), but you get what you get, and if there's some that land with it and some that land without due to some crazy difference, that's something that likely cannot be helped.

Thank you.


-- 
(\___(\___(\______          --=> 8-) EHM <=--          ______/)___/)___/)
 \BS (    |         ***@***.***  PGP 87145445         |    )   /
  \_CS\   |  _____  -O #include <stddisclaimer.h> O-   _____  |   /  _/
8A19\___\_|_/58D2 7E3D DDF4 7BA6 <-PGP-> 41D1 B375 37D0 8714\_|_/___/5445

PR #1127, opened 2024-02-27
Typo in s_exp2f function comment

hchataing:


bsdimp:
This is an actual math error in the comment. It's also present in s_exp2.c as well. I fixed that and wrote up a more verbose commit message, as well as confirming the basic math with the original paper (which I put into the commit message).

PR #1128, opened 2024-03-01
read.2 - resolved a self-reference link in manual page

chrisdavidson:
read.2 - Identified a self-reference in read(2) with .Xr and transitioned them to .Nm to resolve the mandoc -Tlint error message

### DIFF
--- a/lib/libsys/read.2
+++ b/lib/libsys/read.2
@@ -129,15 +129,23 @@ before the end-of-file, but in no other case.
 In accordance with
 .St -p1003.1-2004 ,
 both
+<<<<<<< HEAD
 .Xr read 2
+=======
+.Fn read
+>>>>>>> 29f97d973eec (read.2 - udpated both write / read references from .Xr to .Fn to align to function calls)
 and
-.Xr write 2
+.Fn write
 syscalls are atomic with respect to each other in the effects on file
 content, when they operate on regular files.
 If two threads each call one of the
+<<<<<<< HEAD
 .Xr read 2
+=======
+.Fn read
+>>>>>>> 29f97d973eec (read.2 - udpated both write / read references from .Xr to .Fn to align to function calls)
 or
-.Xr write 2 ,
+.Fn write ,
 syscalls, each call will see either all of the changes of the other call,
 or none of them.
 The


lib/libsys/read.2
```diff
@@ -129,13 +129,13 @@ before the end-of-file, but in no other case.
 In accordance with
 .St -p1003.1-2004 ,
 both
-.Xr read 2
+.Nm read 2
```
bsdimp, line 132: Now that I see this, I think ".Fn read"

lib/libsys/read.2
```diff
@@ -129,13 +129,13 @@ before the end-of-file, but in no other case.
...
 and
 .Xr write 2
```
bsdimp, line 134: I'd consider making this ".Fn write" here too.

And the same below.

lib/libsys/read.2
chrisdavidson: all fixed.

bsdimp:
Looks good now. @bsdbcr can you give me a final sanity check on the advice here?

bsdbcr:
@bsdimp Looks good to me!

lib/libsys/read.2
```diff
@@ -129,16 +129,15 @@ before the end-of-file, but in no other case.
...
-.Xr write 2
+.Fn write
 syscalls are atomic with respect to each other in the effects on file
 content, when they operate on regular files.
 If two threads each call one of the
-.Xr read 2
 .Fn read 2
 or
-.Xr write 2 ,
+.Fn write .
```
bsdimp, line 140: This should be a comma `','` not a period '.'.

bsdimp:
So this needs to be curated a bit.
The last two commits add nothing since one reverts the other, so should be dropped.
and I think that the other commits should be squashed. this seems like it should just be one commit.

chrisdavidson:
> So this needs to be curated a bit. The last two commits add nothing since one reverts the other, so should be dropped. and I think that the other commits should be squashed. this seems like it should just be one commit.

No problem, let me go figure out how to consolidate these into one commit.

PR #1137, opened 2024-03-11
pow,powf(3),__ieee754_rem_pio2(f): Avoid undefined behavior from left shift of negative signed integer values

karlt:
A compiler clever enough to know that z is positive with a non-zero
biased exponent could, for example, optimize away the scalbnf(z,n) in
pow() because behavior for left shift of negative values is undefined.
`n` is negative when y*log2(|x|) < -0.5.  i.e. |x^y| < sqrt(0.5)

The intended behavior for operator<< in this code is to shift the two's
complement representation of the first operand.

In the pow() functions, the result is added to the IEEE 754 exponent of
z = 2^y'.  n may be negative enough to underflow the biased IEEE 754
exponent below zero, which is manifested in the sign bit of j
(which would correspond to the IEEE 754 sign bit).

The conversion from uint32_t to int32_t for out-of-int32_t-range values
is implementation defined.  The assumed behavior of interpreting the
uint32_t value as a two's complement representation of a signed value
is already assumed in many parts of the code, such as uses of
GET_FLOAT_WORD() with signed integers.

### DIFF
--- a/lib/msun/src/e_pow.c
+++ b/lib/msun/src/e_pow.c
@@ -299,7 +299,11 @@ pow(double x, double y)
 	r  = (z*t1)/(t1-two)-(w+z*w);
 	z  = one-(r-z);
 	GET_HIGH_WORD(j,z);
-	j += (n<<20);
+    /*
+     * sign bit of z is 0.
+     * sign bit of j will indicate sign of 0x3ff-biased exponent.
+     */
+	j += (int32_t)((uint32_t)n<<20);
 	if((j>>20)<=0) z = scalbn(z,n);	/* subnormal output */
 	else SET_HIGH_WORD(z,j);
 	return s*z;
--- a/lib/msun/src/e_powf.c
+++ b/lib/msun/src/e_powf.c
@@ -242,7 +242,11 @@ powf(float x, float y)
 	r  = (z*t1)/(t1-two)-(w+z*w);
 	z  = one-(r-z);
 	GET_FLOAT_WORD(j,z);
-	j += (n<<23);
+    /*
+     * sign bit of z is 0.
+     * sign bit of j will indicate sign of 0x7f-biased exponent.
+     */
+	j += (int32_t)((uint32_t)n<<23);
 	if((j>>23)<=0) z = scalbnf(z,n);	/* subnormal output */
 	else SET_FLOAT_WORD(z,j);
 	return sn*z;
--- a/lib/msun/src/e_rem_pio2.c
+++ b/lib/msun/src/e_rem_pio2.c
@@ -162,7 +162,7 @@ medium:
     /* set z = scalbn(|x|,ilogb(x)-23) */
 	GET_LOW_WORD(low,x);
 	e0 	= (ix>>20)-1046;	/* e0 = ilogb(z)-23; */
-	INSERT_WORDS(z, ix - ((int32_t)(e0<<20)), low);
+	INSERT_WORDS(z, ix - ((int32_t)((uint32_t)e0<<20)), low);
 	for(i=0;i<2;i++) {
 		tx[i] = (double)((int32_t)(z));
 		z     = (z-tx[i])*two24;
--- a/lib/msun/src/e_rem_pio2f.c
+++ b/lib/msun/src/e_rem_pio2f.c
@@ -67,7 +67,7 @@ __ieee754_rem_pio2f(float x, double *y)
 	}
     /* set z = scalbn(|x|,ilogb(|x|)-23) */
 	e0 = (ix>>23)-150;		/* e0 = ilogb(|x|)-23; */
-	SET_FLOAT_WORD(z, ix - ((int32_t)(e0<<23)));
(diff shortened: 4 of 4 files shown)


lib/msun/src/e_pow.c
```diff
@@ -299,7 +299,9 @@ pow(double x, double y)
 	r  = (z*t1)/(t1-two)-(w+z*w);
 	z  = one-(r-z);
 	GET_HIGH_WORD(j,z);
-	j += (n<<20);
+    /* sign bit of z is 0.
+       sign bit of j will indicate sign of 0x3ff-biased exponent */
```
bsdimp, line 303: Style(9) wants this comment to look more like
```
/*
 * sign bit of z is 0
 * sign bit of j will indicate sign of 0x3ff-biased exponent.
 */
```

bsdimp:
This commit looks right to my eye, modulo the style stuff. I'd be happier if I could get @jrtc27 or @markjdb or @DimitryAndric to double check me since it is a signed-bug. We know n is positive, but I think the compiler might not and thus is allowed to do any old silly thing, hence the need for the double cast.

lib/msun/src/e_pow.c
```diff
@@ -299,7 +299,9 @@ pow(double x, double y)
...
+	j += (int32_t)((u_int32_t)n<<20);
```
bsdimp, line 304: We prefer uint32_t instead of u_int32_t generally in the FreeBSD tree, especially for new code.

bsdimp:
Also maybe a better title would be 'Avoid left shift of a signed integer' instead?
Sorry for the flurry of comments for such a simple commit.

lib/msun/src/e_pow.c
karlt: Ah, thanks. There was a mix of style in this file and i missed that this is preferred.

lib/msun/src/e_pow.c
karlt: I can change that too.
`u_int32_t` is used elsewhere in this file and in most of the occurrences in this directory.
`uint32_t` is used in a few files. Both `u_int32_t` and `uint32_t` are used in only two.

karlt:
Thanks very much for taking a lot.
This code takes quite some effort to understand.

> We know n is positive, but I think the compiler might not and thus is allowed to do any old silly thing, hence the need for the double cast.

`n` may be (and often is) negative here.

`n` is always non-negative when set at e_pow.c line 287, but its sign is changed on the next line when `j<0` (i.e. `y*log2(x) < -0.5`) to share the same sign as `j`.

`y*log2(x) = n+y'` where `|y'|<=0.5.`

> Also maybe a better title would be 'Avoid left shift of a signed integer' instead?

So, 'Avoid undefined behavior from left shift of negative signed integer values' would be appropriate.

DimitryAndric:
Lots of msun functions are using undefined (or at least skirting the edge of it) behavior, but I'm reluctant to start messing in there without having a good method to ensure there are no regressions. In this particular case it looks like the code tries pretty hard to ensure that n is nonnegative, so it should normally never be an issue.

Has anybody tried building msun with UBSan, and then ran the full test suite?

DimitryAndric:
FWIW I built `lib/msun` with `MK_UBSAN=yes` and then ran the whole test suite. No failures. So that *seems* at least hopeful. :)

phoddie:
FWIW – while fuzz testing XS, our JavaScript engine for embedded devices, we encountered the same UBSAN report for `pow`. The change here solved it. Thank you! 

We had a similar error reported in [e_rem_pio2.c](https://github.com/freebsd/freebsd-src/blob/3f115b864b1fba951ce60970b5febb4e593b1f77/lib/msun/src/e_rem_pio2.c#L165) and solved it using the [same casting](https://github.com/Moddable-OpenSource/moddable/commit/3d6d0f86745a3f1f217540d40785f4a5fe45be8a#diff-b037893a61e416df49560b7bf7573f7b55dbb77c79323c3506eda024788e1dba) as this PR. I thought I'd report that here in case it makes sense to fix that problem as part of this PR.

kargl:
Warner asked me (in a private email) to look at the suggested patch as my fingerprints are all over libm source code.  (Unfortunately, I fat-fingered my reply to him and accidentally deleted the email.)

Two things.  Use `u_int32_t` it is used everywhere else in the file and fdlibm-derived code is far from style(9).   Neither NetBSD nor OpenBSD have this casting, and in the past we managed to keep the source repositories in sync.

I haven't done any extensive testing, but believe that the patch is OK.  It times like this that I miss bde and his 3 long paragraphs about the correctness of the current code or the patch.

bsdimp:
OK. I'll change it back to `u_int32_t` then when I land it. Since I asked for the change originally, and it's the only change I see that's needed, I'm not going to ask for it to be changed back. I'll fix it myself. Thanks @kargl. I miss bde too... and not just for his 3 long paragraphs with code corrections.

bsdimp:
closed! Thanks. e75a1bbc233bf1

PR #1138, opened 2024-03-16
mv: Set file flags after setting file times

ricardobranco777:
Certain file flags prevent the setting of the file times, thus we should set the file times first.

This is done right in NetBSD: https://github.com/NetBSD/src/blob/trunk/bin/mv/mv.c#L327

In cp we also do the right thing: https://github.com/freebsd/freebsd-src/blob/main/bin/cp/utils.c#L300

### DIFF
--- a/bin/mv/mv.c
+++ b/bin/mv/mv.c
@@ -319,6 +319,12 @@ err:		if (unlink(to))
 	 */
 	preserve_fd_acls(from_fd, to_fd, from, to);
 	(void)close(from_fd);
+
+	ts[0] = sbp->st_atim;
+	ts[1] = sbp->st_mtim;
+	if (futimens(to_fd, ts))
+		warn("%s: set times", to);
+
 	/*
 	 * XXX
 	 * NFS doesn't support chflags; ignore errors unless there's reason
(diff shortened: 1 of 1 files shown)


markjdb:
Which particular flags prevent updating file times? I'm wondering if we can easily add a test case for this.

bin/mv/mv.c
```diff
@@ -319,6 +319,12 @@ err:		if (unlink(to))
 	 */
 	preserve_fd_acls(from_fd, to_fd, from, to);
 	(void)close(from_fd);
+
+	ts[0] = sbp->st_atim;
+	ts[1] = sbp->st_mtim;
+	if (futimens(to_fd, ts))
+		warn("%s: set times", to);
```
markjdb, line 326: Do we potentially want to do this before copying ACLs as well?

ricardobranco777:
> Which particular flags prevent updating file times? I'm wondering if we can easily add a test case for this.

Testing with uappend like this `touch ~/test ; chflags uappend ~/test`.

Without this patch:
```
mv ~/test /mnt/
mv: /mnt/test: set times: Operation not permitted
mv: /home/ricardo/test: remove: Operation not permitted
```

With this patch:

```
mv ~/test /mnt/
mv: /home/ricardo/test: remove: Operation not permitted
```

bin/mv/mv.c
ricardobranco777: I haven't played with extended attributes or ACL's yet, so I don't know...

markjdb:
Merged as of commit 428f86fd2ff1fd2b073d556fac273c8c7f457376, thank you.

PR #1139, opened 2024-03-22
genet: add module (if_genet.ko)

llfw:
The driver already had the appropriate module macros, it just wasn't hooked into the build system.

Since this device is specific to the Raspberry Pi 4, only build it for AArch64.

---

tested with monolithic (GENERIC) kernel and modular kernel without miibus (using miibus.ko) on an RPi4.

### DIFF
--- a/sys/modules/Makefile
+++ b/sys/modules/Makefile
@@ -154,6 +154,7 @@ SUBDIR=	\
 	if_edsc \
 	${_if_enc} \
 	if_epair \
+	${_genet} \
 	${_if_gif} \
 	${_if_gre} \
 	${_if_me} \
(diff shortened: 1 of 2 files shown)


emaste (approved):
Looks good

bsdimp:
This isn't ready. It fails to build standalone.
bus_if.h device_if.h ofw_bus_if.h and opt_platform.h were not listed.
Thankfully, listing all of these produces the correct result. opt_platform.h has #define FDT 1 in it and the *if.h files are generated correctly when they are missing.
LINT works
LINT-FDT works
LINT-ACPI works
standalone (make buildenv TARGET=arm64 ; cd sys/modules/genet; make) works
I suggest `make universe -DWITH_ONLY_KERNELS TARGETS=arm64` for future submissions as well as a standalone build.

PR #1142, opened 2024-03-23
bcm2838_xhci: add module

llfw:
bcm2838_xhci(4) is a shim for the XHCI controller on the Raspberry Pi 4B SoC.  It loads the controller's firmware before passing control to the normal xhci(4) driver.

When xhci(4) is built as a module (and not in the kernel), bcm2838_xhci is not built at all and the RPi4's XHCI controller won't attach due to missing firmware.

To fix this, build a new module, bcm2838_xhci.ko, which depends on xhci.ko.  For the dependency to work correctly, also modify xhci to provide the 'xhci' module in addition to the 'xhci_pci' module it already provided.

Since bcm2838_xhci is specific to a quirk of the RPi4 SoC, only build the module for AArch64.

### DIFF
--- a/sys/modules/usb/Makefile
+++ b/sys/modules/usb/Makefile
@@ -44,7 +44,7 @@ MAKE+=" DEBUG_FLAGS+=-DUSB_REQ_DEBUG"
 
 SUBDIR = usb
 SUBDIR += ${_dwc_otg} ehci ${_musb} ohci uhci xhci ${_uss820dci} \
-	  ${_atmegadci} ${_avr32dci} ${_rsu} ${_rsufw}
+	  ${_atmegadci} ${_avr32dci} ${_rsu} ${_rsufw} ${_bcm2838_xhci}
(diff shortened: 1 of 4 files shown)


sys/modules/usb/Makefile
```diff
@@ -44,7 +44,7 @@ MAKE+=" DEBUG_FLAGS+=-DUSB_REQ_DEBUG"
 
 SUBDIR = usb
 SUBDIR += ${_dwc_otg} ehci ${_musb} ohci uhci xhci ${_uss820dci} \
-	  ${_atmegadci} ${_avr32dci} ${_rsu} ${_rsufw}
+	  ${_atmegadci} ${_avr32dci} ${_rsu} ${_rsufw} ${_bcm2838_xhci}
```
bsdimp, line 47: Normally, we sort things alphabetically, but this file is a mess...

bsdimp:
ac83063d37e5e92ad048cc4ed958654c02103f74

PR #1143, opened 2024-03-23
nvme(4), nvd(4): build modules on aarch64

llfw:
tested with a modular kernel (no 'nvme' or 'nvd' devices), with a QEMU emulated NVMe disk on a macOS host; with nvd.ko and nvme.ko loaded in loader.conf, the system is able to boot from an NVMe root device.

### DIFF
--- a/sys/modules/Makefile
+++ b/sys/modules/Makefile
@@ -704,6 +704,12 @@ _e6000sw=	e6000sw
 _neta=		neta
 .endif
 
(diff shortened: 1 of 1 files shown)


bsdimp:
@bsdjhb has a change that enables these for all architectures that I just reviewed. Closing this one since it will interfere and you'll get that functionality from his commit. Thank you for submitting this. My apologies for our tardiness that made it obsolete.

PR #1145, opened 2024-03-24
Add libxo support to du

nhuff:
Convert du to use libxo enabling structured output.

Signed-off-by: Nathan Huff <nhuff@acm.org>

### DIFF
--- a/usr.bin/du/du.c
+++ b/usr.bin/du/du.c
@@ -48,6 +48,7 @@
 #include <string.h>
 #include <sysexits.h>
 #include <unistd.h>
+#include <libxo/xo.h>
 
 #define SI_OPT	(CHAR_MAX + 1)
 
@@ -62,7 +63,7 @@ struct ignentry {
 
 static int	linkchk(FTSENT *);
 static void	usage(void);
-static void	prthumanval(int64_t);
+static void	prthumanval(const char *, int64_t);
 static void	ignoreadd(const char *);
 static void	ignoreclean(void);
 static int	ignorep(FTSENT *);
@@ -107,6 +108,10 @@ main(int argc, char *argv[])
 	depth = INT_MAX;
 	SLIST_INIT(&ignores);
 
+	argc = xo_parse_args(argc, argv);
+	if (argc < 0)
+		exit(EX_USAGE);
+
 	while ((ch = getopt_long(argc, argv, "+AB:HI:LPasd:cghklmnrt:x",
 	    long_options, NULL)) != -1)
 		switch (ch) {
@@ -117,7 +122,7 @@ main(int argc, char *argv[])
 			errno = 0;
 			cblocksize = atoi(optarg);
 			if (errno == ERANGE || cblocksize <= 0) {
-				warnx("invalid argument to option B: %s",
+				xo_warnx("invalid argument to option B: %s",
 				    optarg);
 				usage();
 			}
@@ -147,7 +152,7 @@ main(int argc, char *argv[])
 			errno = 0;
 			depth = atoi(optarg);
 			if (errno == ERANGE || depth < 0) {
-				warnx("invalid argument to option d: %s",
+				xo_warnx("invalid argument to option d: %s",
 				    optarg);
 				usage();
 			}
@@ -181,7 +186,7 @@ main(int argc, char *argv[])
 		case 't' :
 			if (expand_number(optarg, &threshold) != 0 ||
 			    threshold == 0) {
-				warnx("invalid threshold: %s", optarg);
+				xo_warnx("invalid threshold: %s", optarg);
 				usage();
 			} else if (threshold < 0)
 				threshold_sign = -1;
@@ -254,6 +259,8 @@ main(int argc, char *argv[])
 	if ((fts = fts_open(argv, ftsoptions, NULL)) == NULL)
 		err(1, "fts_open");
 
+	xo_open_container("disk-usage-information");
+	xo_open_list("paths");
 	while (errno = 0, (p = fts_read(fts)) != NULL) {
 		switch (p->fts_info) {
 		case FTS_D:			/* Ignore. */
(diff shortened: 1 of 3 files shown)


usr.bin/du/du.c
```diff
@@ -48,10 +48,11 @@
 #include <string.h>
 #include <sysexits.h>
 #include <unistd.h>
+#include <libxo/xo.h>
 
 #define SI_OPT	(CHAR_MAX + 1)
 
-#define UNITS_2		1
+#define UNITS_2	1
```
bsdimp, line 55: This seems gratuitous.

usr.bin/du/du.c
```diff
@@ -62,7 +63,7 @@ struct ignentry {
 
 static int	linkchk(FTSENT *);
 static void	usage(void);
-static void	prthumanval(int64_t);
+static void	prthumanval(const char*, int64_t);
```
bsdimp, line 66: FreeBSD's style is `const char *` with a space after the char.

usr.bin/du/du.c
```diff
@@ -273,15 +280,17 @@ main(int argc, char *argv[])
 			if (p->fts_level <= depth && threshold <=
 			    threshold_sign * howmany(p->fts_bignum *
 			    cblocksize, blocksize)) {
+				xo_open_instance("paths");
```
bsdimp, line 283: unless I'm missing something, and maybe I am, why are we opening this twice? Here and at line 263?

usr.bin/du/du.c
```diff
@@ -309,36 +318,40 @@ main(int argc, char *argv[])
 			    howmany(p->fts_statp->st_blocks, cblocksize);
 
 			if (aflag || p->fts_level == 0) {
+				xo_open_instance("paths");
 				if (hflag > 0) {
```
bsdimp, line 322: Same question.

usr.bin/du/du.c
```diff
@@ -453,9 +466,10 @@ linkchk(FTSENT *p)
 		/* Pull a node from the free list if we can. */
 		le = free_list;
 		free_list = le->next;
-	} else
+	} else {
 		/* Malloc one if we have to. */
 		le = malloc(sizeof(struct links_entry));
+	}
```
bsdimp, line 472: This is a good change, but it's gratuitous since you're not changing adjacent lines.

bsdimp (changes requested):
two minor nits that I see, and either a double push of paths, or something that I don't understand.

nhuff:
I've updated the pull request to address the style issues.

The two xo_open_instance("paths") calls happen in two different code paths during the filesystem traversal. One path is taken if you are reporting on a file and one path if you are reporting on a directory.

markjdb (commented):
Sorry, I had left these comments last week, but apparently failed to click submit. :(

usr.bin/du/du.c
```diff
@@ -453,9 +466,10 @@ linkchk(FTSENT *p)
...
 	if (le == NULL) {
 		stop_allocating = 1;
 		warnx("No more memory for tracking hard links");
```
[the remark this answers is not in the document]
nhuff: I'm using df as a template.  It didn't look like they converted the warnx on malloc failures.  Does xo_warnx allocate?  I figured it might and that is why df does it the way it does, but maybe I should check for sure.

usr.bin/du/du.c
```diff
@@ -107,6 +108,10 @@ main(int argc, char *argv[])
 	depth = INT_MAX;
 	SLIST_INIT(&ignores);
 
+	argc = xo_parse_args(argc, argv);
+	if (argc < 0)
+		exit(-1);
```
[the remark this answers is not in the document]
nhuff: Do you have an in tree example of what you would like to see here?   I'm trying to be consistent and all of the examples I'm finding just exit in this case.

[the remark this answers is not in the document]
markjdb: I meant, this line should be `exit(EX_USAGE)`.

usr.bin/du/du.c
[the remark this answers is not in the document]
markjdb: By default I believe that xo_warnx() etc. just output unstructured data to standard error, and they don't appear to allocate memory. (And libxo will print an error if it hits an internal memory allocation failure.)

usr.bin/du/du.c
```diff
@@ -488,13 +502,13 @@ prthumanval(int64_t bytes)
 
 	humanize_number(buf, sizeof(buf), bytes, "", HN_AUTOSCALE, flags);
 
-	(void)printf("%4s", buf);
```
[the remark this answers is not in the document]
nhuff: I don't think this PR changes the behavior here.  It changes from printf("%4s", buf) to an xo_emit with a format string that also always uses "%4s".  There might be an issue here, but if so I think it is pre-existing.

usr.bin/du/du.c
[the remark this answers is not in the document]
nhuff: In tree examples seem somewhat inconsistent here so I don't have any issues converting the rest of the calls.

nhuff:
I've updated the pull request to address most of the comments from @markjdb.  I didn't address the last comment since I don't think the behavior differs from the current implementation in any relevant way.

bsdimp:
my eyeballs say that this is ready. Will run some tests and either give feedback or push.

usr.bin/du/du.c
```diff
@@ -309,36 +318,41 @@ main(int argc, char *argv[])
...
 		} else {
-			(void)printf("%jd\ttotal\n", (intmax_t)howmany(
+			xo_emit("{:total-blocks/%jd}\ttotal\n", (intmax_t)howmany(
 			    savednumber * cblocksize, blocksize));
 		}
 	}
 
 	ignoreclean();
+	xo_close_container("disk-usage-information");
+	if(xo_finish() <0)
```
bsdimp, line 354: Not sure why this wasn't flagged with the style checker. if space ( and there needs to be a space after <.

usr.bin/du/du.c
```diff
@@ -273,15 +280,17 @@ main(int argc, char *argv[])
 			if (p->fts_level <= depth && threshold <=
 			    threshold_sign * howmany(p->fts_bignum *
 			    cblocksize, blocksize)) {
+				xo_open_instance("paths");
 				if (hflag > 0) {
-					prthumanval(p->fts_bignum);
-					(void)printf("\t%s\n", p->fts_path);
+					prthumanval("{:blocks/%4s}",p->fts_bignum);
```
bsdimp, line 285: this needs a space after the comma.

usr.bin/du/du.c
```diff
@@ -309,36 +318,41 @@ main(int argc, char *argv[])
...
 
 	if (errno)
-		err(1, "fts_read");
+		xo_err(1, "fts_read");
 
 	if (cflag) {
 		if (hflag > 0) {
-			prthumanval(savednumber);
-			(void)printf("\ttotal\n");
+			prthumanval("{:total-blocks/%4s}\ttotal\n",savednumber);
```
bsdimp, line 345: this also needs a space before saved number.

bsdimp (changes requested):
Please fix the style issues tagged.

bsdimp:
this looks like it will work... and  spot checking shows that it does. Just a couple style issue to fix.

bsdimp:
On second thought, the style issues were relatively minor and easy to fix while staging.
I also bumped the date on the man page, but that's OK: many man pages are active enough that dates conflict a lot.
I went ahead and fixed them this time, but next time please use checkstyle9.pl to make future submissions better.
Thanks for this  new functionality.

PR #1146, opened 2024-03-24
Instrument kernel boot time with capped DELAY

Subhechcha:
In order to compute the potential headroom of FreeBSD kernel boot time optimisation by tuning the DELAY calls, cap the max delay to 50ms and profile the boot time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@
 *.so
 *.sw[nop]
 *~
-.vscode
 _.tinderbox.*
 _.universe-toolchain
 _.amd64.*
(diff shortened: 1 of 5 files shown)


markjdb:
The commit logs in this PR are too vague, and I don't see a reason to have so many individual commits. They also make some unrelated changes, e.g., adding vscode files.

The change to delay_tc() is simply wrong. Device drivers may rely on specific values, it is not correct to clamp them like that.

markjdb:
These patches won't be accepted as they are. Feel free to submit a new PR if things change.

PR #1147, opened 2024-03-24
Instrumentation of zfsctrlmodule using TSLOG for fine grained observability

Subhechcha:


rilysh:
You should add some context in that commit message. Also, commit https://github.com/freebsd/freebsd-src/pull/1147/commits/d218c7b02fd20b3a85822fdfb94c83130b7142ad is adding arbitrary files related to the editor which needs to be removed.

markjdb:
Indeed, this patch needs to go upstream (to OpenZFS) in some form, which will involve adding TSLOG stubs for Linux. This can't be merged as-is.

PR #1153, opened 2024-04-04
usb: Fix and cleanup quirks list

VexedUXR:
The `USB_QUIRK_VP` macro was being misused in the usb_quirks list, e.g `USB_QUIRK_VP(USB_VENDOR_MPMAN, 0, UQ_MSC_NO_SYNC_CACHE, UQ_MATCH_VENDOR_ONLY)` would set lo_res and hi_res to the first two specified quirks.

I've replaced it with a new macro, `USB_QUIRK_VO`, which removes the unneeded verbosity. Having this macro also removes the need for the dummy products used to specify vendor only quirks.

I also separated the usb quirks that target specific revisions from those that dont. A lot of the quirks dont use
lo_rev and hi_rev, so we can abstract the 0x0000, 0xffff into a macro.

Edit: Forgot to mention that I also replaced a `|` with `,`. Looks like it was left over from eb58441
https://github.com/freebsd/freebsd-src/blob/main/sys/dev/usb/quirk/usb_quirk.c#L335


VexedUXR:
I separated the changes into two commits:
one replacing `USB_QUIRK_VP` and another adding `USB_QUIRK_REV`.

A large part of the changes are in the second commit, not sure I can separate them though.

bsdimp:
the latest changes look like they can be reviewed, so I'll go over them with a fine toothed comb to see if anything stands out.

VexedUXR:
That's new :)
Thanks.

PR #1154, opened 2024-04-04
Revert xz to 5.4.5

delphij:
This reverts 8db56defa766eacdbaf89a37f25b11a57fd9787a and cherry-picks xzutils [8f236574986e7c414c0ea059f441982d1387e6a4](https://git.tukaani.org/?p=xz.git;a=commitdiff;h=8f236574986e7c414c0ea059f441982d1387e6a4;hp=d0f33d672a4da7985ebb5ba8d829f885de49c171) and [d0f33d672a4da7985ebb5ba8d829f885de49c171](https://git.tukaani.org/?p=xz.git;a=commitdiff;h=d0f33d672a4da7985ebb5ba8d829f885de49c171;hp=c247d06e1f6cada9a76f4f6225cbd97ea760f52f) .


bsdimp:
I'd do this as two commits. Before the next vendor import, I'd revert the compat symbol, reapply 5.6.x, and then go from there. Otherwise you'll hit a fair number of complications when you do the next vendor import. I hit this when I backed out awk. And while it wasn't too bad, it was bad enough that I'm offering you that advice in the hopes it will be better. I don't know that it will be for sure, though.

As for this commit, direction and choice: I like it:
Reviewed-By: imp

delphij:
Revert pushed to `main`.

PR #1155, opened 2024-04-08
sys/netinet6/in6_pcb.c: fix compile without INET

llfw:
in6_mapped_sockaddr() and in6_mapped_peeraddr() both define a local variable named 'inp', but in the non-INET case, this variable is set and never used, causing a compiler error:

/src/freebsd/src/lf/sys/netinet6/in6_pcb.c:547:16: error:
	variable 'inp' set but not used [-Werror,-Wunused-but-set-variable]
  547 |         struct  inpcb *inp;
      |                        ^
/src/freebsd/src/lf/sys/netinet6/in6_pcb.c:573:16: error:
	variable 'inp' set but not used [-Werror,-Wunused-but-set-variable]
  573 |         struct  inpcb *inp;

Fix this by guarding all the INET-specific logic, including the variable definition, behind #ifdef INET.

While here, tweak formatting in in6_mapped_peeraddr() so both functions are the same.

### DIFF
--- a/sys/netinet6/in6_pcb.c
+++ b/sys/netinet6/in6_pcb.c
@@ -544,13 +544,13 @@ in6_getpeeraddr(struct socket *so, struct sockaddr *sa)
 int
 in6_mapped_sockaddr(struct socket *so, struct sockaddr *sa)
 {
-	struct	inpcb *inp;
 	int	error;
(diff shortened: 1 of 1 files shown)


bsdimp:
Looks sane to me.

PR #1156, opened 2024-04-08
sys/dev/irdma/irdma_cm: fix compile without INET

llfw:
irdma_get_vlan_ipv4() calls ip_ifp_find() even if INET isn't defined, in which case this function isn't available.

Stub this out for the non-INET case to return an error (0xffff) instead.


bsdimp:
merged.

PR #1157, opened 2024-04-08
sys/netpfil/pf: fix non-INET module build

llfw:
pf.ko, when built as a module without 'options INET' but with 'options VIMAGE', won't load:

link_elf_obj: symbol vnet_entry_in_loopback_mask undefined

This is because it uses IN_LOOPBACK(), which in the VIMAGE case uses INET-specific symbols.

Fix by making this check conditional on #ifdef INET.

---

i'm not entirely sure this is the right fix, because i don't understand why it builds fine as part of the kernel (with 'device pf') but won't load as a module.  however, this patch does make pf work here.

### DIFF
--- a/sys/netpfil/pf/pf.c
+++ b/sys/netpfil/pf/pf.c
@@ -7956,7 +7956,10 @@ pf_dummynet_route(struct pf_pdesc *pd, struct pf_kstate *s,
 
 		if (s != NULL && s->nat_rule.ptr != NULL &&
 		    s->nat_rule.ptr->action == PF_RDR &&
-		    ((pd->af == AF_INET && IN_LOOPBACK(ntohl(pd->dst->v4.s_addr))) ||
+		    (
+#ifdef INET
+		    (pd->af == AF_INET && IN_LOOPBACK(ntohl(pd->dst->v4.s_addr))) ||
+#endif
 		    (pd->af == AF_INET6 && IN6_IS_ADDR_LOOPBACK(&pd->dst->v6)))) {
 			/*
 			 * If we're redirecting to loopback mark this packet


igalic:
cc @kprovost

sys/netpfil/pf/pf.c
```diff
@@ -7956,7 +7956,10 @@ pf_dummynet_route(struct pf_pdesc *pd, struct pf_kstate *s,
 
 		if (s != NULL && s->nat_rule.ptr != NULL &&
 		    s->nat_rule.ptr->action == PF_RDR &&
-		    ((pd->af == AF_INET && IN_LOOPBACK(ntohl(pd->dst->v4.s_addr))) ||
+		    (
+#ifdef INET
+		    (pd->af == AF_INET && IN_LOOPBACK(ntohl(pd->dst->v4.s_addr))) ||
+#endif
 		    (pd->af == AF_INET6 && IN6_IS_ADDR_LOOPBACK(&pd->dst->v6)))) {
```
bsdimp, line 7963: Don't you need similar medicine for inet6 too?

llfw: i don't think so, because `IN6_IS_ADDR_LOOPBACK` is a straightforward macro (`sys/netinet6/in6.h`):

```
#define IN6_IS_ADDR_LOOPBACK(a)         \
        ((a)->__u6_addr.__u6_addr32[0] == 0 &&  \
         (a)->__u6_addr.__u6_addr32[1] == 0 &&  \
         (a)->__u6_addr.__u6_addr32[2] == 0 &&  \
         (a)->__u6_addr.__u6_addr32[3] == ntohl(1))
```

the problem with `IN_LOOPBACK` is it pulls in INET symbols from VIMAGE by using `V_in_loopback_mask` (`sys/netinet/in.h`):

```
#define IN_LOOPBACK(i) \
    (((in_addr_t)(i) & V_in_loopback_mask) == 0x7f000000)
```

gmshake: So will `pd->af == AF_INET6` be a noop for non-INET6 build ?

llfw: as far as i'm aware, packets for an AF that isn't supported in the kernel can never enter netpfil because they don't enter the IP stack in the first place; even with bridge filtering, bridge won't run pfil for unsupported AFs.

that means the AF_INET check is no-op for INET6-only kernerls, and AF_INET6 check is no-op for INET-only builds.

but i'm not an except on this code...

bsdimp:
Looks good... last call...

bsdimp:
landed

PR #1158, opened 2024-04-08
sys/netipsec: fix IPSEC_SUPPORT for non-INET kernels

llfw:
The functions ipsec_kmod_udp_input() and ipsec_kmod_udp_pcbctl() are used by netinet6 for IPSEC_SUPPORT, but are guarded behind #ifdef INET.

Since neither of these require INET, remove the guard so they're built even without INET.


bsdimp:
merged

PR #1159, opened 2024-04-09
sys/net/if_bridge: support non-INET kernels

llfw:


sys/net/if_bridge.c
```diff
@@ -3556,8 +3557,8 @@ bridge_pfil(struct mbuf **mp, struct ifnet *bifp, struct ifnet *ifp, int dir)
 		}
 
 		/* Recalculate the ip checksum. */
-		ip = mtod(*mp, struct ip *);
-		hlen = ip->ip_hl << 2;
+		struct ip *ip = mtod(*mp, struct ip *);
+		int hlen = ip->ip_hl << 2;
 		if (hlen < sizeof(struct ip))
```
bsdimp, line 3562: These should be at the start of the block. Or at least declared there (make the case ETHERTYPE_IP: start a new block).

llfw:
updated for C89 style.

> I was concerned about #define abuse initially

i did spend a while thinking about the best way to do this and i couldn't come up with a better method.  there is a lot of code in src (not just if_bridge) which does things like:

```
if (AF_INET code...
#ifdef INET6
    || AF_INET6 code...
#endif
    ) {
  ...
```

and it's really awkward to convert this to support INET6-only kernels.  in this case i thought the additional macros were the least awful solution, but i'm open to other suggestions.

sys/net/if_bridge.c
```diff
@@ -3470,6 +3470,7 @@ bridge_pfil(struct mbuf **mp, struct ifnet *bifp, struct ifnet *ifp, int dir)
 	}
 
 	/* Run the packet through pfil before stripping link headers */
+#ifdef INET
```
markjdb, line 3473: Why is link-layer filtering conditional on INET? I'm not too familiar with this bit of the code, but it looks wrong.

sys/net/if_bridge.c
```diff
@@ -3643,6 +3650,7 @@ bridge_pfil(struct mbuf **mp, struct ifnet *bifp, struct ifnet *ifp, int dir)
  * XXX Right now we update ipstat but not
  * XXX csum_counter.
  */
+#ifdef INET
```
markjdb, line 3653: IMO the ifdef should come before the comment which applies to the conditionally compiled code.

markjdb (commented):
Broadly seems ok to me. I had a couple of inline comments.

I'm all for not having preprocessor conditionals in the middle of expressions. :)

sys/net/if_bridge.c
llfw: if we're an INET6-only kernel and the packet is an IPv4 packet, or vice versa, it will already be handled by the previous check on line 3448, which either drops the packet or unconditionally accepts it based on the value of `net.link.bridge.pfil_onlyip`, and we should never get here.

however, it does seem like there might be a pre-existing bug here when an INET6 kernel won't pass IPv6 packets to the filter.

i'm not sure under what circumstances `PFIL_HOOKED_OUT(V_link_pfil_head)` can be true while `PFIL_HOOKED_OUT(V_inet6_pfil_head)` is false -- perhaps this should be replaced by a check for `PFIL_HOOKED_OUT_ANY`?

sys/net/if_bridge.c
llfw: fixed.

sys/net/if_bridge.c
llfw: i tested this code:

```
        if (PFIL_HOOKED_OUT_ANY && V_pfil_ipfw != 0 && dir == PFIL_OUT && ifp != NULL) {
                switch (pfil_mbuf_out(V_link_pfil_head, mp, ifp, NULL)) {
                case PFIL_DROPPED:
                        return (EACCES);
                case PFIL_CONSUMED:
                        return (0);
                }
        }
```

and it does compile, but since i don't have any systems using bridge filtering, i can't really test it.

glebius:
I like the proposed style more than what we got, like the example given:

> ```
> if (AF_INET code...
> #ifdef INET6
>     || AF_INET6 code...
> #endif
>     ) {
>   ...
> ```

I didn't do a thorough review, but change looks good to me. I'd be glad to see it pushed. Thanks!

sys/net/if_bridge.c
```diff
@@ -134,6 +134,25 @@
...
+
+#ifdef INET
+#define	PFIL_HOOKED_IN_INET	PFIL_HOOKED_IN(V_inet_pfil_head)
+#define	PFIL_HOOKED_OUT_INET	PFIL_HOOKED_OUT(V_inet_pfil_head)
+#else
+#define	PFIL_HOOKED_IN_INET	false
+#define	PFIL_HOOKED_OUT_INET	false
+#endif
+
+#define PFIL_HOOKED_IN_ANY	(PFIL_HOOKED_IN_INET6 || PFIL_HOOKED_IN_INET)
```
gmshake, line 153: Actually PFIL can operate on link-layer hooks, so I'd prefer `PFIL_HOOKED_IN_46` rather than `PFIL_HOOKED_IN_ANY`.

llfw: fixed (i hope -- had some git issues locally...)

bsdimp:
Need to verify the latest patches to see if we need it.

sys/net/if_bridge.c
markjdb: I have no good way to test this at the moment either. I'd prefer to see this patch split into two: one which adds preprocessor symbols that let us avoid the ugly inline ifdefs, and one for the rest of the change. The former seems uncontroversial and a positive change, but I am not sure about the latter.

bsdimp:
I like @markjdb 's idea of splitting this. That's the change I'm flagging.

llfw:
the uncontroversial changes have been moved #1191.  i'll update/rebase this PR once that's merged.

llfw:
#1191 was merged (thanks Mark) so i've rebased this on current main.

based on previous discussion i believe this is the correct logic for `bridge_pfil()` (and my previous patch was wrong) but i've only determined this by code inspection, not testing.  i'll bring up a local VM here with a couple of vnet jails to test this properly.

bsdimp:
This looks good to me... @glebius @gmshake What do you think? I'll land this if there's no objection soon.

glebius:
LGTM

sys/net/if_bridge.c
```diff
@@ -403,14 +403,19 @@ static int	bridge_ioctl_sproto(struct bridge_softc *, void *);
...
+#endif /* INET */
+
 #ifdef INET6
 static int	bridge_ip6_checkbasic(struct mbuf **mp);
 #endif /* INET6 */
-static int	bridge_fragment(struct ifnet *, struct mbuf **mp,
-		    struct ether_header *, int, struct llc *);
-static void	bridge_linkstate(struct ifnet *ifp);
-static void	bridge_linkcheck(struct bridge_softc *sc);
+
```
gmshake, line 418: No need for extra blank.

sys/net/if_bridge.c
```diff
@@ -403,14 +403,19 @@ static int	bridge_ioctl_sproto(struct bridge_softc *, void *);
 static int	bridge_ioctl_stxhc(struct bridge_softc *, void *);
 static int	bridge_pfil(struct mbuf **, struct ifnet *, struct ifnet *,
 		    int);
+static void	bridge_linkstate(struct ifnet *ifp);
+static void	bridge_linkcheck(struct bridge_softc *sc);
```
gmshake, line 407: These two prototype can be untouched, so the diff is smaller.

sys/net/if_bridge.c
```diff
@@ -403,14 +403,19 @@ static int	bridge_ioctl_sproto(struct bridge_softc *, void *);
...
+
+#ifdef INET
 static int	bridge_ip_checkbasic(struct mbuf **mp);
+static int	bridge_fragment(struct ifnet *, struct mbuf **mp,
+		    struct ether_header *, int, struct llc *);
+#endif /* INET */
+
```
gmshake, line 414: No need for extra blank.

sys/net/if_bridge.c
```diff
@@ -3728,6 +3747,7 @@ bridge_ip_checkbasic(struct mbuf **mp)
 	*mp = m;
 	return (-1);
 }
+#endif
```
gmshake, line 3750: The `#ifdef #endif` block is long, and I think comment `/* INET */` after `#endif` will improve readability.

sys/net/if_bridge.c
```diff
@@ -3859,6 +3880,7 @@ bridge_fragment(struct ifnet *ifp, struct mbuf **mp, struct ether_header *eh,
 	}
 	return (error);
 }
+#endif
```
gmshake, line 3883: Ditto.

gmshake (approved):
Looks good to me ( but for some nits ).

gmshake:
Oops, it has been merged before my review. Anyway still good :)

emaste:
> Oops, it has been merged before my review. Anyway still good :)

Post-commit review is fine too :)

I agree with your notes, do you want to just make those changes in the tree?

gmshake:
> > Oops, it has been merged before my review. Anyway still good :)
> 
> Post-commit review is fine too :)
> 
> I agree with your notes, do you want to just make those changes in the tree?

Committed in 73585176ffd84c13d68cad67c2ca81643f09075c .

bsdimp:
This can be closed per subitter

gmshake:
Hi @bsdimp , I'd like to MFC this to stable branches. Do you have plan to do that ?

bsdimp:
Just this one orvthe series? I have no plans at tge moment... but i could make plans with lists of hashes.. or if you have a commit bit, you can.

gmshake:
> Just this one orvthe series? 

The series indeed.

> I have no plans at tge moment... but i could make plans with lists of hashes.. or if you have a commit bit, you can.

OK, I'll do that.

PR #1160, opened 2024-04-10
Fix ahc(4) in target mode

hpvb:
This reverts part of 5e63cdb45, as I could only test with AIC-7890 and AIC-7892 cards (AHA-2940U2W and ASC-29160N) I only reverted the changes in the aic7xxx.c driver and left aic79xx.c and the other cards touched by the change alone.

With this change when trying to use ctl(4) the following error occurs when trying to do anything but a trivially small I/O:

ahc0: Connection stuck awaiting busfree or Identify Msg.

This then crashes the card and will eventually lead to a panic():

panic: Timed-out target SCB but bus idle

I have tested this change on the aforementioned cards as targets on x86_64 and aarch64 (through qemu) and I get wire-speeds with both Linux and FreeBSD as the initiator through both an AHA-2940U2W and an LSI 53c1030.

### DIFF
--- a/sys/dev/aic7xxx/aic7xxx.c
+++ b/sys/dev/aic7xxx/aic7xxx.c
@@ -7794,9 +7794,9 @@ ahc_handle_target_cmd(struct ahc_softc *ahc, struct target_cmd *cmd)
 		/* Tag was included */
 		atio->tag_action = *byte++;
 		atio->tag_id = *byte++;
-		atio->ccb_h.flags |= CAM_TAG_ACTION_VALID;
+		atio->ccb_h.flags = CAM_TAG_ACTION_VALID;
 	} else {
-		atio->ccb_h.flags &= ~CAM_TAG_ACTION_VALID;
+		atio->ccb_h.flags = 0;
 	}
 	byte++;
 


bsdimp:
So "It works really well with this patch" is necessary, but not sufficient. Can you explain why we need to clear all the other flags at the point in code you changed? What makes it work?

Also, aic(4) is an old driver. This is a ahc(4), so the commit message needs to be tweaked a little to name the proper driver.

hpvb:
@bsdimp I fixed the commit message, sorry for the oversight.

As for why it works with clearing all the flags, I arrived at this solution by trying various versions of FreeBSD until I hit one that works with /usr/share/examples/scsi_target. When I identified that I ported the driver to main and reverted changes until I was left with just this change.

This was changed in 2017 apparently and I don't think ahc(4) in target mode has worked since. If you'd like I could try to figure out which flag(s) break the controller. But as far as I can see I fixed a regression?

I don't know a lot about cam(4), if you have suggestions on how I should go about figuring out why this fixes the regression I'd be happy to do that though.

amotin:
First let me say that I am impressed by somebody using parallel SCSI target mode in 2024. ;)

Second, I agree with @bsdimp that revert is not necessary a right fix.  Looking on the code, I see that driver touches not so many bits of ccb_h.flags.  One of the bits that I see can be set, but never cleared by ahc_handle_target_cmd() is CAM_DIS_DISCONNECT.  It could definitely affect operation if the bit was not cleared after some previous command.  Could instead of reverting we try this patch:

`diff --git a/sys/dev/aic7xxx/aic7xxx.c b/sys/dev/aic7xxx/aic7xxx.c
index 4b62f76b9cf8..8d000d44f302 100644
--- a/sys/dev/aic7xxx/aic7xxx.c
+++ b/sys/dev/aic7xxx/aic7xxx.c
@@ -7843,7 +7843,8 @@ ahc_handle_target_cmd(struct ahc_softc *ahc, struct target_cmd *cmd)
                ahc->pending_device = lstate;
                aic_freeze_ccb((union ccb *)atio);
                atio->ccb_h.flags |= CAM_DIS_DISCONNECT;
-       }
+       } else
+               atio->ccb_h.flags &= ~CAM_DIS_DISCONNECT;
        xpt_done((union ccb*)atio);
        return (0);
 }

mav@srv:/sources/FreeBSD/head/sys/dev/aic7xxx% git diff aic7xxx.c
diff --git a/sys/dev/aic7xxx/aic7xxx.c b/sys/dev/aic7xxx/aic7xxx.c
index 4b62f76b9cf8..8d000d44f302 100644
--- a/sys/dev/aic7xxx/aic7xxx.c
+++ b/sys/dev/aic7xxx/aic7xxx.c
@@ -7843,7 +7843,8 @@ ahc_handle_target_cmd(struct ahc_softc *ahc, struct target_cmd *cmd)
                ahc->pending_device = lstate;
                aic_freeze_ccb((union ccb *)atio);
                atio->ccb_h.flags |= CAM_DIS_DISCONNECT;
-       }
+       } else
+               atio->ccb_h.flags &= ~CAM_DIS_DISCONNECT;
        xpt_done((union ccb*)atio);
        return (0);
 }
`

hpvb:
> First let me say that I am impressed by somebody using parallel SCSI target mode in 2024. ;)

It's much better than trying to work with spinning disks in 2024! :D You'd be amused to know that I'm trying to get this to work ultimately with a raspberry pi, it's almost working. I can use ahc(4) on an rpi4 with some fixes to the pcib driver, but in target mode it doesn't work yet. It seems to be some kind of dma window issue but I'm still trying to work that out.

> Could instead of reverting we try this patch:

That doesn't work, with this change target mode stops working entirely, the Initiator now doesn't see any devices on the SCSI bus. But I think you pointed me in the right direction! I will try to see if just clearing that bit fixes the problem, and then probably try to work out where it really should be cleared.

amotin:
@hpvb I'll commit the patch myself, if you say it works.

amotin:
https://cgit.freebsd.org/src/commit/?id=f1e4c095777db088645008848b3bfb87146c13ec

hpvb:
@amotin thank you! I'm glad we're keeping this important technology alive :)

PR #1162, opened 2024-04-10
sys/netinet/cc: Switch from deprecated random() to prng32()

hhartzer:
Related: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=277655

Signed-off-by: henrichhartzer@tuta.io

### DIFF
--- a/sys/netinet/cc/cc_cdg.c
+++ b/sys/netinet/cc/cc_cdg.c
@@ -57,6 +57,7 @@
 #include <sys/malloc.h>
 #include <sys/module.h>
 #include <sys/queue.h>
+#include <sys/prng.h>
 #include <sys/socket.h>
 #include <sys/socketvar.h>
 #include <sys/sysctl.h>
@@ -507,7 +508,8 @@ cdg_cong_signal(struct cc_var *ccv, ccsignal_t signal_type)
 static inline int
 prob_backoff(long qtrend)
 {
-	int backoff, idx, p;
+	int backoff, idx;
+	uint32_t p;
 
 	backoff = (qtrend > ((MAXGRAD * V_cdg_exp_backoff_scale) << D_P_E));
 
@@ -519,8 +521,8 @@ prob_backoff(long qtrend)
 			idx = qtrend;
 
 		/* Backoff probability proportional to rate of queue growth. */
-		p = (INT_MAX / (1 << EXP_PREC)) * probexp[idx];
-		backoff = (random() < p);
+		p = (UINT32_MAX / (1 << EXP_PREC)) * probexp[idx];
+		backoff = (prng32() < p);
 	}
 
 	return (backoff);
--- a/sys/netinet/cc/cc_chd.c
+++ b/sys/netinet/cc/cc_chd.c
@@ -58,6 +58,7 @@
 #include <sys/limits.h>
 #include <sys/malloc.h>
 #include <sys/module.h>
+#include <sys/prng.h>
 #include <sys/queue.h>
 #include <sys/socket.h>
 #include <sys/socketvar.h>
@@ -85,8 +86,8 @@
  */
 #define	CC_CHD_DELAY	0x02000000
 
-/* Largest possible number returned by random(). */
-#define	RANDOM_MAX	INT_MAX
+/* Largest possible number returned by prng32(). */
+#define	RANDOM_MAX	UINT32_MAX
 
 static void	chd_ack_received(struct cc_var *ccv, ccsignal_t ack_type);
 static void	chd_cb_destroy(struct cc_var *ccv);
@@ -159,9 +160,9 @@ chd_window_decrease(struct cc_var *ccv)
 static __inline int
 should_backoff(int qdly, int maxqdly, struct chd *chd_data)
 {
-	unsigned long p, rand;
+	uint32_t rand, p;
 
-	rand = random();
+	rand = prng32();
 
 	if (qdly < V_chd_qthresh) {
 		chd_data->loss_compete = 0;
--- a/sys/netinet/cc/cc_hd.c
+++ b/sys/netinet/cc/cc_hd.c
@@ -59,6 +59,7 @@
 #include <sys/limits.h>
 #include <sys/malloc.h>
 #include <sys/module.h>
+#include <sys/prng.h>
 #include <sys/queue.h>
 #include <sys/socket.h>
 #include <sys/socketvar.h>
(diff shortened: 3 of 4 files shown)


hhartzer:
Hoping that someone with familiarity of these components can review this. I feel like it looks like an innocent enough set of changes, but I don't want to get something wrong here.

gmshake:
CC @tuexen

bsdimp:
Also cc @cemeyer @amotin @markjdb 

These seem fine, but it looks like the range is changing for these (the range is accounted for, but I want more eyes on it).

sys/netinet/cc/cc_cdg.c
```diff
@@ -519,8 +520,8 @@ prob_backoff(long qtrend)
 			idx = qtrend;
 
 		/* Backoff probability proportional to rate of queue growth. */
-		p = (INT_MAX / (1 << EXP_PREC)) * probexp[idx];
-		backoff = (random() < p);
+		p = (UINT_MAX / (1 << EXP_PREC)) * probexp[idx];
+		backoff = (prng32() < p);
```
amotin, line 524: Why p is still int, not uint32_t or uint_t?

sys/netinet/cc/cc_chd.c
```diff
@@ -161,7 +162,7 @@ should_backoff(int qdly, int maxqdly, struct chd *chd_data)
 {
 	unsigned long p, rand;
 
-	rand = random();
+	rand = prng32();
 
 	if (qdly < V_chd_qthresh) {
```
amotin, line 167: I just wonder whether should_backoff() arguments here and below should really be signed, whether they makes sense and if so, whether it is handled right.

sys/netinet/cc/cc_cdg.c
hhartzer: I do think that's a mistake. Should I do `unit32_t` for it?

sys/netinet/cc/cc_chd.c
hhartzer: I also am not sure why these were signed. I don't know hardly a thing about the code in question. Was just trying to take a stab at removing the deprecated `random()` call where possible.

hhartzer:
I can update this but would prefer a little bit more guidance about how exactly it should be done, if that's possible.

bsdimp:
@amotin I think @hhartzer needs some help with your suggestion

sys/netinet/cc/cc_cdg.c
amotin: Yes.  I'd say p should be unsigned for correct comparison, and it should have the same range for the probability to work.  So there should be `uint32_t` and `UINT32_MAX`.

sys/netinet/cc/cc_chd.c
amotin: I don't know this code either, so it was just a thought.  What you should to though is use UINT32_MAX for RANDOM_MAX (or just inline it) and uint32_t for p.  rand variable I would just remove, inlining the prng32() call.

sys/netinet/cc/cc_cdg.c
hhartzer: Thank you for the feedback! I've adjusted `p` accordingly.

sys/netinet/cc/cc_chd.c
hhartzer: Thanks! I did most of what you suggested, but I found that the rand variable was referenced twice, so it seemed necessary to keep it assigned. Is that alright?

hhartzer: Oh, looks like RANDOM_MAX is still defined in the file. Should I remove it or reference it instead?

sys/netinet/cc/cc_cdg.c
```diff
@@ -308,7 +309,7 @@ cdg_cb_init(struct cc_var *ccv, void *ptr)
 	cdg_data->queue_state = CDG_Q_UNKNOWN;
 	cdg_data->maxrtt_in_rtt = 0;
 	cdg_data->maxrtt_in_prevrtt = 0;
-	cdg_data->minrtt_in_rtt = INT_MAX;
+	cdg_data->minrtt_in_rtt = UINT_MAX;
```
amotin, line 312: minrtt_in_rtt is just int, this change makes no sense to me, same as one below.

sys/netinet/cc/cc_cdg.c
amotin: Thanks. But why did you leave UINT_MAX in its assignment instead of UINT32_MAX I proposed?

sys/netinet/cc/cc_chd.c
amotin: > Is that alright?

Sure.  I've just missed the other use.  Sorry.

Either use RANDOM_MAX everywhere, or remove everywhere (in the two files for consistency).  It is used only in the same small function as prng32(), so it is a bit obvious, but a bit confusing being declared many screens away.

sys/netinet/cc/cc_cdg.c
hhartzer: Oops! I am not sure why I did that. Good catch.

sys/netinet/cc/cc_cdg.c
hhartzer: My bad! I think I missed that.

sys/netinet/cc/cc_cdg.c
```diff
@@ -637,7 +639,7 @@ cdg_ack_received(struct cc_var *ccv, ccsignal_t ack_type)
 		}
 
 		cdg_data->minrtt_in_prevrtt = cdg_data->minrtt_in_rtt;
-		cdg_data->minrtt_in_rtt = INT_MAX;
+		cdg_data->minrtt_in_rtt = UINT_MAX;
```
amotin, line 642: "The one below"

sys/netinet/cc/cc_chd.c
```diff
@@ -85,8 +86,8 @@
  */
 #define	CC_CHD_DELAY	0x02000000
 
-/* Largest possible number returned by random(). */
-#define	RANDOM_MAX	INT_MAX
+/* Largest possible number returned by prng32(). */
+#define	RANDOM_MAX	RANDOM_MAX
```
amotin, line 90: Cool define! ;)

sys/netinet/cc/cc_cdg.c
hhartzer: Oops!

sys/netinet/cc/cc_chd.c
hhartzer: Today is not my day! Thank you for being patient with me.

amotin (approved):

PR #1163, opened 2024-04-11
ec2: homedir bug fixed in cloud-init 24.1.4

igalic:
This bug fix is due to be released in Quarterly:

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=275896

While here, overwrite the `doas` stanza, which needs to correspond to the user that we are creating.


igalic:
cc @cperciva

cperciva:
Merged, thanks!

PR #1164, opened 2024-04-11
Add support for Intel Atom S1200 UART

hhartzer:
PR: 278316

Signed-off-by: Henrich Hartzer <henrichhartzer@tuta.io>

### DIFF
--- a/sys/dev/uart/uart_bus_pci.c
+++ b/sys/dev/uart/uart_bus_pci.c
@@ -135,6 +135,8 @@ static const struct pci_id pci_ns8250_ids[] = {
 { 0x1d0f, 0x8250, 0x1d0f, 0, "Amazon PCI serial device", 0x10 },
 { 0x1fd4, 0x1999, 0x1fd4, 0x0001, "Sunix SER5xxxx Serial Port", 0x10,
 	8 * DEFAULT_RCLK },
+{ 0x8086, 0x0c5f, 0xffff, 0, "Atom Processor S1200 UART",
+	0x10 | PCI_NO_MSI },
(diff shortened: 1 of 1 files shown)


hhartzer:
This did not work until I disabled MSI. If this looks good, I would appreciate it if it could be backported to stable/14.

Thank you!

bsdimp:
Ideally, we'd want to know *WHY* it didn't work with MSI enabled. Did you discover any erratum or other basis for disabling msi?

hhartzer:
It just didn't work for me with MSI enabled. I could either disable MSI or set `debug.uart_force_poll` to `1` to get it going. There were no BIOS options around this.

I imagine this isn't going to be used very commonly. I do think it's not going to hurt much, performance wise, for the typical applications to not use MSI.

There is only one other driver in there that doesn't use MSI.

I do agree that ideally this would be better understood, and if someone else has the same kind of hardware I'd encourage them to test and I'd be happy to check patches on my hardware as well. I've already put several hours into this, (largely because I've never worked around this before) so I'd prefer not to test a whole lot other permutations unless they're straight forward.

bsdimp:
Great! Wished for better, but that's enough

bsdimp:
Thanks for the explanation and contribution. Landed.

hhartzer:
Thank you so much!! I appreciate you merging this!

Might it be possible to backport this to stable/14? It'd be nice if I could use it in 14.1, when it comes in a few months, without recompiling.

bsdimp:
Queued to my mfc queue. I'll push it in Monday or Tuesday.

PR #1165, opened 2024-04-12
usr.bin/units: use else..if to avoid calling the next branch

rilysh:
Even if the first branch succeeds, next time it will still check for the second branch (which will be false) as the first one was true. Add an else..if statement to address this.

Copy of https://github.com/freebsd/freebsd-src/pull/1005 (without the first one).

### DIFF
--- a/usr.bin/units/units.c
+++ b/usr.bin/units/units.c
@@ -497,8 +497,7 @@ lookupunit(const char *unit)
 			}
 		}
 		free(copy);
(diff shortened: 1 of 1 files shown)


bsdimp:
merged. Thanks

bsdimp:
I'll also note that this is on the edge of being a major enough change to be worthwhile, given how tight time for PR landing is.

PR #1167, opened 2024-04-14
src.libnames.mk: fix LIBPFCTL definition

llfw:
Following the convention used in the rest of this file, ${LIBPFCTLDIR} should refer to the directory, and ${LIBPFCTL} to the library itself.

Instead, both values were assigned to ${LIBPFCTL}, and ${LIBPFCTLDIR} was not set at all.

This appears to be a simple typo and not a deliberate choice, so fix it by assigning the directory name to ${LIBPFCTLDIR} instead.


llfw:
committed in 3075939da41a22fc6a4ec592d78f582a711b5e79, thanks!

PR #1168, opened 2024-04-14
rescue: remove routed, rtquery

llfw:
These binaries are used to implement RIP/RIPng.  While it used to be fairly common for hosts to run RIP for routing, nowadays this is a very niche protocol and is not commonly used.

This reduces the size of rescue by 99072 bytes.


llfw:
closing this one as it seems like the proposal to remove routed entirely is going through, making this overcome by events.

PR #1169, opened 2024-04-14
rescue: add ipfw, pfctl

llfw:
Rationale for this change:

- ipf is already there

- if the kernel is configured with a packet filter which drops packets by default, pfctl or ipfw will be required in the rescue environment to make the network functional.

- rescue's stated purpose is to be useful for small/embedded systems (and is also quite useful for small jails); a rescue-based router would want these binaries.

On amd64, this increases the size of rescue from 17346200 to 17907248, or 561048 bytes (3.2%).

### DIFF
--- a/rescue/rescue/Makefile
+++ b/rescue/rescue/Makefile
@@ -119,6 +119,15 @@ CRUNCH_PROGS_sbin+= ipf
 CRUNCH_LIBS_ipf+=	${LIBIPF}
 .endif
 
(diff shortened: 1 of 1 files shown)


llfw:
this requires #1167 to be merged first or the build fails.

emaste:
Size increase seems reasonable and anyone who needs an especially small image can build with only one of pf/ipf/ipfw.

bsdimp:
merged. thanks for the cleanup

PR #1170, opened 2024-04-15
package: move hv_{vss,kvp}_daemon into hyperv-tools

llfw:


bsdimp:

I think these utils are niche enough, but I don't feel that strongly about it.

evadot:
I've likely missed those when creating the hyper-v package, good for me.

bsdimp:
Looks ready... last call.

bsdimp:
merged

PR #1171, opened 2024-04-15
package: move lpr into its own package

llfw:
several of the lp-related utilities are setuid, so this is a concrete security benefit for sites that don't need printing support.


bsdimp:
@evadot What do you think of this?

evadot:
LGTM, that's something I wanted to do a while ago but just forgot about it :)

bsdimp:
merged, forgot to tag manu on reviewed, even though he gave the lgtm. thanks for the cleanup.

PR #1172, opened 2024-04-15
package: move cron into its own package

llfw:
this allows to use functionality of cron (periodic, etc.) without having to install huge FreeBSD-utilities package. 

cc @evadot, @bapt


igalic:
@dfr might also interest you

bsdimp:

This one seems like it could go either way to me, but I lean slightly to 'cron is big enough to be its own thing' but I could easily see that it is really just another util.

evadot:
Yeah it's good for me.

bsdimp:
looks good... last call...

igalic (approved):
👍

bsdimp:
merged

PR #1173, opened 2024-04-15
package: move bsdconfig to its own package

llfw:
this moves around 1MB of files out of FreeBSD-utilities, including a lot of files in /usr/share/bsdconfig.

### DIFF
--- a/usr.sbin/bsdconfig/Makefile
+++ b/usr.sbin/bsdconfig/Makefile
@@ -22,6 +22,7 @@ SUBDIR=	console \
 SUBDIR+= examples
 .endif
 
+PACKAGE=	bsdconfig
 FILESDIR=	${LIBEXECDIR}/bsdconfig
 FILES=		USAGE
 
--- a/usr.sbin/bsdconfig/console/Makefile
+++ b/usr.sbin/bsdconfig/console/Makefile
@@ -1,6 +1,7 @@
 
 SUBDIR= include
 
+PACKAGE=	bsdconfig
 FILESDIR=	${LIBEXECDIR}/bsdconfig/080.console
 FILES=		INDEX USAGE
 
--- a/usr.sbin/bsdconfig/console/include/Makefile
+++ b/usr.sbin/bsdconfig/console/include/Makefile
@@ -1,4 +1,5 @@
 
+PACKAGE=	bsdconfig
 FILESDIR=	${LIBEXECDIR}/bsdconfig/080.console/include
 FILES=		messages.subr
 
--- a/usr.sbin/bsdconfig/diskmgmt/Makefile
+++ b/usr.sbin/bsdconfig/diskmgmt/Makefile
@@ -1,6 +1,7 @@
 
 SUBDIR= include
 
+PACKAGE=	bsdconfig
 FILESDIR=	${LIBEXECDIR}/bsdconfig/050.diskmgmt
 FILES=		INDEX USAGE
 
--- a/usr.sbin/bsdconfig/diskmgmt/include/Makefile
+++ b/usr.sbin/bsdconfig/diskmgmt/include/Makefile
@@ -1,4 +1,5 @@
 
+PACKAGE=	bsdconfig
 FILESDIR=	${LIBEXECDIR}/bsdconfig/050.diskmgmt/include
 FILES=		messages.subr
 
--- a/usr.sbin/bsdconfig/docsinstall/Makefile
+++ b/usr.sbin/bsdconfig/docsinstall/Makefile
@@ -1,6 +1,7 @@
 
 SUBDIR= include
 
+PACKAGE=	bsdconfig
 FILESDIR=	${LIBEXECDIR}/bsdconfig/020.docsinstall
 FILES=		INDEX USAGE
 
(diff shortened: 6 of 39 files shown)


bsdimp:

I don't have a good feel for this one.

evadot:
I guess it make sense but I've never really used bsdconfig.
Also iirc bsdinstall depends on bsdconfig so maybe we need to add a dep in the bsdinstall package.

evadot:
Oh also since this is a new package we need an entry in release/packages/Makefile.package

llfw:
whoops - i thought i'd added a Makefile.package entry but it's easy to forget that, i'll update this one.  that would be a nice thing for CI to check, if feasible

to add a depdendency for bsdinstall, is this just a case of creating `bsdinstall.ucl`?  i can't see any other packages which add manual dependencies to base this on.

(or, does it make sense to simply put bsdconfig into the bsdinstall package?)

llfw:
unfortunately local testing turned up an issue with this which is that service(8) requires bsdconfig at least for enable/disable:

```
# service uuidd enable
.: cannot open /usr/share/bsdconfig/common.subr: No such file or directory
```

which means FreeBSD-runtime would require a dependency on FreeBSD-bsdconfig.  but that means FreeBSD-runtime already requires FreeBSD-utilities, which seems wrong.

i'm not entirely sure what the right fix is here.

bsdimp:
I've added go-nogo to the labels until a path forward is discovered.

I'll give it a week or so and reevaluate and close if there's no path to resolution here (though as always, if a path forward is found after that a new pull request is welcome).

llfw:
yep that's fine.  fwiw i think the current situation is also wrong, this is just wrong in a different way... maybe common.subr should move somewhere else.

bsdimp:
yea, the common subroutines shouldn't create an inverted pyramid of dependencies.

evadot:
> unfortunately local testing turned up an issue with this which is that service(8) requires bsdconfig at least for enable/disable:
> 
> ```
> # service uuidd enable
> .: cannot open /usr/share/bsdconfig/common.subr: No such file or directory
> ```
> 
> which means FreeBSD-runtime would require a dependency on FreeBSD-bsdconfig. but that means FreeBSD-runtime already requires FreeBSD-utilities, which seems wrong.
> 
> i'm not entirely sure what the right fix is here.

Mhm, I can't see where service(8) needs bsdconfig

llfw:
> I can't see where service(8) needs bsdconfig

in fact, it's not service(8) but rather sysrc(8) (which is called by service):

```
BSDCFG_SHARE="/usr/share/bsdconfig"
[ "$_COMMON_SUBR" ] || . $BSDCFG_SHARE/common.subr || exit 1
[ "$_SYSRC_SUBR"  ] || f_include $BSDCFG_SHARE/sysrc.subr
```

sysrc is also in FreeBSD-utilities.

evadot:
> > I can't see where service(8) needs bsdconfig
> 
> in fact, it's not service(8) but rather sysrc(8) (which is called by service):
> 
> ```
> BSDCFG_SHARE="/usr/share/bsdconfig"
> [ "$_COMMON_SUBR" ] || . $BSDCFG_SHARE/common.subr || exit 1
> [ "$_SYSRC_SUBR"  ] || f_include $BSDCFG_SHARE/sysrc.subr
> ```
> 
> sysrc is also in FreeBSD-utilities.

Ah ok, so it's sysrc called in /etc/rc.subr
Having bsdconfig stuff removed from sysrc it really something that someone(tm) should work on :)

evadot:
Ideally sysrc(8) should also live in FreeBSD-rc

llfw:
yes, i think both sysrc(8) and service(8) should be in FreeBSD-rc as they only make sense in the context of the rc system.

that means FreeBSD-rc should have a dependency on FreeBSD-utilities, but that won't be too bad once the rest of the irrelevant stuff has moved out of -utilities; i think it's safe to assume most rc scripts will require the base utilities to work.

what do we want to do with this PR then?   i feel like bsdconfig is "logically" separate from utilities, but perhaps it doesn't make sense to have a separate package if it'll always be installed.

igalic:
> (or, does it make sense to simply put bsdconfig into the bsdinstall package?)

yes, i think that might make sense

evadot:
> yes, i think both sysrc(8) and service(8) should be in FreeBSD-rc as they only make sense in the context of the rc system.
> 
> that means FreeBSD-rc should have a dependency on FreeBSD-utilities

 No, we need to make sysrc not depends on bsdconfig.

> but that won't be too bad once the rest of the irrelevant stuff has moved out of -utilities; i think it's safe to assume most rc scripts will require the base utilities to work.
> 
> what do we want to do with this PR then? i feel like bsdconfig is "logically" separate from utilities, but perhaps it doesn't make sense to have a separate package if it'll always be installed.

evadot:
> > (or, does it make sense to simply put bsdconfig into the bsdinstall package?)
> 
> yes, i think that might make sense

No it doesn't.

llfw:
i think this should be closed for now, since rewriting sysrc to not use bsdconfig will be a separate change and needs to happen first.

bsdimp:
closed, per OP

PR #1174, opened 2024-04-16
tests: move atf_python/sys/ into the tests package

llfw:
these were already certainly meant to be in the existing tests package and were just missed by accident (since `PACKAGE=` isn't inherited by subdirs).


markjdb (approved):

markjdb:
Merged, thanks.

PR #1176, opened 2024-04-16
make usr/share/examples more packagey

llfw:
3 commits here: 1 adds the Makefile soup to allow examples to live in packages, and two commits move the pf and HAST examples to their proper home.

i can do more commits for the rest of the examples, but i wanted to check the basic idea is sound first.

cc: @evadot, @bapt, @igalic


igalic (changes requested):
this is all happening in the same file, so i don't think it's necessary to split anything.
Just do them all in one go. (and i mean all, not just three three)

llfw:
i'm happy to squash them but i prefer not to do this initially because it makes it difficult to merge into my local branch (which i run in production).

if this passes review / is okay to commit i can squash it, just lmk.

bsdimp:

evadot:
I think it's good. We will have the examples in the right package and if no matching package is present it will be in the FreeBSD-examples packages.
People can still remove all example by adding WITHOUT_EXAMPLES so it's fine.

igalic:
> I think it's good. We will have the examples in the right package and if no matching package is present it will be in the FreeBSD-examples packages. People can still remove all example by adding WITHOUT_EXAMPLES so it's fine.

i think this means: "i approve".

bsdimp:
Looks good... Last call...

bsdimp:
merged

PR #1178, opened 2024-04-16
dumpon: fix set but not used variable

inkeliz:
That is my attempt to fix one issue, when compiling with WITHOUT_OPENSSL or WITHOUT_CRYPTO:

```
/usr/home/x/freebsd/sbin/dumpon/dumpon.c:478:21: error: variable 'cipher' set but not used [-Werror,-Wunused-but-set-variable]
        int ch, error, fd, cipher
```

I'm not sure if that is the proper way to fix that.

### DIFF
--- a/sbin/dumpon/dumpon.c
+++ b/sbin/dumpon/dumpon.c
@@ -463,20 +463,23 @@ main(int argc, char *argv[])
 	struct diocskerneldump_arg ndconf, *kdap;
 	struct addrinfo hints, *res;
 	const char *dev, *pubkeyfile, *server, *client, *gateway;
-	int ch, error, fd, cipher;
+	int ch, error, fd;
 	bool gzip, list, netdump, zstd, insert, rflag;
 	uint8_t ins_idx;
+#ifdef HAVE_CRYPTO
+	int cipher = KERNELDUMP_ENC_NONE;
+#endif
 
 	gzip = list = netdump = zstd = insert = rflag = false;
 	kdap = NULL;
 	pubkeyfile = NULL;
 	server = client = gateway = NULL;
 	ins_idx = KDA_APPEND;
-	cipher = KERNELDUMP_ENC_NONE;
 
 	while ((ch = getopt(argc, argv, "C:c:g:i:k:lrs:vZz")) != -1)
 		switch ((char)ch) {
 		case 'C':
+#ifdef HAVE_CRYPTO
 			if (strcasecmp(optarg, "chacha") == 0 ||
 			    strcasecmp(optarg, "chacha20") == 0)
 				cipher = KERNELDUMP_ENC_CHACHA20;
@@ -487,6 +490,11 @@ main(int argc, char *argv[])
 				errx(EX_USAGE, "Unrecognized cipher algorithm "
 				    "'%s'", optarg);
 			break;
+#else
+			errx(EX_USAGE, "Built without crypto support, "
+			    "-C is unhandled.");
+			break;
+#endif
 		case 'c':
 			client = optarg;
 			break;


sbin/dumpon/dumpon.c
```diff
@@ -463,19 +463,25 @@ main(int argc, char *argv[])
 	struct diocskerneldump_arg ndconf, *kdap;
 	struct addrinfo hints, *res;
 	const char *dev, *pubkeyfile, *server, *client, *gateway;
-	int ch, error, fd, cipher;
+	int ch, error, fd;
 	bool gzip, list, netdump, zstd, insert, rflag;
 	uint8_t ins_idx;
+	#ifdef HAVE_CRYPTO
+	int cipher
+	#endif
```
markjdb, line 471: This should be removed, it won't compile.

sbin/dumpon/dumpon.c
```diff
@@ -463,19 +463,25 @@ main(int argc, char *argv[])
...
 
 	gzip = list = netdump = zstd = insert = rflag = false;
 	kdap = NULL;
 	pubkeyfile = NULL;
 	server = client = gateway = NULL;
 	ins_idx = KDA_APPEND;
-	cipher = KERNELDUMP_ENC_NONE;
+	#ifdef HAVE_CRYPTO
```
markjdb, line 478: Please keep `#ifdef` lines unindented.

sbin/dumpon/dumpon.c
```diff
@@ -487,6 +493,7 @@ main(int argc, char *argv[])
 				errx(EX_USAGE, "Unrecognized cipher algorithm "
 				    "'%s'", optarg);
 			break;
+		#endif
```
markjdb, line 496: We should print a more useful error message here if `-C` isn't implemented. I suggest adding `errx(EX_USAGE, "built without crypto support, -C is unhandled");` or so.

markjdb (changes requested):

inkeliz:
I think I fix that. However, while testing with `dumpon -C something`, it don't returns any error. Also, the `usage` didn't mention `-C` at all:

https://github.com/freebsd/freebsd-src/blob/7d03ec330684a4e82e6113df3f40305849e5fb6c/sbin/dumpon/dumpon.c#L74-L83

markjdb:
> I think I fix that. However, while testing with `dumpon -C something`, it don't returns any error.

The patch looks right. Are you sure you tested the right executable? What does dumpon print in this case?

> Also, the `usage` didn't mention `-C` at all:
> 
> https://github.com/freebsd/freebsd-src/blob/7d03ec330684a4e82e6113df3f40305849e5fb6c/sbin/dumpon/dumpon.c#L74-L83

That is a bug. I'd be happy to see a patch for that.

markjdb (approved):

sbin/dumpon/dumpon.c
```diff
@@ -487,6 +493,11 @@ main(int argc, char *argv[])
 				errx(EX_USAGE, "Unrecognized cipher algorithm "
 				    "'%s'", optarg);
 			break;
+#else
+                        errx(EX_USAGE, "Built without crypto support, "
+                                    "-C is unhandled.");
+			break;
```
markjdb, line 499: I missed it earlier, but please use tabs for indentation, not spaces.

sbin/dumpon/dumpon.c
```diff
@@ -463,20 +463,26 @@ main(int argc, char *argv[])
 	struct diocskerneldump_arg ndconf, *kdap;
 	struct addrinfo hints, *res;
 	const char *dev, *pubkeyfile, *server, *client, *gateway;
-	int ch, error, fd, cipher;
+	int ch, error, fd;
 	bool gzip, list, netdump, zstd, insert, rflag;
 	uint8_t ins_idx;
+#ifdef HAVE_CRYPTO
+	int cipher;
```
bsdimp, line 470: I'd be tempted to just initialize here to safe 3 lines later

markjdb (commented):
The commit's author name is "inkeliz@inkeliz.com", is it intentional? This is also the email address, so it looks strange.

bsdimp:
I think this is good, possibly with the one tweak i suggested

bsdimp:
I can land

markjdb (approved):

markjdb:
Thank you for the patch.

PR #1180, opened 2024-04-17
[NOMERGE] Test an auto-request-review github action

zxombie:
Don't merge, testing an action to add developers to GitHub reviews


zxombie:
It looks like the default public repo `secrets.GITHUB_TOKEN` is read-only. For this to work we would need to use `pull_request_target`. This only works on the base commit, i.e. something already in git. This is for security reasons as the token will have read/write access.

PR #1181, opened 2024-04-17
ng_pipe: Various improvements

gartens:
Various small improvements to `ng_pipe(4)`:

 - Fix various whitespace issues
 - Replace deprecated `random(9)` with `arc4random(9)`
 - Remove the node when all hooks are disconnected.  This is the behavior described in the man page
 - Do not panic when memory allocations fail

### DIFF
--- a/sys/netgraph/ng_pipe.c
+++ b/sys/netgraph/ng_pipe.c
@@ -641,12 +641,14 @@ ngp_rcvdata(hook_p hook, item_p item)
 	}
 
 	/* Populate the packet header */
-	ngp_h = uma_zalloc(ngp_zone, M_NOWAIT);
-	KASSERT((ngp_h != NULL), ("ngp_h zalloc failed (1)"));
 	NGI_GET_M(item, m);
-	KASSERT(m != NULL, ("NGI_GET_M failed"));
-	ngp_h->m = m;
 	NG_FREE_ITEM(item);
+	ngp_h = uma_zalloc(ngp_zone, M_NOWAIT);
+	if (ngp_h == NULL) {
+		NG_FREE_M(m);
+		return (ENOMEM);
+	}
+	ngp_h->m = m;
 
 	if (hinfo->cfg.fifo)
 		hash = 0;	/* all packets go into a single FIFO queue */
@@ -659,7 +661,11 @@ ngp_rcvdata(hook_p hook, item_p item)
 			break;
 	if (ngp_f == NULL) {
 		ngp_f = uma_zalloc(ngp_zone, M_NOWAIT);
-		KASSERT(ngp_h != NULL, ("ngp_h zalloc failed (2)"));
+		if (ngp_f == NULL) {
+			NG_FREE_M(ngp_h->m);
+			uma_zfree(ngp_zone, ngp_h);
+			return (ENOMEM);
+		}
 		TAILQ_INIT(&ngp_f->packet_head);
 		ngp_f->hash = hash;
 		ngp_f->packets = 1;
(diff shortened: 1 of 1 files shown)


bsdimp:
Authorized ci

sys/netgraph/ng_pipe.c
```diff
@@ -771,7 +771,7 @@ pipe_dequeue(struct hookinfo *hinfo, struct timeval *now) {
 		 * the original packet...
 		 */
 		if (hinfo->cfg.duplicate &&
-		    random() % 100 <= hinfo->cfg.duplicate) {
+		    arc4random() % 100 <= hinfo->cfg.duplicate) {
```
markjdb, line 779: Why arc4random() instead of prng32() or prng32_bounded()? The former is a cryptographic PRNG, which doesn't look to be what you want.

gartens: Changed as well as the other `random(9)` replacement.

sys/netgraph/ng_pipe.c
```diff
@@ -814,7 +828,7 @@ pipe_dequeue(struct hookinfo *hinfo, struct timeval *now) {
 		/* Randomly discard the frame, according to BER setting */
 		if (hinfo->cfg.ber) {
 			oldrand = rand;
-			rand = random();
+			rand = prng32() & 0x7fffffff;
```
markjdb, line 831: This calls for prng32_bounded() as well.

sys/netgraph/ng_pipe.c
```diff
@@ -976,7 +998,7 @@ ngp_modevent(module_t mod, int type, void *unused)
 		    sizeof (struct ngp_fifo)), NULL, NULL, NULL, NULL,
 		    UMA_ALIGN_PTR, 0);
 		if (ngp_zone == NULL)
-			panic("ng_pipe: couldn't allocate descriptor zone");
+			error = ENOMEM;
```
markjdb, line 1001: This check can simply be removed: uma_zcreate() never fails.

sys/netgraph/ng_pipe.c
```diff
@@ -658,7 +660,8 @@ ngp_rcvdata(hook_p hook, item_p item)
 			break;
 	if (ngp_f == NULL) {
 		ngp_f = uma_zalloc(ngp_zone, M_NOWAIT);
-		KASSERT(ngp_h != NULL, ("ngp_h zalloc failed (2)"));
+		if (ngp_f == NULL)
+			return (ENOMEM);
```
markjdb, line 664: Here you're leaking ngp_h (and maybe other things).

sys/netgraph/ng_pipe.c
gartens: Done.

sys/netgraph/ng_pipe.c
gartens: Good catch!

sys/netgraph/ng_pipe.c
gartens: A quick grep shows that there are various places where the result of `uma_zcreate()` is checked (e.g. [here](https://github.com/freebsd/freebsd-src/blob/464b1ab216d5f78054626524addf99f5c622da96/sys/dev/dpaa/if_dtsec_rm.c#L313-L316) or [here](https://github.com/freebsd/freebsd-src/blob/464b1ab216d5f78054626524addf99f5c622da96/sys/dev/wg/if_wg.c#L2994-L2996)).  How are they different?

markjdb: They aren't different, that code doesn't need to check either. It's just cargo-culting.

gartens: Removed

markjdb:
This looks good to me. Maybe @glebius could comment on the functional change (destroying the node when hooks are removed)?

glebius:
Given this behavior is documented and also it is typical for many other nodes, I believe the patch does the right thing.

markjdb:
Landed, thanks.

gartens:
Thank you for your help!

PR #1182, opened 2024-04-18
add dtb support for RPI CM4, CM4s, CM4_ioBoard

klaus4:
Since the test of PR #1179 was successful 
it should make sense to support these models.
That wasn't possible before because cm4-models crashed by pcie devmatch .

While in my tests I had manually added only bcm2711-rpi-cm4.dtb to config.txt 
without the corresponding bcm2711-rpi-cm4-io.dtb for the io-board,
I presume there normally should be no caveats with bcm2711-rpi-cm4-io.dtb .

### DIFF
--- a/release/arm64/RPI.conf
+++ b/release/arm64/RPI.conf
@@ -3,7 +3,8 @@
 #
 
 DTB_DIR="/usr/local/share/rpi-firmware"
-DTB="bcm2710-rpi-2-b.dtb bcm2710-rpi-3-b.dtb bcm2710-rpi-3-b-plus.dtb bcm2710-rpi-cm3.dtb bcm2711-rpi-4-b.dtb"
+DTB="bcm2710-rpi-2-b.dtb bcm2710-rpi-3-b.dtb bcm2710-rpi-3-b-plus.dtb bcm2710-rpi-cm3.dtb \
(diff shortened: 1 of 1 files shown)


emaste:
This looks reasonable, thank you for this. Can you please include a Signed-off-by: trailer in the commit, certifying agreement with https://developercertificate.org/.

klaus4:
accidentally 😁, will rebase

klaus4:
> .....  Can you please include a Signed-off-by: trailer in the commit....
✅  
  thank you for the tip and sorry for close/reopen- accidentally

bsdimp:
merged. thanks for the contribution... the other rpi one is waiting on ed making contact with the broadcom linux kernel engineer, if possible...

klaus4:
thank you ! 
should be noted that without #1179  the CM4-models will panic by default,
with or without physical pcie-device connected.
can be handled for CM4 users by a temporarily devmatch_enable=„NO“ in case that #1179 won`t reach the next snapshot.
Regards

bsdimp:
Should i revert until Ed can land #1179?

klaus4:
perhaps better reverting (only if Ed thinks it could last a bit longer to land #1179)  .

klaus4:
but perhaps prefer to land #1179 now ,…
while of course it`s always good contacting a broadcom engineer..

PR #1184, opened 2024-04-18
package: move remaining libraries out of -utilities

llfw:
Move all libraries currently left in FreeBSD-utilities to their own packages.

Exceptions:

- libmenuw goes into FreeBSD-clibs with the rest of ncurses.
- the various googletest libraries all go into FreeBSD-libprivategest.


llfw:
cc @bapt, @evadot, @igalic

igalic (commented):
once this spilt is complete, how many of these libraries does the FreeBSD-utilities package pull in on install?

llfw:
> once this spilt is complete, how many of these libraries does the FreeBSD-utilities package pull in on install?

most (but not all) of them.  but that's because i only addressed libraries in this PR; i intend to examine the binaries later

i can produce an exact list if you like, but this is quite difficult to do since pkg(8) offers no easy way to do that.

evadot:
I don't like that, in fact when I started doing pkgbase things the first things I did was to remove a lot of package that only shipped libs. We still have a few that I haven't cleanup (or left like that on purpose) but this shouldn't be the norm.

bapt:
I agree with @evadot here, the point of pkgbase is not to ship as many small packages as possible, but to find the right balance between a giant packages and thousands of small packages.

bsdimp:
@evadot @bapt so is that no on the idea, or no on this specific PR?

llfw:
okay, feel free to close this one, and i'll come up with a better solution.  (i do think most of these libraries shouldn't be in -utilities, but we can find a better place for them.)

bsdimp:
Closing... this isn't quite right.

PR #1185, opened 2024-04-18
alc(4): disable MSI-X by default on Killer cards

llfw:
Several users with alc(4)-based "Killer" Ethernet cards have reported issues with this driver not passing traffic, which are solved by disabling MSI-X using the provided tunable.

To work around this issue, disable MSI-X by default on this card.

This is done by having msix_disable default to 2, which means "auto-detect".  The user can still override this to either 0 or 1 as desired.

Since these are slow (1Gbps) Ethernet ICs used in low-end systems, it's unlikely this will cause any practical performance issues; on the other hand, the card not working by default likely causes issues for many new FreeBSD users who find their network port doesn't work and have no idea why.

PR:		230807
MFC after:	1 week

---

cc @emaste as you seem interested in user onboarding issues

### DIFF
--- a/sys/dev/alc/if_alc.c
+++ b/sys/dev/alc/if_alc.c
@@ -91,8 +91,14 @@ MODULE_DEPEND(alc, miibus, 1, 1, 1);
 
 /* Tunables. */
 static int msi_disable = 0;
-static int msix_disable = 0;
 TUNABLE_INT("hw.alc.msi_disable", &msi_disable);
+
+/*
+ * The default value of msix_disable is 2, which means to decide whether to
+ * enable MSI-X in alc_attach() depending on the card type.  The operator can
+ * set this to 0 or 1 to override the default.
+ */
+static int msix_disable = 2;
 TUNABLE_INT("hw.alc.msix_disable", &msix_disable);
 
 /*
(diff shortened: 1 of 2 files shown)


sys/dev/alc/if_alc.c
```diff
@@ -1410,6 +1416,14 @@ alc_attach(device_t dev)
 	case DEVICEID_ATHEROS_E2400:
 	case DEVICEID_ATHEROS_E2500:
 		sc->alc_flags |= ALC_FLAG_E2X00;
+
+		/*
+		 * Disable MSI-X by default on Killer devices, since this is
+		 * reported by several users to not work well.
+		 * */
```
bsdimp, line 1423: Please fix

emaste:
Thanks for the CC, this makes sense to me. Do we know off hand if other operating systems (e.g. Linux or OpenBSD) do this?

llfw:
on Linux the bug isn't present, but i spent several hours comparing their driver (alx) to ours and couldn't work out what they're doing differently.  i did mail the listed maintainer of the Linux driver to ask for documentation on the chip (Qualcomm is not very forthcoming with this) but no luck.

NetBSD from what i remember doesn't support MSI-X on this card.  i'm not sure what OpenBSD does.

i've fixed the comment typo.

bsdimp (approved):
Looks good to me...

bsdimp:
Seems ready. Last call ...

bsdimp:
merged

PR #1186, opened 2024-04-19
ifconfig tunnelfib is implemented in wg(4)

gshapiro:
It appears from the wg(4) source code that ifconfig's tunnelfib parameter is available so adding it to the man page.

### DIFF
--- a/sbin/ifconfig/ifconfig.8
+++ b/sbin/ifconfig/ifconfig.8
@@ -549,9 +549,10 @@ A FIB
 .Ar fib_number
 is assigned to all packets encapsulated by tunnel interface, e.g.,
 .Xr gif 4 ,
-.Xr gre 4
+.Xr gre 4 ,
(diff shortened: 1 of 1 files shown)


emaste:
CC @kevans91

gmshake (approved):
Looks good to me.

llfw:
if this line is supposed to enumerate all drivers which support the option, should it say "i.e." rather than "e.g."?

gmshake:
> if this line is supposed to enumerate all drivers which support the option,

If all ( in-tree ) drivers are enumerated, good.
If not, acceptable.

> should it say "i.e." rather than "e.g."?

So I think "e.g." is appropriate here.

kevans91:
This is indeed expected to work; thanks! (LGTM)

bsdimp:
Seems ready. Last call...

bsdimp:
merged

PR #1187, opened 2024-04-20
nanobsd: Set a GPT label for EFI and cfg partitions

jlduran:
This change is dependent on https://github.com/freebsd/poudriere/pull/1143, just to keep things similar (or it should be the other way around).


jlduran:
Small rant on the topic: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=278480 .

bsdimp (approved):
I love this... Maybe we could change bsdinstall to use efiboot, efiboot1, etc.
And I'd love to see a change to vmimage too. It's new enough we don't have a huge legacy to be compatible with.

bsdimp:
On second thought, maybe we should standardize on `efiboot0`

bsdimp:
landed. Thanks for the submission.

PR #1189, opened 2024-04-21
ahc(4) clean up old Linux defines

hpvb:
Linux removed theirs starting in 2018 in commit:
"scsi: aic7xxx: Fix build using bare-metal toolchain"

these now serve only to confuse.

### DIFF
--- a/sys/dev/aic7xxx/aic79xx.c
+++ b/sys/dev/aic7xxx/aic79xx.c
@@ -42,16 +42,9 @@
  * $Id: //depot/aic7xxx/aic7xxx/aic79xx.c#246 $
  */
 
-#ifdef __linux__
-#include "aic79xx_osm.h"
-#include "aic79xx_inline.h"
-#include "aicasm/aicasm_insformat.h"
-#else
-#include <sys/cdefs.h>
 #include <dev/aic7xxx/aic79xx_osm.h>
 #include <dev/aic7xxx/aic79xx_inline.h>
 #include <dev/aic7xxx/aicasm/aicasm_insformat.h>
-#endif
 
 /******************************** Globals *************************************/
 struct ahd_softc_tailq ahd_tailq = TAILQ_HEAD_INITIALIZER(ahd_tailq);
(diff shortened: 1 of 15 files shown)


sys/dev/aic7xxx/aic79xx.c
```diff
@@ -42,16 +42,10 @@
  * $Id: //depot/aic7xxx/aic7xxx/aic79xx.c#246 $
  */
 
-#ifdef __linux__
-#include "aic79xx_osm.h"
-#include "aic79xx_inline.h"
-#include "aicasm/aicasm_insformat.h"
-#else
 #include <sys/cdefs.h>
```
bsdimp, line 45: Do we still need sys/cdefs.h here?

sys/dev/aic7xxx/aic79xx_pci.c
```diff
@@ -43,14 +43,9 @@
  * $Id: //depot/aic7xxx/aic7xxx/aic79xx_pci.c#88 $
  */
 
-#ifdef __linux__
-#include "aic79xx_osm.h"
-#include "aic79xx_inline.h"
-#else
 #include <sys/cdefs.h>
```
bsdimp, line 46: Same

sys/dev/aic7xxx/aic7xxx.c
```diff
@@ -42,16 +42,10 @@
  * $Id: //depot/aic7xxx/aic7xxx/aic7xxx.c#155 $
  */
 
-#ifdef __linux__
-#include "aic7xxx_osm.h"
-#include "aic7xxx_inline.h"
-#include "aicasm/aicasm_insformat.h"
-#else
 #include <sys/cdefs.h>
```
bsdimp, line 45: Same

sys/dev/aic7xxx/aic7xxx_93cx6.c
```diff
@@ -64,16 +64,10 @@
  *   bit to be sent from the chip.
  */
 
-#ifdef __linux__
-#include "aic7xxx_osm.h"
-#include "aic7xxx_inline.h"
-#include "aic7xxx_93cx6.h"
-#else
 #include <sys/cdefs.h>
```
bsdimp, line 67: Same

amotin (approved):
I have no strong opinion. My only interest in the driver was its rare target mode support.  I doubt this driver sync with Linux matters much nowadays, and as I understand from the description it may not be there already any way.

emaste:
Indeed, Linux has removed the FreeBSD defines and there is no need for us to keep Linux defines -- for quite some time neither of us directly shared this file.

Pedantically the cdefs change isn't part of cleaning up old Linux defines but reasonable to include with this.

bsdimp:
landed. Thanks!

PR #1190, opened 2024-04-21
ahc(4)/ahd(4): target mode: cancel outstanding AIOs and INOTs

hpvb:
When disabling a lun there can still be outstanding AIOs and INOTs, when this happens previously the lun would just fail to disable and trying to re-use the lun would break the card.

isp(4) in target mode does the same thing when disabling a lun, in testing this allows re-starting of ctld(8) with connected initiators and allows initiators to gracefully resume afterwards.

~I currently don't have any ahd(4) compatible hardware, but I purchased some. Once I have those, I will see if the equivalent changes can and should be made there too.~

I just applied to ahd(4) as well, the code is pretty much identical.

### DIFF
--- a/sys/dev/aic7xxx/aic79xx.c
+++ b/sys/dev/aic7xxx/aic79xx.c
@@ -9836,6 +9836,7 @@ ahd_handle_en_lun(struct ahd_softc *ahd, struct cam_sim *sim, union ccb *ccb)
 	struct	   ahd_tmode_tstate *tstate;
 	struct	   ahd_tmode_lstate *lstate;
 	struct	   ccb_en_lun *cel;
+	union      ccb* cancel_ccb;
 	cam_status status;
 	u_int	   target;
 	u_int	   lun;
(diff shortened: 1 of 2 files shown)


bsdimp:
@amotin This seems sane to me and matches ISP which I know you've used recently in target mode. Before, we'd just fail the request if there's outstanding I/O of these types, but after we'll fail the I/Os and there's no need to fail the disable request. At least that's my read.

@hpvb The commit message for this is a little ambiguous. It should state more clearly that you are changing the code to match isp, which is working. As it's written now, isp could be doing the same thing as the old code.  It'd suggest `does the same thing when disabling a lun` to `aborts pending AIO and IONTs when disabling a lun` to make it clearer.

bsdimp:
wrong button krank...

amotin (approved):
The change looks right.  It would just be nice to go further and allow LUN disable under a load, that now still may return an error, as I can see.

bsdimp:
Landed, thanks!

bsdimp:
> The change looks right. It would just be nice to go further and allow LUN disable under a load, that now still may return an error, as I can see.

That would be cool...  I'll leave that for followup submissions.

hpvb:
> The change looks right. It would just be nice to go further and allow LUN disable under a load, that now still may return an error, as I can see.

Yeah, I haven't yet been able to reliably trigger that condition, as soon as I can I will take care of that also.

PR #1191, opened 2024-04-21
sys/net/if_bridge: clean up INET/INET6 handling

llfw:
The if_bridge contains several instances of:

	if (AF_INET code ...
	#ifdef INET6
	    AF_INET6 code ...
	#endif
	) {
		...

Clean this up by adding a couple of macros at the top of the file that are conditionally defined based on whether INET and/or INET6 are enabled, which makes the code more readable and easier to maintain.

---

split from #1159.  cc: @bsdimp, @markjdb, @glebius, @gmshake.

### DIFF
--- a/sys/net/if_bridge.c
+++ b/sys/net/if_bridge.c
@@ -135,6 +135,31 @@
 #include <net/route.h>
 
 /*
+ * At various points in the code we need to know if we're hooked into the INET
+ * and/or INET6 pfil.  Define some macros to do that based on which IP versions
+ * are enabled in the kernel.  This avoids littering the rest of the code with
+ * #ifnet INET6 to avoid referencing V_inet6_pfil_head.
+ */
+#ifdef INET6
+#define		PFIL_HOOKED_IN_INET6	PFIL_HOOKED_IN(V_inet6_pfil_head)
+#define		PFIL_HOOKED_OUT_INET6	PFIL_HOOKED_OUT(V_inet6_pfil_head)
+#else
+#define		PFIL_HOOKED_IN_INET6	false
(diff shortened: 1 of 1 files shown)


sys/net/if_bridge.c
```diff
@@ -134,6 +134,31 @@
...
+
+#ifdef INET
+#define		PFIL_HOOKED_IN_INET	PFIL_HOOKED_IN(V_inet_pfil_head)
+#define		PFIL_HOOKED_OUT_INET	PFIL_HOOKED_OUT(V_inet_pfil_head)
+#else
+#define		PFIL_HOOKED_IN_INET	false
+#define		PFIL_HOOKED_OUT_INET	false
+#endif
+
+#define	PFIL_HOOKED_IN_46	(PFIL_HOOKED_IN_INET6 || PFIL_HOOKED_IN_INET)
```
markjdb, line 159: There should be a tab after `#define` for consistency with the other definitions.

markjdb (commented):
Looks good to me aside from one nit.

gmshake (approved):
Looks good to me.

sys/net/if_bridge.c
llfw: fixed.

markjdb:
Merged, thank you.

bsdimp:
ah cool... I was going to include this in my bulk run, but now I don't have to. thanks @markjdb

markjdb:
> ah cool... I was going to include this in my bulk run, but now I don't have to. thanks @markjdb

Oh, sorry, I didn't check the assignee. Will do next time before merging things.

PR #1193, opened 2024-04-23
packages: add package for NTP

llfw:


llfw:
cc @evadot, @bapt

evadot:
> > LGTM, but please add an UPDATING entry so people will know that they need to pkg install FreeBSD-ntp after an upgrade.
> 
> Should we have a single updating entry for all these things? Or one per. Seems like a single one to catch-up to the changes so far makes sense with followup ones as they land.

One UPDATING entry for all pkgbase changes is good a long as the changes are commited at the same time (more or less).

llfw:
i'll add UPDATING with each PR in the future, since they might not get all committed at the same time (depending on review), they can be merged into one before commit if needed.

do you want me to make a PR with an UPDATING for the recent ones?  (i think that's this one, cron, and lpr at least?)

evadot:
> i'll add UPDATING with each PR in the future, since they might not get all committed at the same time (depending on review), they can be merged into one before commit if needed.
> 
> do you want me to make a PR with an UPDATING for the recent ones? (i think that's this one, cron, and lpr at least?)

Yes please.

llfw:
#1196

llfw:
#1196 was merged and includes the `UPDATING` entry for ntp package, but this commit (which contains the actual ntp package) hasn't been merged.  that might be my fault for not noting that explicitly, but this should probably be merged as well.

PR #1197, opened 2024-04-24
package: move OpenBSM auditing into its own package

llfw:
Move auditing runtime (audit, etc.) into the new FreeBSD-audit package. Also move the runtime OpenBSM manual based from libbsm into auditd so they get installed with the right package.

Add an UPDATING entry noting the new packages.


llfw:
cc: @evadot, @bapt

brooksdavis:
Does this split make sense given that su requires libbsm and rc.subr requires su?

llfw:
this doesn't affect libbsm, that was already in its own packages and this commit doesn't change that.

this only affects users who explicitly enable OpenBSM auditing and are running auditd(8).

the benefit here is that auditd(8) et al don't need to be installed for users who are not using OpenBSM auditing, which is probably 99% of users.

evadot:
> Does this split make sense given that su requires libbsm and rc.subr requires su?

As Lexi said libbsm is in its own package and it's fine, it's also used by sshd for example and we don't want to bring all those utilities just for sshd.

Reviewed by: manu

PR #1198, opened 2024-04-24
package: move authpf into the FreeBSD-pf package

llfw:
no UPDATING entry for this since authpf is only useful for pf(4) users, who must already have FreeBSD-pf installed.


llfw:
cc @evadot, @bapt

evadot:
Reviewed by: manu

PR #1199, opened 2024-04-25
sys/amd64/conf/MINIMAL: remove virtio

llfw:
There is no reason for virtio support to be in minimal since it's all built as modules.  Remove it from the kernel config.

---

cc @igalic

### DIFF
--- a/sys/amd64/conf/MINIMAL
+++ b/sys/amd64/conf/MINIMAL
@@ -127,13 +127,6 @@ device		ether			# Ethernet support
 # Note that 'bpf' is required for DHCP.
 device		bpf			# Berkeley packet filter
 
-# VirtIO support
-device		virtio			# Generic VirtIO bus (required)
-device		virtio_pci		# VirtIO PCI device
-device		vtnet			# VirtIO Ethernet device
-device		virtio_blk		# VirtIO Block device
-device		virtio_balloon		# VirtIO Memory Balloon device
-
 # Linux KVM paravirtualization support
 device		kvm_clock		# KVM paravirtual clock driver
 


igalic (commented):
I like this idea, and i would like it even more if i could figure out how to make it boot now.

llfw:
> I like this idea, and i would like it even more if i could figure out how to make it boot now.

the trick is you have to load virtio_pci.ko.  that doesn't get autoloaded (but it should) and without that the virtio pci devices aren't detected.

igalic:
here's my loader.conf,
```loader.conf
# bsdinstall defaults
security.bsd.allow_destructive_dtrace=0
kern.geom.label.disk_ident.enable="0"
kern.geom.label.gptid.enable="0"
cryptodev_load="YES"
zfs_load="YES"
kernels_autodetect="YES"
boot_verbose="YES"
# boot *my* kernel
kernel="kernel.MINIMAL"
verbose_loading="YES"
# we'll need this to actually boot MINIMAL
virtio_pci_load="YES"
virtio_mmio_load="YES"
virtio_scsi_load="YES"
autoboot_delay="1"
# do not load virtio_socket
module_blacklist="virtio_socket"
```

which still leads to:

image

as soon as i figure out how to scroll up to see what's actually happening… since I'm pretty sure i won't find a log file given that the system can't find a disk to which to write logs to…

Crest:
It's a really bad idea to remove those tiny drivers because they're required to boot with an empty /boot/loader.conf in common virtual machine deployments. Is this really the lowest hanging fruit and saves enough memory to be worth breaking what I expect to be quite a lot of workflows?

kevans91:
> It's a really bad idea to remove those tiny drivers because they're required to boot with an empty /boot/loader.conf in common virtual machine deployments. Is this really the lowest hanging fruit and saves enough memory to be worth breaking what I expect to be quite a lot of workflows?
> 

+1, I don't think we should be removing virtio until loader can do pnp autoloading beyond FDT.  Until then, we should really not remove core bits (perhaps with the exception for user choices that are elected via the installer; e.g., root fs) from MINIMAL. CC @bsdimp 

> Did you check the resulting kernel size with and without the VirtIO drivers?

bsdimp:
I'm leaning towards closing this as well, for the reasons @kevans91 and @Crest stated. Ideally we'd remove it, but we're not quite to the place of being able to do that with current loader technology, so this falls into 'except what's needed to mount root' category.

I'd like to make the loader loader ufs.ko or zfs.ko based on the root fileystem's device that we read the kernel from, controlled by a rootfs_autoload variable, but that's also beyond the scope here.

I'll leave it open for a day or so to see if some other compelling argument is made.

llfw:
> this falls into 'except what's needed to mount root' category.

but you already don't get that with MINIMAL - there's no nvme(4) or ahci(4) or mps(4), you have to load both the storage driver and the filesystem driver manually.  we should definitely fix that, but meanwhile, i'm not sure there's any justification for virtio being in MINIMAL when no other block/network devices are.*

\* except Xen because it's weird.

bsdimp:
I think the consensus is that this change is premature, though with a small grumbling. We should look to addressing the concerns around minimal in 15 / 14.2. It would be cool to get the PCI loading stuff into the loader, but that has issues with modules that are already in the kernel that need to be sorted first. We might also want to go add nvme and ahci (and maybe mpt since it's used by virtualbox IIRC) to minimal as a stop gap for now. I'm sorry.
https://reviews.freebsd.org/D45012 makes more things included until we sort out the boot loader issues.

PR #1201, opened 2024-04-26
Fix CTL GET EVENT STATUS NOTIFICATION valid bits

hpvb:
Linux as an initiator periodically sends the following to SCSI cdrom devices: 4a 01 00 00 10 00 00 00 08 00

According to the ctl_cmd_entry for this command this is invalid which leads to a lot of failed SCSI commands.

The mask in this commit is based off of the t10 spec available here: https://www.t10.org/ftp/t10/document.97/97-108r0.pdf

And @bsdimp's research.

This solves the issue with Linux initiators.

### DIFF
--- a/sys/cam/ctl/ctl_cmd_table.c
+++ b/sys/cam/ctl/ctl_cmd_table.c
@@ -1227,7 +1227,7 @@ const struct ctl_cmd_entry ctl_cmd_table[256] =
 					  CTL_CMD_FLAG_ALLOW_ON_PR_RESV |
 					  CTL_FLAG_DATA_IN,
  CTL_LUN_PAT_NONE,
- 10, {0x02, 0x01, 0, 0, 0, 0xff, 0xff, 0xff, 0x07}},
+ 10, {0xe1, 0, 0, 0x7e, 0, 0, 0xff, 0xff, 0x07}},
 
 /* 4B PAUSE/RESUME */
 {NULL, CTL_SERIDX_INVLD, CTL_CMD_FLAG_NONE, CTL_LUN_PAT_NONE},


sys/cam/ctl/ctl_cmd_table.c
```diff
@@ -1227,7 +1227,7 @@ const struct ctl_cmd_entry ctl_cmd_table[256] =
 					  CTL_CMD_FLAG_ALLOW_ON_PR_RESV |
 					  CTL_FLAG_DATA_IN,
  CTL_LUN_PAT_NONE,
- 10, {0x02, 0x01, 0, 0, 0, 0xff, 0xff, 0xff, 0x07}},
+ 10, {0xe1, 0, 0, 0x54, 0, 0x0, 0xff, 0xff, 0xc7}},
```
bsdimp, line 1230: OK... MMC4 says this mask should be:
{ 0x01, 0, 0, 0xff, 0, 0, 0xff, 0xff, ??? } (I couldn't find control byte in mmc and didn't want to go hunting)
MMC3 -- same.
MMC2 -- same.
MMC -- not present

So that's somewhat different... Though MMC2 does have the upper three bits separated from the rest, so that kinda makes sense. I had trouble snagging your document

bsdimp:
So this one was fun to track down... Glad I spent a fair amount of time downloading the scsi specs ages ago... gotta go look for mmc5 and mmc6 it turns out...

sys/cam/ctl/ctl_cmd_table.c
amotin: I haven't looked deep yet, but just for a note, here should be not simply the bits defined in specs, but only the bits supported by the ctl_get_event_status() function.  I see it checks only one bit from the byte2.  We should check that no other bits left allowed but not handled do anything wrong.

amotin: I agree with Warner's version.

hpvb: Okay, I'll adjust the PR

hpvb:
I've adjusted the PR based on your input.

amotin:
@hpvb I'm sorry, but I don't see the changes.

hpvb:
> @hpvb I'm sorry, but I don't see the changes.

I forgot at git add the file when I amended the commit. Sorry! Fixed.

sys/cam/ctl/ctl_cmd_table.c
```diff
@@ -1227,7 +1227,7 @@ const struct ctl_cmd_entry ctl_cmd_table[256] =
 					  CTL_CMD_FLAG_ALLOW_ON_PR_RESV |
 					  CTL_FLAG_DATA_IN,
  CTL_LUN_PAT_NONE,
- 10, {0x02, 0x01, 0, 0, 0, 0xff, 0xff, 0xff, 0x07}},
+ 10, {0xe1, 0, 0, 0x7e, 0, 0x0, 0xff, 0xff, 0xc7}},
```
bsdimp, line 1231: This style checker message is a bit too zealous.

bsdimp (approved):
I think this is good, but will defer to mav@ on the details since my ctl fu is still weak.

sys/cam/ctl/ctl_cmd_table.c
```diff
@@ -1227,7 +1227,7 @@ const struct ctl_cmd_entry ctl_cmd_table[256] =
...
```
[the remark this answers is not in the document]
hpvb: The spec says to throw an error when unknown vendor specific bits are set. So I will change the PR once more. I misunderstood it initially.

amotin (approved):

bsdimp:
Landed with tweaked commit message to spell out what my research is / was.

PR #1202, opened 2024-04-26
CTL: READ(6) should be allowed on CD devices

hpvb:
DOS ASPI drivers use this, with this change applied it is possible to load a DOS ASPI CDROM driver and run the Windows 98 installer.


bsdimp:
Approved. But grudgingly.
I want to eliminate all READ(6) since it would trim a lot off the list of quirks...
However, this on the 'accepting' side, not the 'generation' side, so there's no harm in accepting it, even though the READ6 command has been deleted from the standard.

bsdimp:
Since you're boot Window98, that's actually from a time when READ6 was only deprecated, but all the stacks generated it to optimize the CDB sizes because SCSI controllers of the era had limited CDB buffers and being able to squeeze another command into it was a win....

PR #1204, opened 2024-04-26
package: FreeBSD-periodic should depend on FreeBSD-cron

llfw:
Reported by: des

---

#1203 should be merged along with this otherwise the new dependency won't work.


bsdimp:
Landed, thanks. I landed in the wrong order, though, since I didn't see the #1203 dependency hadn't yet been merged (I thought it had, but didn't actually check first <doh!>).

PR #1205, opened 2024-04-26
glabel: Add support for Linux swap

ricardobranco777:
This will benefit dual-boot installations.

This is possible without corrupting the Linux swap header because FreeBSD, like NetBSD, skips the first page when adding swap:
https://github.com/freebsd/freebsd-src/blob/main/sys/vm/swap_pager.c#L2432

How I tested it:

1. Compile kernel without GEOM_LABEL
1. `cd /sys/modules/geom/geom_label`
1. `make`
1. `sudo make install`
1. `sudo kldunload -v geom_label`
1. `sudo kldload -v geom_label`
1. Attach virtual disk to VM.
1. `echo -e 'n\np\n1\n\n\nt\n82\nw' | sudo chroot /compat/linux fdisk /dev/vtbd1` to create a Linux swap partition.
1. `sudo chroot /compat/linux mkswap -L MYSWAP /dev/vtbd1s1`
1. `sudo swapon /dev/swaplinux/MYSWAP`

To use:
1. `sudo sh -c 'echo /dev/swaplinux/MYSWAP none swap sw 0 0' >> /etc/fstab`
1. `sudo sh -c 'echo geom_label_load=\"YES\" >> /boot/loader.conf'`
1. Reboot
1. `swapinfo | grep MYSWAP`

### DIFF
--- a/lib/geom/label/glabel.8
+++ b/lib/geom/label/glabel.8
@@ -115,6 +115,9 @@ EXT2FS (directory
 .It
 NTFS (directory
 .Pa /dev/ntfs/ ) .
+.It
+Swap Linux (directory
+.Pa /dev/swaplinux/ ) .
 .El
 .Pp
 Support for partition metadata is implemented for:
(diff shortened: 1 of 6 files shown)


kostikbel (approved):

bsdimp:
merged. thanks!

PR #1206, opened 2024-04-26
rtadvd(8): support PREF64 (RFC 8781)

llfw:
the first commit adds the necessary icmp6.h bits, the second is the implementation in rtadvd.

### DIFF
--- a/sys/netinet/icmp6.h
+++ b/sys/netinet/icmp6.h
@@ -63,6 +63,8 @@
 #ifndef _NETINET_ICMP6_H_
 #define _NETINET_ICMP6_H_
 
+#include <stdint.h>
+
 #define ICMPV6_PLD_MAXLEN	1232	/* IPV6_MMTU - sizeof(struct ip6_hdr)
 					   - sizeof(struct icmp6_hdr) */
 
@@ -375,10 +377,10 @@ struct nd_opt_dnssl {		/* DNSSL option (RFC 6106) */
 } __packed;
 
 struct nd_opt_pref64 {		/* PREF64 option (RFC 8781) */
-	u_int8_t	nd_opt_pref64_type;
-	u_int8_t	nd_opt_pref64_len;
+	uint8_t		nd_opt_pref64_type;
+	uint8_t		nd_opt_pref64_len;
 	/* bits 0-12 are the SL, bits 13-15 are the PLC */
-	u_int16_t	nd_opt_pref64_sl_plc;
+	uint16_t	nd_opt_pref64_sl_plc;
 	char		nd_opt_prefix[12];
 } __packed;
 
--- a/sys/netinet/icmp6.h
+++ b/sys/netinet/icmp6.h
@@ -307,7 +307,8 @@ struct nd_opt_hdr {		/* Neighbor discovery option header */
 #define ND_OPT_ROUTE_INFO		24	/* RFC 4191 */
 #define ND_OPT_RDNSS			25	/* RFC 6106 */
 #define ND_OPT_DNSSL			31	/* RFC 6106 */
-#define ND_OPT_MAX			31
+#define ND_OPT_PREF64			38	/* RFC 8781 */
+#define ND_OPT_MAX			38
 
 struct nd_opt_prefix_info {	/* prefix information */
 	u_int8_t	nd_opt_pi_type;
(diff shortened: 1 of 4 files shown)


bsdimp:
This needs a better reviewer than me. @markjdb

sys/netinet/icmp6.h
```diff
@@ -373,6 +374,14 @@ struct nd_opt_dnssl {		/* DNSSL option (RFC 6106) */
 	/* followed by list of DNS search domains */
 } __packed;
 
+struct nd_opt_pref64 {		/* PREF64 option (RFC 8781) */
+	u_int8_t	nd_opt_pref64_type;
+	u_int8_t	nd_opt_pref64_len;
+	/* bits 0-12 are the SL, bits 13-15 are the PLC */
+	u_int16_t	nd_opt_pref64_sl_plc;
```
glebius, line 381: Can you please change to C99 uintXX_t types instead of non-standard BSD 80-ish u_intXX_t.

usr.sbin/rtadvd/config.c
```diff
@@ -1435,6 +1490,19 @@ make_packet(struct rainfo *rai)
 		buf += sizeof(struct nd_opt_mtu);
 	}
 
+	if (rai->rai_pref64.p64_enabled) {
+		ndopt_pref64 = (struct nd_opt_pref64 *)buf;
+		ndopt_pref64->nd_opt_pref64_type = ND_OPT_PREF64;
+		ndopt_pref64->nd_opt_pref64_len = 2;
+		ndopt_pref64->nd_opt_pref64_sl_plc =
+			(htons(rai->rai_pref64.p64_sl << 3))
+			| htons((rai->rai_pref64.p64_plc & 0x7));
```
glebius, line 1499: Per style(9) the operator (| in this case) should be at the end of a line.

usr.sbin/rtadvd/config.c
```diff
@@ -912,6 +912,58 @@ getconfig(struct ifinfo *ifi)
...
+				break;
+			default:
+				syslog(LOG_ERR, "prefix length %" PRIi64
+				       "on %s is invalid; disabling PREF64",
+				       val64, ifi->ifi_ifname);
+				rai->rai_pref64.p64_enabled = 0;
+				break;
+			}
+
+			/* this logic is from RFC8781 section 4.1 */
```
hhartzer, line 956: Very minor nit: I feel like "RFC 8781" would be more consistent here, rather than "RFC8781". I think this a great comment, though.

usr.sbin/rtadvd/config.c
```diff
@@ -912,6 +912,58 @@ getconfig(struct ifinfo *ifi)
...
 		free(dns);
 	}
+
+	/*
+	 * handle pref64
+	 */
+	rai->rai_pref64.p64_enabled = false;
+
+	if ((addr = (char *)agetstr("pref64", &bp))) {
+		if (inet_pton(AF_INET6, addr, &rai->rai_pref64.p64_prefix) != 1) {
```
hhartzer, line 922: I'm no expert with C, but in general I feel it reads better to start with the success case (`== 1`) in a conditional like this. If my opinion goes against `style(9)` or any other convention, please ignore.

sys/netinet/icmp6.h
llfw: do you want me to change every instance in the file (for consistency) or just the new ones?

glebius: Just the new ones in the scope of this pull request. However, if you do a sweep across the entire file as a separate change, that would be great.

usr.sbin/rtadvd/config.c
bzfbd: And "This" and a "." at the end if you are touching it for whitespace already.

bzfbd:
I have another question: what are you using on FreeBSD to test this?

llfw:
> I have another question: what are you using on FreeBSD to test this?

nothing; as far as i'm aware, FreeBSD as a client has no support for PREF64.

i'm testing with macOS and iOS, which have support out of the box.

sys/netinet/icmp6.h
llfw: done.

usr.sbin/rtadvd/config.c
llfw: both fixed.

usr.sbin/rtadvd/config.c
llfw: i disagree, but i understand where you're coming from.

the problem is that this function is horrible to begin with, with a convoluted control controlled by multiple different 'goto' labels, and is quite difficult to follow.  with that in mind, i think erroring early here makes the new code easier to read than putting the success case first.

the properly fix here is to rewrite the entire code in a more modern style, but that's out of scope for this feature.

usr.sbin/rtadvd/config.c
llfw: fixed.

usr.sbin/rtadvd/config.c
hhartzer: I get that, thanks for explaining. Probably best to keep it like this, until it can be entirely rewritten.

bsdimp:
This looks ready. Anything additional can be done as a followup.

PR #1207, opened 2024-04-27
rights.4: various corrections on capability rights

CismonX:
- A file descriptor obtained from `accept(2)`, `accept4(2)` and `openat(2)` is not always assigned all capability rights.  Instead, it inherits capability rights from the "parent" socket/dir file descriptor.
- `getdents(2)` and `getdirentries(2)` requires `CAP_READ`.
- `openat(2)` with `O_WRONLY|O_TRUNC` does not require `CAP_SEEK`. See 943c3bb9688d6d84f9e1314525cfa638003817b2.


bsdimp:
Quick checks of the kernel match the manual, as well as common sense...
Landing this Thanks!

PR #1208, opened 2024-04-27
aic7xxx: make target mode enable a device hint

hpvb:
Previously it was only possible to enable target mode for these drivers by rebuilding the kernel with `AHC_TMODE_ENABLE` or `AHD_TMODE_ENABLE` and a bitmask of which units to statically enable for target mode.

There is no space-savings in the driver by not having `AHC_TMODE_ENABLE` set, so in addition to the compile time option lets also introduce some tunables:

`hint.ahc.<unit>.tmode_enable=0/1`
`hint.ahd.<unit>.tmode_enable=0/1`

For compatibility the old behavior is retained, but it can be overridden with tunables

### DIFF
--- a/sys/dev/aic7xxx/aic79xx.c
+++ b/sys/dev/aic7xxx/aic79xx.c
@@ -6229,6 +6229,9 @@ ahd_init(struct ahd_softc *ahd)
 	size_t		 driver_data_size;
 	int		 i;
 	int		 error;
+#ifdef AHD_TARGET_MODE
+	int		 tmode_enable;
+#endif
 	u_int		 warn_user;
 	uint8_t		 current_sensing;
 	uint8_t		 fstat;
@@ -6261,8 +6264,22 @@ ahd_init(struct ahd_softc *ahd)
 	/*
 	 * Only allow target mode features if this unit has them enabled.
 	 */
-	if ((AHD_TMODE_ENABLE & (0x1 << ahd->unit)) == 0)
+#ifdef AHD_TARGET_MODE
+	tmode_enable = ((AHD_TMODE_ENABLE & (0x1 << ahd->unit)) != 0);
+	resource_int_value(device_get_name(ahd->dev_softc),
+			       device_get_unit(ahd->dev_softc),
+			       "tmode_enable", &tmode_enable);
+
+	if (tmode_enable == 0) {
 		ahd->features &= ~AHD_TARGETMODE;
+	} else {
+		if (bootverbose && ((ahd->features & AHD_TARGETMODE) != 0))
+			printf("%s: enabling target mode\n", ahd_name(ahd));
+	}
+
+#else
+	ahd->features &= ~AHD_TARGETMODE;
+#endif
 
 	/* DMA tag for mapping buffers into device visible space. */
 	if (aic_dma_tag_create(ahd, ahd->parent_dmat, /*alignment*/1,
--- a/share/man/man4/ahc.4
+++ b/share/man/man4/ahc.4
@@ -40,12 +40,6 @@ kernel configuration file:
 .Pp
 For one or more PCI cards:
 .Cd "device pci"
-.Pp
-To allow PCI adapters to use memory mapped I/O if enabled:
-.Cd options AHC_ALLOW_MEMIO
-.Pp
-To configure one or more controllers to assume the target role:
-.Cd options AHC_TMODE_ENABLE <bitmask of units>
 .Ed
 .Pp
 Alternatively, to load the driver as a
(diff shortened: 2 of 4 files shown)


hpvb:
Sorry for the many edits and force-pushes. I changed my mind several times. I'm pretty sure this is good now though!

amotin (commented):
Why did you keep AHC_TARGET_MODE in code, while removed from configs?  How it is supposed to be set and what for?  I would understand if you kept it as alternative method of enabling target to keep compatibility with existing configs, but this way I just don't understand.

hpvb:
This is how it has always been. AHC_TARGET_MODE gets unconditionally enabled in aic7xxx_osm.h here:

```
/* This driver supports target mode */
#define AHC_TARGET_MODE 1
```

If you want a smaller driver I guess you can set that to 0?

hpvb:
@amotin Do you still feel this should be done differently? To address your concern I'd have to rip out all of `AHC_TARGET_MODE` from the driver. This has been in there since 2002 in this form, not possible to be disabled by any config setting.

Alternatively I could introduce a way to switch it in config and allow the driver to be (re) built without target support entirely. But I feel that this should be a separate PR then.

amotin:
Thanks for the clarification.  I indeed mixed AHC_TARGET_MODE with AHC_TMODE_ENABLE.  I do not insist in massive driver modifications.  But it seems we could trivially keep AHC_TMODE_ENABLE for compatibility, using it as default value to override with tunables.

sys/dev/aic7xxx/aic79xx.c
```diff
@@ -6261,8 +6264,32 @@ ahd_init(struct ahd_softc *ahd)
...
 	 * Only allow target mode features if this unit has them enabled.
 	 */
-	if ((AHD_TMODE_ENABLE & (0x1 << ahd->unit)) == 0)
+#ifdef AHD_TARGET_MODE
+        if (resource_int_value(device_get_name(ahd->dev_softc),
+                               device_get_unit(ahd->dev_softc),
+                               "tmode_enable", &tmode_enable) != 0) {
+		if (bootverbose)
+			printf("%s: No ahd.%i.tmode_enable hint found\n",
+				ahd_name(ahd), ahd->unit);
```
amotin, line 6273: I think it is too verbose to report every hint that was not set in a system.

hpvb: The thinking there was to alert users who used to use `AHC_TMODE_ENABLE` that something had changed. But I see your point. If I keep the `AHC_TMODE_ENABLE` in addition to the new hint I guess that won't be a problem.

hpvb:
@amotin I think I addressed your concerns.

amotin (commented):
The direction looks good to me, but I would further cut the printfs.  BTW, there is device_printf() to print something for specific device without crafting local.

hpvb:
> The direction looks good to me, but I would further cut the printfs. BTW, there is device_printf() to print something for specific device without crafting local.

I have removed most of the debug prints now, I thought they were kinda useful but perhaps too chatty even for verbose.

I'm aware of device_printf() but the driver uses this throughout, I do kind of want to replace all of the existing hadcrafted ones with device_printf() but for consistency it seemed better to just keep using the local convention until I got around to replacing all of them.

amotin (approved):

sys/dev/aic7xxx/aic79xx.c
```diff
@@ -6261,8 +6264,22 @@ ahd_init(struct ahd_softc *ahd)
...
+	tmode_enable = ((AHD_TMODE_ENABLE & (0x1 << ahd->unit)) != 0);
+	resource_int_value(device_get_name(ahd->dev_softc),
+			       device_get_unit(ahd->dev_softc),
+			       "tmode_enable", &tmode_enable);
+
+	if (tmode_enable == 0) {
 		ahd->features &= ~AHD_TARGETMODE;
+	} else {
+		if (bootverbose)
+			printf("%s: enabling target mode\n", ahd_name(ahd));
```
amotin, line 6277: On a second thought, this message would be confusing if (ahd->features & AHD_TARGETMODE) is false for some other reason, like not supported by hardware/firmware.  Can it be?

hpvb: You are right, it can be for cards that do not support it at all. Good catch, thank you. I will fix that.

hpvb: Fixed now, thank you.

hpvb:
@amotin are you happy with this now? I'd love to see this merged as I have some other changes pending that depend on these changes to the man pages :)

amotin (approved):

bsdimp:
I'll add this to today's batch.

bsdimp:
Dang, was hoping I'd pushed the right thing to have this show up as merged.

PR #1210, opened 2024-04-27
mdconfig: Print options when listing devices

ricardobranco777:
Currently, we have no way of seeing options set to md(4) devices.

Also fix usage string (missing `async` option).

Example:

`# mdconfig -a -f /root/backup.iso -o readonly -o verify`

Before:
```
# mdconfig -l -v
md0	vnode	 1536K	/root/backup.iso
```

After:
```
# mdconfig -l -v
md0	vnode	 1536K	/root/backup.iso	-	readonly,verify
```

Fixes https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=278624

### DIFF
--- a/sbin/mdconfig/mdconfig.c
+++ b/sbin/mdconfig/mdconfig.c
@@ -56,7 +56,7 @@
 
 static struct md_ioctl mdio;
 static enum {UNSET, ATTACH, DETACH, RESIZE, LIST} action = UNSET;
-static int nflag;
+static int md_fd, nflag;
 
 static void usage(void) __dead2;
 static void md_set_file(const char *);
@@ -65,6 +65,7 @@ static int md_query(const char *, const int, const char *);
 static int md_list(const char *, int, const char *);
 static char *geom_config_get(struct gconf *g, const char *name);
 static void md_prthumanval(char *length);
+static void print_options(const char *s, const char *);
 
 #define OPT_VERBOSE	0x01
 #define OPT_UNIT	0x02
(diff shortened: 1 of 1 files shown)


sbin/mdconfig/mdconfig.c
```diff
@@ -86,9 +87,9 @@ usage(void)
...
 "       mdconfig file\n");
 	fprintf(stderr, "\t\ttype = {malloc, vnode, swap}\n");
-	fprintf(stderr, "\t\toption = {cache, cluster, compress, force,\n");
-	fprintf(stderr, "\t\t          mustdealloc, readonly, reserve, ro,\n");
-	fprintf(stderr, "\t\t          verify}\n");
+	fprintf(stderr, "\t\toption = {async, cache, cluster, compress,\n");
+	fprintf(stderr, "\t\t          force, mustdealloc, readonly, ro,\n");
+	fprintf(stderr, "\t\t          reserve, verify}\n");
 	fprintf(stderr, "\t\tsize = %%d (512 byte blocks), %%db (B),\n");
 	fprintf(stderr, "\t\t       %%dk (kB), %%dm (MB), %%dg (GB), \n");
```
bsdimp, line 94: Should delete this space after the , and before the \n

bsdimp:
Otherwise, I like this.

sbin/mdconfig/mdconfig.c
ricardobranco777: > Should delete this space after the , and before the \n

Fixed, thanks!

bsdimp:
Thanks for this! Please close the PR now that this is in.

asomers:
@ricardobranco777 what is the purpose of the `\0` separator?  Should that initially be `''`?  Currently, there is always an actual `\0` character printed at the beginning of the option list.

ricardobranco777:
> @ricardobranco777 what is the purpose of the `\0` separator? Should that initially be `''`? Currently, there is always an actual `\0` character printed at the beginning of the option list.

You're right.  It must be ''.  My bad.  Will submit a PR tomorrow.

bsdimp:
> > @ricardobranco777 what is the purpose of the `\0` separator? Should that initially be `''`? Currently, there is always an actual `\0` character printed at the beginning of the option list.
> 
> You're right. It must be `''`. My bad. Will submit a PR tomorrow.

Doh! Looks like I missed seeing that. Thanks for fixing.

PR #1211, opened 2024-05-01
Address module reproducibility issues

dankm:
Use .PATH & bare filename. This prevents the real source path from being included in the built object, which improves reproducibility.

### DIFF
--- a/sys/modules/ice_ddp/Makefile
+++ b/sys/modules/ice_ddp/Makefile
@@ -1,5 +1,7 @@
 
+.PATH: ${SRCTOP}/sys/contrib/dev/ice
+
 KMOD=	ice_ddp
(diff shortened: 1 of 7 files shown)


igalic:
are these the only modules that use this pattern, or are you using them as an example?

dankm:
> are these the only modules that use this pattern, or are you using them as an example?

These are the only ones that I caught doing it. Most are doing it in a similar way to what I changed these to. Some are "building" the firmware by extracting a uuencoded file and then loading that. I searched for more, and didn't find any in-tree. Most look like [this file](https://github.com/freebsd/freebsd-src/blob/main/sys/modules/qat_c2xxxfw/Makefile#L2) in main.

bsdimp:
This builds, and produces what looks good. Should autoclose in a few minutes

PR #1212, opened 2024-05-02
.github: Update the path used for the homebrew LLVM install on macOS

bsdjhb:
Using a PR to test


bsdimp:
Invented a new tag for the test-only pull requests so people landing pull requests stay hands off. Hope the test works!

PR #1213, opened 2024-05-04
Enable nomatch events when one_nomatch is empty

ica574:
To avoid redundant nomatch events on boot, a new sysctl is introduced, which is set to 0 initially. This blocks nomatch events within devctl_nomatch_handler in /sys/kern/devctl.c. When $one_nomatch in /etc/rc.d/devmatch is empty, the value of the sysctl is set to 1, allowing nomatch events to be processed. Out of three samples, the boot processes is sped up by approximately 170 ms.

### DIFF
--- a/sys/kern/kern_devctl.c
+++ b/sys/kern/kern_devctl.c
@@ -89,6 +89,9 @@ static int sysctl_devctl_queue(SYSCTL_HANDLER_ARGS);
 static int devctl_queue_length = DEVCTL_DEFAULT_QUEUE_LEN;
 SYSCTL_PROC(_hw_bus, OID_AUTO, devctl_queue, CTLTYPE_INT | CTLFLAG_RWTUN |
     CTLFLAG_MPSAFE, NULL, 0, sysctl_devctl_queue, "I", "devctl queue length");
+static bool nomatch_enabled = false;
+SYSCTL_BOOL(_hw_bus, OID_AUTO, devctl_nomatch_enabled, CTLFLAG_RWTUN,
+    &nomatch_enabled, 0, "enable nomatch events");
 
 static void devctl_attach_handler(void *arg __unused, device_t dev);
 static void devctl_detach_handler(void *arg __unused, device_t dev,
@@ -208,7 +211,8 @@ devctl_detach_handler(void *arg __unused, device_t dev, enum evhdev_detach state
 static void
 devctl_nomatch_handler(void *arg __unused, device_t dev)
 {
-	devaddq("?", "", dev);
+	if (nomatch_enabled)
+		devaddq("?", "", dev);
 }
 
 static int
--- a/libexec/rc/rc.d/devmatch
+++ b/libexec/rc/rc.d/devmatch
@@ -46,6 +46,7 @@ devmatch_start()
 	if [ -n "$one_nomatch" ]; then
 		list=$(devmatch -p "${one_nomatch}" | sort -u)
 	else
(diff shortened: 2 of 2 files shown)


sys/kern/kern_devctl.c
```diff
@@ -89,6 +89,9 @@ static int sysctl_devctl_queue(SYSCTL_HANDLER_ARGS);
 static int devctl_queue_length = DEVCTL_DEFAULT_QUEUE_LEN;
 SYSCTL_PROC(_hw_bus, OID_AUTO, devctl_queue, CTLTYPE_INT | CTLFLAG_RWTUN |
     CTLFLAG_MPSAFE, NULL, 0, sysctl_devctl_queue, "I", "devctl queue length");
+static bool nomatch_enabled = 0;
```
bsdimp, line 92: Please use `false` instead of `0`.

bsdimp:
This looks like two commits, but only one commit message. Maybe this commit message would fix this:

```
Introduce hw.bus.devctl_nomatch_enabled and use it to suppress NOMATCH until devmatch runs

There's a lot of NOMATCH events generated at boot. We also run devmatch once during early boot to load unmatched devices. To avoid redundant work, don't start generating NOMATCH events until after devmatch runs. Set hw.bus.devctl_nomatch_enabled=1 just before we run devmatch. The kernel will suppress NOMATCH events until this is set to true.

This saves about 170ms from the boot on <a word or two about the setup you tested on>
```

Also, I think we should document this somewhere, but I'm not sure where. And characterizing the hardware or env where 170ms was saved would be great too.

bsdimp:
@cperciva The code looks good, the commit message need some work. Do you want me to land this, or did you want to once we have that and the one style nit I pointed out sorted out?

cperciva:
> @cperciva The code looks good, the commit message need some work. Do you want me to land this, or did you want to once we have that and the one style nit I pointed out sorted out?

I'll land it.  @ica574 is my GSoC student.

bsdimp:
> > @cperciva The code looks good, the commit message need some work. Do you want me to land this, or did you want to once we have that and the one style nit I pointed out sorted out?
> 
> I'll land it. @ica574 is my GSoC student.

Works for me too. I have a script that works pretty well...

bsdimp:
The changes look good, from a code level. Can you rebase and squash rather than merge? It works a little for our workflow. @cperciva can help you if you have questions.

cperciva:
@bsdimp Just to confirm, you think there's nothing wrong with this commit except that it exposes something which was already broken?

bsdimp:
> @bsdimp Just to confirm, you think there's nothing wrong with this commit except that it exposes something which was already broken?

That's my current working theory. I've yet to get confirmation, though, and need to update my laptop to test for myself.

khorben:
> I don't know why but ums.ko is not autoloaded after this commit. I have to unplug / replug my mouse or kldload ums.ko manually to have my mouse detected.

I have the same issue, however unplugging / replugging my mouse does not solve the issue; it seems to be seen as a ukbd(4) device instead. My mouse and keyboard are connected on a USB hub provided by a KVM device; maybe this influences the driver selection?

bsdimp:
I backed it out already, so that is almost certainly a different bug. Still worth fixing but wrong driver is different than no driver...

khorben:
I've been bisecting on `stable/14` between 0452659e53db3cf488848345a5fab98f85526be7 and 5dced41439109368a8f3c9a818fe518542d7063e. I probably missed your revert by a few hours 😖

PR #1215, opened 2024-05-04
arp.8: don't document a syntax that doesn't exist

llfw:
The arp.8 manpage documents 'arp -d <addr> pub', but the 'pub' flag is not accepted by the arp command.  Remove this incorrect documentation.

### DIFF
--- a/usr.sbin/arp/arp.8
+++ b/usr.sbin/arp/arp.8
@@ -44,7 +44,6 @@
 .Fl a
 .Nm
 .Fl d Ar hostname
-.Op Cm pub
 .Nm
 .Fl d
 .Op Fl i Ar interface
@@ -95,15 +94,8 @@ A super-user may delete an entry for the host called
 with the
 .Fl d
 flag.
-If the
-.Cm pub
-keyword is specified, only the
-.Dq published
-.Tn ARP
-entry
-for this host will be deleted.
 .Pp
-Alternatively, the
+The
 .Fl d
 flag may be combined with the
 .Fl a


bsdimp:
"pub" is still in the sources:
```
                } else if (strcmp(argv[0], "pub") == 0) {
			opts.flags |= RTF_ANNOUNCE;
                        if (argc && strcmp(argv[1], "only") == 0) {
                                /*                                                                                                                                                                                                          
                                 * Compatibility: in pre FreeBSD 8 times                                                                                                                                                                    
                                 * the "only" keyword used to mean that                                                                                                                                                                     
                                 * an ARP entry should be announced, but                                                                                                                                                                    
                                 * not installed into routing table.                                                                                                                                                                        
                                 */
                                argc--; argv++;
                        }
                } else if (strcmp(argv[0], "blackhole") == 0) {
```
so why isn't it accepted?

3f844a222194598bd866daa7667b5be44d354fe4 added it, but I haven't seen anything that removed it, at least on purpose.

Ah, looks like 9711a168b937802c369ef4241590f671cbee9243 removed it with the code that parsed the command line option. I wonder what @glebius  was thinking. Tagged him for comment to see if this is a doc bug or a code bug... But it's been like this since 2013 :(.

llfw:
right, it's still accepted for `arp -s`, but not for `arp -d`.

glebius:
This is a doc bug and submission is correct. Thanks!

P.S. I'm having a hard time looking for "Approve" button in GitHub Web UI.

markpeek:
@glebius Files Changed tab > Review Changes button > Approve

llfw:
pushed a minor change to replace "This -d flag" with "The -d flag" which is more standard English.

glebius (approved):

bsdimp:
OK. The asymmetry is not a bug, so the doc change is good.

PR #1216, opened 2024-05-04
ifconfig(8): change default IP address format to 'cidr'

llfw:
'netmasks' haven't been used in IP networking for decades.  Change the default address format for both IPv4 and IPv6 addreses in ifconfig(8) to 'cidr', which prints addreses in the format most users will be more familiar with.

The previous format is still available using -finet:hex or -finet6:numeric.

### DIFF
--- a/UPDATING
+++ b/UPDATING
@@ -27,6 +27,12 @@ NOTE TO PEOPLE WHO THINK THAT FreeBSD 15.x IS SLOW:
 	world, or to merely disable the most expensive debugging functionality
 	at runtime, run "ln -s 'abort:false,junk:false' /etc/malloc.conf".)
 
+20240504:
+	The default output format of ifconfig(8) has changed to use CIDR
+	notation for IPv4 and IPv6 addresses.  If you have shell scripts that
+	rely on parsing ifconfig(8) output, you may want to set
+	"IFCONFIG_FORMAT=inet:hex,inet6:numeric".
+
 20240428:
 	OpenBSM auditing runtime (auditd, etc.) has been moved into the new
 	package FreeBSD-audit.  If you use OpenBSM auditing and pkgbase, you
--- a/sbin/ifconfig/af_inet.c
+++ b/sbin/ifconfig/af_inet.c
@@ -71,7 +71,6 @@ static struct in_pdata in_add, in_del;
 #endif
 
 static char addr_buf[NI_MAXHOST];	/*for getnameinfo()*/
-extern char *f_inet, *f_addr;
 
 static void
 print_addr(struct sockaddr_in *sin)
@@ -116,7 +115,9 @@ in_status(if_ctx *ctx __unused, const struct ifaddrs *ifa)
 	sin = satosin(ifa->ifa_netmask);
 	if (sin == NULL)
 		sin = &null_sin;
-	if (f_inet != NULL && strcmp(f_inet, "cidr") == 0) {
+
+	switch (f_inet) {
+	case INET_CIDR: {
 		int cidr = 32;
 		unsigned long smask;
 
@@ -128,10 +129,15 @@ in_status(if_ctx *ctx __unused, const struct ifaddrs *ifa)
 				break;
 		}
 		printf("/%d", cidr);
-	} else if (f_inet != NULL && strcmp(f_inet, "dotted") == 0)
+		break;
+	}
+	case INET_DOTTED:
 		printf(" netmask %s", inet_ntoa(sin->sin_addr));
-	else
+		break;
+	case INET_HEX:
 		printf(" netmask 0x%lx", (unsigned long)ntohl(sin->sin_addr.s_addr));
+		break;
+	}
 
 	if (ifa->ifa_flags & IFF_BROADCAST) {
 		sin = satosin(ifa->ifa_broadaddr);
@@ -168,12 +174,18 @@ in_status_nl(if_ctx *ctx __unused, if_link_t *link, if_addr_t *ifa)
 
 		printf(" --> %s", inet_ntoa(dst->sin_addr));
 	}
-	if (f_inet != NULL && strcmp(f_inet, "cidr") == 0) {
+
+	switch (f_inet) {
+	case INET_CIDR:
 		printf("/%d", plen);
-	} else if (f_inet != NULL && strcmp(f_inet, "dotted") == 0)
+		break;
+	case INET_DOTTED:
 		printf(" netmask %s", inet_ntoa(get_mask(plen)));
-	else
+		break;
+	case INET_HEX:
 		printf(" netmask 0x%lx", (unsigned long)ntohl(get_mask(plen).s_addr));
+		break;
+	}
 
 	if ((link->ifi_flags & IFF_BROADCAST) && plen != 0)  {
 		struct sockaddr_in *brd = satosin(ifa->ifa_broadcast);
--- a/sbin/ifconfig/af_inet6.c
+++ b/sbin/ifconfig/af_inet6.c
@@ -85,7 +85,6 @@ static	int prefix(void *, int);
 #endif
 static	char *sec2str(time_t);
 static	int explicit_prefix = 0;
-extern	char *f_inet6, *f_addr;
 
 extern void setnd6flags(if_ctx *, const char *, int);
 extern void setnd6defif(if_ctx *,const char *, int);
@@ -246,10 +245,14 @@ print_p2p(struct sockaddr_in6 *sin)
 static void
 print_mask(int plen)
 {
-	if (f_inet6 != NULL && strcmp(f_inet6, "cidr") == 0)
+	switch (f_inet6) {
+	case INET6_CIDR:
 		printf("/%d", plen);
-	else
+		break;
+	case INET6_NUMERIC:
 		printf(" prefixlen %d", plen);
+		break;
+	}
 }
 
 static void
--- a/sbin/ifconfig/ifconfig.8
+++ b/sbin/ifconfig/ifconfig.8
@@ -173,15 +173,15 @@ Adjust the display of inet address subnet masks:
 .It Cm cidr
 CIDR notation, for example:
 .Ql 203.0.113.224/26
-.It Cm default
-Default format,
-.Cm hex
 .It Cm dotted
 Dotted quad notation, for example:
 .Ql 255.255.255.192
 .It Cm hex
 Hexadecimal format, for example:
 .Ql 0xffffffc0
+.It Cm default
+Default format,
+.Cm cidr
 .El
 .It Cm inet6
 Adjust the display of inet6 address prefixes (subnet masks):
@@ -192,12 +192,12 @@ CIDR notation, for example:
 .Ql ::1/128
 or
 .Ql fe80::1%lo0/64
-.It Cm default
-Default format,
-.Cm numeric
 .It Cm numeric
 Integer format, for example:
 .Ql prefixlen 64
+.It Cm default
+Default format,
+.Cm cidr
 .El
 .El
 .It Fl G Ar groupname
--- a/sbin/ifconfig/ifconfig.c
+++ b/sbin/ifconfig/ifconfig.c
@@ -90,7 +90,9 @@ int	exit_code = 0;
 static char ifname_to_print[IFNAMSIZ]; /* Helper for printifnamemaybe() */
 
 /* Formatter Strings */
-char	*f_inet, *f_inet6, *f_ether, *f_addr;
+inet_format_t f_inet;
+inet6_format_t f_inet6;
+char	*f_ether, *f_addr;
 
 #ifdef WITHOUT_NETLINK
 static void list_interfaces_ioctl(if_ctx *ctx);
@@ -313,10 +315,6 @@ cmpifaddrs(struct ifaddrs *a, struct ifaddrs *b, struct ifa_queue *q)
 static void freeformat(void)
 {
 
-	if (f_inet != NULL)
-		free(f_inet);
-	if (f_inet6 != NULL)
-		free(f_inet6);
 	if (f_ether != NULL)
 		free(f_ether);
 	if (f_addr != NULL)
@@ -344,10 +342,19 @@ static void setformat(char *input)
 			f_addr = strdup(modifier);
 		else if (strcmp(category, "ether") == 0)
 			f_ether = strdup(modifier);
-		else if (strcmp(category, "inet") == 0)
-			f_inet = strdup(modifier);
-		else if (strcmp(category, "inet6") == 0)
-			f_inet6 = strdup(modifier);
+		else if (strcmp(category, "inet") == 0) {
+			if (strcmp(modifier, "hex") == 0)
+				f_inet = INET_HEX;
+			else if (strcmp(modifier, "cidr") == 0)
+				f_inet = INET_CIDR;
+			else if (strcmp(modifier, "dotted") == 0)
+				f_inet = INET_DOTTED;
+		} else if (strcmp(category, "inet6") == 0) {
+			if (strcmp(modifier, "cidr") == 0)
+				f_inet6 = INET6_CIDR;
+			else if (strcmp(modifier, "numeric") == 0)
+				f_inet6 = INET6_NUMERIC;
+		}
 	}
 	free(formatstr);
 }
@@ -612,7 +619,7 @@ main(int ac, char *av[])
 		.io_s = -1,
 	};
 
-	f_inet = f_inet6 = f_ether = f_addr = NULL;
+	f_ether = f_addr = NULL;
 
 	lifh = ifconfig_open();
 	if (lifh == NULL)
--- a/sbin/ifconfig/ifconfig.h
+++ b/sbin/ifconfig/ifconfig.h
@@ -253,7 +253,21 @@ void	opt_register(struct option *);
 extern	ifconfig_handle_t *lifh;
 extern	int allmedia;
 extern	int exit_code;
-extern	char *f_inet, *f_inet6, *f_ether, *f_addr;
+
+typedef enum {
+	INET_CIDR = 0,
+	INET_HEX,
+	INET_DOTTED
+} inet_format_t;
+extern inet_format_t f_inet;
+
+typedef enum {
+	INET6_CIDR = 0,
+	INET6_NUMERIC
+} inet6_format_t;
+extern inet6_format_t f_inet6;
+
+extern	char *f_ether, *f_addr;
 
 void	clearifcap(if_ctx *ctx, const char *, int value);
 void	setifcap(if_ctx *ctx, const char *, int value);


llfw:
arch@ thread https://lists.freebsd.org/archives/freebsd-arch/2024-May/000653.html

jhowie:
You are right, netmasks have not been used on the Internet for quite some time, but you are ignoring the vast number of companies that still use /8, /16. and /32 on their RFC 1918 address-space networks. This change could be a significant issue for many of them, and breaks your premise of "most people will be familiar".  I recommend against the change, especially as cidr is already supported in ifconfig with the -f flag

llfw:
could you explain how this would be an issue for people who use /8, /16 or /32 addresses?  i use 10/8 in our local network which i subnet into /24s, and i've had this IFCONFIG_FORMAT set in login.conf since forever with no issues.

note that that this doesn't prevent someone using 'ifconfig em0 10.1.1.1 netmask 255.255.255.0' if they want to do that, and it won't break anyone's /etc/rc.conf.  it only affects the printed output format.

jhowie:
Hi Lexi,

Any script that expects the current format will be broken, and you are ignoring all the runbooks and incident handling guides that would have to be updated as a consequence of the change. That could be quite significant.

Best regards,

John


From: Lexi ***@***.***>
Date: Saturday, May 4, 2024 at 11:11 AM
To: freebsd/freebsd-src ***@***.***>
Cc: John Howie ***@***.***>, Comment ***@***.***>
Subject: Re: [freebsd/freebsd-src] ifconfig(8): change default IP address format to 'cidr' (PR #1216)

could you explain how this would be an issue for people who use /8, /16 or /32 addresses? i use 10/8 in our local network which i subnet into /24s, and i've had this IFCONFIG_FORMAT set in login.conf since forever with no issues.

note that that this doesn't prevent someone using 'ifconfig em0 10.1.1.1 netmask 255.255.255.0' if they want to do that, and it won't break anyone's /etc/rc.conf. it only affects the printed output format.

llfw:
we can note this change in RELNOTES for 15.0-RELEASE for people to update their scripts.  i don't expect this change will be MFC'd (nor should it).

the number of people who rely on parsing 'netmask' in ifconfig(8) in a shell script is much, much smaller than the number of people who expect to see IP addresses printed in a modern format -- it's not 1996 anymore.

dag-erling:
This is a bad idea as it will break thousands of scripts and you can achieve the same result by adding a single line to your shell startup script:

```
export IFCONFIG_FORMAT=inet:cidr,inet6:cidr
```

MikeKarels:
I am in favor of this change; it should have been made long ago.  I have had IFCONFIG_FORMAT=inet:cidr in my environment for a long time.  I find it hard to believe that so many scripts will be broken, but they can trivially be fixed by setting IFCONFIG_FORMAT.  On the other hand, new users will not know about the option to use cidr, and the change to the default will benefit most of the people, who do not set the format and deal with the default.  If there is significant fallout, there is plenty of time before 15.0-RELEASE to consider reverting.

llfw:
added an UPDATING entry to the PR.

jhowie:
If you really think this change should be made, then IFCONFIG_FORMAT should be set globally to default to current output format. Anyone who wants cidr format output can remove it. That way you do not break everyone’s scripts and experience. Is that acceptable?Sent from my iPhoneOn May 4, 2024, at 12:15, MikeKarels ***@***.***> wrote:﻿
I am in favor of this change; it should have been made long ago.  I have had IFCONFIG_FORMAT=inet:cidr in my environment for a long time.  I find it hard to believe that so many scripts will be broken, but they can trivially be fixed by setting IFCONFIG_FORMAT.  On the other hand, new users will not know about the option to use cidr, and the change to the default will benefit most of the people, who do not set the format and deal with the default.  If there is significant fallout, there is plenty of time before 15.0-RELEASE to consider reverting.

rwgbsd:
The value IS a netmask, it is stored as a netmask, it should be displayed by default how it is stored and implemented.  CIDR is a notation, a notation that does not cover all possible values, it should ONLY be used if asked for.   As Jessica pointed out on the mailing list the current code incorrectly calculates a non contiguous netmask.  I have not seen any other ifconfig that does this, yes linux "ip address" command outputs it in cidr, but not there ifconfig, nor does windows 10.  Please do not do this.

llfw:
>  As Jessica pointed out on the mailing list the current code incorrectly calculates a non contiguous netmask.

can you show an example of a non-contiguous netmask configured on a FreeBSD system and the commands you used to configure it?  i was not able to achieve this when i tested it (see my reply to Jessica's mail).

llfw:
also, let me add that if FreeBSD does still support non-contiguous netmasks in some way, the correct fix for that is to remove support for non-contiguous netmasks, which have been obsolete for over 30 years.

ppaeps:
I agree that this is a good change, but there are an awful lot of scripts out there that will need to be fixed -- many of which haven't been touched in decades.  "ifconfig | awk" is a very common pattern.

If we're going to make this change, we should set IFCONFIG_FORMAT=inet:cidr,inet6:cidr in the default environment for 15.0 and change ifconfig in 16.0.

I don't think discontiguous netmasks are something to worry about in 2024.  I've only ever encountered them in load balancers, where you could direct traffic to each interface based on them.  And I can't even remember the last time I encountered one of these.

bsdimp:
I'll remind everybody that ifconfig has had IFCONFIG_FORMAT since
```
commit 7c2aa744374aa3449ad81f60852e74ad73d823e6
Author: Allan Jude <allanjude@FreeBSD.org>
Date:   Tue May 31 17:30:08 2016 +0000

    ifconfig(8) now supports some output formatting options
```
so we've already 7 years into this process. This is nothing new and the desire to migrate has been around a long time. As such, we can transition in 15, and put a release note into 14.1 release that's coming up and still give people years worth of notice. We don't need to give 4 or 5 years notice to do this in 16. That's too long and people will take their eye off the ball and it will be 17 or 18 before it happens. I've had several things I "gratuitously" had a long horizon on, only to forget about until it was too late (14.0 was supposed to have an awk that didn't decode hex number, for example).

Also, broken scripts are a 1 line fix away from working, so the impact is rather small, and that one line can be set globally for people that have a lot of these scripts. The fix can be placed into the release notes as well. In addition, there's plenty of time for us to get a feel on how many scripts there actually are here. We can back off to a longer time horizon if it proves to be too disruptive. It would be even better if ifconfig, or a similar tool, could produce easy to use in scripts data (like json via libxo). I'm surprised there's no libxo option to it today. Of course these would be a more intrusive change, but would insulate us from data change like we've done in nearly every other system.

So I agree with Mike Karels on this.

rwgbsd:
> > As Jessica pointed out on the mailing list the current code incorrectly calculates a non contiguous netmask.
> 
> can you show an example of a non-contiguous netmask configured on a FreeBSD system and the commands you used to configure it? i was not able to achieve this when i tested it (see my reply to Jessica's mail).

Something (my best guess would be the netlink rewrites) has broken this:
`rgrimes_bgp.dnsmgr.net:root {1001}# uname -a
FreeBSD rgrimes_bgp.dnsmgr.net 13.2-RELEASE FreeBSD 13.2-RELEASE releng/13.2-n25
4617-525ecfdad597 GENERIC amd64
rgrimes_bgp.dnsmgr.net:root {1002}# ifconfig bridge2 create
rgrimes_bgp.dnsmgr.net:root {1003}# ifconfig bridge2 192.0.2.1 netmask 255.0.255
.0
rgrimes_bgp.dnsmgr.net:root {1004}# ifconfig bridge2
bridge2: flags=8843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST> metric 0 mtu 1500
        ether 58:9c:fc:10:ed:64
        inet 192.0.2.1 netmask 0xff00ff00 broadcast 192.255.2.255
        id 00:00:00:00:00:00 priority 32768 hellotime 2 fwddelay 15
        maxage 20 holdcnt 6 proto rstp maxaddr 2000 timeout 1200
        root id 00:00:00:00:00:00 priority 32768 ifcost 0 port 0
        groups: bridge
        nd6 options=9<PERFORMNUD,IFDISABLED>
`
`afbr1.dnsmgr.net:root {167}# uname -a
FreeBSD afbr1.dnsmgr.net 14.0-RELEASE FreeBSD 14.0-RELEASE #0 releng/14.0-n26538
0-f9716eee8ab4: Fri Nov 10 05:57:23 UTC 2023     root@releng1.nyi.freebsd.org:/u
sr/obj/usr/src/amd64.amd64/sys/GENERIC amd64
afbr1.dnsmgr.net:root {168}# ifconfig bridge2 create
afbr1.dnsmgr.net:root {169}# ifconfig bridge2 192.0.2.1 netmask 255.0.255.0
afbr1.dnsmgr.net:root {170}# ifconfig bridge2
bridge2: flags=1008843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST,LOWER_UP> metric 0
 mtu 1500
        options=0
        ether be:24:56:0f:fd:78
        inet 192.0.2.1 netmask 0xffff0000 broadcast 192.0.255.255
        id 00:00:00:00:00:00 priority 32768 hellotime 2 fwddelay 15
        maxage 20 holdcnt 6 proto rstp maxaddr 2000 timeout 1200
        root id 00:00:00:00:00:00 priority 32768 ifcost 0 port 0
        groups: bridge
        nd6 options=9<PERFORMNUD,IFDISABLED>
`

llfw:
> Something (my best guess would be the netlink rewrites) has broken this

has it "broken" this or has it *fixed* this by not allowing non-contiguous netmasks?  like i said - these have been obsolete for over 30 years.  it's time to move on.  CIDR is your frient.

rwgbsd:
> > Something (my best guess would be the netlink rewrites) has broken this
> 
> has it "broken" this or has it _fixed_ this by not allowing non-contiguous netmasks? like i said - these have been obsolete for over 30 years. it's time to move on. CIDR is your friend.

Silently doing something different than what the user asked to be done is "broken!"

llfw:
> Silently doing something different than what the user asked to be done is "broken!"

that's a fair point but not directly related to this PR - i'd be happy to submit another one to produce a warning/error when trying to add non-contiguous netmasks. i'm not sure if that's better done in ifconfig(8) or in the kernel, though -- probably the latter.

rwgbsd:
Another data point, people are asserting that CIDR obsoleted non contiguous netmask, yet the IETF documents do NOT refect this fact.  RFC950 was NOT updated by the CIDR RFC's, It has only ever been updated by RFC6918 which was to obsolete some ICMP, and with respect to netmask it simply obsoletes the ICMP request/reply mask icmp types.  So as it stands today,  the IETF standards (RFC950 is part of STD5) out weight anything in an RFC or BCP.

llfw:
what IGPs can carry non-contiguous netmasks?

what OSs can configure non-contiguous netmasks on an interface address?  (FreeBSD cannot, as we already established.)

what practical purpose is there to configure a non-contiguous netmask on an interface address?

is anyone here actually using a non-contiguous netmask on an interface address in a production system?

rwgbsd:
> what IGPs can carry non-contiguous netmasks?
As far as I am aware ALL ipv4 IGP's carry routes as Address:Netmask pairs, they do NOT carry this data as a prefix length.  I also know that OSPFv3 for IPv6 did infact do away with the netmask and now carries a prefix length in LSA's.   BGP4 also uses prefix length.

> 
> what OSs can configure non-contiguous netmasks on an interface address? (FreeBSD cannot, as we already established.)

That would need time to investigate, and FreeBSD was NOT an intentional removal, someone borked some code doing some other change so I do not think we should be asserting that as a basis for this change.
> 
> what practical purpose is there to configure a non-contiguous netmask on an interface address?
> 
> is anyone here actually using a non-contiguous netmask on an interface address in a production system?
Good questions, which put those details in what I would call the known unknowns.

Crest:
I understand the temptation to have "nicer" output by default, but **please** resist. This **will** break untold numbers of scripts that parse the `ifconfig` output. This output line has been this way it for decades. If you want to improve the default output for human users please set the `IFCONFIG_FORMAT` environment variable and export it in the default shell profile files instead.

michaeldexter:
> I find it hard to believe that so many scripts will be broken, but they can trivially be fixed by setting IFCONFIG_FORMAT.

This came up on a recent Production User Call and yes, it would break many systems. Users have been required to parse the output for a long time, for better or for worse, and institutionalized that parsing.

If not mistaken, this is output is identical on macOS and the Linux kernel, systemd aside.

This syntax is highly-documented, existing in books, manuals, search results, "stack overflow results", blogs, and other formal and semi-formal locations.

One suggestion to find potential breakage in ports is to run:

grep -r -l ’^\. /etc/rc\.subr’ /usr/ports/*

This change sounds like NextBSD territory.

As for "it's trivial to fix", it is trivial to override the defaults on your systems. Having individual users push their personal configurations upstream is a very long, slippery slope and the adding of directory colors comes to mind. Plus syslog compression, the new default root shell... Please configure your system exactly as you like. Consider zsh on it, or BASH.

These changes feel like progress, but asking administrators to continually revise their working production workflows is not advancement. Different is not always better. New features are great but please only enable them by default when you have broad consensus.

Rephrased: Exactly what well-defined problem does this solve, given the unseen administrator cost involved in adjusting to the change?

igalic:
I just checked my own parser code, and it already transparently handled both, but I added test-case for future proofing: https://github.com/canonical/cloud-init/pull/5272

Crest:
The problem is that there're so many ad-hoc parsers for ifconfig's output in the wild because it's the least bad way to get this information unless you're prepared to deal with operating system specific ioctl()s, netlink sockets etc. Is "it looks prettier" a good enough reason to change the default *notation* to represent the same information? Heck I run the environment variable myself on my personal systems yet I would be **pissed** to learn about this change because some remote system is unreachable after an upgrade because a script blew up.

Has anyone checked how this interacts with `/etc/network.subr`?

bsdimp:
> Has anyone checked how this interacts with `/etc/network.subr`?

Looks like it will work, since it uses regexp to pick out the IP address. Anything trailing the X.Y.Z.W is ignored.

The point about the multiple ad-hoc parsers, though, is interesting. There's three different parsers in network.subr...

MikeKarels:
It appears that the consensus is that this change should not be made, and I think I have been convinced.  Setting IFCONFIG_FORMAT to inet:cidr in root's .profile and .login seems to be a reasonable compromise, is easy to undo, and is unlikely to affect scripts run by cron, etc.

ppaeps:
I support added IFCONFIG_FORMAT to root's default shell configuration.

dag-erling:
I would suggest placing it in the `default` section of `login.conf` instead.

MikeKarels:
I don't think .profile is sourced by shell scripts unless they do it manually; sh(1) says it is used on login.  However, I think login.conf is better, as it applies to all users.

bsdimp:
While the .profile isn't sourced for every shell script (only login shells), exported env variables will "infect" those scripts regardless of how we do it. But there's a big difference.

So setting it in login.conf vs .profile has the advantage of doing all users, and not assuming that the root user's shell is /bin/sh. tcsh, zsh, etc users will get this from login when they are exec'd.

So cron jobs won't be affected. /etc/rc and subshells won't be affected. But 'service netif start' even won't be affected because service scrubs the env before execing the commands. Jails won't necessarily be affected, which is likely right. sudo will be affected if the user does 'sudo -E' but not if they do 'sudo' which might be a little confusing.

so this will change the interactive experience w/o changing much the ad-hoc parsers of ifconfig, some of which have no problem with the change, others of which have a problem with the change.

I find this reasonable middle ground OK, and favor it over making the change in ifconfig, or making the behavior conditional on ifconfig being redirected. I'd honestly prefer just to take the plunge, but there's a lot of fear of this change (likely in excess of the actual breakage) and doing it this way would let us make a more radical change in 16 since the release notes for this can warn users to add th proper env to the top of any script that depends on it and/or advice for how to write a parser that doesn't matter (like coincidentally network.subr). Doing it more slowly would dampen the fear and loathing this suggestion brought up and puts everyone explicitly on notice which will cause many people to evaluate their scripts and look to a future where the output may change (in the future the argument that it's identical to other systems won't be given much weight: there's already lots of differences with other systems and this one detail may be the same, one can never absolutely depend on it). The argument that this is a personal preference being foisted on everybody I give zero weight to: it's another output change to make the output better, just like hundreds of others we've made over the years. As for the 'feel like progress but 'inconvenience me' argument, I'd counter that progress is sometimes necessary at the expense of 100% compatibility and to freeze things like they were in 1990 doesn't do us any good: it's purely an emotional argument that's not at all helpful to the discussion since it's devoid of any data and it's hard to counter the sneering tone with anything useful. I do find mildly persuasive the notion that proponents of this change haven't conducted enough of an audit to characterize even the types of parsers that would break, vs the types that would work, but the opponents haven't brought useful, actionable data either beyond 'wow! this feels like it would break a lit' when in fact it won't break everything... we don't know how many of these scripts there are, nor what percentage will break so it's hard to make a good judgement call from data and it devolves to gut feel, which is part of the fear and loathing this invoked. We need to get out of that somehow with data, but disagree on the best path forward (which is why I like this compromise: it will start to bite some scripts run on the command line where their misbehavior will be most likely to be noticed). You can yell at me over these views, but I think we can neither jump in feet first, or can we do nothing here.

bsdimp:
@MikeKarels @dag-erling  So help me out... I see setenv in the example default cap, but I don't see in the code to login and friends where that capability is obtained and processed. Nor do I see documented the right syntax to use. How would one go about setting this in login.cap? thanks!

MikeKarels:
I got it to work after consulting the login.conf man page a couple of times.  I changed the setenv line in the default entry to

         :setenv=BLOCKSIZE=K,IFCONFIG_FORMAT=inet\ccidr:\

and then ran "cap_mkdb /etc/login.conf".  \c is an escaped colon (not obvious!). I didn't look at the code, but it's probably mostly library code.

stesser:
Am 11.05.24 um 23:42 schrieb MikeKarels:
> I got it to work after consulting the login.conf man page a couple of times. I 
> changed the setenv line in the default entry to
> 
> |:setenv=BLOCKSIZE=K,IFCONFIG_FORMAT=inet\ccidr:\ |
> 
> and then ran "cap_mkdb /etc/login.conf". \c is an escaped colon (not obvious!). 
> I didn't look at the code, but it's probably mostly library code.

If you want to have it affect both INET and INET6, quotes are required in
addition to escaping the colon:

	:setenv=BLOCKSIZE=K,IFCONFIG_FORMAT='inet\ccidr,inet6\ccidr':\

dag-erling:
> If you want to have it affect both INET and INET6, quotes are required in addition to escaping the colon: :setenv=BLOCKSIZE=K,IFCONFIG_FORMAT='inet\ccidr,inet6\ccidr':\

I would suggest modifying `ifconfig(8)` to accept `cidr` as a shortcut for `inet:cidr:inet6:cidr` , if that's not already the case.

llfw:
i don't understand the suggestion to add it to login.conf -- won't this also break everyone's scripts, if that's a concern?

jhowie:
My suggestion was to make your changes, but put an entry into login.conf that defaults to current output. Those who want cidr output by default can remove the entry.

Sent from my iPhone

On May 12, 2024, at 18:32, Lexi ***@***.***> wrote:

﻿

i don't understand the suggestion to add it to login.conf -- won't this also break everyone's scripts, if that's a concern?

MikeKarels:
> i don't understand the suggestion to add it to login.conf -- won't this also break everyone's scripts, if that's a concern?

It won't break scripts that run from cron, etc, just those started from a login shell environment.  I think the primary goal is to use cidr for interactive use at least.

bsdimp:
> > i don't understand the suggestion to add it to login.conf -- won't this also break everyone's scripts, if that's a concern?
> 
> It won't break scripts that run from cron, etc, just those started from a login shell environment. I think the primary goal is to use cidr for interactive use at least.

Yes. It is to put any breakage that happens at a time when a human is watching.

Service net start will not be affected. Service cleans the end. /etc/rc.d/net start would be...

dag-erling:
> Service net start will not be affected. Service cleans the end. /etc/rc.d/net start would be...

We can get around that by either setting `IFCONFIG_FORMAT` at the top of `network.subr` or making sure it always passes an explicit `-f` to `ifconfig(8)` when it matters (and not where it doesn't).

bsdimp:
> > Service net start will not be affected. Service cleans the end. /etc/rc.d/net start would be...
> 
> We can get around that by either setting `IFCONFIG_FORMAT` at the top of `network.subr` or making sure it always passes an explicit `-f` to `ifconfig(8)` when it matters (and not where it doesn't).

Yea. As far as my audit went, nithing in network.subr is affected. Nothing looks for netmask and the ip addr parsers tolerate trailing garbage and seemed to work for me.

But it was moreby way of example rather thsn any known issue or concern. 

We likely should add a note to the release notes for 14.1 too future proof...

bsdimp:
des has landed his cdir alias fix.
So this is close to obe, and likely just needs the login.conf changes to be closed.

llfw:
> So this is close to obe, and likely just needs the login.conf changes to be closed

what is obe?

by login.conf changes, do you mean changing login.conf to use the old formats for inet/inet6 by default?

jhowie:
Yes. If you really must change ifconfig, put an entry in login.conf that defaults to the current/legacy output of netmask. I still think things will break for users, though, and so I continue to recommend to not make the change at all.Sent from my iPhoneOn May 16, 2024, at 09:10, Lexi ***@***.***> wrote:﻿

So this is close to obe, and likely just needs the login.conf changes to be closed

what is obe?
by login.conf changes, do you mean changing login.conf to use the old formats for inet/inet6 by default?

bsdimp:
> > So this is close to obe, and likely just needs the login.conf changes to be closed
> 
> what is obe?

Sorry, project jargon for "Overcome by events" meaning no longer relevant.

> by login.conf changes, do you mean changing login.conf to use the old formats for inet/inet6 by default?

@MikeKarels suggested that we just put the IFCONFIG_FORMAT=cidr in login.conf as the default entry, and there seems to be support for that in general.

llfw:
> @MikeKarels suggested that we just put the IFCONFIG_FORMAT=cidr in login.conf as the default entry, and there seems to be support for that in general.

i don't really understand this proposal.  if cidr is supposed to be the default, then why not make that the default in ifconfig and, if necessary, include the old format in login.conf temporarily (until 16.0 or whatever)?

i think having login.conf override the default format in ifconfig will be much more confusing to users, i would rather not make any change at all rather than that.  plus this still breaks users' scripts if those scripts run under login.conf, so it doesn't fix any of the objections raised here.

MikeKarels:
> > @MikeKarels suggested that we just put the IFCONFIG_FORMAT=cidr in login.conf as the default entry, and there seems to be support for that in general.
> 
> i don't really understand this proposal. if cidr is supposed to be the default, then why not make that the default in ifconfig and, if necessary, include the old format in login.conf temporarily (until 16.0 or whatever)?

The intent of the proposal is to default to cidr in interactive sessions, not in things like cron.  This was discussed earlier.  Changing ifconfig and then restoring the old default in login.conf wouldn't have much benefit and could break scripts run from cron.

> i think having login.conf override the default format in ifconfig will be much more confusing to users, i would rather not make any change at all rather than that. plus this still breaks users' scripts if those scripts run under login.conf, so it doesn't fix any of the objections raised here.

If the scripts are run in a login session and break, there is at least someone watching to notice.  It's not a perfect compromise, but it is a compromise.

llfw:
> If the scripts are run in a login session and break, there is at least someone watching to notice. It's not a perfect compromise, but it is a compromise.

but what's the first thing you do when you write a script to parse ifconfig(8) output?  - you run `ifconfig` in a terminal and look at its output.

then you write a script to parse that output, put it in periodic(8) or whatever, and suddenly it doesn't work and you have no idea why, until you investigate and find what's going on in `/etc/login.conf`.

i was prepared to break some scripts with this change on the basis that existing scripts can be fixed.  IMO, adding this to `/etc/login.conf` will cause users to write new, broken scripts forever.  this is extremely poor UX.

MikeKarels:
It's not a perfect compromise.  It might encourage people to write scripts for the cidr format.  It should certainly be documented in UPDATING and release notes.  Hopefully people who are confused will check ifconfig(8), where the environment variable is mentioned in the -f description.

llfw:
> It's not a perfect compromise. It might encourage people to write scripts for the cidr format. It should certainly be documented in UPDATING and release notes.

people who want to write scripts involving ifconfig are not going to read UPDATING or RELNOTES for previous releases.  why would they?  they're going to type "ifconfig" and see what its output look like. 

i really cannot describe how much i am opposed to this solution and i hope previous commentors like des@ will back me up on this.

if this solution ends up being committed, then i have made a terrible mistake submitting this PR.

llfw:
i am closing this PR because i don't want to risk the chance of the bad login.conf PR being committed with my name on it.

hhartzer:
I wanted to say that I'm in favor of this change.

I did find that this behavior would have broken salt-ssh, so I will submit a bug there tracking that issue.

PR #1218, opened 2024-05-04
linprocfs: Add support for proc/sysvipc/{msg,sem,shm}

ricardobranco777:
Add support for proc/sysvipc/{msg,sem,shm}

Tracking in https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=278812

To build:

1. `cd /sys/modules/linprocfs`
1. `make && doas make install && doas cp -f /boot/modules/linprocfs.ko /boot/kernel/`
1. `doas kldxref /boot/kernel/`
1. `doas umount -a -t linprocfs`
1. `doas kldunload -v linprocfs`
1. `doas kldload -v linprocfs`
1. `doas mount -t linprocfs linprocfs /compat/linux/proc/`

How I tested it (needs `ipcmk` from the `util-linux` package):

```
$ ipcmk --shmem 128K
Shared memory id: 131072

$ cat /compat/linux/proc/sysvipc/shm 
       key      shmid perms                  size  cpid  lpid nattch   uid   gid  cuid  cgid      atime      dtime      ctime                   rss                  swap
-264176145     131072  4644                131072 39330     0      0  1000  1000  1000  1000          0          0 1714846976                    -1                    -1

$ ipcmk --semaphore 7 
Semaphore id: 196608

$ cat /compat/linux/proc/sysvipc/sem
       key      semid perms      nsems   uid   gid  cuid  cgid      otime      ctime
1038146433     196608  1644          7  1000  1000  1000  1000          0 1714847009

$ ipcmk --queue
Message queue id: 131072

$ cat /compat/linux/proc/sysvipc/msg
       key      msqid perms      cbytes       qnum lspid lrpid   uid   gid  cuid  cgid      stime      rtime      ctime
 756788372     131072   644           0          0     0     0  1000  1000  1000  1000          0          0 1714847025

$ ipcs -a
Message Queues:
T           ID          KEY MODE        OWNER    GROUP    CREATOR  CGROUP                 CBYTES                 QNUM               QBYTES        LSPID        LRPID STIME    RTIME    CTIME
q       131072    756788372 --rw-r--r-- ricardo  ricardo  ricardo  ricardo             0            0                 2048            0            0 no-entry no-entry 20:23:45

Shared Memory:
T           ID          KEY MODE        OWNER    GROUP    CREATOR  CGROUP         NATTCH        SEGSZ         CPID         LPID ATIME    DTIME    CTIME
m       131072   -264176145 --rw-r--r-- ricardo  ricardo  ricardo  ricardo             0       131072        39330            0 no-entry no-entry 20:22:56

Semaphores:
T           ID          KEY MODE        OWNER    GROUP    CREATOR  CGROUP          NSEMS OTIME    CTIME
s       196608   1038146433 --rw-r--r-- ricardo  ricardo  ricardo  ricardo             7 no-entry 20:23:29

```

The filtering is the same as done by `ipcs`:
- msg: https://github.com/freebsd/freebsd-src/blob/main/usr.bin/ipcs/ipcs.c#L215
- shm: https://github.com/freebsd/freebsd-src/blob/main/usr.bin/ipcs/ipcs.c#L247
- sem: https://github.com/freebsd/freebsd-src/blob/main/usr.bin/ipcs/ipcs.c#L276

Format string adapted from:
- shm: https://github.com/torvalds/linux/blob/master/ipc/shm.c#L1857
- sem: https://github.com/torvalds/linux/blob/master/ipc/sem.c#L2468
- msg: https://github.com/torvalds/linux/blob/master/ipc/msg.c#L1349


sys/compat/linprocfs/linprocfs.c
```diff
@@ -126,6 +126,9 @@
 #define P2K(x) ((x) << (PAGE_SHIFT - 10))		/* pages to kbytes */
 #define TV2J(x)	((x)->tv_sec * 100UL + (x)->tv_usec / 10000)
 
+/* Value defined in sys/kern/sysv_sem.c */
```
emaste, line 129: Where?

ricardobranco777: > Where?

Sorry. It's `sys/kern/sysv_shm.c`:

https://github.com/freebsd/freebsd-src/blob/main/sys/kern/sysv_shm.c#L114

Fixed.

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2095,158 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+	int error;
+
+	sbuf_printf(sb,
+	    "%10s %10s %4s  %10s %10s %5s %5s %5s %5s %5s %5s %10s %10s %10s\n",
+	    "key", "msqid", "perms", "cbytes", "qnum", "lspid", "lrpid",
+	    "uid", "gid", "cuid", "cgid", "stime", "rtime", "ctime");
+
+	msgmni = msginfo.msgmni;
+	sz = sizeof(struct msqid_kernel) * msgmni;
+	msqids = malloc(sz, M_TEMP, M_NOWAIT);
```
bsdimp, line 2116: this looks racy. What if msqids are added after we read msgifno.msgmni? Can we detect that below?

bsdimp (changes requested):
A couple of poorly handled races I think?

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2095,158 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+		return (ENOMEM);
+
+	error = kernel_sysctlbyname(curthread, "kern.ipc.shmsegs", shmids, &sz,
+	    NULL, 0, 0, 0);
+	if (error != 0) {
+		free(shmids, M_TEMP);
+		return (error);
+	}
+
+	for (id = 0; id < shmmni; id++)
```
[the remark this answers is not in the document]
ricardobranco777: > same as above.

Fixed, thanks!

sys/compat/linprocfs/linprocfs.c
ricardobranco777: > this looks racy. What if msqids are added after we read msgifno.msgmni? Can we detect that below?

Fixed.

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2095,158 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+		return (ENOMEM);
+
+	error = kernel_sysctlbyname(curthread, "kern.ipc.msqids", msqids, &sz,
+	    NULL, 0, 0, 0);
+	if (error != 0) {
+		free(msqids, M_TEMP);
+		return (error);
+	}
+
+	for (id = 0; id < msgmni; id++)
```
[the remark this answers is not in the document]
ricardobranco777: > You should check against sz, otherwise we're reading random memory thing.

Fixed.

bsdimp:
I've had to revert this. There's a race and some issues with it that need to be resolved. Please either reopen with them fixed, or create a new review and reference this one in the first comment.

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2095,176 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+			    msqids[id].u.msg_lspid,
+			    msqids[id].u.msg_lrpid,
+			    msqids[id].u.msg_perm.uid,
+			    msqids[id].u.msg_perm.gid,
+			    msqids[id].u.msg_perm.cuid,
+			    msqids[id].u.msg_perm.cgid,
+			    msqids[id].u.msg_stime,
+			    msqids[id].u.msg_rtime,
+			    msqids[id].u.msg_ctime);
+
```
bsdimp, line 2151: These times need to be cast to (intmax_t) and %jd should be used to print them.

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2095,176 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+again:
+	msgmni = msginfo.msgmni;
+	sz = sizeof(struct msqid_kernel) * msgmni;
+	msqids = malloc(sz, M_TEMP, M_NOWAIT);
+	if (msqids == NULL)
+		return (ENOMEM);
+	if (msgmni != msginfo.msgmni) {
+		free(msqids, M_TEMP);
+		goto again;
+	}
```
bsdimp, line 2123: Why is msginfo.msgmni changing? It would be better to just reply on the sysctl to get the data. It will return ENOMEM when the data size is too small. The case where it's too big is now handled corrected.

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2095,176 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+			    IXSEQ_TO_IPCID(id, semids[id].u.sem_perm),
+			    semids[id].u.sem_perm.mode,
+			    semids[id].u.sem_nsems,
+			    semids[id].u.sem_perm.uid,
+			    semids[id].u.sem_perm.gid,
+			    semids[id].u.sem_perm.cuid,
+			    semids[id].u.sem_perm.cgid,
+			    semids[id].u.sem_otime,
+			    semids[id].u.sem_ctime);
+
```
bsdimp, line 2204: same as above: intmax_t cast.

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2095,176 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+	semmni = seminfo.semmni;
+	sz = sizeof(struct semid_kernel) * semmni;
+	semids = malloc(sz, M_TEMP, M_NOWAIT);
+	if (semids == NULL)
+		return (ENOMEM);
+	if (semmni != seminfo.semmni) {
+		free(semids, M_TEMP);
+		goto again;
+	}
+
```
bsdimp, line 2181: Same as above. I think removing this check and keying off ENOMEM in the sysctl return value is the cleaner way to deal with this since we don't seem to have a cleaner, race-free interface to extract this info.

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2095,176 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+	shmmni = shminfo.shmmni;
+	sz = sizeof(struct shmid_kernel) * shmmni;
+	shmids = malloc(sz, M_TEMP, M_NOWAIT);
+	if (shmids == NULL)
+		return (ENOMEM);
+	if (shmmni != shminfo.shmmni) {
+		free(shmids, M_TEMP);
+		goto again;
+	}
+
```
bsdimp, line 2235: same again race.

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2095,176 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+			    shmids[id].u.shm_lpid,
+			    shmids[id].u.shm_nattch,
+			    shmids[id].u.shm_perm.uid,
+			    shmids[id].u.shm_perm.gid,
+			    shmids[id].u.shm_perm.cuid,
+			    shmids[id].u.shm_perm.cgid,
+			    shmids[id].u.shm_atime,
+			    shmids[id].u.shm_dtime,
+			    shmids[id].u.shm_ctime,
+			    0, 0);	/* XXX rss & swp are not supported */
```
bsdimp, line 2262: same again intmax_t cast.

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2095,176 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+		free(shmids, M_TEMP);
+		return (error);
+	}
+	shmmni = sz / sizeof(struct shmid_kernel);
+
+	for (id = 0; id < shmmni; id++)
+		if ((shmids[id].u.shm_perm.mode & SHMSEG_ALLOCATED) != 0)
+			sbuf_printf(sb,
+			    "%10d %10lu  %4o %21lu %5u %5u  %5u %5u %5u %5u %5u %10ld %10ld %10ld %21d %21d\n",
+			    (int) shmids[id].u.shm_perm.key,
```
bsdimp, line 2248: the %21lu should be %21zu because this is a size_t.

bsdimp:
Also from kib: "The recheck is more problematic, I do agree.  Also, I realized one more
issue there: the data structures like msginfo are from sysvmsg.ko modules.
Then, the change make linprocfs depending on the sysv*.ko, but this is
not directly recorded with MODULE_DEPENDS().  It worked because GENERIC
compiles SysV support statically."

ricardobranco777:
> Also from kib: "The recheck is more problematic, I do agree. Also, I realized one more issue there: the data structures like msginfo are from sysvmsg.ko modules. Then, the change make linprocfs depending on the sysv*.ko, but this is not directly recorded with MODULE_DEPENDS(). It worked because GENERIC compiles SysV support statically."

We already have these at EOF:

```
MODULE_DEPEND(linprocfs, sysvmsg, 1, 1, 1);
MODULE_DEPEND(linprocfs, sysvsem, 1, 1, 1);
MODULE_DEPEND(linprocfs, sysvshm, 1, 1, 1);

```

sys/compat/linprocfs/linprocfs.c
ricardobranco777: > Same as above. I think removing this check and keying off ENOMEM in the sysctl return value is the cleaner way to deal with this since we don't seem to have a cleaner, race-free interface to extract this info.

What do you mean by "keying off ENOMEM" ?

PR #1219, opened 2024-05-05
ahc(4): Default to memory mapped IO

hpvb:
When this driver was written it made sense to make this default to off, but these days almost all BIOSses will do the right thing. Furthermore non-mmio communication only works on Intel architectures.

So lets default to allowing mmio, but not change the semantics of the `AHC_ALLOW_MEMIO` flag to not break existing installs. Also document the already existing `hint.ahc.<unit>.allow_memio`.

### DIFF
--- a/sys/dev/aic7xxx/ahc_pci.c
+++ b/sys/dev/aic7xxx/ahc_pci.c
@@ -138,7 +138,7 @@ ahc_pci_map_registers(struct ahc_softc *ahc)
 	struct	resource *regs;
 	int	regs_type;
 	int	regs_id;
-	int	allow_memio;
+	int	allow_memio = 1;
 
 	regs = NULL;
 	regs_type = 0;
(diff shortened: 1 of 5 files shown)


sys/dev/aic7xxx/ahc_pci.c
```diff
@@ -144,20 +144,23 @@ ahc_pci_map_registers(struct ahc_softc *ahc)
 	regs_type = 0;
 	regs_id = 0;
 
+	// Default to memio
+	allow_memio = 1;
+
```
bsdimp, line 149: This is better off as an initializer, imho. Also , we try not to do C++ comments. Tweaked before the push.

hpvb: Ah,.I didn't realize that initializers were allowed. I'll keep that in mind for the future, also the comment style, thanks!

PR #1221, opened 2024-05-05
ipfw: correctly report table manipulation errors

ttyva:
Let ipfw(8) report the actual error instead of the error that may have come from calling printf().

Without this, ipfw emits some confusing errors when, for example, stdout is redirected.

&#8203;# ipfw table 1 add 1.2.3.4 >/dev/null
&#8203;# ipfw table 1 add 1.2.3.4 >/dev/null
ipfw: Adding record failed: Inappropriate ioctl for device

### DIFF
--- a/sbin/ipfw/tables.c
+++ b/sbin/ipfw/tables.c
@@ -1037,6 +1037,9 @@ table_modify_record(ipfw_obj_header *oh, int ac, char *av[], int add,
 		}
 	}
 
+	/* Get real OS error */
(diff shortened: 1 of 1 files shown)


lytboris:
@ttyva 

This PR created regression: https://lists.freebsd.org/archives/freebsd-current/2025-October/009048.html

There is a differential to fix it: https://reviews.freebsd.org/D53050 . Looking forward for any comments there.

PR #1222, opened 2024-05-07
Fix memory corruption in C++ demangler

jart:
The __cxa_demangle_gnu3() and cpp_demangle_gnu3() functions segfault on various libcxxabi test cases due to a copy and paste error. This change fixes that.


emaste:
See also https://github.com/libcxxrt/libcxxrt/pull/34. It seems reasonable to bring in the small fix now and let more extensive changes (adding tests etc.) happen via the upstream projects.

PR #1223, opened 2024-05-08
netstat(1): increase width of Netif column

llfw:
The previous width of Netif (10 or 8) was too short for modern interface names; make it 12, which is long enough to display "epair0a.1000".

This came up in practice with genet(4) interfaces, since the base interface name is long enough that with the previous limit, VLAN identifiers would be truncated at 1 character in the IPv6 output: "genet0.100" becomes "genet0.1".

The width is now fixed, and doesn't depend on the address family, because there's no reason that length of the interface name would vary based on the AF.

### DIFF
--- a/usr.bin/netstat/nhops.c
+++ b/usr.bin/netstat/nhops.c
@@ -64,7 +64,7 @@
 #include "common.h"
 
 /* column widths; each followed by one space */
-#define WID_IF_DEFAULT		12	/* width of netif column */
+#define WID_IF_DEFAULT		(Wflag ? IFNAMSIZ : 12)	/* width of netif column */
 #ifndef INET6
 #define	WID_DST_DEFAULT(af) 	18	/* width of destination column */
 #define	WID_GW_DEFAULT(af)	18	/* width of gateway column */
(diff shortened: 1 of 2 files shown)


igalic (approved):

usr.bin/netstat/nhops.c
```diff
@@ -64,16 +64,15 @@
 #include "common.h"
 
 /* column widths; each followed by one space */
+#define WID_IF_DEFAULT		12	/* width of netif column */
 #ifndef INET6
 #define	WID_DST_DEFAULT(af) 	18	/* width of destination column */
 #define	WID_GW_DEFAULT(af)	18	/* width of gateway column */
-#define	WID_IF_DEFAULT(af)	(Wflag ? 10 : 8) /* width of netif column */
 #else
```
gmshake, line 71: I'm wondering if the *wide* mode should have `IFNAMSIZ` which is default to 16 as the width.

llfw: this is a good idea, i will re-submit the patch to work like this.

llfw:
updated to use IFNAMSIZ if `-W` is specified, per @gmshake.

bsdimp (approved):
Looks good to my eye and seems to address review comments.

gmshake:
> updated to use IFNAMSIZ if `-W` is specified, per @gmshake.

Please give me some time to have a test ...

gmshake:
Looks good to me. And @bsdimp has landed the commits .

PR #1224, opened 2024-05-09
rman updates/cleanup

ehem:
Upon examination it turned out `rman_reserve_resource_bound()` was completely unused.  This is distinctly troublesome since it serves to misdirect people who were instead looking for `RF_ALIGNMENT_LOG2()`.  As such nuke `rman_reserve_resource_bound()` and fold everything back into `rman_reserve_resource()`.

Also spot `rman_make_alignment_flags()` was rather out date and really needed an update.


ehem (commented):
I'm awfully tempted to turn this into an inline in `sys/sys/rman.h`.

ehem:
Take care of the one style issue the checker legitimately flagged.  The others are due to this being legacy and I wasn't aiming to do full update.

I'm still eyeing `rman_make_alignment_flags()` and thinking it should be `inline` instead.

ehem:
Actually, the other issue is: Should `DPRINTF()` be in non-debug kernels?   Seems it is mostly well-tested at this point.

share/man/man9/rman.9
```diff
@@ -296,31 +289,20 @@ The alignment requirement
...
-.Fa bound
-argument must be a power of two.
-It may be set to zero to specify no boundary restriction.
+Often the
+.Dv RF_ALIGNMENT_LOG2
+macro is used to specify alignment to a power-of-2 size, or
+.Fn rman_make_alignment_flags
+can be used to compute the
+.Fa flags
+value at runtime.
```
ehem, line 298: Note, `roff` isn't something I've worked with so this may or may not be right.  I'm kind of guessing, but this seemed to be right given what was here before.

ehem:
Rebased onto main.  Adjust to leave 2 more words in the man page behind.  I was thinking I was going to propose an adjustment which needed the resource manager functions marked constant.  Yet upon further checking seems this is unnecessary, though worthwhile.

bsdjhb (approved):
These all look good to me

PR #1225, opened 2024-05-10
kern/intr: remove support for passing trap frame as argument

ehem:
While otherwise a handy potential approach, getting the trap frame via the argument isn't documented and isn't supposed to be used.  With all uses removed, now remove support to end the mixed calling conventions.

Differential Revision: https://reviews.freebsd.org/D37688

### DIFF
--- a/sys/kern/kern_intr.c
+++ b/sys/kern/kern_intr.c
@@ -1346,8 +1346,8 @@ ithread_loop(void *arg)
  *
  * Input:
  * o ie:                        the event connected to this interrupt.
- * o frame:                     some archs (i.e. i386) pass a frame to some.
- *                              handlers as their main argument.
+ * o frame:                     the current trap frame.
+ *
  * Return value:
  * o 0:                         everything ok.
  * o EINVAL:                    stray interrupt.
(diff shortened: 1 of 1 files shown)


ehem:
This was last verified 12-18 months ago.  At that time there were 3 outstanding uses of this construct.  Now those 3 are gone and as long as no more were added this backwards compatibility from i386 can be removed.

ehem:
Last verified at ee8b0c436d7221c25e8be3c3fe1f9da78b9d5b16, so just under 12 months ago.  Hopefully no more snuck in, but there is a small chance.

bsdimp:
@markjdb or @bsdjhb I think this is good to go. Can I get a second OK from one of you?

mhorne (approved):
This is a good wart to remove.

You should note in the commit message the preferred API for accessing the trapframe from an interrupt handler: de-referencing `curthread->td_intr_frame`.

bsdimp:
OK. Trying to land it with updated scripts, so hopefully the author dates aren't changed by it....

ehem:
jhibbits approved it on Phabricator.

ehem:
@mhorne's commit message adjustment seems appropriate, though I suspect it is simpler to let you handle that at this point.

bsdimp:
Yea, And I screwed up the commit message change.

bsdimp:
Updated the comments so that this is documented and pushed another commit.
Sorry for not updating the commit message.

ehem:
Happens, not much can be done about such occurrences.

PR #1228, opened 2024-05-10
Remove COMPAT_FREEBSD4/5/6/7/9 from MINIMAL and FIRECRACKER kernel configurations

hhartzer:
PR: [231768](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=231768)

### DIFF
--- a/sys/amd64/conf/FIRECRACKER
+++ b/sys/amd64/conf/FIRECRACKER
@@ -57,11 +57,6 @@ options 	GEOM_RAID		# Soft RAID functionality.
 options 	GEOM_LABEL		# Provides labelization
 options 	EFIRT			# EFI Runtime Services support
 options 	COMPAT_FREEBSD32	# Compatible with i386 binaries
-options 	COMPAT_FREEBSD4		# Compatible with FreeBSD4
-options 	COMPAT_FREEBSD5		# Compatible with FreeBSD5
-options 	COMPAT_FREEBSD6		# Compatible with FreeBSD6
-options 	COMPAT_FREEBSD7		# Compatible with FreeBSD7
-options 	COMPAT_FREEBSD9		# Compatible with FreeBSD9
 options 	COMPAT_FREEBSD10	# Compatible with FreeBSD10
 options 	COMPAT_FREEBSD11	# Compatible with FreeBSD11
 options 	COMPAT_FREEBSD12	# Compatible with FreeBSD12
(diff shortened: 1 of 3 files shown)


bsdimp:
left a comment in the bug, but basically "think it's a great idea, but we should run this by arch@, I don't think anybody will object. Proactively state this won't affect rust and it should be smooth sailing"

bsdimp:
In an ideal world, we'd have the compat stuff be an autoload, but there's a few places where it's not just wrapping the older ABI to call the new APIs.

ehem (changes requested):
The MINIMAL configuration for i386 also features compatibility back to FreeBSD 4.

PowerPC also features compatibility back to FreeBSD 6 in its GENERIC, GENERIC64 and MPC85XXSPE configurations.

Overall seems reasonable to me.  Just need to make sure not to miss anything.

hhartzer:
Good catch on those ones I missed. The old [Phabricator review](https://reviews.freebsd.org/D17375) had a couple others as well.

Should be addressed now.

I've emailed arch@ about this topic.

ehem (approved):
Seems reasonable to me.  Concerns about things using older ABIs being less likely to have been checked for security issues is a significant concern.

bsdimp:
My sense of the arch@ thread is that there's not a lot of support for this.

hhartzer:
Yeah, I think you're right. I probably shouldn't have mentioned deleting these, although I'm really surprised there's such pushback on this. This isn't something like C89, C99, POSIX, etc. Just very end-of-lifed FreeBSD ABI support.

Do you think the changes in MINIMAL might be acceptable? And possibly FIRECRACKER?

brooksdavis:
It's worth talking to @cperciva about FIRECRACKER, but it's hard to imagine that any compat options beyond those required for rust/go support are needed there. Likewise I think removing from MINIMAL makes sense.

emaste:
> It's worth talking to @cperciva about FIRECRACKER, but it's hard to imagine that any compat options beyond those required for rust/go support are needed there. Likewise I think removing from MINIMAL makes sense.

Rust is 12+ https://github.com/rust-lang/rust/issues/89058
I'm not sure about Go but I think it's the same

cperciva:
I'm ok with removing these from the FIRECRACKER kernel.

bsdimp:
It's my read of the arch@ thread that there's no community support for removing these from GENERIC, so please drop that part of this change.

hhartzer:
Should be ready for re-review.

PR #1229, opened 2024-05-10
Add MYKERNEL to .gitignore

hhartzer:
This matches the handbook's suggestion to have custom kernels named `MYKERNEL`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ tags
 .clangd
 .ccls-cache
 sys/*/compile
+MYKERNEL


gmshake:
> This matches the handbook's suggestion to have custom kernels named `MYKERNEL`.

I would recommend against that. The file `.gitignore` is for developers and it is mostly useful to exclude obj / generated files those are not to be ( incidentally ) committed into SCM. I do not think committers will commit files such as `MYKERNEL` that is obviously should not be.

bsdimp:
I have mixed feelings about this.

My first reaction was "Wow! This is a good idea" so that git doesn't complain about what people are doing at our recommendation.

Then I thought "well, if they have a branch, then it will be hard to have a kernel config file name MYKERNEL on that branch" and though "oh, I don't like it".

So I think I'll close this soon, unless there's a good case made that I'm not currently seeing.

hhartzer:
I understand. It definitely cuts both ways. I'll close this out.

PR #1231, opened 2024-05-12
Resolve mandoc style error rpcgen.1 manual page

chrisdavidson:
Resolve mandoc style error useless macro Tn in rpcgen manual page

### DIFF
--- a/usr.bin/rpcgen/rpcgen.1
+++ b/usr.bin/rpcgen/rpcgen.1
@@ -51,14 +51,11 @@
 .Sh DESCRIPTION
 The
 .Nm
-utility is a tool that generates C code to implement an
-.Tn RPC
-protocol.
+utility is a tool that generates C code to implement an RPC protocol.
 The input to
 .Nm
-is a language similar to C known as
-.Tn RPC
-Language (Remote Procedure Call Language).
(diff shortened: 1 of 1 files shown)


concussious:
I'm beginning to understand that nobody is writing scripts to automate zaping deprecated macros to hush the linter because it's a sign the page has aged a lot and could use some tlc.

It seems like we always put a .\"- at the top of the copyright, and a blank comment after the copyright. We can't tag SPDX because there's no license.

RPC could be spelled out in the document description without going over one line, the manual search engine already parses the term rpc from the name section, so that would promote visibility. The other rpc pages are also short enough.

There's a ton more lines that could be joined here. I like to break them at the commas for grep.

I think it's really clean to combine Pq and Xr onto the same line, that's more standard compliant, legible, and also highlights nicely on github.

Some of these Em's need to be Dv's or Ql's. The point of the mdoc(7) language is that we mark it up semantically so parsers can do things with it. There's always a better choice for Li and Em, but you have to evaluate each case.

At least rpc(5), rpcbind(8), and rpcinfo(8) should also go in the see also section.

I didn't tag these because github doesn't let me tap numbers that aren't immediately near what you've edited.

PR #1232, opened 2024-05-12
linprocfs: Add support for proc/sysvipc/{msg,sem,shm} v2

ricardobranco777:
Add support for proc/sysvipc/{msg,sem,shm}

Tracking in https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=278812

Retry of https://github.com/freebsd/freebsd-src/pull/1218


sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2095,158 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+		free(msqids, M_TEMP);
+		return (error);
+	}
+	msgmni = sz / sizeof(struct msqid_kernel);
+
+	for (id = 0; id < msgmni; id++)
+		if (msqids[id].u.msg_qbytes != 0)
+			sbuf_printf(sb,
+			    "%10d %10lu  %4o  %10lu %10lu %5u %5u %5u %5u %5u %5u %jd %jd %jd\n",
+			    (int) msqids[id].u.msg_perm.key,
```
kostikbel, line 2131: No space between ')' and 'm'

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2095,158 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+	size_t sz;
+	int error;
+
+	sbuf_printf(sb,
+	    "%10s %10s %4s  %10s %10s %5s %5s %5s %5s %5s %5s %10s %10s %10s\n",
+	    "key", "msqid", "perms", "cbytes", "qnum", "lspid", "lrpid",
+	    "uid", "gid", "cuid", "cgid", "stime", "rtime", "ctime");
+
+	sz = sizeof(struct msqid_kernel) * msginfo.msgmni;
+	msqids = malloc(sz, M_TEMP, M_NOWAIT);
```
kostikbel, line 2115: still M_NOWAIT. Why?

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2095,158 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+	u_long id, msgmni;
+	size_t sz;
+	int error;
+
+	sbuf_printf(sb,
+	    "%10s %10s %4s  %10s %10s %5s %5s %5s %5s %5s %5s %10s %10s %10s\n",
+	    "key", "msqid", "perms", "cbytes", "qnum", "lspid", "lrpid",
+	    "uid", "gid", "cuid", "cgid", "stime", "rtime", "ctime");
+
+	sz = sizeof(struct msqid_kernel) * msginfo.msgmni;
```
kostikbel, line 2114: msginfo is not available if sysvmsg module is not compiled into the kernel static.  You need to add dependency on sysvmsg module and check that linprocfs is still loadable even if the sysvmsg module is not statically compiled.

kostikbel (changes requested):

sys/compat/linprocfs/linprocfs.c
ricardobranco777: > msginfo is not available if sysvmsg module is not compiled into the kernel static. You need to add dependency on sysvmsg module and check that linprocfs is still loadable even if the sysvmsg module is not statically compiled.

There are 3 lines at EOF:

```
MODULE_DEPEND(linprocfs, sysvmsg, 1, 1, 1);
MODULE_DEPEND(linprocfs, sysvsem, 1, 1, 1);
MODULE_DEPEND(linprocfs, sysvshm, 1, 1, 1);
```

sys/compat/linprocfs/linprocfs.c
ricardobranco777: > still M_NOWAIT. Why?

Fixed.

sys/compat/linprocfs/linprocfs.c
ricardobranco777: > No space between ')' and 'm'

Fixed.

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2095,158 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+	msqids = malloc(sz, M_TEMP, M_WAITOK);
+	if (msqids == NULL)
```
kostikbel, line 2116: msqids cannot be NULL from malloc(M_WAITOK)

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2095,158 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+		return (ENOMEM);
+
+	error = kernel_sysctlbyname(curthread, "kern.ipc.msqids", msqids, &sz,
```
kostikbel, line 2119: This is quite strange way of fetching the data. For instance, if the 32bit process is issuing this sysctl, then the handler adapts and provide quite different 32bit structures instead of what you expect there.

sys/compat/linprocfs/linprocfs.c
kostikbel: ok, sorry

kostikbel (changes requested):

sys/compat/linprocfs/linprocfs.c
ricardobranco777: > This is quite strange way of fetching the data. For instance, if the 32bit process is issuing this sysctl, then the handler adapts and provide quite different 32bit structures instead of what you expect there.

What do you suggest in this case?

sys/compat/linprocfs/linprocfs.c
ricardobranco777: > msqids cannot be NULL from malloc(M_WAITOK)

Fixed.

sys/compat/linprocfs/linprocfs.c
kostikbel: I would consider refactoring the sysctls handlers to split them into functions that returns the array of descriptions (perhaps taking the bool compat32 arg or such), and handlers that call that functions. Then the linprocfs code can reuse the array functions.

I did not looked too close how easy such split is, or what other problems could arise.

ricardobranco777: > I would consider refactoring the sysctls handlers to split them into functions that returns the array of descriptions (perhaps taking the bool compat32 arg or such), and handlers that call that functions. Then the linprocfs code can reuse the array functions.
> 
> I did not looked too close how easy such split is, or what other problems could arise.

Too much hassle considering that `WARNING: 32-bit kernels are deprecated and may be removed in FreeBSD 15.0.`.

I installed the 32-bit version and will check what the issue is.

ricardobranco777: This compiled ok on i386.

kostikbel: compat32 is not going anywhere.  Problem is not with 32bit kernels, but with 32bit processes running on host (64bit) kernels.  The kernels with compat32 option adapt the result to bitness of the process that queried the sysctl, and the reply is no longer structured in the way you supposed.

kostikbel: The problem is not with binaries, but with kernel that internally (and rudely) issues these sysctls and expects that the result is always is in host format.  32bit binaries doing the sysctls are fine because they do want 32bit structures, and kernel code expected that only 32bit binaries do that, not kernel.

bsdimp:
This is dependent on kib's changes to export this data.

ricardobranco777:
https://reviews.freebsd.org/D45175

kostikbel (commented):
So you did not updated the PR to use the new functions.  I want to commit my change after it is verified somehow.

sys/compat/linprocfs/linprocfs.c
```diff
@@ -126,6 +126,9 @@
 #define P2K(x) ((x) << (PAGE_SHIFT - 10))		/* pages to kbytes */
 #define TV2J(x)	((x)->tv_sec * 100UL + (x)->tv_usec / 10000)
 
+/* Value defined in sys/kern/sysv_shm.c */
+#define SHMSEG_ALLOCATED	0x0800
```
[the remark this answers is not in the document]
ricardobranco777: > This should go into the header file; I will update my changes.

Fixed.

ricardobranco777:
> So you did not updated the PR to use the new functions. I want to commit my change after it is verified somehow.

Wouldn't it be easier to test by also patching the sysctl function called by ipcs?

kostikbel:
> > So you did not updated the PR to use the new functions. I want to commit my change after it is verified somehow.
> 
> Wouldn't it be easier to test by also patching the sysctl function called by ipcs?
No, the functions I provided are bad fit for the sysctls.  They do not handle COMPAT32, they allocate large chunk of memory which is not needed for sysctl.
> 
> Will test your changes and report back.
I will commit my bits after your confirmation.

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2092,149 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+	size_t sz;
+	int error;
+
+	sbuf_printf(sb,
+	    "%10s %10s %4s  %10s %10s %5s %5s %5s %5s %5s %5s %10s %10s %10s\n",
+	    "key", "msqid", "perms", "cbytes", "qnum", "lspid", "lrpid",
+	    "uid", "gid", "cuid", "cgid", "stime", "rtime", "ctime");
+
+	sz = sizeof(struct msqid_kernel) * msginfo.msgmni;
+	msqids = malloc(sz, M_TEMP, M_WAITOK);
```
kostikbel, line 2112: Why do you still malloc?  kern_get_XXXids() allocates the memory and returns the size in elements (in sz).

ricardobranco777: Fixed.

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2092,149 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+
+	error = kern_get_msqids(curthread, &msqids, &sz);
+	if (error != 0) {
+		free(msqids, M_TEMP);
+		return (error);
+	}
+	msgmni = sz / sizeof(struct msqid_kernel);
+
+	for (id = 0; id < msgmni; id++)
+		if (msqids[id].u.msg_qbytes != 0)
```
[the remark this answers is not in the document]
ricardobranco777: > if (... == 0) continue;
> 
> Then you can un-indent the printf below.

Fixed.

ricardobranco777:
> I will commit my bits after your confirmation.

It works for SHM.  I can see the same information in 32 & 64-bits Linux chroot.  It works both for /proc/sysvshm/shm & `ipcs`.  But I can't see MSG & SEM, even after I changed `security.bsd.see_other_uids`, `security.bsd.see_other_gids` & `security.bsd.see_jail_proc` to 1.  

```
$ sudo chroot /compat/ubuntu32 ipcs -a

------ Message Queues --------
key        msqid      owner      perms      used-bytes   messages    

------ Shared Memory Segments --------
key        shmid      owner      perms      bytes      nattch     status      
0xd7dda8b0 65536      1000       644        4096       0                       

------ Semaphore Arrays --------
key        semid      owner      perms      nsems     

$ sudo chroot /compat/ubuntu32 cat /proc/sysvipc/shm
       key      shmid perms                  size  cpid  lpid nattch   uid   gid  cuid  cgid      atime      dtime      ctime                   rss                  swap
-673339216      65536  4644                  4096 43204     0      0  1000  1000  1000  1000 0 0 1715879396      

$ sudo chroot /compat/ubuntu cat /proc/sysvipc/shm
       key      shmid perms                  size  cpid  lpid nattch   uid   gid  cuid  cgid      atime      dtime      ctime                   rss                  swap
-673339216      65536  4644                  4096 43204     0      0  1000  1000  1000  1000 0 0 1715879396                     0                     0

$ ipcs -a
Message Queues:
T           ID          KEY MODE        OWNER    GROUP    CREATOR  CGROUP                 CBYTES                 QNUM               QBYTES        LSPID        LRPID STIME    RTIME    CTIME   
q        65536    909169119 --rw-r--r-- ricardo  ricardo  ricardo  ricardo             0            0                 2048            0            0 no-entry no-entry 19:09:56

Shared Memory:
T           ID          KEY MODE        OWNER    GROUP    CREATOR  CGROUP         NATTCH        SEGSZ         CPID         LPID ATIME    DTIME    CTIME   
m        65536   -673339216 --rw-r--r-- ricardo  ricardo  ricardo  ricardo             0         4096        43204            0 no-entry no-entry 19:09:56

Semaphores:
T           ID          KEY MODE        OWNER    GROUP    CREATOR  CGROUP          NSEMS OTIME    CTIME   
s        65536  -1803737408 --rw-r--r-- ricardo  ricardo  ricardo  ricardo             1 no-entry 19:09:56

```

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2092,140 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+
+	sbuf_printf(sb,
+	    "%10s %10s %4s  %10s %10s %5s %5s %5s %5s %5s %5s %10s %10s %10s\n",
+	    "key", "msqid", "perms", "cbytes", "qnum", "lspid", "lrpid",
+	    "uid", "gid", "cuid", "cgid", "stime", "rtime", "ctime");
+
+	error = kern_get_msqids(curthread, &msqids, &sz);
+	if (error != 0)
+		return (error);
+	msgmni = sz / sizeof(struct msqid_kernel);
```
kostikbel, line 2114: sz already contains the count of elements, not the size. No need to delete by sizeof().  Same for all other places.

ricardobranco777: Fixed.

ricardobranco777: Now it works!  It shows the same information for all SYSVIPC mechanisms!

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2092,134 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+	error = kern_get_shmsegs(curthread, &shmids, &size);
+	if (error != 0)
+		return (error);
+
+	for (id = 0; id < size; id++) {
+		if ((shmids[id].u.shm_perm.mode & SHMSEG_ALLOCATED) == 0)
+			continue;
+		sbuf_printf(sb,
+		    "%10d %10lu  %4o %21zu %5u %5u  %5u %5u %5u %5u %5u %jd %jd %jd %21d %21d\n",
+		    (int) shmids[id].u.shm_perm.key,
```
kostikbel, line 2202: no space between ')' and 's'

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2092,134 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+	error = kern_get_sema(curthread, &semids, &size);
+	if (error != 0)
+		return (error);
+
+	for (id = 0; id < size; id++) {
+		if ((semids[id].u.sem_perm.mode & SEM_ALLOC) == 0)
+			continue;
+		sbuf_printf(sb,
+		    "%10d %10lu  %4o %10u %5u %5u %5u %5u %jd %jd\n",
+		    (int) semids[id].u.sem_perm.key,
```
kostikbel, line 2162: again no space

sys/compat/linprocfs/linprocfs.c
ricardobranco777: Fixed.

sys/compat/linprocfs/linprocfs.c
ricardobranco777: Fixed.

kostikbel (approved):

PR #1233, opened 2024-05-13
Avoid use after free in spa_export_common()

khorben:
Note that I am not familiar with this code; after a quick analysis this looks like what was initially intended.

This code is still found upstream at
https://github.com/openzfs/zfs/blob/master/module/zfs/spa.c#L7105.

Reported by:	Coverity
CID:		1545056


emaste:
Per the upstream issue it looks like this will be addressed as a side effect of https://github.com/openzfs/zfs/pull/16153

khorben:
> I'm inclined to just close this pull request since it's already upstream and pull requests aren't really for that stuff since that interferes with the vendor import process (or can).

Yes indeed, this has been addressed upstream a different way - closing now.

PR #1234, opened 2024-05-13
bhyve: avoid resource leak

khorben:
In bhyve_parse_config_option(), a string is allocated and passed to nvlist_add_string() but not free'd afterwards.

Reported by:	Coverity
CID:		1544049


bsdimp (approved):
So this fits the pattern elsewhere for the set_config_value and other strndup list.
nvlists appear to be allocating / freeing the storage it needs independently.
So this looks good to me.

PR #1236, opened 2024-05-15
sys/modules/dpdk_lpm4: do not build without INET

llfw:


llfw:
please MFC this to stable/14 and consider it for releng/14.1

bsdimp (approved):

PR #1238, opened 2024-05-16
cp: avoid a resource leak

khorben:
In copy_file(), make sure the from_fd file descriptor is closed even when the operation failed early.

Reported by:	Coverity Scan
CID:		1545036

### DIFF
--- a/bin/cp/utils.c
+++ b/bin/cp/utils.c
@@ -112,6 +112,8 @@ copy_file(const FTSENT *entp, int dne)
 		if ((from_fd = open(entp->fts_path, O_RDONLY, 0)) < 0 ||
 		    fstat(from_fd, &sb) != 0) {
 			warn("%s", entp->fts_path);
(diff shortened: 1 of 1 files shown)


bin/cp/utils.c
```diff
@@ -112,6 +112,8 @@ copy_file(const FTSENT *entp, int dne)
 		if ((from_fd = open(entp->fts_path, O_RDONLY, 0)) < 0 ||
 		    fstat(from_fd, &sb) != 0) {
 			warn("%s", entp->fts_path);
+			if (from_fd >= 0)
```
bsdimp, line 115: You could just close it since you're casting away the return value anyway.

emaste: I'd rather avoid calling close with an invalid fd, and am happy with the change as proposed

bsdimp: OK

PR #1239, opened 2024-05-16
libnvmf: avoid resource leak

khorben:
In nvmf_host_fetch_discovery_log_page(), the log variable may have been allocated on the heap during a first loop cycle, and should be free()'d again before exiting upon errors.

Reported by:	Coverity
CID:		1545034

### DIFF
--- a/lib/libnvmf/nvmf_host.c
+++ b/lib/libnvmf/nvmf_host.c
@@ -653,19 +653,23 @@ nvmf_host_fetch_discovery_log_page(struct nvmf_qpair *qp,
 	log = NULL;
 	for (;;) {
 		error = nvmf_get_discovery_log_page(qp, 0, &hdr, sizeof(hdr));
-		if (error != 0)
+		if (error != 0) {
+			free(log);
 			return (error);
+		}
 		nvme_discovery_log_swapbytes(&hdr);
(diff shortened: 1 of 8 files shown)


khorben:
Cc @bsdjhb

bsdimp (approved):
Looks fine..  borderline worth it because this usually is adjacent to an exitvpath, but it is in a library.

bsdimp:
Tagged with 'ready' but I'll give it until tomorrow for @bsdjhb to take. peek

bsdjhb:
The other error cases in the loop after the realloc all call free, so this is more consistent.

bsdjhb (approved):

bsdjhb:
Would maybe s/first loop cycle/previous loop iteration/ in the log message.  "again" might also be confusing as you aren't calling free multiple times, so I would perhaps omit it from the log message.

khorben:
> Would maybe s/first loop cycle/previous loop iteration/ in the log message. "again" might also be confusing as you aren't calling free multiple times, so I would perhaps omit it from the log message.

The `log` variable is always allocated in the first loop cycle initially (if there is more than one).
I have removed "again" from the log message.
Thanks!

PR #1244, opened 2024-05-17
bhyve: avoid side effect in assertion

khorben:
An assert() was setting the error variable instead of checking it.

Reported by:	Coverity Scan
CID:		1521431


bsdjhb (approved):

emaste:
f7d45c5443edc99857fdda19c68301b5ec4a8971

bsdimp:
Yea. I wish we could push this to the submitter's branch to get this marked as merged... but so far all my attempts to do this after git-freebsd push have failed on other pull requests

gmshake:
MFCed into stable/14 with 049b3c953e1e .

PR #1245, opened 2024-05-17
fix C++ compile errors (FreeBSD-current)

cnbatch:
Trying to fix the errors of [Bug 279054 - Compile failure when #include <netlink/netlink_snl.h> in C++ code (edit)](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=279054)

### DIFF
--- a/sys/netlink/netlink_snl.h
+++ b/sys/netlink/netlink_snl.h
@@ -63,7 +63,7 @@
 
 #define	_NLA_END(_start, _len)	((char *)(_start) + (_len))
 #define NLA_FOREACH(_attr, _start, _len)      \
-        for (_attr = (_start);		\
+        for (_attr = (struct nlattr *)(_start);		\
 		((char *)_attr < _NLA_END(_start, _len)) && \
 		((char *)NLA_NEXT(_attr) <= _NLA_END(_start, _len));	\
 		_attr =  NLA_NEXT(_attr))
@@ -80,7 +80,7 @@ struct linear_buffer {
 static inline struct linear_buffer *
 lb_init(uint32_t size)
 {
-	struct linear_buffer *lb = calloc(1, size);
+	struct linear_buffer *lb = (struct linear_buffer *)calloc(1, size);
 
 	if (lb != NULL) {
 		lb->base = (char *)(lb + 1);
@@ -102,7 +102,7 @@ lb_allocz(struct linear_buffer *lb, int len)
 	len = roundup2(len, alignof(__max_align_t));
 	if (lb->offset + len > lb->size)
 		return (NULL);
-	void *data = (void *)(lb->base + lb->offset);
+	char *data = (lb->base + lb->offset);
 	lb->offset += len;
 	return (data);
 }
@@ -275,7 +275,7 @@ snl_init(struct snl_state *ss, int netlink_family)
 	}
 
 	ss->bufsize = rcvbuf;
-	ss->buf = malloc(ss->bufsize);
+	ss->buf = (char *)malloc(ss->bufsize);
 	if (ss->buf == NULL) {
 		snl_free(ss);
 		return (false);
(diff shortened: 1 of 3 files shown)


sys/netlink/netlink_snl.h
```diff
@@ -63,7 +63,7 @@
 
 #define	_NLA_END(_start, _len)	((char *)(_start) + (_len))
 #define NLA_FOREACH(_attr, _start, _len)      \
-        for (_attr = (_start);		\
+        for (_attr = (typeof(_attr))(_start);		\
```
bsdimp, line 66: Does this work with C still?

sys/netlink/netlink_snl.h
```diff
@@ -104,7 +104,7 @@ lb_allocz(struct linear_buffer *lb, int len)
 		return (NULL);
 	void *data = (void *)(lb->base + lb->offset);
```
bsdimp, line 105: why not just make this char *?
Or change the return type of this to void *?

sys/netlink/netlink_snl.h
rilysh: typeof is a compiler extension, it was added in C23.

bsdimp: So no then. We don't enable c23 for the kernel at all.

cnbatch: changed to use `__typeof` instead

rilysh: Uh how about doing...
```cpp
#ifdef __cplusplus
# define NLA_FOREACH(_attr, _start, _len)      \
        for (_attr = (decltype(_start))(_start);	\
		((char *)_attr < _NLA_END(_start, _len)) && \
		((char *)NLA_NEXT(_attr) <= _NLA_END(_start, _len));	\
		_attr =  NLA_NEXT(_attr))
#else
# define NLA_FOREACH(_attr, _start, _len)      \
        for (_attr = (_start);		\
		((char *)_attr < _NLA_END(_start, _len)) && \
		((char *)NLA_NEXT(_attr) <= _NLA_END(_start, _len));	\
		_attr =  NLA_NEXT(_attr))
#endif
```

This one would be valid C and C++ code (without using any compiler extensions)... but the only issue (visually) is that two similar macros, one for C and one for C++

sys/netlink/netlink_snl.h
```diff
@@ -102,7 +102,7 @@ lb_allocz(struct linear_buffer *lb, int len)
 	len = roundup2(len, alignof(__max_align_t));
 	if (lb->offset + len > lb->size)
 		return (NULL);
-	void *data = (void *)(lb->base + lb->offset);
+	char *data = (lb->base + lb->offset);
 	lb->offset += len;
 	return ((char *)data);
```
rilysh, line 107: Since you've used char *data ... you can safely remove this cast

sys/netlink/netlink_snl.h
cnbatch: Splitting it into two macros might make future maintenance inconvenience.

How about I change it to `(struct nlattr *)` just like `NLA_NEXT` at line 62?
`NLA_FOREACH` macro uses `NLA_NEXT` internally, and `NLA_NEXT` is directly cast as `(struct nlattr *)`.

brooksdavis: > So no then. We don't enable c23 for the kernel at all.

`typeof()` should be fine in practice. Compiling the following fragment on compiler explorer says clang >= 11 and gcc >= 10 support it with default flags.

```C
int a;
typeof(a) b;
```

sys/netlink/netlink_snl.h
bsdimp: This looks to be the only active comment, and the only thing that has caught my eye that's in the way of committing.

bsdimp (approved):
I think this is now good

PR #1247, opened 2024-05-17
Fix C++ compile errors (FreeBSD Stable/14) 

cnbatch:
Trying to fix the errors of [Bug 279054 - Compile failure when #include <netlink/netlink_snl.h> in C++ code](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=279054)


bsdimp:
This duplicates #1245. FreeBSD's workflow also is that things go into main first, and then are merged to stable some time later. As such, we don't accept pull requests to the stable branch here unless there's something particularly difficult or unique about the changes being merged (we've had nothing like that here, but that mirrors practices we've done with phabricator).

PR #1248, opened 2024-05-17
linux: Support POSIX message queues

ricardobranco777:
This PR adds preliminary support for POSIX message queues.  It also adds the respective /proc/sys/mqueue entries to `linprocfs`.

Not supported:
- `mq_notify` with `SIGEV_THREAD_ID` as it must call a Linux function.

TODO:
- Make new sysctl's read-write?
- Support 32-bit on 64-bits?

Ref: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=279069

### DIFF
--- a/usr.sbin/arp/arp.c
+++ b/usr.sbin/arp/arp.c
@@ -144,7 +144,7 @@ main(int argc, char *argv[])
 	if (!func)
 		func = F_GET;
 	if (opts.rifname) {
-		if (func != F_GET && !(func == F_DELETE && opts.aflag))
+		if (func != F_GET && func != F_SET && !(func == F_DELETE && opts.aflag))
 			xo_errx(1, "-i not applicable to this operation");
 		if ((opts.rifindex = if_nametoindex(opts.rifname)) == 0) {
 			if (errno == ENXIO)
@@ -375,7 +375,7 @@ set(int argc, char **argv)
 		}
 	}
 #ifndef WITHOUT_NETLINK
-	return (set_nl(0, dst, &sdl_m, host));
+	return (set_nl(opts.rifindex, dst, &sdl_m, host));
 #else
 	return (set_rtsock(dst, &sdl_m, host));
 #endif


sys/compat/linux/linux_misc.c
```diff
@@ -2963,3 +2966,163 @@ linux_ioprio_set(struct thread *td, struct linux_ioprio_set_args *args)
...
+	if (l_flags & LINUX_O_DIRECTORY)
+		bsd_flags |= O_DIRECTORY;
+	if (l_flags & LINUX_O_PATH)
+		bsd_flags |= O_PATH;
+	/* XXX LINUX_O_NOATIME: unable to be easily implemented. */
+	return (bsd_flags);
+}
+
+/* The only flag is O_NONBLOCK */
+#define B2L_MQ_FLAGS(bflags)	((bflags != 0) ? LINUX_O_NONBLOCK : 0)
```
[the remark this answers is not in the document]
ricardobranco777: > #define B2L_MQ_FLAGS(bflags) ((bflags) != 0 ? LINUX_O_NONBLOCK : 0)

Do you suggest changing a tab for a space?

sys/compat/linux/linux_misc.c
```diff
@@ -2963,3 +2966,163 @@ linux_ioprio_set(struct thread *td, struct linux_ioprio_set_args *args)
 	}
 	return (error);
 }
+
+/* Taken from sys/compat/linux/linux_misc.c */
```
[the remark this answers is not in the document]
ricardobranco777: You're right. Fixed.

sys/compat/linux/linux_timer.c
```diff
@@ -43,8 +43,8 @@
 #endif
 #include <compat/linux/linux_time.h>
 
-static int
-linux_convert_l_sigevent(struct l_sigevent *l_sig, struct sigevent *sig)
+int
```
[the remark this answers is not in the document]
ricardobranco777: Fixed.

ricardobranco777:
It's passing the LTP & POSIX tests on both 64-bits & 32-bits except for:

- `mq_open01.c:165: TBROK: Failed to close FILE '/proc/sys/fs/mqueue/queues_max': EBADF (9)`
Mostly due not being able to open this file read-write.  Expected.
- `mq_notify03.c:57: TFAIL: mq_notify(m, &sev) failed: EINVAL (22)
mq_notify03.c:64: TFAIL: mq_notify(m, &sev) failed: EINVAL (22)`
Not supported calling Linux functions.  Expected.
- `mq_timedsend01.c:206: TFAIL: mq_timedsend() failed unexpectedly, expected SUCCESS: EMSGSIZE (90)
mq_timedsend01.c:206: TFAIL: mq_timedsend() failed unexpectedly, expected SUCCESS: EINVAL (22)`
This may be a bug in the FreeBSD implementation.

MISSING: Support for 32-bits on 64.

---

To test it I had to mount mqueuefs on a Ubuntu chroot created with `sudo debootstrap jammy /compat/ubuntu`.

```
sudo mkdir /compat/ubuntu/mnt/queue
sudo mount -t mqueuefs none /compat/ubuntu/mnt/queue
```

On an Ubuntu 22.04 VM I compiled LTP like this:

```
sudo apt install build-essential automake autoconf pkg-config m4
git clone https://github.com/linux-test-project/ltp
cd ltp
make autotools
./configure --with-open-posix-testsuite --with-realtime-testsuite
make -j8
sudo make install
```

Then copied `/opt/ltp` to the chroot and ran:
`for d in /opt/ltp/testcases/open_posix_testsuite/functional/mqueues testcases/open_posix_testsuite/conformance/interfaces/mq_* ; do cd $d ; ./run.sh ; cd - ; done`

Also `for i in /opt/ltp/testcases/bin/mq_* ; do $i ; done`

kostikbel (approved):

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2092,6 +2092,97 @@ linprocfs_domax_map_cnt(PFS_FILL_ARGS)
...
+/*
+ * Filler function for proc/sys/fs/mqueue/msg_default
+ */
+static int
+linprocfs_domqueue_msg_default(PFS_FILL_ARGS)
+{
+	int res, error;
+	size_t size = sizeof(res);
+
+	error = kernel_sysctlbyname(curthread, "kern.mqueue.default_maxmsg",
```
kostikbel, line 2104: Why going through the sysctl instead of just exporting the variables and directly using their values in place?

sys/kern/uipc_mqueue.c
```diff
@@ -2012,7 +2012,7 @@ kern_kmq_open(struct thread *td, const char *upath, int flags, mode_t mode,
 	struct pwddesc *pdp;
 	struct file *fp;
 	struct mqueue *mq;
-	int fd, error, len, cmode;
+	int fd, error, len, cmode, slash;
```
kostikbel, line 2015: Instead of adding the slash index, could you rename path to e.g. pathbuf, and have char *path pointing to the slash-less string?

kostikbel (commented):
I am not sure about wisdom of allowing '/' for FreeBSD native, although I do not object.  Can you think of (negative) consequences of changing native behavior?

sys/i386/linux/syscalls.master
```diff
@@ -1554,7 +1554,7 @@
 281	AUE_NULL	STD {
 		int linux_mq_notify(
 		    l_mqd_t mqd,
-		    const struct l_timespec *abs_timeout
+		    const struct sigevent *sevp
```
kostikbel, line 1557: Why sigevent and not l_sigevent?

kostikbel (commented):
Please remove generated diffs from the commit.

kostikbel (approved):

kostikbel (approved):

kostikbel (commented):
kern_ functions are normally put into sys/syscallsubr.h

sys/kern/uipc_mqueue.c
```diff
@@ -2267,58 +2267,77 @@ sys_kmq_setattr(struct thread *td, struct kmq_setattr_args *uap)
...
 {
 	struct mqueue *mq;
 	struct file *fp;
+	int error, waitok;
+
+	AUDIT_ARG_FD(mqd);
+	error = getmq_read(td, mqd, &fp, NULL, &mq);
+	if (error != 0)
+		return (error);
+	waitok = !(fp->f_flag & O_NONBLOCK);
```
kostikbel, line 2281: `waitok = (fp->f_flag & O_NONBLOCK) == 0`;

kostikbel (commented):
Same, please put kern_ functions into sys/syscallsubr.h

sys/i386/linux/syscalls.master
ricardobranco777: > Why sigevent and not l_sigevent?

Goog catch.  Btw, we have sigevent in sys/amd64/linux/syscalls.master for linux_timer_create.

sys/compat/linprocfs/linprocfs.c
ricardobranco777: > Why going through the sysctl instead of just exporting the variables and directly using their values in place?

The problem is that they have generic names like maxmsg.  I can change them to have a `mqueue_` prefix.  Also, I've seen this practice of calling sysctl in other places.

sys/kern/uipc_mqueue.c
ricardobranco777: > Instead of adding the slash index, could you rename path to e.g. pathbuf, and have char *path pointing to the slash-less string?

Fixed.

ricardobranco777:
> Please remove generated diffs from the commit.

You mean the ones generated by `make sysent` ?

ricardobranco777:
> I am not sure about wisdom of allowing '/' for FreeBSD native, although I do not object. Can you think of (negative) consequences of changing native behavior?

I don't see any.  In fact, I was surprised mq_open allowed paths starting with `//`.

kostikbel (approved):

kostikbel (approved):

kostikbel (approved):

sys/compat/linux/linux_misc.c
```diff
@@ -2963,3 +2966,124 @@ linux_ioprio_set(struct thread *td, struct linux_ioprio_set_args *args)
...
+}
+
+int
+linux_mq_unlink(struct thread *td, struct linux_mq_unlink_args *args)
+{
+	struct kmq_unlink_args bsd_args = {
+		.path = PTRIN(args->name)
+	};
+
+	return (sys_kmq_unlink(td, &bsd_args));
```
kostikbel, line 3002: Still you are using sys_kmq... instead of adding yet another kern_kmq function

sys/sys/syscallsubr.h
```diff
@@ -385,6 +386,15 @@ int	kern_writev(struct thread *td, int fd, struct uio *auio);
 int	kern_socketpair(struct thread *td, int domain, int type, int protocol,
 	    int *rsv);
 int	kern_unmount(struct thread *td, const char *path, int flags);
+int	kern_kmq_notify(struct thread *, int, struct sigevent *);
```
kostikbel, line 389: Please put functions by alphabetical order in the file

sys/compat/linux/linux_misc.c
ricardobranco777: > Still you are using sys_kmq... instead of adding yet another kern_kmq function

In this case we can because a pointer is the only argument mq_unlink() uses, and sys_kmq_unlink does the heavy-lifting well...

sys/sys/syscallsubr.h
ricardobranco777: > Please put functions by alphabetical order in the file

Fixed.

sys/compat/linux/linux_misc.c
```diff
@@ -29,13 +29,16 @@
  * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include "opt_posix.h"
```
kostikbel, line 32: Why is this header needed?

Code you changed previously wrapped sys_kmq_ calls into #ifdefs.  Either we need to be same careful there, or I do not see a use for the option.h

sys/compat/linux/linux_misc.c
```diff
@@ -2963,3 +2964,124 @@ linux_ioprio_set(struct thread *td, struct linux_ioprio_set_args *args)
...
+	flags = FFLAGS(flags);
+	if ((flags & O_CREAT) != 0 && args->attr != NULL) {
+		error = copyin(args->attr, &attr, sizeof(attr));
+		if (error != 0)
+			return (error);
+		attr.mq_flags = L2B_MQ_FLAGS(attr.mq_flags);
+	}
+
+	return (kern_kmq_open(td, args->name, flags, args->mode,
+	    args->attr != NULL ? &attr : NULL));
```
kostikbel, line 2990: What if the O_CREAT flag is not set, but args->attr is not NULL?  Wouldn't we pass a pointer to the stack garbage to kern_kmq_open?

ricardobranco777: > What if the O_CREAT flag is not set, but args->attr is not NULL? Wouldn't we pass a pointer to the stack garbage to kern_kmq_open?

sys_kmq_open also does this.  This function is based on sys_kmq_open.  Fortunately, kern_kmq_open does the right thing.

kostikbel (approved):

bsdimp:
I'm landing this... There's a minor whitespace conflict that I can ignore....

bsdimp:
```Building /usr/home/imp/obj/usr/home/imp/git/head/amd64.amd64/sys/GENERIC/modules/usr/home/imp/git/head/sys/modules/linux64/linux_sysvec.o
/usr/home/imp/git/head/sys/compat/linux/linux_misc.c:3048:12: error: no member named 'sevp' in 'struct linux_mq_notify_args'
 3048 |         if (args->sevp == NULL)
      |             ~~~~  ^
/usr/home/imp/git/head/sys/compat/linux/linux_misc.c:3051:24: error: no member named 'sevp' in 'struct linux_mq_notify_args'
 3051 |                 error = copyin(args->sevp, &l_ev, sizeof(l_ev));
      |                                ~~~~  ^
Building /usr/home/imp/obj/usr/home/imp/git/head/amd64.amd64/sys/GENERIC/modules/usr/home/imp/git/head/sys/modules/linuxkpi/linux_schedule.o
2 errors generated.
/usr/home/imp/git/head/sys/compat/linux/linux_misc.c:3048:12: error: no member named 'sevp' in 'struct linux_mq_notify_args'
 3048 |         if (args->sevp == NULL)
      |             ~~~~  ^
/usr/home/imp/git/head/sys/compat/linux/linux_misc.c:3051:24: error: no member named 'sevp' in 'struct linux_mq_notify_args'
 3051 |                 error = copyin(args->sevp, &l_ev, sizeof(l_ev));
      |                                ~~~~  ^
       35.91 real       282.14 user        21.16 sys```

ricardobranco777:
> ```
> /usr/home/imp/git/head/sys/compat/linux/linux_misc.c:3048:12: error: no member named 'sevp' in 'struct linux_mq_notify_args'
>  3048 |         if (args->sevp == NULL)
>       |             ~~~~  ^
> /usr/home/imp/git/head/sys/compat/linux/linux_misc.c:3051:24: error: no member named 'sevp' in 'struct linux_mq_notify_args'
>  3051 |                 error = copyin(args->sevp, &l_ev, sizeof(l_ev));
>       |                                ~~~~  ^
> Building /usr/home/imp/obj/usr/home/imp/git/head/amd64.amd64/sys/GENERIC/modules/usr/home/imp/git/head/sys/modules/linuxkpi/linux_schedule.o
> 2 errors generated.
> /usr/home/imp/git/head/sys/compat/linux/linux_misc.c:3048:12: error: no member named 'sevp' in 'struct linux_mq_notify_args'
>  3048 |         if (args->sevp == NULL)
>       |             ~~~~  ^
> /usr/home/imp/git/head/sys/compat/linux/linux_misc.c:3051:24: error: no member named 'sevp' in 'struct linux_mq_notify_args'
>  3051 |                 error = copyin(args->sevp, &l_ev, sizeof(l_ev));
>       |                                ~~~~  ^
>        35.91 real       282.14 user        21.16 sys```
> ```

You have to run `make sysent`.  I was told to remove the auto-generated entries.  I've no idea how these files would then be updated and make it to version control.

ricardobranco777:
> I'm landing this... There's a minor whitespace conflict that I can ignore....

Also fixed it. Hope I didn't interfere with any rebasing.

bsdimp:
For Phabricator reviews, the advise to remove auto generated files is good.
For pull requests, it's terrible (at least for now). They should be included, otherwise automated testing is impossible without a crazy amount of DWIM. I'm going to respectfully disagree with that bit of advice in this context. If we wanted to go to signed pull requests, say, then that would make those impossible to land since it will break the build guaranteed (though honestly we should be generating most of the files make sysent does on the fly, not as checked in artifacts, but I digress).

brooksdavis:
> For Phabricator reviews, the advise to remove auto generated files is good.

I'll disagree with this statement.  Phabricator hides the generated diff by default due to the `@generated` string so there is little value in not including them.  It's unfortunate that ignoring files matching certain patterns isn't a git feature yet.

bsdimp:
> > For Phabricator reviews, the advise to remove auto generated files is good.
> 
> I'll disagree with this statement. Phabricator hides the generated diff by default due to the `@generated` string so there is little value in not including them. It's unfortunate that ignoring files matching certain patterns isn't a git feature yet.

ah, I'll retract it for phab then. I'd forgotten it does that.

kostikbel:
> > > For Phabricator reviews, the advise to remove auto generated files is good.
> > 
> > 
> > I'll disagree with this statement. Phabricator hides the generated diff by default due to the `@generated` string so there is little value in not including them. It's unfortunate that ignoring files matching certain patterns isn't a git feature yet.
> 
> ah, I'll retract it for phab then. I'd forgotten it does that.

Autogenerated files make reading both reviews/PRs and actual commits awful *.  They should be changed in the separate commit.

In principle, such commit can be included into the series forming the PR, but then it leaves the possibility of conflicting with the changes in upstream.  I do not see any problem with committer doing `make sysent` and adding this last commit before pushing.

IMO this is the best procedure, and I insist on it due to (*).

bsdimp:
Needing to do a `make sysent` is a magic step. There are, sadly, several other autogenerated files in our tree. I don't want to play whack-a-mole with automation to catch them all. I strongly believe it should be included, as a separate commit, in the review chain. Otherwise, we grow an ever increasing amount of DWIM when doing automation which we're already having trouble getting going. Ideally, we'd need to commit none of these files to the tree. Most of them could be autogenerated today, if we hacked the generation tool (I have a GSoC student working on this) which would lessen the issue. But today we do need the commits and we do need them to keep our github actions working.

I think the tradeoff of giving feedback to the submitter that certain things are broken (which we need to do more of, I think) and an 'all green' so worth the time to go through the process of landing out-weighs the ugliness in the review of having to skip over generated files.

bsdimp:
(for this PR, though, I'm doing the make sysent and generating a commit to do the testing locally)

bsdimp:
I also wonder how the cross builds are showing 'green' when they don't build for me natively w/o the make sysent.

bsdimp:
Doh! my 'what to do when we lose the race' case of my automation failed to do the right thing, so this isn't showing up as merged.

PR #1249, opened 2024-05-20
mqueuefs: Do not call taskqueue_enqueue

ricardobranco777:
Fixes crash https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=278936

### DIFF
--- a/sys/kern/uipc_mqueue.c
+++ b/sys/kern/uipc_mqueue.c
@@ -85,7 +85,6 @@
 #include <sys/sysent.h>
 #include <sys/sx.h>
 #include <sys/sysctl.h>
-#include <sys/taskqueue.h>
 #include <sys/unistd.h>
 #include <sys/user.h>
 #include <sys/vnode.h>
@@ -788,7 +787,6 @@ found:
 	(*vpp)->v_data = vd;
 	vd->mv_vnode = *vpp;
 	vd->mv_node = pn;
-	TASK_INIT(&vd->mv_task, 0, do_recycle, *vpp);
 	LIST_INSERT_HEAD(&pn->mn_vnodes, vd, mv_link);
 	mqnode_addref(pn);
 	switch (pn->mn_type) {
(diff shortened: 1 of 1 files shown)


bsdimp:
I think we need an explanation about why this fixes that bug in the commit message.

ricardobranco777:
> I think we need an explanation about why this fixes that bug in the commit message.

I wish I knew that lol.

bsdimp:
> > I think we need an explanation about why this fixes that bug in the commit message.
> 
> I wish I knew that lol.

what made you try it then?

ricardobranco777:
> > > I think we need an explanation about why this fixes that bug in the commit message.
> > 
> > 
> > I wish I knew that lol.
> 
> what made you try it then?

The stack trace in https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=278936 contains this call.

bsdimp:
So, the task of the taskqueue was do_recyle:
```
/*                                                                                                                                                                                                                                                                              
 * task routine                                                                                                                                                                                                                                                                 
 */
static void
do_recycle(void *context, int pending __unused)
{
        struct vnode *vp = (struct vnode *)context;

        vn_lock(vp, LK_EXCLUSIVE | LK_RETRY);
        vrecycle(vp);
        VOP_UNLOCK(vp);
        vdrop(vp);
}
```
so we're "leaking" vnodes I think by not calling it since we never drop the hold we take out in the loop (at the very least) and we never recycle it.

I see there's a sx lock protecting things, but I don't see what protects a race with mqfs_reclaim. Once it gets the lock, it assumes that it can just remove the vd from the list without checking to see if the mqfs_node is still valid or not... So we call do_unlink, queue all this work up, drop the lock, then mqfs_reclaim runs, frees the node and then the task thread runs and BOOM since it's run on a now-free'd mqfs_vdata? And by removing the taskqueue queueing we avoid the panic?

kostikbel:
https://reviews.freebsd.org/D45305 is an attempt to provide the proper fix.

PR #1251, opened 2024-05-20
Update option_survey.sh

michaeldexter:
Increase image size from 4G to 5G to accommodate current OS. Tested on 14.1-BETA2.


emaste:
commit subject line should include a brief description of the change, so include e.g. `increase image size` or `increase image size to 5GB`. You can update the commit and push the update to the pull request (or if I'm going to land this I can just make the change).

michaeldexter:
By the way... syncing this with the VM-IMAGE size in /usr/src/release would solve this in the long term as that code gets much more attention than the build option survey.

michaeldexter:
FYI, builds reach 85% or so.

michaeldexter:
51 builds are complete, for those who celebrate: 

https://callfortesting.org/results/14.1-BETA2/

PR #1252, opened 2024-05-22
Fix "version introduced" in two manual pages

tomhukins:
I discovered these mistakes by running `git grep -F '.Fx 15'` against the `releng/14.1` branch.

Please merge these changes to `stable/14` and `releng/14.1`.


emaste:
Looks good. From the style checker
> Missing Signed-off-by: line

Can you push an updated commit message with that added?

We can definitely merge this to stable/14. I'm not sure if we'll be able to get it into 14.1 or not.

PR #1254, opened 2024-05-23
cross: Move Solaris API64 defines to common

bsdimp:
off64_t is needed for both Linux and MacOS, so move them to the common area.

For testing purposes only...


bsdimp:
So all green, that's a good sign.

bsdimp:
All the tests are green, except the Cirrus CI ones that always fail.

PR #1256, opened 2024-05-24
ovpn: "openvpn" group name concept is misplaced

fichtner:
The driver name upon device creation was "openvpn0" but the resulting ifconfig name is "opvn0".  The default group added was "openvpn" but since it's ovpn(4) keep consistency with all other network devices.

### DIFF
--- a/sys/net/if_ovpn.c
+++ b/sys/net/if_ovpn.c
@@ -252,7 +252,6 @@ SDT_PROBE_DEFINE2(if_ovpn, tx, route, ip4, "struct in_addr *", "struct ovpn_kpee
 SDT_PROBE_DEFINE2(if_ovpn, tx, route, ip6, "struct in6_addr *", "struct ovpn_kpeer *");
 
 static const char ovpnname[] = "ovpn";
-static const char ovpngroupname[] = "openvpn";
 
 static MALLOC_DEFINE(M_OVPN, ovpnname, "OpenVPN DCO Interface");
 #define	MTAG_OVPN_LOOP		0x6f76706e /* ovpn */
@@ -2454,7 +2453,7 @@ ovpn_clone_create(struct if_clone *ifc, char *name, size_t len,
 	ifp = sc->ifp;
 	ifp->if_softc = sc;
 	strlcpy(ifp->if_xname, name, IFNAMSIZ);
-	ifp->if_dname = ovpngroupname;
+	ifp->if_dname = ovpnname;
 	ifp->if_dunit = unit;
 
 	ifp->if_addrlen = 0;
@@ -2544,7 +2543,7 @@ vnet_ovpn_init(const void *unused __unused)
 		.create_f = ovpn_clone_create,
 		.destroy_f = ovpn_clone_destroy,
 	};
-	V_ovpn_cloner = ifc_attach_cloner(ovpngroupname, &req);
+	V_ovpn_cloner = ifc_attach_cloner(ovpnname, &req);
 }
 VNET_SYSINIT(vnet_ovpn_init, SI_SUB_PSEUDO, SI_ORDER_ANY,
     vnet_ovpn_init, NULL);


gmshake:
I guess that is intentional. CC @kprovost  .

bsdimp:
From kp@ when I asked him to look at this:

"Changing the default group for the interface means that users may get caught out by their firewall rules (pf rules can use both interface names or interface groups) suddenly not doing what they expect."

So I'm going to close this. It sounds like a bad idea and will break users setups. At the very least, that requires a conversation to reach consensus on the right way to solve this problem in the appropriate mailing list. At worst, this change is a bad one and shouldn't be made. Either way, it seems not an ideal pull request candidate since there's questions about whether it's a good idea at all, and no clear alternative. Once those are resolved, then code flowing from that resolution would be welcome here. I'm sorry.

fichtner:
I do disagree with the notion that this idiosyncrasy helps users use the group correctly. If it’s intended it should have been documented. At it stands at best this is a workaround for a closed source project that @kp is involved in.

fichtner:
Could we at least change the driver name to be consistent?

bsdimp:
I understand. However @kprovost has most of the github mail go to a spam filter, so this isn't the best place to debate the finer points of this change. I'm not sure if net@ or arch@ would be better for this discussion. I'm not saying "never," just that there's at least some disagreement between you and Kristof, so you two need to work it out in the larger community first rather than fighting this out in a commit war or via proxies like me that aren't close enough to this issue to have the context to know what's right.

fichtner:
Ok, so just to sum it up this patch making ovpn(8) behave like all other network devices in FreeBSD on creation is (rather indirectly) blocked by kp@ for _legacy_ reasons (November 20, 2023 is when this came out in 14.0) and possibly third party use.

As an outside contributor of the last decade think calling it a "commit war" all of a sudden is inappropriate and unconstructive.

In any case thanks for your consideration and review.

bsdimp:
> Ok, so just to sum it up this patch making ovpn(8) behave like all other network devices in FreeBSD on creation is (rather indirectly) blocked by kp@ for _legacy_ reasons (November 20, 2023 is when this came out in 14.0) and possibly third party use.
> 
> As an outside contributor of the last decade I think calling it a "commit war" all of a sudden is inappropriate and unconstructive.

Yea. That was a bit much. It was a bad attempt at a little humor that backfired into an insult, which I wasn't trying to do. I don't have the context here to judge this fine point. If I did commit and it was bad, it would be backed out. I want to avoid that. It was a bit much to describe that as a commit war. I'm sorry.

I don't know or understand all the reasons for Kristof's reasoning. Since it's an issue that seems like it needs a more public discussions than just a github pull request, i closed the pull request. 

I also wouldn't say blocked. I'd say this needs more discussion in some place like net@ to resolve. There is a clear difference of opinion that I think would benefit from the more specialized group at net@. I'm not familiar enough with the network stack to judge.

THINK ICMP redirect rather than a NAK.

Finally, the whole pull request process is new to freebsd and I'm still trying to work out the best norms here. If this isn't the best way to resolve issues like this, please let me know so we can refine in the future. I already have "be careful with hyperbole in my responses" on the list.

> In any case thanks for your consideration and review.

bsdimp:
Also, I forgot to mention, just this past week I landed something that I thought was good, no controversy, etc. Only come to discover afterwards I'd misread one of the comments as positive that was really negative, so I had to have a conversation to sort that out as well, which may have biased me to closing this sooner rather than later.

fichtner:
Thanks for your insights into the process here. Open source is challenging in general. :)

I might take this to net@ as `ifconfig -C` also reports "openvpn" and there is no good way in documentation to describe this (and have the user find it quickly) as well as UX compatibility with device workflows such as tun(4) that it aims to replace.

bsdimp:
> Thanks for your insights into the process here. Open source is challenging in general. :)
> 
> I might take this to net@ as `ifconfig -C` also reports "openvpn" and there is no good way in documentation to describe this (and have the user find it quickly) as well as UX compatibility with device workflows such as tun(4) that it aims to replace.

Please do. They'd be much better suited to know and understand the issues than I can here. Thanks!

PR #1259, opened 2024-05-27
remove vm_pageout's local variable %first

wy-chung:
It looks like that the local variable first will always be 0 so it is not needed.

### DIFF
--- a/sys/vm/vm_pageout.c
+++ b/sys/vm/vm_pageout.c
@@ -2362,23 +2362,21 @@ vm_pageout(void)
 {
 	struct proc *p;
 	struct thread *td;
-	int error, first, i, j, pageout_threads;
+	int error, i, j, pageout_threads;
 
 	p = curproc;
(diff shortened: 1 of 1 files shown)


sys/vm/vm_pageout.c
```diff
@@ -2362,23 +2362,21 @@ vm_pageout(void)
...
 			if (bootverbose)
 				printf("domain %d empty; skipping pageout\n",
 				    i);
 			continue;
 		}
-		if (first == -1)
-			first = i;
-		else {
+		if (i != 0) {
 			error = kthread_add(vm_pageout_worker,
```
bsdimp, line 2380: I don't think this is correct at all.
(a) first will be != 0 if VM_DOMAIN_EMPTY(0) is true.

wy-chung: You are right. I missed considering the case when vm domain 0 is empty.

PR #1260, opened 2024-05-28
linux: implement PR_CHILD_SET_SUBREAPER

sonphantrung:
This PR adds the [syscall](https://man.archlinux.org/man/prctl.2#PR_SET_CHILD_SUBREAPER), which is needed for process reaping stuffs. `PR_GET_CHILD_SUBREAPER` is currently quite quirky for me. Additionally, I've bumped the version as to reflect the change.


sys/compat/linux/linux_misc.c
```diff
@@ -1819,6 +1819,14 @@ linux_prctl(struct thread *td, struct linux_prctl_args *args)
 #endif
 		error = EINVAL;
 		break;
+	case LINUX_PR_SET_CHILD_SUBREAPER:
+		if (args->arg2 == 0){
+			return (kern_procctl(td, P_PID, 0, PROC_REAP_RELEASE,
+			    NULL));
+		};
```
bsdimp, line 1826: No ';' here.

lemul (approved):

sys/compat/linux/linux_misc.c
lemul: It's unnecessary here.

sys/sys/param.h
```diff
@@ -73,7 +73,7 @@
  * cannot include sys/param.h and should only be updated here.
  */
 #undef __FreeBSD_version
-#define __FreeBSD_version 1500018
+#define __FreeBSD_version 1500019
```
lemul, line 76: hmm, what is the reason for this?

sonphantrung: Because packages like `games/linux-steam-utils` will benefit from version bump (it has a workaround for avoiding Steam's `reaper` script [altogether](https://github.com/shkhln/linuxulator-steam-utils/blob/master/src/steamfix.c#L63))

lemul: > Because packages like `games/linux-steam-utils` will benefit from version bump (it has a workaround for avoiding Steam's `reaper` script [altogether](https://github.com/shkhln/linuxulator-steam-utils/blob/master/src/steamfix.c#L63))

in that case, it should be a separate comit.

sonphantrung: Alrighty, I've left the version bump for now.

sys/compat/linux/linux_misc.c
```diff
@@ -1819,6 +1819,14 @@ linux_prctl(struct thread *td, struct linux_prctl_args *args)
 #endif
 		error = EINVAL;
 		break;
+	case LINUX_PR_SET_CHILD_SUBREAPER:
+		if (args->arg2 == 0){
```
jrtc27, line 1823: We’re just going to ignore style(9) here?

bsdimp: > We’re just going to ignore style(9) here?

Please state the problem rather than assume intent.

Stlye(9) is a huge pain in the ass to get right all the time. My whiny script about didn't catch this lack of space and I didn't see it until you pointed it out just now.

jrtc27: My point was that your handy script *did* fail here:

image
image

jrtc27: But sorry for being grumpy, I was just annoyed that I saw the style issue in the commit in my inbox and that, when I went to check the PR, it had been clearly flagged by the automation

bsdimp: Somehow it wasn't flagged by the script I was using... Oh... I see what went wrong with my scripting...

PR #1261, opened 2024-05-28
export $autoboot var when running from $boottrace_cmd to fix root check root check if boottrace is enabled

samm-git:
At the moment, if bootrace profiling is enabled, autoboot is not exported to the rc scripts. This causes fsck to not check the root filesystem.  See https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=278993

### DIFF
--- a/libexec/rc/rc.subr
+++ b/libexec/rc/rc.subr
@@ -1856,7 +1856,8 @@ boottrace_fn()
 		set $_arg; . $_file
 		boottrace_sysctl "$_file done"
 	else
-		$boottrace_cmd "$_file" "$_arg"
+		_boot="${_boot}" rc_fast="${rc_fast}" autoboot="${autoboot}" \
+		    $boottrace_cmd "$_file" "$_arg"
 	fi
 }
 


bsdimp:
So part of the problem is that the boottrace stuff doesn't export any of the variables.  What about _boot or rc_fast? It looks like rc_fast is reconstructed whenver _boot is fast* in rc.subr and only set when _boot has been set to 'faststart'. The reason we have an issue is because the boottrace program is forked as a shell, while other the other startups are done with '.' which doesn't create a subshell.

The fix is a bit of a hack in that it it exports the one known bad variable for this one case. Is there a more generalized way to do it?

bsdimp:
(tagged with a combo that indicates maybe some more thought is needed here and a quick discussion about that).

samm-git:
> So part of the problem is that the boottrace stuff doesn't export any of the variables. What about _boot or rc_fast? It looks like rc_fast is reconstructed whenver _boot is fast* in rc.subr and only set when _boot has been set to 'faststart'. The reason we have an issue is because the boottrace program is forked as a shell, while other the other startups are done with '.' which doesn't create a subshell.
> 
> The fix is a bit of a hack in that it it exports the one known bad variable for this one case. Is there a more generalized way to do it?

It is actually a question of what is a bigger hack - fsck script that uses undocumented `$autoboot` or my variable substitution. I agree that the solution is not optimal, and that's why I am looking for advice. But now situation is like that - if boottrace is enabled and a hard reset is performed system is not able to boot at all, due to fsck never starting on the `/`.

samm-git:
another option i see is to modify bootrace(8) tool to be able to run 2 times, on start of the rc script and after end. This would still allow to use `.` command to load rc scripts but at the same time will enable benchmarking.

bsdimp:
I'm thinking that something more like
```
diff --git a/libexec/rc/rc b/libexec/rc/rc
index b23b0f35f263..53ba407ae0a2 100644
--- a/libexec/rc/rc
+++ b/libexec/rc/rc
@@ -57,6 +57,8 @@ fi

 _localbase=`/sbin/sysctl -n user.localbase 2> /dev/null`

+export autoboot _boot _localbase rc_fast
+
 dlv=`/sbin/sysctl -n vfs.nfs.diskless_valid 2> /dev/null`
 if [ ${dlv:=0} -ne 0 -o -f /etc/diskless ]; then
        sh /etc/rc.initdiskless
```
might be a better route.

bsdimp:
I'm not sure about the fast and loose means, so I'm @cperciva so he can look as swell.

bsdimp:
Ah, not Colin. I'm just going to add _boot and rc_fast to the list in case there's other bugs for these variables that are set in /etc/rc.

samm-git:
@bsdimp amazing, thank you :)

PR #1262, opened 2024-05-29
bhyve: correctly remove a memory mapping

khorben:
In a normal code path while setting up GPU passthrough, the size parameter to munmap() is wrong and its operation not checked for errors, therefore leaking resources.

Reported by:	Coverity Scan
CID:		1519830


bsdimp:
Looks good to my eye, and matches adjacent code. Pushing.

PR #1265, opened 2024-05-31
iwl-devtrace.c: remove leftover FBSDID

sonphantrung:
I found it when browsing the iwlwifi source code.


bsdimp:
I'm very sorry. However, I'm afraid that I can't land this pull request. It's to contrib code that I know is being actively updated (I've had private email with bz@, and landing it now would interefere with that. In addition, when deleting the $FreeBSD$ tag, often times sys/defs.h should be deleted as well next time you find something like this.

PR #1268, opened 2024-05-31
mdconfig: don't print NUL in the options list

asomers:
MFC with:	b01988d5e570158534110ca309fe93112c5c62f1
Sponsored by:	Axcient


asomers:
Currently if there aren't any options nothing will be printed.  The line will end with a tab character.  Do you think we should print `-` to be consistent with the other columns?

ricardobranco777 (approved):
LGTM. Thanks!

bsdimp:
Landing with a slightly tweaked commit message (Fixes: instead of MFC With:).

PR #1271, opened 2024-06-01
mdconfig: remove the "cluster" option.

asomers:
It's never had any effect.  The kernel ignores it.  Remove it from the documentation.  But continue to parse it on the command line, for backwards-compatibility.

### DIFF
--- a/sbin/mdconfig/mdconfig.8
+++ b/sbin/mdconfig/mdconfig.8
@@ -33,7 +33,7 @@
 .\" SUCH DAMAGE.
 .\" from: src/usr.sbin/vnconfig/vnconfig.8,v 1.19 2000/12/27 15:30:29
 .\"
-.Dd August 27, 2021
+.Dd June 1, 2024
 .Dt MDCONFIG 8
 .Os
 .Sh NAME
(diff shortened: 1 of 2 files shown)


bsdimp (approved):
Nice cleanup. Maybe, if you really want to, warn about this and #ifdef the detection < 16 or something... Or not and commit as is... (also, should I let you land this, or would you like me to land, I'm good either way).

asomers:
> Nice cleanup. Maybe, if you really want to, warn about this and #ifdef the detection < 16 or something... Or not and commit as is... (also, should I let you land this, or would you like me to land, I'm good either way).

Warn where, in the mdconfig binary?  And I'm comfortable landing it myself.

bsdimp:
> > Nice cleanup. Maybe, if you really want to, warn about this and #ifdef the detection < 16 or something... Or not and commit as is... (also, should I let you land this, or would you like me to land, I'm good either way).
> 
> Warn where, in the mdconfig binary? And I'm comfortable landing it myself.

Yes.  warnx("Option cluster ignored");

Cool. I'll let you do the landing then. No worries.

asomers:
Committed in [e461c7b](https://github.com/freebsd/freebsd-src/commit/e461c7bc7dde5dc84dc9e04cb623c6f567bfe6e5) .

PR #1277, opened 2024-06-05
loader: Increase buffer size to accommodate longer commands

VexedUXR:
The longest command we have is "efi-autoresizecons". That combined with the two spaces before and after the command gives us a total of 23 characters including the null-terminator.

Increasing this caused alignment issues since `efi-autoresizecons` is > 15 chars ("%-15s"). So I also increased that to 20 and removed the two spaces at the end since they aren't really needed.

Before:
```
Available commands:
  copy_staging     copy staging
  staging_slop     set staging slop
  efi-autoresizeconEFI Auto-resize Console <-- Cuts off
  gop              graphics output protocol
  uga              universal graphics adapter
  efi-seed-entropy try to get entropy from the EFI RNG
  poweroff         power off the system
  ...
```

After:
```
Available commands:
  copy_staging        copy staging
  staging_slop        set staging slop
  efi-autoresizecons  EFI Auto-resize Console
  gop                 graphics output protocol
  uga                 universal graphics adapter
  efi-seed-entropy    try to get entropy from the EFI RNG
  poweroff            power off the system
  ...
```

### DIFF
--- a/stand/common/commands.c
+++ b/stand/common/commands.c
@@ -229,7 +229,7 @@ command_commandlist(int argc __unused, char *argv[] __unused)
 {
 	struct bootblk_command	**cmdp;
 	int	res;
-	char	name[20];
+	char	name[23];
 
 	res = 0;
 	pager_open();
@@ -238,9 +238,10 @@ command_commandlist(int argc __unused, char *argv[] __unused)
 		if (res)
 			break;
 		if ((*cmdp)->c_name != NULL && (*cmdp)->c_desc != NULL) {
-			snprintf(name, sizeof(name), "  %-15s  ",
+			snprintf(name, sizeof(name), "  %-20s",
 			    (*cmdp)->c_name);
 			pager_output(name);
+			pager_output("  ");
 			pager_output((*cmdp)->c_desc);
 			res = pager_output("\n");
 		}


bsdimp (approved):
This is fine.. though it might be good to compute this dynamically ...

stand/common/commands.c
```diff
@@ -238,7 +238,7 @@ command_commandlist(int argc __unused, char *argv[] __unused)
 		if (res)
 			break;
 		if ((*cmdp)->c_name != NULL && (*cmdp)->c_desc != NULL) {
-			snprintf(name, sizeof(name), "  %-15s  ",
+			snprintf(name, sizeof(name), "  %-20s",
```
jrtc27, line 241: Previously the buffer was sized to exactly the right size (2 + 15 + 2 + 1). Now this is printing 2 + 20 + 1 = 23 characters, but the buffer is 24 characters long. Also, why is it correct to remove the trailing two spaces? You definitely want at least one to separate the name from the description. I imagine what you really want to do is put the spaces in their own pager_output calls so they don't get swallowed when a name is truncated.

VexedUXR: Sorry for the late response. It would definitely be better to put the trailing spaces in their own pager call, I didn't think of that.

As for why I removed the trailing spaces to begin with, it's because, as you said, they get truncated away anyways if the command is too long.

Oh, and I completely missed that buffer size thing.

Thanks!

netchild:
> This is fine.. though it might be good to compute this dynamically ...

Or at least a named constant instead of an unnamed value if the work involved to compute it dynamically provides an unreasonable return of investment.

VexedUXR:
> Or at least a named constant instead of an unnamed value if the work involved to compute it dynamically provides an unreasonable return of investment.

Ideally we'd only compute it once, so I don't think it would be too bad.

PR #1278, opened 2024-06-06
u3g: Add support for SIM7600G

0xThiebaut:
Add support for [SIM7600G](https://www.waveshare.com/wiki/SIM7600G-H-PCIE) LTE modem.


bsdimp (approved):
Looks good

kprovost:
Merged in https://cgit.freebsd.org/src/commit/?id=b5b90ff9844d1cb780ff777fc42ad393e4683563

0xThiebaut:
Thanks @kprovost ! First-timer question: As the intention is to get this into 14.1 (used by OPNsense) is this something that will happen automatically or should I have added an `MFC to:` tag?

kprovost:
14.1 was released, so it's too late to get this in.
It can end up in stable/14 and thus eventually in 14.2. We'd usually add an 'MFC after' tag if that was planned. That's mostly so the tooling reminds us though, so it's not too late to MFC this anyway.
I'll add a note to my todo list to MFC it next week.

PR #1279, opened 2024-06-06
Ctladm: misc fixes regarding port creation and deletion

asomers:
* Improve the man page regarding port creation and deletion
* Don't require "-p" when it isn't truly needed
* Print the port number of newly created frontend ports
* Print better error messages when a required argument is missing from a port creation or deletion request
* Add ATF tests

The motivation for these changes is https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=271460 .  Basically, I've gotten into a situation where the only way to recover was a reboot.  But if I'd known how to manually remove iscsi ports with ctladm, I could've recovered without a reboot.  Between the man page changes and better error messages, I think that future users in my position will be able to figure it out.

I elected to review this on Github instead of Phabricator so I can split it into multiple commits.  But if you prefer, I can go back to Phabricator and create separate reviews in sequence.

### DIFF
--- a/usr.sbin/ctladm/ctladm.8
+++ b/usr.sbin/ctladm/ctladm.8
@@ -673,8 +673,7 @@ and
 Specify the frontend port number.
 The port numbers can be found in the frontend port list.
 .It Fl r
-Remove port specified with
-.Pq Fl p Ar targ_port .
+Remove a port.
 .It Fl t Ar fe_type
 Specify the frontend type used by the
 .Pq Fl o ,
--- a/usr.sbin/ctladm/ctladm.8
+++ b/usr.sbin/ctladm/ctladm.8
@@ -35,7 +35,7 @@
 .\"
 .\" $Id: //depot/users/kenm/FreeBSD-test2/usr.sbin/ctladm/ctladm.8#3 $
 .\"
-.Dd June 5, 2024
+.Dd June 6, 2024
 .Dt CTLADM 8
 .Os
 .Sh NAME
@@ -166,7 +166,7 @@
 .Op Fl o Ar on|off
 .Op Fl w Ar wwpn
 .Op Fl W Ar wwnn
-.Op Fl O Ar pp|vp
+.Op Fl O Ar name=value
 .Op Fl p Ar targ_port
 .Op Fl r
 .Op Fl t Ar fe_type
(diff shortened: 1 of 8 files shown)


asomers:
cc @kendmerry

usr.sbin/ctladm/ctladm.8
```diff
@@ -644,13 +644,36 @@ If no port number or port type is specified, all ports are turned on or
...
+.Dq nvmf
+driver recognizes
+.Dq subnqn ,
+.Dq portid ,
+.Dq max_io_qsize ,
+.Dq enable_timeout ,
+.Dq ioccsz ,
+.Dq nn ,
+and
+.Dq serial .
```
asomers, line 671: These lines seem like candidates to move into cfiscsi(4) and nvmf(4).  But those man pages don't currently have any detail on this level, and there is no cfioctl(4).

usr.sbin/ctladm/ctladm.c
```diff
@@ -619,6 +619,20 @@ cctl_port(int fd, int argc, char **argv, char *combinedopt)
...
+				req.result_nvl = nvlist_unpack(result_buf, req.result_len, 0);
+				if (req.result_nvl == NULL) {
+					warnx("error unpacking result nvlist");
+					break;
+				}
+				created_port = nvlist_get_number(req.result_nvl, "port_id");
+				printf("Port created successfully\n");
+				printf("frontend: %s\n", driver);
+				printf("port:     %ju\n",
+				    (uintmax_t) created_port);
```
amotin, line 632: May be I would compress it into one message.

amotin (approved):

asomers:
Committed just now, along with mav's suggested change.

bsdimp:
FYI: You can get this to show up as merged by doing the following (assuming github the name of the github remote and freebsd is your FreeBSD source of truth one). It will also close automatically.
<make your changes to main, and get ready to push to `freebsd` like you normally would, but push like so instead>
```
git push github HEAD:github-branch
git push freebsd HEAD:main
```
if you lose the race, rebase and then try the two above commands again. It's critical that the commit be pushed to the github branch before freebsd-git pushes FreeBSD's main back to github for it to appear as 'Merged'. If the order is wrong, it's not the end of the world: it will autoclose and show up as 'Closed'.

Normally I do this dance for people whose repos are otherwise not writeable, but are due to the pull requests they've submitted. Since you have permissions for your own repo, you can do this.

asomers:
Ahh @bsdimp I'll try that next time.  I'm working on a related change to ctld.

bsdimp:
Thanks! I'm happy to help if you hit any snags with it, but I don't want it to be a large burden if it comes down to it...

PR #1280, opened 2024-06-07
Intrng: adjusting pic_init_secondary for further use cases

ehem:
Issue is `pic_init_secondary()` was implemented for root PICs-only.  The author's theory appears to have been the root calls the function on child PICs.  At least one secondary PIC needs the function called, but the device-tree doesn't have it as a child of the root PIC.  As such instead simply calling the function on all registered PICs appears to make more sense.

### DIFF
--- a/sys/arm/arm/gic.c
+++ b/sys/arm/arm/gic.c
@@ -205,6 +205,13 @@ arm_gic_init_secondary(device_t dev, uint32_t rootnum)
 	struct arm_gic_softc *sc = device_get_softc(dev);
 	u_int irq, cpu;
 
+	if (rootnum >= INTR_ROOT_COUNT) {
+		/*
+		 * Per-processor setup for devices with PPI interrupts?
+		 */
+		return;
+	}
+
 	/* Set the mask so we can find this CPU to send it IPIs */
 	cpu = PCPU_GET(cpuid);
 	MPASS(cpu < GIC_MAXCPU);
--- a/sys/arm/broadcom/bcm2835/bcm2836.c
+++ b/sys/arm/broadcom/bcm2835/bcm2836.c
@@ -543,6 +543,13 @@ bcm_lintc_init_secondary(device_t dev, uint32_t rootnum)
 	u_int cpu;
 	struct bcm_lintc_softc *sc;
 
+	if (rootnum >= INTR_ROOT_COUNT) {
+		/*
+		 * Per-processor setup for devices with PPI interrupts?
+		 */
+		return;
+	}
+
 	cpu = PCPU_GET(cpuid);
 	sc = device_get_softc(dev);
 
--- a/sys/arm64/arm64/gic_v3.c
+++ b/sys/arm64/arm64/gic_v3.c
@@ -1103,6 +1103,13 @@ gic_v3_init_secondary(device_t dev, uint32_t rootnum)
 	u_int cpu, irq;
 	int err, i;
 
+	if (rootnum >= INTR_ROOT_COUNT) {
+		/*
+		 * Per-processor setup for devices with PPI interrupts?
+		 */
+		return;
+	}
+
 	sc = device_get_softc(dev);
 	cpu = PCPU_GET(cpuid);
 
--- a/sys/arm64/arm64/gicv3_its.c
+++ b/sys/arm64/arm64/gicv3_its.c
@@ -1297,6 +1297,13 @@ gicv3_its_init_secondary(device_t dev, uint32_t rootnum)
 {
 	struct gicv3_its_softc *sc;
 
+	if (rootnum >= INTR_ROOT_COUNT) {
+		/*
+		 * Per-processor setup for devices with PPI interrupts?
+		 */
+		return;
+	}
+
 	sc = device_get_softc(dev);
 
 	/*
--- a/sys/kern/subr_intr.c
+++ b/sys/kern/subr_intr.c
@@ -1570,12 +1570,10 @@ dosoftints(void)
 void
 intr_pic_init_secondary(void)
 {
+	struct intr_pic *pic;
 	device_t dev;
 	uint32_t rootnum;
 
-	/*
-	 * QQQ: Only root PICs are aware of other CPUs ???
-	 */
 	//mtx_lock(&isrc_table_lock);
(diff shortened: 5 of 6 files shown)


ehem:
The conversion from `SLIST` to `STAILQ` is since `RISC-V` is expected to need the functionality and have it called on the first registered PIC before secondary ones.  The driver I have on-hand doesn't need this, but since it was brought up with D40474 I have no concern with this (which my case didn't need this, I foresaw others needing it).

There was an objection to the initial implementation, but no updates have been seen in Phabricator, so I'm dumping this on GitHub which seems better at getting these pushed through.

ehem:
This was simply rebased to aid keeping track of local branches, everything is identical to the original.  Notably the CI pass is still valid.

ehem:
Note to @jrtc27, you had pointed to an issue and objected in Phabricator.  This has a simple solution to the issue, but you never updated in Phabricator.  I'm unsure how GitHub handles mentions, but your review would be appropriate.

ehem:

ehem:
The issue in the final commit is still there.  The present mechanism is `intr_pic_init_secondary()` calls `PIC_INIT_SECONDARY()` then those calls need to propagate the call.  Unfortunately only the GICv3 driver attempts to do this and the default implementation is a NOP.  Additionally this requires all PICs to be presented as children of the root PIC and trees for some devices don't match this assumption.

As the comment in `intr_pic_init_secondary()` says,
> QQQ: Only root PIC is aware of other CPUs ???
Alas with theoretical objections to the proposal and no suggestions for alternatives, despite known real world cases, I'm rather stuck.

Good news is #1363 can be leveraged as an alternative for the case I'm looking at, but I imagine someone else is going to end up stuck here in the future.  Since I've got a workaround, I guess I have to leave this as a draft.

ehem:
Given how things are shaping up, this might be superior for my purpose.  Alas as the comments mention, what about setup for non-PIC devices with PPI interrupts?

ehem:
Huh.  Hadn't actually realized the extent of that dependency.  Yet it is there since needs `INTR_ROOT_COUNT` to be visible elsewhere to work.  I guess this is how this works.

ehem:
Since I need someone, how about @mhorne this time?  I was thinking you were planning for the first commit as part of #1363, but now it seems I need that anyway.

ehem:
Key phrase there "existing PIC drivers implementing `pic_init_secondary()`".  Which means if a PIC is a child of PIC driver which does not implement `pic_init_secondary()`, then that child won't have the call propagated to it.  This also assumes a device tree perfectly matching the model of all PIC-like devices in the PIC hierarchy.

You should be able to guess the one I'm thinking of.  While it would function to declare it as a root PIC, that is a bit ugly as the only root PIC functionality it uses is to hook into `pic_init_secondary()`.  Worse that then creates the ugly preprocessor mess trying to count the number of roots.  Having the type be `enum` works far better, but that was decided against.

In fact this is an issue for any device connected via PPI interrupt.  Any device connected via PPI might need to be informed as processors come online to know interrupts can now be sent to that processor.

Essentially the invalid root is for informing PICs/devices which don't currently receive `pic_init_secondary()`.  Alas this means we get the ugliness of PPI devices registering as PICs merely to capture `pic_init_secondary()`, but at least now they won't have to declare themselves a degenerate root PICs.

mhorne:
> Key phrase there "existing PIC drivers implementing `pic_init_secondary()`". Which means if a PIC is a child of PIC driver which does not implement `pic_init_secondary()`, then that child won't have the call propagated to it. This also assumes a device tree perfectly matching the model of all PIC-like devices in the PIC hierarchy.
> 
> You should be able to guess the one I'm thinking of. While it would function to declare it as a root PIC, that is a bit ugly as the only root PIC functionality it uses is to hook into `pic_init_secondary()`. Worse that then creates the ugly preprocessor mess trying to count the number of roots. Having the type be `enum` works far better, but that was decided against.
> 
> In fact this is an issue for any device connected via PPI interrupt. Any device connected via PPI might need to be informed as processors come online to know interrupts can now be sent to that processor.
> 
> Essentially the invalid root is for informing PICs/devices which don't currently receive `pic_init_secondary()`. Alas this means we get the ugliness of PPI devices registering as PICs merely to capture `pic_init_secondary()`, but at least now they won't have to declare themselves a degenerate root PIC.

Something is wonky about your model. If it is indeed a PIC, give it PIC methods, and if necessary, update it's parent PIC to propagate the `pic_init_secondary` call.

Otherwise, if it is just a matter of taking some action after all processors are online, use a custom `SYSINIT` after `SI_SUB_SMP`.

Sorry, but a strong NACK to what is here now. Closing.

ehem:
> Something is wonky about your model. If it is indeed a PIC, give it PIC methods, and if necessary, update it's parent PIC to propagate the `pic_init_secondary` call.

This particular device tree, all the parent PIC sees is a child with a PPI interrupt.  This seems appropriate as this device is a special PIC with limited functionality (no, I'm not about to lobby for a change to that device-tree).

> Otherwise, if it is just a matter of taking some action after all processors are online, use a custom `SYSINIT` after `SI_SUB_SMP`.

Seems hideous having to create a thread and then schedule it onto each processor in turn.  Guess it is the uglier approach then.

jrtc27:
> > Something is wonky about your model. If it is indeed a PIC, give it PIC methods, and if necessary, update it's parent PIC to propagate the `pic_init_secondary` call.
> 
> This particular device tree, all the parent PIC sees is a child with a PPI interrupt. This seems appropriate as this device is a special PIC with limited functionality (no, I'm not about to lobby for a change to that device-tree).
> 
> > Otherwise, if it is just a matter of taking some action after all processors are online, use a custom `SYSINIT` after `SI_SUB_SMP`.
> 
> Seems hideous having to create a thread and then schedule it onto each processor in turn. Guess it is the uglier approach then.

smp_rendezvous exists...

ehem:
Checking confirmed `smp_rendezvous()` did serve to fill the need.

Now I need someone to look at #1281 (one definite bug, several concerning bits) and #1285.  I only need the first two commits of #1285, but the problem is all the existing infrastructure matches that lower-level interface not the interface of a full hardware PIC.

Then #1358 needs a bunch of squashing/cleanup.

PR #1281, opened 2024-06-07
fix INTRNG interrupt release

ehem:
SSIA.

While it is good `intr_isrc_deregister()` exists, it hasn't ever actually been used.  Due to a driver which needs this, fix issues found by reviewing the source and finding solutions.

### DIFF
--- a/sys/kern/kern_intr.c
+++ b/sys/kern/kern_intr.c
@@ -530,8 +530,10 @@ int
 intr_event_destroy(struct intr_event *ie)
 {
 
-	if (ie == NULL)
-		return (EINVAL);
+	if (ie == NULL) {
+		printf("ERROR: %s(): passed NULL event!\n", __func__);
+		return (0);
+	}
 
 	mtx_lock(&event_lock);
 	mtx_lock(&ie->ie_lock);
--- a/sys/kern/subr_intr.c
+++ b/sys/kern/subr_intr.c
@@ -320,7 +320,18 @@ isrc_release_counters(struct intr_irqsrc *isrc)
 
 	mtx_assert(&isrc_table_lock, MA_OWNED);
 
+	if (isrc->isrc_count == NULL || isrc->isrc_index > nintrcnt) {
+		/* already cleared */
+		printf("WARNING: %s(): counter release called for \"%.*s\" with"
+		    " invalid counters\n", __func__,
+		    (int)sizeof(isrc->isrc_name), isrc->isrc_name);
+		return;
+	}
+
 	bit_nclear(intrcnt_bitmap, idx, idx + 1);
+
+	isrc->isrc_index = ~0;
+	isrc->isrc_count = NULL;
 }
 
 /*
--- a/sys/kern/subr_intr.c
+++ b/sys/kern/subr_intr.c
@@ -404,7 +404,8 @@ intr_isrc_dispatch(struct intr_irqsrc *isrc, struct trapframe *tf)
 			return (0);
 	} else
 #endif
-	if (isrc->isrc_event != NULL) {
+	{
+		MPASS(isrc->isrc_event != NULL);
 		if (intr_event_handle(isrc->isrc_event, tf) == 0)
 			return (0);
 	}
@@ -413,6 +414,113 @@ intr_isrc_dispatch(struct intr_irqsrc *isrc, struct trapframe *tf)
 		isrc_increment_straycount(isrc);
 	return (EINVAL);
 }
+
+/*
+ *  Interrupt source pre_ithread method for MI interrupt framework.
+ */
+static void
+intr_isrc_pre_ithread(void *arg)
+{
+	struct intr_irqsrc *isrc = arg;
+
+	PIC_PRE_ITHREAD(isrc->isrc_dev, isrc);
+}
+
+/*
+ *  Interrupt source post_ithread method for MI interrupt framework.
+ */
+static void
+intr_isrc_post_ithread(void *arg)
+{
+	struct intr_irqsrc *isrc = arg;
+
+	PIC_POST_ITHREAD(isrc->isrc_dev, isrc);
+}
+
+/*
+ *  Interrupt source post_filter method for MI interrupt framework.
+ */
+static void
+intr_isrc_post_filter(void *arg)
+{
+	struct intr_irqsrc *isrc = arg;
+
+	PIC_POST_FILTER(isrc->isrc_dev, isrc);
+}
+
+/*
+ *  Interrupt source assign_cpu method for MI interrupt framework.
+ */
+static int
+intr_isrc_assign_cpu(void *arg, int cpu)
+{
+#ifdef SMP
+	struct intr_irqsrc *isrc = arg;
+	int error;
+
+	mtx_lock(&isrc_table_lock);
+	if (cpu == NOCPU) {
+		CPU_ZERO(&isrc->isrc_cpu);
+		isrc->isrc_flags &= ~INTR_ISRCF_BOUND;
+	} else {
+		CPU_SETOF(cpu, &isrc->isrc_cpu);
+		isrc->isrc_flags |= INTR_ISRCF_BOUND;
+	}
+
+	/*
+	 * In NOCPU case, it's up to PIC to either leave ISRC on same CPU or
+	 * re-balance it to another CPU or enable it on more CPUs. However,
+	 * PIC is expected to change isrc_cpu appropriately to keep us well
+	 * informed if the call is successful.
+	 */
+	if (irq_assign_cpu) {
+		error = PIC_BIND_INTR(isrc->isrc_dev, isrc);
+		if (error) {
+			CPU_ZERO(&isrc->isrc_cpu);
+			mtx_unlock(&isrc_table_lock);
+			return (error);
+		}
+	}
+	mtx_unlock(&isrc_table_lock);
+	return (0);
+#else
+	return (EOPNOTSUPP);
+#endif
+}
+
+/*
+ *  Create interrupt event for interrupt source.
+ */
+static int
+isrc_event_create(struct intr_irqsrc *isrc)
+{
+	int error;
+
+	error = intr_event_create(&isrc->isrc_event, isrc, 0, INTR_IRQ_INVALID,
+	    intr_isrc_pre_ithread, intr_isrc_post_ithread, intr_isrc_post_filter,
+	    intr_isrc_assign_cpu, "%s:", isrc->isrc_name);
+
+	return (error);
+}
+
+#ifdef notyet
+/*
+ *  Destroy interrupt event for interrupt source.
+ */
+static void
+isrc_event_destroy(struct intr_irqsrc *isrc)
+{
+	struct intr_event *ie;
+
+	mtx_lock(&isrc_table_lock);
+	ie = isrc->isrc_event;
+	isrc->isrc_event = NULL;
+	mtx_unlock(&isrc_table_lock);
+
+	if (ie != NULL)
+		intr_event_destroy(ie);
+}
+#endif
 
 /*
  *  Alloc unique interrupt number (resource handle) for interrupt source.
@@ -447,6 +555,7 @@ isrc_alloc_irq(struct intr_irqsrc *isrc)
 	return (ENOSPC);
 
 found:
+	isrc->isrc_event->ie_irq = irq;
 	isrc->isrc_irq = irq;
 	irq_sources[irq] = isrc;
 
@@ -464,6 +573,7 @@ isrc_free_irq(struct intr_irqsrc *isrc)
 {
 
 	mtx_assert(&isrc_table_lock, MA_OWNED);
+	MPASS(isrc->isrc_event != NULL);
 
 	if (isrc->isrc_irq >= intr_nirq)
 		return (EINVAL);
@@ -512,10 +622,18 @@ intr_isrc_register(struct intr_irqsrc *isrc, device_t dev, u_int flags,
 	vsnprintf(isrc->isrc_name, INTR_ISRC_NAMELEN, fmt, ap);
 	va_end(ap);
 
+	error = isrc_event_create(isrc);
+	if (error)
+		return (error);
+
 	mtx_lock(&isrc_table_lock);
 	error = isrc_alloc_irq(isrc);
 	if (error != 0) {
+		int rc;
 		mtx_unlock(&isrc_table_lock);
+		if ((rc = intr_event_destroy(isrc->isrc_event)) != 0)
+			printf("ERROR: %s(): intr_event_destroy() rc = %d!\n",
+			    __func__, rc);
 		return (error);
 	}
 	/*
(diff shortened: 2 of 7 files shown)


ehem:
@markjdb prefers `intr_event_destroy()` returning an error for `ie == NULL`.  Problem is this is answering the **wrong** question for most callers.  Since this is called from cleanup/error paths, the caller wants to know `Has everything been properly cleaned up?`  Whereas returning `EINVAL` is answering `Has everything been cleaned up and was everything in **perfect** condition?`

I'm pretty sure most, if not all callers will want the answer to the first since `EBUSY` means memory is likely being leaked.  Whereas for INTRNG, the event being `NULL` simply means no event was ever allocated and things are fine.

ehem:
Putting this a different way, given a random `struct intr_irqsrc *isrc;`, the following **must** always terminate no matter the initial state of `isrc`:

```
while(intr_isrc_deregister(isrc)!=0)
    sleep(1);
```

(this would be a hypothetical PIC which implemented lazy release of interrupts)

If the filter hadn't yet been removed from the event, the first call to `intr_isrc_deregister()` would successfully release `isrc->isrc_irq`, but then return `EBUSY` from `intr_event_destroy()`.  For subsequent calls currently `isrc_free_irq()` would return `EINVAL` and the loop would never terminate.

ehem:
This was simply rebased to aid keeping track of local branches, everything is identical to the original.  Notably the CI pass is still valid.

sys/kern/subr_intr.c
```diff
@@ -528,7 +544,11 @@ intr_isrc_deregister(struct intr_irqsrc *isrc)
 		isrc_release_counters(isrc);
 	error = isrc_free_irq(isrc);
 	mtx_unlock(&isrc_table_lock);
-	return (error);
+
+	if (error != 0)
+		return (error);
+
+	return (isrc_event_destroy(isrc));
```
bsdimp, line 551: As noted above, this routine is never called. Why do we need to call isrc_event_destroy() here? That's new and not quite what the commit message says. How did you test this?

sys/kern/subr_intr.c
```diff
@@ -460,12 +460,11 @@ isrc_free_irq(struct intr_irqsrc *isrc)
...
-		return (EINVAL);
-	if (irq_sources[isrc->isrc_irq] != isrc)
-		return (EINVAL);
+		return (0);
 
-	irq_sources[isrc->isrc_irq] = NULL;
-	isrc->isrc_irq = INTR_IRQ_INVALID;	/* just to be safe */
+	if (irq_sources[isrc->isrc_irq] == isrc)
+		irq_sources[isrc->isrc_irq] = NULL;
+	isrc->isrc_irq = INTR_IRQ_INVALID;	/* mark removed */
```
bsdimp, line 468: We shouldn't be papering over programing errors.  It's a programming error when isrc_irq is >= intr_nirq. That normally can't happen. Likewise with setting things to NULL in this way. This is a bad change. NAK.

Hard to say for sure, though, since there's not actually any callers of this in the tree. This is only called from `intr_isrc_deregister`, which is global, but has no actual callers. So how did you test this?

But this is unsafe. If they don't match, it must be an error.

sys/kern/kern_intr.c
```diff
@@ -531,7 +531,7 @@ intr_event_destroy(struct intr_event *ie)
 {
 
 	if (ie == NULL)
-		return (EINVAL);
+		return (0);
```
bsdimp, line 534: This change is wrong. We want this to be an error. It's a programming error to call this with NULL.

sys/kern/subr_intr.c
```diff
@@ -479,6 +478,23 @@ isrc_free_irq(struct intr_irqsrc *isrc)
...
+/*
+ *  Destroy interrupt event for interrupt source.
+ */
+static int
+isrc_event_destroy(struct intr_irqsrc *isrc)
+{
+	int rc;
+
+	MPASS(isrc->isrc_irq >= intr_nirq);
+
```
bsdimp, line 490: So this is backwards. MPASS(ex...) means ex must be true. However, this can never be true (since IRQs are necessarily smaller than intr_nirq. So this will always panic. Again, how did you test this?

bsdimp (changes requested):
So I'm inclined to just close this as not desired. There's so much here that needs to change, I'm not sure that there's anything left

bsdimp:
> @markjdb prefers `intr_event_destroy()` returning an error for `ie == NULL`. Problem is this is answering the **wrong** question for most callers. Since this is called from cleanup/error paths, the caller wants to know `Has everything been properly cleaned up?` Whereas returning `EINVAL` is answering `Has everything been cleaned up and was everything in **perfect** condition?`
> 
> I'm pretty sure most, if not all callers will want the answer to the first since `EBUSY` means memory is likely being leaked. Whereas for INTRNG, the event being `NULL` simply means no event was ever allocated and things are fine.

I agree with mark. All the callers want no errors at all. They all assume that it's OK to tear down this intr_event. EBUSY is an error here (not a try again considtion). It's for when the handlers haven't been unregistered. This leads to a panic in most of the callers as a programming error. I disagree with this analysis. That makes the whole basis for this change in question. And the change itself isn't just this, but has other bits in it as well. I'm inclined to just close this and something we don't want, at least not in the current form.

sys/kern/subr_intr.c
ehem: > We shouldn't be papering over programing errors. It's a programming error when isrc_irq is >= intr_nirq. That normally can't happen. Likewise with setting things to NULL in this way. This is a bad change. NAK.
A hypothetical scenario, but one which I believe _should_ be handled as it could nominally be used.  Imagine a highly hot-plug bus on a system with >1000 processors.  On such a hot-plug bus it could nominally be valid to have one thread which calls `intr_isrc_deregister()` at 1 second intervals and a separate thread doing `intr_teardown_irq()` calls.  In such a case the sequence:
```
intr_isrc_deregister();
intr_teardown_irq();
intr_isrc_deregister();
```
Issue is the first `intr_isrc_deregister()` call would get `EBUSY` from `intr_event_destroy()` since the filter wasn't yet removed.  Presently the second would get `EINVAL` from `isrc_free_irq()` since the `isrc` had already been removed from the table by the first call.  When it comes down to it, either `isrc_free_irq()` or `isrc_event_destroy()` **must** tolerate being called multiple times with the same `isrc`.

> Hard to say for sure, though, since there's not actually any callers of this in the tree. This is only called from `intr_isrc_deregister`, which is global, but has no actual callers. So how did you test this?
By reading the source and concluding this was a nominally valid course of action.  I've got a driver which is expected to call `intr_isrc_deregister()` during normal operation.  As such I'm examining `intr_isrc_deregister()` and discovering it is rather buggy.  The x86 equivalent driver is unable to release interrupts and is thus forced to keep them on a list.  Whereas INTRNG has the function partially implemented, but untested.

I have no expectation to need this functionality, but when trying to fix the problems with `intr_isrc_deregister()` this issue shows up.  Presently I'm unable to submit the driver since 3 key areas it needs working aren't in usable shape.

sys/kern/subr_intr.c
ehem: Notice another function called near this function?  What does that function do?  Having considered the situation, I do believe it is appropriate for this one to be called second.

sys/kern/subr_intr.c
ehem: Notice how the event is created by `isrc_event_create()`?  Given that, it is INTRNG's job to destroy the event, yet this call was missing.  I simply needed to examine `intr_isrc_deregister()` and notice the missing call.  The driver is sitting in my tree due it needing `intr_isrc_deregister()` to actually work.

sys/kern/kern_intr.c
ehem:
```
% find sys -type f -print0 | xargs -0 grep -eintr_event_destroy -l
sys/netpfil/pf/if_pfsync.c
sys/netpfil/pf/pf_ioctl.c
sys/netpfil/pf/pflow.c
sys/kern/kern_intr.c
sys/kern/subr_intr.c
sys/sys/interrupt.h
% 
```
The ones in `sys/netpfil/pf/` are unlikely to trigger this.  The only present way for them to trigger this is for something to scribble zeros over memory or some other sort of memory corruption.  My concern is they're all calling `intr_event_destroy()` during shutdown, then either doing `MPASS(rc == 0);` or returning the error to the caller.  If this isn't a debug kernel then the error could simply be lost (`sys/netpfil/pf/if_pfsync.c:vnet_pfsync_uninit()`).  If this is a debug kernel, is a `panic()` truly a desirable result?  I suspect _these_ callers would instead prefer:
```
if (ie == NULL) {
        printf("ERROR: %s(): called with NULL event\n", __func__);
        return (0);
}
```
As the rest of their shutdowns can still occur.

The situation for INTRNG (`sys/kern/subr_intr.c`) is very different.  INTRNG does lazy event allocation.  After `intr_isrc_register()` returns, `->isrc_event` is still `NULL`.  `->isrc_event` doesn't become non-`NULL` until `isrc_add_handler()` is called.  As such:
```
struct intr_irqsrc *isrc;
int rc;
intr_isrc_register(isrc);
rc = intr_isrc_deregister(isrc);
printf("intr_isrc_deregister() returned: %d\n", rc);
```
Will reliably output `intr_isrc_deregister() returned: 22`.  So there is a disagreement in the implementation about what `intr_event_destroy(NULL)` should do/return.  Since I'm looking at a driver which needs `intr_isrc_deregister()` to work, this is blocking for me.

I suppose _you_ could modify `isrc_event_destroy()` to do `if (isrc->isrc_event != NULL) return(0);`.  My concern with this is you're testing the same value twice and giving exactly opposite results for a problematic value.  I cannot abide this, it is just too ugly (I will hold my nose and ignore it if someone else implements, but I will not implement).

Another viable approach would be to remove lazy event allocation from INTRNG.  I don't think lazy event allocation significantly helps INTRNG.  I also observe major downsides to lazy event allocation (both here and other places you must first test `->isrc_event`, you cannot assume non-`NULL`).  I think it was a mistake to allow PowerPC and INTRNG to use lazy event allocation.
Issue is this would basically invalidate the premise behind `INTR_SOLO` and someone would be rather annoyed with that (`D35607`).

ehem:
> That makes the whole basis for this change in question. And the change itself isn't just this, but has other bits in it as well. I'm inclined to just close this and something we don't want, at least not in the current form.

Wrong.  The one and only basis for this pull is that `intr_deregister_source()` is unreliable and really should be fixed.

sys/kern/kern_intr.c
ehem: Actually, one other approach _I_ would be willing to implement.  It would be acceptable to me to revert 39888ed7a3dca81c5d5c886e699dbcf58953137f and then have `isrc_event_destroy()` check for `isrc->isrc_event == NULL`.  It is the double checking for `NULL` which I have a problem with.

I though would be quite happy if lazy event allocation was deprecated, and assured removing that would be accepted.

ehem:
An alternative patch set to fix the issue: [altfix](https://github.com/ehem/freebsd-src/tree/altfix)  As GitHub functions on branch names, I could readily push that on top of "intrfix" for this pull.

It is nominally feasible to omit the first patch in that series.  Problem is that leaves `INTR_SOLO` in a broken state.  The premise of `INTR_SOLO` is to try to avoid allocating the event, which is incompatible with removing lazy allocation.  The last 3 are also optional for solving the issue, though the fourth looks like a kind of Good Idea.

ehem:
Okay, try this.  Note the later commits are cleaning out some bits which I would tend to, but they are optional to the goal of this pull request.

I suspect `INTR_SOLO` retains its level of functionality.  The worth may be degraded, but as near as I can tell it still does what it did before (I'm unable to test since there are no test cases).

I'm trying to check whether `isrc->isrc_handlers` can be removed.  _Looks_ like it _should_ work, but testing failed.  I'm left suspicious the PIC drivers may have placed some steps which should have been in `pic_enable_intr()` hooks into `pic_setup_intr()` hooks.

ehem:
It actually only takes the two at the front to achieve my primary objective.  With those `intr_isrc_deregister()` would genuinely function and fully release the portions allocated by INTRNG.

The as not yet submitted driver would simply use `intr_isrc_deregister()` and emit a warning if it gets an error code.  It simply won't handle recovery beyond that.  Yet I feel it is my duty to allow other driver writers to go further on their recovery approaches.

ehem:
Okay, figured out what was going on with `->isrc_handlers`, so that can now be nuked.  The goal for this pull is achieved at 3068c822dcb784d5f66f73343d86510fbd359a6b, though the cleanup in 56aaa294392145dd64476da6ab3335204d52f978 is kind of nice too.

I would _like_ to purge `INTR_SOLO`, but that is not a _requirement_ for this to be a success.  The main goal is simply for:
```
intr_isrc_register(isrc);
intr_isrc_deregister(isrc);
```
To fully cleanup the structure with no leaks.  Notably this requires `->isrc_event` to be released by INTRNG (which allocates it).  I would prefer for it to be possible to recover after `EBUSY`, but that is not actually a requirement for my goal.

sys/kern/subr_intr.c
```diff
@@ -1867,8 +1864,6 @@ intr_ipi_setup(u_int ipi, const char *name, intr_ipi_handler_t *hand,
 	if (error != 0)
 		return;
 
-	isrc->isrc_handlers++;
-
```
ehem, line 1868: This turns out to be an issue.  With `intr_setup_irq()` the handler can be detected via `CK_SLIST_EMPTY()`, but here the IPI interrupt handler is never passed to the event core.  This makes sense due to the differences between IPI and SPI interrupts, but complicates the detection.

Would be possible to synthesize a NOP handler function to hand to the event core.  Nominally that _should_ never be called and isn't much more than a placeholder for this purpose.  Likely a worthwhile trade-off as this shrinks every `struct intr_irqsrc` by 4 bytes by creating an extra `struct intr_handler` for IPI interrupts (roughly +64 bytes).

Due to this issue I'm pulling this commit from the series for now.

ehem:
I finally realized `INTR_SOLO` isn't actually incompatible with always allocating events.  It is doing using `INTR_SOLO` and event-style handlers at the same time which is incompatible.

As a result removing lazy event allocation works fine.  This then means there is no need to check for `->isrc_event` being `NULL` before calling `intr_event_destroy()`.  The only need is to actually call `isrc_event_destroy()` so the event is released.

This version has been tested.  The final commit isn't absolutely needed, but I do wish to express the point that with no activity on `INTR_SOLO` in more than 5 years it is overdue for removal.  Its presence confuses developers into thinking it might be the future (it is not) and is thereby wasting a great deal of developer time.

ehem:

ehem:
In fact, the two at the front really do fix the problem sufficiently.  If the event is always allocated, then `intr_event_destroy()` won't ever be called with a `NULL` event.  Then it simply becomes an issue of needing `isrc_event_destroy()` to be fixed and be called.

I still feel `intr_event_destroy()` returning success and doing a `printf()` is the way to go.  Since the function is called from cleanup paths, you don't want to abort those if no event was ever allocated or cleanup was already done.

ehem:
Given how things look now, I plan to rebase in a bit.  Some of what is in here now looks worthy of being split into a separate branch.

sys/kern/kern_intr.c
ehem: In different terms, this situation is highly similar to c1287a3bb1b7b7ccc494d6718db52cb443bd0d38,  Yes, it might be an error, but handing the error off to the caller drastically escalates the damage.  Whereas handling it mitigates things.

ehem:
The issue 5aa7e6e8ea778851433795dba32deb2e4d0370ae is aimed at is `intr_event_destroy()` is a shutdown/cleanup function.  As such only **urgent** errors should be returned.  The event already being `NULL` is legitimately bad, but if the caller is cleaning up due to other error situation there is likely no recovery path.  Whereas if `printf()` is done there is hope the system won't immediately need to `panic()`.

In theory simply including the string @strejda can summon people.  The main goal here is `intr_isrc_deregister()` leaks memory.  INTRNG may allocate for `->isrc_event`, but presently `intr_isrc_deregister()` fails to clean that up.

ehem:
Some of what is here would be rendered obsolete by #1457.

ehem:
Mostly I rebase to keep everything near HEAD.  This time I also grabbed one commit from #1457 where I noticed a lurking issue.  In most cases this will never trigger, but it it a lurking issue of concern.

PR #1284, opened 2024-06-10
Fix a typo in manual page.

gevorgyana:


gevorgyana:
See man `namei(9)`

emaste:
The change proposed in https://reviews.freebsd.org/D45546 should make it clear

PR #1285, opened 2024-06-11
split INTRNG main file into two files

ehem:
Hopefully SSIA.  INTRNG presents multiple interfaces and tries to hide some.  There seems to be a border between the PIC side and the event side.  Split these two apart.

The PIC side bears fair resemblance to what is in `x86/nexus.c`.  I suspect it might be possible to have this take over some of what is done for x86.  Theory is to shrink the architecture specific bits.


ehem (commented):
Note to reviewers, I notice `intr_event_describe_handler()` is one of the few `intr_event` functions not to the cookie for `NULL`.  As such handling this outside.

ehem:
I'll confess this is mostly meant to suggest there really is another interface being hidden here.  Mainly `intr_add_handler()` and `intr_describe()` very definitely point to a border.  Some drivers would benefit from access to this extra interface.

ehem:
Gah!  Keep trying to draw a borderline and some bits are quite unclear.

ehem:
This is an alternative.  I'm pretty sure I've got `intr_describe_irq()` split correctly.  Problem is I'm rather less confident of how `intr_isrc_assign_cpu()` should be split.  Then there are a bunch of pieces which could fall on either side.

ehem:
I'm rather unsure where the border goes here.  To tell the truth, right now I'm primarily interested in the two at the front.  I'm trying to port a driver for an unusual PIC.  Due to extra knowledge about its bus, the implementation skips some steps and therefore doesn't match to INTRNG's existing interface.

PR #1286, opened 2024-06-12
work towards unifying interrupt interfaces

ehem:
This is some work towards unifying the architectural interrupt interfaces.  Simply tries to identify an identical portion and take advantage of what having a common interface provides.

### DIFF
--- a/sys/arm64/arm64/gic_v3.c
+++ b/sys/arm64/arm64/gic_v3.c
@@ -54,7 +54,7 @@
 
 #include <machine/bus.h>
 #include <machine/cpu.h>
-#include <machine/intr.h>
+#include <machine/interrupt.h>
 
 #ifdef FDT
 #include <dev/fdt/fdt_intr.h>
--- a/sys/arm64/arm64/gic_v3.c
+++ b/sys/arm64/arm64/gic_v3.c
@@ -36,6 +36,7 @@
 #include <sys/systm.h>
 #include <sys/bitstring.h>
 #include <sys/bus.h>
+#include <sys/intrtab.h>
 #include <sys/kernel.h>
 #include <sys/ktr.h>
 #include <sys/malloc.h>
--- a/sys/arm64/arm64/gic_v3.c
+++ b/sys/arm64/arm64/gic_v3.c
@@ -355,12 +355,18 @@ gic_v3_attach(device_t dev)
 	if (sc->gic_nirqs > GIC_I_NUM_MAX)
 		sc->gic_nirqs = GIC_I_NUM_MAX;
 
-	sc->gic_irqs = malloc(sizeof(*sc->gic_irqs) * sc->gic_nirqs,
-	    M_GIC_V3, M_WAITOK | M_ZERO);
+	sc->gic_intrs = intrtab_alloc_intr(dev, sc->gic_nirqs);
+	if (sc->gic_intrs == NULL) {
+		device_printf(dev, "could not allocate interrupts\n");
+		return (ENOSPC);
+	}
+
 	name = device_get_nameunit(dev);
 	for (irq = 0; irq < sc->gic_nirqs; irq++) {
-		struct gic_v3_irqsrc *gi = &GIC_INTR(sc, irq);
+		struct gic_v3_irqsrc *gi;
 		struct intr_irqsrc *isrc;
+
+		gi = malloc(sizeof(*gi), M_GIC_V3, M_WAITOK | M_ZERO);
 
 		gi->gi_irq = irq;
 		gi->gi_pol = INTR_POLARITY_CONFORM;
@@ -369,19 +375,28 @@ gic_v3_attach(device_t dev)
 		isrc = &gi->gi_isrc;
 		if (irq <= GIC_LAST_SGI) {
 			err = intr_isrc_register(isrc, sc->dev,
-			    INTR_ISRCF_IPI, "%s,i%u", name, irq - GIC_FIRST_SGI);
+			    INTR_ISRCF_NOIRQ | INTR_ISRCF_IPI, "%s,i%u", name,
+			    irq - GIC_FIRST_SGI);
 		} else if (irq <= GIC_LAST_PPI) {
 			err = intr_isrc_register(isrc, sc->dev,
-			    INTR_ISRCF_PPI, "%s,p%u", name, irq - GIC_FIRST_PPI);
+			    INTR_ISRCF_NOIRQ | INTR_ISRCF_PPI, "%s,p%u", name,
+			    irq - GIC_FIRST_PPI);
 		} else {
-			err = intr_isrc_register(isrc, sc->dev, 0,
-			    "%s,s%u", name, irq - GIC_FIRST_SPI);
+			err = intr_isrc_register(isrc, sc->dev,
+			    INTR_ISRCF_NOIRQ, "%s,s%u", name,
+			    irq - GIC_FIRST_SPI);
 		}
 		if (err != 0) {
 			/* XXX call intr_isrc_deregister() */
-			free(sc->gic_irqs, M_DEVBUF);
+			free(gi, M_GIC_V3);
 			return (err);
 		}
+
+		if (intrtab_set(sc->gic_intrs,
+		    rman_get_start(sc->gic_intrs) + irq, &gi->gi_isrc, NULL) !=
+		    0)
+			panic("%s(): Failed when setting up IRQ %u (%lu)",
+			    __func__, irq, rman_get_start(sc->gic_intrs) + irq);
 	}
 
 	mtx_init(&sc->gic_mbi_mtx, "GICv3 mbi lock", NULL, MTX_DEF);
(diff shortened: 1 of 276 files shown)


ehem:
I now have a better feel for the complaints about Hyper-V for arm64.

sys/dev/hyperv/pcib/vmbus_pcib.c
```diff
@@ -50,7 +50,7 @@
 #include <vm/pmap.h>
 
 #if defined(__aarch64__)
-#include <arm64/include/intr.h>
+#include <machine/intr_machdep.h>
```
ehem, line 53: Wow.  Just wow.  The qualities of this are just...

Alas, I'm quite unsure why the original author did things this way.  On further consideration, perhaps this should simply be `#include <arm64/include/intr_machdep.h>` to fulfill the minimum.  Let someone involved with this do the proper job.

ehem: This comes from 6e5b082cab8d9dd0eb0bce65cf8cf268f11c7a25.  Looks like they wanted ac29ddabea3fedaace5c7dab54287ee2edd203a7 (or the inverse of `intr_machdep.h` => `intr.h`) as well.  Alas the qualities of that implementation are poor.

ehem:
Another note to reviewers.  I think `intrtab_lookup()` is a better name for the function due to that being the minimum guarantee behind it.  Many MIPS devices would have used `typedef struct intr_event interrupt_t;`, so the thing being returned is _something_ interrupt and not necessarily something which could be called "source".

Hmm, I was planning to remove since I'm unsure.  I'm not quite not 100% sure whether the `intrtab_lookup()` functions should be marked `__pure`.  There are no side-effects and their returns tend to be stable, yet behind the scenes they are relying on global tables which can be modified by calling _other_ functions.

ehem:
_This_ is roughly where I figure to end #1286.  Everything up to 06660ff10fc5c17bbb873d5d046a633f72867dc7 is pretty much acceptable.  Everything after 06660ff10fc5c17bbb873d5d046a633f72867dc7 is near-certain not to work.  Problem is the documentation for `rman` doesn't really give me an appropriate feel for some of its workings.  Ideally a userspace `librman.so` which could be used to explore its capabilities and interface would be really handy.

Getting a ruling on 83653a7e2b0a6fe22645a6bc606b4d28464b2d5a and the issue of `machine/intr_machdep.h` versus `machine/intr.h` sooner would be nice though.  That really doesn't need to wait and is an albatross for any effort to unify the interrupt interfaces.

Even though this is a draft, comments are invited.

ehem:
Figuring I need to update the second patch.  I'm assuming either `machine/intr.h` or `machine/intr_machdep.h` will be chosen.  I'm using `machine/machintr.h` simply to express neutrality, issue is a decision is needed if there is any hope FreeBSD will ever have a single interrupt system.  For drivers which are independent of architecture this is a major pain.

I think the second and third commits would be highly valuable *now*.  Looks rather feasible to start sharing more simply by having a common type definition.

The remaining issue is I don't have a good feel for the resource manager.  As a result the implementation is very likely flawed.  The issue is which approach should be used to pass the table size to the core.

sys/arm64/arm64/gic_v3.c
ehem, line 1: Indeed, the motivation behind this commit and the next commit is it makes the conversion to the proposed common interrupt table simpler.  Yet this does pretty distinctly shrink the files.  Nominally these two could go in independent of the rest of the pull.

ehem:
Hrmm, find all the existing occurrences, but them miss the two internal ones.  Still looking for feedback.

ehem:
A distinct issue is the documentation for the resource manager seems unclear to me.  There are the `->rm_start` and `->rm_end` fields, but what effect do these actually have?  I'm getting the impression they're checked against `rman_manage_region()`, but not enforced elsewhere.

I'm left suspecting for x86 the setup should be roughly `->rm_start = 0; ->rm_end = ~0; rman_manage_region(rman, 0, ~0);`, then let the various allocators allocate blocks.  After all blocks are allocated do `rman_last_free_region(rman, &start, &end);`, then set `rman.rm_end = start;`.  This would have `intrtab_init()` allocate the appropriate sized array.

Anyway the first key issue is the commit at the start.  My impression of PowerPC interrupts is that is correct since `->vector` is the dense table whose characteristics match other architectures, while `->irq` is a sparsely populated rather more similar to PCI bus/device/function.  If so then the shared table implementation is always a dense table and this works.

The second issue is the second commit and the standard name to use.  The long disagreement of `machine/intr.h`/`machine/intr_machdep.h` needs to be settled once and for all.  Once that is done the third commit allows a surprising amount of convergence all by itself.

ehem:
There seems to be disagreement on the interpretation of "draft" here.  I was choosing "draft" to indicate that while this is not 100% ready, but it is _mostly_ ready and could use some review/feedback.  Whereas the potential reviewers seem to think "draft" indicates no need for review yet?

Unless there is something from reviewers I'm going to select the "Ready for review" in 3-7 days...   Commit 2 needs a choice made, while commit 3 is 100% ready **now** (and doesn't actually have any ordering issues).

ehem:
The second and third are surprisingly effective at pushing merging further.  There are a large number of spots which don't really care _which_ interrupt structure they use, they simply need to reference the one in use.

The big question mark is 8f3c6ddf6a85c13badff1df1530f901d8230cabf and adjusting x86 to use the interface.  Instead of requiring every PIC interface to be friends with each other, they could simply allocate their range and be protected from everyone else.

ehem (commented):
I _prefer_ the interface of `intr_{get|set}affinity()` being passed the `struct intr_event *`.  This is more consistent with the architecture interrupt interfaces.  I do though admit the _current_ situation all callers instead have the vector.

As a result this second change is optional and not actually required for this pull.  This one could be skipped when bringing into the tree and still get most of the benefits.

ehem (commented):
Having played with things a bit, I now think it best for `intr_event_create()` _not_ to be passed the interrupt number.  Issue is both INTRNG and PowerPC allocate the interrupt number rather later than x86.  It is possible to remove lazy allocation of events from the other architectures, but they would then need to set `->ie_irq` _after_ allocation.

ehem (commented):
A core question of this series is, "Is 9b33b154b531606eccfc091faf50467eca0573f1 the future?"  16 years ago, 9b33b154b531606eccfc091faf50467eca0573f1 copied interrupt numbers to `struct intr_event` and then pretty well nothing has happened.  Either 9b33b154b531606eccfc091faf50467eca0573f1 should be reverted or else the places which are copied to `->ie_irq` should be removed.

I've been given the impression the way forward is to push things towards using `->ie_irq` and removing the places which are copied to `->ie_irq`.  Here and the next commit we see 2 of the x86 PICs had variables which stored the value simply for copying into `->ie_irq`.  Using `->ie_irq` allows the x86 PIC interface to be simplified.

ehem:
Hopefully this is a bit more obvious.  "a_bikeshed_string_for_sed_to_target" is simply a handy string for `sed` to target, not an actual suggestion.  Theory is this collects all the information together for someone with a commit bit to do:
```
git format-patch 7f0f39b57d08b14420823e7276c9222f37886468 -1
sed -i -es,a_bikeshed_string_for_sed_to_target,interrupt, 0001-intrcompat-rename-machine-intr-.h-to-machine-a_bikes.patch
git am 0001-intrcompat-rename-machine-intr-.h-to-machine-a_bikes.patch
```

Issue is "intr_machdep.h" versus "intr.h" is too trivial for consensus to work.  This appears to have been argued about for more than 2 decades already, due to being so trivial.

I suppose one might setup a bunch of eggshell shaped spheres at BSDCon (or some other appropriate event).  Inside have 3 buttons, "intr.h", "intr_machdep.h", "other".  At noon of some day a vacuum pump attached to all spheres would start to operate.  This pump would continue operation until only "intr.h" or "intr_machdep.h" buttons were depressed, if the "other" button was depressed the pump would continue operation until it was only depressed in a single sphere.  By having everyone place their lives on the line an end could be achieved.  Yet I fear this would end up with a mass casualty event.

Could we just bring this most stupidly trivial of issues to a close?

ehem:
Appears the issue started at 7fed2e3a32d6b44adca73179b98730d1938c7767.  Even for consensus on a trivial issue that is impressive.

ehem:
I really was kind of hoping for activity on this.  The first two can readily be done before trying to review the later commits (GitHub might automatically close the pull request, but it can be readily reopened then rebased).  This would also give hope that waiting 2 months for activity wasn't in vain...

The situation with `machine/intr.h` versus `machine/intr_machdep.h` appears a bit more complex on further review.  Alpha introduced `machine/intr.h` at 7fed2e3a32d6b44adca73179b98730d1938c7767 (1998-01).  x86 introduced `machine/intr_machdep.h` at ecee5704ed500a18a49850acb275aa4ff463bdd3 (2003-11).
PowerPC introduced `machine/intr.h` at f9bac91b18fc344fe5967f4367a1994588a26b10 (2001-06), then created `machine/intr_machdep.h` at f6a7723dff5b8049d9674e05f6bee7b5c43afb35 (2002-07).  I believe the theory was `intr.h` was supposed to be all PowerPC variants, while `intr_machdep.h` was machine-specific tweaks.  Yet at 8a35d194f2a3d472a22cecc0de0491f2651c7a96 (2010-07) the machine-specific header had absorbed everything.

Standardizing on **one** header name has a surprisingly large benefit.  Any efforts to converge the subsystems will eventually settle on one name, but sooner turns out to be better.

PR #1287, opened 2024-06-12
ctld: parse config file independently of getting kernel info

asomers:
Separate the parsing of the config file from the reading of kernel port information.  This has three benefits:

* Separation of concerns makes future changes easier.
* Allows the config file to be read earlier, which is necessary for fixing PR 271460.
* Reduces total line count, by eliminating duplication between parse.y (for traditional config file) and uclparse.c (for UCL config file).

MFC after:	2 weeks
Sponsored by:	Axcient

### DIFF
--- a/usr.sbin/ctld/ctld.c
+++ b/usr.sbin/ctld/ctld.c
@@ -98,7 +98,6 @@ conf_new(void)
 	TAILQ_INIT(&conf->conf_auth_groups);
 	TAILQ_INIT(&conf->conf_ports);
 	TAILQ_INIT(&conf->conf_portal_groups);
-	TAILQ_INIT(&conf->conf_pports);
 	TAILQ_INIT(&conf->conf_isns);
 
 	conf->conf_isns_period = 900;
@@ -117,7 +116,6 @@ conf_delete(struct conf *conf)
 	struct target *targ, *tmp;
 	struct auth_group *ag, *cagtmp;
 	struct portal_group *pg, *cpgtmp;
-	struct pport *pp, *pptmp;
 	struct isns *is, *istmp;
 
 	assert(conf->conf_pidfh == NULL);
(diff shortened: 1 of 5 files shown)


amotin (commented):
My only concern is that postponing ports processing can make error reporting less informative.  Though I don't remember whether the old code actually reported the line numbers.

asomers:
No, errors like this wouldn't get line numbers.  A syntax error in the config file would and still will report its line number.  But an error like "unknown port" never did.  That sort of error message is unchanged by this PR:
```
ctld: unknown port "isp0" for target "iqn.2018-10.myhost:testvol3"
```

asomers:
ping @amotin do you have any more concerns, given that the error messages for unknown ports won't change?

usr.sbin/ctld/ctld.c
```diff
@@ -2852,6 +2900,10 @@ main(int argc, char **argv)
 					tmpconf->conf_debug = debug;
 				oldconf = newconf;
 				newconf = tmpconf;
+				if (new_pports_from_conf(newconf, &kports))
+					log_errx(1, "Error associating "
+					    "physical ports");
```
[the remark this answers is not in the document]
asomers: Good catch.  Fixed.

usr.sbin/ctld/ctld.c
```diff
@@ -2841,17 +2890,20 @@ main(int argc, char **argv)
...
-			tmpconf = conf_new_from_file(config_path, newconf,
-			    use_ucl);
+			tmpconf = conf_new_from_file(config_path, use_ucl);
 
 			if (tmpconf == NULL) {
 				log_warnx("configuration error, "
 				    "continuing with old configuration");
+			} else if (new_pports_from_conf(tmpconf, &kports)) {
+				log_warnx("Error associating physical ports, "
+				    "continuing with old configuration");
```
amotin, line 2900: What happens to tmpconf in this case?

asomers: It will leak.  Fixed.

asomers:
ping @amotin is there anything else you'd like to see here?

usr.sbin/ctld/ctld.h
```diff
@@ -305,16 +312,19 @@ void			isns_register(struct isns *isns, struct isns *oldisns);
...
+struct pport		*pport_copy(struct pport *pp, struct kports *kports);
 void			pport_delete(struct pport *pport);
 
 struct port		*port_new(struct conf *conf, struct target *target,
 			    struct portal_group *pg);
-struct port		*port_new_ioctl(struct conf *conf, struct target *target,
+struct port		*port_new_ioctl(struct conf *conf,
+			    struct kports *kports, struct target *target,
 			    int pp, int vp);
+			    //int pp, int vp);
```
amotin, line 327: What is this commented line?

amotin (approved):
If you say it works, so be it.

asomers:
Thanks mav

PR #1292, opened 2024-06-15
loader: Correct terminal size

VexedUXR:
The UEFI spec says:
> It is required that all output devices support at least 80x25 text mode. This mode is defined to be mode 0.

So change the EFI `TEXT_ROWS` macro to reflect this. This also makes it more inline with the BIOS version:
https://github.com/freebsd/freebsd-src/blob/d19851f002862a5510bf31fae4083fab979258be/stand/i386/libi386/vbe.h#L136

Also clarify the man page. The only time we have a fixed resolution terminal is in text mode with the BIOS loader, and that uses 80x25. With the other configurations, the terminal size isn't really guaranteed to be anything.

### DIFF
--- a/stand/defaults/loader.conf.5
+++ b/stand/defaults/loader.conf.5
@@ -21,7 +21,7 @@
 .\" LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
-.Dd February 2, 2024
+.Dd July 8, 2024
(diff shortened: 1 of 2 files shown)


bsdimp:
A real vt100 is only 24 lines. This is commonly emulated like that.... but i know of almost no one that's so restrictive on the consoles they run. So we can live with that historical mismatch. While the successor terminals to the vt100 also had this limitation, the last time i heard of one connected to a FreeBSD machine was before smpng. And that was a gag....

bsdimp:
Looks good otherwise.  Any chance you'd do sob lines for the commit?

VexedUXR:
Sure. Was I supposed to do that for my other commits :/

bsdimp:
Thanks for your contribution. My apologies that it took so long to land.

PR #1295, opened 2024-06-16
Makefile.inc1: Also cleanup (un)compressed manpages

ricardobranco777:
Cleanup (un)compressed manpages in the `check-old-files` & `delete-old-files` targets.

Also cleanup manpages for arches not included in MAN_ARCH, if specified.

PR:https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=279792

### DIFF
--- a/Makefile.inc1
+++ b/Makefile.inc1
@@ -3516,6 +3516,13 @@ delete-old-files: .PHONY
 		fi; \
 	done
 .endif
+.if !empty(MAN_ARCH) && ${MAN_ARCH} != "all"
+	@exec 3<&0; \
+	find -E /usr/share/man/man4/ -mindepth 1 -type d ! -regex '.*/(${MAN_ARCH:C/[[:space:]]+/|/gW})' | \
+	while read dir; do \
+		find $$dir ! -type d -exec rm ${RM_I} {} + <&3; \
+	done
+.endif
 	@echo ">>> Old files removed"
 
 check-old-files: .PHONY
@@ -3556,6 +3563,12 @@ check-old-files: .PHONY
 		if [ -e "$${manpage}.gz" ]; then \
 			echo $${manpage}; \
 		fi; \
+	done | sort
+.endif
+.if !empty(MAN_ARCH) && ${MAN_ARCH} != "all"
+	@find -E /usr/share/man/man4/ -mindepth 1 -type d ! -regex '.*/(${MAN_ARCH:C/[[:space:]]+/|/gW})' | \
+	while read dir; do \
+		find $$dir ! -type d; \
 	done | sort
 .endif
 
--- a/Makefile.inc1
+++ b/Makefile.inc1
@@ -3497,6 +3497,25 @@ delete-old-files: .PHONY
 			rm ${RM_I} $${catpage} <&3; \
 	        fi; \
 	done
+# Remove compressed copies of uncompressed manpages
+.if ${MK_MANCOMPRESS} != "yes"
+	@exec 3<&0; \
+	find ${DESTDIR}/usr/share/man ${DESTDIR}/usr/share/openssl/man ! -type d ! -name \*.gz 2>/dev/null | \
+	while read manpage; do \
+		if [ -e "$${manpage}.gz" ]; then \
+			rm ${RM_I} $${manpage}.gz <&3; \
+		fi; \
+	done
+# Remove uncompressed copies of compressed manpages
+.else
+	@exec 3<&0; \
+	find ${DESTDIR}/usr/share/man ${DESTDIR}/usr/share/openssl/man ! -type d ! -name \*.gz 2>/dev/null | \
+	while read manpage; do \
+		if [ -e "$${manpage}.gz" ]; then \
+			rm ${RM_I} $${manpage} <&3; \
+		fi; \
+	done
+.endif
 	@echo ">>> Old files removed"
 
 check-old-files: .PHONY
@@ -3522,6 +3541,23 @@ check-old-files: .PHONY
 			echo $${catpage}; \
 	        fi; \
 	done | sort
+# Check for compressed copies of uncompressed manpages
+.if ${MK_MANCOMPRESS} != "yes"
+	@find ${DESTDIR}/usr/share/man ${DESTDIR}/usr/share/openssl/man ! -type d ! -name \*.gz 2>/dev/null | \
+	while read manpage; do \
+		if [ -e "$${manpage}.gz" ]; then \
+			echo $${manpage}.gz; \
+		fi; \
+	done | sort
+# Check for uncompressed copies of compressed manpages
+.else
+	@find ${DESTDIR}/usr/share/man ${DESTDIR}/usr/share/openssl/man ! -type d ! -name \*.gz 2>/dev/null | \
+	while read manpage; do \
+		if [ -e "$${manpage}.gz" ]; then \
+			echo $${manpage}; \
+		fi; \
+	done | sort
+.endif
 
 list-old-libs: .PHONY
 	@cd ${.CURDIR}; \
--- a/share/mk/bsd.man.mk
+++ b/share/mk/bsd.man.mk
@@ -237,11 +237,9 @@ maninstall: ${MAN}
 # On MacOS, assume case folding FS, and don't install links from foo.x to FOO.x.
 .if ${.MAKE.OS} != "Darwin" || ${l:tu} != ${t:tu}
 .if ${MK_MANSPLITPKG} == "no"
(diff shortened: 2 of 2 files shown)


concussious:
(I don't really understand where is the appropriate place to have this conversation so I'm posting both places -- guidance welcome)

Maybe that's a feature?

With compressed manual pages, you can install a locally modified manual page alongside the compressed ones. Man(1) chooses the uncompressed version first, and the compressed versions are still tracked and updated with the system, while leaving the locally modified manuals alone. There are other ways to do this, but this way is zero conf.

Nobody told me to do it this way, but I've been doing this for years.

If others agree that it's better to clean them up instead, I'm okay with that.

ricardobranco777:
> With compressed manual pages, you can install a locally modified manual page alongside the compressed ones. Man(1) chooses the uncompressed version first, and the compressed versions are still tracked and updated with the system, while leaving the locally modified manuals alone. There are other ways to do this, but this way is zero conf.

Why would you have a locally modified manual page?  

I enabled ZFS compression and having compressed manual pages is a waste.

concussious:
> Why would you have a locally modified manual page?

hier(7): "A well-maintained installation will include  a  customized version of this document."

Best reason: man page developmemt and testing.

Corner reason: upstream and user are opinionated about doc in divergent ways.

Secret reason: keeps their playbook in the form of examples in custom manpages 😎 `man foo` is more elegant than `m ~/notes/foo`.

> I enabled ZFS compression and having compressed manual pages is a waste.

I have had this setup sometimes but that's pretty custom. Iirc without_mancompress is a src.conf setup.

I could manpath +/man/working just as easily as you could rm */*.gz.

How much of a waste is it? Zfs compression knows to bypass compressed files.

Again, I don't object to changing it, but I do think it's important that this is mentioned as part of the discussion.

ricardobranco777:
> > Why would you have a locally modified manual page?
> 
> hier(7): "A well-maintained installation will include a customized version of this document."
> 
> Best reason: man page developmemt and testing.

This is the worst reason. Why use a staging manpage in production?  You could manpath for that.

Makefile.inc1
```diff
@@ -3522,6 +3537,21 @@ check-old-files: .PHONY
 			echo $${catpage}; \
 	        fi; \
 	done | sort
+.if ${MK_MANCOMPRESS} != "yes"
```
bsdimp, line 3540: I wonder if this could somehow just be on the inner if rather than having two identical calls copies of the loop

ricardobranco777: > I wonder if this could somehow just be on the inner if rather than having two identical calls copies of the loop

The problem are the backslashes.

bsdimp: Let me play with something to overcome that

concussious:
> Why use a staging manpage in production?

I test in production because I'm embarrassed when I proposed flawed changes in default behavior and scared of mutually unknown unknowns.

ricardobranco777:
Added support to also cleanup manpages for arches not included in MAN_ARCH, if specified.

bsdimp:
I wonder... what do you think about making this a script in tools/build somewhere that we run. Then the repetition won't be a problem...

ricardobranco777:
> I wonder... what do you think about making this a script in tools/build somewhere that we run. Then the repetition won't be a problem...

I did just that in a new commit.  Tell me if you like this approach to improve it further and then squash and rebase.

Makefile.inc1
```diff
@@ -3497,6 +3497,23 @@ delete-old-files: .PHONY
...
+	@exec 3<&0; \
+	find ${DESTDIR}/usr/share/man ${DESTDIR}/usr/share/openssl/man ! -type d ! -name \*.gz 2>/dev/null | \
+	while read manpage; do \
+		if [ -e "$${manpage}.gz" ]; then \
+			rm ${RM_I} $${manpage}.gz <&3; \
+		fi; \
+	done
+.else
+	@exec 3<&0; \
```
bsdjhb: I think comments here would be good.  For this clause I think it would be something like:
```
# Remove uncompressed copies of compressed manpages.
```

Makefile.inc1
```diff
@@ -3497,6 +3497,23 @@ delete-old-files: .PHONY
 			rm ${RM_I} $${catpage} <&3; \
 	        fi; \
 	done
+.if ${MK_MANCOMPRESS} != "yes"
```
bsdjhb: I think the comment here would be:
```
# Remove compressed copies of uncompressed manpages.
```

Makefile.inc1
```diff
@@ -3522,6 +3539,21 @@ check-old-files: .PHONY
 			echo $${catpage}; \
 	        fi; \
 	done | sort
+.if ${MK_MANCOMPRESS} != "yes"
+	@find ${DESTDIR}/usr/share/man ${DESTDIR}/usr/share/openssl/man ! -type d ! -name \*.gz 2>/dev/null | \
```
bsdjhb: Similar "Check for ..." comments would be helpful here I think

bsdjhb (commented):
I'm less certain of a need for cleanup script vs doing it inline.

bsdjhb (commented):
Or at least, if we had a cleanup script, the cleanup script should do all the things including removing catpages, dealing with mismatched MAN_ARCH pages, etc.  To that end, I would probably just leave the last commit off and focus on the first ones.  I think the duplication is ok TBH.

Makefile.inc1
ricardobranco777: Fixed.

ricardobranco777:
> Or at least, if we had a cleanup script, the cleanup script should do all the things including removing catpages, dealing with mismatched MAN_ARCH pages, etc. To that end, I would probably just leave the last commit off and focus on the first ones. I think the duplication is ok TBH.

I applied your suggestions. Hopefully fixed now..

concussious:
IIUC @jrtc27's comment in the bug tracker, this is actually a bug fix. I would like to humbly request changes to commit msg:
```
MFC after: 3 days
Fixes: 460f173dad6e
Reported by: jrtc
Reviewed by: imp, jhb
Pull Request: https://github.com/freebsd/freebsd-src/pull/1295
```

ricardobranco777:
> IIUC @jrtc27's comment in the bug tracker, this is actually a bug fix. I would like to humbly request changes to commit msg:

We all care about commit messages.  And I'll change it after:
1. The code review is finished.
2. Maybe I'm asked to squash 2 commits into one.
3. Last but not least, know why the old commit fails.

I've been doing PR's before and know the drill.;)

concussious:
> I've been doing PR's before and know the drill.;)

Sorry, I did not mean to imply otherwise. I've been enjoying following your work, which is very cool and well above my skill level!

ricardobranco777:
> > I've been doing PR's before and know the drill.;)
> 
> Sorry, I did not mean to imply otherwise. I've been enjoying following your work, which is very cool and well above my skill level!

Actually, I've been never asked to add the ` by:` fields to the commit messages.  Those have been added by the reviewers so far.

ricardobranco777:
We must have a conversation about the wastefulness of compressed manpages.  They don't save that much space and the way we install them fills up space in ZFS snapshots by removing manpages and then installing them:

```
rm -f /usr/share/man/man9/callout_active.9 /usr/share/man/man9/callout_active.9.gz;  install -C -l h -o root -g wheel -m 444  /usr/share/man/man9/callout.9 /usr/share/man/man9/callout_active.9
```

bsdimp:
*DO NOT ADD PULL REQUEST: LINES TO YOUR COMMIT MESSAGES*

OK. Please don't add these, as it slows me down when I'm landing the pull request. My landing software automatically adds them and I have to go in and remove the duplicates.

concussious:
I'm sorry for suggesting the incorrect information.

Makefile.inc1
bsdimp: I've not found time to do this, so please don't let it block this. We/I can refine after the comment when I find time.

bsdimp:
Looking at it again, this looks good enough for now. Is there some reason for @bapt to do the review?

bsdjhb (approved):

ricardobranco777:
I've been using this patch on STABLE-14 for weeks without issues.

ricardobranco777:
It's been months and this hasn't been merged. What happened?

bsdimp:
For me, it's just no time.

ricardobranco777:
This PR has been working for me on 14 & 15.  Can we merge at least for 15?

bsdimp:
I plan on merging this on Friday this week.

ricardobranco777:
@bsdimp sorry to ping you lol. I know you guys must be tired after 14.3-RELEASE.  I missed this one in the latest round of pings.

bsdimp:
Taking another look...  It applied cleanly for me...

PR #1296, opened 2024-06-16
seems stdbool.h was missed to be included in tzsetup.c, causing build failures

smurfd:
noticed this, should be a quick fix :)


emaste:
Reverted for now in c6030b380469f928c8cae87ed53bcb234cb3486f as I am unsure if AArch64 Windows systems are keeping UTC or local time.

PR #1301, opened 2024-06-21
Interrupt counters on `struct intr_event`

ehem:
Every architecture has counters just outside `struct intr_event`.  As such it seems appropriate to try to move them onto `struct intr_event` to reduce differences between architectures.

There is also an issue of the interface between architecture and core.  The core has been implementing `hw.intrcnt` and `hw.intrnames`, yet forcing architectures to handle the tables behind these.  This results in rather troublesome mixing.

### DIFF
--- a/sys/kern/kern_clock.c
+++ b/sys/kern/kern_clock.c
@@ -88,6 +88,8 @@ static struct mtx time_lock;
 SDT_PROVIDER_DECLARE(sched);
 SDT_PROBE_DEFINE2(sched, , , tick, "struct thread *", "struct proc *");
 
+struct	intr_event *clk_intr_event;
+
 static int
 sysctl_kern_cp_time(SYSCTL_HANDLER_ARGS)
 {
@@ -548,6 +550,23 @@ hardclock_sync(int cpu)
 	t = DPCPU_ID_PTR(cpu, pcputicks);
 	*t = ticksl;
 }
+
+/*
+ * Start standard software interrupt threads
+ */
+static void
+start_softintr(void *dummy)
+{
+
+	KASSERT(clk_intr_event == NULL, ("clk_intr_event non-NULL at %s()",
+	    __func__));
+
+	if (swi_add(&clk_intr_event, "clk", NULL, NULL, SWI_CLOCK,
+	    INTR_MPSAFE, NULL))
+		panic("died while creating clk swi ithread");
+}
+SYSINIT(start_softintr, SI_SUB_SOFTINTR, SI_ORDER_FIRST, start_softintr,
+    NULL);
 
 /*
  * Regular integer scaling formula without losing precision:
--- a/sys/kern/kern_clock.c
+++ b/sys/kern/kern_clock.c
@@ -40,6 +40,7 @@
 #include "opt_hwpmc_hooks.h"
 #include "opt_ntp.h"
 #include "opt_watchdog.h"
+#include "opt_acpi.h"
 
 #include <sys/param.h>
 #include <sys/systm.h>
@@ -88,7 +89,9 @@ static struct mtx time_lock;
 SDT_PROVIDER_DECLARE(sched);
 SDT_PROBE_DEFINE2(sched, , , tick, "struct thread *", "struct proc *");
 
+#ifdef DEV_ACPI
 struct	intr_event *clk_intr_event;
+#endif
 
 static int
 sysctl_kern_cp_time(SYSCTL_HANDLER_ARGS)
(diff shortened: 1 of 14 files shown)


ehem:
I do not guarantee I've sufficiently resolved issues brought up via Phabricator, even though I have tried.  This seems the way to go.

sys/kern/kern_intr.c
ehem, line 1: I'm quite unsure of this.  This kind of stands out, but I'm unsure of exactly the reasoning for this being "notyet".

sys/kern/kern_intr.c
```diff
@@ -1519,11 +1519,10 @@ intr_event_handle(struct intr_event *ie, struct trapframe *frame)
 	}
 	critical_exit();
 	td->td_intr_nesting_level--;
-#ifdef notyet
+
 	/* The interrupt is not aknowledged by any filter and has no ithread. */
 	if (!thread && !filter)
 		return (++ie->ie_stray);
-#endif
```
ehem: I'm quite unsure of this.  This sticks out, but I'm not absolutely sure under what exact condition this is meant to be reenabled.

sys/kern/kern_intr.c
```diff
@@ -1521,8 +1522,10 @@ intr_event_handle(struct intr_event *ie, struct trapframe *frame)
 	td->td_intr_nesting_level--;
 
 	/* The interrupt is not aknowledged by any filter and has no ithread. */
-	if (!thread && !filter)
-		return (++ie->ie_stray);
+	if (!thread && !filter) {
+		++ie->ie_stray;
+		return (stray ? ie->ie_stray : 0);
+	}
```
ehem, line 1545: Kind of a wishlist item.  I'm looking at a device which is known to generate a stray interrupt during start.  I would rather like to have it counted, yet returning `FILTER_STRAY` causes all interrupt controllers to mask the interrupt.  Given the device, this would be Bad, as a result the present driver simply always returns `FILTER_HANDLED`.

sys/kern/kern_intr.c
```diff
@@ -629,6 +629,19 @@ intr_event_add_handler(struct intr_event *ie, const char *name,
...
+		    "uniprocessor interrupt\n", name);
+#endif
+		return (ENODEV);
+	}
+#if defined(INVARIANTS) || defined(DEBUG)
+	if (!(flags & INTR_MULTIPROC) && (ie->ie_flags & IE_MULTIPROC))
+		printf("%s: Requested uniprocessor interupt, but got "
+		    "multi-processor interrupt\n", name);
+#endif
+
```
ehem, line 644: Seems something which really should have some way for a driver to indicate, "I'd damn well better get a PPI interrupt, or I'll panic()!"

ehem:

sys/kern/kern_intr.c
ehem: Ah, commit 8381e9f49ec733437754a822ef2e8344115289ac resolves this observation.

sys/kern/kern_clock.c
```diff
@@ -531,7 +534,10 @@ hardclock(int cnt, int usermode)
 			if (i > 0 && i <= newticks)
 				watchdog_fire();
 		}
-		intr_event_handle(clk_intr_event, NULL);
+#ifdef DEV_ACPI
+		if (!CK_SLIST_EMPTY(&clk_intr_event->ie_handlers))
+			intr_event_handle(clk_intr_event, NULL);
```
ehem, line 539: @jrtc27 was objecting to this on #1358, due to looking inside `struct intr_event`.  Issue would be, what approach should instead be used?  Perhaps clock interrupts aren't considered stray if there are no handlers?  Perhaps a flag to indicate no handler shouldn't be considered stray?

ehem:
I don't recall who forced per-processor for PPI counters due to indicating they would veto atomic increment.  This does provide the valuable feature of showing which processors the PPI is occurring on.  Otherwise you're left unsure whether your PPI interrupt being triggered on a single processor, or several.  Also tells you how evenly it is spread.  On balance I think this is distinctly superior to INTRNG's choice to use atomic increment.

ehem:
Note, ARM64 has been tested and was helpful with confirming a fully operational PPI interrupt.  x86 and PowerPC are untested.  I guess the point here is, why is this information being copied back and forth between the events and the architecture tables, and simply being duplicated in both places?  All of this really should be in **one** place.

ehem:
Upon checking I remember why I originally tried this.  Still potentially viable as an alternative solution to a pesky issue.

Note, this has simply been rebased and not properly checked.  There is a good chance some of the commits have become broken in the interim, so exercise care with this.

ehem:
Been some time since I had looked at this.  Some things in the kernel look distinctly concerning.  Seems the best ordering would be #1457, this (will need adaptation) and then adapting #1358.

ehem:
Right now I'm focusing on #1457, but I hope that completes at which point this will be next.  The very first commit should likely be grabbed as that is a bugfix which might effect other places.  Failing that I would kind of like it in sooner to avoid other places starting to rely on buggy behavior.

PR #1302, opened 2024-06-23
timerfd.2: document CLOCK_UPTIME flag

c7skasku:


lib/libsys/timerfd.2
```diff
@@ -97,6 +97,9 @@ argument specifies the clock used as a timing base and may be:
 Increments as a wall clock should.
 .It Dv CLOCK_MONOTONIC
 Increments monotonically in SI seconds.
+.It Dv CLOCK_UPTIME
+Starts at zero when the kernel boots and increments monotonically in SI seconds
+while the machine is running.
```
concussious, line 102: This is cool!
```suggestion
SI seconds since kernel boot.
```
Does this read the same to everyone else?
Sorry, I don't have deep understanding of timing.

bsdimp: Uptime includes while suspended, while monotonic stops while suspended.

concussious: Thank you! What about this?

bsdimp: I stand corrected. Boot time and monotonic time are identical and do tick while the system is suspended. uptime only ticks when the system is up and not suspended. I got the two confused in my earlier comments, sorry.
The clock_gettime.2 man page is wrong about BOOTTIME. I just checked the code.

bsdimp:
Merged, slightly edited, to main. 172328c477b13458c5faa4cf81ad8669f6852aee

bsdimp:
In the end, I had to tweak this a little bit, and updated clock_gettime man page to be clear. Thanks for the submission!

PR #1304, opened 2024-06-26
remove a redundant check on PGA_WRITEABLE

wy-chung:
"#!pmap_page_is_write_mapped(m)#" is equal to "(m->a.flags & PGA_WRITEABLE) == 0" so one of the if statements can be removed.

### DIFF
--- a/sys/riscv/riscv/pmap.c
+++ b/sys/riscv/riscv/pmap.c
@@ -4614,11 +4614,6 @@ pmap_clear_modify(vm_page_t m)
 	    ("%s: page %p is not managed", __func__, m));
 	vm_page_assert_busied(m);
 
-	/*
-	 * If the page is not PGA_WRITEABLE, then no PTEs can have PG_M set.
-	 * If the object containing the page is locked and the page is not
-	 * exclusive busied, then PGA_WRITEABLE cannot be concurrently set.
-	 */
 	if (!pmap_page_is_write_mapped(m))
 	        return;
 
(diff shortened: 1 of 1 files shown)


sys/riscv/riscv/pmap.c
```diff
@@ -4611,19 +4611,17 @@ pmap_clear_modify(vm_page_t m)
...
 	vm_page_assert_busied(m);
 
-	if (!pmap_page_is_write_mapped(m))
-	        return;
-
 	/*
 	 * If the page is not PGA_WRITEABLE, then no PTEs can have PG_M set.
 	 * If the object containing the page is locked and the page is not
 	 * exclusive busied, then PGA_WRITEABLE cannot be concurrently set.
 	 */
```
mhorne, line 4621:
```suggestion

```
This comment should be removed, it is no longer accurate and not present in other pmap implementations.

In fact it seems this comment and check were missed in the previous change, see: 638f867814.

mhorne (changes requested):
Thanks for finding this! A tweak is required but the change is good.

Please confirm or adjust the Author field of the commit.
Currently it reads: `wy-chung <wy-chung@outlook.com>`

sys/riscv/riscv/pmap.c
bsdimp: Should we do this here, or as a separate commit?

wy-chung: The user.name and user.email are correct.

mhorne (approved):

mhorne:
Thanks. Merged to `main` as 2eb64df7f53d6. This PR can be closed.

PR #1305, opened 2024-06-26
Update default pool/path in pam_zfs_key.c

eborisch:
While _rpool_ is the default name throughout OpenZFS codebase -- in comments and tests, primarily -- this instance directly impacts the ease of use of pam_zfs_key, improving POLA.

With this change, the default value for the pam_zfs_key paramater `homes=...` matches what _bsdinstall_ (https://cgit.freebsd.org/src/tree/usr.sbin/bsdinstall/scripts/zfsboot#n44)  creates by default on a zfs install: _zroot/home_.

### DIFF
--- a/sys/contrib/openzfs/contrib/pam_zfs_key/pam_zfs_key.c
+++ b/sys/contrib/openzfs/contrib/pam_zfs_key/pam_zfs_key.c
@@ -449,7 +449,7 @@ static int
 zfs_key_config_load(pam_handle_t *pamh, zfs_key_config_t *config,
     int argc, const char **argv)
 {
-	config->homes_prefix = strdup("rpool/home");
+	config->homes_prefix = strdup("zroot/home");
 	if (config->homes_prefix == NULL) {
 		pam_syslog(pamh, LOG_ERR, "strdup failure");
 		return (PAM_SERVICE_ERR);


bsdimp (approved):
Looks good to me.

sys/contrib/openzfs/contrib/pam_zfs_key/pam_zfs_key.c
```diff
@@ -449,7 +449,7 @@ static int
 zfs_key_config_load(pam_handle_t *pamh, zfs_key_config_t *config,
     int argc, const char **argv)
 {
-	config->homes_prefix = strdup("rpool/home");
+	config->homes_prefix = strdup("zroot/home");
```
bsdimp, line 452: Though it's ok, it would be better to have a FreeBSD ifdef and bring in the change via upstream.

eborisch: Should I just take this directly (with ifdefs) to OpenZFS, then? (Am I reading that right?) It doesn't seem super robust on FreeBSD to begin with; I can get it to mostly work with console logins, but SSH appears to be a non-starter (session complains about being unable to get password from PAM stack.)

bsdimp: My apologies for the tardy response.... I've been away due to some personal issues...

I think that you likely should... But I'm not familiar with all the bits that are needed to make this functionality work completely on FreeBSD.

I've not tried this feature out at all...

bsdimp:
Sorry for the long delay. I think the best way forward is to submit it upstream, #ifdef'd for FreeBSD, and we'll bring it into the tree that way. Once it's upstreamed, we'll figure out the best way to land it.

sys/contrib/openzfs/contrib/pam_zfs_key/pam_zfs_key.c
emaste: For upstream should we have a more extensive change to use `zroot` rather than `rpool` generally on FreeBSD?

bsdimp:
Ping? What's the status of this upstream?

eborisch:
> Ping? What's the status of this upstream?

Sorry; out of sight, out of mind... I'll try to do this and link back.

eborisch:
Fiinally got back to this and submitted an upstream PR. https://github.com/openzfs/zfs/pull/17600

bsdimp:
> Fiinally got back to this and submitted an upstream PR. [openzfs/zfs#17600](https://github.com/openzfs/zfs/pull/17600)

Thanks

bsdimp:
This has merged upstream and flowed back down to FreeBSD. Closing! Thanks for making that happen.

PR #1307, opened 2024-06-30
hier.7: installations by pkg(8), not pkg(7)

grahamperrin:
```text
Consistent with
https://github.com/freebsd/freebsd-src/pull/831
```

### DIFF
--- a/share/man/man7/hier.7
+++ b/share/man/man7/hier.7
@@ -479,7 +479,7 @@ Z file system utilities
 .Pp
 .It Pa local/
 local executables, libraries, etc, installed by
-.Xr pkg 7
+.Xr pkg 8
 or
 .Xr ports 7
 .Pp


grahamperrin:
Section 8 – the _real_ pkg. 

Not pkg(7), which is integral to base and is used to bootstrap pkg(8).

Consistent with <https://github.com/freebsd/freebsd-src/commit/03ee09148950c9ea99987bfee8a769473079745b#diff-8abb4c93d0b75b28f1302098529c9e4b2e261df509386123b92a2d5992d76488L71-L73>.

@concussious @mhorne

share/man/man7/hier.7
```diff
@@ -479,7 +479,7 @@ Z file system utilities
 .Pp
 .It Pa local/
 local executables, libraries, etc, installed by
-.Xr pkg 7
+.Xr pkg 8
 or
 .Xr ports 7
 .Pp
```
grahamperrin, line 485:
```suggestion
(including use of the package management system for
.Xr ports 7
)
.Pp
```

The current edition implies that for installations, ports(7) is an alternative to pkg. Related misunderstandings are commonplace.

From <https://man.freebsd.org/cgi/man.cgi?query=ports&sektion=7&manpath=freebsd-current#TARGETS>: 

> … Usually, you will only use the install target. 

– and: 

> Install the port and register	it with the package system.

That is, the package management system. From Warren Block's [What is pkg on FreeBSD?](http://www.wonkity.com/~wblock/docs/html/pkg.html) (2014) point 4: 

> …  When a port is compiled, it actually creates a package. The package is then installed and managed by the package management system. A binary package is the same, it has just been compiled on a different computer. 

Specifically [pkg-install(8)](https://man.freebsd.org/cgi/man.cgi?query=pkg-install&sektion=8&manpath=freebsd-ports), I guess. pkg(8) is close enough.

concussious:
Thanks for tagging me!

Perhaps your problem could be solved by pkg(7) could use a clearer introductory paragraph?

Rationale:

Hier is an index.
Compiling ports is a first class use case.

Hier must be as concise as possible or this document becomes illegible. Mentioning only pkg(7)/ports(7) is as concise as possible. If you don't think the introductory paragraphs to those pages are good enough then I would love to review a PR on that.

You can't just read only the index. The index points to where you've got to look.

If they've never touched pkg, then they don't have pkg(8), they need to look at pkg(7).

If the introductory paragraph of pkg(7) isn't clear enough, and you think this is a real problem, IMHO that is the place to solve this.

concussious:
> Not pkg(7)

We did this on purpose because the pkg(7) manual will be available when the pkg(8) manual is not installed.

share/man/man7/hier.7
concussious: > Either (a) clarify the mention of ports, or (b) don't mention ports(7) here.

This is an index. Index points to the overview on the topic. ports(7) is the topic overview for contributed applications.

Previously we only linked to ports(7). If the compromise I proposed is causing issues, we should revert to the previous behavior of only mentioning ports(7).

concussious:
Maybe instead we can work together to fix the problem in the description of ports(7). Here's what I've got so far:

https://github.com/freebsd/freebsd-src/pull/1312

PR #1309, opened 2024-06-30
freebsd-update: rsautl is deprecated, use pkeyutl

jlduran:
The command [openssl-rsautl(1)](https://www.openssl.org/docs/man1.1.1/man1/rsautl.html) has been deprecated in OpenSSL 3.0.  The [openssl-pkeyutl(1)](https://www.openssl.org/docs/man3.1/man1/openssl-pkeyutl.html) command should be used instead.

### DIFF
--- a/usr.sbin/freebsd-update/freebsd-update.sh
+++ b/usr.sbin/freebsd-update/freebsd-update.sh
@@ -1307,7 +1307,7 @@ fetch_tag () {
 		return 1
 	fi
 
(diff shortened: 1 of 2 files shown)


jlduran:
Sorry about that. I could add a test plan:

Before:

    # freebsd-update --debug fetch
    ...
    The command rsautl was deprecated in version 3.0. Use 'pkeyutl' instead.

After:
Without the warning.

PR #1310, opened 2024-07-01
kvprintf(): Fix '+' conversion handling

sebhub:
For example, printf("%+i", 1) prints "+1".  However, kvprintf() did print just "1" for this example.  According to PRINTF(3):

  A sign must always be placed before a number produced by a signed conversion.

For "%+r" radix conversions, keep the "+" handling as it is, since this is a non-standard conversion.

This change allows to support the ' ' conversion modifier in the future.

### DIFF
--- a/UPDATING
+++ b/UPDATING
@@ -2022,6 +2022,8 @@ COMMON ITEMS:
 	make -DALWAYS_CHECK_MAKE buildkernel KERNCONF=YOUR_KERNEL_HERE
 	make -DALWAYS_CHECK_MAKE installkernel KERNCONF=YOUR_KERNEL_HERE
 
+	If you are running kernel modules from ports, see FOOTNOTE [1].
+
 	To test a kernel once
 	---------------------
 	If you just want to boot a kernel once (because you are not sure
@@ -2038,8 +2040,7 @@ COMMON ITEMS:
 	<make sure you have good level 0 dumps>
 	make buildworld
 	make buildkernel KERNCONF=YOUR_KERNEL_HERE
-	make installkernel KERNCONF=YOUR_KERNEL_HERE
-							[1]
+	make installkernel KERNCONF=YOUR_KERNEL_HERE	[1]
 	<reboot in single user>				[3]
 	etcupdate -p					[5]
 	make installworld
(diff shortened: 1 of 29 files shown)


bsdimp:
This looks like it changes the meaning of both '%-p' and '%+p'. before it would always set sign to 0. Now it doesn't which seems like it would be weird.

sebhub:
I have to admit that I did test the '%+p' using glibc. It seems that it is not defined by POSIX.
```c
#include <stdio.h>

void print(void *p)
{
   printf("'%20p' '%-20p' '%+20p'\n", p, p, p);
}

int main()
{
  print(0);
  print((void*)1L);
  print((void*)-1L);
}
```
On FreeBSD:
```cc -Wall -Wextra -pedantic main.c
main.c:5:29: warning: flag '+' results in undefined behavior with 'p' conversion specifier [-Wformat]
   printf("'%20p' '%-20p' '%+20p'\n", p, p, p);
                           ~^~~~
1 warning generated.
./a.out
'                 0x0' '0x0                 ' '                 0x0'
'                 0x1' '0x1                 ' '                 0x1'
'  0xffffffffffffffff' '0xffffffffffffffff  ' '  0xffffffffffffffff'

```
On glibc:
```
cc -Wall -Wextra -pedantic main.c
main.c: In function ‘print’:
main.c:5:32: warning: '+' flag used with ‘%p’ gnu_printf format [-Wformat=]
    printf("'%20p' '%-20p' '%+20p'\n", p, p, p);
                                ^
./a.out
'               (nil)' '(nil)               ' '               (nil)'
'                 0x1' '0x1                 ' '                +0x1'
'  0xffffffffffffffff' '0xffffffffffffffff  ' ' +0xffffffffffffffff'
```
```

sebhub:
Please let me know if you are in principle fine with this change. I can change the '%+p' behaviour to match with the FreeBSD user space.

sebhub:
I changed the patch to keep the current "%+p" behaviour.

bsdimp:
Sorry missed these comments... I think I'm fine with the change (either way, but happier with it matching user space better). I started looking at this, then got distracted and didn't get back to it.

sebhub:
Anything left to do on my side?

bsdimp:
> Anything left to do on my side?

No. I think it's ready to go, but am not sure. Was going to look closely on my next pr landing day... but last friday and this have gotten away from me so it will be maybe monday or tiesday this next week when i hope to do a mini one.

PR #1311, opened 2024-07-01
riscv: Don't need to check the return value with NULL

wy-chung:
The return value from pmap_l1_to_l2 and pmap_l2_to_l3 will never be NULL
so don't need to check their return value with NULL.


bsdimp:
@mhorne @markjdb comments?

markjdb (approved):
This looks good to me.

mhorne (approved):
Thanks, I will merge.

PR #1314, opened 2024-07-04
Stop forcing -g in mpi3mr module because it breaks non-debug install.

ericfreebsd:
Stop forcing -g in mpi3mr module because it breaks non-debug install.


ericfreebsd:
It is probably more correct to pass on whatever debug settings were specified, but this makes my builds work again.

ericfreebsd:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=280133

bsdimp (approved):

PR #1317, opened 2024-07-05
Potential typo/copy issue

tspence:
This test here appears to be clamping the start and stop values to a minimum and maximum. But it looks like one of these lines may have been copied incorrectly.

This code pattern is probably intended to test both start and stop against sensible values, but accidentally only changes the start value.

looking at the logic, I suspect that the impact of this bug might be negligible in most cases since it would simply reduce the performance impact of packet coalescing by avoiding aggressive start values.


bsdimp:
Seems legit... added jhb

bsdimp:
Thanks for your contribution!

PR #1319, opened 2024-07-06
freebsd-version.1: corrections and suggested improvements

grahamperrin:
For four, the word 'neither' is inappropriate.

Make clearer that with multiple options, user-specified order is disregarded.

Various other suggestions.

### DIFF
--- a/bin/freebsd-version/freebsd-version.1
+++ b/bin/freebsd-version/freebsd-version.1
@@ -81,16 +81,15 @@ userland
 jail userland.
 .El
 .Sh IMPLEMENTATION NOTES
-The
 .Nm
-utility should provide the correct answer in the vast majority of
-cases, including those where
+should provide the correct answer in the vast majority of cases,
+including those where
 .Xr freebsd-update 8 
 changes the patch level of the kernel but not userland (or vice versa).
 .Pp
-To determine the name (and hence the location) of a custom kernel, the
+To determine the name (and hence the location) of a custom kernel,
 .Nm
-utility will attempt to parse
+will attempt to parse
 .Pa /boot/defaults/loader.conf
 and
 .Pa /boot/loader.conf ,
@@ -131,7 +130,7 @@ env ROOT=/mnt /mnt/bin/freebsd-version -ku
 .Sh HISTORY
 The
 .Nm
-command appeared in
+utility first appeared in
 .Fx 10.0 .
 .Sh AUTHORS
 The
--- a/bin/freebsd-version/freebsd-version.1
+++ b/bin/freebsd-version/freebsd-version.1
@@ -114,7 +114,8 @@ Path to the root of the filesystem in which to look for
 and the kernel.
 .El
 .Sh EXAMPLES
-To determine the version of the installed userland:
+Show userland information alone, without specifying option
+.It Fl u :
 .Bd -literal -offset indent
 freebsd-version
 .Ed
--- a/bin/freebsd-version/freebsd-version.1
+++ b/bin/freebsd-version/freebsd-version.1
@@ -127,6 +127,7 @@ env ROOT=/mnt /mnt/bin/freebsd-version -ku
 .Ed
 .Sh SEE ALSO
 .Xr uname 1 ,
+.Xr freebsd-update 8 ,
 .Xr loader.conf 5
 .Sh HISTORY
 The
--- a/bin/freebsd-version/freebsd-version.1
+++ b/bin/freebsd-version/freebsd-version.1
@@ -40,7 +40,6 @@ utility makes a best effort to determine the version and patch level
 of the kernel and / or userland.
 .Pp
 .Ss Options
-.Pp
 .Bl -tag -width Fl
 .It Fl k
 Print the version and patch level of the installed kernel.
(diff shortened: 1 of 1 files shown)


grahamperrin:
Affects: <https://man.freebsd.org/cgi/man.cgi?query=freebsd-version&sektion=1&manpath=freebsd-current>.

Partly related: <https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=279947#c2>.

@dag-erling

grahamperrin:
<details>
<summary>Rendered</summary>

```text
% man -P cat /tmp/freebsd-version.1
FREEBSD-VERSION(1)      FreeBSD General Commands Manual     FREEBSD-VERSION(1)

NAME
     freebsd-version – print the version and patch level of the installed
     system

SYNOPSIS
     freebsd-version [-kru] [-j jail]

DESCRIPTION
     The freebsd-version utility makes a best effort to determine the version
     and patch level of the kernel and / or userland.

   Options
     -k          Print the version and patch level of the installed kernel.
                 Unlike uname(1), if a new kernel has been installed but the
                 system has not yet rebooted, freebsd-version will print the
                 version and patch level of the new kernel.

     -r          Print the version and patch level of the running kernel.
                 Unlike uname(1), this is unaffected by environment variables.

     -u          Print the version and patch level of the installed userland.
                 The same information will be printed if no option is
                 specified.

     -j jail     Print the version and patch level of the installed userland
                 in the given jail specified by jid or name.  This option can
                 be specified multiple times.

     If multiple options are specified, information will be in a preset order
     (regardless of user-specified order):

     -   installed kernel
     -   running kernel
     -   userland
     -   jail userland.

IMPLEMENTATION NOTES
     freebsd-version should provide the correct answer in the vast majority of
     cases, including those where freebsd-update(8) changes the patch level of
     the kernel but not userland (or vice versa).

     To determine the name (and hence the location) of a custom kernel,
     freebsd-version will attempt to parse /boot/defaults/loader.conf and
     /boot/loader.conf, looking for definitions of the kernel and bootfile
     variables, both with a default value of “kernel”.  It may however fail to
     locate the correct kernel if either or both of these variables are
     defined in a non-standard location, such as in /boot/loader.rc.

     Userland information is hardcoded into freebsd-version at build time.

ENVIRONMENT
     ROOT  Path to the root of the filesystem in which to look for loader.conf
           and the kernel.

EXAMPLES
     Show userland information alone, without specifying option -u:

           freebsd-version

     To inspect a system being repaired using a live system:

           mount -rt ufs /dev/ada0p2 /mnt
           env ROOT=/mnt /mnt/bin/freebsd-version -ku

SEE ALSO
     uname(1), freebsd-update(8), loader.conf(5)

HISTORY
     The freebsd-version utility first appeared in FreeBSD 10.0.

AUTHORS
     The freebsd-version utility and this manual page were written by
     Dag-Erling Smørgrav <des@FreeBSD.org>.

FreeBSD 15.0-CURRENT             July 6, 2024             FreeBSD 15.0-CURRENT
% 
```
</details>

concussious:
> 14.1-RELEASE-p1 (patch level 1) with pkgbase is inconsistent with the 
14.1-RELEASE    (non-patched)   that results from use of freebsd-update.

> We can begin with simple, non-contentious freebsd-src pull requests for freebsd-version(1).

> .Ss Options

You are trolling.

concussious:
> With four options (not two), the word 'neither' is inappropriate.

This is a lie.

concussious:
> With four options (not two), the word 'neither' is inappropriate.

No.

"1. Not either; generally used to introduce the first of two or more coördinate clauses of which those that follow begin with nor."
~ The GNU version of the Collaborative International Dictionary of English

> Make clearer that with multiple options, user-specified order is disregarded

No. The order is quite clear and written with very strong language. The proposed change is insulting the intended audience. This is not a tutorial or handbook. Further, it reads like our beloved security officer wrote it and you wanna change that, while there's nothing wrong with it... is... shame on you!

> a simple, non-contentious pr

".Ss Options" has been the subject of flame wars. Where do you see us doing that? Why would you want to change it from most other manual says? "The following options are available" is *FreeBSD manual culture*. $otherOS all have a different style. This so unnecessary, it's a stylistic regression, it's triggering, and to call it "simple (and) non-contentious" when you didn't have to say anything, seems like (as a former comitter) on top of everything your underlying intention is... doesn't have enough respect. I am extremely suspicious of anyone wants to change anything on planet earth without clear evidence that they appreciate and respect it first.

Please extend to me the grace you extend to random people who publicly disparage our OS on Reddit for the best effort nature of this response. I have explained *why* this doesn't work and is offensive, to the best of my emotional capacity, for your actionable benefit, at peril to my reputation, in the hopes that we will all be improved (including maybe someone will come and coach me how to do this better) so that we can all play better in the future. Thanks.

dag-erling:
You are both extremely tedious.

grahamperrin:
> extremely tedious.

If you want something _truly_ tedious, I might publicly explain why I resigned my `doc` commit bit. 

In the meantime: 

- <https://cgit.freebsd.org/doc/log/access?h=refs/internal/admin>

Rejoice.

PR #1320, opened 2024-07-06
virtio_p9fs: fix panic on qemu/kvm

daniloegea:
When the module is loaded on a system running on qemu/kvm the "modern" virtio infrastructure is used and virtio_read_device_config() will end up calling vtpci_modern_read_dev_config(). This function cannot read values of arbitrary sizes and will panic if the p9fs mount tag size is not supported by it.

Use virtio_read_device_config_array() instead. It was tested on both bhyve and qemu/kvm.

PR:	280098
Co-authored-by:	Mark Peek <mp@FreeBSD.org>

Please see https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=280098 for more details on the investigation.


markpeek (approved):
Thank you for making the PR. I had not gotten around to reproducing the issue to verify the fix.

bsdimp:
I'll land this... any reason you didn't?

PR #1322, opened 2024-07-08
DVD image: repo name: installation_image

grahamperrin:
A DVD is not a CD-ROM.


grahamperrin:
### Context

<https://blendit.bsd.cafe/post/158549> ▶ <https://blendit.bsd.cafe/comment/588885>

> archaic name

----

There's also the name of the file:

https://github.com/freebsd/freebsd-src/blob/39bda097c03780e26e6a25ff59a3e8e77c77563f/release/Makefile#L307

– we can keep this PR simple by not renaming.

PR #1323, opened 2024-07-09
p9fs: remove duplicated code

daniloegea:
This code is using the vnode after it has been released and causing a panic when a p9fs shared volume is unmounted. In fact, it seems like it's just duplicated code left behind from a bad merge.

PR:		279887
Reported by:	Michael Dexter

---

See https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=279887 for more details.

It seems to me that this code originated from here https://github.com/Juniper/virtfs/blob/jnpr/virtfs/sys/dev/virtio/9pfs/virtfs_vnops.c#L97

With this patch I can mount/umount p9fs shares reliably.

Feedback from someone familiar with the code would be appreciated.

### DIFF
--- a/share/man/man7/networking.7
+++ b/share/man/man7/networking.7
@@ -1,7 +1,7 @@
 .\"-
 .\" SPDX-License-Identifier: BSD-2-Clause
 .\"
-.\" Copyright (c) 2024 Alexander Ziaee <concussious@runbox.com>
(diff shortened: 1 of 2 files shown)


markpeek (approved):
Agreed on the duplicated code and this change aligning with the original. Thanks for making this PR.

bsdimp:
this looks good to me. I'll land it after I use it to test running a CI cycle over multiple PRs.

PR #1324, opened 2024-07-11
siphash: allow zero values for final && len in SipBuf()

shiftee:
This change is required for the first loop iteration of the unit test in siphash_test.c to pass.
The current assert corresponds to an XOR, but it should also allow the case where both values are zero.
I have therefore simplified it to asserting that at least one of the values is zero.

### DIFF
--- a/sys/crypto/siphash/siphash.c
+++ b/sys/crypto/siphash/siphash.c
@@ -92,16 +92,21 @@ SipBuf(SIPHASH_CTX *ctx, const uint8_t **src, size_t len, int final)
 {
 	size_t x = 0;
 
-	KASSERT((!final && len > 0) || (final && len == 0),
-	    ("%s: invalid parameters", __func__));
+	/* handle hashing 0 length buffer - needed for test vectors */
+	if (len == 0 && final == 0)
+		return (0);
 
-	if (!final) {
+	if (final) {
+		KASSERT(len == 0, ("%s: invalid len param", __func__));
+		ctx->buf.b8[7] = (uint8_t)ctx->bytes;
+	} else {
+		KASSERT((len > 0) && src && *src,
+			("%s: invalid parameters", __func__));
 		x = MIN(len, sizeof(ctx->buf.b64) - ctx->buflen);
 		bcopy(*src, &ctx->buf.b8[ctx->buflen], x);
 		ctx->buflen += x;
 		*src += x;
-	} else
-		ctx->buf.b8[7] = (uint8_t)ctx->bytes;
+	}
 
 	if (ctx->buflen == 8 || final) {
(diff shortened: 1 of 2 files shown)


bsdimp:
I'd like @cperciva to take a look at this one.
In today's tree, we only ever use it with non-zero hash objects. All the len parameters are sizeof(blah) except one case where a variable is passed in after an if where we test if it is > 0....
So this can't introduce a regression.
I'm not 100% sure about the replacement. for !final, len can be >= 0, at least according to the test code, while we currently assert len > 0. For final, it must be 0 (which matches the only way I see in the code for it to be called with final == 1). That seems to make the assert true, but I think the intent is better matched with the slightly redundant:
KASSERT((!final && len >= 0) || (final && len == 0),...
but that's likely to run afoul of the tautology checks since len >= 0 is always true.
KASSERT(!final || (final && len == 0),...
seems like a better choice to replace it to express the intent more clearly, even though it's the same logically as the proposed change.

Also, maybe siphash_test should be ATF-ified and added to the kyua test suite?

bsdimp:
Also, since this is crypto code, I did some searching....

shiftee@posteo.net appears to be a real person, has code in Linux. It looks like I landed two minor-issue/cleanup pull requests last summer from him as well. He also has at least a 4 year long online presence (though his name is common enough it took a little digging). I hate cyber-stalking submitters, but after xz.... So passes the first round of "seems legit" and the fix seems real, though right on the edge of the "interesting enough" test...

cperciva:
In the case of len == 0 and final == 0 we would end up bcopying zero bytes.  Which is fine... as long as the source isn't a NULL pointer, in which case we could have undefined behaviour.  (I'm not sure about bcopy, but memcpy definitely asserts that its pointers are non-NULL even if copying zero bytes.)

So I'm a bit wary of this change just because it might lead a compiler to deduce that a pointer is non-NULL.  If we really want to allow for len and final to both be zero, we should probably just add an `if (len == 0 && final == 0) return;` to ensure that we cover the edge case safely.

bsdimp:
Yea. Likely we should add that return to SipBuf. No state changes for it, and it avoids a big can of worms around ambiguity (at best) or UB (at worst) around passing NULL to bcopy (which we define, but we define it to memmove which in the kernel has _Nonnull annotations which moves it to clearly UB, though the standard doesn't define it as _Nonnull)...

shiftee:
I have added the if statement as suggested.
I have also broken up the KASSERT to make it simpler and add checks for Null.
Let me know if you think that is a good idea.
main appears to be not building on the CI at the moment due to some ssp changes added a few hours ago.

bsdimp (approved):
I like the update. Thanks!

PR #1327, opened 2024-07-16
usb: increase USB_PORT_RESET_RECOVERY

VexedUXR:
10ms seems to be a little too short for some configurations.

I have two USB devices that needed this change (both Sinowealth mice) to work. From my tests, I could only get the issue to show up on both an `AMD 500 Series Chipset USB 3.1 XHCI Controller` and it's 400 series counterpart. Using other hubs on the system works fine. Note that these hubs work perfectly fine for every other USB device I've tried. I also tried it on two different 500 series controllers, getting the exact same result.

Someone also had a similar issue (I'm 90% sure it was also caused by this) on discord a while ago, and it was fixed by moving the usb device to another hub. I don't have enough info, so I don't know what the problematic device was though.

As a sanity check, I looked at what NetBSD was doing and it looks like they came to the same conclusion.
See https://github.com/NetBSD/src/commit/e988d633755d89def420163bf5224becd55c3bea

### DIFF
--- a/UPDATING
+++ b/UPDATING
@@ -2022,6 +2022,8 @@ COMMON ITEMS:
 	make -DALWAYS_CHECK_MAKE buildkernel KERNCONF=YOUR_KERNEL_HERE
 	make -DALWAYS_CHECK_MAKE installkernel KERNCONF=YOUR_KERNEL_HERE
 
+	If you are running kernel modules from ports, see FOOTNOTE [1].
+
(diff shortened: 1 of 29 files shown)


emaste:
Sounds reasonable

bsdimp (approved):
Absent being able to ask Hans if there's something subtle this would break or some unintended long delay at boot, I think it's fine.

PR #1328, opened 2024-07-16
Replaced the usage of asctime with strftime

RSilicon:
Replaced the usage of the obsolete asctime function with strftime for generating timestamp strings.

### DIFF
--- a/usr.bin/at/at.c
+++ b/usr.bin/at/at.c
@@ -880,8 +880,9 @@ main(int argc, char **argv)
 
 	if (atverify)
 	{
-	    struct tm *tm = localtime(&timer);
-	    fprintf(stderr, "%s\n", asctime(tm));
+	    char time_string[26];
+	    strftime(time_string, sizeof(time_string), "Www Mmm dd hh:mm:ss yyyy\n", localtime(&timer));
+	    fprintf(stderr, "%s\n", time_string);
 	}
 	writefile(timer, queue);
 	break;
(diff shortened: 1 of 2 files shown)


usr.bin/at/at.c
```diff
@@ -880,8 +880,9 @@ main(int argc, char **argv)
 
 	if (atverify)
 	{
-	    struct tm *tm = localtime(&timer);
-	    fprintf(stderr, "%s\n", asctime(tm));
+	    char time_string[26];
+	    strftime(time_string, sizeof(time_string), "Www Mmm dd hh:mm:ss yyyy\n", localtime(&timer));
```
bsdimp, line 884: This makes no sense. Did you test these changes? That's not a valid format string.

bsdimp (changes requested):
Please use valid strftime strings.
And please provide proof that this is working the same before/after.

concussious:
Hi, I am a satisfied lock user. I use lock every single day. Thank you for your interest in lock. Please [don't break my heart](https://docs.freebsd.org/en/articles/committers-guide/#commit-log-message).

jrtc27:
> Hi, I am a satisfied lock user. I use lock every single day. Thank you for your interest in lock. Please [don't break my heart](https://docs.freebsd.org/en/articles/committers-guide/#commit-log-message).

The first three sentences are spam. The final one links to some documentation that may be worth reading for the submitter, but the text used for the link is completely unrelated. This is extremely unhelpful and makes the entire message look like spam. Please do not do this again.

concussious:
Copy, sorry, thank you.

PR #1330, opened 2024-07-17
vmstat: Add root element to libxo output

bramton:
Partial fix for [Bug 254635](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=254635). 

On a side note: The conventions around libxo output seem to be a bit inconsistent. For instance `netstat` and `jls` do not use the program name as root element, but use `statistics` and `jail-information` instead. The `procstat` program uses the `getprogname()` function to set the name of the root element, whilst `arp` hard codes the program name. What are the best practices regarding libxo output?

### DIFF
--- a/usr.bin/vmstat/vmstat.c
+++ b/usr.bin/vmstat/vmstat.c
@@ -66,7 +66,7 @@
 #include <libutil.h>
 #include <libxo/xo.h>
 
-#define VMSTAT_XO_VERSION "1"
+#define VMSTAT_XO_VERSION "2"
 
 static char da[] = "da";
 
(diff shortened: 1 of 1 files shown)


jlduran:
I wonder if such change would warrant a version bump?

bramton:
Agreed. Version number has been bumped.

bsdimp:
I think this is fine...

In an ideal world, we'd document the schemas and versions in the man page, but that's not done elsewhere.

markjdb:
This looks good to me, thank you. The two commits can be squashed together IMO.

> What are the best practices regarding libxo output?

As far as I know, we don't have any documented conventions beyond what's in the libxo documentation itself. I do think it makes sense to have a root element, and that's what most programs do. It would be nice to document these version bumps in the man page as well, explaining what changed, etc., but I think this patch is ok as it is.

bramton:
> This looks good to me, thank you. The two commits can be squashed together IMO.
 
Done!

delphij (approved):

delphij:
Merged as c7dd97ec99687996de49ec0b1a23bf007e5199b9 , thanks!

PR #1331, opened 2024-07-17
netstat: Fix duplicate keys in libxo output

bramton:
Resolves [Bug 254672](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=254672) and [Bug 244589](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=244589)

### DIFF
--- a/usr.bin/netstat/if.c
+++ b/usr.bin/netstat/if.c
@@ -501,7 +501,10 @@ intpr(void (*pfunc)(char *), int af)
 		    IFA_STAT(ipackets), link|network, 1);
 		show_stat("lu", nerr_len, "received-errors", IFA_STAT(ierrors),
 		    link, 1);
+		/* Below is kept for backwards compatibility. Will be removed in the future. */
 		show_stat("lu", nerr_len, "dropped-packets", IFA_STAT(iqdrops),
+		    link, 1);
+		show_stat("lu", nerr_len, "dropped-packets-in", IFA_STAT(iqdrops),
 		    link, 1);
 		if (bflag)
 			show_stat("lu", nbyte_len, "received-bytes",
@@ -516,7 +519,7 @@ intpr(void (*pfunc)(char *), int af)
 		show_stat("NRSlu", nerr_len, "collisions", IFA_STAT(collisions),
 		    link, 1);
 		if (dflag)
-			show_stat("LSlu", nerr_len, "dropped-packets",
+			show_stat("LSlu", nerr_len, "dropped-packets-out",
 			    IFA_STAT(oqdrops), link, 1);
 		xo_emit("\n");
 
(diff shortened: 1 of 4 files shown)


usr.bin/netstat/sctp.c
```diff
@@ -711,7 +711,7 @@ sctp_stats(u_long off, const char *name, int af1 __unused, int proto __unused)
 	    "secret\n");
 	p(sctps_timopathmtu, "\t\t{:pmtu-timer/%ju} "
 	    "{N:/PMTU timer%s fired}\n");
-	p(sctps_timoshutdownack, "\t\t{:shutdown-timer/%ju} "
+	p(sctps_timoshutdownack, "\t\t{:shutdown-ack-timer/%ju} "
```
gmshake, line 714: This looks good to me.

usr.bin/netstat/if.c
```diff
@@ -705,7 +705,7 @@ sidewaysintpr(void)
 	    new->ift_ip - old->ift_ip, 1, 1);
 	show_stat("lu", 5, "received-errors",
 	    new->ift_ie - old->ift_ie, 1, 1);
-	show_stat("lu", 5, "dropped-packets",
+	show_stat("lu", 5, "dropped-packets-in",
```
gmshake, line 708: Convert `dropped-packets` to `dropped-packets-in` breaks IMO. I'd prefer keep `dropped-packets` as is, and add redundant `dropped-packets-in`. The `dropped-packets` can be eventually removed in future release.

gmshake (commented):
I'm OK with the change `shutdown-ack-timer`. That can be committed fast if you want ( and I'd recommend you ) to split this change into two parts.

bramton:
Thanks for reviewing. I have split the commit into two seperate commits.

usr.bin/netstat/if.c
bramton: Done. In the future I will add a `xo_set_version` to this command so that users can track changes in the content.

usr.bin/netstat/if.c
```diff
@@ -501,8 +501,11 @@ intpr(void (*pfunc)(char *), int af)
 		    IFA_STAT(ipackets), link|network, 1);
 		show_stat("lu", nerr_len, "received-errors", IFA_STAT(ierrors),
 		    link, 1);
+		/* Below is kept for backwards compatibility. Will be removed in the future. */
```
bsdimp, line 504: It might be good to s/in the future/before FreeBSD 16/

bsdimp:
And question: do we need to bump the version for this? libxo man pages are kinda silent about when to do that.

gmshake (approved):
Looks good to me.

gmshake:
> And question: do we need to bump the version for this? libxo man pages are kinda silent about when to do that.

@bsdimp  IMO, for the fixes, `dropped-packets-out` and `shutdown-ack-timer`, no need to bump the version.
But for the new introduced `dropped-packets-in` I guess it will be easy to track the change if bump the version.

bramton:
> And question: do we need to bump the version for this? libxo man pages are kinda silent about when to do that.

netstat didn't have version information yet. This is added in 0d2e74b1e6a67907f2668ed85286f78a1091d2e7.

bsdimp:
I've had to (partially) revert this pull request. For the bits I reverted, please resubmit as a new PR. Thanks!

PR #1334, opened 2024-07-21
imgact_elf: Load APE executables faster

jart:
Actually Portable Executable (APE) is a file format that polyglots Windows Portable Executable with a UNIX Sixth Edition shell script (which predates the shebang line). It was introduced in 2020 and is specified formally in: https://github.com/jart/cosmopolitan/blob/master/ape/specification.md v0.1 and we even got POSIX to change their rules to let us use this clever hack https://www.austingroupbugs.net/view.php?id=1250 Our file format is widely used on sites like Hugging Face to distribute ML models, and gets hundreds of thousands of downloads per month.

This change improves the performance of loading APE binaries since they'll no longer need to be interpreted by /bin/sh. APE executables are basically just ELF executables, except with a shell script prepended to the start of the file. All we need to do is tell the FreeBSD kernel's normal ELF loader how to find the ELF header.

### DIFF
--- a/sys/kern/imgact_elf.c
+++ b/sys/kern/imgact_elf.c
@@ -85,7 +85,7 @@
 
 static int __elfN(check_header)(const Elf_Ehdr *hdr);
 static Elf_Brandinfo *__elfN(get_brandinfo)(struct image_params *imgp,
-    const char *interp, int32_t *osrel, uint32_t *fctl0);
+    const char *interp, int32_t *osrel, uint32_t *fctl0, const Elf_Ehdr *hdr);
 static int __elfN(load_file)(struct proc *p, const char *file, u_long *addr,
     u_long *entry);
 static int __elfN(load_section)(const struct image_params *imgp,
@@ -97,7 +97,7 @@ static bool __elfN(freebsd_trans_osrel)(const Elf_Note *note,
 static bool kfreebsd_trans_osrel(const Elf_Note *note, int32_t *osrel);
 static bool __elfN(check_note)(struct image_params *imgp,
     Elf_Brandnote *checknote, int32_t *osrel, bool *has_fctl0,
-    uint32_t *fctl0);
+    uint32_t *fctl0, const Elf_Ehdr *hdr);
 static vm_prot_t __elfN(trans_prot)(Elf_Word);
 static Elf_Word __elfN(untrans_prot)(vm_prot_t);
 static size_t __elfN(prepare_register_notes)(struct thread *td,
@@ -335,9 +335,8 @@ __elfN(brand_inuse)(Elf_Brandinfo *entry)
 
 static Elf_Brandinfo *
 __elfN(get_brandinfo)(struct image_params *imgp, const char *interp,
-    int32_t *osrel, uint32_t *fctl0)
+    int32_t *osrel, uint32_t *fctl0, const Elf_Ehdr *hdr)
 {
-	const Elf_Ehdr *hdr = (const Elf_Ehdr *)imgp->image_header;
 	Elf_Brandinfo *bi, *bi_m;
 	bool ret, has_fctl0;
 	int i, interp_name_len;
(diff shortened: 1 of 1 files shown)


concussious:
Holy shit you are so cool

bsdimp:
Does APE define an ABI? Or does the binary have to be compiled specifically for FreeBSD? Or do details like that live in the printf ELF header?

jart:
I've spent the last four hours greatly expanding the [APE Specification Document](https://github.com/jart/cosmopolitan/blob/master/ape/specification.md) to answer your questions and more. APE software conforms to the System V ABI, just like FreeBSD, except it needs to adopt the restrictions of other OSes too. For example, we use `-mno-red-zone` because APE programs boot from BIOS and run on Windows too. The main thing that's hairy is TLS. Although that's not an issue with FreeBSD, since your OS is one of the most awesome and permissive ones when it comes to granting control and power to change things to userspace programs!

bsdimp:
Given that this is a non-standard image format, I wonder if there's a case to be made to make it optional?
Maybe options IMGACT_APE? in opt_imgact.h?

jart:
I'm happy to add an option to the kernel build config that lets users disable APE loading if they don't want it. For example, the IT department at an enterprise might want to have a policy where their FreeBSD kernel images is hardened to only load ELF. The issue is that this might leave them in a worse spot, because APE would fallback to running as a shell script, which is less reliable. For example, when the 9kb `ape` interpreter is extracted to the home folder, it has to drop an executable file, which might make tools like CrowdStrike unhappy. When the `ape` interpreter loads the binary, it lacks the ability to protect the mapped memory in such a way that `ETXTBSY` will be raised, should the executable image be accidentally modified. That's bad, because if a file is `cp`'d over, it'll cause the pages loaded off disk to be blended between two incompatible images, because only the kernel is able to offer protection against that. So please consider that FreeBSD _already supports_ the APE format. All we're doing here is making it faster and more reliable.

sys/kern/imgact_elf.c
```diff
@@ -999,6 +998,103 @@ __elfN(enforce_limits)(struct image_params *imgp, const Elf_Ehdr *hdr,
...
+	int desired_machine;
+
+#if defined(__aarch64__)
+	desired_machine = EM_AARCH64;
+#elif defined(__powerpc64__)
+	desired_machine = EM_PPC64;
+#elif defined(__riscv)
+	desired_machine = EM_RISCV;
+#else
+	desired_machine = EM_X86_64;
```
brooksdavis, line 1063: I'd prefer an explicit match for x86-64 and a `#error` default case.  It's almost certainly wrong that armv7 and i386 will get EM_X86_64. Is the APE format 64-bit only and thus this all needs a appropriate guard?

sys/kern/imgact_elf.c
```diff
@@ -999,6 +998,103 @@ __elfN(enforce_limits)(struct image_params *imgp, const Elf_Ehdr *hdr,
...
+#endif
+
+	i = 0;
+keep_looking:
```
brooksdavis, line 1067: Not a requirement, but this feels like it would be a pretty natural use of `do {...} while (ehdr->e_machine != desired_machine);`

brooksdavis (commented):
Except for not handling 32-bit this change seems generally fine.

I agree with the desire to make this an option.

I'd probably make the explicit `hdr` argument change in a separate commit.

Tagging @bsdjhb in case he sees issues for our CHERI support downstream.

bsdimp:
We need this to have a proper option. Plus it looks like some of @brooksdavis 's concerns aren't address yet and they need to be answered.

sys/kern/imgact_elf.c
bsdjhb: Isn't this just `ELF_ARCH`?

arthepsy:
Hey. This looks really interesting! Could we have some movement on this, i.e., what is missing to move forward? Does this require more input from Justine or something else?

bsdimp:
> Hey. This looks really interesting! Could we have some movement on this, i.e., what is missing to move forward? Does this require more input from Justine or something else?

Yes. As I said in my Oct 4th message: many of brooks' concerns haven't been addressed. Plus it's not been made an option.

bsdimp:
OK. I asked for changes months ago, brooks asked before that. I'm going to close this due to being inactive. Please resubmit if you address the issues brought up.

bsdimp:
I think there's some reason to reconsider this....  I'll look into it. I closed it prematurely.

clausecker:
I think this feature might be better as a loadable kernel module supplied as a port for now; while I think APE is cool, I am not sure if we should support it as an official FreeBSD feature with all the responsibility that comes with doing so, especially given that it works fine without the kernel support.

PR #1335, opened 2024-07-21
fix getrandom() not declared if crosscompile from macOS

smurfd:
Hey
When crosscompiling `buildworld` from macOS i got an error saying that getrandom() was not declared
Added a include `sys/random.h`

```
freebsd-src/crypto/openssl/providers/implementations/rands/seeding/rand_unix.c:399:12: error: call to undeclared function 'getrandom'; ISO C99 and later do not support implicit
  399 |     return getrandom(buf, buflen, 0);
      |            ^
1 error generated.
*** [rand_unix.o] Error code 1
```

### DIFF
--- a/crypto/openssl/providers/implementations/rands/seeding/rand_unix.c
+++ b/crypto/openssl/providers/implementations/rands/seeding/rand_unix.c
@@ -43,6 +43,7 @@
 
 #if (defined(OPENSSL_SYS_UNIX) && !defined(OPENSSL_SYS_VXWORKS)) \
      || defined(__DJGPP__)
+# include <sys/random.h>
 # include <sys/types.h>
 # include <sys/stat.h>
 # include <fcntl.h>


jlduran:
Hi! Thank you for noticing it.
The patch you are submitting modifies code under `crypto` directly, at best, it should be upstreamed. It is likely already noticed and tracked in https://github.com/openssl/openssl/pull/24903.
I'll cc/ @kevans91 because of 838b6caababbaaab65659d40a41c2dd46b3a5fd2, which likely introduced this issue.

crypto/openssl/providers/implementations/rands/seeding/rand_unix.c
```diff
@@ -43,6 +43,7 @@
 
 #if (defined(OPENSSL_SYS_UNIX) && !defined(OPENSSL_SYS_VXWORKS)) \
      || defined(__DJGPP__)
+# include <sys/random.h>
```
bsdimp, line 46: why unconditional?

bsdimp:
https://reviews.freebsd.org/D45995 might be better

bsdimp:
Thank you so much for this submission. @bsdjhb had a better patch that he's landed from the code review above, so I'll close this request.

PR #1336, opened 2024-07-22
iflib(4): Replace admin taskqueue group with per-interface taskqueues

kgalazka:
Using one taskqueue group with single thread to execute all admin
tasks may lead to unexpected timeouts when long running task (e.g.
handling a reset after FW update) for one interface prevents
tasks from other interfaces being executed. Taskqueue group API
doesn't let to dynamically add threads, and pre-allocating thread
for each CPU as it's done for traffic queues would be a waste
of resources on systems with small number of interfaces. Replace
global taskqueue group for admin tasks with taskqueue allocated
for each interface to allow independent execution.

To enable that refactor iflib_legacy_setup to follow the pattern from
iflib_irq_alloc_generic function and use iflib_fast_intr as a handler
for RX only interrupts.

### DIFF
--- a/sys/net/iflib.c
+++ b/sys/net/iflib.c
@@ -178,8 +178,9 @@ struct iflib_ctx {
 	struct resource *ifc_msix_mem;
 
 	struct if_irq ifc_legacy_irq;
-	struct grouptask ifc_admin_task;
-	struct grouptask ifc_vflr_task;
+	struct task ifc_admin_task;
+	struct task ifc_vflr_task;
+	struct taskqueue *ifc_tq;
 	struct iflib_filter_info ifc_filter_info;
 	struct ifmedia	ifc_media;
 	struct ifmedia	*ifc_mediap;
@@ -1640,7 +1641,7 @@ static int
 iflib_fast_intr_ctx(void *arg)
 {
 	iflib_filter_info_t info = arg;
-	struct grouptask *gtask = info->ifi_task;
+	if_ctx_t ctx = info->ifi_ctx;
 	int result;
 
 	DBG_COUNTER_INC(fast_intrs);
@@ -1650,8 +1651,7 @@ iflib_fast_intr_ctx(void *arg)
 			return (result);
 	}
 
-	if (gtask->gt_taskqueue != NULL)
-		GROUPTASK_ENQUEUE(gtask);
+	taskqueue_enqueue(ctx->ifc_tq, &ctx->ifc_admin_task);
 	return (FILTER_HANDLED);
 }
 
@@ -4176,7 +4176,7 @@ skip_rxeof:
 }
 
 static void
-_task_fn_admin(void *context)
+_task_fn_admin(void *context, int pending)
 {
 	if_ctx_t ctx = context;
 	if_softc_ctx_t sctx = &ctx->ifc_softc_ctx;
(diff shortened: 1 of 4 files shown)


bsdimp:
Sorry for the tardy feedback. This looks good to me, but I'm not an expert in this area, so I've ask a few other folks to look at this. @bsdjhb

sys/net/iflib.c
```diff
@@ -6428,33 +6428,29 @@ iflib_legacy_setup(if_ctx_t ctx, driver_filter_t filter, void *filter_arg, int *
...
 	info->ifi_filter = filter;
 	info->ifi_filter_arg = filter_arg;
 	info->ifi_task = gtask;
-	info->ifi_ctx = rx_only ? ctx : q;
+	info->ifi_ctx = rxq;
 
 	dev = ctx->ifc_dev;
 	/* We allocate a single interrupt resource */
-	err = _iflib_irq_alloc(ctx, irq, tqrid, rx_only ? iflib_fast_intr_ctx :
+	err = _iflib_irq_alloc(ctx, irq, tqrid, rx_only ? iflib_fast_intr :
```
bsdjhb, line 6459: I don't really understand why we have both `iflib_fast_intr` and `iflib_fast_intr_ctx` TBH.  The only difference is the `ctx` variant checks if the gtask taskqueue is NULL.

sys/net/iflib.c
```diff
@@ -6428,33 +6428,29 @@ iflib_legacy_setup(if_ctx_t ctx, driver_filter_t filter, void *filter_arg, int *
...
-	struct taskqgroup *tqg;
-	void *q;
 	int err, tqrid;
 	bool rx_only;
 
-	q = &ctx->ifc_rxqs[0];
-	info = &rxq[0].ifr_filter_info;
-	gtask = &rxq[0].ifr_task;
-	tqg = qgroup_if_io_tqg;
+	info = &rxq->ifr_filter_info;
```
bsdjhb, line 6446: The newer version here is cleaner.  It's not very clear in the old code why `q` exists at all since it is just an alias of `rxq`.

sys/net/iflib.c
```diff
@@ -1650,8 +1651,12 @@ iflib_fast_intr_ctx(void *arg)
 			return (result);
 	}
 
-	if (gtask->gt_taskqueue != NULL)
-		GROUPTASK_ENQUEUE(gtask);
+	/*
```
bsdjhb, line 1654: The driver should probably be doing bus_teardown_intr() first before draining the taskqueue to close this race instead.

bsdjhb (commented):
I think the idea sounds sensible.

sys/net/iflib.c
kgalazka: iflib_fast_intr_ctx was added in #95246abb21da along with iflib_fast_intr_rxtx. I my understanding the intention was to have separate handler for admin/link interrupts, though at that point there was no actual difference between iflib_fast_intr and iflib_fast_intr_ctx. With this patch iflib_fast_intr_ctx starts using taskqueue while iflib_fast_intr keeps using taskqgroup.

sys/net/iflib.c
kgalazka: So taskqueue_free in iflib_device_deregister should go after IFDI_DETACH, which tears down MSI-X, and iflib_free_intr_mem, which takes care of MSI and legacy interrupts. Kevin Bowling merged first patch from this PR. Do you want me to update it anyway or open a new one?

bsdjhb (approved):

bsdimp:
meged. Couldn't push for some reason though.

PR #1338, opened 2024-07-22
mac_veriexec_parser: Fix open_file error handling

losfair:
NDFREE_PNBUF should be called after checking the return value of vn_open(), and should only be called once.

### DIFF
--- a/sys/security/mac_veriexec_parser/mac_veriexec_parser.c
+++ b/sys/security/mac_veriexec_parser/mac_veriexec_parser.c
@@ -241,9 +241,9 @@ open_file(const char *path, struct nameidata *nid)
 
 	NDINIT(nid, LOOKUP, 0, UIO_SYSSPACE, path);
 	rc = vn_open(nid, &flags, 0, NULL);
-	NDFREE_PNBUF(nid);
 	if (rc != 0)
 		return (rc);
+	NDFREE_PNBUF(nid);
 
 	return (0);
 }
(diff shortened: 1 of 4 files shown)


sys/security/mac_veriexec_parser/mac_veriexec_parser.c
```diff
@@ -241,9 +241,9 @@ open_file(const char *path, struct nameidata *nid)
 
 	NDINIT(nid, LOOKUP, 0, UIO_SYSSPACE, path);
 	rc = vn_open(nid, &flags, 0, NULL);
-	NDFREE_PNBUF(nid);
 	if (rc != 0)
 		return (rc);
+	NDFREE_PNBUF(nid);
 
 	return (0);
```
ehem, line 248: I would tend towards instead doing:
```
if (rc != 0)
        NDFREE_PNBUF(nid);

return (rc);
```
I don't know about other reviewers though.

ehem (approved):
Looks good to me.

gmshake (approved):
Looks good to me.

gmshake:
CC @wojtas-marcin

wojtas-marcin:
cc: @kornelduleba

sys/security/mac_veriexec_parser/mac_veriexec_parser.c
kornelduleba: Did you mean something like this?
```
if (rc == 0)
        NDFREE_PNBUF(nid);
```

The problem here is that if `vn_open` fails it'll free `cn_pnbuf` by itself, causing a double free in the error path.
We might want to do something like:
```
if (rc != 0)
	return (rc);
else
        NDFREE_PNBUF(nid);
```

but IMO the version that was submitted is fine as is.

kornelduleba:
LGTM

bsdimp:
lgtm. Will land.

PR #1339, opened 2024-07-23
nuageinit: Accept plain text passwords

jlduran:
Per [pw(8)], when `-H` is set, the password should be supplied already encrypted in a form suitable for writing directly to the password database (`passwd` in cloud-init tems); `-h` provides a special interface by which interactive scripts can set an account password using [pw(8)] in plain text (`plain_text_passwd` in cloud-init terms).

The default user (_freebsd_) is defined with a `plain_text_passwd` (_freebsd_), not with an encrypted one.

[pw(8)]: https://man.freebsd.org/cgi/man.cgi?query=pw&sektion=8&format=html

---

I suppose this is why the default user/password (freebsd/freebsd) of the image [FreeBSD-14.1-RELEASE-amd64-BASIC-CLOUDINIT-ufs.qcow2.xz](https://download.freebsd.org/releases/VM-IMAGES/14.1-RELEASE/amd64/Latest/FreeBSD-14.1-RELEASE-amd64-BASIC-CLOUDINIT-ufs.qcow2.xz) is not working.

### DIFF
--- a/libexec/nuageinit/nuage.lua
+++ b/libexec/nuageinit/nuage.lua
@@ -112,7 +112,7 @@ local function adduser(pwd)
 		postcmd = " -H 0 "
 	elseif pwd.plain_text_passwd then
 		precmd = "echo "..pwd.plain_text_passwd .. "| "
-		postcmd = " -H 0 "
+		postcmd = " -h 0 "
 	end
 	cmd = precmd .. "pw "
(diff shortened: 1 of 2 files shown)


jlduran:
cc/ @bapt

bapt:
merged thank you!

PR #1340, opened 2024-07-23
nuageinit: Set recommended SSH permissions

jlduran:
As stated in [sshd(8)](https://man.freebsd.org/cgi/man.cgi?query=sshd&&sektion=8&format=html), the recommended permissions for `~/.ssh` are read/write/execute for the user, and not accessible by others; and the recommended permissions for `~/.ssh/authorized_keys` are read/write for the user, and not accessible by others.

---

The administrator will likely enforce these permissions using some orchestration tool. At any rate, it is a good practice, and we are already doing it for other release images.


bapt:
merged thank you!

PR #1341, opened 2024-07-23
nuageinit: Allow special characters in passwords

jlduran:
When passwords are declared in plain text, the vertical bar (`|`), and other characters should be valid.

Adapt the script accordingly by serializing the password.  While here, change the encrypted path, as well.

---

Take this commit with a pinch of salt. I'm not 100% sure this is the proper way to escape things in lua.


jlduran:
@bapt I wonder if it would be safer to serialize everything in `cmd`, as almost everything is user-input?

bapt:
it does not hurt to do it

jlduran:
I'll close this PR, and submit a new one then...

PR #1343, opened 2024-07-24
Avoid division in round_up

VoxSciurorum:
Round up to page size using add and mask instead of divide and multiply.  This is slightly faster, but more importantly it avoids a function call when division is not implemented in hardware.  Looking up the target of the call here can lead to deadlock in rtld.  See recent discussion on -arm and -current.

Some older armv7 CPUs do not have hardware divide.  Theoretically FreeBSD could also target a RISC-V CPU without hardware multiply and divide.

### DIFF
--- a/lib/libthr/thread/thr_stack.c
+++ b/lib/libthr/thread/thr_stack.c
@@ -126,7 +126,7 @@ static char *last_stack = NULL;
 static inline size_t
 round_up(size_t size)
 {
-	return (roundup2(size, _thr_page_size - 1));
+	return (roundup2(size, _thr_page_size));
 }
 
 void
--- a/lib/libthr/thread/thr_stack.c
+++ b/lib/libthr/thread/thr_stack.c
@@ -126,10 +126,7 @@ static char *last_stack = NULL;
 static inline size_t
 round_up(size_t size)
 {
-	if (size % _thr_page_size != 0)
-		size = ((size / _thr_page_size) + 1) *
-		    _thr_page_size;
-	return size;
+	return (roundup2(size, _thr_page_size - 1));
 }
 
 void


lib/libthr/thread/thr_stack.c
```diff
@@ -126,10 +126,7 @@ static char *last_stack = NULL;
 static inline size_t
 round_up(size_t size)
 {
-	if (size % _thr_page_size != 0)
-		size = ((size / _thr_page_size) + 1) *
-		    _thr_page_size;
-	return size;
+	return (size + _thr_page_size - 1) & ~(size_t)(_thr_page_size - 1);
```
brooksdavis, line 129: Any reason not to use `roundup2(size, _thr_page_size)`? It's either a compiler buildit or the same code.

VoxSciurorum: I forgot about `roundup2`.   Konstantin Belousov sent a diff to the mailing lists using that function.

brooksdavis (approved):

lib/libthr/thread/thr_stack.c
```diff
@@ -126,10 +126,7 @@ static char *last_stack = NULL;
 static inline size_t
 round_up(size_t size)
 {
-	if (size % _thr_page_size != 0)
-		size = ((size / _thr_page_size) + 1) *
-		    _thr_page_size;
-	return size;
+	return (roundup2(size, _thr_page_size - 1));
```
brooksdavis, line 129:
```suggestion
	return (roundup2(size, _thr_page_size));
```

I'd merged this, but Doug Moore correctly pointed out that this is wrong.

VoxSciurorum: You're right.  Too many versions of code going around and I stopped thinking about each new one.  It will usually work, but the `- 1` is wrong.

VoxSciurorum: I pushed a fix to the branch associated with this pull request.

lwhsu:
FYI this was reverted in f5c894479e5986323a0d5a5a698c7d2177eec20f as:
https://lists.freebsd.org/archives/dev-commits-src-all/2024-July/043646.html

brooksdavis:
> FYI this was reverted in [f5c8944](https://github.com/freebsd/freebsd-src/commit/f5c894479e5986323a0d5a5a698c7d2177eec20f) as: https://lists.freebsd.org/archives/dev-commits-src-all/2024-July/043646.html

kib beat to to remerging it with 31f688a26d82ce00d1ec7ca9ed17b9914bf5176f

PR #1344, opened 2024-07-24
vmstat: Fix xo_emit format string for percentages

VoxSciurorum:
The "%lld" format string was used with a value of type long. See https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=278532.

### DIFF
--- a/usr.bin/vmstat/vmstat.c
+++ b/usr.bin/vmstat/vmstat.c
@@ -1105,7 +1105,7 @@ percent(const char *name, long pctv, int *over)
 {
 	char fmt[64];
 
-	snprintf(fmt, sizeof(fmt), " {:%s/%%%ulld/%%lld}", name,
+	snprintf(fmt, sizeof(fmt), " {:%s/%%%uld/%%ld}", name,
 	    (*over && pctv <= 9) ? 1 : 2);
 	xo_emit(fmt, pctv);
 	if (*over && pctv <= 9)


amotin (approved):
Seems fine.  I've tested it only on 64-bit amd64, that is likely why it worked for me.

strejda:
> Seems fine. I've tested it only on 64-bit amd64, that is likely why it worked for me.

Unfortunately, vmstat is broken in multiple places. Another example is processor usage.  The xo_emit() function is a pure disaster and should be converted to xo_emit_f() at every point.

bsdimp:
> > Seems fine. I've tested it only on 64-bit amd64, that is likely why it worked for me.
> 
> Unfortunately, vmstat is broken in multiple places. Another example is processor usage. The xo_emit() function is a pure disaster and should be converted to xo_emit_f() at every point.

This makes things less bad, should we not land it because it doesn't do this as well? @strejda

strejda:
I definitely agree, even that using %ul on a signed long argument  is not formally correct. That wasn't meant to be a negative reaction, but a sigh of relief at the poor design of the xo library, especially the existence of the xo_emit() function, which is undoubtedly a step back several decades. And of course also with the hidden hope that someone else will fix the rest  :)

namiltd:
The problem still exists. Tested on FreeBSD 14.1-p7 (32bit) and 14.3-p1 (32bit).
```
 procs    memory    page                      disks       faults       cpu
 r  b  w  avm  fre  flt  re  pi  po   fr   sr ada0  cd0   in   sy   cs us sy id
 0  0  0 153202688 902914048 1994   0  22   0 2228    3    0    0   74 3948  736 47244640264 47244640281 47244640322
```

PR #1347, opened 2024-07-24
if_ethersubr.c: fix ether_8021q_frame to respect VHID from ether_vtag

rozhuk-im:


gmshake:
I do not quite follow you. There's no VHID in ether_vtag. Do you mean VID (vlan id) ?

rozhuk-im:
Yes, we have some code that was work fine starting FreeBSD 10, and now it stop work after:
https://github.com/freebsd/freebsd-src/commit/49d6743da15fe378782e43776df8b4fd4f84c8d0
https://github.com/freebsd/freebsd-src/commit/c7cffd65c5d858425e90b847d2e8e583e3b13bf7
we got packets tagged with VLAN=0 instead of VID that we set in ```ether_vtag```.

sys/net/if_ethersubr.c
```diff
@@ -1439,6 +1439,7 @@ ether_8021q_frame(struct mbuf **mp, struct ifnet *ife, struct ifnet *p,
 	 * If PCP is set in mbuf, use it
 	 */
 	if ((*mp)->m_flags & M_VLANTAG) {
+		vid = EVL_VLANOFTAG((*mp)->m_pkthdr.ether_vtag);
```
gmshake, line 1442: This is wrong. The PCP set in mbuf is from upper layer, aka from ip_output(). It will be layer violation for Upper layer to set vlan id in mbuf.

gmshake (changes requested):

sys/net/if_ethersubr.c
rozhuk-im: Then IP layer should stop write PCP to Ethernet layer.

gmshake: > This is wrong. The PCP set in mbuf is from upper layer, aka from ip_output(). It will be layer violation for Upper layer to set vlan id in mbuf.

Regardless the layer violation, this will break vlan(4). The vlan(4) relies `ether_8021q_frame()` to properly inject VID.

gmshake: > Then IP layer should stop write PCP to Ethernet layer.
That was introduced via 868aabb4708d to support priority per-flow . 

> Upper layer will always write VID=0, as it hardcoded or respect already set VID.
I do not quite follow. I do not see any issues with that, to always write VID=0.

> or respect already set VID.
Upper layers should not do that. From the perspective from upper layers, the VID does not exist, or it can be treated to be 0, exactly.

gmshake:
> Yes, we have some code that was work fine starting FreeBSD 10, and now it stop work after: [49d6743](https://github.com/freebsd/freebsd-src/commit/49d6743da15fe378782e43776df8b4fd4f84c8d0) [c7cffd6](https://github.com/freebsd/freebsd-src/commit/c7cffd65c5d858425e90b847d2e8e583e3b13bf7) we got packets tagged with VLAN=0 instead of VID that we set in `ether_vtag`.

May you please file a bug report with minimal steps to repeat the issue to https://bugs.freebsd.org/bugzilla/ ?

sys/net/if_ethersubr.c
rozhuk-im: Current code base state violates ```m_pkthdr.ether_vtag``` usage.
Current implementation of ip_output() does respect ```m_pkthdr.ether_vtag``` by using ```EVL_APPLY_PRI()``` macro.

Only ```ether_8021q_frame()``` does respect nothing.

rozhuk-im: Current ```ether_set_pcp()``` set PCP + zeroize VID and CFI.

rozhuk-im:
> May you please file a bug report with minimal steps to repeat the issue to https://bugs.freebsd.org/bugzilla/ ?

No, I could not: this is closed source product based on FreeBSD.
Probably some netgraph configs may have same issues.

sys/net/if_ethersubr.c
gmshake: > Current code base state violates `m_pkthdr.ether_vtag` usage.

For inbound packets, `M_VLANTAG` and `ether_vtag` is used by drivers to indicate that the hardware / firmware / driver has processed the 802.1Q tag.

For outbound traffic,  `M_VLANTAG` is used to direct the driver / firmware / hardware to do 802.1Q tag inserting.

> Current implementation of ip_output() does respect `m_pkthdr.ether_vtag` by using `EVL_APPLY_PRI()` macro.

This should be the only violation in current net stack.

> 
> Only `ether_8021q_frame()` does respect nothing.

> Current ether_set_pcp() set PCP + zeroize VID and CFI.

That is on purpose, to support PCP for untagged ( NULL VLAN ID ) outbound traffic.

rozhuk-im: > That is on purpose, to support PCP for untagged ( NULL VLAN ID ) outbound traffic.

This should not go in this way, like ip_output does not force VID=0, CFI=0.

Any way, I make simple implementation that respect ```ether_vtag``` changes from any code path, and now code more clear and predictable.
I need some time to check that it work.

rozhuk-im: It work with if_vlan and code base from company product.

rozhuk-im:
@AlexanderChernikov @kostikbel @emaste review please

gmshake:
> > May you please file a bug report with minimal steps to repeat the issue to https://bugs.freebsd.org/bugzilla/ ?
> 
> No, I could not: this is closed source product based on FreeBSD. Probably some netgraph configs may have same issues.

IMO this fix should go into your closed source product but not FreeBSD. Anyway works that refine / rework the implementation of vlan(4) and PCP are welcome ( if that happens to meet your needs ).

bsdimp:
Based on @gmshake 's feedback, I'm closing this. Please feel free to open a new pull request if you implement his suggestion.

rozhuk-im:
There is no any suggestion from @gmshake.

PR #1348, opened 2024-07-25
tools/build/make.py: Minor typo fix

jlduran:
If the object directory prefix does not exist, `make.py` exits with a message indicating that the chosen prefix does not exit.


emaste:
committed as d55de30efd09ee011bf4604be1c15ce4155775c1

PR #1350, opened 2024-07-26
Add version information to libxo output of several commands

bramton:
Several commands where still missing libxo version information in their output.


markjdb:
Looks good to me. netstat is also missing a version, for what it's worth.

bramton:
Good catch. I have added this in 0d2e74b1e6a67907f2668ed85286f78a1091d2e7

markjdb (approved):
Looks good, thank you. Just waiting for Warner to land this now.

PR #1356, opened 2024-08-02
amd64: move the right parenthesis to the right place

wy-chung:
Like GPROC0_SEL, GUSERLDT_SEL also occupies two consecutive entries in segment descriptor table. So logically to specify the second descriptor of GUSERLDT_SEL, it shuld be (GUSERLDT_SEL  + 1).

### DIFF
--- a/sys/amd64/amd64/machdep.c
+++ b/sys/amd64/amd64/machdep.c
@@ -1382,7 +1382,7 @@ hammer_time(u_int64_t modulep, u_int64_t physfree)
 	 */
 	for (x = 0; x < NGDT; x++) {
 		if (x != GPROC0_SEL && x != (GPROC0_SEL + 1) &&
-		    x != GUSERLDT_SEL && x != (GUSERLDT_SEL) + 1)
+		    x != GUSERLDT_SEL && x != (GUSERLDT_SEL + 1))
 			ssdtosd(&gdt_segs[x], &gdt[x]);
 	}
 	gdt_segs[GPROC0_SEL].ssd_base = (uintptr_t)&pc->pc_common_tss;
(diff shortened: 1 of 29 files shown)


ehem:
The commit message needs to say more.  I'm wondering why the parentheses were used in the first place.  Older style?  Otherwise seems reasonable.

wy-chung:
I have modified the commit message.

bsdimp (approved):

bsdimp:
This looks good. Sorry it took me so long to review such a trivial and obviously right change

sys/amd64/amd64/machdep.c
```diff
@@ -1382,7 +1382,7 @@ hammer_time(u_int64_t modulep, u_int64_t physfree)
 	 */
 	for (x = 0; x < NGDT; x++) {
 		if (x != GPROC0_SEL && x != (GPROC0_SEL + 1) &&
-		    x != GUSERLDT_SEL && x != (GUSERLDT_SEL) + 1)
+		    x != GUSERLDT_SEL && x != (GUSERLDT_SEL + 1))
```
gmshake, line 1385: According to the precedence, the **un-fixed** one behaves the same with the **fixed** one. But I like the later form, which is more clear.

gmshake (approved):
Looks good to me.

gmshake:
> Like GPROC0_SEL, GUSERLDT_SEL also occupies two consecutive entries in segment descriptor table. So logically to specify the second descriptor of GUSERLDT_SEL, it shuld be (GUSERLDT_SEL + 1).

There should be nothing bad happens. See my inline comment.

PR #1358, opened 2024-08-02
Minimal collection for running on Xen/ARM64

ehem:
Let us see whether I can get the Xen/ARM64 commits together in one place in one form.  Many of these pieces are part of other pull requests, but aiming to create a minimal set here.


sys/kern/kern_clock.c
```diff
@@ -531,7 +534,10 @@ hardclock(int cnt, int usermode)
 			if (i > 0 && i <= newticks)
 				watchdog_fire();
 		}
-		intr_event_handle(clk_intr_event, NULL);
+#ifdef DEV_ACPI
+		if (!CK_SLIST_EMPTY(&clk_intr_event->ie_handlers))
+			intr_event_handle(clk_intr_event, NULL);
+#endif
```
ehem, line 540: This isn't actually required for the implementation.  Issue is without this you end up with a mysterious "swi4:" interrupt producing large numbers of stray interrupts and no clue as to what is happening.

ehem:
Commit cecb5a12d6e55ba6f5ca8aaf33483fa3c7f6898a was tested and confirmed to work.

sys/kern/kern_clock.c
```diff
@@ -531,7 +534,10 @@ hardclock(int cnt, int usermode)
 			if (i > 0 && i <= newticks)
 				watchdog_fire();
 		}
-		intr_event_handle(clk_intr_event, NULL);
+#ifdef DEV_ACPI
+		if (!CK_SLIST_EMPTY(&clk_intr_event->ie_handlers))
```
jrtc27, line 538: Why? intr_event_handle checks CK_SLIST_EMPTY at the start and returns EINVAL if so, with the return value being ignored here.

sys/kern/kern_clock.c
```diff
@@ -531,7 +534,10 @@ hardclock(int cnt, int usermode)
 			if (i > 0 && i <= newticks)
 				watchdog_fire();
 		}
-		intr_event_handle(clk_intr_event, NULL);
+#ifdef DEV_ACPI
```
jrtc27, line 537: Why does this need to be under DEV_ACPI? If it builds and works, just isn't currently doing anything, that's fine.

sys/kern/kern_clock.c
ehem: Oh, I'd been detecting it due to #1301.  Not exactly a proper situation though.

jrtc27: Says who? I'd argue this is working as intended, and having to peek inside the intr_event to know whether to call intr_event_handle is a bad design that unnecessarily exposes internal details.

sys/kern/kern_clock.c
ehem: `intr_event_handle()` isn't exactly the cheapest function call.  Partially this seemed since the lack of other uses seems to suggest perhaps a different approach should be used.

The previous comment though is a better reason for dropping from _this_ series.  I'd been bothered by this for a while and it was quite troublesome to track down.

sys/kern/kern_clock.c
ehem: Perhaps this means the generalization of no handlers meaning stray is wrong?  In which case another flag needs to be added for "no handlers isn't stray"?

ehem:
Before being rebased, 54bf57434899d989ec050fecc15b601547e23830 was tested and confirmed to work.  The delta is simply removing the stray clock interrupt fix and pulling 2 of the intermediates.

Mostly there are 3 problems in need of fixes.  First, since Xen interrupts are very dynamic, release needs to work #1281.  Second, some sort of per-processor hook is needed; `pic_init_secondary()` is basically ideal, but the current implementation is problematic, #1280.

Lastly, INTRNG's interface is appropriate for many PICs.  Problem is due to being an unusual bus, this is well suited to a lower-level interface, #1285.  Trying to work with the existing interface would be quite expensive.

Then simply becomes an issue of figuring out how much squashing should be done (while @jgrall did the original implementation, this has rather a lot of changes to get it into shape).

ehem:
So instead of #1280, #1363 also functions to provide access to the appropriate hooks.  For `sys/arm64/include/intr.h` use:
```
+#ifdef XENHVM
+#define INTR_ROOT_XEN	2
+#define INTR_ROOT_NUM	3
+#else
+
 #define INTR_ROOT_NUM	2
+#endif
```
In `sys/dev/xen/intrng/xen_arch_intr.c` add:
```
+	i = intr_pic_claim_root_num(dev, ofw_bus_get_node(dev),
+	    (intr_irq_filter_t *)panic, NULL, INTR_ROOT_XEN);
+	if (i != 0)
+		return (i);
```
After the `intr_pic_register()` call.  This is somewhat hack-ish as we're barely using INTRNG, but it does seem kind of appropriate as it does amount to a disconnected PIC.

sys/kern/subr_intr.c
```diff
@@ -588,131 +703,6 @@ iscr_setup_filter(struct intr_irqsrc *isrc, const char *name,
...
-
-	mtx_lock(&isrc_table_lock);
-	ie = isrc->isrc_event;
-	isrc->isrc_event = NULL;
-	mtx_unlock(&isrc_table_lock);
-
-	if (ie != NULL)
-		intr_event_destroy(ie);
-}
-#endif
```
markjdb, line 715: Why is all this code moving around?

sys/kern/subr_intr.c
```diff
@@ -772,6 +772,25 @@ pic_lookup(device_t dev, intptr_t xref, u_int flags)
...
+pic_lookup_dev(device_t dev, intptr_t xref, u_int flags)
+{
+	struct intr_pic *pic;
+
+	pic = pic_lookup(dev, xref, flags);
+	if (__predict_false(pic == NULL))
+		return (NULL);
+
+	KASSERT((pic->pic_flags & FLAG_TYPE_MASK) == (flags & FLAG_TYPE_MASK),
+	    ("%s: Found wrong type of controler: %s (flags = %u, type = %u)",
```
markjdb, line 788: Typo in "controler"

sys/kern/subr_intr.c
```diff
@@ -815,7 +815,7 @@ pic_create(device_t dev, intptr_t xref, u_int flags)
 	pic->pic_dev = dev;
 	pic->pic_flags = flags;
 	mtx_init(&pic->pic_child_lock, "pic child lock", NULL, MTX_SPIN);
-	SLIST_INSERT_HEAD(&pic_list, pic, pic_next);
+	STAILQ_INSERT_TAIL(&pic_list, pic, pic_next);
```
markjdb, line 818: You should at the very least have a comment mentioning that the insertion order is important, since it preserves an invariant that you apparently depend on.

sys/kern/subr_intr.c
```diff
@@ -800,8 +800,9 @@ pic_create(device_t dev, intptr_t xref, u_int flags)
 	struct intr_pic *pic;
 
 	mtx_lock(&pic_list_lock);
-	pic = pic_lookup_locked(dev, xref, flags);
+	pic = pic_lookup_locked(dev, xref, flags & ~FLAG_TYPE_MASK);
 	if (pic != NULL) {
+		pic->pic_flags |= flags & FLAG_TYPE_MASK;
```
markjdb, line 805: Which PIC driver requires this? I can't see any uses of intr_msi_register() in Xen.

sys/kern/subr_intr.c
```diff
@@ -1537,15 +1537,12 @@ dosoftints(void)
...
-	 * QQQ: Only root PIC is aware of other CPUs ???
-	 */
-	KASSERT(intr_irq_root_dev != NULL, ("%s: no root attached", __func__));
-
-	//mtx_lock(&isrc_table_lock);
-	PIC_INIT_SECONDARY(intr_irq_root_dev);
-	//mtx_unlock(&isrc_table_lock);
+	//mtx_lock(&pic_list_lock);
+	STAILQ_FOREACH(pic, &pic_list, pic_next)
+		PIC_INIT_SECONDARY(pic->pic_dev);
```
markjdb, line 1544: This isn't minimally required for arm64/Xen, as it changes the behaviour of other PIC drivers which implement PIC_INIT_SECONDARY.

markjdb (commented):
I just looked at intrng, not very much at Xen-specific bits yet.

sys/kern/subr_intr.c
ehem: The issue is all of these are `static` without separate declarations.  As such `intr_isrc_register()` calling `isrc_event_create()` means `isrc_event_create()` needs to be above `intr_isrc_register()`.  `isrc_event_destroy()` is moved to keep it next to `isrc_event_create()`.

In turn `isrc_event_create()` passes pointers to `intr_isrc_pre_ithread()`, `intr_isrc_post_ithread()`, `intr_isrc_post_filter()` and `intr_isrc_assign_cpu()` to `intr_event_create()`.  Since those four have no declarations, they must move earlier in the file than `isrc_event_create()`.

sys/kern/subr_intr.c
ehem: I don't depend on it.  There was a statement the proposed IPI PIC for RISC-V would depend on this though.  In truth there are so few uses of `pic_init_secondary()` that doing anything is simply a guessing game.

Since this seems so controversial I'll likely be switching to declaring the Xen entry as a separate root PIC.  This is _mildly_ abusing that functionality, but does seem a good choice as a backup solution.

sys/kern/subr_intr.c
ehem: This was simply trying to preserve behavior of the `pic_init_secondary()` hooks.  By merging them controllers only have this called _once_.

sys/kern/subr_intr.c
ehem: Have you noticed how many implementations of that there are?  The hook is barely used.  Most PICs leave the empty default.  As a result the hook is effectively broken for anything which isn't a root PIC.

Turning it into a root PIC _mildly_ abuses the functionality, but does sort of match what the device-tree looks like.

sys/kern/subr_intr.c
markjdb: If you don't require this, then it doesn't belong in minimal set of patches for running Xen on arm64. I don't have strong opinions on the issue at hand, but since it's difficult to get consensus on intrng changes, your goal is easier to achieve if we can avoid those that aren't strictly needed.

sys/kern/subr_intr.c
markjdb: "Trying" meaning that it's supposed to work but isn't tested? It's not clear to me.

sys/kern/subr_intr.c
markjdb: Of course I noticed, otherwise I wouldn't have commented. The point is that there is more than zero, and I don't see any explanation in the commit logs of how they were tested, nor any explanation for why it is unlikely to cause problems.

sys/kern/subr_intr.c
ehem: Meaning the case I have the GICv3 doesn't have MSI capability, so this isn't well-tested.  _That_ is where this comes into play.

sys/kern/subr_intr.c
ehem: Being rarely used means it is simple to examine the uses and take actions believed to preserve functionality.  I lack any systems to _fully_ test this, but I can confirm it provides an appropriate hook to get Xen fully functional.

The default hook is a NOP.  Only GICv3 tries to propagate the call, but that requires a device-tree which conforms to the preconceptions of someone.  Since the one provide via Xen differs, there is a problem.

Anyway I was planning to abandon this approach as while the issue in the comment remains "QQQ: Only root PIC is aware of other CPUs ???".  Hooking in as a root also works.

I had hoped it would be part of #1363, but is now #1431.

sys/kern/subr_intr.c
ehem: *I* don't depend on any invariant.  There are claims of there being invariants, but I can't verify any of those.  Notably the call I need doesn't care about ordering.

Some sort of hook into processor startup is needed.  Otherwise we end up with all Xen events on a single processor.  There is also an alternative approach which showed up _after_ this had last been updated.

ehem:
I'm expecting the per-processor startup hook to change in light of other recent commits.  Two issues with INTRNG remain though:

1. Since Xen events are rather more dynamic than hardware interrupts, released events need some sort of handling.  The obvious approach is `intr_isrc_deregister()`, problem is this hadn't previously been tested and it leaks the `struct intr_event`.
2. Since they're a specialized type of interrupt, not all of INTRNG's services are needed.  In particular the lower level interface of `intr_add_handler()` (rename `isrc_add_handler()`) and `intr_describe()` (break `intr_describe_irq()` apart) is rather superior.

sys/kern/subr_intr.c
ehem: This was trying to figure out how to get hooked into `pic_init_secondary()`, #1280.  Due to the changes in how this is handled due to #1363, this no longer applies.

Still a bit hazy as to which direction to go with getting connected to `pic_init_secondary()` though.

sys/kern/subr_intr.c
ehem: Similar to the above #1280 and #1363.  I'm still a bit unsure of how to handle this though.

sys/kern/subr_intr.c
ehem: Similar to the above #1280 and #1363.  I'm still a bit unsure of how to handle this though.

PR #1359, opened 2024-08-03
gpiobus(4): Add an acpi variant of gpiobus

VexedUXR:
Add a simple acpi_gpiobus driver

Currently, this only installs an address space handler and attempts to set pin flags using info obtained from GPIO resources. Child enumeration doesn't seem to be necessary at this point, as it would only complicate things. It can be added in the future if needed.

Tested on an Acer one 10 s1002

### DIFF
--- a/sys/dev/gpio/acpi_gpiobus.c
+++ b/sys/dev/gpio/acpi_gpiobus.c
@@ -1,0 +1,310 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2024 Ahmad Khalifa <ahmadkhalifa570@gmail.com>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHORS AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHORS OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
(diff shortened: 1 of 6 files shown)


VexedUXR:
CC @wulf7

sys/dev/gpio/acpi_gpiobus.c
```diff
@@ -0,0 +1,304 @@
...
+static int
+acpi_gpiobus_attach(device_t dev)
+{
+	struct acpi_gpiobus_softc *sc;
+	struct acpi_gpiobus_ctx ctx;
+	ACPI_HANDLE handle;
+	ACPI_STATUS status;
+	int err;
+
+	if ((err = gpiobus_init_softc(dev)) != 0)
```
wulf7, line 232: Use attach method from base driver

sys/dev/gpio/acpi_gpiobus.c
```diff
@@ -0,0 +1,304 @@
...
+	if (res->Type != ACPI_RESOURCE_TYPE_GPIO)
+		return (AE_OK);
+
+	if (ACPI_FAILURE(AcpiGetHandle(ACPI_ROOT_OBJECT,
+	    gpio_res->ResourceSource.StringPtr, &handle)) ||
+	    handle != ctx->dev_handle)
+		return (AE_OK);
+
+	/* Figure out pin flags */
+	flags = 0;
```
wulf7, line 67: acpi_resource to gpio flags conversion can be factored out to separate function

sys/dev/gpio/acpi_gpiobus.c
```diff
@@ -0,0 +1,304 @@
...
+		}
+
+	switch (gpio_res->PinConfig) {
+	case ACPI_PIN_CONFIG_PULLUP:
+		flags |= GPIO_PIN_PULLUP;
+		break;
+	case ACPI_PIN_CONFIG_PULLDOWN:
+		flags |= GPIO_PIN_PULLDOWN;
+		break;
+	}
```
wulf7, line 100: There is also gpio_res->Shareable field to process. It should be mapped to new GPIO_INTR_SHARED flag in sys/gpio.h

sys/dev/gpio/acpi_gpiobus.c
```diff
@@ -0,0 +1,304 @@
...
+		UINT16 pin = gpio_res->PinTable[i];
+
+		if (__predict_false(pin >= super_sc->sc_npins)) {
+			device_printf(super_sc->sc_busdev,
+			    "invalid pin 0x%x, max: 0x%x (bad ACPI tables?)\n",
+			    pin, super_sc->sc_npins - 1);
+			return (AE_LIMIT);
+		}
+
+		GPIO_PIN_SETFLAGS(super_sc->sc_dev, pin, flags);
```
wulf7, line 119: GPIO_PIN_SETFLAGS sets only pin config and IO direction. Please filter out interupt bits with logical AND

sys/dev/gpio/acpi_gpiobus.c
VexedUXR: Oh, nice catch.

Would we need to change the interrupt handling code in gpioc to accommodate the flag?
But then again, I think this attribute only means "Can this be accessed/is this used by other ACPI devices". So in that case, it shouldn't affect interrupt creation.
Hm, I'm not sure.

bsdimp:
@wulf7 are you happy with this...

wulf7:
> @wulf7 are you happy with this...

Almost. I can handle remaining very minor issues after returning of vacations

wulf7:
Committed. Thanks!

I ifdef-ed out interrupt-related parts as they are not used yet and requires some extra work in gpiobus.c and gpioc.c

VexedUXR:
Thanks!

PR #1360, opened 2024-08-04
riscv: a bug in pmap_growkernel when calculating the pindex for PDP page

wy-chung:
Since there are already macros (pmap_l1_pindex and pmap_l2_pindex) for calculating  the pindex for PDP and PTE pages, we should use thoses macros.

### DIFF
--- a/sys/riscv/riscv/pmap.c
+++ b/sys/riscv/riscv/pmap.c
@@ -1888,8 +1888,8 @@ pmap_growkernel(vm_offset_t addr)
 			nkpg = vm_page_alloc_noobj(VM_ALLOC_INTERRUPT |
 			    VM_ALLOC_NOFREE | VM_ALLOC_WIRED | VM_ALLOC_ZERO);
 			if (nkpg == NULL)
-				panic("pmap_growkernel: no memory to grow kernel");
-			nkpg->pindex = kernel_vm_end >> L1_SHIFT;
+				panic("%s: no memory to grow kernel", __func__);
+			nkpg->pindex = pmap_l1_pindex(kernel_vm_end);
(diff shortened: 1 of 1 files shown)


mhorne (approved):
Thanks, I'll merge this. 

The pindex bug is a good find! I think it is without consequence in the current code, but it is certainly incorrect. arm64 has the same bug, it seems.

sys/riscv/riscv/pmap.c
```diff
@@ -1757,8 +1757,8 @@ pmap_alloc_l2(pmap_t pmap, vm_offset_t va, struct rwlock **lockp)
 		l2pg->ref_count++;
 	} else {
 		/* Allocate a L2 page. */
-		l2pindex = pmap_l2_pindex(va) >> Ln_ENTRIES_SHIFT;
-		l2pg = _pmap_alloc_l3(pmap, NUL2E + l2pindex, lockp);
+		ptdpindex = pmap_l1_pindex(va);
+		l2pg = _pmap_alloc_l3(pmap, ptdpindex, lockp);
```
mhorne, line 1761: Oh by the way, I think there might be more opportunities to use these macros; see `_pmap_alloc_l3()`.

wy-chung: I have checked that functions that are calling _pmap_alloc_l3 are all using pmap_l1/l2_pindex to get the pagetable page index for _pmap_alloc_l3. On the other hand, _pmap_alloc_l3 does not need pmap_l1/l2_pindex. When it recusively calls itself, it needs to deduce the pindex for higher level pagetable page by itself.

mhorne:
This can be closed. Thanks!

PR #1361, opened 2024-08-04
P9fs fix - Handle "." in a path correctly.

bakul:
p9fs: The current logic worked only when "." was the final component but doesn't handle paths like for example "./a.out". Here p9fs_lookup() will be called twice, first for "." (cnp->cn_namelen == 1), and then "a.out" (cn_namelen == 5), but cnp->cn_nameptr is not nul terminated for the current component so much check cn_namelen.

### DIFF
--- a/sys/fs/p9fs/p9fs_vnops.c
+++ b/sys/fs/p9fs/p9fs_vnops.c
@@ -238,7 +238,7 @@ p9fs_lookup(struct vop_lookup_args *ap)
 	if (dnp == NULL)
 		return (ENOENT);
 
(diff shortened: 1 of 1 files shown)


bsdimp (approved):

markjdb:
I tested this, and it appears to fix the problem, thank you. The commit title should be prefixed `p9fs: `.

bakul:
Added prefix

markjdb:
> Added prefix

I think you didn't push the update? Please just push a branch with a single commit, rather than merging `main` into your branch.

markjdb:
This is merged now, thanks for the investigation and patch.

PR #1363, opened 2024-08-05
Add support for ARM64 FIQs

ayrtonm:
This adds support for ARM64 fast interrupt requests (FIQs) based off the [suggested approach](https://github.com/strejda/freebsd/commit/396c0b26c33bf99bf660760d68c68c65215b6b3b) in https://reviews.freebsd.org/D40161. Additionally it allows arch-specific code to define the number of interrupt roots supported with `INTR_ROOT_NUM` to avoid changing the number of roots for archs that don't need it. See the individual commit messages for details.

nits: `INTR_ROOT_` was chosen as the prefix to avoid collisions with the unrelated `intr_type` enum but we can bikeshed the name. Also setting up FIQ vectors for ARM64 platforms that don't expect to take FIQ exceptions should be fine, but I'm open to having configs opt-in to FIQ support. I initially considered this but I'm not sure it's worth the extra complexity in arm64/exception.S.

cc @kevans91

### DIFF
--- a/sys/arm/arm/exception.S
+++ b/sys/arm/arm/exception.S
@@ -48,6 +48,8 @@
 
 #include "assym.inc"
 
+#include <sys/intr.h>
+
 #include <machine/asm.h>
 #include <machine/armreg.h>
 #include <machine/asmacros.h>
@@ -308,7 +310,8 @@ ASENTRY_NP(irq_entry)
 	PUSHFRAMEINSVC			/* mode stack, build trapframe there. */
 	adr	lr, exception_exit	/* Return from handler via standard */
 	mov	r0, sp			/* exception exit routine.  Pass the */
-	b	_C_LABEL(intr_irq_handler)/* trapframe to the handler. */
+	mov	r1, #INTR_ROOT_IRQ	/* trapframe and PIC root to the handler. */
+	b	_C_LABEL(intr_irq_handler)
 END(irq_entry)
 
 /*
--- a/sys/arm64/arm64/exception.S
+++ b/sys/arm64/arm64/exception.S
@@ -26,6 +26,7 @@
  */
 
 #include <sys/elf_common.h>
+#include <sys/intr.h>
 
 #include <machine/asm.h>
 #include <machine/armreg.h>
@@ -232,6 +233,7 @@ ENTRY(handle_el1h_irq)
 	save_registers 1
 	KMSAN_ENTER
 	mov	x0, sp
+	mov	x1, #INTR_ROOT_IRQ
 	bl	intr_irq_handler
 	KMSAN_LEAVE
 	restore_registers 1
@@ -266,6 +268,7 @@ ENTRY(handle_el0_irq)
 	save_registers 0
 	KMSAN_ENTER
 	mov	x0, sp
+	mov	x1, #INTR_ROOT_IRQ
 	bl	intr_irq_handler
 	do_ast
 	KMSAN_LEAVE
--- a/sys/arm64/arm64/exception.S
+++ b/sys/arm64/arm64/exception.S
@@ -30,6 +30,7 @@
 
 #include <machine/asm.h>
 #include <machine/armreg.h>
+#include <machine/intr.h>
 #include "assym.inc"
 
 	.text
@@ -170,7 +171,7 @@
 .macro	do_ast
 	mrs	x19, daif
 	/* Make sure the IRQs are enabled before calling ast() */
-	bic	x19, x19, #PSR_I
+	bic	x19, x19, #(PSR_I | PSR_F)
 1:
 	/*
 	 * Mask interrupts while checking the ast pending flag
@@ -240,6 +241,17 @@ ENTRY(handle_el1h_irq)
 	ERET
 END(handle_el1h_irq)
 
+ENTRY(handle_el1h_fiq)
+	save_registers 1
+	KMSAN_ENTER
+	mov	x0, sp
+	mov	x1, #INTR_ROOT_FIQ
+	bl	intr_irq_handler
+	KMSAN_LEAVE
+	restore_registers 1
+	ERET
+END(handle_el1h_fiq)
+
 ENTRY(handle_el1h_serror)
 	save_registers 1
 	KMSAN_ENTER
@@ -275,6 +287,18 @@ ENTRY(handle_el0_irq)
 	restore_registers 0
 	ERET
 END(handle_el0_irq)
+
+ENTRY(handle_el0_fiq)
+	save_registers 0
+	KMSAN_ENTER
+	mov	x0, sp
+	mov	x1, #INTR_ROOT_FIQ
+	bl	intr_irq_handler
+	do_ast
+	KMSAN_LEAVE
+	restore_registers 0
+	ERET
+END(handle_el0_fiq)
 
 ENTRY(handle_el0_serror)
 	save_registers 0
@@ -318,17 +342,17 @@ exception_vectors:
 
 	vector el1h_sync 1	/* Synchronous EL1h */
 	vector el1h_irq 1	/* IRQ EL1h */
-	vempty 1		/* FIQ EL1h */
+	vector el1h_fiq 1	/* FIQ EL1h */
 	vector el1h_serror 1	/* Error EL1h */
 
 	vector el0_sync 0	/* Synchronous 64-bit EL0 */
 	vector el0_irq 0	/* IRQ 64-bit EL0 */
-	vempty 0		/* FIQ 64-bit EL0 */
+	vector el0_fiq 0	/* FIQ 64-bit EL0 */
 	vector el0_serror 0	/* Error 64-bit EL0 */
 
 	vector el0_sync 0	/* Synchronous 32-bit EL0 */
 	vector el0_irq 0	/* IRQ 32-bit EL0 */
-	vempty 0		/* FIQ 32-bit EL0 */
+	vector el0_fiq 0	/* FIQ 32-bit EL0 */
 	vector el0_serror 0	/* Error 32-bit EL0 */
 
 GNU_PROPERTY_AARCH64_FEATURE_1_NOTE(GNU_PROPERTY_AARCH64_FEATURE_1_VAL)
--- a/sys/arm64/include/intr.h
+++ b/sys/arm64/include/intr.h
@@ -27,6 +27,7 @@
 #ifndef _MACHINE_INTR_H_
 #define	_MACHINE_INTR_H_
 
+#ifndef LOCORE
 #ifdef FDT
 #include <dev/ofw/openfirm.h>
 #endif
@@ -47,5 +48,10 @@ arm_irq_memory_barrier(uintptr_t irq)
 #define	ACPI_MSI_XREF	2
 #define	ACPI_GPIO_XREF	3
 #endif
+
+#endif	/* !LOCORE */
+
+#define	INTR_ROOT_FIQ	1
+#define	INTR_ROOT_NUM	2
 
 #endif	/* _MACHINE_INTR_H */
--- a/sys/kern/subr_intr.c
+++ b/sys/kern/subr_intr.c
@@ -89,6 +89,15 @@
 
 #define	INTRNAME_LEN	(2*MAXCOMLEN + 1)
 
+/*
+ * Archs may define multiple roots with INTR_ROOT_NUM to support different kinds
+ * of interrupts (e.g. arm64 FIQs which use a different exception vector than
+ * IRQs).
+ */
+#if !defined(INTR_ROOT_NUM)
+#define	INTR_ROOT_NUM	1
+#endif
+
 #ifdef DEBUG
 #define debugf(fmt, args...) do { printf("%s(): ", __func__);	\
     printf(fmt,##args); } while (0)
@@ -99,13 +108,14 @@
 MALLOC_DECLARE(M_INTRNG);
 MALLOC_DEFINE(M_INTRNG, "intr", "intr interrupt handling");
 
-/* Main interrupt handler called from assembler -> 'hidden' for C code. */
-void intr_irq_handler(struct trapframe *tf);
-
 /* Root interrupt controller stuff. */
-device_t intr_irq_root_dev;
-static intr_irq_filter_t *irq_root_filter;
-static void *irq_root_arg;
+struct intr_irq_root {
+	device_t dev;
+	intr_irq_filter_t *filter;
+	void *arg;
+};
+
+static struct intr_irq_root intr_irq_roots[INTR_ROOT_NUM];
 
 struct intr_pic_child {
 	SLIST_ENTRY(intr_pic_child)	 pc_next;
@@ -327,12 +337,17 @@ isrc_release_counters(struct intr_irqsrc *isrc)
  *  from the assembler, where CPU interrupt is served.
  */
 void
-intr_irq_handler(struct trapframe *tf)
+intr_irq_handler(struct trapframe *tf, uint32_t rootnum)
 {
 	struct trapframe * oldframe;
 	struct thread * td;
+	struct intr_irq_root *root;
 
-	KASSERT(irq_root_filter != NULL, ("%s: no filter", __func__));
+	KASSERT(rootnum < INTR_ROOT_NUM,
+	    ("%s: invalid interrupt root %d", __func__, rootnum));
+
+	root = &intr_irq_roots[rootnum];
+	KASSERT(root->filter != NULL, ("%s: no filter", __func__));
 
 	kasan_mark(tf, sizeof(*tf), sizeof(*tf), 0);
 	kmsan_mark(tf, sizeof(*tf), KMSAN_STATE_INITED);
@@ -342,7 +357,7 @@ intr_irq_handler(struct trapframe *tf)
 	td = curthread;
 	oldframe = td->td_intr_frame;
 	td->td_intr_frame = tf;
-	irq_root_filter(irq_root_arg);
+	(root->filter)(root->arg);
 	td->td_intr_frame = oldframe;
 	critical_exit();
 #ifdef HWPMC_HOOKS
@@ -477,6 +492,14 @@ isrc_free_irq(struct intr_irqsrc *isrc)
 		irq_next_free = 0;
 
 	return (0);
+}
+
+device_t
+intr_irq_root_device(uint32_t rootnum)
+{
+	KASSERT(rootnum < INTR_ROOT_NUM,
+	    ("%s: invalid interrupt root %d", __func__, rootnum));
+	return (intr_irq_roots[rootnum].dev);
 }
 
 /*
@@ -877,9 +900,10 @@ intr_pic_deregister(device_t dev, intptr_t xref)
  */
 int
 intr_pic_claim_root(device_t dev, intptr_t xref, intr_irq_filter_t *filter,
-    void *arg)
+    void *arg, uint32_t rootnum)
 {
 	struct intr_pic *pic;
+	struct intr_irq_root *root;
 
 	pic = pic_lookup(dev, xref, FLAG_PIC);
 	if (pic == NULL) {
@@ -901,14 +925,17 @@ intr_pic_claim_root(device_t dev, intptr_t xref, intr_irq_filter_t *filter,
 	 * Note that we further suppose that there is not threaded interrupt
 	 * routine (handler) on the root. See intr_irq_handler().
 	 */
-	if (intr_irq_root_dev != NULL) {
+	KASSERT(rootnum < INTR_ROOT_NUM,
+	    ("%s: invalid interrupt root %d", __func__, rootnum));
+	root = &intr_irq_roots[rootnum];
+	if (root->dev != NULL) {
 		device_printf(dev, "another root already set\n");
 		return (EBUSY);
 	}
 
-	intr_irq_root_dev = dev;
-	irq_root_filter = filter;
-	irq_root_arg = arg;
+	root->dev = dev;
+	root->filter = filter;
+	root->arg = arg;
 
 	debugf("irq root set to %s\n", device_get_nameunit(dev));
 	return (0);
@@ -1552,14 +1579,19 @@ dosoftints(void)
 void
 intr_pic_init_secondary(void)
 {
+	device_t dev;
+	uint32_t rootnum;
 
 	/*
-	 * QQQ: Only root PIC is aware of other CPUs ???
+	 * QQQ: Only root PICs are aware of other CPUs ???
 	 */
-	KASSERT(intr_irq_root_dev != NULL, ("%s: no root attached", __func__));
-
 	//mtx_lock(&isrc_table_lock);
-	PIC_INIT_SECONDARY(intr_irq_root_dev);
+	for (rootnum = 0; rootnum < INTR_ROOT_NUM; rootnum++) {
+		dev = intr_irq_roots[rootnum].dev;
+		if (dev != NULL) {
+			PIC_INIT_SECONDARY(dev, rootnum);
+		}
+	}
 	//mtx_unlock(&isrc_table_lock);
 }
 #endif
--- a/sys/riscv/riscv/trap.c
+++ b/sys/riscv/riscv/trap.c
@@ -74,8 +74,6 @@
 #include <ddb/db_sym.h>
 #endif
 
-void intr_irq_handler(struct trapframe *tf);
-
 int (*dtrace_invop_jump_addr)(struct trapframe *);
 
 /* Called from exception.S */
@@ -319,7 +317,7 @@ do_trap_supervisor(struct trapframe *frame)
 	exception = frame->tf_scause & SCAUSE_CODE;
 	if ((frame->tf_scause & SCAUSE_INTR) != 0) {
 		/* Interrupt */
-		intr_irq_handler(frame);
+		intr_irq_handler(frame, INTR_ROOT_IRQ);
 		return;
 	}
 
@@ -400,7 +398,7 @@ do_trap_user(struct trapframe *frame)
 	exception = frame->tf_scause & SCAUSE_CODE;
 	if ((frame->tf_scause & SCAUSE_INTR) != 0) {
 		/* Interrupt */
-		intr_irq_handler(frame);
+		intr_irq_handler(frame, INTR_ROOT_IRQ);
 		return;
 	}
 	intr_enable();
--- a/sys/sys/intr.h
+++ b/sys/sys/intr.h
@@ -33,10 +33,15 @@
 #error Need INTRNG for this file
 #endif
 
+#ifndef LOCORE
 #include <sys/systm.h>
+#endif
 
 #define	INTR_IRQ_INVALID	0xFFFFFFFF
 
+#define INTR_ROOT_IRQ	0
+
+#ifndef LOCORE
 enum intr_map_data_type {
 	INTR_MAP_DATA_ACPI = 0,
 	INTR_MAP_DATA_FDT,
@@ -110,12 +115,13 @@ u_int intr_irq_next_cpu(u_int current_cpu, cpuset_t *cpumask);
 
 struct intr_pic *intr_pic_register(device_t, intptr_t);
 int intr_pic_deregister(device_t, intptr_t);
-int intr_pic_claim_root(device_t, intptr_t, intr_irq_filter_t *, void *);
+int intr_pic_claim_root(device_t, intptr_t, intr_irq_filter_t *, void *,
+    uint32_t);
 int intr_pic_add_handler(device_t, struct intr_pic *,
     intr_child_irq_filter_t *, void *, uintptr_t, uintptr_t);
 bool intr_is_per_cpu(struct resource *);
 
-extern device_t intr_irq_root_dev;
+device_t intr_irq_root_device(uint32_t);
 
 /* Intr interface for BUS. */
 
@@ -163,4 +169,8 @@ void intr_ipi_send(cpuset_t cpus, u_int ipi);
 void intr_ipi_dispatch(u_int ipi);
 #endif
 
+/* Main interrupt handler called from asm on most archs except riscv. */
+void intr_irq_handler(struct trapframe *tf, uint32_t rootnum);
+
+#endif	/* !LOCORE */
 #endif	/* _SYS_INTR_H */
--- a/sys/arm/allwinner/a10/a10_intc.c
+++ b/sys/arm/allwinner/a10/a10_intc.c
@@ -236,7 +236,7 @@ a10_intr_pic_attach(struct a10_aintc_softc *sc)
 	if (pic == NULL)
 		return (ENXIO);
 
-	return (intr_pic_claim_root(sc->sc_dev, xref, a10_intr, sc));
+	return (intr_pic_claim_root(sc->sc_dev, xref, a10_intr, sc, INTR_ROOT_IRQ));
 }
 
 static void
--- a/sys/arm/arm/gic.c
+++ b/sys/arm/arm/gic.c
@@ -200,7 +200,7 @@ gic_cpu_mask(struct arm_gic_softc *sc)
 
 #ifdef SMP
 static void
-arm_gic_init_secondary(device_t dev)
+arm_gic_init_secondary(device_t dev, uint32_t rootnum)
 {
 	struct arm_gic_softc *sc = device_get_softc(dev);
 	u_int irq, cpu;
--- a/sys/arm/arm/gic_acpi.c
+++ b/sys/arm/arm/gic_acpi.c
@@ -227,7 +227,7 @@ gic_acpi_attach(device_t dev)
 	/*
 	 * Controller is root:
 	 */
-	if (intr_pic_claim_root(dev, xref, arm_gic_intr, sc) != 0) {
+	if (intr_pic_claim_root(dev, xref, arm_gic_intr, sc, INTR_ROOT_IRQ) != 0) {
 		device_printf(dev, "could not set PIC as a root\n");
 		intr_pic_deregister(dev, xref);
 		goto cleanup;
--- a/sys/arm/arm/gic_fdt.c
+++ b/sys/arm/arm/gic_fdt.c
@@ -153,7 +153,8 @@ gic_fdt_attach(device_t dev)
 	 */
 	pxref = ofw_bus_find_iparent(ofw_bus_get_node(dev));
 	if (pxref == 0 || xref == pxref) {
-		if (intr_pic_claim_root(dev, xref, arm_gic_intr, sc) != 0) {
+		if (intr_pic_claim_root(dev, xref, arm_gic_intr, sc, INTR_ROOT_IRQ)
+		    != 0) {
 			device_printf(dev, "could not set PIC as a root\n");
 			intr_pic_deregister(dev, xref);
 			goto cleanup;
--- a/sys/arm/broadcom/bcm2835/bcm2835_intr.c
+++ b/sys/arm/broadcom/bcm2835/bcm2835_intr.c
@@ -405,7 +405,8 @@ bcm_intc_attach(device_t dev)
 	sc->intc_irq_res = bus_alloc_resource_any(dev, SYS_RES_IRQ, &rid,
 	    RF_ACTIVE);
 	if (sc->intc_irq_res == NULL) {
-		if (intr_pic_claim_root(dev, xref, bcm2835_intc_intr, sc) != 0) {
+		if (intr_pic_claim_root(dev, xref, bcm2835_intc_intr, sc, INTR_ROOT_IRQ)
+		    != 0) {
 			/* XXX clean up */
 			device_printf(dev, "could not set PIC as a root\n");
 			return (ENXIO);
--- a/sys/arm/broadcom/bcm2835/bcm2836.c
+++ b/sys/arm/broadcom/bcm2835/bcm2836.c
@@ -538,7 +538,7 @@ bcm_lintc_init_pmu_on_ap(struct bcm_lintc_softc *sc, u_int cpu)
 }
 
 static void
-bcm_lintc_init_secondary(device_t dev)
+bcm_lintc_init_secondary(device_t dev, uint32_t rootnum)
 {
 	u_int cpu;
 	struct bcm_lintc_softc *sc;
@@ -646,7 +646,8 @@ bcm_lintc_pic_attach(struct bcm_lintc_softc *sc)
 	if (pic == NULL)
 		return (ENXIO);
 
-	error = intr_pic_claim_root(sc->bls_dev, xref, bcm_lintc_intr, sc);
+	error = intr_pic_claim_root(sc->bls_dev, xref, bcm_lintc_intr, sc,
+	    INTR_ROOT_IRQ);
 	if (error != 0)
 		return (error);
 
--- a/sys/arm/ti/aintc.c
+++ b/sys/arm/ti/aintc.c
@@ -230,7 +230,8 @@ ti_aintc_pic_attach(struct ti_aintc_softc *sc)
 	if (pic == NULL)
 		return (ENXIO);
 
-	return (intr_pic_claim_root(sc->sc_dev, xref, ti_aintc_intr, sc));
+	return (intr_pic_claim_root(sc->sc_dev, xref, ti_aintc_intr, sc,
+	    INTR_ROOT_IRQ));
 }
 
 static int
--- a/sys/arm64/arm64/gic_v3.c
+++ b/sys/arm64/arm64/gic_v3.c
@@ -1093,7 +1093,7 @@ gic_v3_bind_intr(device_t dev, struct intr_irqsrc *isrc)
 
 #ifdef SMP
 static void
-gic_v3_init_secondary(device_t dev)
+gic_v3_init_secondary(device_t dev, uint32_t rootnum)
 {
 	struct gic_v3_setup_periph_args pargs;
 	device_t child;
@@ -1140,7 +1140,7 @@ gic_v3_init_secondary(device_t dev)
 
 	for (i = 0; i < sc->gic_nchildren; i++) {
 		child = sc->gic_children[i];
-		PIC_INIT_SECONDARY(child);
+		PIC_INIT_SECONDARY(child, rootnum);
 	}
 }
 
--- a/sys/arm64/arm64/gic_v3_acpi.c
+++ b/sys/arm64/arm64/gic_v3_acpi.c
@@ -345,8 +345,9 @@ gic_v3_acpi_attach(device_t dev)
 		}
 	}
 
-	if (intr_pic_claim_root(dev, ACPI_INTR_XREF, arm_gic_v3_intr, sc)
-	    != 0) {
+	err = intr_pic_claim_root(dev, ACPI_INTR_XREF, arm_gic_v3_intr, sc,
+	    INTR_ROOT_IRQ);
+	if (err != 0) {
 		err = ENXIO;
 		goto error;
 	}
--- a/sys/arm64/arm64/gic_v3_fdt.c
+++ b/sys/arm64/arm64/gic_v3_fdt.c
@@ -161,7 +161,8 @@ gic_v3_fdt_attach(device_t dev)
 	/* Register xref */
 	OF_device_register_xref(xref, dev);
 
-	if (intr_pic_claim_root(dev, xref, arm_gic_v3_intr, sc) != 0) {
+	err = intr_pic_claim_root(dev, xref, arm_gic_v3_intr, sc, INTR_ROOT_IRQ);
+	if (err != 0) {
 		err = ENXIO;
 		goto error;
 	}
--- a/sys/arm64/arm64/gicv3_its.c
+++ b/sys/arm64/arm64/gicv3_its.c
@@ -1284,7 +1284,7 @@ gicv3_its_setup_intr(device_t dev, struct intr_irqsrc *isrc,
 
 #ifdef SMP
 static void
-gicv3_its_init_secondary(device_t dev)
+gicv3_its_init_secondary(device_t dev, uint32_t rootnum)
 {
 	struct gicv3_its_softc *sc;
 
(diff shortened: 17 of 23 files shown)


kevans91:
I can't seem to tag @strejda or @jrtc27 for review, but CC'ing them as well

ehem:
@kevans91, you're not in the GitHub "FreeBSD organization"; nor are @jrtc27 nor @strejda.  That is why it isn't allowing you to select people as reviewers nor approving workflows.

Likely the membership of the GitHub organization hasn't been getting updated very frequently.  Given how GitHub seems to want to become a social network I'm not at all surprised how many developers have been reluctant to join.

jrtc27:
You seem to have a snarky explanation for everything. It is not at all like that; there just isn't currently an easy official way to be added to the organisation beyond figuring out who has admin rights and asking them. No conspiracy theories needed.

jrtc27:
I'll also remind you once again that an important part of being a contributor to an open-source project is knowing how to get on with people, in particular those in the project you want to be involved with. Being a know-it-all who likes to throw around conspiracies is generally not conducive to that.

ehem:
@jrtc27 please point to a snarky explanation.  I simply see reasonable cause why many people would be reluctant to create an account on GitHub.  Then since there isn't a formal process, as you've just stated, @kevans91 never got added.

Information on who has admin in the organization not being well known is more than I knew.  That in itself seems rather concerning.

jrtc27:
> @jrtc27 please point to a snarky explanation. I simply see reasonable cause why many people would be reluctant to create an account on GitHub. Then since there isn't a formal process, as you've just stated, @kevans91 never got added.

"Given how GitHub seems to want to become a social network I'm not at all surprised how many developers have been reluctant to join."

This is conspiracy theory snark and you know it.

> Information on who has admin in the organization not being well known is more than I knew. That in itself seems rather concerning.

It's really not. I did once know, and I have zero concern about the set of people being trustworthy. Historically it was not important to document, because it was just for read-only mirrors. Now we are accepting pull requests it is more important, but things still work fine despite it. Asking in private developer channels on IRC would quickly yield that information, I've just never been bothered to add myself. It is as simple as that. I suspect the same is true of others.

bsdimp:
Only about half or fewer developers have been added. It's easy to know once you know the right bugzilla categories to use. But it's a mild hassle for an unclear benefit.

Plus we're just learning the rope on GH pull requests, so we'll likely do a lot that's suboptimal in the mean time.  This is one detail we've not pushed. For people aching to do it, the docs are in the handbook. But if not, I've not pushed people to add themselves since the only tangible benefit is i can assign issues and reviews. But since that ptocess is not well defined today, and I look at the big picture and @ the right people at the right times, it's been OK enough as the status quo.

sys/kern/subr_intr.c
```diff
@@ -89,6 +89,16 @@
...
 
+/*
+ * Archs may define multiple roots with INTR_ROOT_NUM to support different kinds
+ * of interrupts (e.g. arm64 FIQs which use a different exception vector than
+ * IRQs).
+ */
+#if !defined(INTR_ROOT_NUM)
+#define INTR_ROOT_IRQ	0
+#define INTR_ROOT_NUM	1
+#endif
```
ehem, line 100: Is the number of root PICs being constant a good choice?  Perhaps this should instead be dynamic?

sys/kern/subr_intr.c
```diff
@@ -876,8 +900,8 @@ intr_pic_deregister(device_t dev, intptr_t xref)
  *     an interrupts property and thus no explicit interrupt parent."
  */
 int
-intr_pic_claim_root(device_t dev, intptr_t xref, intr_irq_filter_t *filter,
-    void *arg)
+intr_pic_claim_root_num(device_t dev, intptr_t xref, intr_irq_filter_t *filter,
+    void *arg, uint32_t root)
```
ehem, line 904: Perhaps there should be a reserved value for `root` which specifies a particular root controller is dynamic and needs `intr_pic_claim_root_num()` to return its index?

sys/kern/subr_intr.c
```diff
@@ -1556,10 +1589,10 @@ intr_pic_init_secondary(void)
 	/*
 	 * QQQ: Only root PIC is aware of other CPUs ???
 	 */
-	KASSERT(intr_irq_root_dev != NULL, ("%s: no root attached", __func__));
+	KASSERT(intr_irq_root_dev() != NULL, ("%s: no root attached", __func__));
 
 	//mtx_lock(&isrc_table_lock);
-	PIC_INIT_SECONDARY(intr_irq_root_dev);
+	PIC_INIT_SECONDARY(intr_irq_root_dev());
 	//mtx_unlock(&isrc_table_lock);
```
ehem, line 1596: I'm rather concerned about this choice, but the case I'm thinking of may be unusual and not apply here.  I'm unfamiliar with the design of this new Apple hardware and I don't know how this is expected to map to future devices.  I suspect this should instead be looping over all the root PICs and calling the function on each of them in turn?

ehem (commented):
Took quite a while to consider some of the implications of this.  Is the count of root PICs being constant appropriate?

The PIC I'm looking at isn't tied to hardware and does act as a root PIC in some ways (#1358).  If the count of root PICs was dynamic it could then register as a root PIC and hook into `pic_init_secondary` and thus eliminate the need for #1280.

This could also allow for a kernel which ran both on this new Apple silicon and devices which had only a single root.

sys/kern/subr_intr.c
ayrtonm: Yeah you have a good point, this happened to work for my use case but it's probably not what we want to do. I'm not sure if just looping over the root PICs is the correct approach either though. This PR allows one device to claim multiple PIC roots and so if we're going to call `pic_init_secondary` multiple times with the same `device_t` it should at least know which root it's for.

ayrtonm:
> Is the count of root PICs being constant appropriate?

The case in your linked PRs does seem a bit unusual like you mentioned, e.g. `xen_setup_vcpu_info` gets cast to `pic_secondary_init_t` ignoring the `device_t` arg so it seems like it's just using that as a convenient way to setup a hook to those calls. I've only taken a brief look at your PRs so far so I don't have the full context though.

For reference on Apple silicon there's [one devicetree node](https://github.com/AsahiLinux/u-boot/blob/asahi/arch/arm/dts/t8112.dtsi#L914) so we have one device that claims two root PICs (for IRQs and FIQs). I think a fixed number of root PICs is also reasonable for riscv use cases mentioned in [D40161](https://reviews.freebsd.org/D40161), but I'm not super familiar with those.

ehem:
You've got it right.  Issue is having the root count dynamic could both allow sharing one kernel between multiple devices and cover this need too.  Feel free to take no action on this.

sys/kern/subr_intr.c
ayrtonm: I added a loop over root PIC devices and modified `pic_init_secondary` to take a `uint32_t root` argument as well. I also replaced the assert for root dev != NULL with just a NULL check, but it might be good to make sure that `pic_init_secondary` is called for at least one device.

@jrtc27 @strejda I pushed the change as a new commit assuming that makes the diff easier to review and plan on squashing the related commits before this is ready to land. Alternatively I'm happy to combine the commits now if you prefer, whatever works for you.

sys/kern/subr_intr.c
```diff
@@ -100,12 +110,13 @@ MALLOC_DECLARE(M_INTRNG);
 MALLOC_DEFINE(M_INTRNG, "intr", "intr interrupt handling");
 
 /* Main interrupt handler called from assembler -> 'hidden' for C code. */
+void intr_irq_handler_root(struct trapframe *tf, uint32_t root);
 void intr_irq_handler(struct trapframe *tf);
```
ehem, line 114: Ideally I'd actually like to get rid of `intr_irq_handler()` _now_.

ehem (commented):
I'm still concerned about the static allocation.  Yet insisting upon dynamic allocation is likely unreasonable for this pull request.  A quick test suggests this could function to supplant #1280.

sys/kern/subr_intr.c
```diff
@@ -479,6 +498,11 @@ isrc_free_irq(struct intr_irqsrc *isrc)
 	return (0);
 }
 
+device_t intr_irq_root_dev(void)
```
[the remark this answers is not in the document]
ayrtonm: done

sys/kern/subr_intr.c
```diff
@@ -1553,13 +1586,18 @@ void
...
-
+	uint32_t root;
+	device_t dev;
 	//mtx_lock(&isrc_table_lock);
-	PIC_INIT_SECONDARY(intr_irq_root_dev);
+	for (root = 0; root < INTR_ROOT_NUM; root++) {
+		dev = intr_irq_root_devs[root];
+		if (dev) {
+			/*
+			 * QQQ: Only root PIC is aware of other CPUs ???
```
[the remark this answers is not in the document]
ayrtonm: done

sys/kern/subr_intr.c
ayrtonm: Just to clarify you mean replace all calls to `intr_irq_handler` with `intr_irq_handler_root(INTR_ROOT_IRQ)`? I kept the former to minimize the changes needed for archs that don't use multiple roots yet but I don't have a strong opinion on this so I can do either.

sys/kern/subr_intr.c
```diff
@@ -100,12 +110,13 @@ MALLOC_DECLARE(M_INTRNG);
...
+void intr_irq_handler_root(struct trapframe *tf, uint32_t root);
 void intr_irq_handler(struct trapframe *tf);
 
 /* Root interrupt controller stuff. */
-device_t intr_irq_root_dev;
-static intr_irq_filter_t *irq_root_filter;
-static void *irq_root_arg;
+static device_t intr_irq_root_devs[INTR_ROOT_NUM];
+static intr_irq_filter_t *irq_root_filter[INTR_ROOT_NUM];
+static void *irq_root_arg[INTR_ROOT_NUM];
```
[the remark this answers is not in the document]
ayrtonm: done

ayrtonm:
> I'm still concerned about the static allocation. Yet insisting upon dynamic allocation is likely unreasonable for this pull request

Yeah this handles the cases mentioned in [D40161](https://reviews.freebsd.org/D40161) which multiple people were on board with and simply stalled because of lack of bandwidth from the original author, but I'll let others who are more familiar with intrng chime in on whether the number of PIC roots should eventually be dynamic.

sys/kern/subr_intr.c
ehem: Note, I am **not** a member of the organization so **do not** consider my opinions the last word.

My concern is compatibility shims tend to hang around a **long** time in FreeBSD.  There are likely plenty of decade-old fixup shims around.  This seems one which might be best to rip the band-aid off now, rather than hanging around for years.

sys/kern/subr_intr.c
```diff
@@ -327,12 +342,17 @@ isrc_release_counters(struct intr_irqsrc *isrc)
...
-intr_irq_handler(struct trapframe *tf)
+intr_irq_handler_root(struct trapframe *tf, uint32_t rootnum)
 {
 	struct trapframe * oldframe;
 	struct thread * td;
+	struct intr_irq_root *root;
 
-	KASSERT(irq_root_filter != NULL, ("%s: no filter", __func__));
+	KASSERT(rootnum < INTR_ROOT_NUM,
+		("%s: invalid interrupt root %d", __func__, rootnum));
```
ehem, line 352: FreeBSD style says this should be "\<tab\>\<space\>\<space\>\<space\>\<space\>".

sys/kern/subr_intr.c
```diff
@@ -352,6 +372,12 @@ intr_irq_handler(struct trapframe *tf)
 #endif
 }
 
+void
+intr_irq_handler(struct trapframe *tf)
+{
+	intr_irq_handler_root(tf, INTR_ROOT_IRQ);
```
ehem, line 378: I believe FreeBSD style is there should be a blank line between these two.

sys/kern/subr_intr.c
```diff
@@ -479,6 +505,12 @@ isrc_free_irq(struct intr_irqsrc *isrc)
 	return (0);
 }
 
+device_t
+intr_irq_root_dev(void)
+{
+	return intr_irq_roots[INTR_ROOT_IRQ].dev;
```
ehem, line 511: Same as above.

sys/kern/subr_intr.c
```diff
@@ -901,19 +934,29 @@ intr_pic_claim_root(device_t dev, intptr_t xref, intr_irq_filter_t *filter,
 	 * Note that we further suppose that there is not threaded interrupt
 	 * routine (handler) on the root. See intr_irq_handler().
 	 */
-	if (intr_irq_root_dev != NULL) {
+	KASSERT(rootnum < INTR_ROOT_NUM,
+		("%s: invalid interrupt root %d", __func__, rootnum));
```
ehem, line 938: I believe style is this should be tab, then 4 spaces.

sys/kern/subr_intr.c
```diff
@@ -901,19 +934,29 @@ intr_pic_claim_root(device_t dev, intptr_t xref, intr_irq_filter_t *filter,
...
 
 	debugf("irq root set to %s\n", device_get_nameunit(dev));
 	return (0);
 }
 
+int
+intr_pic_claim_root(device_t dev, intptr_t xref, intr_irq_filter_t *filter,
+    void *arg)
+{
+	return intr_pic_claim_root_num(dev, xref, filter, arg, INTR_ROOT_IRQ);
```
ehem, line 957: Another blank line.

sys/kern/subr_intr.c
```diff
@@ -1552,14 +1595,18 @@ dosoftints(void)
...
 {
-
 	/*
-	 * QQQ: Only root PIC is aware of other CPUs ???
+	 * QQQ: Only root PICs are aware of other CPUs ???
 	 */
-	KASSERT(intr_irq_root_dev != NULL, ("%s: no root attached", __func__));
-
+	uint32_t rootnum;
+	device_t dev;
```
ehem, line 1602: I would to have these declarations above the comment.  Then FreeBSD style would be blank line between the two.

sys/kern/subr_intr.c
```diff
@@ -1552,14 +1595,18 @@ dosoftints(void)
...
 	//mtx_lock(&isrc_table_lock);
-	PIC_INIT_SECONDARY(intr_irq_root_dev);
+	for (rootnum = 0; rootnum < INTR_ROOT_NUM; rootnum++) {
+		dev = intr_irq_roots[rootnum].dev;
+		if (dev) {
+			PIC_INIT_SECONDARY(dev, rootnum);
+		}
```
ehem, line 1608: I believe FreeBSD style allows omitting the braces.  Style also needs `dev != NULL`.

sys/kern/subr_intr.c
```diff
@@ -327,12 +342,17 @@ isrc_release_counters(struct intr_irqsrc *isrc)
  *  from the assembler, where CPU interrupt is served.
  */
 void
-intr_irq_handler(struct trapframe *tf)
+intr_irq_handler_root(struct trapframe *tf, uint32_t rootnum)
```
ehem, line 345: I concur with changing to "rootnum".

sys/kern/subr_intr.c
ayrtonm: Sorry I should've added a new comment rather than edit my last one. I don't fully consider `intr_irq_handler` a compatibility shim since I expect some archs won't ever need the new version (e.g. ARM32 where the existing FIQ support was recently removed). That's why I'm waiting to see if anyone has opinions on this instead of just making the change you suggested.

sys/arm64/arm64/exception.S
```diff
@@ -315,17 +339,17 @@ exception_vectors:
...
 	vector el0_sync 0	/* Synchronous 64-bit EL0 */
 	vector el0_irq 0	/* IRQ 64-bit EL0 */
-	vempty 0		/* FIQ 64-bit EL0 */
+	vector el0_fiq 0		/* FIQ 64-bit EL0 */
 	vector el0_serror 0	/* Error 64-bit EL0 */
 
 	vector el0_sync 0	/* Synchronous 32-bit EL0 */
 	vector el0_irq 0	/* IRQ 32-bit EL0 */
-	vempty 0		/* FIQ 32-bit EL0 */
+	vector el0_fiq 0		/* FIQ 32-bit EL0 */
```
ehem, line 352: FreeBSD assumes tabs are 8 spaces, so these need single tab removals to remain lined up.

ehem (commented):
Again, note I am a _community reviewer_ my reviews are my own and may not reflect the views of people with commit bits.  Seems I'm onto bringing up the style issues since there isn't much else to comment on.

Since things are at this point, it may be time to start squashing things so they're easier to track.  My preference would be to keep 2065408796005ba38c14b5df93aa32c85175974a as a separate commit.  The rest I would tend to squash.

sys/kern/subr_intr.c
ehem: (blank line between the declarations and the comment)

sys/kern/subr_intr.c
jrtc27: Not since 2019

sys/kern/subr_intr.c
jrtc27: As above, not since 2019, but return x should always have x parenthesised.

ayrtonm:
> My preference would be to keep https://github.com/freebsd/freebsd-src/commit/2065408796005ba38c14b5df93aa32c85175974a as a separate commit. The rest I would tend to squash.

Yeah that was the plan. The ones without a commit message would get squashed into the first, but I wanted to wait until others tagged in the PR got a chance to take a look and let them decide whether they want me to squash it or prefer to do it themselves.

sys/kern/subr_intr.c
ayrtonm: oh thanks for pointing this out! I somehow missed the return value convention when going through the style guide

sys/kern/subr_intr.c
ayrtonm: done

sys/kern/subr_intr.c
ayrtonm: ok left it as-is

sys/kern/subr_intr.c
ayrtonm: fixed

sys/kern/subr_intr.c
ayrtonm: done

sys/kern/subr_intr.c
ayrtonm: done

sys/arm64/arm64/exception.S
ayrtonm: done (in a separate commit from the other style fixes to simplify squashing commits later)

sys/kern/subr_intr.c
ayrtonm: I fixed the `NULL` check. Since the style guide doesn't _require_ omitting the braces I'd prefer to leave them in out of an abundance of caution for when someone needs to change this.

strejda (commented):
I'm almost satisfied with that. But I have a few small nits :)
- move #define INTR_ROOT_IRQ 0 to intr.h, with the appropriate change in subr_intr.c:98. The IRQ interrupt type is mandatory, and I think it should always come first, because of the order loop in intr_pic_init_secondary().
- intr_irq_root_dev() should take rootnum as argument
- I'm not sure if intr_pic_claim_root() , intr_irq_handler() is a useful shortcut or a work complicating workaround. Personally I don't like it, but I can live with it.

With the above change (mainly the first point), this seems like a very good starting commit for further polishing if necessary.

ehem:
> * move #define INTR_ROOT_IRQ 0 to intr.h, with the appropriate change in subr_intr.c:98. The IRQ interrupt type is mandatory, and I think it should always come first, because of the order loop in intr_pic_init_secondary().

Since there is a missing "*", to the second:

I **strongly** disagree.  The PIC interface is consistent in always passing the device first (which then yields the `softc`).  I'm highly adverse to that inconsistency.  The device may also implicitly encode the type since there could be different devices for the different types (I'm unsure whether the Apple approach will be common or uncommon).  As such the device is **far** more important and should be first.

> * I'm not sure if intr_pic_claim_root() , intr_irq_handler() is a useful shortcut or a work complicating workaround. Personally I don't like it, but I can live with it.

In [different terms](https://github.com/freebsd/freebsd-src/pull/1363#discussion_r1708059595).  I doubt it will complicate things, but I also doubt retaining it will help much.

ayrtonm:
> I'm not sure if intr_pic_claim_root() , intr_irq_handler() is a useful shortcut or a work complicating workaround

Yeah I'll remove those. Part of my rationale for keeping them was that we didn't need to expose archs that weren't expected to change to the concept of PIC roots, but like I mentioned in another comment I don't have a strong opinion on this.

> The PIC interface is consistent in always passing the device first...

I think you misunderstood. "[the IRQ interrupt type] should always come first" refers to the values of the `INTR_ROOT_*` macros (`INTR_ROOT_IRQ` should have the lowest value), not the order of arguments. Regardless I think `device_t` in the PIC interface has to come first for the kobj method lookup to work correctly anyway.

ehem:
One good mis-parse deserves another.  Okay, that was strictly a request to have the `INTR_ROOT_*` values in `sys/sys/intr.h`.  Originally I had thought the same thing, but now I'm less sure.

Come to think of it though, shouldn't the types really be an `enum` instead of a `uint32_t`?  I agree `intr_irq_handler_root()` is best using a `uint32_t` (to interface with assembly language), yet several other places `enum` seems better.  Notably this allows a `INTR_ROOT_END` which is implicitly set simply by being at the end.  Trick is having it both be an enum and managing the interface to assembly language.

One other administrative issue, @kevans91 I still see traces of what [D40161](https://reviews.freebsd.org/D40161) looked like.  I suspect there should be a `Original idea: kevans` footer on the first commit.  I also suggest this is effectively commandeering D40161 and perhaps a `Differential Revision: https://reviews.freebsd.org/D40161` footer would also be appropriate?

sys/arm64/include/intr.h
```diff
@@ -27,6 +27,7 @@
 #ifndef _MACHINE_INTR_H_
 #define	_MACHINE_INTR_H_
 
+#ifndef LOCORE
 #ifdef FDT
 #include <dev/ofw/openfirm.h>
 #endif
```
ehem, line 33:
```suggestion
#endif

typedef enum {
	INTR_ROOT_IRQ =	_INTR_ROOT_IRQ,
	INTR_ROOT_FIQ =	_INTR_ROOT_FIQ,
	INTR_ROOT_END
} intr_root_t;
#define INTR_ROOT_COUNT	INTR_ROOT_END
```
This allows the C constructs to handle this as an enumerated value (which it amounts to), while allowing the assembly language portion to still have its values.

ehem (commented):
Upon further consideration, I think I was near the right track, but didn't quite get it right.  Instead of `uint32_t` everything should be using a `intr_root_t` which the architecture can override.

Since this is getting out of hand, I've got this visible at 
[here](https://github.com/ehem/freebsd-src/tree/pull1363).

At this point I would tend to squash as there are too many adjustment commits and this is getting confusing.

ehem:
One more issue.  Perhaps the type value should be passed to the root filter?  If regular and fast interrupts are different PICs, the interface works well.  If they're a shared device, they might use the same argument value and modify handling based on the type.

ayrtonm:
> move #define INTR_ROOT_IRQ 0 to intr.h

done and I had to add `#ifndef LOCORE` to `sys/intr.h` to allow including it in exception.S for arm and arm64 since `intr_irq_handler` takes the rootnum as an argument.

> intr_irq_root_dev() should take rootnum as argument

done and fixed up the callsites

> I'm not sure if intr_pic_claim_root() , intr_irq_handler() is a useful shortcut or a work complicating workaround

I removed these workarounds so there's only one version of each without the new `_num` suffix and with a `uint32_t rootnum` argument. Also riscv calls `intr_irq_handler` from C and previously declared it in riscv/trap.c so I removed that declaration and moved the declaration in kern/subr_intr.c to sys/intr.h to prevent them from getting out of sync.

Since this change is a bit bigger I checked that arm, arm64 and riscv build after each commit. Also @strejda I went ahead and squashed the commits but I still have a local branch with the commits split up in case you find it easier to review the changes chronologically.

sys/arm64/include/intr.h
ayrtonm: I considered it but I'm not a fan of this approach. The problem is that you're setting the enum values and also depending on `INTR_ROOT_END` having the correct value simply because it's at the end. For more complicated cases this seems error-prone.

sys/kern/subr_intr.c
```diff
@@ -100,12 +110,17 @@ MALLOC_DECLARE(M_INTRNG);
 MALLOC_DEFINE(M_INTRNG, "intr", "intr interrupt handling");
 
 /* Main interrupt handler called from assembler -> 'hidden' for C code. */
+void intr_irq_handler_root(struct trapframe *tf, uint32_t rootnum);
```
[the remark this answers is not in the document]
ayrtonm: `style` just says all functions are prototyped somewhere. I don't see the point in adding `extern` before the function since they default to external linkage anyway, but regardless this is irrelevant since I had to move the declaration to sys/intr.h.

ayrtonm:
> Perhaps the type value should be passed to the root filter?

There is a different filter for each root so the "type value" (I assume this refers to the root number) can be known statically. Passing the root number to the filter is only necessary when you have a dynamic number of root PICs (assuming you have a fixed number of functions that can be assigned as the filter) which this PR is not adding.

> If regular and fast interrupts are different PICs, the interface works well. If they're a shared device, they might use the same argument value and modify handling based on the type.

This comment seems to be mixing up PICs and devices. Like I mentioned earlier, the apple silicon case has one "shared" device which claims two root PICs. In each call to `intr_pic_claim_root` we pass a different function as the filter so we don't need to pass pass the root number to the filters themselves.

ehem:
> There is a different filter for each root so the "type value" (I assume this refers to the root number) can be known statically. Passing the root number to the filter is only necessary when you have a dynamic number of root PICs (assuming you have a fixed number of functions that can be assigned as the filter) which this PR is not adding.

> This comment seems to be mixing up PICs and devices. Like I mentioned earlier, the apple silicon case has one "shared" device which claims two root PICs. In each call to `intr_pic_claim_root` we pass a different function as the filter so we don't need to pass pass the root number to the filters themselves.

I'm not mixing up anything.  I was making a suggestion about the interrupt interface.  Mainly pass what you're calling `rootnum` all the way to the root function.  _Your_ proposal does not require this, but I thought it might be worthwhile for some other device which wanted this interface.

Yet predicting future needs is rather imprecise and future needs will likely be wildly different from what I'm suggesting.  This was an idea, feel free to ignore it.

ehem (changes requested):
I'm highly adverse to having `INTR_ROOT_IRQ` in `sys/sys/intr.h`.  That is very much architecture-specific and wrong to have in `sys/sys/intr.h` (which is semi-architecture-independent).  Yes, it does mean the value is in one place, but it may fool someone into thinking it has a cross-architecture significance which it does not.

This is the time to consider what _exactly_ the interface should look like.  Right now what you have is:
```
#if !defined(INTR_ROOT_NUM)
#define	INTR_ROOT_NUM	1
#endif
```
```
void intr_irq_handler(struct trapframe *tf, uint32_t rootnum);
```
First issue, `INTR_ROOT_NUM` is a _count_ of maximum number of root PICs.  In light of that `INTR_ROOT_COUNT` is a rather better name.
Second issue is `uint32_t` isn't a bad choice, but does have some limitations.  Notably for the C portions an `enum` makes rather a lot of sense.  `register_t` or `u_register_t` also make sense as a full register-sized value could have use.

As such I would like to point at two branches, [full](https://github.com/ehem/freebsd-src/tree/pull1363) and [enum](https://github.com/ehem/freebsd-src/tree/pull1363-0).  The main suggested interface being:
```
#if !defined(INTR_ROOT_COUNT)
#define	INTR_ROOT_COUNT	1
typedef	register_t	intr_root_t;
#endif
```

Result is an architecture can substitute most any variable type (notably an `enum`).  I'm fine with simply using `#define` in your pull request, but allowing for them is worthwhile future support (since if you don't use an `enum`, such will be included with _my_ work).

Two tricks from these.  First, `intr_root_t` becomes part of several declarations inside `sys/sys/intr.h`, as a result these become incomplete without `machine/intr.h`.  As a result I had to look into the `#include` situation.  Since `#include <machine/intr_machdep.h` or `#include <machine/intr.h>` have been the way to get the architecture interrupt header for a long time,  There is an extra commit to resolve the few spots which grabbed `sys/intr.h` instead of `machine/intr.h`.

Second, since an architecture _could_ use a signed type for `intr_root_t` this needs to be accommodated.  Promoting to the longest unsigned integer is simplest approach for handling this situation.

sys/riscv/riscv/trap.c
```diff
@@ -319,7 +317,7 @@ do_trap_supervisor(struct trapframe *frame)
 	exception = frame->tf_scause & SCAUSE_CODE;
 	if ((frame->tf_scause & SCAUSE_INTR) != 0) {
 		/* Interrupt */
-		intr_irq_handler(frame);
+		intr_irq_handler(frame, INTR_ROOT_IRQ);
```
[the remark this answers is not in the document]
ayrtonm: RISC-V is not calling a wrapper function, this is the exact same `intr_irq_handler` that ARM{,64} call and equivalent code is generated on these platforms here. The only difference is that one source file is asm while the other is C.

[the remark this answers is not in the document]
jrtc27: I've always assumed it's because Arm's VBAR does the interrupt-vs-exception dispatch in hardware thanks to having multiple trap vectors, whereas RISC-V has just the one in direct mode which requires demuxing in software, and it's best to keep the assembly minimal, so it's done in C.

sys/sys/intr.h
```diff
@@ -33,10 +33,15 @@
...
 
+#ifndef LOCORE
 #include <sys/systm.h>
+#endif
 
 #define	INTR_IRQ_INVALID	0xFFFFFFFF
 
+#define INTR_ROOT_IRQ	0
+
+#ifndef LOCORE
```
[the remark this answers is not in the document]
ehem: No.  While the name isn't my favorite, my objection is simply this shouldn't be here as this is an architecture detail which INTRNG has no business interacting with.

sys/kern/subr_intr.c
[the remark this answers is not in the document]
ehem: I had missed this was due to RISC-V calling `intr_irq_handler()` from C.  That invalidates my concern.

sys/arm64/include/intr.h
ehem: Hrmm, seems I've discovered unexpected behavior in how GitHub handles comments...

As such replacing the mention.  This is no more error-prone than demanding list of functions end with `{0, 0}` or `DEVMETHOD_END`.  Yes, it is possible to goof this, but it is quite easy to get it right.  This isn't severe enough to worry.  Take a look at my proposed examples.

zxombie:
How much do we care about the KPI of these, e.g. to merge to 14? It looks doable, although the `intr_irq_root_dev` function will need to be renamed as it conflicts with the variable.

ayrtonm:
> How much do we care about the KPI of these, e.g. to merge to 14? It looks doable, although the intr_irq_root_dev function will need to be renamed as it conflicts with the variable.

The function signature of `intr_pic_claim_root` also changed so would that also need to be renamed to backport this to 14?

jrtc27:
> > How much do we care about the KPI of these, e.g. to merge to 14? It looks doable, although the intr_irq_root_dev function will need to be renamed as it conflicts with the variable.
> 
> The function signature of `intr_pic_claim_root` also changed so would that also need to be renamed to backport this to 14?

No. https://github.com/freebsd/freebsd-src/commit/e06afdb2856944d6db4e33d4bbe0e2dbbd057eb4 changes that and is overdue being MFC'ed.

ayrtonm:
I pushed it as a new commit if anyone wants to take a look (sorry about the workflow, I'll use phabricator in the future). @zxombie I'm not sure if you meant rename the function and keep the variable decl, but I assumed it should be removed.

I'm also not totally clear on what the requirements for being able to MFC or what the KPI stability requirements are, but I tested a `git cherry-pick` of these commits onto `stable/14`. The intrng commit required a tiny fixup due to a header added in https://github.com/freebsd/freebsd-src/commit/c2e0d56f5e49. The FIQ commit was doable, but more involved because https://github.com/freebsd/freebsd-src/commit/1b60bbfe653f split up one of the exception vector entries.

sys/arm/arm/exception.S
```diff
@@ -48,6 +48,8 @@
 
 #include "assym.inc"
 
+#include <sys/intr.h>
```
ehem, line 51: The only place which should `#include <sys/intr.h>` are the `machine/intr.h` headers.  Everywhere else should pick up `sys/intr.h` by `#include <machine/intr.h>`.

sys/arm64/arm64/exception.S
```diff
@@ -26,6 +26,7 @@
  */
 
 #include <sys/elf_common.h>
+#include <sys/intr.h>
```
ehem, line 29: Identical to above.  The only place which should #include <sys/intr.h> are the machine/intr.h headers. Everywhere else should pick up sys/intr.h by #include <machine/intr.h>.

sys/sys/intr.h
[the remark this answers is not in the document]
ehem: Two instincts are conflicting here.  I readily understand using a `#define` for a constant which gets special treatment in a number of places and needs to be synchronized.  At the same time I find creating yet another `#define` for the well-known constant `0` a bit odd.

sys/arm64/arm64/exception.S
```diff
@@ -30,6 +30,7 @@
```
ehem, line 30: No matter which direction things go, you absolutely need to remove the `#include <sys/intr.h>` in the second commit.

sys/arm64/include/intr.h
```diff
@@ -42,10 +43,16 @@ arm_irq_memory_barrier(uintptr_t irq)
 {
 }
 
+#endif	/* !LOCORE */
+
 #ifdef DEV_ACPI
 #define	ACPI_INTR_XREF	1
 #define	ACPI_MSI_XREF	2
 #define	ACPI_GPIO_XREF	3
 #endif
```
ehem, line 52: I object to this since these will almost certainly **never** be used by assembly language.  I understand the instinct to make as much as possible available to assembly language, but the ACPI constants are too improbable to merit being made available.

ehem (changes requested):
Please take a look at: https://github.com/ehem/freebsd-src/tree/pull1363-0 and https://github.com/ehem/freebsd-src/tree/pull1363
(you can examine them in `git` by doing `git remote add ehem https://github.com/ehem/freebsd-src.git; git fetch ehem pull1363 pull1363-0`)

The two `[SQUASH]` things are the changes I **really** want to see here.  Mainly there should instead be a typedef for the root designation.  The `[consider SQUASH?]` are things I'm unsure about insisting on.

If they're not enumerated values as part of #1363, be assured those will be added later (I have a planned use for enumerated values).

ayrtonm:
@ehem
re: enums or macros for the root number
Like [I said before](https://github.com/freebsd/freebsd-src/pull/1363#discussion_r1712670528) in more involved cases enums seem error-prone when you need to access the values from asm. I'm not a fan of the approach and since I haven't heard from others I'm leaning towards keeping the macros for now. If you have a use case that absolutely needs enums please add them in a separate change.

re: replacing `sys/intr.h` with `machine/intr.h`in `arm{,64}/exception.S`
Sure that sounds reasonable. I used `sys/intr.h` to minimize the `#ifdef`s needed in headers (the ones needed to include `arm/intr.h` look particularly silly), but I don't have a strong opinion here.

re: expanding `arm64/include/intr.h` `#ifdef` to exclude `ACPI_*_XREF`
Sorry the comments in this PR are getting a bit repetitive and I simply missed your initial suggestion. That change seems fine to me.

sys/arm64/arm64/exception.S
ayrtonm: Pointing this out twice on `arm64/exception.S` in one set of comments seems unnecessary. Maybe it was unintentional or you're not used to github's UI, but in general repeating yourself and being overly-insistent will make others less willing to engage with your suggestions (especially given that the review workflow isn't what people are used to).

mhorne (approved):
This looks pretty good overall, I am willing to help land the commits unless @kevans91  or @strejda is planning on it.

A few months back @jrtc27 added logic to `subr_intr.c` for registering an alternate root device for IPI interrupts (`intr_ipi_pic_register()`). This is obviously applicable to this notion of multiple "root types", so my intention would be to merge this functionality so that we have (for riscv):

```
#define INTR_ROOT_IPI 1
```
in `riscv/include/intr.h`, and:
```
intr_pic_claim_root(dev, xref, sbi_ipi_intr, sc, INTR_ROOT_IPI);
```
in `riscv/riscv/sbi_ipi.c`, and future drivers (APLIC) implementing IPIs.

The IPI logic has a notion of "priority", for competing root devices, which could be added to the `intr_pic_claim_root()` path without too much difficulty, I think. So for example we might end up with the variant:
```
int
intr_pic_claim_root_pri(device_t dev, intptr_t xref, intr_irq_filter_t *filter,
    void *arg, uint32_t rootnum, uint32_t priority)
```

I am not asking you to implement this, but it should be considered within the scope of modifications to the INTRNG KPI that are being made here.

I left a few minor comments, mostly matters of taste, to be handled as you see fit.

sys/sys/intr.h
```diff
@@ -33,10 +33,15 @@
 #error Need INTRNG for this file
 #endif
 
+#ifndef LOCORE
 #include <sys/systm.h>
+#endif
 
 #define	INTR_IRQ_INVALID	0xFFFFFFFF
 
+#define INTR_ROOT_IRQ	0
```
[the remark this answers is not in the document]
ayrtonm: > hurts reader comprehension

Agreed and I think I initially had what you're suggesting before I was asked to move it. Maybe a good compromised might be `#define INTR_ROOT_DEFAULT 0` in `sys/intr.h` and `#define INTR_ROOT_IRQ INTR_ROOT_DEFAULT` in `machine/intr.h`. That way MD variants for each arch are all defined in the same places and the value for the default root still matches across archs.

sys/kern/subr_intr.c
```diff
@@ -877,9 +900,10 @@ intr_pic_deregister(device_t dev, intptr_t xref)
  */
 int
 intr_pic_claim_root(device_t dev, intptr_t xref, intr_irq_filter_t *filter,
-    void *arg)
+    void *arg, uint32_t rootnum)
```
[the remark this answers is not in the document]
ayrtonm: Yeah "types" definitely makes more sense but there's an unrelated `intr_type_` enum so I avoided `INTR_TYPE_` for the macros. I agree `rootnum` could be made less ambiguous though, so maybe `roottype` or `intr_type`?

sys/sys/intr.h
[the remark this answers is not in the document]
jrtc27: Having another level of define for something that's realistically always going to be 0 is a waste, just put it back to how it was. Sorry for the busywork induced by non-committers.

sys/kern/subr_intr.c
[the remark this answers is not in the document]
mhorne: > Yeah "types" definitely makes more sense but there's an unrelated `intr_type_` enum so I avoided `INTR_TYPE_` for the macros. I agree `rootnum` could be made less ambiguous though, so maybe `roottype` or `intr_type`?
> 

Oh yeah I forgot about the existing INTR_TYPE_ definitions. Forget that suggestion then.

`roottype` works for me.

sys/arm64/arm64/exception.S
ehem: My apologies if you felt this was unnecessary.  I may not be the greatest reviewer.  I'm not at 100% and I am aware I do annoy a number of people (things which you shouldn't wish on anyone, even your enemies).

ehem:
> @ehem re: enums or macros for the root number Like [I said before](https://github.com/freebsd/freebsd-src/pull/1363#discussion_r1712670528) in more involved cases enums seem error-prone when you need to access the values from asm. I'm not a fan of the approach and since I haven't heard from others I'm leaning towards keeping the macros for now. If you have a use case that absolutely needs enums please add them in a separate change.

I'm guessing no one has commented simply due to no one having an opinion.  At first glance `uint32_t` _seems_ reasonable.  This though has a weakness in most architectures will instead pass a register-sized value.  I doubt any near-future device will have _that many_ roots, but it does suggest `uint32_t` may not be the right type.

Having to reread that comment several times to try to make sure I respond to the correct issue...

Both `enum` and `#define` have some flavor of update failure mode.  For `enum`, if the members get out of order (or more specifically, if the one with the greatest value **isn't** at the end) the end value will be wrong and everything will explode.  Good news is keeping things in numeric order isn't too difficult.

Having the count be a `#define` is actually similarly brittle.  If someone adds a new entry, but forgets to update `INTR_ROOT_COUNT` then things will explode in a similar way.  Getting them out of numeric order will make this more likely.  From my view these failure modes are pretty well equivalent.

For ARM{,64} `#define` has a real advantage of the constants need to be shared with assembly language.  As such having the `enum` alongside the `#define` is mildly disadvantageous for actual hardware root interrupt controllers.

For RISC-V `enum` has an advantage of there is no need to share with assembly language and any extra roots won't need to have their numbers manually assigned.  For instance [here](https://github.com/freebsd/freebsd-src/pull/1363#pullrequestreview-2243372216), @mhorne could add:
```
#ifdef SMP
        INTR_ROOT_IPI,
#endif 
```
There would be no need to manually assign the root number.  If a uniprocessor kernel was built, there wouldn't be an extra root.

_I_ would end up with a commit adding:
```
#if defined(XENHVM)
        INTR_ROOT_HYPERVISOR,
#endif
```
Then all kernels with support for a hypervisor could hook into this root and get `pic_init_secondary()` functionality (I don't know whether the others need the functionality, but it _could_ be shared).

Having multiple optional virtual roots would require unpleasant hacks with `#define` whereas with `enum` these are very simple.

sys/kern/subr_intr.c
```diff
@@ -89,6 +89,15 @@
 
 #define	INTRNAME_LEN	(2*MAXCOMLEN + 1)
 
+/*
+ * Archs may define multiple roots with INTR_ROOT_NUM to support different kinds
+ * of interrupts (e.g. arm64 FIQs which use a different exception vector than
+ * IRQs).
+ */
+#if !defined(INTR_ROOT_NUM)
```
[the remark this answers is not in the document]
ayrtonm: The idea was that only archs/configs that want to opt-in to multiple PIC roots would need to define this and the rest would just use the default, but I'm not totally opposed to the suggestion.

kevans91:
> This looks pretty good overall, I am willing to help land the commits unless @kevans91 or @strejda is planning on it.
> 

I don't have any immediate plans, but I'll definitely pick it up if nobody else does when I find some time- this change in particular will unblock the Apple Silicon aic driver and a bunch of other stuff we have for AS support.

sys/kern/subr_intr.c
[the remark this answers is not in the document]
ehem: I concur with @mhorne here.  Having no default and requiring the architecture to define these keeps them all in one place and will result in less confusion.  Notably ARM64 will deviate from the default due to FIQ and there was discussion of RISC-V adding IPI.  At which point the "default" would only apply to ARM and it is silly to have a default for only a single architecture.

ehem:
There was a burst of discussion on [D40161](https://reviews.freebsd.org/D40161).  The comment may have been more about dislike of using bitfields, but @strejda was suggesting [the use of `enum`](https://reviews.freebsd.org/D40161#953962) there.  I'm unsure whether it is worthwhile to make it possible to build a kernel without FIQ support, but `enum`s do make this rather easier.

I'm in concurrence with @mhorne and not having any defaults and simply letting builds fail if `INTR_ROOT_COUNT` is undefined.  I felt uneasy about those defaults, but I was having trouble figuring out what to suggest.

strejda:
I'm OK with this.  The remaining problems (new type for IPI) can be solved later (and by using phabricator if I may ask :)  ).
@mhorne can you, please, can you commit it? The github environment is for me slightly unfriendly...

bsdimp:
i'll land it today...

bsdimp:
I've added co-authored-by for @kevans91 and @zxombie before the push.

ehem:
Uhmm...   Near the top of the right-side column, "Reviewers".  Under that, @mhorne is listed as "mhorne approved these changes" (checkmark); think "Add Action" => "Accept Revision".  Whereas I'm listed as "ehem requested changes"; think "Add Action" => "Request Changes".

I had been waiting for activity, but then two of you jump in and it is in FreeBSD's tree in less than 2 hours (after I'd been waiting for a full week).

In particular there are 3 things I found **Not Acceptable**.

1. @strejda had [requested the type be `enum`](https://reviews.freebsd.org/D40161#953962) in [D40161](https://reviews.freebsd.org/D40161).  I was suggesting the [use of a `typedef`](https://github.com/ehem/freebsd-src/commit/4c4459296452f91fe3b6e7e90e6e7c41a93370aa), though I was expecting most architectures would end up using `enum`s.
2. I had [already suggested](https://github.com/freebsd/freebsd-src/pull/1363#discussion_r1712187021) using `INTR_ROOT_COUNT`.  I suspect @mhorne independently came to the [same conclusion](https://github.com/freebsd/freebsd-src/pull/1363#discussion_r1755156418).  Yet this wasn't fixed, event when brought into FreeBSD's tree.
3. It hadn't seem quite right, but it was @mhorne who came up with [what seemed the right solution](https://github.com/freebsd/freebsd-src/pull/1363#discussion_r1720217726).  Yet this wasn't fixed in FreeBSD's tree.

Good news is I found a better approach to one concern I'd had with the present state.  Mainly the assembly to C interface didn't look quite right.

ehem:
Looks like #1431 is the number I got.

ayrtonm:
Looks like you got to this before I had time to fix up the last few nits, but what got committed looks good to me, thanks @bsdimp!

PR #1365, opened 2024-08-06
wsp: multi-finger touchpad improvements

MegaManSec:
Currently, the wsp driver limits the ability for a single-click (left click) on the touchpad followed by a partial release (finger still on the trackpad). When you single-click then partially release, the trackpad is frozen and you cannot continue movement until you fully release the button and re-touch the trackpad for movement.

This patch provides the ability to continue movement after a single-click has taken place.

The tunable hw.usb.wsp.enable_single_tap_movement (1=enabled by default) has already been created for those that do not want to allow the movement of the trackpad after a single click.

Likewise, the wsp driver does not currently correctly allow the movement of the trackpad using two fingers while a single button is held down: that is to say, during a single-button click, it is only possible to move the cursor using the same finger that is clicking.

This patch provides support for the movement using a second finger, which is in-line with how MacOS supports this trackpad: a single button click can be moved using either the original finger or a secondary finger on the trackpad following the first click. It appears that this was originally missed by the developer, as the comments suggest that the variables holding the position arrays should be arrays, however they were not:
 int16_t pre_pos_x;              /* previous position array */
 int16_t pre_pos_y;              /* previous position array */
This patch makes them arrays of int16_t, in-line with the "finger index data" and "position array" arrays.

In addition, this patch also creates two tunables:
hw.usb.wsp.max_finger_area
hw.usb.wsp.max_double_tap_distance

max_finger_area is the maximum area that a finger may take up on the trackpad to be registered as a finger. Previously, it was hardcoded 1200. This value was too low to register a thumb-click, which MacOS does correctly.

max_double_tap_distance is the maximum distance between two fingers which permit a two-finger-click to be registered as a double-click. Previously, this was hardcoded as 2500, and that is left as the default.

### DIFF
--- a/UPDATING
+++ b/UPDATING
@@ -2022,6 +2022,8 @@ COMMON ITEMS:
 	make -DALWAYS_CHECK_MAKE buildkernel KERNCONF=YOUR_KERNEL_HERE
 	make -DALWAYS_CHECK_MAKE installkernel KERNCONF=YOUR_KERNEL_HERE
 
+	If you are running kernel modules from ports, see FOOTNOTE [1].
+
 	To test a kernel once
 	---------------------
 	If you just want to boot a kernel once (because you are not sure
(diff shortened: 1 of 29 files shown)


bsdimp:
@wulf7 Can you take a look? This seems up your alley, but if not just say so.

PR #1367, opened 2024-08-06
nmdm: Allow module to be force-unloaded.

MegaManSec:
If an nmdm device was created by accident such as by running
`ls /dev/nmdmAA`, it would restrict the module from being unloaded,
despite no data passing through it.

With this patch, `kldunload nmdm` will still fail if any nmdm device still
exists.

`kldunload -f nmdm` will try to destroy all nmdm devices before unloading.
If an nmdm modem is either opened in blocking mode or data is being
passed through it, after five seconds of trying, it will not be destroyed
and the unload will also fail.

Likewise, fix an incorrect tty_lock, a typo, and the order of locking.

### DIFF
--- a/sys/dev/nmdm/nmdm.c
+++ b/sys/dev/nmdm/nmdm.c
@@ -95,9 +95,43 @@ struct nmdmsoftc {
 	struct nmdmpart	ns_part1;
 	struct nmdmpart	ns_part2;
 	struct mtx	ns_mtx;
+	TAILQ_ENTRY(nmdmsoftc)	ns_list;
 };
 
 static int nmdm_count = 0;
+
+static TAILQ_HEAD(, nmdmsoftc) nmdmsoftc_list = TAILQ_HEAD_INITIALIZER(nmdmsoftc_list);
+static struct mtx nmdmsoftc_lock;
+MTX_SYSINIT(nmdmsoftc_lock, &nmdmsoftc_lock, "nmdmsoftc list", MTX_DEF);
+
+static void
+nmdm_destroy(struct tty *tp)
+{
+	struct nmdmpart *np;
+	struct nmdmpart *onp;
+	struct tty *otp;
+
+	np = tty_softc(tp);
+
+	if (np == NULL)
+		return;
+
+	if (tp) {
+		/* Shut down self. tty_rel_gone() schedules a deferred free callback. */
+		tty_lock(tp);
+		tty_rel_gone(tp);
+
+		/* Shut down second (other) part. */
+		onp = np->np_other;
+		if (onp == NULL)
+			return;
+
+		otp = onp->np_tty;
+
+		tty_lock(otp);
+		tty_rel_gone(otp);
+	}
+}
 
 static void
 nmdm_close(struct tty *tp)
@@ -114,17 +148,9 @@ nmdm_close(struct tty *tp)
 	if (tty_opened(otp))
 		return;
 
-	/* Shut down self. */
-	tty_lock(tp);
-	tty_rel_gone(tp);
-
-	/* Shut down second part. */
-	onp = np->np_other;
-	if (onp == NULL)
-		return;
-	otp = onp->np_tty;
-	tty_lock(otp);
-	tty_rel_gone(otp);
+	mtx_lock(&nmdmsoftc_lock);
+	nmdm_destroy(tp);
+	mtx_unlock(&nmdmsoftc_lock);
 }
 
 static void
@@ -148,6 +174,11 @@ nmdm_free(void *softc)
 		return;
 	}
 	mtx_destroy(&ns->ns_mtx);
+
+	mtx_lock(&nmdmsoftc_lock);
+	TAILQ_REMOVE(&nmdmsoftc_list, ns, ns_list);
+	mtx_unlock(&nmdmsoftc_lock);
+
 	free(ns, M_NMDM);
 	atomic_subtract_int(&nmdm_count, 1);
 }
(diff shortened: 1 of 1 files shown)


bsdimp (approved):
This looks like a good fix to me. @kevans91 is the closest thing we have to a tty expert though.

sys/dev/nmdm/nmdm.c
```diff
@@ -115,16 +115,16 @@ nmdm_close(struct tty *tp)
 		return;
 
 	/* Shut down self. */
+	tty_lock(tp);
```
kevans91: This is wrong, `tp` is locked on entry

sys/dev/nmdm/nmdm.c
```diff
@@ -115,16 +115,16 @@ nmdm_close(struct tty *tp)
...
 	tty_rel_gone(tp);
 
 	/* Shut down second part. */
-	tty_lock(tp);
```
kevans91, line 121: This is necessary because they share a single mutex, and it needs to be held for the below `tty_rel_gone`.  I'd be in favor of switching this one to `otp` to make it more clear, though.

sys/dev/nmdm/nmdm.c
```diff
@@ -115,16 +115,16 @@ nmdm_close(struct tty *tp)
...
 	onp = np->np_other;
 	if (onp == NULL)
 		return;
 	otp = onp->np_tty;
+	tty_lock(otp);
 	tty_rel_gone(otp);
-	tty_lock(tp);
```
kevans91, line 127: This needs to be restored because this tty needs to remain locked on exit, but we dropped it with `tty_rel_gone` again

kevans91 (changes requested):
Some problems with the locking bits, and IMO the other half is working around a bug.  I don't see where the current implementation would be intentionally stopping destruction of a nmdm pair that was only ever half open.

kevans91:
> Some problems with the locking bits, and IMO the other half is working around a bug. I don't see where the current implementation would be intentionally stopping destruction of a nmdm pair that was only ever half open.

Ah, right- we clone it in response to stat(), nothing actually opened it or ever planned on opening it. Hmm.

sys/dev/nmdm/nmdm.c
MegaManSec: Right: I mixed up the locks here, and this effective double locking (in the later commit where this lock is called from nmdm_destroy) hides a bug too. I'll make some changes and post them here.

sys/dev/nmdm/nmdm.c
```diff
@@ -416,6 +452,29 @@ nmdm_outwakeup(struct tty *tp)
 	taskqueue_enqueue(taskqueue_swi, &np->np_task);
 }
 
+static int
+nmdm_unload(void *data)
+{
+	struct nmdmsoftc *ns;
+	sbintime_t start_time = sbinuptime();
+	sbintime_t timeout = SBT_1S * 5; // Try to force unload for 5 seconds
```
kevans91, line 460: I think this is entirely valid, but I'd really prefer to accomplish it without the timeout by adding some plumbing through the tty later to drain the deferred destroy_dev taskqueue. kern_conf.c doesn't seem to have quite the KPI that I think I'd like to see.

Ideally, I think we'd have some kind of `tty_drain_freed` -> `destroy_dev_drain_pending` -> `taskqueue_drain` that we can call to force the async bits to run, then we can just EBUSY it if we still have nmdm devices when we get back.

sys/dev/nmdm/nmdm.c
```diff
@@ -416,6 +452,29 @@ nmdm_outwakeup(struct tty *tp)
...
+
+	mtx_lock(&nmdmsoftc_lock);
+	TAILQ_FOREACH(ns, &nmdmsoftc_list, ns_list) {
+		nmdm_destroy(ns->ns_part1.np_tty);
```
kevans91, line 464: ... and `nmdm_destroy` should ideally return EBUSY if one of them has both ttys and at least one of the ttys is opened, so that we don't bother waiting at all if we already know it's not going to magically go away.

kevans91 (commented):
Actionable feedback on a path forward for the other half...

sys/dev/nmdm/nmdm.c
MegaManSec: "Fixed" within the realm of the second and third patch.

sys/dev/nmdm/nmdm.c
MegaManSec: "Fixed" within the realm of the second and third patch.

sys/dev/nmdm/nmdm.c
MegaManSec: "Fixed" within the realm of the second and third patch.

bsdimp:
@kevans91 Looks like this is ready for you to re-review since it looks like your feedback has been applied.

MegaManSec:
The remaining recommendation "_Ideally, I think we'd have some kind of tty_drain_freed -> destroy_dev_drain_pending -> taskqueue_drain_" is not done yet: I have not had time yet to try this.

bsdimp:
any update?

bsdimp:
last pinged over 6 months ago, with no update. Closing.
Feel free to re-open this or to create a new pull request if the required changes are made. Thanks for your submission. Sorry that it didn't make it in. But we're open to having a fixed version.

MegaManSec:
Unfortunately, I never got time to make the changes requested ("without the timeout by adding some plumbing through the tty layer to drain the deferred destroy_dev taskqueue").

Perhaps a freebsd bug report should be made for this, though.

MegaManSec:
addressed in https://github.com/freebsd/freebsd-src/pull/2207

PR #1368, opened 2024-08-06
dhclient: Set default ARP resolution wait to 250 ms, with an option to make it 0 ms

ica574:
Make the default ARP resolution wait time 250 ms, and add an option within dhclient to remove it altogether.

### DIFF
--- a/libexec/rc/rc.conf
+++ b/libexec/rc/rc.conf
@@ -138,6 +138,7 @@ dhclient_flags=""		# Extra flags to pass to dhcp client.
 #dhclient_flags_em0=""		# Extra dhclient flags for em0 only
 background_dhclient="NO"	# Start dhcp client in the background.
 #background_dhclient_em0="YES"	# Start dhcp client on em0 in the background.
+dhclient_arpwait="YES"		# Wait for ARP resolution
 synchronous_dhclient="NO"	# Start dhclient directly on configured
 				# interfaces during startup.
 defaultroute_delay="30"		# Time to wait for a default route on a DHCP interface.
--- a/libexec/rc/rc.d/dhclient
+++ b/libexec/rc/rc.d/dhclient
@@ -48,6 +48,10 @@ dhclient_prestart()
 		rc_flags="${rc_flags} -b"
 	fi
 
+	dhclient_arpwait=$(get_if_var $ifn dhclient_arpwait_IF $dhclient_arpwait)
+	if ! checkyesno dhclient_arpwait; then
+		rc_flags="${rc_flags} -n"
+	fi
 
 	# /var/run/dhclient is not guaranteed to exist,
 	# e.g. if /var/run is a tmpfs
--- a/sbin/dhclient/dhclient.8
+++ b/sbin/dhclient/dhclient.8
@@ -36,7 +36,7 @@
 .\" see ``http://www.isc.org/isc''.  To learn more about Vixie
 .\" Enterprises, see ``http://www.vix.com''.
 .\"
-.Dd August 4, 2018
+.Dd August 1, 2024
 .Dt DHCLIENT 8
 .Os
 .Sh NAME
@@ -82,6 +82,10 @@ will revert to running in the background.
 Specify an alternate location,
 .Ar file ,
 for the leases file.
+.It Fl n
+Make
+.Nm
+not wait for ARP resolution.
 .It Fl p Ar file
 Specify an alternate location for the PID file.
 The default is
--- a/sbin/dhclient/dhclient.c
+++ b/sbin/dhclient/dhclient.c
@@ -121,7 +121,7 @@ struct pidfh *pidfile;
  */
 #define TIME_MAX        ((((time_t) 1 << (sizeof(time_t) * CHAR_BIT - 2)) - 1) * 2 + 1)
 
-static struct timespec arp_timeout = { .tv_sec = 2, .tv_nsec = 0 };
+static struct timespec arp_timeout = { .tv_sec = 0, .tv_nsec = 250 * 1000 * 1000 };
 static const struct timespec zero_timespec = { .tv_sec = 0, .tv_nsec = 0 };
 int		log_priority;
 static int		no_daemon;
@@ -386,7 +386,7 @@ main(int argc, char *argv[])
 	cap_openlog(capsyslog, getprogname(), LOG_PID | LOG_NDELAY, DHCPD_LOG_FACILITY);
 	cap_setlogmask(capsyslog, LOG_UPTO(LOG_DEBUG));
 
-	while ((ch = getopt(argc, argv, "bc:dl:p:qu")) != -1)
+	while ((ch = getopt(argc, argv, "bc:dl:np:qu")) != -1)
 		switch (ch) {
 		case 'b':
 			immediate_daemon = 1;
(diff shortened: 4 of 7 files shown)


cperciva:
@ica574 Looks like the style checker wants "static const" rather than "const static".  Can you change that?

It's also complaining about the timespeccmp calls but that's a bug in the style checker; please leave those as they are.

@bsdimp Do we ask for signed-off-by lines in github PRs?  I'm sure Isaac can add them, I just hadn't noticed them before and wasn't sure if this was something you(?) put into the style checker deliberately.

ica574:
> @ica574 Looks like the style checker wants "static const" rather than "const static". Can you change that?
> 
> It's also complaining about the timespeccmp calls but that's a bug in the style checker; please leave those as they are.
> 
> @bsdimp Do we ask for signed-off-by lines in github PRs? I'm sure Isaac can add them, I just hadn't noticed them before and wasn't sure if this was something you(?) put into the style checker deliberately.

@cperciva Done! I've changed "const static" to "static const".

sbin/dhclient/dhclient.c
```diff
@@ -120,6 +121,8 @@ struct pidfh *pidfile;
  */
 #define TIME_MAX        ((((time_t) 1 << (sizeof(time_t) * CHAR_BIT - 2)) - 1) * 2 + 1)
 
+static struct timespec arp_timeout = { .tv_sec = 2, .tv_nsec = 0 };
+const struct timespec zero_timespec = { .tv_sec = 0, .tv_nsec = 0 };
```
brooksdavis, line 125: It seems like this could be static?

sbin/dhclient/dhcpd.h
```diff
@@ -360,10 +361,13 @@ char *piaddr(struct iaddr);
 extern cap_channel_t *capsyslog;
 extern const char *path_dhclient_conf;
 extern char *path_dhclient_db;
+extern struct timespec time_now;
 extern time_t cur_time;
 extern int log_priority;
 extern int log_perror;
 
+extern const struct timespec zero_timespec;
```
brooksdavis, line 369: Why declare this when it's only used in dhclient.c?

cperciva:
Oops, I told @ica574 to make `zero_timespec` extern since I thought it was being used in multiple files.  I think I forgot how much code was all in `dhclient.c`.

Isaac, please mark it `static` and remove it from `dhcpd.h`

sbin/dhclient/dhclient.8
```diff
@@ -95,6 +95,9 @@ Forces
 .Nm
 to reject leases with unknown options in them.
 The default behaviour is to accept such lease offers.
+.It Fl n
+Don't wait for ARP resolution within
+.Nm
```
concussious, line 100: Please keep the flags alphabetized by moving the added text to after line 84. Also the sentances explaining other flags are ending with a period. Thanks!

concussious (changes requested):
Everyone doesn't always do it, but it'd be nice if the commit message explains why you want to do that for someone unfamiliar with the topic.

concussious (approved):
My wheelhouse is only manuals, but manual LGTM! Congrats on your project, changing the world is so exciting!

libexec/rc/rc.conf
```diff
@@ -138,6 +138,7 @@ dhclient_flags=""		# Extra flags to pass to dhcp client.
 #dhclient_flags_em0=""		# Extra dhclient flags for em0 only
 background_dhclient="NO"	# Start dhcp client in the background.
 #background_dhclient_em0="YES"	# Start dhcp client on em0 in the background.
+dhclient_noarpwait="NO"	# Trust IP address from DHCP and don't ARP
```
tomhukins, line 141: The double negative in `dhclient_noarpwait="NO"` reads awkwardly.  Reversing its logic, naming it `dhclient_wait_for_arp`, and setting its default value to `"YES"` might make it easier to understand.

cperciva: Good point.  I suggested "noarpwait" since that's what the `-n` flag means (and it needs to be that way around to keep the previous "no flags means wait for ARP" behaviour) but you're quite right that for a new rc.conf setting we can reverse that.

@ica574 How about `dhclient_arpwait="YES" # ARP IP addresses before using them` and a corresponding change in the rc.d script?

ica574: Sounds good. I'll implement it!

libexec/rc/rc.conf
```diff
@@ -138,6 +138,7 @@ dhclient_flags=""		# Extra flags to pass to dhcp client.
 #dhclient_flags_em0=""		# Extra dhclient flags for em0 only
 background_dhclient="NO"	# Start dhcp client in the background.
 #background_dhclient_em0="YES"	# Start dhcp client on em0 in the background.
+dhclient_arpwait="YES"	# Trust IP address from DHCP and don't ARP
```
cperciva, line 141: You forgot to update the comment

libexec/rc/rc.d/dhclient
```diff
@@ -48,6 +48,10 @@ dhclient_prestart()
 		rc_flags="${rc_flags} -b"
 	fi
 
+	dhclient_arpwait=$(get_if_var $ifn dhclient_arpwait_IF $dhclient_arpwait)
+	if checkyesno dhclient_arpwait; then
```
cperciva, line 52: and this is now backwards; we need to add `-n` if `dhclient_arpwait` is set to *no*.

libexec/rc/rc.conf
ica574: My bad, should be fixed now.

concussious:
Should we add an entry for `dhclient_arpwait` to rc.conf.5?

libexec/rc/rc.conf
```diff
@@ -138,6 +138,7 @@ dhclient_flags=""		# Extra flags to pass to dhcp client.
 #dhclient_flags_em0=""		# Extra dhclient flags for em0 only
 background_dhclient="NO"	# Start dhcp client in the background.
 #background_dhclient_em0="YES"	# Start dhcp client on em0 in the background.
+dhclient_arpwait="YES"	# Wait for ARP resolution
```
cperciva, line 141: Looks like we need an extra tab to line the comment up with the rest?  Or is that just github being wonky?

cperciva:
> Should we add an entry for `dhclient_arpwait` to rc.conf.5?

Yes, that's probably a good idea.  @ica574 Can you add this?  Write something about not ARPing IP addresses before using them and this making the system boot faster on networks where the DHCP server is guaranteed to know whether an address is available.

cperciva:
@ica574 Can you add Signed-off-by tags to all of the commits?  I think `git rebase --signoff` might do this.

share/man/man5/rc.conf.5
```diff
@@ -618,6 +618,15 @@ When specified, this variable overrides the
 variable for interface
 .Ar iface
 only.
+.It Va dhclient_arpwait
+.Pq Vt bool
+Set to
+.Dq Li YES
```
cperciva, line 624: YES is the default, so for `rc.conf(5)` we should say what happens if we set it to NO.

cperciva:
Committed, thanks!

PR #1372, opened 2024-08-08
Fix RTL8156 support to use the if_ure(4) driver instead

kenrap:
The problem here is that there are two competing drivers, `if_cdce(4)` and `if_ure(4)`, which are in conflict of each other.

USB Ethernet devices that use either RTL8156 and RTL8156B chips would automatically load the `if_cdce(4)` driver on USB attachment which limits the speed to roughly 1/3 Gbit/s.

Using `if_ure(4)` allows utilization of "2500Base-T \<full-duplex\>" for these devices which `if_cdce(4)` doesn't, according to `ifconfig(8)`. In fact `if_cdce(4)` doesn't mention any link speed when inserting an Ethernet cable to these devices.

References to the problem from the following forum posts:
https://www.truenas.com/community/threads/usb-3-1-gen2-2-5gbps-ethernet-adapter.107273/
https://forums.freebsd.org/threads/plugable-usbc-e2500-unable-to-select-between-if_cdce-and-if_ure.88052/


lwhsu:
Thanks, I have https://reviews.freebsd.org/D45088 and currently under review and test.

PR #1373, opened 2024-08-09
ntptime: Use time_t for tv_sec related variables

sebhub:
The struct timespec tv_sec member is of type time_t.  Make sure that all variables related to this member are of the type time_t.  This is important for targets where long is a 32-bit type and time_t a 64-bit type.


bsdimp (approved):
A good change for 32bit arm and powerpc. Everything else is a nop practically.... i386 is left behind, but it won't be in 15, so will only live until 2028 in a supported branch. Since that's well short of the deadline, even the really long tail of use will be fine enough: i386 deployments will have to sunset well before then and are already only a tiny sliver of new installs.

PR #1374, opened 2024-08-09
sound: Fix SEQ_SYSEX() macro

rozhuk-im:
Correct implementation also found in Linux's soundcard.h:
https://github.com/torvalds/linux/blob/master/include/uapi/linux/soundcard.h#L1178

seq_processevent() interprets event[1] as device number and does not
work properly with current SEQ_SYSEX() implementation.

### DIFF
--- a/sys/sys/soundcard.h
+++ b/sys/sys/soundcard.h
@@ -1400,8 +1400,9 @@ void seqbuf_dump(void);	/* This function must be provided by programs */
 	int i, l=(len); if (l>6)l=6;\
 	_SEQ_NEEDBUF(8);\
 	_seqbuf[_seqbufptr] = EV_SYSEX;\
-	for(i=0;i<l;i++)_seqbuf[_seqbufptr+i+1] = (buf)[i];\
-	for(i=l;i<6;i++)_seqbuf[_seqbufptr+i+1] = 0xff;\
+	_seqbuf[_seqbufptr+1] = (dev);\
+	for(i=0;i<l;i++)_seqbuf[_seqbufptr+i+2] = (buf)[i];\
+	for(i=l;i<6;i++)_seqbuf[_seqbufptr+i+2] = 0xff;\
 	_SEQ_ADVBUF(8);}
 
 #define SEQ_CHN_PRESSURE(dev, chn, pressure) \


markjdb:
@christosmarg could you take a look?

christosmarg:
Not very experienced with this specific codebase, but the change seems like it's only mirroring what the Linux header does. LGTM.

christosmarg:
https://reviews.freebsd.org/D46377

rozhuk-im:
It is not mirroring linux, it makes it how it should be:
1. macro already have ```dev``` arg but it forgot to use
2. ```seq_processevent()``` expect in ```event[1]``` dev index, it check that passed index is correct and route message to device pointed by index.

I found this than make ```sequencer``` in user space, and found that linux have correct version of this macro than write description to this PR.

christosmarg:
Fixed the commit message in the review.

christosmarg:
LGTM. I only think the commit message could be improved a bit, something like:

```
Correct implementation also found in Linux's soundcard.h:
https://github.com/torvalds/linux/blob/master/include/uapi/linux/soundcard.h#L1178

seq_processevent() interprets event[1] as device number and does not
work properly with current SEQ_SYSEX() implementation.
```

christosmarg (approved):
LGTM. I only think the commit title and message could be improved a bit, something like:

```
sound: Fix SEQ_SYSEX() macro

Correct implementation also found in Linux's soundcard.h:
https://github.com/torvalds/linux/blob/master/include/uapi/linux/soundcard.h#L1178

seq_processevent() interprets event[1] as device number and does not
work properly with current SEQ_SYSEX() implementation.
```

bsdimp:
Please take a look at the 'Files Changed' tab. The style filter has a number of issues that need to be corrected first.

rozhuk-im:
I use style from file that I edit.
It will be strange to mix styles in 1 file.

christosmarg:
> I use style from file that I edit. It will be strange to mix styles in 1 file.

I second that. soundcard.h suffers from terrible and inconsistent style in general.

bsdimp:
OK. I'll do a style override and commit this under the 'matches existing' exception. Thanks for the explanation.

PR #1376, opened 2024-08-11
cross-build: fix fake_sysctl/tzsetup

valpackett:
The error was always returned, even after handling the sysctl, breaking installworld under Linux.

Sponsored by:		https://www.patreon.com/valpackett


bsdimp (approved):
Looks right to me

PR #1381, opened 2024-08-12
loader: fix/cleanup staging commands

VexedUXR:
A few fixes and cleanup for the staging commands:
The staging_slop command never actually changes the staging slop, so fix that.

The `res` variable used in both the staging commands just adds unnecessary complexity and can be removed.

Both the commands print their errors directly instead of the usual `command_errmsg = "..."`. This causes problems since `interp_run` in `interp_lua` attempts to run the command twice, once through `loader.command` in `cli_execute` and another through the normal `interp_builtin_cmd`. This causes duplicated output along with the default "no error message". I'm not entirely sure this behavior is correct, but we should be using `command_errmsg` either way.
Also print use the same error message if argc > 2 instead of printing "no error message".


bsdimp:
@markjdb and I  think this look good. @kostikbel does this look good to you?

kostikbel (approved):

PR #1388, opened 2024-08-20
libefivar: Fix AcpiEx print logic

jlduran:
Add logic that checks if the code doesn't overflow ACPI_EXTENDED_HID_DEVICE_PATH node when searching for optional strings. If the string is not provided in the device path node default value of "\0" is used.

Upstream PR:	https://bugzilla.tianocore.org/show_bug.cgi?id=4555
Obtained from:	https://github.com/tianocore/edk2/commit/96ed60dfd7d4818a532216d64ee87b13fae5c726


bsdimp (approved):
lots of somehwat gratuitous churn in here. But its's what's upstream. Kudos on discovering the real upstream for this file that's far outside of contrib...

PR #1389, opened 2024-08-21
0024 pf: Handle m_len < sizeof(struct ether_header) case

ihoro:


ihoro:
It goes here instead: https://reviews.freebsd.org/D46391

PR #1390, opened 2024-08-21
dummynet: fix pie

fichtner:
Since 26b9e1f07fa codel was fixed but traffic was not flowing for
pie too. Apply the same fix.

MFC after:	1 week
Sponsored by:	OPNsense
Differential Revision:	https://reviews.freebsd.org/D46182
Also see:	https://redmine.pfsense.org/issues/13996
Also see:	https://forum.opnsense.org/index.php?topic=41827.0


bsdimp:
@kprovost Seems reasonable explanation. But I'm unsure.... Can you take a look and tell me what you think?

bsdimp:
@markjdb and I looked at it. We're going to land it: (1) it seems to fix a bug, the fix has been verified in a downstream and there were no objections when this was brought up in review. It's imperfect evidence, but strong enough to suggest the risk from this is low.

fichtner:
Thank you.

PR #1391, opened 2024-08-21
mcast: fix leaked igmp packets on multicast cleanup

fichtner:
The commits cherry-picked here have not been added to stable/14.
They have been known to fix IGMP leave messages which are currently
broken on both 14.0 and 14.1. The necessary MFC has been absent from
them for unknown reasons.

Differential Revision: https://reviews.freebsd.org/D46190

### DIFF
--- a/sys/netinet/in_mcast.c
+++ b/sys/netinet/in_mcast.c
@@ -1181,6 +1181,7 @@ inm_purge(struct in_multi *inm)
 		free(ims, M_IPMSOURCE);
 		inm->inm_nsrc--;
 	}
+	mbufq_drain(&inm->inm_scq);
 }
 
 /*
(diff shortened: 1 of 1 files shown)


bsdimp:
Normally we don't take MFC PRs. However, this one looks like it's easy enough to followup on so I'll allow it for now.

fichtner:
Thanks. To avoid chasing up committers for MFC on operational issues would it make sense to make this semi-recurring? One change of note is https://github.com/freebsd/freebsd-src/commit/ad874544d9f that would at least benefit 14.2 in the long run.

bsdimp:
Yea, the original committer is responsible for MFC requests. So that's the first line of attack on that.
Now that we have git, we can likely shift how we do MFC requests, but are still operating under the old ways.

fichtner:
Fair enough, thanks for responding. I'll chase Vincenzo then.

glebius:
These two commits had already been merged. I really don't understand why github doesn't reflect this.

emaste:
> I really don't understand why github doesn't reflect this.

The hashes in this pull request commits (d7d51e11219c0b739792711084a5a85efd2e1789, e2e07611e575a16bbebfb5171180117058786c79) don't match the hashes in stable/14 (31ad232d85656af77726626decce6c2b7fd824f6, e2e07611e575a16bbebfb5171180117058786c79) so GitHub does not know that the changes were merged.

bsdimp:
Even if they did match, it would only close it.
There is no good way to accept MFC pull requests.

PR #1392, opened 2024-08-21
Refactor error handling in lseek operations and remove unused includes

kfv:
This pull request refactors the error handling in lseek operations to ensure proper error checking and POSIX compliance. It also removes unused include statements, improving the overall cleanliness and maintainability of the codebase.


lib/libfigpar/figpar.c
```diff
@@ -268,19 +266,18 @@ parse_config(struct figpar_config options[], const char *path,
...
+			charpos--;
 
 			/*
 			 * Go back so we can read the character before the key
 			 * to check if the character is escaped (which means we
 			 * should continue).
 			 */
-			error = (int)lseek(fd, -2, SEEK_CUR);
-			if (error == -3) {
+			if (lseek(fd, -2, SEEK_CUR) == -1) {
```
bsdimp, line 280: The original code is a bit odd here...  -3 would be Linux-speak for ESRCH (a define that goes back to V7 times, so it's the same everywhere). But there's no ESRCH documented anywhere I can find for lseek (in linux or elsewhere).

bsdimp:
@devinteske Why the two error == -3 that this PR removes?

bsdimp:
I'll defer this util next week's PR run. If there's no more comments, I'll queue it for QA and pushing into main.

lib/libfigpar/figpar.c
rilysh: ESRCH means no process found matching the process ID, so for lseek() this check seems was wrong.

kfv: I spoke with Devin privately, and she mentioned that the use of -3 wasn't actually intended to represent ESRCH as she wrote figpar before ESRCH existed.

bsdimp:
OK. I think these changes are right, so I've added ready.

bsdimp (approved):

lib/libfigpar/figpar.c
rilysh: ESRCH (3) is an errno value, however it's checking lseek()'s return value.

bsdimp: ESRCH has been around since at least V3 Unix (even though userland errno didn't seem to exist until V4... with the caveat that our extant artifacts are limited beyond the man pages).
But it's clear we should check -1, so I'm landing it...

lib/libfigpar/string_m.c
```diff
@@ -24,11 +24,8 @@
  * SUCH DAMAGE.
  */
 
-#include <sys/types.h>
```
bsdimp, line 27: This is actually needed. But stdint.h is the new way to spell this for uintXX_t that this file uses. I'll fix.

PR #1393, opened 2024-08-22
mk: Add a BTI-report linker feature

jlduran:
Add support for specifying how to report the missing Branch Target Identification (BTI) linker feature on AArch64.

For:

- Kernel: `bti-report` on when the linker supports it
- Userspace: `bti-report` on when the linker supports it and `BTI_REPORT_ERROR` is defined

Fixes:	43e8849bc294 ("conf: Enable BTI checking in the arm64 kernel")

### DIFF
--- a/share/mk/bsd.lib.mk
+++ b/share/mk/bsd.lib.mk
@@ -99,7 +99,7 @@ LDFLAGS+= -Wl,-zretpolineplt
 LDFLAGS.bfd+= -Wl,-znoexecstack
 .if ${MK_BRANCH_PROTECTION} != "no"
 CFLAGS+=  -mbranch-protection=standard
-.if ${MACHINE_ARCH} == "aarch64" && defined(BTI_REPORT_ERROR)
+.if ${LINKER_FEATURES:Mbti-report} && defined(BTI_REPORT_ERROR)
 LDFLAGS+= -Wl,-zbti-report=error
 .endif
 .endif
--- a/share/mk/bsd.linker.mk
+++ b/share/mk/bsd.linker.mk
@@ -11,9 +11,11 @@
 # LINKER_FEATURES may contain one or more of the following, based on
 # linker support for that feature:
 #
-# - build-id:  support for generating a Build-ID note
-# - retpoline: support for generating PLT with retpoline speculative
-#              execution vulnerability mitigation
+# - build-id:   support for generating a Build-ID note
+# - retpoline:  support for generating PLT with retpoline speculative
+#               execution vulnerability mitigation
+# - bti-report: support for specifying how to report the missing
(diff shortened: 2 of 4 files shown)


gmshake:
@zxombie CC Andrew

bsdimp:
This also is in bsd.lib.mk and bsd.prog.mk. that suggests we should use the normal bsd.linker.mk file and define a LINKER_FEATURE for this and then use that all three places.

jlduran:
> This also is in bsd.lib.mk and bsd.prog.mk. that suggests we should use the normal bsd.linker.mk file and define a LINKER_FEATURE for this and then use that all three places.

Not sure if this is what you had in mind.

share/mk/bsd.lib.mk
```diff
@@ -99,7 +99,7 @@ LDFLAGS+= -Wl,-zretpolineplt
 LDFLAGS.bfd+= -Wl,-znoexecstack
 .if ${MK_BRANCH_PROTECTION} != "no"
 CFLAGS+=  -mbranch-protection=standard
-.if ${MACHINE_ARCH} == "aarch64" && defined(BTI_REPORT_ERROR)
+.if ${LINKER_FEATURES:Mbti-report}
```
zxombie, line 102: I included the `defined(BTI_REPORT_ERROR)` check as I wasn't sure if it would break ports.

jlduran: I'll add that as well.

jlduran: @zxombie: Just, to confirm... `BTI_REPORT_ERROR` is not defined anywhere, do you want to have this check globally? (i.e. as part of the linker feature).

zxombie: It's passed in on the command line (or appropriate make config file) to enable the check. Until we are happy anything it could break has been fixed I was planning on keeping it off.

jlduran: Done. Thank you for clarifying.
- [x] Guards against old linkers (fixes cross-builds)
- [x] Turned into a linker feature
- [x] Activated by a `BTI_REPORT_ERROR` variable

zxombie:
I think we ad a misunderstanding. I only wanted the `BTI_REPORT_ERROR` check on userspace. The kernel is always built with BTI enable so I added the `bti-report` check to ensure newly imported code is built with BTI so it won't break on hardware that implements it.

The logic sould be:
- Userspace: `bti-report` on when the linker supports it & `BTI_REPORT_ERROR` is defined
- Kernel: `bti-report` on when the linker supports it

jlduran:
I appreciate the clarification. It all makes sense now. Thank you!

jlduran:
Landed as 973bbdab47035ebd16200c63d095904924dc44d9.

PR #1394, opened 2024-08-22
Centralize the kernel and module metadata types

VexedUXR:
Currently, the way we get the kernel metadata pointer is by calling
preload_search_by_type with one of the following three: "elf kernel",
"elf32 kernel" and "elf64 kernel". Which one(s) we use isn't consistent
though. Sometimes we would only try "elf kernel", and other times we
would try one of the latter two if the first failed. However, the loader
only ever sets "elf kernel" as the kernel type.
See: https://github.com/freebsd/freebsd-src/blob/8afae0caf4c4816eb56b732fcd1a4b185e86098a/stand/common/load_elf.c#L92

Instead, make the kmdp a global, preload_kmdp, and initialize it using
preload_initkmdp in machdep.c (or machdep_boot.c on arm/64).
preload_initkmdp takes a single boolean argument that tells us whether
not finding kmdp is fatal or not.

Also, feel free to critique my commit message. I feel like it's not great,
but that's the best I could word it.

### DIFF
--- a/stand/common/modinfo.h
+++ b/stand/common/modinfo.h
@@ -6,6 +6,11 @@
 #ifndef COMMON_MODINFO_H
 #define COMMON_MODINFO_H
 
+extern const char md_modtype[];
+extern const char md_kerntype[];
+extern const char md_modtype_obj[];
+extern const char md_kerntype_mb[];
+
 int md_load(char *args, vm_offset_t *modulep, vm_offset_t *dtb);
 int md_load64(char *args, vm_offset_t *modulep, vm_offset_t *dtb);
 
--- a/sys/kern/subr_module.c
+++ b/sys/kern/subr_module.c
@@ -46,10 +46,14 @@
 vm_offset_t preload_addr_relocate = 0;
 caddr_t preload_metadata, preload_kmdp;
 
+const char preload_modtype[] = MODTYPE;
+const char preload_kerntype[] = KERNTYPE;
+const char preload_modtype_obj[] = MODTYPE_OBJ;
+
 void
 preload_initkmdp(bool fatal)
 {
-	preload_kmdp = preload_search_by_type("elf kernel");
+	preload_kmdp = preload_search_by_type(preload_kerntype);
 
 	if (preload_kmdp == NULL && fatal)
 		panic("unable to find kernel metadata");
--- a/sys/kern/subr_module.c
+++ b/sys/kern/subr_module.c
@@ -44,7 +44,16 @@
  */
 
 vm_offset_t preload_addr_relocate = 0;
-caddr_t preload_metadata;
+caddr_t preload_metadata, preload_kmdp;
+
+void
+preload_initkmdp(bool fatal)
+{
+	preload_kmdp = preload_search_by_type("elf kernel");
+
+	if (preload_kmdp == NULL && fatal)
+		panic("unable to find kernel metadata");
+}
 
 /*
  * Search for the preloaded module (name)
--- a/sys/sys/linker.h
+++ b/sys/sys/linker.h
@@ -270,7 +270,7 @@ void linker_kldload_unbusy(int flags);
  * Module lookup
  */
 extern vm_offset_t	preload_addr_relocate;
-extern caddr_t		preload_metadata;
+extern caddr_t		preload_metadata, preload_kmdp;
 
 extern void *		preload_fetch_addr(caddr_t _mod);
 extern size_t		preload_fetch_size(caddr_t _mod);
@@ -278,6 +278,7 @@ extern caddr_t		preload_search_by_name(const char *_name);
 extern caddr_t		preload_search_by_type(const char *_type);
 extern caddr_t		preload_search_next_name(caddr_t _base);
 extern caddr_t		preload_search_info(caddr_t _mod, int _inf);
+extern void		preload_initkmdp(bool _fatal);
 extern void		preload_delete_name(const char *_name);
 extern void		preload_bootstrap_relocate(vm_offset_t _offset);
 extern void		preload_dump(void);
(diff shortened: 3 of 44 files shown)


VexedUXR:
Fixed some style warnings

bsdimp:
I'll take a look. @kostikbel and @kevans91 should take a look too.

kostikbel:
My opinion is that this is an enormous churn.

Regardless of it, I think even the approach is not taken to the logical completion.  All "elf kernel" literals _must_ be replaced by the global const string which is used by reference, both in loader and in kernel.

Ideally, we would have some enumeration value used to identify type, instead of free-typed string, but for this change at least replacing all individual strings with the same reference to constant string is the least required step.

VexedUXR:
> My opinion is that this is an enormous churn.

Yes, but it centralizes where we get the metadata at least. It also removes a bunch of unnecessary checks where kmdp can't be NULL. IMO most of the changes are trivial.

> Regardless of it, I think even the approach is not taken to the logical completion. All "elf kernel" literals must be replaced by the global const string which is used by reference, both in loader and in kernel.

You're right, they should be. But for the kernel, after this patch, I believe the only reference to "elf kernel" would be in preload_initkmdp.

bsdimp:
I'll take a closer look at this. It's basically OK, but I think I have some misgivings I need to put into an actionable form.

VexedUXR:
Looking more into what @kostikbel said, I've decided to move "elf kernel" and friends to sys/linker.h as macros. The reason they're macros an no longer globals is so that they can be easily shared across both the loader and kernel code.

Other `preload_search_by_type` calls either use macros or use a single free-typed string (i.e it only has one reference), so I think this is better.

Also, I noticed that we also check for "elfN module" and "elfN obj module", so I removed those too.

sys/sys/linker.h
```diff
@@ -32,6 +32,7 @@
 #ifdef _KERNEL
 
 #include <machine/elf.h>
+#include <sys/systm.h>
```
kostikbel, line 35: This is extreme namespace pollution.
I suspect it is due to preload_initkmdp(), and I do not see why it needs to be inline.

sys/sys/linker.h
```diff
@@ -219,6 +219,14 @@ void linker_kldload_unbusy(int flags);
 
 #endif	/* _KERNEL */
 
+/*
+ * ELF file types
+ */
+#define KERNTYPE_MB	"elf multiboot kernel"
```
kostikbel, line 225: I still want these to be constant objects instead of literals hidden under a macro.

kostikbel: No, of course it cannot be shared between kernel and loader.  But inside kernel, and inside loader, you should use a symbol pointing to the same string.  This would avoid spending memory in both kernel and loader for copies of the same string literal.

For kernel, it is nice not to waste memory.  For loader, we have issues with the BIOS limitations, so even a byte saved matters.

VexedUXR: afaik this shouldn't be a problem; we compile with constant merging enabled, so all the string literals should point to the same place.
For GCC, -fmerge-constants is implicitly enabled at -O1 or higher.
For Clang, as far as I can tell, there isn't a way to disable it.
The merging itself is of course done by the linker.

```
$ uname -r
14.1-RELEASE
$ strings /boot/loader | grep 'elf kernel'
elf kernel
$ strings /boot/loader.efi | grep 'elf kernel'
elf kernel
$ strings /boot/kernel/kernel | grep 'elf kernel'
elf kernel
```

I do see your point though.

The reason I went for macros in the first place was because I couldn't find a good place for the global in the loader. It can't be in load_elf.c since that gets compiled twice and would cause a duplicate symbol error. We could wrap it in __elfN, but then we'd have two symbols for no reason and we would need to use __elfN in other files that reference the string. Outside of load_elf.c and load_elf_obj.c, I'm not sure where to put it.

kostikbel: I am curious, and do not have time to dig it myself.  How do linkers do the deduplication of the string literals?  I do not see any way for this to work except putting each literal into a dedicated section with specific attributes.

That said, I do not think that finding a place is a big problem.  Just pointing a finger into the wall, I suggest stand/common/boot.c.  If you do not like it, you can add a new file into libsa or common, to collect literals.

VexedUXR: > I am curious, and do not have time to dig it myself. How do linkers do the deduplication of the string literals? I do not see any way for this to work except putting each literal into a dedicated section with specific attributes.

Almost, but multiple string literals can be in the same section. e.g a file consisting of only
```c
const char *foo = "bar";
const char *bar = "foo";
```
puts both "bar" and "foo" in `.rodata.str1.1`. Using readelf, these are the section's flags: `ALLOC, MERGE, STRINGS`
Why the number at the end? Apparently its so that it can create another section is in case we need different alignment for another set of strings, but I didn't look too much into that.

The `STRINGS` flag means that the section consists of only null-terminated strings, and the `MERGE` flag instructs the linker to attempt to merge them. I'm not sure how exactly the actual merging happens though, that's about where the depth of my knowledge on this ends.

> That said, I do not think that finding a place is a big problem. Just pointing a finger into the wall, I suggest stand/common/boot.c. If you do not like it, you can add a new file into libsa or common, to collect literals.

Hm, I'll have another look around the loader, I feel like I dismissed the idea of a global too quickly. common/boot.c isn't a bad suggestion :)

VexedUXR:
Alright, settled on modinfo.c at the end
Also, needed to make `file_loadraw` accept a const char * for the type string instead of a plain char *

sys/kern/subr_module.c
```diff
@@ -46,10 +46,14 @@
 vm_offset_t preload_addr_relocate = 0;
 caddr_t preload_metadata, preload_kmdp;
 
+const char *preload_modtype = MODTYPE;
```
kostikbel, line 49: Why not `const char preload_modtype[] = MODTYPE'` ?

stand/common/modinfo.h
```diff
@@ -6,6 +6,11 @@
 #ifndef COMMON_MODINFO_H
 #define COMMON_MODINFO_H
 
+extern const char *md_modtype;
```
kostikbel, line 9: Same question, why not use the array type directly, avoiding pointer dereference?

kostikbel (commented):
I am fine with the patch series, modulo two small identical comments

sys/kern/subr_module.c
VexedUXR: Oh, that would be better yes

kostikbel (approved):

VexedUXR:
Update and rebase after b538d4911004ca541507166b8ec9689d2e87d1aa

VexedUXR:
Thanks!

PR #681, opened 2023-03-02
GitHub: Cross-build full matrix

jlduran:
Follow up of #668.

Notice all the DROP COMMITs, those are just to show the full matrix passing. As stated in the commit, it could be used in the future, when a build fails, FreeBSD's GitHub bot (aka Beastie) could notify with the following matrix (if all is green, there's no need):

| OS     | Architecture | clang-12 | clang-13 | clang-14 | clang-15 |
| ------:|:------------:|:--------:|:--------:|:--------:|:--------:|
| macOS  | amd64        | [![Cross-build Kernel][macos-latest-amd64-clang-12-badge]][actions]    | [![Cross-build Kernel][macos-latest-amd64-clang-13-badge]][actions]    | [![Cross-build Kernel][macos-latest-amd64-clang-14-badge]][actions]    | [![Cross-build Kernel][macos-latest-amd64-clang-15-badge]][actions]    |
| macOS  | aarch64      | [![Cross-build Kernel][macos-latest-aarch64-clang-12-badge]][actions]  | [![Cross-build Kernel][macos-latest-aarch64-clang-13-badge]][actions]  | [![Cross-build Kernel][macos-latest-aarch64-clang-14-badge]][actions]  | [![Cross-build Kernel][macos-latest-aarch64-clang-15-badge]][actions]  |
| Ubuntu | amd64        | [![Cross-build Kernel][ubuntu-latest-amd64-clang-12-badge]][actions]   | [![Cross-build Kernel][ubuntu-latest-amd64-clang-13-badge]][actions]   | [![Cross-build Kernel][ubuntu-latest-amd64-clang-14-badge]][actions]   | [![Cross-build Kernel][ubuntu-latest-amd64-clang-15-badge]][actions]   |
| Ubuntu | aarch64      | [![Cross-build Kernel][ubuntu-latest-aarch64-clang-12-badge]][actions] | [![Cross-build Kernel][ubuntu-latest-aarch64-clang-13-badge]][actions] | [![Cross-build Kernel][ubuntu-latest-aarch64-clang-14-badge]][actions] | [![Cross-build Kernel][ubuntu-latest-aarch64-clang-15-badge]][actions] |

[macos-latest-amd64-clang-12-badge]: https://github.com/jlduran/freebsd-src/actions/workflows/cross-bootstrap-tools.yml/badge.svg?branch=github-cross-build-full-matrix&jobname=amd64%20macos-latest%20(clang-12)
[...]

PR #1399, opened 2024-08-29
Use recent clang releases in github actions

zxombie:


jlduran:
Nice!
https://docs.freebsd.org/en/articles/committers-guide/#_current_compiler_versions might also need an update.

bsdimp:
@brooksdavis @jrtc27 Comments on this? @emaste

jlduran:
Maybe I should try to rekindle #681?

emaste (approved):

bsdjhb (approved):

brooksdavis:
We've still got at least one Ubuntu 20.04 runner for CheriBSD builds, but could probably get it upgraded without too much trouble. Not sure were the CL stands on deployed systems. I think oldest clang we care about is 14 for Morello. We've got a completed merge request https://git.morello-project.org/morello/llvm-project/-/merge_requests/293 to move to 15, but it's stalled and there's at least one derivative project that may not have caught up.

PR #1400, opened 2024-08-31
libc/stdio: Increase BUF in vfprintf.c and vfwprintf.c

VexedUXR:
With the new %b format specifier, we need at most 64 characters.

Also, the comment says that we need enough space to write a uintmax_t in octal plus one byte, I'm not sure what that's for. I guess its implying that we need a null terminator(?), but it doesn't look like we do.

This can be easily tested by doing:
```c
printf("%lb\n", 1ul << n); /* Where n is anywhere from 32 to 63 */
```

### DIFF
--- a/lib/libc/stdio/vfprintf.c
+++ b/lib/libc/stdio/vfprintf.c
@@ -291,13 +291,9 @@ vfprintf(FILE * __restrict fp, const char * __restrict fmt0, va_list ap)
 /*
  * The size of the buffer we use as scratch space for integer
  * conversions, among other things.  We need enough space to
- * write a uintmax_t in octal (plus one byte).
+ * write a uintmax_t in binary.
  */
-#if UINTMAX_MAX <= UINT64_MAX
-#define	BUF	32
-#else
-#error "BUF must be large enough to format a uintmax_t"
-#endif
+#define BUF	(sizeof(uintmax_t) * CHAR_BIT)
 
 /*
  * Non-MT-safe version
--- a/bin/sh/sh.1
+++ b/bin/sh/sh.1
@@ -1,4 +1,6 @@
 .\"-
+.\" SPDX-License-Identifier: BSD-3-Clause
+.\"
 .\" Copyright (c) 1991, 1993
 .\"	The Regents of the University of California.  All rights reserved.
 .\"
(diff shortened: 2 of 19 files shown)


VexedUXR:
Thinking about this some more, we should just be able to remove the whole conditional and make BUF `sizeof(uintmax_t) * CHAR_BIT`

lib/libc/stdio/vfprintf.c
```diff
@@ -289,13 +289,9 @@ vfprintf(FILE * __restrict fp, const char * __restrict fmt0, va_list ap)
 /*
  * The size of the buffer we use as scratch space for integer
  * conversions, among other things.  We need enough space to
- * write a uintmax_t in octal (plus one byte).
+ * write a uintmax_t in binary.
```
bsdimp, line 292: Don't you need +2 for the 0b?

bsdimp (changes requested):
I think that the 0b that's prented isn't accounted for, but isn't now.

lib/libc/stdio/vfprintf.c
VexedUXR: Oh, I didn't think of that, but it looks like we print it separately:

https://github.com/freebsd/freebsd-src/blob/6e414739fc95663836a5a10a8cf13e0211fd5efc/lib/libc/stdio/vfprintf.c#L1022

https://github.com/freebsd/freebsd-src/blob/6e414739fc95663836a5a10a8cf13e0211fd5efc/lib/libc/stdio/vfwprintf.c#L1077

bsdimp: Then don't we need +1 for NUL?

VexedUXR: As far as I can tell, we don't.

In the case of %c, we write to it manually (or in the case of a wide char through `wcrtomb`). Otherwise we pass it to `__ujtoa` or `__ultoa` which both stop when they're done writing the number.
After that we just print with `size` which is calculated using `buf + BUF - cp` where `cp` is `buf` plus some offset.

When calling `grouping_print`, to print a number with separators, we pass a start pointer and end pointer (`cp` and `buf+BUF`), which are used to call `io_print` (through `io_printandpad`) with the correct size.

Overall the code looks like it doesn't expect a null terminator. However, I think that the comment I removed was hinting that we did, otherwise I'm not sure what that +1 would be for. Unless there's some edge case I'm missing...

FWIW, the patch also survived the cirrus-ci boot smoke test.

PR #1401, opened 2024-09-02
Backport the security patch of CVE-2024-39894

Crispy-fried-chicken:
Hi,
 we find the recurring vulnerability which shares the similarity of CVE-2024-39894 in you project with the version of release/14.0.0 and release/13.3.0. The patch of this CVE is https://github.com/openssh/openssh-portable/commit/146c420d29d055cc75c8606327a1cf8439fe3a08 and we find that you've already fix it in the main branch, which is the commit https://github.com/freebsd/freebsd-src/commit/b81424adf7181d816c10b1345aaa3305ab0ec304. Is there a need to backport it?

### DIFF
--- a/crypto/openssh/clientloop.c
+++ b/crypto/openssh/clientloop.c
@@ -607,8 +607,9 @@ obfuscate_keystroke_timing(struct ssh *ssh, struct timespec *timeout,
 		if (timespeccmp(&now, &chaff_until, >=)) {
 			/* Stop if there have been no keystrokes for a while */
 			stop_reason = "chaff time expired";
-		} else if (timespeccmp(&now, &next_interval, >=)) {
-			/* Otherwise if we were due to send, then send chaff */
+		} else if (timespeccmp(&now, &next_interval, >=) &&
+		    !ssh_packet_have_data_to_write(ssh)) {
+			/* If due to send but have no data, then send chaff */
 			if (send_chaff(ssh))
 				nchaff++;
 		}


jlduran:
It's already in [stable/14](https://github.com/freebsd/freebsd-src/commit/bf9a275b24f6655616cc691555fe1a36ed5e4338), and in [stable/13](https://github.com/freebsd/freebsd-src/commit/79853e40abd8a63bd53cbf4059032d7143c69300). Patches to `releng` branches must go through the Security Officer.

jlduran:
I can answer the easy part:

This pull request cannot be accepted as is. If you consider that this bug[^1] should have received a Security Advisory, then you should follow this procedure: https://www.freebsd.org/security/.

[^1]: **Logic error in ssh(1) `ObscureKeystrokeTiming`**

    In OpenSSH version 9.5 through 9.7 (inclusive), when connected to an OpenSSH server version 9.5 or later, a logic error in the ssh(1) `ObscureKeystrokeTiming` feature (on by default) rendered this feature ineffective - a passive observer could still detect which network packets contained real keystrokes when the countermeasure was active because both fake and real keystroke packets were being sent unconditionally.

    This bug was found by Philippos Giavridis and also independently by Jacky Wei En Kung, Daniel Hugenroth and Alastair Beresford of the University of Cambridge Computer Lab.

    Worse, the unconditional sending of both fake and real keystroke packets broke another long-standing timing attack mitigation. Since OpenSSH 2.9.9 sshd(8) has sent fake keystoke echo packets for traffic received on TTYs in echo-off mode, such as when entering a password into su(8) or sudo(8). This bug rendered these fake keystroke echoes ineffective and could allow a passive observer of a SSH session to once again detect when echo was off and obtain fairly limited timing information about keystrokes in this situation (20ms granularity by default).

    This additional implication of the bug was identified by Jacky Wei En Kung, Daniel Hugenroth and Alastair Beresford and we thank them for their detailed analysis.

    This bug does not affect connections when ObscureKeystrokeTiming was disabled or sessions where no TTY was requested.

bsdimp:
This is in all stable branches, and secteam didn't think we needed a SA for this for 14.0 (which is almost out of support)

PR #1402, opened 2024-09-03
mfiutil: Fix incorrect accesses of an union member in 'struct mfi_evt_detail'

Low-power:
PR:		[281155](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=281155)
PR:		[281156](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=281156)

From my portable mfiutil project, commit <https://sourceforge.net/p/mfiutil/code/ci/d4d66777e4685813cdc501669275bb7853a32088/>.


delphij (approved):
Looks reasonable to me.

PR #1405, opened 2024-09-03
mfiutil: Fix unsafe assumptions of snprintf(3) return value in function `mfi_autolearn_period`

Low-power:
PR:		[281160](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=281160)

From my portable mfiutil project, commit <https://sourceforge.net/p/mfiutil/code/ci/8ee7bd9112fab7bdf28188fa595f8a8a59cced48/>.

### DIFF
--- a/usr.sbin/mfiutil/mfi_bbu.c
+++ b/usr.sbin/mfiutil/mfi_bbu.c
@@ -50,10 +50,21 @@ mfi_autolearn_period(uint32_t period, char *buf, size_t sz)
 
 	tmp = buf;
 	if (d != 0) {
-		tmp += snprintf(buf, sz, "%u day%s", d, d == 1 ? "" : "s");
+		int fmt_len;
+		fmt_len = snprintf(buf, sz, "%u day%s", d, d == 1 ? "" : "s");
+		if (fmt_len < 0) {
+			*buf = 0;
+			return;
+		}
+		if ((size_t)fmt_len >= sz) return;
+		tmp += fmt_len;
 		sz -= tmp - buf;
 		if (h != 0) {
-			tmp += snprintf(tmp, sz, ", ");
+			fmt_len = snprintf(tmp, sz, ", ");
+			if (fmt_len < 0 || (size_t)fmt_len >= sz) {
+				return;
+			}
+			tmp += fmt_len;
 			sz -= 2;
 		}
 	}


usr.sbin/mfiutil/mfi_bbu.c
```diff
@@ -50,10 +50,21 @@ mfi_autolearn_period(uint32_t period, char *buf, size_t sz)
...
+			*buf = 0;
+			return;
+		}
+		if ((size_t)fmt_len >= sz) return;
+		tmp += fmt_len;
 		sz -= tmp - buf;
 		if (h != 0) {
-			tmp += snprintf(tmp, sz, ", ");
+			fmt_len = snprintf(tmp, sz, ", ");
+			if (fmt_len < 0 || (size_t)fmt_len >= sz) {
```
delphij, line 64: or maybe just `strlcat()`?  I don't see a need to use snprintf here...

usr.sbin/mfiutil/mfi_bbu.c
```diff
@@ -50,10 +50,21 @@ mfi_autolearn_period(uint32_t period, char *buf, size_t sz)
 
 	tmp = buf;
 	if (d != 0) {
-		tmp += snprintf(buf, sz, "%u day%s", d, d == 1 ? "" : "s");
+		int fmt_len;
+		fmt_len = snprintf(buf, sz, "%u day%s", d, d == 1 ? "" : "s");
```
delphij, line 54: The proposed change won't hurt but I don't think `snprintf` can return a negative number here, no?

Low-power:
> The proposed change won't hurt but I don't think `snprintf` can return a negative number here, no?

I don't think so either. Just pedantic, since I need to cast `fmt_len` into `size_t` to compare it with `sz`; better to make sure it isn't negative first.

On other hand, the [man page](https://man.freebsd.org/cgi/man.cgi?query=snprintf&apropos=0&sektion=0&manpath=FreeBSD+14.1-RELEASE&arch=default&format=html) did mentioned a few error conditions for this function. It is theoretically possible for **snprintf(3)** to fail due to an internal buffer allocation failure in the C library.

Low-power:
**strlcpy(3)** should be used. Then the checking for negative number can be eliminated because it returns `size_t`.

I didn't use these functions in my fork because they are not portable.

usr.sbin/mfiutil/mfi_bbu.c
bsdimp: It returns the number of characters that would be written if there was an unlimited buffer.  It might be > sz.
However, it can return < 0 if there's an error in formatting (well, the man page just says "an error")

usr.sbin/mfiutil/mfi_bbu.c
bsdimp: agree'd

bsdimp:

PR #1407, opened 2024-09-03
 Reduce additional console output when muted

RedcomOSS:
While trying to resolve some custom installer issues, we found that using conscontrol(8) or setting kern.always_console_output=0 in sysctl.conf(5) did not always prevent console output. This is in part because some areas of the kernel were outputting to the console device without checking the status of the setting. These changes enforce checking of the value in both locations where console output occurs from kernel and init(8) based callouts.

Details on changes:

- Moves check for mute to earlier in sequence to silence kernel output even if EARLY_PRINTF is defined.
- Modifies call prf_putbuf() and prf_putchar() in subr_prf.c to strip TOCONS flag if muting is enabled, to honor the setting at print level.

This is a rather simple change, which increases areas where flags to silence console output are honored.  We have been running this change since 10/23 in-house without issue (Patching prior to 14.0 also required making cn_mute non-static.)

Signed-off-by: Craig.Woodward@redcom.com

### DIFF
--- a/sys/kern/kern_cons.c
+++ b/sys/kern/kern_cons.c
@@ -525,6 +525,9 @@ cnputc(int c)
 	struct consdev *cn;
 	const char *cp;
 
+	if (cn_mute || c == '\0')
+		return;
+
 #ifdef EARLY_PRINTF
 	if (early_putc != NULL) {
 		if (c == '\n')
@@ -534,8 +537,6 @@ cnputc(int c)
 	}
 #endif
 
-	if (cn_mute || c == '\0')
-		return;
 	STAILQ_FOREACH(cnd, &cn_devlist, cnd_next) {
 		cn = cnd->cnd_cn;
 		if (!kdb_active || !(cn->cn_flags & CN_FLAG_NODEBUG)) {
--- a/sys/kern/subr_prf.c
+++ b/sys/kern/subr_prf.c
@@ -112,6 +112,7 @@ struct snprintf_arg {
 };
 
 extern	int log_open;
+extern	int cn_mute;
 
 static void  msglogchar(int c, int pri);
 static void  msglogstr(char *str, int pri, int filter_cr);
@@ -425,7 +426,7 @@ prf_putchar(int c, int flags, int pri)
 		msgbuftrigger = 1;
 	}
 
-	if (flags & TOCONS) {
+	if ((flags & TOCONS) && !cn_mute) {
 		if ((!KERNEL_PANICKED()) && (constty != NULL))
 			msgbuf_addchar(&consmsgbuf, c);
 
@@ -443,7 +444,7 @@ prf_putbuf(char *bufr, int flags, int pri)
 		msgbuftrigger = 1;
 	}
 
-	if (flags & TOCONS) {
+	if ((flags & TOCONS) && !cn_mute) {
 		if ((!KERNEL_PANICKED()) && (constty != NULL))
 			msgbuf_addstr(&consmsgbuf, -1,
 			    bufr, /*filter_cr*/ 0);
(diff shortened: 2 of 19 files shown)


sys/kern/subr_prf.c
```diff
@@ -425,7 +426,7 @@ prf_putchar(int c, int flags, int pri)
 		msgbuftrigger = 1;
 	}
 
-	if (flags & TOCONS) {
+	if ((flags & TOCONS) && (!cn_mute)){
```
delphij, line 429: style: blank space between `)` and `{`

sys/kern/subr_prf.c
```diff
@@ -443,7 +444,7 @@ prf_putbuf(char *bufr, int flags, int pri)
...
```
delphij, line 447: style: blank space between `)` and `{`

jlduran:
In the commit message, `sysctrl.conf(5)` -> `sysctl.conf(5)`.

RedcomOSS:
Missed the code changes.  I'm a new to the process of pull requests here. Do I need to close this and open a new/clean pull request? The push/merge seems to have gone fine, but it's blocked on compute time limits?

RedcomOSS:
> It is likely that a final rebase/squash on your commits is requested (or the committer may have an automated script that does it for you).

Fine with a squash. I repeated the commit in the second update (with requested correction), for easier squashing.

> Yes, usually the credits run out within the first few days of the month.

The new delta is literally white-space fixes, so compile should be no different.  That would be up to the committer I suppose, but seems unneeded for this delta if they're a scarce resource.

> Is it easy to add a before/after output? Or a way to easily reproduce the differences?

If you mean a clean diff from main to this (post fixes) then yes.  I can generate it from our base tree vs our last merge point.  But doing a checkout/patch/commit would remove the references already in this pull, would it not?

If you mean a test-case for output this patch would block?  That's harder.

We found it when an app was trying to run when a library it wanted was not available.  We could easily catch the error and setup a workaround.  The problem was that the ***console output*** would show the library load failure, even when `cn_mute` was set, messing up the ASCII status screen. We tracked it to this and patched accordingly.

This area of the code has seen little change in the base tree for many years.  The only "conflict" in pulling up to 14.0 was `cn_mute` had changed to non-static, when it was added to the sysctl(8) list.  We had it non-static for this patch since 13.X-ish.  It complicated the last merge, (potential reverse patch detected) which made the argument to upstream changes easier.

sys/kern/kern_cons.c
```diff
@@ -525,6 +525,9 @@ cnputc(int c)
 	struct consdev *cn;
 	const char *cp;
 
+	if (cn_mute || c == '\0')
```
bsdimp, line 528: why add check for NUL? Aren't NULs valid output sometimes?

sys/kern/subr_prf.c
```diff
@@ -443,7 +444,7 @@ prf_putbuf(char *bufr, int flags, int pri)
 		msgbuftrigger = 1;
 	}
 
-	if (flags & TOCONS) {
+	if ((flags & TOCONS) && (!cn_mute)) {
 		if ((!KERNEL_PANICKED()) && (constty != NULL))
 			msgbuf_addstr(&consmsgbuf, -1,
 			    bufr, /*filter_cr*/ 0);
```
bsdimp, line 450: So why here, and before, do you not add it do msgbuf?

sys/kern/kern_cons.c
RedcomOSS: I did not add the check for null, it was already there.  I added the  `cm_mute` flag check only.

sys/kern/subr_prf.c
RedcomOSS: It already exists there, but some things bypass calls, and call this directly.  (Specifically other kernel modules, including the ELF loader and a few other places.)  Placing the check here, at the lowest level, guarantees that anything calling for console output will be intercepted.

sys/kern/kern_cons.c
bsdimp: OK. that makes sense to me then.

RedcomOSS:
Since this has been marked "ready", is there anything more I need to do to get this into a main tree?  I'm not super familiar with the process here, so just want to make sure I'm not missing a step to get changes merged in.

bsdimp:
I've been too busy this month to land things... I just need to make sure this boots (I'm 99% sure it will) and I'll commit it.

PR #1408, opened 2024-09-03
e6000sw: Add support for MV88E6190X switch variant

salekseev:
This commit adds support for MV88E6190X switch variant that is similar to MV88E6190 in its configuration.

It also adds is6190x hint support for enabling on amd64 platform without device tree support.

PR: [281211](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=281211)

### DIFF
--- a/sys/dev/etherswitch/e6000sw/e6000sw.c
+++ b/sys/dev/etherswitch/e6000sw/e6000sw.c
@@ -252,6 +252,10 @@ e6000sw_probe(device_t dev)
 	if (resource_int_value(device_get_name(sc->dev),
 	    device_get_unit(sc->dev), "addr", &sc->sw_addr) != 0)
 		return (ENXIO);
+
+	resource_int_value(device_get_name(sc->dev),
+		device_get_unit(sc->dev), "is6190x", &is_6190x);
+
 	if (resource_int_value(device_get_name(sc->dev),
 	    device_get_unit(sc->dev), "is6190", &is_6190) != 0)
 		/*
@@ -260,12 +264,12 @@ e6000sw_probe(device_t dev)
 		 */
 		resource_int_value(device_get_name(sc->dev),
 		    device_get_unit(sc->dev), "is8190", &is_6190);
-	resource_int_value(device_get_name(sc->dev),
-		device_get_unit(sc->dev), "is6190x", &is_6190x);
+
 	if (is_6190 !=0 && is_6190x != 0)
 		if (bootverbose)
 			device_printf(dev, "Cannot configure conflicting variants\n");
 		return (ENXIO);
+
 	if (is_6190 != 0)
 		sc->swid = MV88E6190;
 	else if (is_6190x != 0)
--- a/sys/dev/etherswitch/e6000sw/e6000sw.c
+++ b/sys/dev/etherswitch/e6000sw/e6000sw.c
@@ -217,6 +217,7 @@ e6000sw_probe(device_t dev)
 	phandle_t switch_node;
 #else
 	int is_6190;
+	int is_6190x;
 #endif
 
 	sc = device_get_softc(dev);
@@ -259,8 +260,16 @@ e6000sw_probe(device_t dev)
 		 */
 		resource_int_value(device_get_name(sc->dev),
 		    device_get_unit(sc->dev), "is8190", &is_6190);
+	resource_int_value(device_get_name(sc->dev),
+		device_get_unit(sc->dev), "is6190x", &is_6190x);
+	if (is_6190 !=0 && is_6190x != 0)
+		if (bootverbose)
+			device_printf(dev, "Cannot configure conflicting variants\n");
+		return (ENXIO);
 	if (is_6190 != 0)
 		sc->swid = MV88E6190;
+	else if (is_6190x != 0)
+		sc->swid = MV88E6190X;
 #endif
 	if (sc->sw_addr < 0 || sc->sw_addr > 32)
 		return (ENXIO);
@@ -302,6 +311,10 @@ e6000sw_probe(device_t dev)
 		description = "Marvell 88E6190";
 		sc->num_ports = 11;
 		break;
+	case MV88E6190X:
+		description = "Marvell 88E6190X";
+		sc->num_ports = 11;
+		break;
 	default:
 		device_printf(dev, "Unrecognized device, id 0x%x.\n", sc->swid);
 		return (ENXIO);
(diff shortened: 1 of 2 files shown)


sys/dev/etherswitch/e6000sw/e6000sw.c
```diff
@@ -261,6 +262,8 @@ e6000sw_probe(device_t dev)
 		    device_get_unit(sc->dev), "is8190", &is_6190);
 	if (is_6190 != 0)
 		sc->swid = MV88E6190;
+	if (is_6190x != 0)
```
rilysh, line 265: Where is is_6190x variable initialized? Searching "is_6190x" only gave me two results

salekseev: I am not a huge fan of increasing number of `if` statements to support additional switch variants but I feel it's a good tradeoff to keep the change small and backwards compatible.

sys/dev/etherswitch/e6000sw/e6000sw.c
```diff
@@ -252,15 +252,27 @@ e6000sw_probe(device_t dev)
...
+		    device_get_unit(sc->dev), "is6190x", &is_6190x)
+		}
+	if (is_6190 !=0 && is_6190x != 0)
+	    if (bootverbose)
+	        device_printf(dev, "Cannot configure conflicting variants\n");
+	    return (ENXIO);
 	if (is_6190 != 0)
 		sc->swid = MV88E6190;
+	if (is_6190 != 0)
+		sc->swid = MV88E6190X;
```
rilysh, line 275: This part seems wrong
Because both can't be available (as it creates a conflict) but on the next line, it's checking if is_6190 != 0 two times, one time to set MV88E6190 and another time to set MV88E6190X. At the end it will set swid value to MV88E6190X, even if MV88E6190 (is_6190 != 0) is only available.

salekseev (commented):
@rilysh do you think [this ](https://github.com/freebsd/freebsd-src/blob/d359a4349a21ae74fb06b42d53797b059c017fe1/sys/dev/etherswitch/e6000sw/e6000sw.c#L255-L264) is the best approach to check for multiple device hints being set and to initalize variables if set?

sys/dev/etherswitch/e6000sw/e6000sw.c
```diff
@@ -259,8 +260,16 @@ e6000sw_probe(device_t dev)
 		 */
 		resource_int_value(device_get_name(sc->dev),
 		    device_get_unit(sc->dev), "is8190", &is_6190);
+	(void) resource_int_value(device_get_name(sc->dev),
```
rilysh, line 263: Return value casting to void isn't necessary, unless it's explicitly has a `[[nodiscard]]` attribute.

```
-	(void) resource_int_value(device_get_name(sc->dev),
-		device_get_unit(sc->dev), "is6190x", &is_6190x);
-	if (is_6190 !=0 && is_6190x != 0)
+
+	resource_int_value(device_get_name(sc->dev),
+	    device_get_unit(sc->dev), "is6190x", &is_6190x);
+	if (is_6190 != 0 && is_6190x != 0)
```

Since both can't be available (according to it), there's no need to check for `is_6190x` if `is_6190` is true.
```
 	if (is_6190 != 0)
 		sc->swid = MV88E6190;
-	if (is_6190x != 0)
+	else if (is_6190x != 0)
 		sc->swid = MV88E6190X;
```

salekseev: Thanks a lot for your review. I've made changes based on your suggestions.

sys/dev/etherswitch/e6000sw/e6000sw.c
```diff
@@ -260,8 +260,9 @@ e6000sw_probe(device_t dev)
 		 */
 		resource_int_value(device_get_name(sc->dev),
 		    device_get_unit(sc->dev), "is8190", &is_6190);
-	resource_int_value(device_get_name(sc->dev),
-		device_get_unit(sc->dev), "is6190x", &is_6190x);
+	if (resource_int_value(device_get_name(sc->dev),
+		device_get_unit(sc->dev), "is6190x", &is_6190x) != 0)
+		return;
```
salekseev, line 265: THis should avoids the following error: `error: misleading indentation; statement is not part of the previous 'if' [-Werror,-Wmisleading-indentation]`

sys/dev/etherswitch/e6000sw/e6000sw.c
```diff
@@ -260,8 +260,9 @@ e6000sw_probe(device_t dev)
 		 */
 		resource_int_value(device_get_name(sc->dev),
 		    device_get_unit(sc->dev), "is8190", &is_6190);
-	resource_int_value(device_get_name(sc->dev),
-		device_get_unit(sc->dev), "is6190x", &is_6190x);
+	if (resource_int_value(device_get_name(sc->dev),
```
rilysh, line 263: If "is6190" is available, but "is6190x" isn't, it will just stop the execution. I'm not exactly sure if MV88E6190 and MV88E6190X are related to each other (without one either one can't work).

Also, return isn't returning any meaningful values, and the function expects an integer.

bsdimp:
rylish's comments need to be addressed.

erikarn:
hi! This has been landed in -HEAD now, thanks!

sys/dev/etherswitch/e6000sw/e6000sw.c
```diff
@@ -259,8 +264,16 @@ e6000sw_probe(device_t dev)
 		 */
 		resource_int_value(device_get_name(sc->dev),
 		    device_get_unit(sc->dev), "is8190", &is_6190);
+
+	if (is_6190 !=0 && is_6190x != 0)
+		if (bootverbose)
+			device_printf(dev, "Cannot configure conflicting variants\n");
+		return (ENXIO);
```
jrtc27, line 271: Given the lack of curly braces amounts to an unconditional `return (ENXIO)` I don't see how this could possibly have been tested.

jrtc27:
There are clear style issues in this patch (lacking space around `!=` and what look to be overly long lines). I don't know why GitHub actions didn't run checkstyle9.pl but please run it yourself locally (tools/build/checkstyle9.pl) and fix the issues that arise.

sys/dev/etherswitch/e6000sw/e6000sw.c
```diff
@@ -217,6 +217,7 @@ e6000sw_probe(device_t dev)
 	phandle_t switch_node;
 #else
 	int is_6190;
+	int is_6190x;
```
jrtc27, line 220: So, for is_6190 the code assumes one of the two properties is present (otherwise this is left uninitialised). That's a bit dodgy, but for is_6190x you definitely can't assume that otherwise you will break existing hints that predate the existence of is_6190x.

bsdimp:
> hi! This has been landed in -HEAD now, thanks!

Thanks!

PR #1409, opened 2024-09-04
ctl: fix uninitialized data used by PREVENT ALLOW MEDIUM REMOVAL

asomers:
This is not considered a security risk because the uninitialized data is never directly exposed to user land.  The worse case scenario is that START STOP UNIT might not work correctly.

Also, add ATF tests for START STOP UNIT and PREVENT ALLOW MEDIUM REMOVAL.

This was discussed in https://reviews.freebsd.org/D45952 , but not included as part of that review.

### DIFF
--- a/sys/cam/ctl/ctl.c
+++ b/sys/cam/ctl/ctl.c
@@ -4630,7 +4630,7 @@ fail:
 	ctl_tpc_lun_init(lun);
 	if (lun->flags & CTL_LUN_REMOVABLE) {
 		lun->prevent = malloc((CTL_MAX_INITIATORS + 31) / 32 * 4,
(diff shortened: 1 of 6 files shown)


markjdb (commented):
Have you run these tests under a `GENERIC-KMSAN` kernel? That should help verify that we're not accessing uninitialized data in an unsafe way.

asomers:
> Have you run these tests under a `GENERIC-KMSAN` kernel? That should help verify that we're not accessing uninitialized data in an unsafe way.

That's a good idea.  I hadn't, but I have now.  They pass.

emaste (approved):

markjdb (approved):

PR #1414, opened 2024-09-11
ps3disk.c: Rewrite ps3disk_transfer

aomsin2526:
This function is bugged since the beginning, but it never hit because its variable doesn't allow.

However, since commit a77e1f0f81df5fa3b4a6a38728ebace599cb18a4 it happen now.

First, it assume that ds_len will always equal to real user requested size. So it being used for sector count calculation.

This is no longer true, and will fail if attempt to read last few sectors.

Use bp->bio_length instead.

Second, this being a loop is pointless because nsegs will never be > 1 as specified at bus_dma_tag_create() call.

And all it doing is to repeat very same command again but with different ds_addr. Since bio_driver2 tag ident pointer are being reused, the result will be discarded at ps3disk_intr().

### DIFF
--- a/sys/powerpc/ps3/ps3disk.c
+++ b/sys/powerpc/ps3/ps3disk.c
@@ -588,9 +588,10 @@ ps3disk_transfer(void *arg, bus_dma_segment_t *segs, int nsegs, int error)
 	struct bio *bp = (struct bio *)(arg);
 	struct ps3disk_softc *sc = (struct ps3disk_softc *)bp->bio_disk->d_drv1;
 	struct ps3disk_region *rp = &sc->sc_reg[bp->bio_disk->d_unit];
+	bus_dma_segment_t *seg = &segs[0];
 	uint64_t devid = ps3bus_get_device(sc->sc_dev);
-	uint64_t block;
-	int i, err;
+	uint64_t block, bio_length, sector_op_count;
+	int err;
 
 	/* Locks already held by busdma */
 	PS3DISK_ASSERT_LOCKED(sc);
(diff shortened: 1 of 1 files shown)


pkubaj:
While I'm not sure about the contents of the patch, permissions seem to change from 644 to 755. C source files should definitely be 644.

pkubaj:
Same as https://github.com/freebsd/freebsd-src/pull/1415
Permissions are changed to 755, that shouldn't be the case.

aomsin2526:
Fixed, thanks.

sys/powerpc/ps3/ps3disk.c
```diff
@@ -588,13 +588,17 @@ ps3disk_transfer(void *arg, bus_dma_segment_t *segs, int nsegs, int error)
...
 
 	/* Locks already held by busdma */
 	PS3DISK_ASSERT_LOCKED(sc);
 
+	KASSERT(error == 0,
+		    ("DMA Error %d", error));
+
 	if (error) {
 		bp->bio_error = error;
 		bp->bio_flags |= BIO_ERROR;
```
amotin, line 604: Does the KASSERT() before normal error recovery makes some sense, or it is a leak of debug code?

aomsin2526: Latter, to me, this error should never hit at anytime.

should i remove?

bsdimp:
This may be ready, but I think @amotin 's comments need a response at the very least.

bsdimp:
I think at the least, we should remove the assert since we have code to handle when it fails and we usually do one or the other.

sys/powerpc/ps3/ps3disk.c
aomsin2526: I'm fine removing that.

aomsin2526:
Should be good now

bsdimp:
my automation messed up, so I couldn't push this to close it as merged.

PR #1415, opened 2024-09-11
powerpc/busdma_machdep.c: Limit sgsize to buflen

aomsin2526:
Before commit a77e1f0f81df5fa3b4a6a38728ebace599cb18a4, total sum of ds_len will always equal to user requested size (buflen), so user assume that this will always be true and used it as reference. (example https://github.com/freebsd/freebsd-src/pull/1414)

Now since this is no longer true, it causes severe memory corruption.

By simply limit sgsize of last segment to buflen will fix it.

### DIFF
--- a/sys/powerpc/powerpc/busdma_machdep.c
+++ b/sys/powerpc/powerpc/busdma_machdep.c
@@ -648,6 +648,7 @@ _bus_dmamap_load_buffer(bus_dma_tag_t dmat,
 		sgsize = MIN(buflen, PAGE_SIZE - (curaddr & PAGE_MASK));
 		if (map->pagesneeded != 0 && must_bounce(dmat, curaddr)) {
 			sgsize = roundup2(sgsize, dmat->alignment);
+			sgsize = MIN(sgsize, buflen);
 			curaddr = add_bounce_page(dmat, map, kvaddr, curaddr,
 			    sgsize);
 		}


pkubaj:
Same as https://github.com/freebsd/freebsd-src/pull/1414
Permissions are changed to 755, that shouldn't be the case.

aomsin2526:
Thanks.

mhorne:
> @mhorne , this is maybe fallout from your earlier commit, and if it is correct it is probably applicable to other architectures such as RISC-V and arm64 as well.

Indeed, I have a piece of RISC-V hardware that exhibits the problem and requires the same fix. (As opposed to the hardware that required the change at fault, https://github.com/freebsd/freebsd-src/commit/a77e1f0f81df5fa3b4a6a38728ebace599cb18a4).

I require a little more time to be sure of what exactly is going wrong, but rest assured I am looking into it and will be able to take action on this PR soon.

mhorne:
FYI: https://reviews.freebsd.org/D47807

bsdimp:
@mhorne This looks like it's a good change, but you've also made it elsewhere... what should I do?

mhorne:
> @mhorne This looks like it's a good change, but you've also made it elsewhere... what should I do?

Thanks for the reminder. We should definitely proceed, so I will commit my expanded version of the change by this Friday. Any further refinements can wait.

https://reviews.freebsd.org/D47807

mhorne:
Finally merged! Thanks @aomsin2526 for your submission and patience.

PR #1416, opened 2024-09-11
Fix gets x

njrizzo:
I found strange behavior when I used gets_s function.
When I type characters less or more than N - 1, this function work fine, but when I type exactly N - 1 characters, this functions report a strange result. My proposal for correcting the error fix these issues.
Below I show the result of test with this code in C.

```
#include <stdio.h>
#include <stdlib.h>

int
main(void)
{
        char a1[MAX_STR_LEN];
        char a2[MAX_STR_LEN];
        unsigned int c;

        // Clean the buffer
        for ( c = 0; c < MAX_STR_LEN; c++)
        {
                          a1[c] = '\0';
                          a2[c] = '\0';
        }

        printf("Type a letter:\n");
        gets_s(a1, 5);
        printf("Type a letter:\n");
        gets_s(a2, 5);
        printf("You typed the letter:\na1 = %s\na2 = %s\n",a1, a2);
        return 0;
}
```

Using original gets_s function
         
% ./gets_s 
Type a letter:
aa
Type a letter:
aaaaa
You typed the letter: 
a1 = aa
a2 = aaaa
% ./gets_s
**Type a letter:
aaaa**

Type a letter:
aaaaa
You typed the letter: 
a1 = aaaa
a2 = aaaa

Using new fix gets_s function
% ./fixgets_s
Type a letter:
aa
Type a letter:
aaaaa
You typed the letter: 
a1 = aa
a2 = aaaa
% ./fixgets_s
Type a letter:
aaaa
Type a letter:
aaaaa
You typed the letter: 
a1 = aaaa
a2 = aaaa

### DIFF
--- a/lib/libc/stdio/gets_s.c
+++ b/lib/libc/stdio/gets_s.c
@@ -87,9 +87,11 @@ _gets_s(char *buf, rsize_t n)
 				}
 		}
 	}
-
 	*s = 0;
-	return (buf);
+	if (n == 0)
+		return (NULL);
+	else
+		return (buf);
 }
 
 /* ISO/IEC 9899:2011 K.3.7.4.1 */
--- a/lib/libc/stdio/gets_s.c
+++ b/lib/libc/stdio/gets_s.c
@@ -54,7 +54,7 @@ _gets_s(char *buf, rsize_t n)
    	 */
 
 	/* This prevents various alert via __throw_constraint_handler_s() if 
- 	 * the user type more than N - 1 characters, when functions flush stdin
+ 	 * the user type more than N - 1 characters, when function flush stdin
   	 * buffer
    	 */
 	signal = 1;
--- a/lib/libc/stdio/gets_s.c
+++ b/lib/libc/stdio/gets_s.c
@@ -49,11 +49,13 @@ _gets_s(char *buf, rsize_t n)
 {
 	int c, signal;
 	char *s;
-	/* This modifications prevents error qhen the user type exactly N - 1
+	/* 
+ 	 * This modifications prevents error qhen the user type exactly N - 1
  	 * Character in stdin buffer and press  <ENTER> key ('\n')
    	 */
 
-	/* This prevents various alert via __throw_constraint_handler_s() if 
+	/* 
+ 	 * This prevents various alert via __throw_constraint_handler_s() if 
  	 * the user type more than N - 1 characters, when function flush stdin
   	 * buffer
    	 */
@@ -66,11 +68,15 @@ _gets_s(char *buf, rsize_t n)
 			} else
 				break;
 		} else {
-			if ( n ) {
+			/*
+   			 * This prevents strange behavior when user types EXATLY N - 1
+			 * character into the stdin buffer
+    			*/
+			if ( n ) {	// if n > 0
 				*s++ = c;
 				n --;
 			} else 
-				if ( signal ) {
+				if ( signal == 1 ) {
 				/*
  	 			* If end of buffer reached, discard until \n or eof.
 	 			* Then throw an error.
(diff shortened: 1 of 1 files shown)


lib/libc/stdio/gets_s.c
```diff
@@ -47,31 +47,47 @@
 static inline char *
 _gets_s(char *buf, rsize_t n)
 {
-	int c;
+	int c, signal;
 	char *s;
+	/* 
+ 	 * This modifications prevents error qhen the user type exactly N - 1
```
gperciva, line 53: Should that `when` be `when`?

njrizzo: Yes, typo wrong

njrizzo: This is the final version, but I edited it in the GitHub editor and some things are wrong, like tabs and spaces.

rbgarga:
I followed @njrizzo first report on brazillian FreeBSD Telegram group.  The problem he found is 100% reproducible and happens every time string length matches size parameter of `get_s()`.  The following code can be used to reproduce it:

```C
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

#define MAX_STR_LEN     4

int
main(void) {
        char a[MAX_STR_LEN];

        printf("Type something:\n");
        gets_s(a, MAX_STR_LEN);
        printf("You typed: %s\n",a);
        return 0;
}
```

If you type `abc` when prompted, `a` will contain `abc\n` and it will match `MAX_STR_LEN`.  On  current `gets_s()` implementation it will leave `for()` statement with empty `stdin` buffer and `n == 0` and when it happens it's going to wait for another `\n` or `EOF` on that `while ((c = __sgetc(stdin)) != '\n' && c != EOF);` statement so user will need to press a second `<ENTER>`

```
Type something:
abc
          <-- Here is where second \n is required
You typed: abc```

**Comment by rbgarga** [comment]
@cschuber you are the original author of this code so maybe you can be interested in this report

**Comment by gperciva** [comment]
@rbgarga:
> The problem he found is 100% reproducible and happens every time string length matches `gets_s()` size parameter. 

That's how `gets_s` is supposed to work:

```
$ man gets_s
...
The fgets() function reads at most one less than the number of characters
specified by size from the given stream and stores them in the string
str.
...
The gets_s() function is equivalent to fgets()...
```

With your `MAX_STR_LEN 4`, the function will read at most 3 characters.  If you type `abc\n`, the `\n` is not seen by that `gets_s` call.

A quick check of the C11 standard is more explicit, in K.3.5.4.1:
```
char *gets_s(char *s, rsize_t n);

Runtime-constraints
...
A new-line character, end-of-file, or read error shall occur within reading
n-1 characters from stdin.
```

In the `abc\n` example, there is no new-line character in the first 3 characters of stdin, so the runtime constraint was violated.

If you add
```
set_constraint_handler_s(abort_handler_s);
```
before the `get_s()` call and enter `abc\n`, you'll get an abort trap.  If you ignore the constraint violation (which is unfortunately the default behaviour), then it's not surprising that `*s` doesn't match what you expect.

Also, note that you said that `a` would contain `abc\n`.  That's not the case; you can check that by replacing the final `printf()` with `for` loop which prints the contents of `a` character-by-character.  There's no newline in there.

rbgarga:
> Also, remember to check for error values!
> 
> Even if we don't check the runtime constraints, you should check
> 
> ```
> if (gets_s(...) == NULL)
>         printf("Error occurred\n");
> ```
> 
> The man page specifies:
> 
> ```
> RETURN VALUES
> ...
> If an error occurs, they return NULL and the buffer contents are indeterminate.
> ```

thanks for all details.  In the end it's not a bug, but a feature ;)

I believe this PR can be closed then

PR #1417, opened 2024-09-11
man: Fix unknown macros

gperciva:


gperciva:
This fixes a few mdoc errors in the man pages.  (Github note: we could add the `man-only` label).

First-time contributor.  I'm confident about the file changes; less confident about the commit message (too verbose?) and metadata at the bottom of the commit message.  Happy to revise as requested!

gperciva:
(Hmm, I thought that force-pushing would remove my original commit entirely.)

The change: my original commit included changes to `usr.sbin/ntp/doc/ntp.conf.5`, but I later noticed that the file included "EDIT THIS FILE WITH CAUTION  (ntp.mdoc)... It has been AutoGen-ed...".

I'm not un-interested in fixing the mdoc-generation template, but that's definitely outside of the scope of this PR.

lib/libsys/statfs.2
```diff
@@ -232,7 +232,7 @@ The fields in the
 .Vt statfs
 structure have been defined to provide the parameters relevant for
 traditional
-.Tm UNIX
+.Tn UNIX
```
concussious, line 235: Tn is deprecated and doesn't do anything, so I usually remove it and join the lines.

concussious (commented):
I usually say "manuals:" in commit message title to disambiguate between man(1). These are all macro typos, so if it was me I'd say "manuals: misc macro typos".

concussious (approved):
Thanks, LGTM! Some comitters have automated tooling that generates the PR and reviewers and then they have to manually remove them, so... there's that.

gperciva:
@concussious Thanks for the review!

As for ntp, I'm stumped trying to build the documentation after I've made fixes -- I think the build system is looking for `mdoc2mdoc`, but AutoGen doesn't supply that.  Anyway, that's very much outside the scope of this PR; so future discussion should go to:
https://bugs.ntp.org/show_bug.cgi?id=3947

bsdimp:
Pushed.

PR #1418, opened 2024-09-12
Remove stray whitespaces from sys/amd64/

MegaManSec:
This commit removes some stray whitespaces from sys/amd64/.

### DIFF
--- a/sbin/ggate/ggatec/ggatec.8
+++ b/sbin/ggate/ggatec/ggatec.8
@@ -34,6 +34,7 @@
 .Op Fl n
 .Op Fl v
 .Op Fl o Cm ro | wo | rw
+.Op Fl o Cm direct
 .Op Fl p Ar port
 .Op Fl q Ar queue_size
 .Op Fl R Ar rcvbuf
@@ -48,6 +49,7 @@
 .Op Fl n
 .Op Fl v
 .Op Fl o Cm ro | wo | rw
+.Op Fl o Cm direct
 .Op Fl p Ar port
 .Op Fl R Ar rcvbuf
 .Op Fl S Ar sndbuf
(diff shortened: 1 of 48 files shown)


gmshake:
@bsdimp  I also have some WIP style fixes. I'm wondering what is the best practices to do these. I had ever made some guideline within my team, and hand crafted CI pipelines to enforce the new code ( yes only the new code ) to follow the code style. I'd personally prefer onetime fixing for all style issues, but FreeBSD's code base is too large and I guess that is not feasible.

What's your opinion?

bsdimp:
> @bsdimp I also have some WIP style fixes. I'm wondering what is the best practices to do these. I had ever made some guideline within my team, and hand crafted CI pipelines to enforce the new code ( yes only the new code ) to follow the code style. I'd personally prefer onetime fixing for all style issues, but FreeBSD's code base is too large and I guess that is not feasible.
> 
> What's your opinion?

This is a tricky issue...  There's dozens of style things that need fixing, but they are minor enough we've not paid the cost to do so.

A one-time fixup is hard, because it @#$^@#^ up MFC  often given our usage patterns. If it was a script, and if that script were run on all supported branches, that might be OK. However, even then my attempts to do this have been, at best, more trouble than they are worth.

Changes like this are likely OK, but run afoul of being too big to properly verify by eye (though `git diff -w` I think could be used to assure ourselves that only trailing whitespace is removed. The only time it's a risk is if it's inside strings (though that's rare) or in code that's really a template (also rare). Both of these are mitigateable.

Enforcing the one true style is hard when there is no true, one true style. style(9) is a range. At best, we should make it an opt-in per file/directory since files with active maintianers that have a slight variation in style to whatever is selected as canonical will start to have their styles change from something consistent to something not. We've had a notion since forever that style should be flexible and as long as a file is consistent, it's fine.

Also, automatic style formatters collide badly with the CAM style of having things like '/* argfp.*/` in front of the args to describe them, and a careful wrapping of args. Much churn would happen in this code, for little benefit (if any).

So I'm a bit cold to this generally. But I in specific instances, I can see it working out well.

MegaManSec:
Just to give some context about this PR: I was doing some work on some files in this directory and kept coming across these stray whitespaces, so thought I would remove all that a quick grep could find. I don't expect that I'll be sending similar PRs to this one in the future.

bsdimp:
Normally, I wouldn't land something like this. We don't generally want whitespace changes via github, and generally eschew them except when other cleanup is happening. I was able to verify these, so I let it in, but I'll be less willing in the future to do so.

PR #1419, opened 2024-09-12
vmm: Correctly suspend and resume the vmm driver.

MegaManSec:
Previously, VMXON would be executed on a resume, contrary to proper initalization. The MSR lock may be lost on suspension, therefore must be taken again. Likewise, the VMX Enable bit may be cleared upon suspend, requiring to be re-set.

Concretely disable VMX on suspend, and re-enable it on resume.

Note: any IOMMU context will remain lost for any enabled vmm devices.

### DIFF
--- a/bin/sh/sh.1
+++ b/bin/sh/sh.1
@@ -1,4 +1,6 @@
 .\"-
+.\" SPDX-License-Identifier: BSD-3-Clause
+.\"
 .\" Copyright (c) 1991, 1993
 .\"	The Regents of the University of California.  All rights reserved.
 .\"
@@ -1074,7 +1076,6 @@ or the end of the
 command.
 .Ss Grouping Commands Together
 Commands may be grouped by writing either
-.Pp
 .Sm off
 .Bd -literal -offset indent
 .Po Ar list Pc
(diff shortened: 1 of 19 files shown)


bsdjhb (approved):
I think it might be good to be explicit that the contents of MSR_IA32_FEATURE_CONTROL need to be restored in resume (I think that's what "MSR lock" is referring to?) in the log message.

emaste:
> Note: any IOMMU context will remain lost for any enabled vmm devices.

Even if this isn't a new issue from your change, should we have this noted somewhere?

MegaManSec:
> I think it might be good to be explicit that the contents of MSR_IA32_FEATURE_CONTROL need to be restored in resume (I think that's what "MSR lock" is referring to?) in the log message.

Yes, MSR lock refers to `MSR_IA32_FEATURE_CONTROL` and specifically the `IA32_FEATURE_CONTROL_LOCK` bit.

MegaManSec:
> > Note: any IOMMU context will remain lost for any enabled vmm devices.
> 
> Even if this isn't a new issue from your change, should we have this noted somewhere?

Probably. FWIW, using vmm with a pci device on my system stalls the system completely upon its use after a resume (more details: https://joshua.hu/brcmfmac-bcm43602-suspension-shutdown-hanging-freeze-linux-freebsd-wifi-bug-pci-passthru) -- when I'm passthru'ing the device to a Linux VM, I have to "remove" the device inside the VM before a suspension, otherwise the system crashes (just completely freezes/stalls) after resume and the device is used again.

As far as I could tell when debugging a bit more lately (and I may have used the incorrect terminology about "IOMMU context"), the irq (at least) of the device is lost on resume, which results in the system freezing when the device is interacted with.

PR #1395, opened 2024-08-23
GSoC 2024 - Improving Installer with Repair and Upgrade Ability

leafoliage:
## Goals of the project

This project aims to improve the FreeBSD installer in three aspects: 
* Adding a repair menu: allows choosing existing partitions in a computer to mount and access for further inspection and repair
* Adding an upgrade menu: allows choosing existing systems to upgrade to newer version
* Providing pkg install support: allows pkg installation within live environment out of the box

### Repair Menu

I implemented a menu that looks for `freebsd-ufs` partitions and zfs pools. They are listed out so that users can choose which partition to mount. After choosing, users can decided whether they want to perform inspections, such as `fsck` or `mtree`, with another menu.

The repair menu can deal with systems that has ufs or zfs installed in the officially recommended way. For a ufs installed system, it mounts the selected partition at `/mnt`. For a zfs installed partition, it imports the selected pool at the mountpoint `/mnt`. In both cases, the selected inspection is run one by one after they're chosen.

### Upgrade Menu

This menu looks for partitions to upgrade. It utilizes the released files tarball and a pre-built etcupdate tree included in the installer to compare with the existing system and upgrade.

The upgrade menu and script prepares an etcupdate tree of the new release in advance and mount the selected partitions at `/mnt/system`. It compares `base.txz` and `kernel.txz` archive contents with `/mnt/system` directory contents to decide whether a file should be added, modified or removed, and install the changes accordingly. It then runs `etcupdate` with the pre-built tree to update etc files, and runs `pkg upgrade` to upgrade packages. 

### Pkg Installation Support

Network configuration and pkg bootstrapping is called upon exiting into live environment. A tmpfs is mounted under `/usr/local` where pkgs are extracted.

## Future Works

For the repair menu, there isn't a comprehensive way to detect existing freebsd partitions so that any customly installed system can be detected. Also, there are currently only 2 inspection options for ufs and zfs each, so any suggestion for inspecting partitions is encouraged.

For the upgrade menu, the script lacks a comprehensive ability to detect existing system. The upgrade process itself requires more testing to rule out any error it might cause, and its efficiency of installing modified changes should be improved.

For both parts, unit testing and end-to-end testing are still unsupported and encouraged to be implemented. 

## Useful Links

See more on the [project's wiki page](https://wiki.freebsd.org/SummerOfCode2024Projects/ImprovingRepairAbilityOfTheFreeBSDInstaller)

PR #1424, opened 2024-09-15
bsdinstall: Add pkg install support in live env

leafoliage:
This patch improves the support of pkg installation in the live environment. Currently, pkg has to be manually bootstrapped before it can install pkgs in the live env. Since the installation media is often used as a tool for inspecting and rescuing systems, it would be great if users can easily install tools in live env using pkg to their needs. 

This patch provides a bsddialog menu for pkg bootstrap after "Live System" option is selected from the installer menu. If "Yes", it goes on asking for network configurations, and bootstraps pkg under `/usr/local`.

This patch is part of the code of the [GSoC 2024 Project - Installer](https://wiki.freebsd.org/SummerOfCode2024Projects/ImprovingRepairAbilityOfTheFreeBSDInstaller), whose work is committed in PR #1395 . This patch is created for the ease of getting this project's work commited.

### DIFF
--- a/usr.sbin/bsdinstall/startbsdinstall
+++ b/usr.sbin/bsdinstall/startbsdinstall
@@ -102,6 +102,17 @@ $BSDDIALOG_OK)	# Install
 	fi
 	;;
 $BSDDIALOG_CANCEL)	# Live System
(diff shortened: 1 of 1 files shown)


grahamperrin:
Cross-reference FreeBSD bug [281518 – bsdinstall: improve pkg installation support in installation media live environment](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=281518)

PR #1425, opened 2024-09-15
inet_ntop returns const char, not char

RSilicon:


sys/libkern/inet_ntop.c
```diff
@@ -37,7 +37,7 @@ static char	*inet_ntop6(const u_char *src, char *dst, socklen_t size);
  * author:
  *	Paul Vixie, 1996.
  */
-char *
+const char *
 inet_ntop(int af, const void *src, char *dst, socklen_t size)
```
gmshake, line 41: `inet_ntop` actually return `pointer to presentation format address (`dst'), or NULL (see errno)`. So why mark it `const char *` ?

gmshake (changes requested):

rilysh (commented):
inet_ntop() function returns either [inet_ntop4()](https://github.com/freebsd/freebsd-src/blob/2a680c5c37ea020467fce545734afea019d3a3ea/sys/libkern/inet_ntop.c#L66) or [inet_ntop6()](https://github.com/freebsd/freebsd-src/blob/2a680c5c37ea020467fce545734afea019d3a3ea/sys/libkern/inet_ntop.c#L87) or NULL. Both  inet_ntop4() and inet_ntop6() returns `dst` pointer which isn't a `const char *`, so why marking it as `const` here?

Edit: Man page seem to refer `const char *` for inet_ntop().
https://man.freebsd.org/cgi/man.cgi?query=inet_ntop&apropos=0&sektion=0&manpath=FreeBSD+14.1-RELEASE+and+Ports&arch=default&format=html

sys/libkern/inet_ntop.c
bsdimp: Posix sez it returns const char *

sys/fs/nfsserver/nfs_nfsdserv.c
```diff
@@ -4085,10 +4085,10 @@ nfsrvd_setclientid(struct nfsrv_descript *nd, __unused int isdgram,
 			else 
 				(void) nfsm_strtom(nd, "udp6", 4);
 			rin6 = (struct sockaddr_in6 *)clp->lc_req.nr_nam;
-			ucp = inet_ntop(AF_INET6, &rin6->sin6_addr, addrbuf,
+			ucp = (u_char *)inet_ntop(AF_INET6, &rin6->sin6_addr, addrbuf,
```
bsdimp, line 4088: This is wrong. ucp and ucp2 should be const char *, which will convert without the casts.

bsdimp (changes requested):
Please make the suggested type change for ucp and ucp2.

sys/libkern/inet_ntop.c
gmshake: > Posix sez it returns const char *

Yes. Indeed.

@rilysh  So the commit message should mention this. That will be much clear.

rilysh: Yeah I just saw that too. It seems like man pages described it right (according to POSIX).

bsdimp:
Any word?

bsdimp:
Timeout. It's been months since I asked for changes. If you want to resubmit with the changes, please open a new PR. Thanks!

PR #1429, opened 2024-09-21
conf: make ZFS depend on XDR sources

svmhdvn:
ZFS nvpair.c source file depends on XDR unconditionally. This fixes the ZFS build with NFSD kernel options disabled.


bsdimp:
this makes me want to transition to a better config system... But this is good.

PR #1431, opened 2024-09-22
Fixes to fixes to 1363 problems

ehem:
~~This is kind of urgent as I had been wanting some of this with #1363.  Finally spotted the overt assembly to C bridge and that works better for things meant to be shared.~~

~~Hopefully got the headers right.  Turned out to need `LINT-ACPI` to find the combination which actually triggered the one condition.~~

~~This is kind of urgent-ish.~~

---
Now doing cleanup from the fallout resulting from people changing things without understanding.

### DIFF
--- a/sys/arm/annapurna/alpine/alpine_pci.c
+++ b/sys/arm/annapurna/alpine/alpine_pci.c
@@ -36,7 +36,6 @@
 #include <sys/module.h>
 #include <sys/bus.h>
 #include <sys/rman.h>
-#include <sys/intr.h>
 
 #include <dev/ofw/openfirm.h>
 #include <dev/ofw/ofw_bus.h>
--- a/sys/arm/arm/exception.S
+++ b/sys/arm/arm/exception.S
@@ -48,12 +48,11 @@
 
 #include "assym.inc"
 
-#include <sys/intr.h>
-
 #include <machine/asm.h>
 #include <machine/armreg.h>
 #include <machine/asmacros.h>
 #include <machine/trap.h>
+
 #ifdef KDTRACE_HOOKS
 	.bss
 	.align 4
--- a/sys/arm/arm/genassym.c
+++ b/sys/arm/arm/genassym.c
@@ -131,6 +131,8 @@ ASSYM(DCACHE_LINE_MASK, offsetof(struct cpuinfo, dcache_line_mask));
 ASSYM(ICACHE_LINE_SIZE, offsetof(struct cpuinfo, icache_line_size));
 ASSYM(ICACHE_LINE_MASK, offsetof(struct cpuinfo, icache_line_mask));
 
+ASSYM(INTR_ROOT_IRQ, INTR_ROOT_IRQ);
+
 /*
  * Emit the LOCORE_MAP_MB option as a #define only if the option was set.
  */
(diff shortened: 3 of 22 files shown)


bsdimp:
It's late here and I'm off to bed. I'm on travel tomorrow and may be jet lagged tuesday, and I'll look at it first of the pull requests. We have a lot of time before the branch to refine and I spent time at EuroBSDCon recruiting reviwers. I wanted to let you know quickly my schedule.

ehem:
On review seems the one which had been in the middle really needed to be first.  Dunno about others, but there seems a phenomenon the first commit usually needs to be last and the last commit needs to be first.

ehem:
I'm unsure the last one really needs to be part of _this_.  Simply a fragment of where I think things are going, simply illustrating what I was thinking of.

ehem:
While I'm not too hopeful about the latter two right now, I hope the two at the front can go in now.

ehem:
> This code is hard to review. It has taken me far to long to understand it.

Originally it was simpler.  Problem is someone broke something which meant more had to get in and the complexity went up.  Had this been reviewed months ago it wouldn't have the extras.

> First, it undoes the consensus ending to the sys/intr.h machine/intr.h debate. Those changes are recjected outright at this time. We may want to revisit it in the future, but there's not strong justification for doing it now. There's also not support for moving all the defines into machine/intr.h and moving them out of sys/intr.h, but that might be possible. Nothing really needs to have machine/intr.h be completely independent, apart from two arch's exception.S. The current situation is fine for the moment and not worth the churn to fix.

Having headers be standalone is standard practice.  This means you can include fewer and development is simpler due to needing fewer `#include` lines.

> Second, the renaming, at this time, of the interrupt files creates a lot more chrun than the benefit from having a uniform name.

This is required for `interrupt_t` to be useful.  This allows source files which merely handle pointers to the structure to `#include` a common name, then use pointers to the structure without looking inside.

> I dislike having interrupt_t. You can't forward declare typedefs. However, given the size of the change, it's hard to know for sure where that's going. I know it enables other things in #1457, so perhaps it would be better there.

It is a step in the direction of commonizing the interrupt frameworks.  Might seem low-ish value at first glance, but suddenly you can handle things as `interrupt_t *` instead of `void *`.

For #1457 this allows creation of alternatives to `intr_event_create()` which by using `interrupt_t *` drastically reduces the amount of casting.  While I am not adverse to _some_ casting, the amount in the FreeBSD kernel is unnecessarily excessive.  A major problem for #1457 was distinguishing the different cases in `io_apic.c`.

> I'm just a volunteer here. You don't get to demand service from me. Especially in the part of the system that can have huge systemic performance regressions if not done properly. So these things take time, and that's the best you can expect.

The initial steps of #1457 definitely have minor reduction.  `interrupt_t *` is part of an interface which regains some.

Alas it is primarily hypervisors which need common interrupt frameworks.  Allows porting to a single system instead of having to treat FreeBSD as 3 distinct OSes.

ehem:
Gah, #1281 also relies on `interrupt_t`, but there it is completely fundamental.  A common type allows nibbling hunks off the interrupt infrastructures in the hope of converging them.

ehem:
One higher value bit ended up going in with #1280, while the rest will go in with #1457.  The one nit which remains is what got into the FreeBSD repository was _significantly_ distinct from what was in the pull.  I'll going to leave this pull where it was at that time, but this pull no longer serves a sufficiently useful purpose, closing.

PR #1434, opened 2024-09-24
stats.3: Fix "skipping end of block" syntax errors

gperciva:
These were reported by `mandoc -T lint ...` as errors.

There was a small misunderstanding about the mdoc format: a 1-line enclosure does not need a closing tag.  For example, Pc is for closing an Po; if a line uses Pq (enclose the remainder of this line), there's no need for Pc.

### DIFF
--- a/share/man/man3/stats.3
+++ b/share/man/man3/stats.3
@@ -102,17 +102,17 @@
 .Fc
 .Ft struct voistatspec
 .Fo "STATS_VSS_CRHIST<32|64>_USR"
-.Fa Sy "HBKTS" Ns Pq Sy "CRBKT" Ns ( Em "lb" ) , "..." Pc ,
+.Fa Sy "HBKTS" Ns Pq Sy "CRBKT" Ns ( Em "lb" ) , "..." ,
 .Fa "vsdflags"
 .Fc
 .Ft struct voistatspec
 .Fo "STATS_VSS_DRHIST<32|64>_USR"
-.Fa Sy "HBKTS" Ns Pq Sy "DRBKT" Ns ( Em "lb" , "ub" ) , "..." Pc ,
+.Fa Sy "HBKTS" Ns Pq Sy "DRBKT" Ns ( Em "lb" , "ub" ) , "..." ,
 .Fa "vsdflags"
 .Fc
 .Ft struct voistatspec
 .Fo "STATS_VSS_DVHIST<32|64>_USR"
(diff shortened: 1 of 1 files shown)


gperciva:
PR created from https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=28165, for @mhorne.

To help with "man cleanup" commits, I created a script to check the ascii and html output of man pages, before and after a commit.  I'm not certain it'll be of interest to anybody else [1], but then again I'm not certain that it *won't* be of interest.

[1] unless somebody's working on fixing mandoc warnings and style?  At the moment, I'm only intending to clean up the errors.

Here's the script; if there's interest, I could clean it up a bit and put it somewhere.

```
#!/usr/bin/env python3

import difflib
import subprocess
import tempfile

import git

def render_man(filename):
    # Render to different formats
    outputs = []
    for fmt in ["ascii", "html"]:
        cmd = ["mandoc", "-c", "-T", fmt, filename]
        proc = subprocess.run(cmd, check=True, capture_output=True)
        output = proc.stdout.decode().splitlines()
        outputs.append(output)

    return filename, outputs[0], outputs[1]


def main():
    repo = git.Repo(".")
    if repo.is_dirty():
        print("Repo is dirty; commit first")
        exit(1)

    filenames = repo.git.diff_tree("--no-commit-id",
                                   "--name-only", "HEAD",
                                   "-r").splitlines()

    # Get current man output.
    mans_new = []
    for filename in filenames:
        mans_new.append(render_man(filename))

    # Get previous man output.
    repo.git.checkout("HEAD^", "--", filenames)
    mans_old = []
    for filename in filenames:
        mans_old.append(render_man(filename))
    repo.git.checkout("HEAD", "--", filenames)

    assert len(mans_new) == len(mans_old)

    # Compare
    err = 0
    for old, new in zip(mans_old, mans_new):
        assert new[0] == old[0]
        assert len(new) == len(old)
        for i in range(1, len(new)):
            output_old = old[i]
            output_new = new[i]
            if output_old != output_new:
                diffgen = difflib.unified_diff(output_old,
                                               output_new,
                                               lineterm="",
                                               n=1)
                diff = "\n".join(diffgen)
                print("Difference in %s:\n%s" % (new[0], diff))
                err = 1

    # Sanity check
    if err == 0:
        print("Checked:")
        for new in mans_new:
            print("\t%s" % (new[0]))


if __name__ == "__main__":
    main()
```

PR #1435, opened 2024-09-24
manuals: Fix "skipping end of block" .Ed errors

gperciva:
These were reported by `mandoc -T lint ...` as errors; this commit only handles unnecessary .Ed commands.


gperciva:
PR created from https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=281653 for @mhorne.

mhorne (approved):

PR #1436, opened 2024-09-24
manuals: Fix typos in -offset for .Bd and .Bl

gperciva:
The intended value is:
    -offset indent
If there's any typo such that the value doesn't match the pre-defined strings, then the offset is the same width as the value.  So by chance, "-offset -ident" ended up being a standard-width indent (since the default indent is 6 chars, and "-ident" also has 6 chars), whereas "-offset -indent" had a longer indent, and "-offset ident" had a shorter one.

### DIFF
--- a/bin/sh/sh.1
+++ b/bin/sh/sh.1
@@ -1074,13 +1074,13 @@ command.
 Commands may be grouped by writing either
 .Pp
 .Sm off
-.Bd -literal -offset -ident
+.Bd -literal -offset indent
 .Po Ar list Pc
 .Ed
 .Sm on
 .Pp
 or
-.Bd -literal -offset -ident
+.Bd -literal -offset indent
 .No { Ar list ; }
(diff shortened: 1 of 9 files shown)


gperciva:
Created from https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=281621 for @mhorne.

This is slightly modified; the bugzilla commit included changes to `-width ident`, which is not strictly what the commit subject says.  I removed those changes from this PR, and I'll bring the `-width TYPO` fixes a bit later.  (Also, while investigating, I discovered a few other `-width TYPO` that weren't fixed by the bugzilla commit.)

gperciva:
This commit *does* change the rendered ascii output, but only in the obvious / intended way (i.e. adding or subtracting a space from the indented portions).

concussious (approved):

mhorne (approved):

PR #1437, opened 2024-09-26
intpm: use MMIO for AMD EFCH CPUs

brian90013:
I observed an issue when attempting to use SMBus devices on an AMD Zen3 7443P CPU where running `smbmsg -p` found no devices. Previously I used a Zen2 7702P CPU in the same chassis/motherboard and had no issue finding or accessing SMBus devices. Debugging the intsmb driver (source file intpm.c) showed all the PMIO reads were returning 0xff on Zen3.

A comment in the Linux i2c-piix4.c source pointed me to the root cause:
```
cd6h/cd7h port I/O accesses can be disabled on AMD processors
w/ SMBus PCI revision ID 0x51 or greater. MMIO is supported on
the same processors and is the recommended access method.
```
Indeed, the revision ID of both the Zen2 and Zen3 CPUs is 0x61. I added a test for this condition and when true, set and allocate resources using MMIO instead of PMIO. I now see valid results from `smbmsg -p` and am able to use SMBus devices on both the Zen2 and Zen3 CPUs using MMIO.


sys/dev/intpm/intpm.c
```diff
@@ -135,28 +136,41 @@ sb8xx_attach(device_t dev)
...
+	 * Comment from Linux i2c-piix4.c:
+	 *
+	 * cd6h/cd7h port I/O accesses can be disabled on AMD processors
+	 * w/ SMBus PCI revision ID 0x51 or greater. MMIO is supported on
+	 * the same processors and is the recommended access method.
+	 */
+	if (devid == AMDCZ_SMBUS_DEVID && revid >= AMDCZ51_SMBUS_REVID) {
+		sc->type = SYS_RES_MEMORY;
+		addr = AMDFCH41_MMIO_ADDR + AMDFCH41_MMIO_PM_OFF;
+	}
```
bsdjhb, line 158: I would set `sc_type` and `addr` in an else clause here as I think that will be clearer to future readers.

bsdjhb: Well, I guess I see why `sc->type` has to be set where it is, I would still move the `addr` into an `else` here.

brian90013: Good point. While there I included `sc->type = SYS_RES_IOPORT` in the else for clarity even though it is set by default to work in non sb8xx cases.

brian90013:
Hello, I made the requested change back on October 7th and barring other feedback believe this MR is ready.

bsdimp (approved):
looks good to my eye

PR #1443, opened 2024-09-30
paste(1): capsicumise the utility

kfv:
This PR applies Capsicum security enhancements to the `paste` utility as part of my initial contribution with Capsicumisation. The user experience and POSIX compliance have been preserved, with adjustments made to improve the code’s structure and error handling.

My goal is to gain a deeper understanding of Capsicum by starting with simpler utilities before progressing to more complex codebases and libraries, so any corrections, suggestions, or feedback on the approach are highly appreciated, as I am eager to refine my work and contribute more broadly to the project.


usr.bin/paste/paste.c
```diff
@@ -44,11 +47,20 @@
 #include <unistd.h>
 #include <wchar.h>
 
+typedef struct _list {
```
gmshake, line 50: I'd recommend, if possible, move non-functional changes ( refactors ) to separated commit. This will make reviewing much easier.

usr.bin/paste/paste.c
```diff
@@ -98,41 +114,22 @@ main(int argc, char *argv[])
...
-	LIST *head, *tmp;
-	int opencnt, output;
-
-	for (cnt = 0, head = tmp = NULL; (p = *argv); ++argv, ++cnt) {
+	for (filecnt = 0, head = tmp = NULL; *argv; ++argv, ++filecnt) {
+		failed = 0;
 		if ((lp = malloc(sizeof(LIST))) == NULL)
 			err(1, NULL);
-		if (p[0] == '-' && !p[1])
+		if ((*argv)[0] == '-' && !(*argv)[1])
```
rilysh, line 121: Maybe just do argv[0][0] == '-' && argv[0][1] == '\0' instead of using a pointer (or !argv[0][1] is okay too).

usr.bin/paste/paste.c
kfv: Almost every change herein was necessary to smoothly introduce Capsicum and enter capability mode, so the refactoring is closely tied to this effort, as it helped streamline file management and error handling.

usr.bin/paste/paste.c
kfv: While both approaches likely generate the very same code, I opted for the pointer dereference since per C23 6.3.2.1p3, to my understanding, the expression is automatically converted from an array to a pointer anyway.

kfv: I referenced C23 as it's the latest approved draft, but section 6.3.2.1p3 is the same in C99 as well. Regarding readability, I have a different opinion, but I’m completely fine if you (or anyone else committing the PR later) decide to change it.

usr.bin/paste/paste.c
kfv: Alright, that makes sense. I’ll take care of it shortly, roger that.

kfv:
Cc: @oshogbo

usr.bin/paste/paste.c
```diff
@@ -113,12 +111,11 @@ typedef struct _list {
 	char *name;
 } LIST;
 
-STAILQ_HEAD(head, _list);
+STAILQ_HEAD(head, _list) static lh;
```
markjdb: This should appear in the previous commit which added the queue head.

`static` comes before the type name by convention, i.e., this should be `static STAILQ_HEAD ...`.

usr.bin/paste/paste.c
```diff
@@ -131,6 +133,16 @@ main(int argc, char *argv[])
 		STAILQ_INSERT_TAIL(&lh, lp, entries);
 	}
 
+	if (cap_enter() < 0 && errno != ENOSYS)
+		err(EXIT_FAILURE, "failed to enter capability mode");
+
+	cap_rights_init(&rights, CAP_READ);
+	STAILQ_FOREACH(lp, &lh, entries) {
+		if (lp->fp && cap_rights_limit(fileno(lp->fp), &rights) < 0 &&
```
markjdb, line 141: I would suggest calling cap_rights_limit() right after each fopen() call, rather than having a separate loop. It's easier to reason about that way (open+rights limiting happens in one go), and it's simpler.

Also, I don't think you need the `lp->fp` null check.

Also, I believe you can use caph_rights_limit() and avoid the ENOSYS check.

markjdb (commented):
This mostly looks ok to me, just some nitpicking.

I'm a bit wary about pre-opening all of the files at once. In general that's a bad idea since it might cause the utility to open more file descriptors than is allowed (small `RLIMIT_NOFILE` limits are not uncommon), but for paste that's probably ok...

I would also drop the sysexits change, as we generally don't add new uses of them.

usr.bin/paste/paste.c
kfv: Makes total sense. Just fixed it.

kfv:
> I would also drop the sysexits change, as we generally don't add new uses of them.

@markjdb: The termination status macros are from the standard C library (defined in `<stdlib.h>` as per C23 7.24p4 or the older C99 7.20p3,) not from `<sysexit.h>`.

usr.bin/paste/paste.c
kfv: > I would suggest calling cap_rights_limit() right after each fopen() call, rather than having a separate loop. It's easier to reason about that way (open+rights limiting happens in one go), and it's simpler.

You're absolutely right, it's totally valid. Done.

> Also, I don't think you need the lp->fp null check.

True.

> Also, I believe you can use caph_rights_limit() and avoid the ENOSYS check.

True, but since it's concise enough for this small use case, I kept it as is. Let me know if you think it needs to be changed.

kfv:
No worries, mate. Thanks!

usr.bin/paste/paste.c
```diff
@@ -98,51 +114,50 @@ main(int argc, char *argv[])
 		delim = tab;
 	}
 
-	if (seq)
-		rval = sequential(argv);
-	else
-		rval = parallel(argv);
+	for (filecnt = 0; *argv; ++argv, ++filecnt) {
```
oshogbo, line 117: Have you considered using fileargs?
man.freebsd.org/cgi/man.cgi?query=cap_fileargs&sektion=3&manpath=freebsd+14.1-release+and+ports
Opening files up front has some advantages, it's faster.
I wonder why you decided to open files upfront and not use fileargs?

usr.bin/paste/paste.c
```diff
@@ -98,51 +114,50 @@ main(int argc, char *argv[])
...
+			err(EXIT_FAILURE, "unable to limit rights for %s", *argv);
+
+		lp->cnt = filecnt;
+		lp->err = failed;
+		lp->name = *argv;
+
+		STAILQ_INSERT_TAIL(&lh, lp, entries);
+	}
+
+	if (cap_enter() < 0 && errno != ENOSYS)
```
oshogbo, line 142: Why not use `caph_enter`?

usr.bin/paste/paste.c
```diff
@@ -98,51 +114,50 @@ main(int argc, char *argv[])
...
+		if ((*argv)[0] == '-' && !(*argv)[1])
+			lp->fp = stdin;
+		else if (!(lp->fp = fopen(*argv, "r"))) {
+			if (!seq)
+				err(EXIT_FAILURE, "%s", *argv);
+			else
+				failed = errno;
+		}
+
+		if (cap_rights_limit(fileno(lp->fp),
```
oshogbo, line 130: Maybe caph_rights_limit?

usr.bin/paste/paste.c
```diff
@@ -98,51 +114,50 @@ main(int argc, char *argv[])
...
-			head = tmp = lp;
-		else {
-			tmp->next = lp;
-			tmp = lp;
-		}
-	}
-
-	for (opencnt = cnt; opencnt;) {
-		for (output = 0, lp = head; lp; lp = lp->next) {
+	for (opencnt = filecnt; opencnt;) {
```
oshogbo, line 158: In opencnt can we do a logical expression?

usr.bin/paste/paste.c
```diff
@@ -98,51 +114,50 @@ main(int argc, char *argv[])
...
+		lp->name = *argv;
+
+		STAILQ_INSERT_TAIL(&lh, lp, entries);
+	}
+
+	if (cap_enter() < 0 && errno != ENOSYS)
+		err(EXIT_FAILURE, "failed to enter capability mode");
+
+	rval = seq ? sequential() : parallel();
+
```
oshogbo, line 146: Do we care about freeing the list?

oshogbo (changes requested):

usr.bin/paste/paste.c
kfv: Lovely! I didn't know that, Mariusz. Thanks a lot, I'll take care of it.

usr.bin/paste/paste.c
kfv: I chose to use the raw Capsicum API instead of `caph` because it felt more concise and appropriate for the small-scale use cases in such a utility. Additionally, I thought having another direct example of capsicumisation with only raw API calls might be beneficial as a reference, considering further complexities observed here. While it does add some verbosity to the code, I don't see it causing any other significant issues. That said, if you still believe using Capsicum helpers would be better for this context, I’m happy to reconsider and follow your guidance.

usr.bin/paste/paste.c
kfv: Absolutely, but for the same reasons mentioned earlier, I felt it might be better to stick with the raw API. I’ve reviewed the helpers you’ve written, and they’re indeed excellent—adding no overhead or unnecessary abstraction. However, I thought it might still be worthwhile to use the raw API directly here. If you believe otherwise, I’m happy to adapt based on your recommendation.

usr.bin/paste/paste.c
kfv: Do you mean using `opencnt > 0`? That’s certainly a valid approach and it’s a good idea if you want to make the code more self-explanatory. It avoids relying on implicit truthiness, which can sometimes be less clear. Personally, I prefer to keep the code brief and concise where possible, which is why I didn’t use that form. However, if you think it would be helpful for clarity, I’d be open to considering the change.

usr.bin/paste/paste.c
kfv: It’s unnecessary to free the list explicitly since the OS will handle all memory cleanup upon program termination in this context. However, if the goal is code perfection and strict correctness, I’m open to adding a cleanup routine to free the list explicitly before exiting. But it's redundant, I believe.

oshogbo: Yea, I guess its fine.

kfv: Thanks! So I mark it as resolved.

usr.bin/paste/paste.c
oshogbo: Yes, I prefer to have a boolean expression in conditions it makes it easier to understand.

kfv: Roger that, I'll take care of it.

usr.bin/paste/paste.c
oshogbo: There is no need to use a raw API, capsicum helpers make it simpler.
Ideally when we make capsicum mandatory in FreeBSD it will also make code cleanup much simpler.

kfv: Makes sense. OK, I can take care of it as well, rog.

kfv:
Apologies for my initial reluctance to adopt `<capsicum_helpers.h>`. I now understand the rationale behind why we, as the FreeBSD Project, prefer using the helper interface over the raw Capsicum API. I’ve addressed all the related suggestions that were raised—thank you all for your patience and guidance.

usr.bin/paste/paste.c
kfv: Thank you for explaining this, Mariusz. I've now resolved the issue, and you're absolutely right — this approach significantly simplifies the implementation. Since performance isn't a critical concern for this utility, prioritising simplicity and maintainability makes far more sense. Cheers!

kfv:
@bsdimp: Hey, Warner! Sorry to jump in uninvited, just wanted to see if this PR might be worth considering as a 15-candidate too.

Cc: @oshogbo

bsdimp:
@oshogbo Looks like changes were made, and this is ready to land. Can you take one last look?

usr.bin/paste/paste.c
```diff
@@ -243,7 +259,8 @@ tr(wchar_t *arg)
 			default:
 				*arg = ch;
 				break;
-		} else
+			}
+		else
```
bsdimp, line 263: This likely should be } else {  and a closing } after the assignment.

bsdimp:
Tagged as both 'changes requested' and 'needs review'. My change is kinda trivial if it's otherwise ready to go

usr.bin/paste/paste.c
kfv: The brace closes the `switch`. Are you suggesting wrapping the `if`/`else` in braces for clarity? If so, please confirm and I’ll update it accordingly.

bsdimp: Yea... This is the 'old style' of one statement no braces... And this makes it a bit better. I'm going to land this as is, I think (I may tweak it for testing my landing tools, but I may not if the tests don't go well).

bsdimp:

PR #1444, opened 2024-10-01
nvme: Add support for Autonomous Power State Transition

eseipi:
APST is an optional NVMe power-saving feature that allows devices to automatically enter higher non-operational power states after a certain amount of idle time, reducing the controller's overall power consumption.

The feature configuration involves filling out the transition table, which then needs to be sent to the controller in a data buffer. Each table entry corresponds to one of the available power states and contains two values: idle transition power state (ITPS) and idle time prior to transition (ITPT). The first specifies the next power state the controller should switch to, and the second specifies the amount of idle time required before that switch.

Two sysctls are added: `apst_itpt_factor` for ITPT calculation (as an integer by which the total latency will be multiplied to get a suitable transition flow), and `apst_max_latency` for cutting off higher states with unwanted latency (by specifying a maximum value in microseconds).

The thing I struggled with the most was the default settings. I know that different controllers set them differently, and it's probably not a good idea to change them without the user's explicit intent. The best solution I could come up with is to check the return values of the `TUNABLE_INT_FETCH` calls and, depending on those, decide whether to touch the feature or not. The issue here is that if the feature is enabled by the vendor, sysctls will show zero values by default, but this would be incorrect. I'm not sure if this is a problem, but it's quite possible that a more elegant solution can be found.


sys/dev/nvme/nvme_ctrlr_cmd.c
```diff
@@ -251,6 +253,65 @@ nvme_ctrlr_cmd_set_interrupt_coalescing(struct nvme_controller *ctrlr,
...
+nvme_ctrlr_cmd_set_apst(struct nvme_controller *ctrlr, uint32_t itpt_factor,
+    uint32_t max_latency, nvme_cb_fn_t cb_fn, void *cb_arg)
+{
+	struct nvme_completion_poll_status status;
+	struct nvme_power_state *ps;
+	uint64_t *entry, total_latency;
+	uint32_t cdw11, itps, itpt;
+
+	if (ctrlr->cdata.apsta == 0 || ctrlr->quirks & QUIRK_DISABLE_APST) {
+		nvme_printf(ctrlr, "APST is not supported by the device\n");
```
bsdimp, line 266: This belongs under boot_verbose

sys/dev/nvme/nvme_ctrlr_cmd.c
```diff
@@ -251,6 +253,65 @@ nvme_ctrlr_cmd_set_interrupt_coalescing(struct nvme_controller *ctrlr,
...
+		return;
+	}
+
+	if (ctrlr->cdata.npss > 31) {
+		nvme_printf(ctrlr, "invalid NPSS, APST is not set\n");
```
bsdimp, line 271: Boot verbose, I think. Or maybe this should be a kassert if it's a programming error that originates in the kernel.

bsdimp:
Generally, I like this. I'm unsure of what we should do by default...

sys/dev/nvme/nvme_ctrlr_cmd.c
eseipi: I'm sorry, could you reconsider this? At this point, `nvme_ctrlr_cmd_set_apst` is only called by `nvme_ctrlr_start` or sysctl handlers. In both cases, the message is shown only when the user explicitly enables or disables the feature. Since it is not supported, the message acts more like a response to the user’s action. Without it, there would be no way to indicate that the operation has failed.

sys/dev/nvme/nvme_ctrlr_cmd.c
eseipi: Thanks, I'll change it to `KASSERT`.

eseipi:
@bsdimp, are you considering the UX issue I described or deciding whether to enable or disable the feature by default?

If it's the former, there's also an option to set both sysctl variables to -1 by default and take no action unless they are changed. This would avoid inconsistency, though it would make the code slightly less straightforward due to handling signed values. I can implement this if needed.

If it's the latter, I strongly believe it's not a good idea to enable the feature by default due to the latency trade-offs and potential flaws in the higher power states. At the same time, disabling power-saving when it's working properly isn't ideal either, and if it doesn't work, that’s not an issue specific to FreeBSD. Adding support for the mechanism without enforcing it won't harm anyone and can only benefit users who are concerned about power consumption.

wigneddoom:
This is a great feature. But for some reason, I think that we can implement it in user space - nvmecontrol(8)

amshafer:
Tested this on a dell 5960 over the weekend. Works as advertised and lets the nvme get into the lower sleep states, which for me are states 3 and 4. I don't see as hefty of savings as others but it seems to roughly save 0.3-0.5W based on `powermon` output. Would definitely be nice if we could get this merged.

One interesting observation is I think this particular nvme ramps down pretty fairly on its own, the actual power usage is always less than what that power state advertises. For example state 0 advertises 8W but doesn't use anywhere near that. I'd guess that the 8W figure is the *max* for that particular state but the nvme doesn't actually ramp that high. Other cards might not do this and that's why this patch saves so much power on them?

bsdimp:
Oh cool! I'll bump the priority of getting back to this then and making sure we have this feature, though maybe not this API (but if I can't come up with a better one, I may just commit this one). Thanks for testing @amshafer

sualbee:
Thanks for testing, Austin!

I'm still using this patch on my main laptop and I haven't noticed any issues with it as well.

However, at some point I realized that it would be much better to be able to configure the feature per drive, not globally for every nvme device in a system, as it is done now.  I think this is especially important when we're talking about transition data (or apst_data kenv, as in the current implementation), since, obviously, different controllers have their own set of power states.

Warner, I'm still ready to work on this patch in any direction you say if something else is needed.  Sorry for the time I might have wasted earlier with this PR, I'll try to do better next time.

bsdimp (approved):
If this were a new review, I'd complain about some style issues. Since it's less than new, and I have github landing guilt, I'll tweak the style on the way in.

Thanks for doing this, sorry I was on the fence so long about it.

bsdimp:

sualbee:
> Thanks for doing this, sorry I was on the fence so long about it.

@bsdimp It's all good, thanks for reviewing!

PR #1445, opened 2024-10-02
Makefile.inc1: allow real-update-packages to be called independently

svmhdvn:
To perform an incremental update of a pkgbase repo, you would call 'make update-packages', which will stage, create, and incrementally choose newer package versions to sign as part of a pkg repo. However, this forces you to stage the kernel and source packages along with the world packages. For a jail-only installation of FreeBSD, these packages are generally not required.

This patch separates the 'update-packages' target from the 'real-update-packages' target to allow choosing world, kernel, and/or source individually at the command line like so:

Jail-only installation:
```sh
make [...] -j$(nproc) buildworld
make [...] PKG_VERSION='15.snap<date>' \
    stage-packages-world create-packages-world \
    real-update-packages
```
Other minor changes are included to fix the PKG_VERSION variable definition logic and order of operations.

### DIFF
--- a/Makefile.inc1
+++ b/Makefile.inc1
@@ -2058,8 +2058,8 @@ PKGMAKEARGS+=	PKG_VERSION=${PKG_VERSION} \
 packages: .PHONY
 	${_+_}${MAKE} -C ${.CURDIR} ${PKGMAKEARGS} real-packages
 
-update-packages: .PHONY
-	${_+_}${MAKE} -C ${.CURDIR} ${PKGMAKEARGS} real-update-packages
+update-packages: stage-packages .PHONY
+	${_+_}${MAKE} -C ${.CURDIR} ${PKGMAKEARGS} create-packages real-update-packages
 
 package-pkg: .PHONY
 	rm -rf /tmp/ports.${TARGET} || :
@@ -2071,8 +2071,7 @@ package-pkg: .PHONY
 
 real-packages:	stage-packages create-packages sign-packages .PHONY
 
-real-update-packages: stage-packages .PHONY
-	${_+_}${MAKE} -C ${.CURDIR} PKG_VERSION=${PKG_VERSION} create-packages
+real-update-packages: .PHONY
 .if defined(PKG_VERSION_FROM_DIR)
 	@echo "==> Checking for new packages (comparing ${PKG_VERSION} to ${PKG_VERSION_FROM})"
 	@for pkg in ${PKG_VERSION_FROM_DIR}/${PKG_NAME_PREFIX}-*; do \
--- a/Makefile.inc1
+++ b/Makefile.inc1
@@ -2026,8 +2026,8 @@ PKG_EXT=	${PKG_FORMAT}
 PKG_EXT=	pkg
 .endif
 
-.if !defined(PKG_VERSION_FROM) && make(real-update-packages)
-.if exists(${PKG_ABI_FILE}) && exists(${REPODIR}/${PKG_ABI}/latest)
+.if !defined(PKG_VERSION_FROM) && make(real-update-packages) && \
+    exists(${PKG_ABI_FILE}) && exists(${REPODIR}/${PKG_ABI}/latest)
 PKG_VERSION_FROM!=/usr/bin/readlink ${REPODIR}/${PKG_ABI}/latest
 PKG_VERSION_FROM_DIR=	${REPODIR}/${PKG_ABI}/${PKG_VERSION_FROM}
 # Determine the name of the branch base on the version
@@ -2049,11 +2049,6 @@ BRANCH_EXT_FROM=	beta
 BRANCH_EXT_FROM=	rc
 .elif ${PKG_VERSION_FROM:M*.a*}
 BRANCH_EXT_FROM=	alpha
-.endif
-.else
-PKG_VERSION_FROM=
-PKG_VERSION_FROM_DIR=
-BRANCH_EXT_FROM=
 .endif
 .endif
 
@@ -2078,9 +2073,7 @@ real-packages:	stage-packages create-packages sign-packages .PHONY
 
 real-update-packages: stage-packages .PHONY
 	${_+_}${MAKE} -C ${.CURDIR} PKG_VERSION=${PKG_VERSION} create-packages
-.if empty(PKG_VERSION_FROM_DIR)
-	@echo "==> Bootstrapping repository, not checking for new packages"
-.else
+.if defined(PKG_VERSION_FROM_DIR)
 	@echo "==> Checking for new packages (comparing ${PKG_VERSION} to ${PKG_VERSION_FROM})"
 	@for pkg in ${PKG_VERSION_FROM_DIR}/${PKG_NAME_PREFIX}-*; do \
 	  pkgname=$$(pkg query -F $${pkg} '%n' | sed 's/${PKG_NAME_PREFIX}-\(.*\)/\1/') ; \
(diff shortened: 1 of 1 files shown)


Makefile.inc1
```diff
@@ -571,7 +571,6 @@ VERSION=	FreeBSD ${_REVISION}-${_BRANCH:C/-p[0-9]+$//} ${TARGET_ARCH} ${SRCRELDA
 .endif
 MAJOR_REVISION=	${_REVISION:R}
 
-.if !defined(PKG_VERSION)
```
svmhdvn, line 574: The below variables should always be defined regardless of what PKG_VERSION is set to, since it is used to compare two repos for incremental updates.

Makefile.inc1
```diff
@@ -593,9 +592,8 @@ EXTRA_REVISION=	.${BRANCH_EXT}.${_TIMENOW}
 BRANCH_EXT=	p${_BRANCH:C/.*-p([0-9]+$)/\1/}
 EXTRA_REVISION=	${BRANCH_EXT}
 .endif
-PKG_VERSION:=	${_PKG_REVISION}${EXTRA_REVISION:C/[[:space:]]//g}
-.endif
-.endif	# !defined(PKG_VERSION)
```
svmhdvn, line 598: This line was wrongly commented as `# !defined(PKG_VERSION)`.

Makefile.inc1
```diff
@@ -2036,16 +2034,11 @@ PKG_EXT=	${PKG_FORMAT}
...
 .endif
 
-.if !defined(PKG_VERSION_FROM) && make(real-update-packages)
-.if exists(${PKG_ABI_FILE}) && exists(${REPODIR}/${PKG_ABI})
+.if !defined(PKG_VERSION_FROM) && make(real-update-packages) && \
+    exists(${PKG_ABI_FILE}) && exists(${REPODIR}/${PKG_ABI}/latest)
 PKG_VERSION_FROM!=/usr/bin/readlink ${REPODIR}/${PKG_ABI}/latest
 PKG_VERSION_FROM_DIR=	${REPODIR}/${PKG_ABI}/${PKG_VERSION_FROM}
 BRANCH_EXT_FROM=	${PKG_VERSION_FROM:C/^[^[:alpha:]]+p?([[:alpha:]]*)[0-9]+$/\1/}
-.else
```
svmhdvn, line 2044: These should be left undefined instead of empty for better logic below.

Makefile.inc1
```diff
@@ -2054,8 +2047,8 @@ PKGMAKEARGS+=	PKG_VERSION=${PKG_VERSION} \
 packages: .PHONY
 	${_+_}${MAKE} -C ${.CURDIR} ${PKGMAKEARGS} real-packages
 
-update-packages: .PHONY
-	${_+_}${MAKE} -C ${.CURDIR} ${PKGMAKEARGS} real-update-packages
+update-packages: stage-packages .PHONY
```
svmhdvn, line 2050: The `update-packages` target continues to have the same behaviour as before. The only difference is that the `real-update-packages` target does not force a stage and creation of world, kernel, AND source package lists.

NOTE: `stage-packages` is listed in the prerequisites because it does not depend on the value of the `PKG_VERSION` variable, whereas `create-packages` does.

Makefile.inc1
```diff
@@ -2067,11 +2060,8 @@ package-pkg: .PHONY
 
 real-packages:	stage-packages create-packages sign-packages .PHONY
 
-real-update-packages: stage-packages .PHONY
-	${_+_}${MAKE} -C ${.CURDIR} PKG_VERSION=${PKG_VERSION} create-packages
-.if empty(PKG_VERSION_FROM_DIR)
-	@echo "==> Bootstrapping repository, not checking for new packages"
-.else
+real-update-packages: .PHONY
+.if defined(PKG_VERSION_FROM_DIR)
```
svmhdvn, line 2064: The logic is now inverted for better flow.

Makefile.inc1
```diff
@@ -2349,14 +2340,13 @@ real-sign-packages:	_pkgbootstrap .PHONY
...
 .endif
-	@[ -L "${REPODIR}/${PKG_ABI}/latest" ] && unlink ${REPODIR}/${PKG_ABI}/latest; \
 	${PKG_CMD} -o ABI=${PKG_ABI} repo -o OSVERSION="${SRCRELDATE}" \
 		-m ${WSTAGEDIR}/meta \
 		-o ${REPODIR}/${PKG_ABI}/${PKG_VERSION} \
 		${REPODIR}/${PKG_ABI}/${PKG_VERSION} \
 		${PKG_REPO_SIGNING_KEY} ; \
 	cd ${REPODIR}/${PKG_ABI}; \
-	ln -s ${PKG_OUTPUT_DIR} latest
+	ln -sf ${PKG_OUTPUT_DIR} latest
```
svmhdvn, line 2349: This avoids deleting the `latest` symlink when the `PKG_CMD` invocation fails.

bsdimp:
Can you look at this @evadot @bapt @emaste ?

Makefile.inc1
bsdjhb: Does the inversion depend on your other changes or can it be merged standalone?

Makefile.inc1
```diff
@@ -2089,9 +2079,10 @@ real-update-packages: stage-packages .PHONY
 	    echo "==> New package $${newpkgname}" ; \
 	  fi ; \
 	done
+.else
+	@echo "==> Bootstrapping repository, not checking for new packages"
 .endif
-	${_+_}@cd ${.CURDIR}; \
```
bsdjhb, line 2093: Using make -C here I think could be its own commit.

Makefile.inc1
```diff
@@ -2349,14 +2340,13 @@ real-sign-packages:	_pkgbootstrap .PHONY
 .if ${PKG_BIN_VERSION} < 11700
 	printf "packing_format = \"${PKG_FORMAT}\";\n" >> ${WSTAGEDIR}/meta
 .endif
-	@[ -L "${REPODIR}/${PKG_ABI}/latest" ] && unlink ${REPODIR}/${PKG_ABI}/latest; \
```
bsdjhb, line 2352: Likewise, this hunk to use a forced `ln -s` to remove the previous link can be its own commit I think.

svmhdvn:
I broke the commit into three separate commits:
* bugfix for the `latest` symlink
* non-functional style changes
* main change to make `real-update-packages` independent

Is this easier to review/merge incrementally now?

Makefile.inc1
svmhdvn: Also, I had this question on the `freebsd-dev` IRC channel at some point, but I don't remember getting a definitive answer on it. Why is the `${_+_}` prefix scattered around everywhere? As I understand, the manpage states that the `+` prefix can be used to force a command to always run, even when `-n` is used at the make command line. So why does it need to be expanded as a variable surrounded by `_` characters?

brooksdavis: I was curious. It's a local oddity to make the `-N` option not recurse. See 56585ab576ff9265a9c0f44b6d4986eb4f5461e3 and it's predecessor acabf29a1bdafe532c71c1f1a8053c7e32251649

svmhdvn:
Updated the PR on top of 458dc7f30310394a571fd4f2f5368acd48684652

svmhdvn:
It's been a while on this one, can we remove the "changes-required" tag at least to ensure accurate tracking of this bug?

Makefile.inc1
```diff
@@ -2049,11 +2049,6 @@ BRANCH_EXT_FROM=	rc
 .elif ${PKG_VERSION_FROM:M*.a*}
```
bsdimp, line 2049: This now happens even when PKG_VERSION_FROM might not be defined, though I suppose we might know it's always defined for other reasons I'm just not seeing. Do you know which it is?

Makefile.inc1
bsdimp: The _ are there to keep it out of the user's usual namespace.

Makefile.inc1
svmhdvn: Funny enough, because of Github's interface, I think I saw the same thing you saw after you wrote this comment :) Once you expand the collapsed hunk, this comment doesn't seem to apply.

bsdimp: Yea. You're right. Don't know what i was thinking

bsdimp:

PR #1446, opened 2024-10-02
Cleanstand

VexedUXR:
Some cleanup commits, as promised in #1098 :)

This series, along with some minor changes, does two main things:

- Move some common macros to headers.
- Use headers instead of doing things ourselves in the various elf64_freebsd.c files.

A note about moving bi_load into loader_efi.h:
Almost every file that includes loader_efi.h proceeds to do some form of
```c
extern int bi_load(char *, vm_offset_t *, vm_offset_t *, bool);
```
Since this is the efi version of bi_load, I moved it info loader_efi.h. However, the kboot files don't
always include loader_efi.h, so it looks like we won't be able to remove the declaration there.
I'm open to suggestions about where else to put bi_load (or any of the other things I moved),
but I don't see many options.

### DIFF
--- a/stand/common/bootstrap.h
+++ b/stand/common/bootstrap.h
@@ -239,10 +239,10 @@ struct preloaded_file
 	size_t f_size;		/* file size */
 	struct kernel_module	*f_modules;	/* list of modules if any */
 	struct preloaded_file	*f_next;	/* next file */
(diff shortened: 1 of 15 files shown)


bsdimp:
I think we may need to snag the 'no brainer yes' bits and iterate. So I'll look at doing that.

PR #1447, opened 2024-10-03
manuals: Fix "skipping end of block" .El errors

gperciva:


gperciva:
Thanks for merging the PRs, @mhorne!

These were reported by `mandoc -T lint ...` as errors; this commit only handles unnecessary .El commands.

PR #1448, opened 2024-10-03
manuals: Fix "missing end of block" errors

gperciva:


gperciva:
Another set of mdoc syntax-error fixes, @mhorne.

These were reported by `mandoc -T lint ...` as errors.

There are a few "missing end of block" syntax errors which this commit does *not* fix.  Those fixes introduce changes to the rendered output, so I figured that I'd save them for a later PR.

lib/libpathconv/abs2rel.3
```diff
@@ -93,6 +93,7 @@ The
 .Fa size
 argument is greater than zero but smaller than the length of the pathname plus 1.
```
gperciva, line 95: Removing this blank line would not change the ascii output, but it *does* change the html output (removing an empty `<p></p>` paragraph).

I therefore left that blank line alone; I might submit a "clean up useless whitespace" PR once all the syntax errors are fixed, but that's probably a month away.

mhorne (approved):

PR #1452, opened 2024-10-07
Update action to use native bmake if possible

kevemueller:
On MacOS, native bmake works fine, refactor the environment to be able to use it. 
While here, update to the latest ubuntu version, clang versions and clean-up YAML formatting.

### DIFF
--- a/.github/workflows/cross-bootstrap-tools.yml
+++ b/.github/workflows/cross-bootstrap-tools.yml
@@ -2,9 +2,13 @@ name: Cross-build Kernel
 
 on:
   push:
-    branches: [ main, 'stable/14', 'stable/13' ]
+    branches: 
+      - main
+      - 'stable/14'
+      - 'stable/13'
   pull_request:
-    branches: [ main ]
+    branches: 
+      - main
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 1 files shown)


kevemueller:
Using private email now 72columns commit message.

.github/workflows/cross-bootstrap-tools.yml
```diff
@@ -51,17 +60,35 @@ jobs:
...
+          echo Using NPROC=${NPROC}
+          echo Using EXTRA_MAKE_ARGS=${EXTRA_MAKE_ARGS}
+
       - name: make kernel-toolchain
-        run: ./tools/build/make.py --debug $EXTRA_BUILD_ARGS TARGET=${{ matrix.target }} TARGET_ARCH=${{ matrix.target_arch }} kernel-toolchain -s -j$NPROC -DWITH_DISK_IMAGE_TOOLS_BOOTSTRAP
+        run: ${{ matrix.bmake }} TARGET=${{ matrix.target }} TARGET_ARCH=${{ matrix.target_arch }} -s -j$NPROC kernel-toolchain -DWITH_DISK_IMAGE_TOOLS_BOOTSTRAP
+
       - name: make buildkernel
-        run: ./tools/build/make.py --debug $EXTRA_BUILD_ARGS TARGET=${{ matrix.target }} TARGET_ARCH=${{ matrix.target_arch }} KERNCONF=GENERIC NO_MODULES=yes buildkernel -s -j$NPROC $EXTRA_MAKE_ARGS
+        run: ${{ matrix.bmake }} TARGET=${{ matrix.target }} TARGET_ARCH=${{ matrix.target_arch }} -s -j$NPROC buildkernel KERNCONF=GENERIC NO_MODULES=yes
```
bsdimp, line 94: Why elinminate EXTRA_MAKE_ARGS here?

.github/workflows/cross-bootstrap-tools.yml
```diff
@@ -17,22 +21,27 @@ jobs:
...
-        os: [ ubuntu-22.04, ubuntu-24.04, macos-latest ]
+        target_arch:
+          - amd64
+          - aarch64
+        os: 
+          - ubuntu-24.04
+          - macos-latest
         include:
-          # TODO: both Ubuntu and macOS have bmake packages, we should try them instead of bootstrapping our own copy.
-          - os: ubuntu-22.04
```
bsdimp, line 24: This also seems to eliminate ubuntu-22 from the matrix. Is that intentional?

.github/workflows/cross-bootstrap-tools.yml
kevemueller: ```EXTRA_MAKE_ARGS``` was referenced in the original script, it was never set to any value, it could not be set in the call environment. I have left the ```echo``` and would remove that in a subsequent change, could be done right away.
I found it confusing to have to find what ```EXTRA_MAKE_ARGS``` _in this context_ is actually used for, hence I removed the reference.

.github/workflows/cross-bootstrap-tools.yml
bsdimp: We'll have to ask @jrtc27 to see which ones we can drop, since she added this originally.
I think that ubuntu-20 is safe to drop, but I think she (or some other CheriBSD person) had said that they still needed to support ubuntu-22.

jrtc27:
I don't think the native bmake is configured correctly. Parts of the build, in particular release/, rely on it being configured with .../share/mk as the first item in its path (see tools/build/make.py's configure line). The comment in the workflow predates that configuration and is likely no longer accurate. I would be very concerned with doing anything other than using tools/build/make.py to wrap bootstrapping the in-tree bmake, and do not see the benefit of it when it takes a fraction of the total build time to bootstrap, only the downside of more headaches from subtle configuration differences.

bsdimp:
I'm going to close this... There's no consensus for it at the moment, it's languished here for a while... If things change, we can do another PR to land this or something like this that's more acceptable..

PR #1453, opened 2024-10-07
manuals: Misc syntax fixes

gperciva:


gperciva:
These were reported by `mandoc -T lint ...` as errors.  Tagging @mhorne.

In contrast to my recent PRs, this one **does** change the ascii (and html) output.

If it helps reviewing, here's the diff in the ascii *output*, other than the changes to `orders (8)` because those are too verbose.

```
Difference in lib/librss/librss.3:
---
+++
@@ -91,3 +91,3 @@
      The rss_sock_set_recvrss() function enables or disables receiving RSS
-     related information as socket options in.  calls.
+     related information as socket options in recvmsg(2) calls.

Difference in share/man/man4/bpf.4:
---
+++
@@ -417,3 +417,3 @@
      bpf now supports several standard ioctl(2)'s which allow the user to do
-     async and/or non-blocking I/O to an open file descriptor.
+     async and/or non-blocking I/O to an open bpf file descriptor.

Difference in share/man/man8/rc.subr.8:
---
+++
@@ -25,2 +25,3 @@
      rc_log message
+     rc_trace level message
      rc_usage command ...
Difference in share/man/man9/cdefs.9:
---
+++
@@ -362,4 +362,4 @@
      selects which C environment is used.  However, when C11 dialect is
-     selected with IEEE Std 1003.1-2008 ("POSIX.1"), definitions for are
-     included.
+     selected with IEEE Std 1003.1-2008 ("POSIX.1"), definitions for ISO/IEC
+     9899:2011 ("ISO C11") are included.

Difference in share/man/man9/zone.9:
---
+++
@@ -261,4 +261,4 @@
	   Create a zone whose items will be synchronized using the smr(9)
-	   mechanism.  Upon creation the zone will have an associated structure
-	   which can be fetched using uma_zone_get_smr().
+	   mechanism.  Upon creation the zone will have an associated smr_t
+	   structure which can be fetched using uma_zone_get_smr().
```

mhorne (approved):
To state the obvious, the rendering changes here are correct/desirable... :)

PR #1454, opened 2024-10-07
manuals: Fix "missing end of block" errors

gperciva:

PR #1455, opened 2024-10-08
manuals: Fix syntax of list width

gperciva:


gperciva:
This changes the rendered ascii but not html output.  The actual changes are adding 1 or 2 spaces to the indentations, exactly as we'd expect from the source diff.

Tagging @mhorne.

17 mandoc errors remaining after this, #1454, and #1453 are merged.

share/man/man9/vmem.9
```diff
@@ -159,7 +159,7 @@ A bitwise OR of an allocation strategy and a
 .Xr malloc 9
 wait flag.
 The allocation strategy is one of:
-.Bl -tag width indent
+.Bl -tag -width indent
```
concussious, line 162:
```suggestion
.Bl -tag -width "qcache_max"
```
Is ten too many? We could align every table on this page with `qcache_max`, which is quite handsome, to me.

gperciva: I'm not certain what the policy is for "use the standard indent" vs. "specify a customized number of letters".

There's two issues here:
- are you sure that you commented on the right line?  That particular `.Bl` is a nested list with only 3 entries, the longest of which is `M_FIRSTFIT`.

- I'm fairly certain that you meant to comment on a different line (and a few others), to suggest that all 6 functions in the DESCRIPTION should have the same width.

  I'm not opposed to that, but it seems overkill to me.  If it only had 2 functions, then sure, it would look typographically nicer to align the widths.  But 6 functions is more, and there's the added cost of maintainability -- if any function gets a longer argument, then the committer would need to notice that the other 5 `.Bl` should also have their widths increased.

I'm also biased in that when I'm fixing syntax errors, I prefer to keep the changes as minimal as possible.  So another option would be to merge the syntax fixes as-is, then you could edit this later (and perhaps other man pages) to align widths.

mhorne: I'm happy to take just the syntax fixes for now, but I agree that the formatting for this page in particular should be improved. @concussious will you take that on?

mhorne (approved):

share/man/man9/vmem.9
concussious: Yes, thanks for including me!

concussious: > I'm not certain what the policy is for "use the standard indent" vs. "specify a customized number of letters".

You can find it in style.mdoc(5), and an example in freebsd/#1496. The narrowest width I'm aware of someone in our community using seriously is 59 char, so I test it in standard(80) and narrow(59) and just subjectively look at it. Does it fold the lines over too much? Is it too far over to the side? Does it look more or less pleasant one way or the other? If it looks neater, it's less stressful to read, and easier for people to comprehened, and leaves a better impression of the professionalism of the project.

A big part of technical writing is about how it feels when you're reading it, so I just look at current vs draft side by side in tmux, and evaluate the cohesive flow of the entire document. Imho it needs to be considered that way. Doc is written for humans to read, and it is an honor to be a part of this tradition.

> are you sure that you commented on the right line?

Oops! I did have the incorrect line, sorry!

PR #1457, opened 2024-10-09
x86 PICs as devices

ehem:
A major reason behind non-x86 PICs being devices is to utilize the module and newbus infrastructure.  This does make x86 the odd one out, so turning the x86 PICs into full devices will likely aid converging the interrupt systems.

Having PICs for all architectures be proper devices allows the interrupt event core use device methods for interrupt callbacks.  This shrinks the interrupt event data structures as everything can be done via a single pointer, rather than a pointer for each callback.

If interrupt events could be embedded inside the architecture interrupt structures, `struct intr_event` could be further shrunk by removing another pointer.  The result would make the architecture core <=> event core relationship similar to the PIC <=> architecture core relationship.


ehem:
This has ended up going rather further than I had expected before terminating.

The old Marvell GPIO driver has needed a severe update for some time.  It has become so rotten I can no longer work around the problems.

The x86 local-APIC and IO-APIC drivers have been tested.  The AT PIC driver _may_ work, but I haven't yet gotten a good recipe for `qemu`.  The impact on x86/Xen and x86/MSI drivers is small enough for them to likely still be working, but they're not yet tested.

ehem:
This is certainly going to need some adjustment, but I need advice as to which direction to go with this.  There are 3 places I see where this series could be reasonably terminated.

First, it could be terminated after converting the x86 PICs to devices.  This was the main goal.  This fails to pick up most of the benefits, but does make the interrupt infrastructures consistent in PICs being devices.

Second, it could terminate after the conversion to `int intr_event_create_device(struct intr_event **event, device_t pic, interrupt_t *source, u_int irq, int flags, const char *fmt, ...)`.  This shrinks the structures for hardware interrupts by 24 bytes (3 fewer pointers) and enlarges software interrupts by 16 bytes.  Many systems will have >100 hardware interrupts and with 3-7 software interrupts being the maximum, this is a notable benefit.  This requires conversion to `#include <machine/interrupt.h>` in many places.

Third, it could terminate after converting to `int intr_event_init(struct intr_event *ie, device_t pic, u_int irq, int flags, const char *fmt, ...)`.  I feel having the PIC <=> arch core interface being similar to the arch core <=> shared <core> interface is a good thing.  I've found it rather bizarre how the PIC structure wraps the architecture core, but the architecture core then has a pointer to the shared core.  Not having memory allocation in `intr_event_init()` allows the architectures to initialize their `intr_event` structures sooner as they can be setup while the appropriate mutex is held.  This also reduces memory usage for hardware interrupts by 48 bytes and enlarges software interrupts by 8 bytes.

Overall this seems a rather good idea.  Yet now I'm waiting on reviewers...

bsdimp:
This change is too large. It has a lot of gratuitous changes that aren't strictly necessary. 49 changes and 244 files is simply too much to review. A quick survey of the changes shows lots of header file renames (which aren't needed) and kobj function name renaming (also need needed). In it's current form, it's not possible to review this.

ehem:
That is what happens if you look at a squash of 49 commits.  Take almost any sequence of 49 commits in FreeBSD's repository, look at the delta and it will be confusing.  There is also some infrastructure which has been waiting 2.5 years and had that been brought in in a timely fashion it wouldn't need to be included here.

For some overview of the earlier ones, 5a4854b05487ab763760eb7928a9fea3e6c53f4d ended up getting in via other means.  It is here as I was expecting it to be brought in with this, but then unexpected it came in via other means.

The [next 3](https://github.com/freebsd/freebsd-src/pull/1457/files/5a4854b0548..495cdd552da) are 2 bits of cleanup and one thing I've got an unpleasant suspicion will explode in the not too distant future.

[Commit #5](https://github.com/freebsd/freebsd-src/pull/1457/commits/5f1a8cba4b4) is a key commit.  Without that the approach is impossible.  Notably x86 presently initializes PICs at `SI_SUB_INTR` while newbus was doing all initialization at `SI_SUB_DRIVERS`.  Initializing the root bus earlier works fine, but I'm unsure whether that will be accepted.

Advice on the question asked immediately above might allow abandoning some pieces.  Might be clearer to point to the [primary commit](https://github.com/freebsd/freebsd-src/pull/1457/files/60f224c125b..8c03f301536) for `intr_event_create_device()`.  The point is the arguments to `intr_event_create()` are rarely independent.  For any given PIC, the functions will _always_ be the same.  As such if the interface moves to `intr_event` there is substantial savings on hardware interrupts since only a single pointer is used.

There isn't a good commit to point to for `intr_event_init()`, though perhaps f2abf738c8abca7b990c3fc650de0786c46cbe0f can give some idea.  I've long been bothered by how INTRNG and x86 PICs wrap their data around the architecture interrupt type, then cast back and forth.  Whereas both INTRNG and x86 include a `struct intr_event *<prefix>_event;` inside their respective structures.

Why is one boundary embedding the structure inside the other, while the other is handled via pointer?  These two relationships are similar, so shouldn't their interfaces be similar?  Embedding the structure results in 16 bytes of savings since two pointers are removed (pointer to `intr_event` and there is no longer any need for `->ie_source`).

sys/kern/subr_bus.c
```diff
@@ -1205,7 +1205,12 @@ static int
 devclass_alloc_unit(devclass_t dc, device_t dev, int *unitp)
 {
 	const char *s;
-	int unit = *unitp;
+	u_int unit = *unitp;
+
+	_Static_assert((int)DEVICE_UNIT_MAX > 0,
+	    "DEVICE_UNIT_MAX must be non-negative when converted to integer");
+	_Static_assert(DEVICE_UNIT_MAX < DEVICE_UNIT_ANY,
+	    "DEVICE_UNIT_MAX must be less than DEVICE_UNIT_ANY");
```
bsdimp, line 1212: These static asserts are misplaced.
Also, won't the second one fire since INT_MAX is not < -1?
Also, why not just use INT_MAX?

sys/kern/subr_bus.c
```diff
@@ -1226,7 +1231,7 @@ devclass_alloc_unit(devclass_t dc, device_t dev, int *unitp)
 	} else {
 		/* Unwired device, find the next available slot for it */
 		unit = 0;
-		for (unit = 0;; unit++) {
+		for (unit = 0; unit < DEVICE_UNIT_MAX; unit++) {
```
bsdimp, line 1233: INT_MAX

sys/kern/subr_bus.c
```diff
@@ -1240,6 +1245,9 @@ devclass_alloc_unit(devclass_t dc, device_t dev, int *unitp)
 		}
 	}
 
+	if (__predict_false(unit >= DEVICE_UNIT_MAX))
+		return (ERANGE);
```
bsdimp, line 1248: EINVAL is better, plus this needs to be added to the docs here and in this routine's callers.

sys/sys/bus.h
```diff
@@ -160,6 +160,7 @@ struct devreq {
 #define	DEVF_RESET_DETACH	0x0000001	/* Detach drivers vs suspend
 						   device */
 #define DEVICE_UNIT_ANY		(-1)
+#define DEVICE_UNIT_MAX		(~(unsigned int)0 >> 1)
```
bsdimp, line 163: Not needed if you just use INT_MAX

sys/kern/subr_bus.c
```diff
@@ -1250,8 +1258,8 @@ devclass_alloc_unit(devclass_t dc, device_t dev, int *unitp)
 		int newsize;
 
 		oldlist = dc->devices;
-		newsize = roundup((unit + 1),
-		    MAX(1, MINALLOCSIZE / sizeof(device_t)));
+		newsize = MIN(DEVICE_UNIT_MAX, roundup((unit + 1),
+		    MAX(1, MINALLOCSIZE / sizeof(device_t))));
```
bsdimp, line 1261: I've redone this a little:
 https://reviews.freebsd.org/D47359
and added you as a co-author. I've also audited all the places this could have fallout, which I document in the commit message.

Also, unsigned doesn't make sense: 2^31-1 devices is plenty.

sys/kern/subr_bus.c
ehem: Second doesn't trigger since `DEVICE_UNIT_MAX` was declared `unsigned`.  This seemed a reasonable place since everything was in `devclass_alloc_unit()`.

sys/kern/subr_bus.c
ehem: `DEVICE_UNIT_MAX` could reasonably be set less than `INT_MAX`.  The most crucial factor is to ensure it doesn't overflow into the negative range, but I haven't yet seen any devices which had unit numbers in excess of 16 (which would be way too low as a setting, but `1<<31` is likely excessive).

sys/kern/subr_bus.c
ehem: Hopefully all the spots have been checked, but indexing an array with a signed type makes me nervous.  I'm fine with [D47359](https://reviews.freebsd.org/D47359).  This was mostly a case of during the bring-up phase I noticed this and it left me a bit uneasy.  Might not be worthy of much concern, but I did want to flag this as being a bit worrisome.

bsdimp: They all have been checked. As have all the maxunit checked, and all callers.

sys/kern/subr_bus.c
```diff
@@ -5108,7 +5109,6 @@ root_bus_module_handler(module_t mod, int what, void* arg)
 {
 	switch (what) {
 	case MOD_LOAD:
-		TAILQ_INIT(&bus_data_devices);
```
bsdimp, line 5111: This change doesn't move the needle enough to take at this time. bus_data_devices is the first thing that does this, so that makes this a little inconsistent. It's fine how it is, and there's not a compelling reason to  do this. Please drop this change from this commit series.

sys/kern/subr_bus.c
```diff
@@ -472,14 +472,28 @@ driver_register_pass(struct driverlink *new)
...
+
 /**
  * @brief Raise the current bus pass
  *
  * Raise the current bus pass level to @p pass.  Call the BUS_NEW_PASS()
  * method on the root bus to kick off a new device tree scan for each
  * new pass level that has at least one driver.
  */
-void
+static void
```
bsdimp, line 495: This whole change looks good, I'll pick it up. However, we may need to make this non-static

sys/kern/subr_bus.c
bsdimp: now this this has landed, please drop this from this series.

sys/kern/subr_bus.c
```diff
@@ -5156,7 +5154,7 @@ static moduledata_t root_bus_mod = {
 	root_bus_module_handler,
 	NULL
 };
-DECLARE_MODULE(rootbus, root_bus_mod, SI_SUB_DRIVERS, SI_ORDER_FIRST);
+DECLARE_MODULE(rootbus, root_bus_mod, SI_SUB_INTR, SI_ORDER_FIRST);
```
bsdimp, line 5157: this is premature at this time. I'm not going to land it just yet.

bsdimp:
OK. I've snagged the newbus parts of this that we're going to take. The cleaned I didn't pull isn't enough of a change to consider at this time and we need to prune this very large pull request if we're going to make progress on it. Each additional change adds time to testing and risk, so I'd like to keep the ones we pull in more directly related to the topic of this pull request.

The interrupt stuff looks good on its surface, but I haven't had time to more than scratch the surface or test the affected files, or evaluate if there's potential performance issues that I'll need test and measure in our environment at work.

So please rebase, drop the newbus cleanup stuff I didn't accept for now (you'll likely need to retain the one change i said was premature since it looks to be needed later in this series). Thanks for your patience.

sys/kern/subr_bus.c
ehem: Okay.  I really hope out-of-tree architectures have moved away from this, but that wasn't something I was aware of.  It looked effectively deprecated since there weren't any other uses.

sys/kern/subr_bus.c
ehem: I has been considering squashing it into the move to => `SI_SUB_INTR`.  Mainly I had looking at what could be pulled out of `root_bus_module_handler()`.

sys/kern/subr_bus.c
ehem: Fine by me.  Soon to be later in the series.

ehem:
> OK. I've snagged the newbus parts of this that we're going to take. The cleaned I didn't pull isn't enough of a change to consider at this time and we need to prune this very large pull request if we're going to make progress on it. Each additional change adds time to testing and risk, so I'd like to keep the ones we pull in more directly related to the topic of this pull request.

They're all related, though a few are a bit more distant.  Removing `->pic_vector()` shrinks the compatibility macros.  The size of the temporary transition compatibility layer is a bit large.

The PowerPC cleanup is related to the much later `intr_event_create_device()`/`intr_event_init()` interfaces.  That isn't important up-front, but should be easy and mass cleanup is always valuable.

> The interrupt stuff looks good on its surface, but I haven't had time to more than scratch the surface or test the affected files, or evaluate if there's potential performance issues that I'll need test and measure in our environment at work.

I haven't gotten full testing either, but I have gotten most of it tried.  I've confirmed the devices for local-APIC, IO-APIC, and Xen PIC work.  I lack a handy recipe for `qemu-system-*` to test the AT PIC and MSI PICs.  The delta for the MSI PIC is fairly small so I think that has good odds of working.
The AT PIC though is very uncertain.  The driver is very quirky and there is a good chance this breaks it.

As for performance, it depends on which end point is chosen and which architecture you're concerned with.  I believe the overall impact should be fairly small.

For x86, using devices is likely a very mild performance reduction.  According to the documentation `kobj` is very cheap, but that will have an impact.  The question is how critical are f9acc6410e5e and the other quirks in the AT PIC driver?  This might fully grind an i80386 into dust, but I suspect the (trivial) memory reduction may help an amd64 system.
Later on the question becomes whether the shrinkage of `struct intr_event` has more or less impact than switching to `device_t`/`kobj`.  On a modern system with lots of interrupts there may be a trivial gain from the new interface.

For PowerPC, its interrupt infrastructure is rather different from the other architectures.  The question is how much impact is there from having to update the 0c50edff52f2.  My guess is the impact will be small.

INTRNG will definitely **like** `intr_event_create_device()`/`intr_event_init()`.  Shrinking `struct intr_event` will be minor.  Yet the interface of those two proposed interfaces is very friendly to INTRNG.  As long as the INTRNG team is okay with the adjustments needed, they mesh perfectly and this should be positive.

Still an issue of which endpoint to use...

> So please rebase, drop the newbus cleanup stuff I didn't accept for now (you'll likely need to retain the one change i said was premature since it looks to be needed later in this series). Thanks for your patience.

Done.  Reordering things often happens when trying to get things in.

sys/x86/x86/io_apic.c
ehem, line 1: In case anyone wonders, the issue is there are 19 casts to/from `struct ioapic *` to/from `void *` in `sys/x86/x86/io_apic.c`.  Trick was 10 of these are simply keeping the caller's fingers out of a private structure.  While 9 were interfacing with the outside via generics.  The last bug with the IO-APIC was I was having to examine every cast to figure out which type it was.  This was very useful for isolating the others.

ehem:
I've mentioned it a few times and not gotten a response, so trying to collect my thoughts and go into the two key points.

```
int    intr_event_create(struct intr_event **event, void *source,
           int flags, u_int irq, void (*pre_ithread)(void *),
           void (*post_ithread)(void *), void (*post_filter)(void *),
           int (*assign_cpu)(void *, int), const char *fmt, ...)
           __printflike(9, 10);
```

It is notable the four arguments, `pre_ithread`, `post_ithread`, `post_filter` and `assign_cpu` are rarely independent.  In fact the only two cases where they vary are from x86.  First, the interrupt core where `post_ithread` and `post_filter` are copied from the PIC's function table.  Second, the AT PIC driver where `pic_eoi_source` changes depending on which of the two chips is triggering the interrupt.

For INTRNG they're always the same function pointers.  For PowerPC they're always the same function pointers.  In light of this it makes sense to bundle them together and replace the 4 pointers with a single pointer.  If parts of the interrupt core can assume callers are `device_t` then using `kobj` methods allows them to bundled together.

This yields:
```
int     intr_event_create_device(struct intr_event **event, device_t pic,
	    interrupt_t *source, u_int irq, int flags, const char *fmt, ...)
	    __printflike(6, 7);
```
A compatibility `intr_event_create()` was readily created (mostly for SWIs).

The other observation is how wildly different the interfaces between architecture and shared core, versus architecture and PIC are.  Notably both INTRNG and x86 the PIC allocates the appropriate structure and the architecture core simply fills it in (the PIC allocates extra to use for its own data, then casts to/from its larger structure).  Whereas the shared core is always a pointer from the architecture structure to the shared structure.

I was given the impression the thinking was `struct intr_event` was special and therefore always allocated via `M_ITHREAD`.  This seems wrong to me.  I would suggest the shared core is merely a baseline of things which every architecture can agree on (not much).  They're all kept in a list.  They all have a name.  They all have an "`irq`".  They can always be bound to a cpu.  If we think of `struct intr_event` as simply a shared portion of the interrupt structure then we get:
```
int    intr_event_init(struct intr_event *event, device_t pic, u_int irq,
           int flags, const char *fmt, ...) __printflike(5, 6);
```
Which results in calls of the form `INTR_EVENT_POST_ITHREAD(ie->ie_pic, (interrupt_t *)ie);` and `INTR_EVENT_ASSIGN_CPU(ie->ie_pic, (interrupt_t *)ie, cpu);`.  This removes the back and forth pointers from `struct intr_event` and the architecture interrupt type.  For PowerPC and INTRNG this has the distinct advantage of `intr_event_init()` can be called while mutexes are held.  This also reduces the complexity of recovery paths.

`intr_event_init()` represents a significant change in how things have been done.  I think the change is good, but I do have `intr_event_create_device()` as a fallback if reviewers aren't up for this magnitude of change.

ehem:
Guess I reopened this.  Decided I really wanted the favored statement and this is a rather potent demonstration.

Presently the five at the start are cleanup-ish.  Reviewers urgently need to be careful though.  The third through fifth include some spots which are directly leading into later commits and changing those will break later commits.

Notably the loop in `atpic_init()` will get expanded later.  `intr_disable_src()` was deliberately kept since `isrc->is_pic->pic_disable_source` disappears with `kobj`.

ehem:
Another day, another bit of cleanup spotted and a goof with conversion spotted.

Now up to 6-ish cleanup commits at the start.  Note the 5th and 6th still have those quirks leading into later commits.  This is then followed by a 13.5 commit marathon of the actual conversion.

Then followed by 3 more refinements, then a huge aftershock.

ehem:
This needs regular rebasing onto HEAD since portions are rather sensitive to commit activity on main.  Certain types of changes mean this series has to be updated.

This can actually be approached from two directions, the original x86 PICs as devices, but also as converting interrupt events to kernel object methods.  The former is aiming at convergences of the interrupt frameworks, this is quite valuable for hypervisor usage.  The latter shrinks interrupt events and provides a tiny performance boost to INTRNG and PowerPC (which already have PICs as devices).

Commit 3d3aaea725aa351f6939dc2a99af6d547dbc8f36 is where x86 is converted to PICs as devices.  There is likely a tiny performance loss due to adding a shim layer and kernel object methods being slightly slower than function pointers.  Commit 53f55463de7490164448d6193aa2dde6d9d23ed1 is where x86 plugs into the new kernel object method for interrupt events and this should be close to performance-neutral (shim layer is removed).  On a system with large numbers of interrupts this may enhance performance due to shrinking interrupt events reducing cache usage.

Commit 9797d1287117cfaba221b1840c908c5fb2b564fb is purely the conversion of interrupts events to kernel object methods.  Passing events to the functions as `interrupt_t *` is ideal since this matches what INTRNG already does and matches the conversion of x86 to PIC devices.  The one quirk is the device root bus needs to be initialized at `SI_SUB_INTR` since that is when x86 starts creating PICs/events.  Commit 3734dbe20d7f4822f221858fd23cf330e91a5aa0 is the conversion of INTRNG to the interrupt event methods, should increase performance due to removal of the shim layer and shrinkage of interrupt events.
Commit 025793b662ed8bafef66c149a218626fbaae25c3 is a similar conversion for PowerPC.  PowerPC likely doesn't gain nearly as much x86 or INTRNG do, but reduced cache helps everyone.

ehem:
Presently commit 7018d815105bac2d95a5da5430fec234683625c5 is the end of the x86-centric path.  If someone with a commit bit has only a little bit of time, the commits through 169dab82e6314fffe10bf540b5e7741aa2f9a722 are pretty quick standalones.

The main course of the interrupt event => kobj method series is fdd1f887b141e36ab5cde4519191b0764f760603.  INTRNG runs through 38660f74a806ccafd1186c63b2a7627e224c199e, PowerPC goes to a909236514b913a1a446c916e65bcc07130e85c9.

ehem:
This branch is rather sensitive to certain changes, so needs updating fairly often.

The x86-centric path now terminates at eedbfb45b997268dc19a4ec3e5256e93a3b306d2.  The short standalone path is a28e1e8b8da21899353f19500ae419d13130783a.

The main sequence of the interrupt event => kobj method series is at f1ae360e6ad094070109361f8c8f4a5165056a3a.  INTRNG is adapted at 8f4f337bfe23847eee49153947a3dcd002b86c4b, while PowerPC goes to f0cec98adc49af8d32943f064cde4478d747959b.

The interrupt event => kobj method looks kind of overdue.  It may take a _lot_ of interrupts for that to reach break-even for x86, but for all other architectures the benefit is immediately rather compelling.

ehem:
This significant (therefore large and with many steps) forward series is still here for now.  The underlying idea is pretty simple and the gains are large, but it does need review.

The x86 path now terminates at 356c719afc4e92b19a902ec1ccc87a788034813b for the device-first approach.  Otherwise a5096d39fb8884ba09ee6a6f2b9b9677e2dd392b for the core-first approach.

The interrupt event => kobj method series finishes at a2e6ddccd6434626494aede940e7055324f57dc8.  INTRNG is adapted at 9e17cde719844655e0d896291a7bb1c2e6a9865c, while PowerPC goes to cb3e91b6a64cc91ab766c6f64ca89da3ec3b7362.  f86e470edacac046dd5f6e2a319afbfbde283f3c is the tweak for SWIs.

Finally figured out what needed to be done about `intr_handler_source()`.  Now need a reviewer to choose which direction to take at a few points and bring into FreeBSD's repository.

ehem:
Rebased, no significant changes.

The x86 path now terminates at c1e7dde0af23931a8645fcf81c9072738e174527 for the device-first approach. Otherwise 5825a5e88003fc06682b8fd4e7766e08c969cd67 for the core-first approach.

The interrupt event => kobj method series finishes at 8957858faef021f7c027ec2b411dc428af01a3b9.  INTRNG is adapted at 526425636cbfa99a91cfcf8b743f57eca233d2ad, while PowerPC goes to 5d06dff1950fe6dd449be65ad9843235883d6505.  013ea0bf87c587a2808bc2338026f01330317900 is the tweak for SWIs.

Quite a few of the preparation steps can be done without significant effects elsewhere.  Unfortunately that does need someone with commit access...

ehem:
Hmm, someone else broke the build.

ehem:
Rebased, no changes.  Simply let the CI system retry since last time caught someone's goof.

The x86 path now terminates at e4b78f4dbde46ed45f8d94a72975272ca0e1c171 for the device-first approach. Otherwise 30a47eeb0628c19feecb4a30ce0fd4d90f3c278e for the core-first approach.

The interrupt event => kobj method series finishes at 6f9069dd92a27c20d9dc235ca5a6e0de74fb0f51. INTRNG is adapted at 5a820e41fc8781e3135e125ab84955a988475078, while PowerPC goes to a8bfe9151c464e61d1d1769bb6e33b52c92cb0be.  6c7ee03d5d2c53ced0ae924ba44edf45ec6f7395 is the tweak for SWIs.

Quite a few of the preparation steps can be done without significant effects elsewhere. Unfortunately that does need someone with commit access...

ehem:
Rebased, small update for minor upstream changes.

The x86 path now terminates at 9304d34e8a7f91b7f850d16e797386fdae8efd5b for the device-first approach. Otherwise badb2fbb04bde84f553b4d011206a4049078abb3 for the core-first approach.

The interrupt event => kobj method series finishes at f0ef8be3c6d6981c834782e84731feb06cfd8d0a. INTRNG is adapted at 1990bc180eaa789fced3a73325afce14d67c6304, while PowerPC goes to 852b5172a425f38e8931073dbe0c3882fe85ce99.  e81bc41461690f4ebcfbd7675f2cf46597f604e1 is the tweak for SWIs.

Quite a few of the preparation steps can be done without significant effects elsewhere. Unfortunately that does need someone with commit access...

PR #1459, opened 2024-10-10
manuals: Misc syntax fixes

gperciva:


gperciva:
These were reported by `mandoc -T lint ...` as errors.  The ascii output is not affected by this commit.  There are some minor changes to the html.

@mhorne.

gperciva:
> [gperciva](https://github.com/gperciva) requested review from [bsdjhb](https://github.com/bsdjhb) and [bsdimp](https://github.com/bsdimp) as [code owners](https://github.com/freebsd/freebsd-src/blob/d5e9faf1f3291d0af9dcdd53ccdaf0971e5d8f63/.github/CODEOWNERS#L116) [2 minutes ago](https://github.com/freebsd/freebsd-src/pull/1459#event-14581141535)

My apologies, I think I triggered something accidentally.  @mhorne asked me to submit these fixes via github, and he's been reviewing&merging my PRs.

sbin/nvmecontrol/nvmecontrol.8
```diff
@@ -998,13 +998,13 @@ If the
 .Dq nvmecontrol identify -n 1 nvme2
 command ended with
 .Pp
-.Bl -verbatim
+.Bd -literal
```
gperciva, line 1001: There's no `.Bl -verbatim`; it seems clear to me that the original author intended `.Bd -literal` here.

(This results in different `<div>` classes and a `<pre>` in the html output, but that's a desired change.  No affect in the ascii output.)

share/man/man4/pfsync.4
```diff
@@ -147,7 +147,6 @@ E.g.:
 # ifconfig pfsync0 version 1301
 .Ed
 .Pp
-.Bd
```
gperciva, line 150: This tag was ignored because it didn't have a `-type`.  None of the display types really seemed necessary, so I opted for keeping the output unchanged (while fixing the syntax error).

stand/libsa/libsa.3
```diff
@@ -597,6 +597,7 @@ string after the device specification is written.
 .It Xo
 .Ft int
 .Fn devinit void
+.Xc
```
gperciva, line 600: This causes the other difference in the html output: the "Calls all the..." sentence is no longer part of the `<dt>` (which it shouldn't be).

Adding the `.Xc` brings `int devinit(void)` in line with all the other functions in this file.

mhorne, line 600:
```suggestion
.Xc
.Pp
```
This is also needed for consistency with the other items.

mhorne (changes requested):

stand/libsa/libsa.3
gperciva: Thanks!  Fixed in my git, but I've removed the changes to that file out of this PR.

I also removed the changes to `nvmecontrol.8`.

This PR is now a "no changes at all to the rendered output"; the next PR will have the fixes I removed (and other output-changing fixes).

mhorne (approved):

PR #1460, opened 2024-10-11
ice(4): properly use kernel option RSS

brian90013:
Header ice_rss.h uses the kernel RSS interface if option RSS is defined. However when ice_rss.h is included by ice_lib.h there is no prior inclusion of ice_opts.h to set RSS causing ifdef RSS to always fail. Add ice_opts.h to the top of ice_lib.h (like ice_iflib.h) so RSS can be defined when ice_rss.h is parsed.

With that in place, compilation fails due to a missing defintion of ICE_DEFAULT_RSS_HASH_CONFIG. It is defined in ice_rss.h only when RSS is not defined. Since this define is not part of the kernel RSS interface but ice-specific, it should always be defined. Move its definition outside of ifdef RSS.

This was reported as [bugzilla 255309](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=255309) back in April 2021. erj said in June 2021, "This patch looks valid to me. I'll get it in after this update" but it never was merged. I've been using this patch ever since.

### DIFF
--- a/sys/dev/ice/ice_lib.h
+++ b/sys/dev/ice/ice_lib.h
@@ -40,6 +40,9 @@
 #ifndef _ICE_LIB_H_
 #define _ICE_LIB_H_
 
+/* include kernel options first */
+#include "ice_opts.h"
+
 #include <sys/types.h>
 #include <sys/bus.h>
 #include <sys/rman.h>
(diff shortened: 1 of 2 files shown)


mhorne (approved):
LGTM.

The general pattern is to place `#include opt_*.h` lines at the very top of a `.c` file; they are less common to appear in headers. However I think what you have done is preferable, both for consistency with `ice_iflib.h`, and since nothing in `ice_lib.c` checks for these kernel options.

concussious:
Cc @ricera

mhorne:
Thanks Brian for the fix, and for your patience!

PR #1466, opened 2024-10-13
manuals: Fix position of SPDX license line

gperciva:


gperciva:
A very low-priority PR.

Based on @bsdimp's comment [1], I added a lint check to my doc scripts [2] to check the SPDX line.  Since there were only 9 non-contrib files that were incorrect, I manually edited them.

[1] https://github.com/freebsd/freebsd-src/pull/1458#discussion_r1796088898
[2] https://github.com/Tarsnap/freebsd-doc-scripts/blob/main/lint.py#L14

bsdimp:
These are all likely good, but their sources are (a) jhb and (b) the foundation, so we may want to chat with john and likely ed or someone he redirects us to... and it would likely be good for me to create a canonical source of this guidance.

gperciva:
Sure thing!  I viewed this as starting / continuing a conversation, not the end point.

FWIW, the only problems in contrib/ are the 32 files in `contrib/libfido2/man/*.[1-9]`.  Everything else passes the linter with no complaints.

bsdimp:
@bsdjhb can we do this?

PR #1467, opened 2024-10-14
manuals: Fix some "missing section argument"

gperciva:


gperciva:
These were reported by `mandoc -T lint ...` as warnings.

This PR fixes roughly 2/3 of these warnings; the remaining ones were not immediately obvious to me.  Also, I figured the PR was already nudging the boundary of being too long.

@mhorne

share/man/man7/mitigations.7
```diff
@@ -189,13 +189,13 @@ is a FreeBSD-specific extension to
...
-.Xr mmap
+.Xr mmap 2
 and later altered by
-.Xr mprotect .
+.Xr mprotect 2 .
 For example, memory allocated originally with an mmap prot argument of
 PROT_MAX(PROT_READ | PROT_WRITE) | PROT_READ
 may be made writable by a future
-.Xr mprotect
+.Xr mprotect 2
```
concussious, line 198: These four are already addressed in https://github.com/freebsd/freebsd-src/pull/1428, please don't make me redo it.

concussious (changes requested):
Everything else looks great though!

share/man/man7/mitigations.7
gperciva: If the changes are exactly the same (and they appear to be!), then git will handle it automagically, so there wouldn't be any need to redo anything.

concussious (approved):

PR #1470, opened 2024-10-14
fusefs: support FUSE_IOCTL

CismonX:


tests/sys/fs/fusefs/ioctl.cc
```diff
@@ -0,0 +1,190 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2019 The FreeBSD Foundation
+ *
+ * Portions of this software was developed by BFF Storage Systems, LLC under
+ * sponsorship from the FreeBSD Foundation.
```
asomers, line 7: You should update this copy/pasted copyright section.

asomers (commented):
Interesting PR!  Do you know of any real-world fuse file systems that implement FUSE_IOCTL?  I couldn't find any back in 2019, which is why I never implemented this command.

asomers:
> Recently I came across two FUSE filesystems that implement `FUSE_IOCTL`.
> 
> One of them is bindfs. When launched with the `-o enable-ioctl` option, it enables ioctl forwarding.
> 
> However, it only works for the ioctls that do not require copying extra data between kernel and user address space, since FUSE servers are not allowed to perform "unrestricted ioctl".

That's certainly an interesting use case.

> 
> Another is ntfs-3g. The only supported ioctl is `FITRIM`.
> 
> Although there is currently no `FITRIM` equivalent on FreeBSD, it's not difficult to port: Just adding a new entry in `sys/filio.h`, and porting `fstrim(8)` from util-linux would suffice.
> 
> Note: For the TRIM feature, ntfs-3g server requires `BLKDISCARD` internally. Since we already have `DIOCGDELETE` as its equivalent, this won't be a problem.

That wouldn't work very well, because few to no FreeBSD applications will use FITRIM.  A far better option would be for ntfs-3g to switch from libfuse2 to libfuse3.  Then it could use the native FUSE_FALLOCATE request for hole-punching.

tests/sys/fs/fusefs/ioctl.cc
```diff
@@ -0,0 +1,187 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2024 The FreeBSD Foundation
```
asomers, line 4: As emaste said, no copyright assignment is expected.  Better to put your own name on this line.

sys/fs/fuse/fuse_vnops.c
```diff
@@ -368,6 +368,84 @@ fuse_inval_buf_range(struct vnode *vp, off_t filesize, off_t start, off_t end)
...
+	fdisp_make_vp(&fdi, FUSE_IOCTL, vp, td, cred);
+	fii = fdi.indata;
+	fii->fh = fufh->fh_id;
+	fii->flags = flags;
+	fii->cmd = cmd;
+	fii->arg = (uintptr_t)arg;
+	fii->in_size = insize;
+	fii->out_size = outsize;
+	if (insize > 0)
+		memcpy((char *)fii + sizeof(*fii), arg, insize);
```
asomers, line 416: Is this copy coming from userspace?  If so you should use `copying` instead of `memcpy`.

sys/fs/fuse/fuse_vnops.c
```diff
@@ -368,6 +368,84 @@ fuse_inval_buf_range(struct vnode *vp, off_t filesize, off_t start, off_t end)
...
+
+	fio = fdi.answ;
+	if (fdi.iosize > sizeof(*fio)) {
+		size_t realoutsize = fdi.iosize - sizeof(*fio);
+
+		if (realoutsize > outsize) {
+			err = EIO;
+			goto out;
+		}
+		memcpy(arg, (char *)fio + sizeof(*fio), realoutsize);
```
asomers, line 433: Similarly, this must be `copyout` if `arg` points to userspace.

sys/fs/fuse/fuse_vnops.c
```diff
@@ -368,6 +368,84 @@ fuse_inval_buf_range(struct vnode *vp, off_t filesize, off_t start, off_t end)
...
+	struct fuse_dispatcher fdi;
+	struct fuse_ioctl_in *fii;
+	struct fuse_ioctl_out *fio;
+	struct fuse_filehandle *fufh;
+	uint32_t flags = 0;
+	uint32_t insize = 0;
+	uint32_t outsize = 0;
+	int err;
+
+	err = fuse_filehandle_get_anyflags(vp, &fufh, cred, td->td_proc->p_pid);
```
asomers, line 385: Should this be fuse_filehandle_getrw instead?

asomers (changes requested):
You should send FUSE_HAS_IOCTL_DIR during fuse_internal_send_init.
This looks like a good start, but I haven't reviewed the tests yet.

tests/sys/fs/fusefs/ioctl.cc
emaste: (That said, if there is a reason you do not want to be the copyright holder I can discuss that separately.)

sys/fs/fuse/fuse_vnops.c
asomers: That's true for all _good_ fuse file systems.  But one (anti-, IMHO) feature of the fuse protocol is that the file system itself can choose to take responsibility for permission checks on open(). That's why fuse file handles exist.  So we need to be careful to get a file handle of the correct type.

sys/fs/fuse/fuse_internal.c
```diff
@@ -1115,7 +1114,7 @@ fuse_internal_send_init(struct fuse_data *data, struct thread *td)
 	 * FUSE_MAX_PAGES: not yet implemented
 	 */
 	fiii->flags = FUSE_ASYNC_READ | FUSE_POSIX_LOCKS | FUSE_EXPORT_SUPPORT
-		| FUSE_BIG_WRITES | FUSE_WRITEBACK_CACHE
+		| FUSE_BIG_WRITES | FUSE_HAS_IOCTL_DIR | FUSE_WRITEBACK_CACHE
```
asomers, line 1117: I think we also need to check for FUSE_HAS_IOCTL_DIR during fuse_internal_init_callback.  If not present, then we should return ENOTTY whenever a user tries to use an ioctl on a directory.

tests/sys/fs/fusefs/ioctl.cc
```diff
@@ -0,0 +1,187 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2024 CismonX <admin@cismon.net>
```
asomers, line 4: Is CismonX your legal name?  An alias cannot own a copyright.

asomers (changes requested):

tests/sys/fs/fusefs/ioctl.cc
bsdimp: Legally they can. Pseudonyms are explicitly allowed. The project strongly prefers a legal entity, though.

CismonX:
Just rebased HEAD without other changes.
Fortunately the patch still works :)

sys/fs/fuse/fuse_vnops.c
CismonX: I finally get what `fuse_filehandle_getrw()` is for:

When opening a vnode with `O_RDONLY` or `O_WRONLY`, if there is an existing file handle opened with `O_RDWR`, that file handle is reused, instead of creating a new one.

So you're right, using `fuse_filehandle_getrw()` here is the correct thing to do. Sorry it takes me so long to figure this out.

Patch updated (and squashed into a single commit).

bsdimp:
@asomers this looks reasonable to my eye. Any reason I shouldn't push it into head before 15?

tests/sys/fs/fusefs/ioctl.cc
asomers: I stand corrected, @imp .  And @CismonX , multibyte characters are not a problem in the source code, as long as they're UTF-8.

asomers (commented):
The only problem, @bsdimp, is that the tests use a different style than the rest of the fusefs test suite.  That makes them harder for me to review.  That's why I've been procrastinating on this for so long.  If `with_expect_ioctl_rw` and friends were inlined it would match the existing style better.  I would prefer that, even if it leads to a little bit of code duplication.

CismonX:
@asomers I've refactored the tests as you suggested.

tests/sys/fs/fusefs/ioctl.cc
CismonX: CONTRIBUTING.md now [explicitly allows pseudonyms](https://github.com/freebsd/freebsd-src/blob/main/CONTRIBUTING.md#author-name-and-email).

If that also applies to copyright notices, this shouldn't be a problem any more.

asomers:
I'd like you to know that I haven't been ignoring this.  I've been looking at it today, and adding a few more test cases, which I'll push soon.  Sometime next week, I think.

CismonX:
@asomers

15.0-ALPHA has been around for a while, and I'm wondering if this patch could still make it into 15.
If there are anything else I need to do, please let me know.

bsdimp:
Sorry for the delay... This looks ready, modulo a merge conflict that I've resolved, I think, correctly.

PR #1471, opened 2024-10-15
wsp/evdev/sysmouse: Add wsp(4) horizontal scrolling, improve new tunables, and improve documentation.

MegaManSec:
This PR works on 6 issues:

1) Separates much of the wsp.c constants to a separate header file. This allows for, for example, the `MAX_FINGER_ORIENTATION` macro to be referenced anywhere in the wsp.c file.
2) Changes the `scr_hor_threshold` tunable default. Previously, it was so low than even a few millimeters movement of two fingers on the mousepad would register as a right/left t-axis mouse click.
3) Creates a new tunable, `max_scroll_finger_distance`, which can be used to specify the maximum distance between two fingers which can be used for a z-axis movement. This information and some other missing details in the manual have also been included.
4) Adds a new T-axis reporting method to sysmouse(4) for the wsp(4). This allows wsp to report variable distances of movement in the t-axis by using a fake button-5 press.
6) Add support for horizontal (t-axis) scrolling to wsp(4), and separate forward/backwards swiping from that functionality, which now uses three-fingers instead of two by default -- however this is variable with new tunables.
7) Vastly improve documentation and explanations of each tunable.

Individual commits have longer explanations.

### DIFF
--- a/share/man/man4/wsp.4
+++ b/share/man/man4/wsp.4
@@ -107,8 +107,8 @@ Enables movement on the trackpad follow a partially-released left-click.
 Default is 1 (enabled).
 .El
 .Bl -tag -width indent
-.It Va hw.usb.wsp.max_finger_area
-Specifies the maximum area on the trackpad which is registered as a
+.It Va hw.usb.wsp.max_finger_diameter
+Specifies the maximum finger diameter on the trackpad that is registered as a
 finger (a lower value is used for palm detection). Default is 1900.
 .El
 .Bl -tag -width indent
--- a/share/man/man4/wsp.4
+++ b/share/man/man4/wsp.4
@@ -63,24 +63,39 @@ through nodes under
 Pointer sensitivity can be controlled using the sysctl tunable
 .Nm hw.usb.wsp.scale_factor .
 Tap to left-click can be controlled using the sysctl tunable
-.Nm hw.usb.wsp.enable_single_tap_clicks ,
-set to 0 to disable single tap clicks or 1 to enable them (default).
+.Nm hw.usb.wsp.enable_single_tap_clicks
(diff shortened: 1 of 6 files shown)


MegaManSec:
Note that I lowered the default threshold again from commit after ef64180633f43335b9e8f635e5451c1dbc89d032. This is because the swipe forward/back movement is now using (by default) three-fingers, which is likely much less common to be using than two-fingers (therefore will not accidentally be executed while attempting to do a two-finger z-axis/vertical scroll).

MegaManSec:
I'm converting this to a draft because I'm experiencing a full-system freeze when using it. Since I'm not in ttyv0, I cannot see the error messages (if any).
I've looked at the source and can't see any reason the system should be freezing, but will think about this more.

MegaManSec:
Changing back to a completed PR. The full system freeze is not due to this code.

Finally, I have also clarified that the sysctl `hw.usb.wsp.max_finger_area` refers to the maximum diameter of a finger, not any "area". Since this sysctl is a _change_, it is a breaking change for users.

wulf7:
https://github.com/freebsd/freebsd-src/pull/1471/commits/979f53f9135eebc6abbb496675f37820c8c45062 is wrong. It is bad idea to initialize array in a header file.

Other commits LGTM.

MegaManSec:
Sure. Fixed!

wulf7:
LGTM

wulf7 (approved):

concussious (commented):
Hey! Can you change these Nm's to Va? Nm is only for the name of the topic, but it sometimes gets abused to create bold. Sy makes bold, but since these are sysctls, they should be Va.

wulf7:
Committed. Thanks!

PR #1472, opened 2024-10-15
manuals: Misc syntax fixes

gperciva:


gperciva:
These were reported by `mandoc -T lint ...` as errors. @mhorne 

Once this is merged, there will be only 3 detected errors [1] left in the non-contrib / non-insane man pages [2].

[1] I'm defining "detected errors" as "detected on 14.1".  1 of those errors is `St -p1003.1-2024`, which is included in mandoc in 15.0-CURRENT but not 14.1.

[2] full list of excluded man pages is https://github.com/Tarsnap/freebsd-doc-scripts/blob/main/generate-man-problem-lists.sh#L20.  I'm not ruling out re-examining the problematic man pages at some point, but it's not a priority.

stand/man/loader.efi.8
```diff
@@ -393,10 +393,10 @@ Boot to FW : false
 BootCurrent: 0001
 Timeout    : 2 seconds
 BootOrder  : 0000, 0001, 0003, 0004, 0005, 0006, 0001, 0008, 000A, 000B, 000C, 000E, 0007
-...
+\&...
 +Boot0001* FreeBSD ZPOOL HD(1,GPT,b5d0f86b-265d-1e1b-18aa-0ed55e1e73bd,0x28,0x96000)/File(\EFI\FREEBSD\LOADER.EFI)
```
gperciva, line 397: man interprets `\E`, `\F`, and `\L` as special commands.  I have a fix waiting for another PR; I don't think it belongs in this one since the commit says that they were reported by `mandoc`.

If you'd rather that I include the fixes now, I could make the message "Most of these were reported by `mandoc...`".

sbin/ipf/ipf/ipf.5
```diff
@@ -295,7 +295,7 @@ proto-unr
 (protocol unreachable)
 the IP protocol specified in the packet is not available to receive
 packets.
-.DE
+.RE
```
gperciva, line 298: There's no `.DE` in man(7).

Given the context, I think this should be `.RE` to close the `.RS` on line 257.

sbin/ipf/ipf/ipf.8
```diff
@@ -158,7 +158,7 @@ Zero global statistics held in the kernel for filtering only (this doesn't
 affect fragment or state statistics).
 .DT
 .SH ENVIRONMENT
-.NM utilizes the following environment variable.
+ipf utilizes the following environment variable.
```
gperciva, line 161: There's no `.NM` in man(7).  The author was clearly thinking of mandoc; for plain old man, we have to write out the name.

concussious, line 161:
```suggestion
.Nm
utilizes the following environment variable.
```

concussious: Sorry for the noise, I was confused because mandoc(1) is the compiler which speaks both man(7) and mdoc(7) roff(7) dialects (plus some others like tbl(7), roff(7) is an amazing language).

gperciva: I haven't seen your "earlier deleted comment", but this comment made me realize that I'd written "mandoc" instead of "mdoc(7)".

I've edited my earlier comment; hopefully it makes more sense now.

gperciva:
BTW @concussious : could I convince you to fix two `-width` places?  They're the remaining 2 errors.

```
mandoc: sbin/setkey/setkey.8:284:20: ERROR: skipping excess arguments: Bl ... natt_mtu
mandoc: share/man/man4/pci.4:374:20: ERROR: skipping excess arguments: Bl ... struct
```

In both cases, somebody thought that they could use formatting commands for the argument to `-width`:

```
.Bl -tag -width Fl natt_mtu -compact
```
```
.Bl -tag -width Vt struct pcise pbm_sel
```

I could just slap a `-width indent` in there, but I think a better solution would be to have a look at the lengths of the lists with different numbers of columns, but that's the kind of subjective judgement that I prefer to avoid.

concussious:
I'll take a look when I get home. Thanks for inviting me!

sbin/ipf/ipf/ipf.8
concussious, line 161:
```suggestion
.B ipf
utilizes the following environment variable.
```

concussious (changes requested):

sbin/ipf/ipf/ipf.8
gperciva: Thanks, adjusted.

concussious (approved):

stand/man/loader.efi.8
mhorne: I'll take what is here now and you can submit the other fix later :+1:

mhorne (approved):

PR #1480, opened 2024-10-17
libxo support for devinfo (take 2)

ktullavik:
Since last time I have
* Fixed issues raised by @bsdimp in my last pull request
* Dealt with recent change to libxo by jhb
* Reworked the -u path, as it used duplicate json keys in certain cases
* Tried to make a cleaner history, doing more prep work up front
* Fixed a few small things discovered along the way

Many of the prep changes only makes sense once libxo are added.
I had to do quite a few contortions to retrofit libxo into this thing.

### DIFF
--- a/usr.sbin/devinfo/devinfo.c
+++ b/usr.sbin/devinfo/devinfo.c
@@ -255,6 +255,17 @@ print_rman(struct devinfo_rman *rman, void *arg __unused)
 	return(0);
 }
 
+static void
+print_device_path_entry(struct devinfo_dev *dev)
+{
+	const char *devname = dev->dd_name[0] ? dev->dd_name : "unknown";
+
+	printf("%s", devname);
+	print_device_props(dev);
+	if (vflag)
+		printf("\n");
+}
+
 static int
 print_device_path(struct devinfo_dev *dev, void *xname)
 {
@@ -262,20 +273,14 @@ print_device_path(struct devinfo_dev *dev, void *xname)
 	int rv;
 
 	if (strcmp(dev->dd_name, name) == 0) {
-		printf("%s", dev->dd_name[0] ? dev->dd_name : "unknown");
-		print_device_props(dev);
-		if (vflag)
-			printf("\n");
+		print_device_path_entry(dev);
 		return (1);
 	}
 
 	rv = devinfo_foreach_device_child(dev, print_device_path, xname);
 	if (rv == 1) {
 		printf(" ");
-		printf("%s", dev->dd_name[0] ? dev->dd_name : "unknown");
-		print_device_props(dev);
-		if (vflag)
-			printf("\n");
+		print_device_path_entry(dev);
 	}
 	return (rv);
 }
--- a/usr.sbin/devinfo/devinfo.c
+++ b/usr.sbin/devinfo/devinfo.c
@@ -155,12 +155,19 @@ print_dev(struct devinfo_dev *dev)
 {
 
 	printf("%s", dev->dd_name[0] ? dev->dd_name : "unknown");
-	if (vflag && *dev->dd_desc)
-		printf(" <%s>", dev->dd_desc);
-	if (vflag && *dev->dd_pnpinfo)
-		printf(" pnpinfo %s", dev->dd_pnpinfo);
-	if (vflag && *dev->dd_location)
-		printf(" at %s", dev->dd_location);
+
+	if (vflag) {
+		if (*dev->dd_desc) {
+			printf(" <%s>", dev->dd_desc);
+		}
+		if (*dev->dd_pnpinfo) {
+			printf(" pnpinfo %s", dev->dd_pnpinfo);
+		}
+		if (*dev->dd_location) {
+			printf(" at %s", dev->dd_location);
+		}
+	}
+
 	if (!(dev->dd_flags & DF_ENABLED))
 		printf(" (disabled)");
 	else if (dev->dd_flags & DF_SUSPENDED)
--- a/usr.sbin/devinfo/devinfo.c
+++ b/usr.sbin/devinfo/devinfo.c
@@ -45,14 +45,17 @@
 static bool	rflag;
 static bool	vflag;
 
+static void	print_indent(int);
 static void	print_resource(struct devinfo_res *);
 static int	print_device_matching_resource(struct devinfo_res *, void *);
 static int	print_device_rman_resources(struct devinfo_rman *, void *);
+static void	print_dev(struct devinfo_dev *);
 static int	print_device(struct devinfo_dev *, void *);
 static int	print_rman_resource(struct devinfo_res *, void *);
 static int	print_rman(struct devinfo_rman *, void *);
 static int	print_device_path(struct devinfo_dev *, void *);
 static void	print_path(struct devinfo_dev *, char *);
+static void	usage(void);
 
 struct indent_arg
 {
--- a/usr.sbin/devinfo/devinfo.c
+++ b/usr.sbin/devinfo/devinfo.c
@@ -31,7 +31,8 @@
  * Print information about system device configuration.
  */
 
-#include <sys/types.h>
+#include <sys/param.h>
+
 #include <err.h>
 #include <errno.h>
 #include <stdbool.h>
@@ -56,6 +57,20 @@ struct indent_arg
 	int	indent;
 	void	*arg;
 };
+
+
+static void
+print_indent(int n)
+{
+	static char	buffer[1024];
+
+	if (n < 1)
+		return;
+	n = MIN((size_t)n, sizeof(buffer) - 1);
+	memset(buffer, ' ', n);
+	buffer[n] = '\0';
+	printf("%s", buffer);
+}
 
 /*
  * Print a resource.
@@ -86,14 +101,12 @@ print_device_matching_resource(struct devinfo_res *res, void *arg)
 {
 	struct indent_arg	*ia = (struct indent_arg *)arg;
 	struct devinfo_dev	*dev = (struct devinfo_dev *)ia->arg;
-	int			i;
 
 	if (devinfo_handle_to_device(res->dr_device) == dev) {
 		/* in 'detect' mode, found a match */
 		if (ia->indent == 0)
 			return(1);
-		for (i = 0; i < ia->indent; i++)
-			printf(" ");
+		print_indent(ia->indent);
 		print_resource(res);
 		printf("\n");
 	}
@@ -107,7 +120,7 @@ int
 print_device_rman_resources(struct devinfo_rman *rman, void *arg)
 {
 	struct indent_arg	*ia = (struct indent_arg *)arg;
-	int			indent, i;
+	int			indent;
 
 	indent = ia->indent;
 
@@ -117,8 +130,7 @@ print_device_rman_resources(struct devinfo_rman *rman, void *arg)
 	    print_device_matching_resource, ia) != 0) {
 
 		/* there are, print header */
-		for (i = 0; i < indent; i++)
-			printf(" ");
+		print_indent(indent);
 		printf("%s:\n", rman->dm_desc);
 
 		/* print resources */
@@ -155,12 +167,11 @@ static int
 print_device(struct devinfo_dev *dev, void *arg)
 {
 	struct indent_arg	ia;
-	int			i, indent;
+	int			indent;
 
 	if (vflag || (dev->dd_name[0] != 0 && dev->dd_state >= DS_ATTACHED)) {
 		indent = (int)(intptr_t)arg;
-		for (i = 0; i < indent; i++)
-			printf(" ");
+		print_indent(indent);
 		print_dev(dev);
 		printf("\n");
 		if (rflag) {
--- a/usr.sbin/devinfo/devinfo.c
+++ b/usr.sbin/devinfo/devinfo.c
@@ -150,7 +150,7 @@ print_dev(struct devinfo_dev *dev)
 /*
  * Print information about a device.
  */
-int
+static int
 print_device(struct devinfo_dev *dev, void *arg)
 {
 	struct indent_arg	ia;
--- a/usr.sbin/devinfo/devinfo.c
+++ b/usr.sbin/devinfo/devinfo.c
@@ -80,13 +80,21 @@ print_resource(struct devinfo_res *res)
 {
 	struct devinfo_rman	*rman;
 	bool			hexmode;
+	rman_res_t		end;
 
 	rman = devinfo_handle_to_rman(res->dr_rman);
 	hexmode =  (rman->dm_size > 1000) || (rman->dm_size == 0);
-	printf(hexmode ? "0x%jx" : "%ju", res->dr_start);
-	if (res->dr_size > 1)
-		printf(hexmode ? "-0x%jx" : "-%ju",
-		    res->dr_start + res->dr_size - 1);
+	end = res->dr_start + res->dr_size - 1;
+
+	if (hexmode) {
+		printf("0x%jx", res->dr_start);
+		if (res->dr_size > 1)
+			printf("-0x%jx", end);
+	} else {
+		printf("%ju", res->dr_start);
+		if (res->dr_size > 1)
+			printf("-%ju", end);
+	}
 }
 
 /*
--- a/usr.sbin/devinfo/devinfo.c
+++ b/usr.sbin/devinfo/devinfo.c
@@ -183,8 +183,10 @@ print_device(struct devinfo_dev *dev, void *arg)
 {
 	struct indent_arg	ia;
 	int			indent;
+	bool			printit = vflag || (dev->dd_name[0] != 0 &&
+				    dev->dd_state >= DS_ATTACHED);
 
-	if (vflag || (dev->dd_name[0] != 0 && dev->dd_state >= DS_ATTACHED)) {
+	if (printit) {
 		indent = (int)(intptr_t)arg;
 		print_indent(indent);
 		print_dev(dev);
--- a/usr.sbin/devinfo/devinfo.c
+++ b/usr.sbin/devinfo/devinfo.c
@@ -51,6 +51,7 @@ static int	print_device_rman_resources(struct devinfo_rman *, void *);
 static int	print_device(struct devinfo_dev *, void *);
 static int	print_rman_resource(struct devinfo_res *, void *);
 static int	print_rman(struct devinfo_rman *, void *);
+static int	print_device_path(struct devinfo_dev *, void *);
 
 struct indent_arg
 {
@@ -254,7 +255,7 @@ print_rman(struct devinfo_rman *rman, void *arg __unused)
 }
 
 static int
-print_path(struct devinfo_dev *dev, void *xname)
+print_device_path(struct devinfo_dev *dev, void *xname)
 {
 	const char *name = xname;
 	int rv;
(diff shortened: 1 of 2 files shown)


ktullavik:
Seems I forgot to put "devinfo:" in front of some of my commit messages. I hope this is easily fixed if someone decides to commit it.

ktullavik:
There's two typos in the commit message for a2a3a74.
Should be:
devinfo: Put expression in variable in prep for libxo

usr.sbin/devinfo/devinfo.c
```diff
@@ -57,6 +57,18 @@ struct indent_arg
...
 
+
+static void
+print_indent(int n)
+{
+	char *s;
+
+	if (n < 1)
+		return;
+	asprintf(&s, "%*s", n, " ");
```
rilysh: `asprintf()` causes allocation, but it doesn't seem like there's any need for this?
equivalent to
```
	while (n--)
		putchar(' ');
```

usr.sbin/devinfo/devinfo.c
```diff
@@ -70,8 +75,38 @@ print_indent(int n)
...
+/*
+ * Takes a list of key-value pairs in the form
+ * "key1=val1 key2=val2 ..." and prints them according
+ * to xo formatting.
+ */
+static void
+print_kvlist(char *s)
+{
+	char *kv;
+	char *copy = strdup(s);
```
rilysh, line 92: `strdup()` just a wrapper over `malloc()` and `memcpy()`, so it should be checked if it returns NULL

usr.sbin/devinfo/devinfo.c
```diff
@@ -83,21 +118,31 @@ print_resource(struct devinfo_res *res)
 	struct devinfo_rman	*rman;
 	bool			hexmode;
 	rman_res_t		end;
+	char			*safe_desc;
 
 	rman = devinfo_handle_to_rman(res->dr_rman);
 	hexmode =  (rman->dm_size > 1000) || (rman->dm_size == 0);
 	end = res->dr_start + res->dr_size - 1;
 
+	safe_desc = (char*) malloc(strlen(rman->dm_desc));
```
rilysh, line 128: cast to `char *` is useless here. `malloc()` errors should be checked here too

usr.sbin/devinfo/devinfo.c
```diff
@@ -83,21 +118,31 @@ print_resource(struct devinfo_res *res)
...
+	strcpy(safe_desc, rman->dm_desc);
```
rilysh, line 129: `strlcpy()` seems better here, with the previously calculated length of rman->dm_desc. maybe length + 1 (in malloc()) is needed (for the null terminator).

usr.sbin/devinfo/devinfo.c
```diff
@@ -311,6 +434,11 @@ main(int argc, char *argv[])
 	bool			uflag;
 	char			*path = NULL;
 
+	argc = xo_parse_args(argc, argv);
+	if (argc < 0) {
```
rilysh, line 439: how argument count can be lower than zero? program name is counted as an argument as well.

usr.sbin/devinfo/devinfo.c
ktullavik: It's done this way to minimize diff when libxo is added later. With libxo we want to emit the whole padding in one call. Otherwise the html get spammed with <div class="padding"> elements.

usr.sbin/devinfo/devinfo.c
ktullavik: xo_parse_args returns -1 on error.

usr.sbin/devinfo/devinfo.c
ktullavik: This good?

```
	char *copy = strdup(s);

	if (copy == NULL) {
		xo_err(1, "No memory!");
		exit(1);
	}
```

usr.sbin/devinfo/devinfo.c
ktullavik: Like this?

```
	char			*safe_desc;
	int			safe_desc_len;

	rman = devinfo_handle_to_rman(res->dr_rman);
	hexmode =  (rman->dm_size > 1000) || (rman->dm_size == 0);
	end = res->dr_start + res->dr_size - 1;

	safe_desc_len = strlen(rman->dm_desc) + 1;
	safe_desc = malloc(safe_desc_len);
	if (safe_desc == NULL) {
		xo_err(1, "No memory!");
		exit(1);
	}
	strlcpy(safe_desc, rman->dm_desc, safe_desc_len);
```

usr.sbin/devinfo/devinfo.c
ktullavik: Noted. See below.

usr.sbin/devinfo/devinfo.c
bsdimp: still, you need to free(s) here...

usr.sbin/devinfo/devinfo.c
rilysh: Yeah (with bsd style)
```
char *copy;
if ((copy = strdup(s)) == NULL)
        xo_err(1, "No memory!");
```

usr.sbin/devinfo/devinfo.c
rilysh: Yes like this. Also `safe_desc_len` should be a size_t not an int. I think you can rename the variable just len (safe_desc_len seems too long)

usr.sbin/devinfo/devinfo.c
rilysh: Oh that's interesting. I just saw the man page of this function and you're absolutely right!

usr.sbin/devinfo/devinfo.c
ktullavik: Fixed. Along with a few other cases. Sorry, I'm not used to C.

usr.sbin/devinfo/devinfo.c
ktullavik: Fixed

usr.sbin/devinfo/devinfo.c
ktullavik: Fixed

usr.sbin/devinfo/devinfo.c
ktullavik: Fixed. Thanks!

ktullavik:
I'm learning about git rebase --interactive. Hold your horses.

ktullavik:
Ok, I've spent some time learning new git tricks and cleaning up the history. Let me know if there is anything else @bsdimp @rilysh

usr.sbin/devinfo/devinfo.c
```diff
@@ -57,6 +57,19 @@ struct indent_arg
...
 
+
+static void
+print_indent(int n)
+{
+	char *s;
+
+	if (n < 1)
+		return;
+	asprintf(&s, "%*s", n, " ");
```
bsdimp: This is going to be allocating and freeing this a lot (once per line).
And asprintf is an expensive way to do this.
And we don't need to be thread safe here.
It would be better to have a 1k static buffer. then this becomes:
n = min(n, sizeof(buffer) - 1); memset(buffer, ' ', n); buffer[n] = '\0';
Then there's no leaks, and 1k is so far beyond what you'll need it's not funny.

usr.sbin/devinfo/devinfo.c
```diff
@@ -78,13 +78,22 @@ print_resource(struct devinfo_res *res)
 {
 	struct devinfo_rman	*rman;
 	bool			hexmode;
+	rman_res_t		end;
```
bsdimp, line 125: Don't know how to add a comment on the commit message...

But this change looks good, I'd say 'unwind the ternary operator to better fit future libxo structure. No functional change intended.' instead of just no functional change intended.

usr.sbin/devinfo/devinfo.c
ktullavik: I have to define min myself right? So like this then?
I'd rather not change the type of the argument to avoid the cast to size_t, because there is already some casting involved with the indentation that I don't properly understand.

```
#define	min(a, b)	((a < b) ? a : b)

static char buffer[1024];
...
static void
print_indent(int n)
{
	if (n < 1)
		return;

	n = min((size_t)n, sizeof(buffer) - 1);
	memset(buffer, ' ', n);
	buffer[n] = '\0';
	xo_emit("{Pa:%s}", buffer);
}
```

usr.sbin/devinfo/devinfo.c
ktullavik: Fixed.

usr.sbin/devinfo/devinfo.c
rilysh: You should include `sys/param.h` (there MIN() as a macro is already defined).
Also, you don't have to explicitly cast, as buffer size is smaller than INT_MAX anyway. With `-Wextra` you'll get a harmless compiler warning, though to silent it you have to cast (either with size_t) or with int e.g. `(int)sizeof(buffer) - 1`

usr.sbin/devinfo/devinfo.c
```diff
@@ -200,8 +291,13 @@ print_device(struct devinfo_dev *dev, void *arg)
 		}
 	}
 
-	return(devinfo_foreach_device_child(dev, print_device,
+	ret = (devinfo_foreach_device_child(dev, print_device,
```
rilysh, line 294: ret seems to be assigned here only?
isn't it similar to
```
if (printit) {
	xo_close_container(devname);
}

return (devinfo_foreach_device_child(dev, print_device,
                           (void *)((char *)arg + 2)));
```

usr.sbin/devinfo/devinfo.c
```diff
@@ -213,34 +309,53 @@ print_rman_resource(struct devinfo_res *res, void *arg __unused)
...
 	rman = devinfo_handle_to_rman(res->dr_rman);
 	hexmode =  (rman->dm_size > 1000) || (rman->dm_size == 0);
 	end = res->dr_start + res->dr_size - 1;
 
-	printf("    ");
-
 	if (hexmode) {
 		if (res->dr_size > 1)
-			printf("0x%jx-0x%jx", res->dr_start, end);
+			asprintf(&res_str, "0x%lx-0x%lx", res->dr_start, end);
```
rilysh, line 322: btw res->dr_start is a `*max_t` (uintmax_t) and %j seems to be standard prefix to format a uintmax_t
but i'm not too sure about it

usr.sbin/devinfo/devinfo.c
ktullavik: Ah, param.h it is then. It's compiled with -Werror., so the cast is needed.

ktullavik: Fixed

usr.sbin/devinfo/devinfo.c
ktullavik: Fixed. Not shure how I messed that up. It was right to begin with. Good catch!

usr.sbin/devinfo/devinfo.c
```diff
@@ -31,46 +32,126 @@
  * Print information about system device configuration.
  */
 
-#include <sys/types.h>
+#include <sys/param.h>
```
bsdimp, line 35: this should sort after unistd.h. We're inconsistent about whether it should be before or after libxo/xo.h, but there's slightly more code in the tree that uses all sys entries after the normal includes w/o directories.

usr.sbin/devinfo/devinfo.c
```diff
@@ -31,46 +32,126 @@
...
 #include <err.h>
 #include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <stdbool.h>
```
bsdimp, line 42: this sorts before stdio.h

usr.sbin/devinfo/devinfo.c
```diff
@@ -31,46 +32,126 @@
...
+#include <libxo/xo.h>
 #include "devinfo.h"
 
-static int	rflag;
-static int	vflag;
+static bool	rflag;
+static bool	vflag;
+static int	open_tag_count;
+static char	*last_res;
+static char	buffer[1024];
```
bsdimp, line 50: should this be private to print_indent?

usr.sbin/devinfo/devinfo.c
```diff
@@ -31,46 +32,126 @@
...
-	printf(hexmode ? "0x%jx" : "%ju", res->dr_start);
-	if (res->dr_size > 1)
-		printf(hexmode ? "-0x%jx" : "-%ju",
-		    res->dr_start + res->dr_size - 1);
+	end = res->dr_start + res->dr_size - 1;
+
+	len = strlen(rman->dm_desc) + 1;
+	if ((safe_desc = malloc(len)) == NULL)
+		xo_err(1, "No memory!");
+	strlcpy(safe_desc, rman->dm_desc, len);
```
bsdimp, line 136: why not strdup? instead of malloc + strlcpy?

usr.sbin/devinfo/devinfo.c
```diff
@@ -116,52 +197,92 @@ print_device_rman_resources(struct devinfo_rman *rman, void *arg)
...
 		/* there are, print header */
-		for (i = 0; i < indent; i++)
-			printf(" ");
-		printf("%s:\n", rman->dm_desc);
+
+		len = strlen(rman->dm_desc) + 1;
+		if ((safe_desc = malloc(len)) == NULL) {
+			xo_err(1, "No memory!");
+		}
+		strlcpy(safe_desc, rman->dm_desc, len);
```
bsdimp, line 205: same as above.

usr.sbin/devinfo/devinfo.c
```diff
@@ -116,52 +197,92 @@ print_device_rman_resources(struct devinfo_rman *rman, void *arg)
...
-int
+static int
 print_device(struct devinfo_dev *dev, void *arg)
 {
 	struct indent_arg	ia;
-	int			i, indent;
+	int			indent, ret;
+	const char*		devname = dev->dd_name[0] ? dev->dd_name : "unknown";
+	bool			printit = vflag || (dev->dd_name[0] != 0
+					    && dev->dd_state >= DS_ATTACHED);
```
bsdimp, line 275: we have a preference for && at the end of the prior line rather than the start of this line. And the indent should align the 'dev' under the first 't' in printit.

usr.sbin/devinfo/devinfo.c
```diff
@@ -181,16 +307,54 @@ int
...
+			asprintf(&res_str, "%ju", res->dr_start);
+	}
+
+	xo_emit("{P:    }");
+
+	if (last_res == NULL) {
+		// First resource
+		xo_open_list(res_str);
+	}
+	else if (strcmp(res_str, last_res) != 0) {
```
bsdimp, line 339: '} else if' is how this should be formatted, on one line.

usr.sbin/devinfo/devinfo.c
```diff
@@ -181,16 +307,54 @@ int
...
 	if ((dev != NULL) && (dev->dd_name[0] != 0)) {
-		printf(" (%s)", dev->dd_name);
-	} else {
-		printf(" ----");
+		asprintf(&entry, "{el:%s}{D:%s} {D:(%s)}\n",
+		    res_str, res_str, dev->dd_name);
+		xo_emit(entry, dev->dd_name);
 	}
-	printf("\n");
+	else {
```
bsdimp, line 352: same

usr.sbin/devinfo/devinfo.c
```diff
@@ -200,65 +364,115 @@ print_rman_resource(struct devinfo_res *res, void *arg __unused)
...
-	printf("%s:\n", rman->dm_desc);
+	char*	safe_desc;
+	size_t	len;
+
+	len = strlen(rman->dm_desc) + 1;
+	if ((safe_desc = malloc(len)) == NULL) {
+		xo_err(1, "No memory!");
+	}
+	strlcpy(safe_desc, rman->dm_desc, len);
+	xml_safe_string(safe_desc);
```
bsdimp, line 375: This is the third instance of this, that suggests a common function and a little refactoring would be in order.

bsdimp:
OK. That's this round. It looks to be mostly style. I'll test it on the next round to see if I can find some issues I'm not seeing in glancing over this.

bsdimp:
Also, the 3rd-5th commit messages need to have more descriptive titles. They are a little generic. Let me know if you want more specific suggestions.

usr.sbin/devinfo/devinfo.c
ktullavik: Style(9) seemed pretty clear to me:
"Kernel include files (sys/*.h) come first.  If either <sys/types.h> or <sys/param.h> is needed, include it before other include files." What am I missing? I became a dad a few day ago. Please excuse the pace going forward.

usr.sbin/devinfo/devinfo.c
ktullavik: I'm not used to tab-stops indenting. The results seem inconsistent between editors. What editor/config combo can I use to make shure it's right?

usr.sbin/devinfo/devinfo.c
ktullavik: Fixed

usr.sbin/devinfo/devinfo.c
ktullavik: Fixed

usr.sbin/devinfo/devinfo.c
ktullavik: Fixed

usr.sbin/devinfo/devinfo.c
```diff
@@ -31,46 +32,127 @@
...
-#include <sys/types.h>
+#include <sys/param.h>
 #include <err.h>
 #include <errno.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <libxo/xo.h>
```
bsdimp, line 43: style(9) also suggests a blank line after unistd.h
And after sys/param.h. (but if that's the only issue, I'll just fix it before pushing)

usr.sbin/devinfo/devinfo.c
bsdimp: I have an emacs config file that I've used for a long time. The important part is && at the end. The extra indent is less important, to be honest. If you are using vi/vim/nvi, I'm afraid I can't help you, but maybe tools/tools/editing/freebsd.vim can help? I don't know first hand how well it works.

bsdimp:
Other than a couple of very minor style nits, the only thing I see is the strdup/refactoring suggestion I made.

usr.sbin/devinfo/devinfo.c
ktullavik: Fixed by putting more logic into xml_safe_string(). Shaved of 15 lines in total. Thanks.

usr.sbin/devinfo/devinfo.c
ktullavik: I'm using vscode with no relevant config. But default vi seems more in agreement with github diffs so I'll go with that for now.

ktullavik:
I think it's ready for the next round. Let me know if there's anything else @bsdimp

bsdimp:
Oh no. Gotta review this. Missed it when it came in

bsdimp:
Landing this, after fixing the conflict.

PR #1486, opened 2024-10-18
tools/build/cross-build: Don't include sys/uio.h from linux limits.h

jrtc27:
This creates a circular dependency for OpenZFS's libspl in sys/uio.h,
and it shouldn't be needed since the system limits.h already defines
IOV_MAX, so delete it, and unconditionally assert that to be the case.
Otherwise the re-include of libspl's sys/uio.h tries to use PAGESIZE
before it has been defined by OpenZFS's own sys/param.h.

Fixes:		7a7741af18d6 ("zfs: merge openzfs/zfs@b10992582")
MFC after:	1 week


jrtc27:
For CI only

bsdimp:
This looks good to me... Sorry for the delay :)

PR #1487, opened 2024-10-22
ukbd: Enabling ukbd_apple_swap for all Apple ISO keyboards

frarohi:
Key code swapping between [<>] and [^°] key is enabled for all Apple ISO type keyboards.
Before, swapping was enabled when the Eject key was detected in HID usage.
This did not correlate well with the swapped keys presence.

usbdevs file is extended by several Apple keyboard models to support the ISO model identification.


bsdimp:
Why close?

frarohi:
Sorry, I accidently cleaned up the wrong branch.
I will try to re-issue the pull request later

frarohi:
Re-issued as #1506

PR #1488, opened 2024-10-24
Migrate from old repo

i-am-gizm0:
Copied in module from old repo. Applied patches to get it building on main


i-am-gizm0:
Whoops. I blame GitHub for their default being that PRs should open on the upstream of a fork, not the fork itself. Who thought that was a good idea?

PR #1489, opened 2024-10-24
look(1): Capsicumise

kfv:


lwhsu:
@oshogbo can you review this?

usr.bin/look/look.c
```diff
@@ -132,21 +138,35 @@ main(int argc, char *argv[])
...
+	for (size_t idx = 0; idx < sizeof fds / sizeof fds[0]; file = argv[idx++]) {
+		if ((fds[idx] = open(file, O_RDONLY, 0)) < 0)
+			err(2, "%s", file);
+#ifndef WITHOUT_CAPSICUM
+		if (cap_rights_limit(fds[idx], &rights) < 0)
+			err(2, "unable to limit rights for %s", file);
+	}
+	if (cap_enter() < 0)
+		err(EXIT_FAILURE, "failed to enter capability mode");
+	for (size_t idx = 0; idx < sizeof fds / sizeof fds[0]; file = argv[idx++]) {
```
oshogbo, line 154: We use sizeof().
You can also use nitems.

oshogbo: This if is under WITHOUT_CAPSICUM?
Have you tried to build this with this flag defined?

oshogbo (changes requested):

usr.bin/look/look.c
```diff
@@ -132,21 +138,35 @@ main(int argc, char *argv[])
 
 	match = 1;
 
-	do {
-		if ((fd = open(file, O_RDONLY, 0)) < 0 || fstat(fd, &sb))
+#ifndef WITHOUT_CAPSICUM
+	cap_rights_init(&rights, CAP_MMAP_R, CAP_READ, CAP_FSTAT);
+#endif
+	int fds[argc > 1 ? argc - 1 : argc];
```
oshogbo, line 144: We don't use dynamic stack variable initialization from newer C. You should use malloc or calloc.

oshogbo (changes requested):

usr.bin/look/look.c
kfv: Yes, and it compiled and worked as expected. But I'll double check shortly.

kfv: And you're absolutely right regarding `sizeof()`, we have it clearly stated in style(9), my apologies.

usr.bin/look/look.c
```diff
@@ -132,21 +138,35 @@ main(int argc, char *argv[])
 
 	match = 1;
 
-	do {
-		if ((fd = open(file, O_RDONLY, 0)) < 0 || fstat(fd, &sb))
+#ifndef WITHOUT_CAPSICUM
+	cap_rights_init(&rights, CAP_MMAP_R, CAP_READ, CAP_FSTAT);
```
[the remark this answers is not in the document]
kfv: I used WITHOUT_CAPSICUM assuming it is required to handle cases where the kernel is built without Capsicum support, rather than strictly for tools in contrib. Could you clarify if my understanding is off? I'd appreciate any insights!

usr.bin/look/look.c
kfv: I didn't find anything in `style(9)` explicitly disallowing VLAs, and technically they’re not from 'newer' C—VLAs have been part of the standard since C99 and are mandatory again in C23 (after being optional since C11.) I believe using a VLA here would simplify our approach significantly and avoid unnecessary heap allocation, improving memory efficiency. I’d like to request reconsideration for this specific case, but if there’s a strong rationale against VLAs in general, I'd appreciate any guidance on it.

usr.bin/look/look.c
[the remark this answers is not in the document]
oshogbo: On FreeBSD if you build a kernel without the CAPSICUM support the `cap_*` functions still exists however they will fail and return ENOSYS.
To avoid checking errno you can use capsicum helpers.
The WITHOUT_CAPSICUM is for platforms that aren't aware of capsicum's existence, like Linux. So we have to add them to contrib source code which isn't managed by FreeBSD but not in the code that is dedicated for FreeBSD.

usr.bin/look/look.c
oshogbo: Yes, you are right. However, I (maybe I should say We as project) try to avoid them. Stack is limited and I don't want to even analyze if we won't exceed it on all supported platforms or if the command line lenght limit will be enough.

usr.bin/look/look.c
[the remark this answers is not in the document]
kfv: Thanks, I didn't know that. I'll take care of it.

usr.bin/look/look.c
kfv: All right, it makes sense.

oshogbo:
@kfv Can you also fix the style issues reported by GitHub Actions?

kfv:
@oshogbo: The style checks are already passing, but I assume you're referring to the warnings for lines exceeding 80 characters. I’ll go ahead and address those as well, sure.

kfv:
@oshogbo: I applied soft wrapping but kept `mmap()` on line 162 unwrapped for clarity. I also removed `WITHOUT_CAPSICUM` by handling ENOSYS as suggested. I’ll just need to go over the code again to address the dynamic stack allocation we discussed.

oshogbo:
Sorry, why haven't we used capsicum_helpers here?

bsdimp:
@markjdb any final comments?

markjdb:
> @markjdb any final comments?

@oshogbo 's comments still apply here: these should just use capsicum helpers (i.e., `caph_*` routines which handle ENOSYS), and the dynamically allocated stack array is kind of iffy and should be converted to an explicit allocation.

bsdimp:
@kfv What are your plans here? This seems to be stuck waiting for addressing the helper feedback comments.

kfv:

kfv:
I think this is ready for final review. Let me know if I’ve missed anything or if there’s anything else you'd like me to adjust.

usr.bin/look/look.c
```diff
@@ -132,22 +137,38 @@ main(int argc, char *argv[])
 
 	match = 1;
 
-	do {
-		if ((fd = open(file, O_RDONLY, 0)) < 0 || fstat(fd, &sb))
+	cap_rights_init(&rights, CAP_MMAP_R, CAP_READ, CAP_FSTAT);
+	nfiles = argc > 1 ? argc - 1 : argc;
+	if ((fds = malloc(nfiles * sizeof(int))) == NULL)
+		err(2, NULL);
+	for (size_t idx = 0; idx < nfiles; file = argv[idx++]) {
```
oshogbo: Hym, I'm wondering if we care about change of behavior in this utility.

So currently if I do:
```
look string a_file b_file c_file d_file
```
and `c_file` doesn't exists I still get result from `a_file` and `b_file`.
However, after this change I will get immediately an error that the `c_file` doesn't exists without results from `a_file` and `b_file`.
I'm not a big user of look so I would say I don't care, but maybe somebody uses this behavior.
What do you think @markjdb and @bsdimp ?

oshogbo:
What do you think @markjdb and @bsdimp ?

usr.bin/look/look.c
markjdb: Is it challenging to support the old behaviour? It's hard to say whether it's important or not, so without more info I'd avoid changing how this works.

oshogbo: Imho it's not. We need to store errors as negative values and report them later on.

kfv: Agreed. I'll make it support the old behaviour in two days when I get back to work.

kfv: The capsicumised code now preserves the original behaviour.

Although the original implementation has a known limitation where it stops processing at the first file error (which could be considered suboptimal from a UX perspective, I might say) I've maintained this behaviour to ensure backward compatibility at the minute.

This approach allows us to ship the new enhancements while deferring the behavioural improvements to a future release. When that time comes, I'll implement proper error handling with detailed documentation of any breaking changes.

usr.bin/look/look.c
```diff
@@ -132,22 +137,38 @@ main(int argc, char *argv[])
...
+		if (caph_rights_limit(fds[idx], &rights) != 0)
+			err(2, "unable to limit rights for %s", file);
+	}
+
+	caph_cache_catpages();
+	if (caph_enter() != 0)
+		err(EXIT_FAILURE, "failed to enter capability mode");
+
+	for (size_t idx = 0; idx < nfiles; file = argv[idx++]) {
+		if (fstat(fds[idx], &sb))
```
oshogbo: I think this is still wrong. I have not run this code, so I might be wrong here.

The open(2) returns `-1` for failed open. We ignore actual error codem which is in `errno`. Then we pass -1 to `fstat(2)`. In result instead of meaningful error like "File doesn't exists", "No access" ect. we provide user with a same error "Invalid file descriptor" when the `fstat(2)` fails.

usr.bin/look/look.c
```diff
@@ -132,22 +137,38 @@ main(int argc, char *argv[])
...
+			continue;
+		if (caph_rights_limit(fds[idx], &rights) != 0)
+			err(2, "unable to limit rights for %s", file);
+	}
+
+	caph_cache_catpages();
+	if (caph_enter() != 0)
+		err(EXIT_FAILURE, "failed to enter capability mode");
+
+	for (size_t idx = 0; idx < nfiles; file = argv[idx++]) {
```
oshogbo: Actually I just noticed that this magic is also wrong.
You forgot to reset file to `argv[0]`. So if I read this correctly your first `file` is actually the last from argv. Later it works.

This is why I actually don't like such magical initalization, I think such code like:
```
for (size_t idx = 0; idx < nfiles; idx++) { 
     file = argv[idx];
```

I haven't read into more details but it seems that we also move argv by one:
```
	if (argc >= 2)
		file = *argv++;
```

So I'm not sure if we can reuse it in the previous snipped. This all seems quite magical.

kfv: You are absolutely right, Mariusz — that was my oversight, and thank you for catching it. My intention was to preserve the style of this decades-old codebase, which is why I avoided introducing structural modifications, and in doing so, I miscalculated the handling of `argv`. My apologies.

I have just pushed a new commit that addresses this by removing the implicit `argv++` side effect and introducing a more explicit indexed loop over the files. And to keep track of `errno`, I added a small structure with `fd` and `err` members. I considered adding a `name` member as well, in which case the iteration domain would shift entirely to that structure, eliminating the need for `argv` in the second loop. It could be refactored even further if we decide it’s worthwhile.

If the preference is for a more substantial redesign, I am more than happy to work on that as well. Otherwise, please let me know if there are any remaining gaps for our strict _"capsicumisation"_ goal here, or other areas that still need fixing, and I’ll take care of them.

usr.bin/look/look.c
kfv: Right, that was my mistake. It’s now fixed, with the details explained here:  https://github.com/freebsd/freebsd-src/pull/1489#discussion_r2319811140

usr.bin/look/look.c
```diff
@@ -128,26 +138,48 @@ main(int argc, char *argv[])
...
 	if (argc >= 2)
-		file = *argv++;
+		file = *argv;
 
 	match = 1;
 
-	do {
-		if ((fd = open(file, O_RDONLY, 0)) < 0 || fstat(fd, &sb))
+	cap_rights_init(&rights, CAP_MMAP_R, CAP_READ, CAP_FSTAT);
+	nfiles = argc > 1 ? argc - 1 : argc;
```
oshogbo, line 146: I don't think I understand what is happening here.

usr.bin/look/look.c
```diff
@@ -128,26 +138,48 @@ main(int argc, char *argv[])
...
+	if ((files = malloc(nfiles * sizeof(struct files))) == NULL)
+		err(2, NULL);
+	for (size_t idx = 0; idx < nfiles; idx++) {
```
oshogbo, line 147: The code (at least for me) seem a little bit puzzling.
First I don't like the play with file. Its seems a little bit hackish.
Maybe we can have something like:

```
static char *_path_words[] = { _PATH_WORDS };

...

file_list = _path_words;
nfiles = 1;
if (argc >= 2) {
    file_list = argv;
    nfiles = argc - 1;
}
```
Then using file_list to iterate you don't have to guess which list you are iterating.


To be honest I don't like the fact that argv/argc gets desynchronized.
So I would propose also this change:
```
key = prepkey(*argv++, termchar);
argc -= 1;
```

usr.bin/look/look.c
```diff
@@ -85,12 +87,20 @@ static struct option longopts[] = {
 	{ NULL,		0,		NULL, 0 },
 };
 
+struct files {
+	int fd;
+	int err;
+};
```
oshogbo, line 94: Thats a good idea.

usr.bin/look/look.c
```diff
@@ -128,26 +138,48 @@ main(int argc, char *argv[])
...
+
+	caph_cache_catpages();
+	if (caph_enter() != 0)
+		err(EXIT_FAILURE, "failed to enter capability mode");
+
+	for (size_t idx = 0; idx < nfiles; idx++) {
+		if (idx || nfiles > 1)
+			file = argv[idx];
+		if (files[idx].err || fstat(files[idx].fd, &sb))
 			err(2, "%s", file);
```
oshogbo: What `errno` will be used here?
I guess you have to restore errno from `files[idx].err` to print correct msg.

```
if (files[idx].err != 0) {
   errno = files[idx].err;
    err(2, "%s", file);
}

if (fstat(files[idx].fd, &sb))
    err(2, "%s", file);
```

usr.bin/look/look.c
```diff
@@ -128,26 +138,48 @@ main(int argc, char *argv[])
...
+		if (caph_rights_limit(files[idx].fd, &rights) != 0)
+			err(2, "unable to limit rights for %s", file);
+	}
+
+	caph_cache_catpages();
+	if (caph_enter() != 0)
+		err(EXIT_FAILURE, "failed to enter capability mode");
+
+	for (size_t idx = 0; idx < nfiles; idx++) {
+		if (idx || nfiles > 1)
```
oshogbo: Maybe we should keep the name of file in the `struct` to not reinterpret argv and files?
If you don't like this approach, with the `file_list` you also don't need to guess if you are using argv or `_path_words`.

bsdimp:
Is there a new version of this that addresses the review feedback?

kfv:
Don't know why I forgot this PR, thought it's landed. Working on it now.

usr.bin/look/look.c
kfv: Agreed. I'm on it.

usr.bin/look/look.c
kfv: You are right to question this. That would either use whatever global `errno` is set without ever reading `files[idx].err` in case of the short-circuit, or use the `errno` set by `fstat()` on the right-hand side — where only the latter is correct. For the former case, by the way, I prefer to use `errc()` instead of restoring `errno` first and then calling `err()`.

usr.bin/look/look.c
kfv: Cheers ;-)

usr.bin/look/look.c
kfv: You're right. Can you take a look and see if the recent updates make it clean enough, or if you still think further improvements are needed? I've kept the ternaries as-is; I think with the rest of the modifications it should be clean enough, but I'm open to any ideas.

kfv:
Cc: @oshogbo

oshogbo:
LGTM. I will build it and test.

kfv:
May I proceed with this patch?

Cc: @freebsdfrau, @clausecker

oshogbo:
Yes, please go ahead.

clausecker:
Looking good here!
Approved for commit.

usr.bin/look/look.c
```diff
@@ -127,27 +136,48 @@ main(int argc, char *argv[])
...
+	cap_rights_init(&rights, CAP_MMAP_R, CAP_READ, CAP_FSTAT);
+	nfiles = !argc ? 1 : argc;
+	if ((files = malloc(nfiles * sizeof(struct files))) == NULL)
+		err(2, NULL);
+	for (size_t idx = 0; idx < nfiles; idx++) {
+		files[idx].path = !argc ? _path_words : argv[idx];
+		if ((files[idx].fd = open(files[idx].path, O_RDONLY, 0)) < 0) {
+			files[idx].err = errno;
+			continue;
+		}
```
kfv, line 152: While I was doing a final check before landing the patch, I noticed I'd forgotten `files[idx].err = 0` on the success path, so `files[idx].err` would retain whatever `malloc()` left behind. The later if `(files[idx].err)` check would then trigger, causing `errc()` to fail even though the file opened fine.

kfv:
It's ready now. I'll land it shortly.

kfv:
Closed by commit f66c8680e804c282af66c93b13766801c6f85e7f

PR #1492, opened 2024-10-26
perror(1): Capsicumise

kfv:


markjdb (approved):

markjdb:
This looks ok, but in general we don't bother with sysexits.h for new code. I can't point to a particular policy which forbids it, but I would just drop such commits. Or was that specifically recommended?

emaste:
sysexits.h is in a bit of an unfortunate state, with differing opinions on whether it should be used. There's a review open to discourage its use: https://reviews.freebsd.org/D27176

I find it hard to disagree with @dag-erling's stance expressed in that review, as a baseline:
1. sysexits are permitted but not required
2. if you do use them, use them consistently
3. don't add them to existing code without good reason
4. conversely, don't remove them from existing code without good reason
5. mixed usage (within a single program) is a good reason (to either remove or apply them consistently, as you wish)

I'd apply point 3 and not add them here. And in particular when a program exits with only status 0 or 1 I don't think changing to `EXIT_SUCCESS` and `EXIT_FAILURE` adds value.

emaste:
Hrm, it seems that as of 5ccaab1797e5bdc404f2299ebaec1265547a4744 we actually claim that sysexits.h is deprecated.

dag-erling:
> Hrm, it seems that as of [5ccaab1](https://github.com/freebsd/freebsd-src/commit/5ccaab1797e5bdc404f2299ebaec1265547a4744) we actually claim that sysexits.h is deprecated.

Considering that is based on a1432b4c9920de1bb864b42f3995049431a0c123 which I authored, I think I'm entitled to change my mind

bsdimp:
Also des has a review somewhere that codogies what ed said originally.

oshogbo (approved):

oshogbo:
Done: 4dd6597a56f8c7ff03d4f451b612a9e45442c549

PR #1497, opened 2024-10-27
Update rc.initdiskless, fix error handling of remount_optional

kevemueller:
chkerr() ignores the exit code of a preceding mount command in case a file ```remount_optional``` exists. The check is performed and a subshell is launched to log the informational message and return. The return is executed in the context of the subshell, not the context of the chkerr() function, hence is a NOP. The remount_optional check is hence ineffective.

Change the code to if/then/fi, so the return is evaluated in the context of the  chkerr function, to make the check effective.

### DIFF
--- a/libexec/rc/rc.initdiskless
+++ b/libexec/rc/rc.initdiskless
@@ -174,7 +174,10 @@ log() {
 chkerr() {
     lastitem () ( n=$(($# - 1)) ; shift $n ; echo $1 )
     mountpoint="$(lastitem $2)"
-    [ -r $mountpoint/remount_optional ] && ( echo "$2 failed: ignoring due to remount_optional" ; return )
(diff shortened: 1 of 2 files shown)


bsdimp (approved):
Subtle but wrong before

emaste (approved):

bsdimp:
I staged this, but got the error
```
ERROR: Real email adress is needed
#2:
Author: Keve Müller <kevemueller@users.noreply.github.com>
```
Can you offer a better email?

Also, while you're at it, can you wrap the body of the commit message at 72 columns (+/-) please if you do this by updating the commit?

bsdimp:
We need the email for this and #1452

kevemueller:
pushed with private email address.

bsdimp:
merged, but couldn't push to show as merged

PR #1498, opened 2024-10-27
lsvfs(1): Capsicumise

kfv:


kfv:
This must be ready, I believe.

Cc: @oshogbo

oshogbo (changes requested):
We don't need to change Makefile to add casper support?

usr.bin/lsvfs/lsvfs.c
```diff
@@ -40,10 +45,29 @@ main(int argc, char **argv)
...
 
 	argc--, argv++;
 
+	if ((capcas = cap_init()) == NULL)
+		err(EXIT_FAILURE, "unable to contact Casper");
+	if ((capsysctl = cap_service_open(capcas, "system.sysctl")) == NULL)
+		err(EXIT_FAILURE, "unable to open system.sysctl servic");
+	cap_close(capcas);
+
+	if ((limit = cap_sysctl_limit_init(capsysctl)) == NULL)
```
oshogbo, line 60: We cannot check here for a NULL. As the non-casper version of the cap_sysctl_limit_init will return name as well. I guess we need to update man page for this (the return value section). The idea is that you ignore errors from `cap_sysctl_limit_init` ect, and verify only `cap_sysctl_limit`. See examples section of man page.

oshogbo (changes requested):

usr.bin/lsvfs/lsvfs.c
kfv: Alright, I'll take care of it now, and will also take care of the man page later today.

kfv:
> We don't need to change Makefile to add casper support?

You're right; I had changed it but forgot to commit. My apologies.

usr.bin/lsvfs/lsvfs.c
oshogbo: I guess we/I need to update the `cap_sysctl_limit_name` of the real function to ignore NULL limit.

kfv: Good idea, makes sense. Alright.

kfv: The latest commit should meet your call now. Let me know if any other change's required as well.

oshogbo (approved):

usr.bin/lsvfs/lsvfs.c
```diff
@@ -40,10 +45,27 @@ main(int argc, char **argv)
...
+	int cnt, rv = 0;
+	cap_channel_t *capcas, *capsysctl;
+	cap_sysctl_limit_t *limit;
 
 	argc--, argv++;
 
+	if ((capcas = cap_init()) == NULL)
+		err(EXIT_FAILURE, "unable to contact Casper");
+	if ((capsysctl = cap_service_open(capcas, "system.sysctl")) == NULL)
+		err(EXIT_FAILURE, "unable to open system.sysctl servic");
```
asomers, line 57:
```suggestion
		err(EXIT_FAILURE, "unable to open system.sysctl service");
```

asomers (changes requested):

usr.bin/lsvfs/lsvfs.c
kfv: Thanks for the catch! Resolved.

asomers (approved):

bsdimp:
@markjdb any final comments?

bsdimp:
@kfv Any comments? Otherwise, I plan on closing soon.

kfv:

kfv:
@bsdimp, @markjdb: Hey! I believe it's now ready to ship.

usr.bin/lsvfs/lsvfs.c
```diff
@@ -82,10 +87,9 @@ static const char *
 fmt_flags(int flags)
 {
 	static char buf[sizeof(struct flaglist) * sizeof(fl)];
-	int i;
 
 	buf[0] = '\0';
-	for (i = 0; i < (int)nitems(fl); i++) {
+	for (int i = 0; i < (int)nitems(fl); i++) {
```
oshogbo: Instead of casting it to int maybe we can use `size_t` here?
What do you think?

usr.bin/lsvfs/lsvfs.c
```diff
@@ -40,10 +42,22 @@ main(int argc, char **argv)
...
 
 	argc--, argv++;
 
+	if (sysctlbyname("vfs.conflist", NULL, &buflen, NULL, 0) < 0)
+		err(EXIT_FAILURE, "sysctl(vfs.conflist)");
+	if ((xvfsp = malloc(buflen)) == NULL)
+		errx(EXIT_FAILURE, "malloc failed");
+	if (sysctlbyname("vfs.conflist", xvfsp, &buflen, NULL, 0) < 0)
+		err(EXIT_FAILURE, "sysctl(vfs.conflist)");
+	cnt = buflen / sizeof(struct xvfsconf);
```
oshogbo, line 55: As we are here to do we want `cnt` keep as `int`?
We are dividing size_t by size_t, so potentially 64bits unsigned numbers and we store it in 32bit signed.

usr.bin/lsvfs/lsvfs.c
kfv: Good catch — you're absolutely right. Thanks for pointing that out.

usr.bin/lsvfs/lsvfs.c
kfv: Sharp eye — you're spot on, mate. Appreciate you flagging that.

usr.bin/lsvfs/lsvfs.c
```diff
@@ -58,16 +58,16 @@ main(int argc, char **argv)
 			}
 		}
 	} else {
+		for (size_t i = 0; i < cnt; i++) {
```
oshogbo: I think the indentation here is broken.

kfv: I just checked – all looks fine. I only changed the initialisation clause; nothing else was modified.

oshogbo: image

Please verify commit: lsvfs(1): Make slight refactor

kfv: All set – you've got my nod, mate.

oshogbo: Sorry mate, but I think we are still not on the same page :)
In this PR you have two commits:
- e9e7ef72f28da5c3313b0b873ac1c9aba21d05fe: lsvfs(1): Make slight refactor
- 47757166ca6c4d48cf7e35eceb3dbded8c33d943: lsvfs(1): Capsicumise

In the first commit where you are doing a refactoring you are changing also the indentation. Here is the link:
https://github.com/freebsd/freebsd-src/commit/e9e7ef72f28da5c3313b0b873ac1c9aba21d05fe.patch

Here is the patch fragment:
```
+		for (size_t i = 0; i < cnt; i++) {
 		if (sysctlbyname("vfs.conflist", NULL, &buflen, NULL, 0) < 0)
-			err(1, "sysctl(vfs.conflist)");
-		xvfsp = malloc(buflen);
-		if (xvfsp == NULL)
-			errx(1, "malloc failed");
+			err(EXIT_FAILURE, "sysctl(vfs.conflist)");
+		if ((xvfsp = malloc(buflen)) == NULL)
+			errx(EXIT_FAILURE, "malloc failed");
 		if (sysctlbyname("vfs.conflist", xvfsp, &buflen, NULL, 0) < 0)
-			err(1, "sysctl(vfs.conflist)");
+			err(EXIT_FAILURE, "sysctl(vfs.conflist)");
 		cnt = buflen / sizeof(struct xvfsconf);
 
-		for (i = 0; i < cnt; i++) {
+		for (size_t i = 0; i < cnt; i++) {
 			printf(FMT, xvfsp[i].vfc_name, xvfsp[i].vfc_typenum,
 			    xvfsp[i].vfc_refcount,
 			    fmt_flags(xvfsp[i].vfc_flags));
``` 

This ends up with the wrong indentation of code:
image

Only after applying the first patch.
Then when you apply the second patch you are fixing the indentation.

We don't need to change the indention in the first commit.

kfv: Oh, my bad! We absolutely should not, indeed. Sorry, I checked but didn't notice (too sleepy ATM.) You're right. I'll fix it shortly.

kfv: Apologies for the confusion earlier. While working on the latest domain iteration, I inadvertently copied and pasted the code above the conditionals, which caused the issue. I wasn't touching any indentation, it's likely that I accidentally used `yy` and `P` without realising. The issue has now been resolved.

usr.bin/lsvfs/lsvfs.c
```diff
@@ -44,6 +46,18 @@ main(int argc, char **argv)
 
 	argc--, argv++;
 
+	if (sysctlbyname("vfs.conflist", NULL, &buflen, NULL, 0) < 0)
```
oshogbo, line 49: I was about to commit this change, when I have noticed one more thing. Sorry about that.

Do we want to always fetch the full list.
When we do:
```
lsvfs ufs
```

We don't need to fetch the `vfs.conflist` to check for UFS fs. Now we are adding a small overhead to the utility that we always fetch the list. @kfv what do you think?

oshogbo:
The first commit landed: a3eab01304884e13342bdaca4cf5204fa97ba456

usr.bin/lsvfs/lsvfs.c
kfv: I’ll look into this in more detail and get back to you shortly.

kfv: You are right. Resolved.

kfv:
@oshogbo: I’m pretty sure we’re finally ready to ship the Capsicumise commit as well. I’ve completely reworked it since it had some issues. Thanks a lot for catching that and for bearing with my slip-ups. It’s been a great journey!

usr.bin/lsvfs/lsvfs.c
```diff
@@ -65,14 +70,17 @@ main(int argc, char **argv)
...
+	}
 
-		for (size_t i = 0; i < cnt; i++) {
-			printf(FMT, xvfsp[i].vfc_name, xvfsp[i].vfc_typenum,
-			    xvfsp[i].vfc_refcount,
-			    fmt_flags(xvfsp[i].vfc_flags));
-		}
-		free(xvfsp);
+	caph_cache_catpages();
+	if (caph_enter() != 0)
```
oshogbo, line 58: Sorry I don't like this new change.
After this change it seems like totally useless to enter Capability mode as we have done all the heavy lifting (parsing the list) in unsandboxed part. What we are sandboxing now? Just printing the names?

This is bad on my side as well because, I just now checked how the `getvfsbyname` works. It is overlook on my side. Sorry about that.

I think we should rewrite this. Maybe we can't use `getvfsbyname`. Maybe add a function to libc like: `getvfslist`. Reuse it in the `getvfsbyname` and use it here. Maybe we can split it like that:

```
list = getvfslist();
caph_enter();

// check for argv or print all list.
for (size_t i = 0; i < cnt; i++) {
      found = false;
      for (; argc > 0; argc--, argv++) {
         if (strcmp()) {
            found = true;
         }
      }
      if (found || argc == 0) {
         printf(...);
      }
}
```

Or just drop usage of `getvfsbyname` here without adding `getvfslist`.

oshogbo:
@kfv Thoughts?

usr.bin/lsvfs/lsvfs.c
kfv: You’re right — and thanks a lot for pointing out. After moving the `getvfsbyname` calls outside capability mode, the code does indeed end up sandboxing only the printing phase, which weakens the whole point of Capsicum, agreed.

My intention with the change was to make the flow a bit clearer and avoid failures from `getvfsbyname` after entering capability mode. In that sense, it was a more workable version at the time, I still believe, even if not the right long-term solution.

Do I also agree with your suggestion: a `getvfslist`-style interface would be the right direction, so that we can fetch once and then enter capability mode before filtering/printing. I’ll start working on it shortly along those lines.

kfv:
@oshogbo: Reworked this based on your feedback (and also shared some thoughts on the thread). Sorry for the delay — I was away on vacation.

This version now fetches the full `vfs.conflist` list up front, then enters capability mode before any filtering/printing. That way sandboxing is meaningful, while keeping the overall approach close to what we had with `getvfsbyname`, but done manually here without further `sysctl` calls under the hood.

Could you take a look and let me know what you think?

oshogbo:
Yes, it looks much better. I will build it and test and let you know if it runs as expected.

oshogbo:
```
usr.bin/lsvfs/lsvfs.c:67:18: error: comparison of integers of different signs: 'size_t' (aka 'unsigned long') and 'int' [-Werror,-Wsign-compare]
   67 |                         for (j = 0; j < argc; j++) {
      |                                     ~ ^ ~~~~
usr.bin/lsvfs/lsvfs.c:71:10: error: comparison of integers of different signs: 'size_t' (aka 'unsigned long') and 'int' [-Werror,-Wsign-compare]
   71 |                         if (j == argc)
      |                             ~ ^  ~~~~
2 errors generated.
*** Error code 1
```

kfv:
Right, `j` should have been declared as `int` since it is compared against `argc`. My apologies for the oversight.

oshogbo:
Commited: a0c709ab5af4e87ce4579404c4ffbd4295ad12c5

PR #1503, opened 2024-10-30
sysent: add a NOLIB syscall flag

brooksdavis:
I added this in CheriBSD as part of porting to the new syscall generation code.

CC: @agge3


sys/tools/syscalls/scripts/syscalls_map.lua
```diff
@@ -38,7 +38,7 @@ function syscalls_map.generate(tbl, config, fh)
 
 	for _, v in pairs(s) do
 		--print("num " .. v.num .. " name " .. v.name)
-		if v:native() and not v.type.NODEF and v.name ~= "yield" then
+		if v:native() and not v.type.NODEF and not v.type.NOLIB then
 			if v.name ~= "exit" and v.name ~= "vfork" then
```
brooksdavis, line 42: I've got a plan for these as well (dropping most `_foo` format symbols by adding an INTERPOSED flag), but that will generate more churn so I want to do it separately.

emaste:
A commit message typo: yeild

Overall looks good to me

As an aside we lose some useful information in the pull request view, at least at first look - the commit message for the `sysent: add a NOLIB syscall flag` change and the change itself explains what the flag does, but the pull request / conversation view doesn't capture this. Not so much for this set of changes, but something we ought to think about for pull requests in general - perhaps we should add a template.

brooksdavis:
> As an aside we lose some useful information in the pull request view, at least at first look - the commit message for the `sysent: add a NOLIB syscall flag` change and the change itself explains what the flag does, but the pull request / conversation view doesn't capture this. Not so much for this set of changes, but something we ought to think about for pull requests in general - perhaps we should add a template.

This is a result of me being lazy. It would be nice if there were a way to say "populate the PR description from *this* commit" instead of making you write your own description any time there is more than one commit. (Alternatively this could all be one commit since the separation of concerns isn't particularly interesting.)

emaste (approved):

brooksdavis:
Merged as of 59a8b439aca5e6ef3c603f3f9814cdbd234d4947

PR #1504, opened 2024-10-31
resolver.5: Fix example Google DNS server address

pulsemod:
The previouly written example of 4.4.4.4 is not a Google DNS server and shouldn't be used, corrected to their secondary DNS address.


emaste (approved):

emaste:
Merged as e7f0f4f9206e8b0f411847b293c7b79eb84f0e51

PR #1509, opened 2024-11-02
fusefs: ignore FUSE_NO_OPEN(DIR)_SUPPORT flags

CismonX:
The `FUSE_NO_OPEN_SUPPORT` and `FUSE_NO_OPENDIR_SUPPORT` flags
are only meant to indicate kernel features, and should be ignored
if they appear in the `FUSE_INIT` reply flags.

See: <https://github.com/torvalds/linux/blob/v6.11/fs/fuse/inode.c#L1232>

### DIFF
--- a/sys/fs/fuse/fuse_vnops.c
+++ b/sys/fs/fuse/fuse_vnops.c
@@ -1945,10 +1945,9 @@ fuse_vnop_readdir(struct vop_readdir_args *ap)
 	tresid = uio->uio_resid;
 	err = fuse_filehandle_get_dir(vp, &fufh, cred, pid);
 	if (err == EBADF && mp->mnt_flag & MNT_EXPORTED) {
-		KASSERT(fuse_get_mpdata(mp)->dataflags
-				& FSESS_NO_OPENDIR_SUPPORT,
-			("FUSE file systems that don't set "
-			 "FUSE_NO_OPENDIR_SUPPORT should not be exported"));
+		KASSERT(!fsess_is_impl(mp, FUSE_OPENDIR),
+			("FUSE file systems that implement "
+			 "FUSE_OPENDIR should not be exported"));
 		/* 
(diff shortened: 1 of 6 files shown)


asomers:
> @asomers would you be able to look at this?

Yes, but probably not until the weekend after next.

sys/fs/fuse/fuse_vnops.c
```diff
@@ -3201,24 +3197,6 @@ fuse_vnop_vptofh(struct vop_vptofh_args *ap)
...
-		 * guarantee that d_off will be valid after a directory is
-		 * closed and reopened.  So prohibit exporting FUSE file
-		 * systems that don't set that flag.
-		 *
-		 * But userspace NFS servers don't have this problem.
-                 */
-		SDT_PROBE2(fusefs, , vnops, trace, 1,
-			"VOP_VPTOFH without FUSE_NO_OPENDIR_SUPPORT");
-		return EOPNOTSUPP;
-	}
```
asomers, line 3221: The check is still important, even if the condition is not correct.  I think you should reinstate the check, but change the condition to `fsess_is_impl(mp, FUSE_OPENDIR)`.  If a fuse filesystem implements FUSE_OPENDIR, then it cannot be correctly exported via NFS.  Then you'll also need to add a `fsess_set_impl` to `fuse_filehandle_open`.

Finally, we should probably stop returning an error.  Instead, we should probably allow the operation to succeed, but use a `fuse_warn` to warn the sysadmin.  But that's beyond the scope of this PR.

asomers (changes requested):
Thanks for finding this!  I agree with your change, and it should make a small boost to users of libfuse (though I see that some non-libfuse users do set that flag in the fuse_init_out response).  I just have one request regarding NFS exports.

asomers (approved):

asomers:
Merged as f0f596bd955e5b48c55db502e79fc652ac8970d3 .

PR #1512, opened 2024-11-03
Add logger for /src/usr.bin/mdo/mdo.c: a logger like sudo.

Rin0913:
We should add a logger for MAC mechanism. However, the logger for mac_do needs further discussions, and the tool use MAC framework should also be logging itself.

The output log format is like following:
```
Nov  4 03:46:54 bsd-workstation mdo[26618]: USER: rin; failed to call initgroups: 1
Nov  4 03:47:12 bsd-workstation mdo[26622]: USER: root; COMMAND=/bin/sh
```


delphij:
I think this (adding logging support) is the right direction but logging in the utility is probably not a very good choice because when mdo is loaded and enabled, an approved users can simply call `setuid()` and/or `setgid()` in their applications to the allowed user / groups themselves, which will not trigger any logging...

delphij:
> I think this (adding logging support) is the right direction but logging in the utility is probably not a very good choice because when mdo is loaded and enabled, an approved users can simply call `setuid()` and/or `setgid()` in their applications to the allowed user / groups themselves, which will not trigger any logging...

Scratch that, I have misread the code.  The module does require caller to be the hardcoded `/usr/bin/mdo`.

delphij (changes requested):
Please see comments inline.

usr.bin/mdo/mdo.c
```diff
@@ -22,10 +23,27 @@ usage(void)
 	exit(EXIT_FAILURE);
 }
 
+static char*
+join_strings(char **arr, int size, const char *delimiter)
```
delphij, line 27: Since this function is only used once, it can be folded into `main()`

Rin0913: I think the separation would make the code easier to read and without loss the original code logic?

usr.bin/mdo/mdo.c
```diff
@@ -45,31 +63,58 @@ main(int argc, char **argv)
...
 	if (*argv == NULL) {
 		const char *sh = getenv("SHELL");
 		if (sh == NULL)
 			sh = _PATH_BSHELL;
+		syslog(LOG_AUTH | LOG_INFO,
+			   "USER: %s; COMMAND=%s",
+			   original_username, sh);
 		execlp(sh, sh, "-i", NULL);
 	} else {
+		syslog(LOG_AUTH | LOG_INFO, "USER: %s; COMMAND=%s",
```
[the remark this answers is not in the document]
Rin0913: Because the memory block of new generated command string is allocated by malloc, maybe I could just free the memory block after logging just like this?

image

As for using sbuf, I believe if we can guarantee that the `join_strings` function is well-implemented, we don't have to use it because there is no possibility of buffer overflows.

What do you think?

usr.bin/mdo/mdo.c
```diff
@@ -22,10 +23,29 @@ usage(void)
 	exit(EXIT_FAILURE);
 }
 
+static char*
+join_strings(char **arr, int size, const char *delimiter)
+{
+	int total_length = 0, delimiter_length = strlen(delimiter);
```
rilysh, line 29: There are several branches in these loops, also I think it's a good idea to print the function name where it fails.
As suggested by [delphij](https://github.com/freebsd/freebsd-src/pull/1512/files#r1845332811), you can also use sbuf(9) so that you can probably eliminate this wrapper function.

Slightly changed
```c
static char *join_strings2(char **arr, int nlen, const char *delim)
{
	char *buf;
	size_t dlen, tlen, i;

	tlen = 0;
	dlen = strlen(delim);
	for (i = 0; i < nlen; i++)
		tlen += strlen(arr[i]) + dlen;

	if ((buf = malloc(tlen + 1)) == NULL)
		err(1, "malloc()");

	*buf = '\0';
        strcat(buf, arr[0]);
	for (i = 1; i < nlen; i++) {
		strcpy(buf + strlen(buf), delim);
		strcpy(buf + strlen(buf), arr[i]);
	}

	return (buf);
}
```

usr.bin/mdo/mdo.c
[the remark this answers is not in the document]
delphij: Since the VM space is going to be discarded anyway soon (with a successful `execvp()` or the `err()` below), the `free()` can be omitted here.

`sbuf(9)` is a "string builder", it keeps track of the length of string and handles buffer management.  On the other hand, doing it your own, especially doing `strcat()` in a loop would have to iterate each characters in the string built so far in order to append to them.  Therefore in most cases, using `sbuf(9)` is preferred for this kind of scenario.

usr.bin/mdo/mdo.c
delphij: I don't really have strong opinion here, it's more of a personal preference.

However, this approach also matches existing patterns of other similar situations: from readability point of view, while separating a more complex function is generally helpful (e.g. this program would do some complicated operations A, B, and C in order, and these operations are implemented as separate functions), it can actually hurt readability to overly wrap simple operations.  My personal rule of thumb is to not extract an operation into a function if a) it's only used once, and b) the function itself plus the caller's code block would fit one page or screen (about 20-25 lines) worth of code block.

Note that there are other similar situations where it's done in place instead of in a separate function, for example in `apply(1)` and `newsyslog(8)`.

Rin0913:
@OlCe2 

Hi, could you please help me review this PR?
Thank you very much!

bsdimp:
@OlCe2 Any final comments? This seems reasonable to me. I'm on the fence whether @bapt 's comments would help or not... it seems like maybe only a marginal improvement unless I'm badly misreading.

usr.bin/mdo/mdo.c
```diff
@@ -6,14 +6,14 @@
 
 #include <sys/limits.h>
 #include <sys/ucred.h>
-
```
OlCe2, line 9: Please keep this newline to separate headers in 'sys/' from the others.

usr.bin/mdo/mdo.c
```diff
@@ -23,10 +23,33 @@ usage(void)
...
 
+static char *join_strings(char **arr, size_t nlen, const char *delim)
+{
+	char *buf;
+	size_t dlen, tlen, i;
+
+	tlen = 0;
+	dlen = strlen(delim);
+	for (i = 0; i < nlen; i++)
+		tlen += strlen(arr[i]) + dlen;
```
OlCe2, line 34: We are allocating `dlen` too many bytes.  But that's minor (doesn't affect functional correctness).

usr.bin/mdo/mdo.c
```diff
@@ -23,10 +23,33 @@ usage(void)
...
+
+	if ((buf = malloc(tlen)) == NULL)
+		err(1, "malloc()");
+
+	*buf = '\0';
+	strcpy(buf, arr[0]);
```
OlCe2, line 40: Undefined behavior on `nlen` being 0 (minor, but still).

usr.bin/mdo/mdo.c
```diff
@@ -48,31 +71,29 @@ main(int argc, char **argv)
 	argc -= optind;
 	argv += optind;
 
+	openlog("mdo", LOG_PID | LOG_CONS, LOG_USER);
+	strcpy(original_username, pw->pw_name);
```
OlCe2, line 75: Risk of overflow. Please use strlcpy() instead.

OlCe2, line 75: `pw` can be NULL, have to handle it.  Perhaps by just putting the raw ID in `original_username`.

usr.bin/mdo/mdo.c
```diff
@@ -89,16 +110,24 @@ main(int argc, char **argv)
 		    SETCREDF_SUPP_GROUPS;
 	}
 
-	if (setcred(setcred_flags, &wcred, sizeof(wcred)) != 0)
+	if (setcred(setcred_flags, &wcred, sizeof(wcred)) != 0) {
+		syslog(LOG_ERR, "calling setcred() failed");
```
OlCe2, line 114: It is strange not to output user information here, whereas you do it anywhere else.

usr.bin/mdo/mdo.c
```diff
@@ -48,31 +71,29 @@ main(int argc, char **argv)
...
+
 	if ((pw = getpwnam(username)) == NULL) {
 		if (strspn(username, "0123456789") == strlen(username)) {
 			const char *errp = NULL;
 			uid_t uid = strtonum(username, 0, UID_MAX, &errp);
-			if (errp != NULL)
-				err(EXIT_FAILURE, "invalid user ID '%s'",
-				    username);
+			if (errp != NULL) {
+				syslog(LOG_ERR, "Failed to login: %s", username);
```
OlCe2, line 82: Why this different error message?  `Failed to login` in this context doesn't make any sense to me.

usr.bin/mdo/mdo.c
```diff
@@ -48,31 +71,29 @@ main(int argc, char **argv)
...
+				err(EXIT_FAILURE, "invalid user ID '%s'", username);
+			}
 			pw = getpwuid(uid);
 		}
-		if (pw == NULL)
+		if (pw == NULL) {
+			syslog(LOG_ERR, "Failed to login: %s", username);
```
OlCe2, line 88: Same as above.

usr.bin/mdo/mdo.c
```diff
@@ -48,31 +71,29 @@ main(int argc, char **argv)
...
 
 	if (!uidonly) {
-		/*
-		 * If there are too many groups specified for some UID, setting
-		 * the groups will fail.  We preserve this condition by
-		 * allocating one more group slot than allowed, as
-		 * getgrouplist() itself is just some getter function and thus
-		 * doesn't (and shouldn't) check the limit, and to allow
-		 * setcred() to actually check for overflow.
-		 */
```
OlCe2, line 75: Please do not remove useful comments.

usr.bin/mdo/mdo.c
```diff
@@ -23,10 +23,33 @@ usage(void)
...
+
+	if ((buf = malloc(tlen)) == NULL)
+		err(1, "malloc()");
+
+	*buf = '\0';
+	strcpy(buf, arr[0]);
+	for (i = 1; i < nlen; i++) {
+		strcpy(buf + strlen(buf), delim);
+		strcpy(buf + strlen(buf), arr[i]);
+	}
```
OlCe2, line 44: This is terribly inefficient (quadratic)... As advised by others, please use `open_memstream()` instead.

OlCe2 (changes requested):
Sorry, forgot to click on "Request changes" while submitting.

OlCe2:
Given the lack of answers, and the fact that since then the code of `mdo(1)` has considerably changed, closing this.

Feel free to re-open, rebase changes and cater to comments when you have the time.

PR #1515, opened 2024-11-05
ppp.8: Fix typo cuad0 -> cuau0

jlduran:
Move [PR #248359](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=248359) here as part of a cleanup of personal old bug reports.

### DIFF
--- a/usr.sbin/ppp/ppp.8
+++ b/usr.sbin/ppp/ppp.8
@@ -23,7 +23,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd June 27, 2022
+.Dd November 18, 2024
 .Dt PPP 8
 .Os
 .Sh NAME
@@ -263,7 +263,8 @@ will force it to exit.
 .It Supports client callback.
 .Nm
 can use either the standard LCP callback protocol or the Microsoft
-CallBack Control Protocol (https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-CBCP/[MS-CBCP].pdf).
+CallBack Control Protocol
+.Pq Lk https://winprotocoldocs-bhdugrdyduf5h2e4.b02.azurefd.net/MS-CBCP/%5bMS-CBCP%5d.pdf .
 .It Supports NAT or packet aliasing.
 Packet aliasing (a.k.a.\& IP masquerading) allows computers on a
 private, unregistered network to access the Internet.
@@ -339,7 +340,7 @@ to check the packet flow over the
 link.
 .It Supports PPP over TCP and PPP over UDP.
 If a device name is specified as
-.Em host Ns No : Ns Em port Ns
+.Em host : Ns Em port Ns
 .Xo
 .Op / Ns tcp|udp ,
 .Xc
@@ -353,7 +354,7 @@ into synchronous mode.
 If
 .Nm
 is given a device specification of the format
-.No PPPoE: Ns Ar iface Ns Xo
+.No PPPoE\&: Ns Ar iface Ns Xo
 .Op \&: Ns Ar provider Ns
 .Xc
 and if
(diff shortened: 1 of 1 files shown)


usr.sbin/ppp/ppp.8
```diff
@@ -65,7 +65,7 @@ Refer to
 .Xr libalias 3
 for details on the technical side of the NAT engine.
 Refer to the
-.Sx NETWORK ADDRESS TRANSLATION (PACKET ALIASING)
+.Sx NETWORK ADDRESS TRANSLATION (PACKET ALIASING)\&
```
concussious, line 68:
```suggestion
.Sx NETWORK ADDRESS TRANSLATION Pq PACKET ALIASING
```

usr.sbin/ppp/ppp.8
```diff
@@ -2166,13 +2166,15 @@ set device /dev/cuau0
 set speed 115200
 .Ed
 .Pp
-Cuad0 is the first serial port on
+.Pa cuau0
```
concussious, line 2169: Nice catch! The reader is definitely not going to find `/dev/Cuad0`. However, I don't think they're going to find `cuau0` either.
```suggestion
.Pa cuaU0
```
This needs changed on line 545,604,1209,1212,2165,2166,2529,2563,2564,2565, and 4063 too.

This one is a major and serious typo that would make these instructions unusable for someone who didn't already know, so I'm wondering if we can mfc this to the beta.

concussious (commented):
Critical catch!

Removing redundant normal text macros is ok with me. There are a few more syntax errors on this page, but none user facing, except:

The linter complains about line 266 being too long, which is a seemingly dead link. Are you seeing that too?

usr.sbin/ppp/ppp.8
```diff
@@ -2166,13 +2166,15 @@ set device /dev/cuau0
...
+is the first serial port on
 .Fx .
```
bsdimp, line 2171: Yes and no. cuau0 is the first serial port that's connected to ACPI, ISA or PCI bus via one of the uart attachments FreeBSD has. This is true on all our tier1 platforms, in the absence of wiring or other issues.
I'd be mildly tempted to say that cuaU0 is the first USB serial port, but I'm not sure how to work that in, exactly.
ppp(8) was written when there were no USB attached serial ports to note (and none good enough to run PPP). Now, you're more likely to have a USB serial port than a hard wired one (modems, what are those :).

jlduran:
> The linter complains about line 266 being too long, which is a seemingly dead link. Are you seeing that too?

Ah yes, I guess the file is now at:
https://winprotocoldocs-bhdugrdyduf5h2e4.b02.azurefd.net/MS-CBCP/%5bMS-CBCP%5d.pdf

I would just:
```mdoc
.Pq https://winprotocoldocs-bhdugrdyduf5h2e4.b02.azurefd.net/MS-CBCP/%5bMS-CBCP%5d.pdf .
```

concussious:
> > The linter complains about line 266 being too long, which is a seemingly dead link. Are you seeing that too?
> 
> Ah yes, I guess the file is now at: https://winprotocoldocs-bhdugrdyduf5h2e4.b02.azurefd.net/MS-CBCP/%5bMS-CBCP%5d.pdf
> 
> I would just:
> 
> ```
> .Pq https://winprotocoldocs-bhdugrdyduf5h2e4.b02.azurefd.net/MS-CBCP/%5bMS-CBCP%5d.pdf .
> ```

Nice find! If you want to Pq, please also use Lk so that the link is clickable.

jlduran:
> Edit: although, this should really go in the see also section, but I do not object to fixing it in place.

Yes, my main objective is to silence the linter before applying the actual fix, not to thoroughly clean up the manual.
But I do plan on doing some clean up for other man pages in the near future™.

usr.sbin/ppp/ppp.8
jlduran: Yes, because it can be confusing.
Device naming in manual page changelog(-ish):

| Name | Introduced | Manual Page Fix |
| ---- | ---------- | ----- |
| `cua00` (SIO) | af57ed9fdcc4e2a81940fc2243c0d3949fc0e00f | - |
| `cuaa0` | ab9ea4a7256ce2e1289bd82a31ec5d210a9f21aa | c853fdb82db495d0ec307f1f81a9003d670747b0 |
| `cuad0` | 9cb9afbd68d2088291ec4017cf48dde099b68582 | 37c510c7c0c46b344026d7b0eb65d0cf91420c39 |
| `cuau0` (UART) | ae59037978535fd06e84ab87a298fc54ee084bbb | this PR[^1] |


[^1]: Missed in 839f11a4fe18e4ae2dd930766b551fa67e354735.

concussious: This feels a little rough, but it is a little rough already:
```
.Pa cuaU0
and
.Pa cuau0
are the first
.Xr usb 4
and
.Xr uart 4
serial ports found by
.Fx ,
respectively.
```

jlduran: Good! thank you for the input.

concussious (approved):

mhorne (approved):
Please go ahead with this.

PR #1517, opened 2024-11-07
Fix comment typos in ipfw's struct dyn_data

DamjanJovanovic:
The bcnt_fwd and bcnt_rev fields are the byte counters, while the pcnt_fwd and pcnt_rev fields are the packet counters.
Fix the comments that were swapped around.

### DIFF
--- a/bin/sh/sh.1
+++ b/bin/sh/sh.1
@@ -1,4 +1,6 @@
 .\"-
+.\" SPDX-License-Identifier: BSD-3-Clause
+.\"
 .\" Copyright (c) 1991, 1993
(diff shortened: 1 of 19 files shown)


jlduran:
This change is correct. From `sys/netpfil/ipfw/ip_fw_private.h`:
`pcnt` means `Packet counter`
`bcnt` means `Byte counter`

We may have to change the subject of the commit to fit within 50 characters or less, per our standards.

Thank you for your contribution!

DamjanJovanovic:
> We may have to change the subject of the commit to fit within 50 characters or less, per our standards.

It's now within 50 characters.

> Thank you for your contribution!

It's a pleasure.

PR #1520, opened 2024-11-07
p9fs.5: rename to p9fs.4

llfw:
This is a filesystem (driver), not a file format, so it should be in section 4.


Rin0913:
As my personal opinion, I think it seems to be reasonable to categorize in man 5 because P9 is nothing but a protocol instead of a device driver in the very begin. However, rename to p9f.4 is acceptable now while a manual should still exist describing the protocol itself.

llfw:
> In my opinion, it seems reasonable to categorize it under man 5 because P9 was originally a protocol instead of a device driver.

are protocols normally in section 5 ('File Formats')?  ip(4) is in section 4, for example.

concussious:
Section 5 is the File Formats Manual. All filesystem driver manuals have been moved to section 4 with the other drivers for FreeBSD 15.

Rin0913:
Sorry for that, I got confused. Section 5 is indeed for file format only.
I was thinking that the protocol format was included inside.
I apologized for my mistake.

bsdimp:
Filesystems have been moved to section 4 recently. They used to be in section 5.
So I'm not sure why this isn't correct.
p9fs landed in the middle of that transition.

markjdb:
> can we get this merged as it's obviously correct?

It's incomplete. Obsoletefiles.inc needs an entry `OLD_FILES+=usr/share/man/man5/p9fs.5.gz`.

llfw:
fixed

concussious (approved):

PR #1523, opened 2024-11-13
man: Misc syntax fixes

gperciva:


gperciva:
@mhorne 

This includes the fix to `\E`, `\F`, and `\L` that I mentioned in https://github.com/freebsd/freebsd-src/pull/1472#discussion_r1811129940.

stand/man/loader.efi.8
```diff
@@ -394,7 +394,7 @@ BootCurrent: 0001
 Timeout    : 2 seconds
 BootOrder  : 0000, 0001, 0003, 0004, 0005, 0006, 0001, 0008, 000A, 000B, 000C, 000E, 0007
 \&...
-+Boot0001* FreeBSD ZPOOL HD(1,GPT,b5d0f86b-265d-1e1b-18aa-0ed55e1e73bd,0x28,0x96000)/File(\EFI\FREEBSD\LOADER.EFI)
++Boot0001* FreeBSD ZPOOL HD(1,GPT,b5d0f86b-265d-1e1b-18aa-0ed55e1e73bd,0x28,0x96000)/File(\\EFI\\FREEBSD\\LOADER.EFI)
```
concussious, line 397: No?

lib/libpmc/pmc.dmc-620.3
```diff
@@ -130,7 +130,6 @@ The following PMC events are available:
 .Xr pmc.corei7 3 ,
 .Xr pmc.corei7uc 3 ,
 .Xr pmc.iaf 3 ,
-.Xr pmc.iaf 3 ,
```
concussious, line 133: Verrified, these two are a great catch.

usr.bin/gzip/gzip.1
```diff
@@ -189,7 +189,7 @@ Options on the command line will override anything in
 .Sh EXIT STATUS
 The
 .Nm
-utility exits 0 on success,
+utility exits 0 on success,
```
concussious, line 192: How do I compare spaces? Git diff only reports a diff if I change only this to a space, but... what is it?

stand/man/loader.efi.8
gperciva: What do you mean?

The current man-page renders like this:
```
+Boot0001* FreeBSD ZPOOL HD(1,GPT,b5d0f86b-265d-1e1b-18aa-0ed55e1e73bd,0x28,0x96000)/File(EEBSD
                                       nda0p1:/EFI/FREEBSD/LOADER.EFI /boot/efi//EFI/FREEBSD/LOADER.EFI
```

Are you suggesting that it's supposed to say `File(EEBSD)`?

I think it's supposed to display `File(\EFI\FREEBSD\LOADER.EFI)`.  In order to achieve that, the `\` need to be escaped.

usr.bin/gzip/gzip.1
gperciva: A normal space in 0x20 (in both ascii and utf-8).

UTF-8 also includes a "non-breaking space", which is the character pair 0xC2A0.  Non-breaking spaces are supposed to look the same as regular spaces (hence why they look the same on your console).

If you look at the raw bytes, then you can see the difference:
```
git diff | xxd
```
(hopefully your console shows color; look for `c2a0` near the end of the output)

stand/man/loader.efi.8
concussious: Huh, excuse me (now we're even 🤣). I didn't see that it's not rendering correctly.

You want `\e`.

Addl. Context: mandoc_char(7): `note that doubling it ('\\') is not the right way to output a backslash.`

gperciva: Aw, crap.  In pretty much every language, doubling-up a backslash is the way to print it.  There's 528 other `\\` in the man pages.  Easy to fix, but a pain to review.  I'll add that to my list.

Thanks for the fix!

concussious: No, thank you!

concussious: It's now rendering correctly. Sure wish we could get some Pp's after these .Ed's in the examples section. I think that's a syntax fix, but I don't object if you don't want to do it.

gperciva: I'm a bit leery of making that change, because it's a bit subjective.  For example, in
```
Copy the loader:

    # cp /boot/loader.efi /boot/efi/EFI/FREEBSD/LOADER.EFI
replacing the all caps part of the example with the proper path.
```

one change could be to remove the space before the example -- I mean, the "replacing" is clearly in the same "sentence" as the example.  (That's a style of writing that's relatively common in math and science, where an equation is "part of the same sentence".)

IMO, a better fix would be to write:
```
Copy the loader, after replacing the all-caps part of the example with the proper path:

    # cp /boot/loader.efi /boot/efi/EFI/FREEBSD/LOADER.EFI
```

... but that's getting into even more subjective judgement, rather than fixing objectively incorrect syntax.

I'll feel much more comfortable staying focused on objectively bad stuff (such as `\\`), and only move on to subjective writing after that's done.

concussious: Yes, I agree that the second example is a better fix.

PR #1524, opened 2024-11-13
manuals: fix "PP after SS | SH" warnings

gperciva:


gperciva:
@mhorne 

This fixes 82 instances of two mandoc warnings:
    skipping paragraph macro: PP after SS
    skipping paragraph macro: PP after SH

Fixes made by script in https://github.com/Tarsnap/freebsd-doc-scripts

sbin/ipf/ipmon/ipmon.8
```diff
@@ -27,7 +27,6 @@ ipmon \- monitors /dev/ipl for logged packets
 .B <filename>
 ]
 .SH DESCRIPTION
-.LP
```
gperciva, line 30: I was initially confused by this: `.LP` is a synonym for `PP`, and `mandoc -T lint` reports "PP after SH" even if it's actually `.LP`.

(This seems like slightly sub-optimal behaviour from `mandoc`, but it only applies to a few lines, so oh well.)

jlduran: Only of historical relevance:
- The `.LP` macro produces a justified, block paragraph
- The `.PP` macro produces a paragraph with a temporary indent for the first line

Depending on the manual of style used, the first paragraph after a heading should not be indented (https://practicaltypography.com/first-line-indents.html).

For `mdoc`, `Lp` is a deprecated synonym for `Pp`.

jlduran (approved):
I like these cross sectional standardizations.  Landing them, however is another story.
But given its triviality, I'm inclined to accept it.

Thank you!

sbin/ipf/ipmon/ipmon.8
gperciva: Interesting!  `man(7)` just says `LP  A synonym for PP`.  On Linux, `man(7)` says `Same as .PP (begin a new paragraph)`.  It looks like some nuance was lost when they were writing those man pages.

Did they change the first-line indent for `.PP` at some point?  `ipf.4` has multiple `.PP` paragraphs, but I can't see any first-line indents in the rendered `ipf(4)`.

... I'm wondering if that "temporary indent for the first line" gets merged with the normal indent within a `.SH` section (like html margin collapsing).

jlduran: It is/was an `ms` macro... Don't know exactly when it was introduced, so I wouldn't restrict my search to only manual pages. This macro is/was heavily used for technical reports.
- https://man.freebsd.org/cgi/man.cgi?query=groff_ms&apropos=0&sektion=0&manpath=FreeBSD+14.1-RELEASE+and+Ports&arch=default&format=html
- http://n-t-roff.github.io/heirloom/doctools/man.7.pdf

As for the evolution of the manuals in base, can't comment much, as I wasn't involved.

mhorne (approved):

PR #1527, opened 2024-11-15
man: Fix some "new sentence, new line" warnings

gperciva:


gperciva:
@mhorne

The "new sentence, new line" warnings are arguably the least interesting ones.  But they're by far the most numerous now (163 warnings; the next most frequent is 76) so I'm tackling it now.

While I was reviewing my script's output when fixing all of them, I realized that a good number (35 files) had exactly the same problem -- the (non-comment) copyright license didn't start a new line.  I then wrote a new function which handled those cases, and text-wrapped the remainder of the paragraph.

This PR only handles those cases -- licenses at the ends of man pages in lib/libgssapi/*.3.  Exactly the same fix in every file.

lib/libgssapi/gss_accept_sec_context.3
```diff
@@ -463,12 +463,12 @@ others, and derivative works that comment on or otherwise explain it
...
-included on all such copies and derivative works.  However, this
-document itself may not be modified in any way, such as by removing
-the copyright notice or references to the Internet Society or other
-Internet organizations, except as needed for the purpose of
-developing Internet standards in which case the procedures for
-copyrights defined in the Internet Standards process must be
+included on all such copies and derivative works.
+However, this document itself may not be modified in any way, such as
+by removing the copyright notice or references to the Internet
+Society or other Internet organizations, except as needed for the
```
concussious, line 469: These are ridiculous places to wrap these lines because of grep. You're going to split isoc into two lines? Really?

This is why you can't write doc with scripts.

lib/libgssapi/gss_acquire_cred.3
```diff
@@ -215,12 +215,12 @@ others, and derivative works that comment on or otherwise explain it
...
-included on all such copies and derivative works.  However, this
-document itself may not be modified in any way, such as by removing
-the copyright notice or references to the Internet Society or other
-Internet organizations, except as needed for the purpose of
-developing Internet standards in which case the procedures for
-copyrights defined in the Internet Standards process must be
+included on all such copies and derivative works.
+However, this document itself may not be modified in any way, such as
+by removing the copyright notice or references to the Internet
+Society or other Internet organizations, except as needed for the
```
concussious, line 221: Really?

lib/libgssapi/gss_add_cred.3
```diff
@@ -315,12 +315,12 @@ others, and derivative works that comment on or otherwise explain it
...
-included on all such copies and derivative works.  However, this
-document itself may not be modified in any way, such as by removing
-the copyright notice or references to the Internet Society or other
-Internet organizations, except as needed for the purpose of
-developing Internet standards in which case the procedures for
-copyrights defined in the Internet Standards process must be
+included on all such copies and derivative works.
+However, this document itself may not be modified in any way, such as
+by removing the copyright notice or references to the Internet
+Society or other Internet organizations, except as needed for the
```
concussious, line 321: You must be joking

lib/libgssapi/gss_add_oid_set_member.3
```diff
@@ -107,12 +107,12 @@ others, and derivative works that comment on or otherwise explain it
...
-included on all such copies and derivative works.  However, this
-document itself may not be modified in any way, such as by removing
-the copyright notice or references to the Internet Society or other
-Internet organizations, except as needed for the purpose of
-developing Internet standards in which case the procedures for
-copyrights defined in the Internet Standards process must be
+included on all such copies and derivative works.
+However, this document itself may not be modified in any way, such as
+by removing the copyright notice or references to the Internet
+Society or other Internet organizations, except as needed for the
```
concussious, line 113: If you know reviewer time is extemely limited, and you know that there are meaningful content patches to the manual pending review, is it appropriate to edit docs with scripts and not even bother testing or thinking critically about the result?

lib/libgssapi/gss_canonicalize_name.3
```diff
@@ -114,12 +114,12 @@ others, and derivative works that comment on or otherwise explain it
...
-included on all such copies and derivative works.  However, this
-document itself may not be modified in any way, such as by removing
-the copyright notice or references to the Internet Society or other
-Internet organizations, except as needed for the purpose of
-developing Internet standards in which case the procedures for
-copyrights defined in the Internet Standards process must be
+included on all such copies and derivative works.
+However, this document itself may not be modified in any way, such as
+by removing the copyright notice or references to the Internet
+Society or other Internet organizations, except as needed for the
```
concussious, line 120: No

lib/libgssapi/gss_display_name.3
```diff
@@ -128,12 +128,12 @@ others, and derivative works that comment on or otherwise explain it
...
-included on all such copies and derivative works.  However, this
-document itself may not be modified in any way, such as by removing
-the copyright notice or references to the Internet Society or other
-Internet organizations, except as needed for the purpose of
-developing Internet standards in which case the procedures for
-copyrights defined in the Internet Standards process must be
+included on all such copies and derivative works.
+However, this document itself may not be modified in any way, such as
+by removing the copyright notice or references to the Internet
+Society or other Internet organizations, except as needed for the
```
concussious, line 134: No

lib/libgssapi/gss_display_status.3
```diff
@@ -187,12 +187,12 @@ others, and derivative works that comment on or otherwise explain it
...
-included on all such copies and derivative works.  However, this
-document itself may not be modified in any way, such as by removing
-the copyright notice or references to the Internet Society or other
-Internet organizations, except as needed for the purpose of
-developing Internet standards in which case the procedures for
-copyrights defined in the Internet Standards process must be
+included on all such copies and derivative works.
+However, this document itself may not be modified in any way, such as
+by removing the copyright notice or references to the Internet
+Society or other Internet organizations, except as needed for the
```
concussious, line 193: No

concussious (changes requested):

lib/libgssapi/gss_accept_sec_context.3
gperciva: I originally wrote that as
```
included on all such copies and derivative works.
However, this
document itself may not be modified in any way, such as
```

but I thought people would object to the short `However, this` line.  I'm quite happy to submit that version if it's preferred.  (That would make the script considerably simpler!)

Also, this is boilerplate license text; very unlikely to attract grep.  And the phrase "Internet Society" appears in both the previous and subsequent paragraphs.

concussious: People's use cases for doc can be incredibly diverse. See the resolution of the `man after midnight` easter egg.

I would fold after `way,` and then start the next newline with isoc.

Commas are a great place to wrap. Wrapping has a surprising amount of considerations. Here's an excerpt from a recent discussion on https://reviews.freebsd.org/D47178

"Line folding is kinda a whole thing (no change suggested, just fyi):

First, roff asks nothing goes over 80 characters unless there's an emergency (see #4).

Second, in a standard bsd console, the bsd editor nvi uses 8 characters to the left of the screen, so you have 72 characters before wrapping occurs. So personally, I like to limit it there, because then it's a logical extension of self-hosting: our doc is written with our tooling in mind. Of course, people are welcome to use whatever they like, but this is BSD culturally conscious best practice.

Third, the mdoc style guide for authors at mandoc.bsd.lv teaches that when lines are wrapped, they're as wrapped in as close to the middle as possible, for many reasons.

Fourth, keep in mind not breaking up phrases, or breaking them up in a logical way for grep, since grep searches a line at a time.

So, afaik those are all the main considerations when choosing how to wrap lines."

concussious:
I am not going to cover my bad behavior by deleting it... in my defense, everyone who has been here for a good amount of time has done this at some point, but that doesn't make it okay. My behavior ITT was a violation of the hard-won and holy CoC by being unfriendly, and I apologize.

bsdimp:
I'm closing this. We need to have a better way of dealing with these disputes, to be honest, since this isn't the right place for hashing it out. I'll contact Graham as well for good ways to proceed.

PR #1528, opened 2024-11-15
manuals: Fix some .Bl -tag lists

gperciva:


gperciva:
@mhorne 

While working on my script to fix all `.Bl -tag` "missing width" problems, I realized that these files needed special handling.

The files originally had empty `.It` lines, which (with `.Bl -tag`) meant that it was essentially just a highly-indented display.  This PR makes them actual lists with an "item head" and "item body".

lib/msun/man/cospi.3
```diff
@@ -62,27 +62,21 @@ where p is the floating\(enpoint precision of
 .Ar x ,
 then the returned value is 1 and it has no significance.
 .Sh SPECIAL VALUES
-.Bl -tag
-.It
-.Fn cospi \*(Pm0
+.Bl -tag -width 11n
```
gperciva, line 65: I know that there's a preference for something like `-width "cospi blargle"`.  However, the item heads in this list contain rather special escaped roff commands, which don't make sense inside a `""`.  I therefore opted to go with the lowest `n` such that it displayed correctly in the terminal.

Optionally, we could make them all `indent-two` instead; that's 12 rather than 10 or 11 chars.

gperciva:
@concussious this is your speciality.  :)

concussious:
Graham. I really appreciate the work you are trying to do. If the errors are fixed in the manual, then we can introduce meaningful CI on them. This is a really big deal, the good work you are trying to do.

However. You trolling me and that hurts my feelings. I feel that this is extremely inappropriate because you are much older and more educated than me.

I explained to you very recently why @mhorne explained to me that we *do not* use *n as a width arguement, because it forces people to count. We never do that, and instead you could use a perfectly reasonable `-width "cospi(+0)"` or `-width "cospi(-0)"`, which lets someone see how wide it is if they are looking at the documentation, which is how good technical writers write documentation.

Even further, you could actually clean these pages up, because *every element in the list is broken*, and I showed you how I do that by tagging you in reviews, which you instead just criticized my work on baseless grounds.

Please, do not do this. If you want to fix these pages, I want that too, and I will help you. As it stands, you are jeering me, and I object to this change on grounds that you are not fixing the list and actively choosing to undermine our style conventions, and trolling me.

What do you gain out of making me embarass myself and leave? Do you want one less next generation in the project? Should it only be programmers and no technical writers, in your view? Or a good old boys club?

concussious (changes requested):
Graham. I really appreciate the work you are trying to do. If the errors are fixed in the manual, then we can introduce meaningful CI on them. This is a really big deal, the good work you are trying to do.

However. You trolling me and that hurts my feelings. I feel that this is extremely inappropriate because you are much older and more educated than me.

I explained to you very recently why @mhorne explained to me that we do not use *n as a width arguement, because it forces people to count. We never do that, and instead you could use a perfectly reasonable `-width "cospi(+0)"` or `-width "cospi(-0)"`, which lets someone see how wide it is if they are looking at the documentation, which is how good technical writers write documentation.

Even further, you could actually clean these pages up, because *every element in the list is broken*, and I showed you how I do that by tagging you in reviews, which you instead just criticized my work on baseless grounds.

Please, do not do this. If you want to fix these pages, I want that too, and I will help you. As it stands, you are jeering at me, and **I object to this change on grounds that you are not fixing the list and actively choosing to undermine our style conventions**, and trolling me.

What do you gain out of making me embarass myself and leave? Do you want one less next generation in the project? Should it only be programmers and no technical writers, in your view? Or a good old boys club?

concussious:
Someone will merge your stupid commit refusing to fixing the clearly visible problem, while I have been loudly making noise about a finished patch to man(1) for *months*, which has only recieved minimal bikeshedding while the options for search go unexplained in another release.

I am begging anyone that cares about the health of this project, this is the reason doc is critically understaffed. This is *extemely* abusive.

concussious:
While I'm here misbehaving, this kind of complete social disrespect caused the maintainer of steam_utils to *quit* several weeks ago.

Mark my words, there will be *no* attracting next generation talent without doc and video games.

gperciva:
... I am not trolling.  I honestly thought that I was being helpful.

I did not forget about the reasons for not liking `-width Xn`, but I really thought that this case was worth an exception to the general rule.  I specifically commented about this, so that it would be obvious to reviewers and they could tell me that even in this case, it wasn't appropriate to use `-width Xn`.

I will need more time to reflect on the rest of your comments.

lib/msun/man/cospi.3
gperciva: I rewrote these to use `-width "cospi(+n/2)"`.

The suggestion to use rendered output for the `-width` argument is brilliant.

concussious:
No rushed response requested, just additional context:

What you are intending/doing fixing the errors in the manual pages is *incredibly valuable*. We have a doc CI problem: when a manual page is changed, we build *every compiler toolchain* on *every architecture*. This is a massive waste of money and electricity/non-renewable resources, and still doesn't catch the errors. What we need to do (to start[^1]) is put a filter on the front e.g. `if [ *.[1-9] ]; mandoc -Tlint; exit 0`. However we can not do that yet with a usefil S/N because there are too many errors already. Your work to reduce the errors will make meaningful manual page CI possible, and it will save hundreds of pounds of coal per year. Your intent is litterally paying technical debt, which needs to happen to do other things. If I had a commit bit, I would merge all of these with adjustments or a second commit on top just fixing the things I discovered while testing your proposed changes. But that isn't the situation we have... and it maybe has nothing to do with you. The decision was made in the beginning of this project to give doc their own separate commit bits, because technical writers and programmers look at things with such massively different eyes, that these types of social problems have been consistently happening to everyone at the intersection between programming and technical writing since long before I was born (according to recentish fireside chat with freebsd founder Jordan Hubbard, if you like lore, highly recommended).

[^1]: Jan Bramkampf has also suggested some brilliant refinements in the ZFS Production User Call once we get that up.

gperciva:
A quick technical note:  out of the 3198 man files in my "not-contrib, not-insane" list, there are only 2 errors remaining.

If we tweak the file inclusion list to ignore those contrib files (or if FreeBSD wants to start keeping local modifications to those man pages), then

```
if [ *.[1-9] ]; mandoc -Tlint -Werror; exit 0
```

could be implemented within a few days.  (Note the `-Werror` argument, which ignores style and warnings.)

Since it wasn't at all clear how the project would want to approach errors in contrib files, I moved on to fixing warnings.

lib/msun/man/cospi.3
concussious: I'm sorry I wasn't clear when I explained it the previous times, this is entirely my failure to communicate this with you. I didn't come up with this, this is how it's always been afaik. It's the width of the rendered longest item in the list.

concussious (approved):
These lists are now entirely correct.

concussious:
The reason this was especially hurtful to me is because I write meaningful content all the time, and it often just sits there for months, including when I've promised users in scheduled, recorded talks that I'd get an explanation in, or fix broken search keywords.

List width alignment gets in very quickly because there's really no room for a judgement call there. There is a very clear cut way it is supposed to be done.

I promised users we would get manual search explained in the next release during a scheduled, recorded talk four months ago and it's been finished and sitting for three months.

With the imminent release, I've been especially distressed about being stonewalled into failing to deliver a promise, made on video no less. It's completely humiliating. Being called a div centering specialist, during this time, was quite literally adding insult to a permanent injury.

gperciva:
I've spent a few days thinking about it.

As far as this PR goes,
1. I (stupidly) misunderstood the instructions regarding the `-width` argument.
2. Based on my misunderstanding, I asked for an exemption by specifying `-width 11n`.
3. You disagreed that this needed an exemption, and showed me how it should be done.
4. The next morning when I read your comment, I immediately agreed that your method was superior, and uploaded a PR which fixed those 3 lines within 10 minutes.
5. You submitted a PR which clarifies style.mdoc(5).

To me, that seems an ideal way to handle submissions.  Granted, it would be better if I hadn't made a mistake, but it happens [1].  My error was spotted early in the review process, and I fixed it very quickly.

[1] Just a few days ago, a senior FreeBSD person committed an mdoc with a broken macro which would have triggered a `mandoc` error if it was checked.  I assume no malice on their part; such things are simply a consequence of not including `mandoc -Tlint -Werror` in CI.

However, you made many other comments.
- you accused me of trolling.
- you claimed that "every element in the list is broken", which I still don't understand.
- you questioned whether I only wanted "programmers and no technical writers" in FreeBSD.
- you referred to my work as a "stupid commit".

You also mentioned that I was "much older and more educated".  While that is probably true, I am quite aware that I am your junior in FreeBSD.  Your first commit was 11 months ago; mine was 2 months ago.  You are also much more familiar with mdoc and especially groff.

You also made many complaints which (as far as I can tell) have nothing to do with me:
- you've been "loudly making noise about a finished patch to man(1) for months".  I do not have a commit bit, and based on my understanding, it's normally years until somebody get a commit bit.  You're much closer to having one than me!  Also, if "months" is more than 2, then that patch was before I got involved at all.
- you mentioned problems with supporting new contributors.  I might agree with those complaints in principle, but I am not responsible for them, and I do not appreciate being blamed for them.
- Just this morning, you said that somebody called you a "div centering specialist".  That was certainly not me!
- You also mentioned that you made a promise in a recorded video four months ago, which in jeopardy.  I have no knowledge of that, and again, I only began contributing 2 months ago.

I will not abandon any PRs that I have submitted, but yesterday I began looking through bugzilla for anything in `Base System -> bin` and `Base System -> standards` that's within my skill set.  So far they all look beyond me, but on the other hand, tackling a difficult challenge is a good way to learn.

concussious:
Mistakes do happen to all of us. I hope you (and our community) will forgive, and excuse my mistake.

I even created this misunderstanding by failing to fix the instructions for list width alignment when I attempted it and failed in May.

mhorne (approved):

mhorne:
I only skimmed the comments here, rather than trying to read in detail. I hope some understanding has been reached.

I would like to affirm that both of you, @gperciva and @concussious, have provided valuable contributions recently, and I personally hope you will both continue to do so. For me it is obvious that you have different approaches to the same type of work, and such diversity is greatly beneficial, as long as we can disagree respectfully.

But I warn against taking anything _I_ say about how things "should" be done as gospel (hard rule). Even though I am an experienced committer, I am still learning as I go, especially concerning the minutia of man pages.

It is discouraging as a reviewer to approach a GitHub review for a "simple" change, and see dozens of back-and-forth replies. It makes me not want to get involved at all. Do not make me into a baby-sitter; I will not accept this role.

Here are the two things I would like to stress, and you can reflect on how each applies to you personally:
 - Perfection, in code or docs, does not exist and can never be fully satisfied. We can aim only for "better", but aiming for perfect is a recipe for disappointment. In this sense there can be no universal rules, only guidelines.
 - A mistake only becomes a "failure" when you attach your self-worth to the works you produce. In the same way, a reply only becomes a personal attack when you take it personally. We can choose to take a breath, step away, and re-approach a challenging reply in good faith.

Thanks again to you both.

PR #1536, opened 2024-11-20
Add syslogd message format tests and fix PR 220246

michal-os:
Added syslogd -O flag tests covering issues raised in PR 220246 [https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=220246](url) regarding the syslogd RFC 3164 message format not meeting the RFC recommendations in certain situations. Also included tests covering the RFC 5424 format.

To resolve PR 220246, added a new syslogd -O flag value that meets the RFC 3164 recommendations. Old format is kept as backwards compatibility was mentioned in the PR discussion.

800 lines is a bit large, but it is basically the same 3 tests repeated for each possible value of the -O flag. 

Commit summary:
1. (added after rebase) remove is_exclusive="true" from test metadata.
2. small refactor: pulled out some existing test functions for reuse; and added basic log file regex tests.
3. added tests for a syslog message forwarded from a "remote" host. Checks regex on log file and tcpdump (PRIVAL is only in packet for the bsd/rfc3164 format). These tests cover the main problem case in PR 220246.
4. added tests for a syslog message relayed by an intermediate host. Checks regex on log file and tcpdump. The existing bsd/rfc3164 format is most deficient in this case.
5. added new "strict" format with the same basic/forwarded/relayed tests.


markjdb:
Would it be possible to rebase this PR? Some changes landed in `main` this week which make syslogd run in capability mode, but they were fairly invasive. If you need some help, please let me or @jakesfreeland know and we can help with the rebase.

michal-os:
Hi, I made the updates to make it work with the recent syslogd changes. I was able to reuse more of the existing test functions.

I also removed the "is_exclusive=true" from the test metadata. Based on this status report [https://www.freebsd.org/status/report-2024-07-2024-09/kyua/](url) the switch to the jail execution environment should make is_exclusive no longer necessary.

markjdb:
> Hi, I made the updates to make it work with the recent syslogd changes. I was able to reuse more of the existing test functions.

Thank you!

> I also removed the "is_exclusive=true" from the test metadata. Based on this status report [https://www.freebsd.org/status/report-2024-07-2024-09/kyua/](url) the switch to the jail execution environment should make is_exclusive no longer necessary.

Yes, that makes sense to me.

The change looks good to me. Thanks for your patience. A few minor comments:
- What's the purpose of defining `rfc3164-strict` and `strict` when they're handled identically? It seems unnecessarily confusing.
- If what happens if `-O rfc5424 -O rfc3164-strict` is specified? I'm not sure that the result makes sense. Rather than having two boolean flags, perhaps a single tri-state enum value is better?
- There's no need to set `required.progs` for logger and tcpdump, as they're always available in the base system.

michal-os:
Thanks for reviewing my changes.

> The change looks good to me. Thanks for your patience. A few minor comments:
> 
> What's the purpose of defining `rfc3164-strict` and `strict` when they're handled identically? It seems unnecessarily confusing.

This ultimately came to a style choice because the existing implementation uses two option values for each format setting (e.g. `syslog` or `rfc5424` for setting RFC 5424 format).

I will remove `strict` and keep `rfc3164-strict`, as I agree, two seem unnecessary.

I contemplated a broader change to only permit the values `bsd|rfc3164|rfc5424` and use them for setting the legacy RFC 3164 format, this new strict RFC 3164 format, and RFC 5424 format, respectively. But the impact to backwards-compatibility of removing `syslog` and modifying `rfc3164` seemed too high.

> If what happens if `-O rfc5424 -O rfc3164-strict` is specified? I'm not sure that the result makes sense. Rather than having two boolean flags, perhaps a single tri-state enum value is better?

I did not consider this scenario. I will review the behaviour, and make changes/add a test if necessary. Using an enum does make more sense.

> There's no need to set `required.progs` for logger and tcpdump, as they're always available in the base system.

Thanks, I will remove those lines.

I will make the proposed changes, and I will update this pull request once they are ready.

michal-os:
I have completed the changes to address the minor comments.

> If what happens if -O rfc5424 -O rfc3164-strict is specified? I'm not sure that the result makes sense. Rather than having two boolean flags, perhaps a single tri-state enum value is better?

It was possible to set RFC 5424 mode with the RFC 3164-strict mode. It functionally just operated in the RFC 5424 mode, but I implemented the enum so the internal logic make more sense and no longer permits that edge case.

markjdb (approved):

PR #1537, opened 2024-12-02
sfxge: defer ether_ifattach to when ifmedia_init is done

fichtner:
Otherwise the ifmedia subsystem may call unguarded NULL function pointers.  Same issue that was fixed for cxgb(4) in f2daf8995.

Also see:	https://github.com/opnsense/src/issues/228
MFC after:	1 week


gmshake (approved):
Looks good to me.

kev009:
0482974b2c021a8abbc044d3e2b1240250f791e7 committed

fichtner:
@gmshake @kev009 superb, thanks! shall I propose a workaround for https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=277849 too?

gmshake:
> @gmshake @kev009 superb, thanks! shall I propose a workaround for https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=277849 too?
The original reporter [Martin](mailto:mm@FreeBSD.org) does not response to my first request, lets be a little patient.

PR #1538, opened 2024-12-02
github: optimize style workflow

VexedUXR:
Optimize the style workflow by limiting the fetch depth to the number of commits in the PR + 1. Unfortunately, the only way to add one seems to be to create a new step and do it there.
I conducted a small test here: https://github.com/VexedUXR/freebsd-src/pull/3

This reduces the execution time to ~50s.

Also make it so the workflow only triggers on source files.

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -1,10 +1,6 @@
 name: Style Checker
 
-# Runs my simple style(9) checker on any pushes or pull requests.  It could be
-# optimized by fetching the pull request head branch back to main revisions and
-# running on that. That would reduce the run time from 3-4 minutes down to 30-40
-# seconds. Getting the right series of clone + fetches to get that iteratively
-# is proving elusive, so optimizations welcome.
+# Runs my simple style(9) checker on pull requests.
 
 on:
   pull_request: # maybe pull_request_target
@@ -19,10 +15,15 @@ jobs:
     name: Style Checker
     runs-on: ubuntu-latest
     steps:
+      # Unfortunately there doesn't seem to be a way to
+      # do this without an extra step.
+      - name: Get depth
+        run: |
+          echo "DEPTH=$(( ${{ github.event.pull_request.commits }} + 1 ))" >> $GITHUB_ENV
       - name: checkout
         uses: actions/checkout@v4
         with:
-          fetch-depth: 0
+          fetch-depth: ${{ env.DEPTH }}
           ref: ${{ github.event.pull_request.head.sha }}
       - name: Install packages
         run: |
@@ -30,5 +31,5 @@ jobs:
           sudo apt-get -yq --no-install-suggests --no-install-recommends install perl
       - name: Run checker
         run: |
-          sha=$(git merge-base ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }})
+          sha=$(git rev-parse HEAD~${{ github.event.pull_request.commits }})
           tools/build/checkstyle9.pl --github ${sha}..${{ github.event.pull_request.head.sha }}
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -6,6 +6,7 @@ on:
   pull_request: # maybe pull_request_target
     branches: [ main ]
     types: [ opened, reopened, edited, synchronize ]
+    paths: [ '**.S', '**.c', '**.cc', '**.cpp', '**.h', '**.hh', '**.hpp' ]
 
 permissions:
   contents: read


.github/workflows/style.yml
```diff
@@ -1,15 +1,12 @@
...
-# running on that. That would reduce the run time from 3-4 minutes down to 30-40
-# seconds. Getting the right series of clone + fetches to get that iteratively
-# is proving elusive, so optimizations welcome.
+# Runs my simple style(9) checker on pull requests.
 
 on:
   pull_request: # maybe pull_request_target
     branches: [ main ]
     types: [ opened, reopened, edited, synchronize ]
+    paths: [ '**.c', '**.cc', '**.h', '**.hh' ]
```
jlduran, line 9: I think `checkstyle9.pl` lints/checks the following source file types:
https://github.com/freebsd/freebsd-src/blob/6749f059a586ff62f2909134c4f4599c82fabc5e/tools/build/checkstyle9.pl#L15

bsdimp: We need .S at least in the github path statement.

.github/workflows/style.yml
```diff
@@ -19,16 +15,21 @@ jobs:
     name: Style Checker
     runs-on: ubuntu-latest
     steps:
+      # Unfortunately there doesn't seem to be a way to
+      # do this without an extra step.
+      - name: Get depth
+        run: |
+          echo "DEPTH=$(( ${{ github.event.pull_request.commits }} + 1 ))" > $GITHUB_ENV
```
emaste, line 23: I guess `$GITHUB_ENV` is magic and might be worth a reference in the commit message.

jlduran: And it should be `>>` in order not to overwrite it (even if it doesn't matter if it's overwritten).
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-environment-variable

.github/workflows/style.yml
VexedUXR: > I think `checkstyle9.pl` lints/checks the following source file types:
> 
> https://github.com/freebsd/freebsd-src/blob/6749f059a586ff62f2909134c4f4599c82fabc5e/tools/build/checkstyle9.pl#L15

Hm, maybe we should add `cc` and `hh` here. Though, I don't know what the style rules are with C++ specific things (do we have any?).

I'll only add `.S`, as Warner suggested, since we don't really enforce a style on the others.

.github/workflows/style.yml
VexedUXR: > I guess `$GITHUB_ENV` is magic and might be worth a reference in the commit message.

I added it to the commit message, let me know if it's too vague. :)

.github/workflows/style.yml
```diff
@@ -1,15 +1,12 @@
...
-# running on that. That would reduce the run time from 3-4 minutes down to 30-40
-# seconds. Getting the right series of clone + fetches to get that iteratively
-# is proving elusive, so optimizations welcome.
+# Runs my simple style(9) checker on pull requests.
 
 on:
   pull_request: # maybe pull_request_target
     branches: [ main ]
     types: [ opened, reopened, edited, synchronize ]
+    paths: [ '**.S', '**.c', '**.cc', '**.h', '**.hh' ]
```
bsdimp, line 9: Needs **.cpp and **.hpp as well.
Though the style script also needs an update for those two. I'll take care of that and have it stop commenting on perl and python.

.github/workflows/style.yml
```diff
@@ -19,16 +16,21 @@ jobs:
     name: Style Checker
     runs-on: ubuntu-latest
     steps:
+      # Unfortunately there doesn't seem to be a way to
+      # do this without an extra step.
+      - name: Get depth
+        run: |
+          echo "DEPTH=$(( ${{ github.event.pull_request.commits }} + 1 ))" >> $GITHUB_ENV
```
bsdimp, line 23: I'm surprised we need this.

bsdimp:
OK. Almost there. Checked on something and we need .cpp and .hpp since we have those in our non-contrib tree as well.

bsdimp:
OK. Updated checkstyle9.pl regexp in 0207010f595d. Please rebase and let me know how it works out. :)

.github/workflows/style.yml
VexedUXR: Same here.
I would've thought there would at least be basic mathematical operators here: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#operators

VexedUXR:
Rebased and added added `**.cpp` and `**.hpp`

bsdimp:
OK. I'm happy

VexedUXR:
Thanks!

PR #1540, opened 2024-12-06
find: Re-enable the -acl primary on FreeBSD

bsdjhb:
This was made conditional to support cross-builds, but the relevant header wasn't included so it was never enabled for native builds.

PR:		278124


bsdjhb:
PR is just for testing cross-builds

PR #1541, opened 2024-12-08
iwlwifi.4: bugs: tracking and clarification

grahamperrin:
```text
Enhancements to the word 'certainly' might certainly include a link to
at least one of the impressively well-organised tracking reports in
Bugzilla.

MFC: certainly.
```

### DIFF
--- a/share/man/man4/iwlwifi.4
+++ b/share/man/man4/iwlwifi.4
@@ -27,7 +27,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd November 10, 2024
+.Dd December 8, 2024
 .Dt IWLWIFI 4
 .Os
 .Sh NAME
@@ -250,11 +250,21 @@ The
 driver first appeared in
 .Fx 13.1 .
 .Sh BUGS
-Certainly.
+.%U <https://bugs.freebsd.org/bugzilla/showdependencytree.cgi?id=iwlwifi&hide_resolved=1>
+presents the dependency tree for the iwlwifi tracking meta-bug, which:
 .Pp
-While
-.Nm
-supports 802.11a/b/g/n/ac/ax modes,
-the compatibility code currently only supports 802.11a/b/g modes.
-Support for 802.11n/ac/ax is yet to come.
-802.11ax and 6Ghz support are planned.
+.Bl -bullet -compact
+.It
+blocks the LinuxKPI based wireless drivers tracking meta-bug
+.It
+indirectly blocks the general wireless tracking meta-bug.
+.Pp
+Notable limitations: 
+.Pp
+.Bl -bullet -compact
+.It
+802.11a/b/g modes are supported
+.It
+802.11n/ac/ax and 6Ghz are not yet supported
+.It
+operating system resume (wake from sleep) may be unreliable.


share/man/man4/iwlwifi.4
```diff
@@ -250,11 +250,21 @@ The
...
 .Pp
-While
-.Nm
-supports 802.11a/b/g/n/ac/ax modes,
-the compatibility code currently only supports 802.11a/b/g modes.
-Support for 802.11n/ac/ax is yet to come.
-802.11ax and 6Ghz support are planned.
+.Bl -bullet -compact
+.It
+blocks the LinuxKPI based wireless drivers tracking meta-bug
```
grahamperrin, line 258:
```suggestion
blocks the LinuxKPI-based wireless drivers tracking meta-bug
```

share/man/man4/iwlwifi.4
```diff
@@ -250,11 +250,21 @@ The
...
+.It
+indirectly blocks the general wireless tracking meta-bug.
```
grahamperrin, line 260:
```suggestion
indirectly blocks the general wireless drivers tracking meta-bug.
```

For consistency with the preceding list item, add the word _drivers_.

grahamperrin:
Affects the _bugs_ section of **iwlwifi(4)**:

- <https://man.freebsd.org/cgi/man.cgi?query=iwlwifi&sektion=4&manpath=freebsd#BUGS>

I do always smile at the dryness of "Certainly." … equally, I am certain that a broad audience will positively appreciate the highly organised approach that is demonstrated by the trees of the tracking reports. 

----

@bzfbd

bsdimp:
@erikarn @bzfbd Can you two look at this and comment to see if this is still relevant or not?

bzfbd:
Half, I'll handle it

share/man/man4/iwlwifi.4
```diff
@@ -250,11 +250,21 @@ The
...
+.Pp
```
bzfbd, line 261: I believe this sections is overcome by events as the man page has been updated.
The suspend problem can be found in the list of bugs.

share/man/man4/iwlwifi.4
```diff
@@ -250,11 +250,21 @@ The
 driver first appeared in
 .Fx 13.1 .
 .Sh BUGS
-Certainly.
+.%U <https://bugs.freebsd.org/bugzilla/showdependencytree.cgi?id=iwlwifi&hide_resolved=1>
```
bzfbd, line 253: Can we simplify this to:
```
.%U <https://bugs.freebsd.org/bugzilla/showdependencytree.cgi?id=iwlwifi>
presents the known open and closed bugs related to
.Nm .
```
I do not think the entire bug tree is interesting to anyone here. If at best these links would go into the other respective man pages but then they are not entirely maintained as iwlwifi is so I am not sure.

bzfbd (changes requested):
Let me know what you think;  we can likely reduce it to three lines these days, which I do like as an experiment to see if it helps adding the bugzilla URL to the man page.

share/man/man4/iwlwifi.4
concussious: Yes, we absolutely can, I would see how it looks (away from desk ATM) to take it even further like so:
```
.Lk https://bugs.freebsd.org/bugzilla/showdependencytree.cgi?id=iwlwifi Nm known bugs
```
Then, in graphical manual viewers, there would just be a clickable link that says "iwlwifi known bugs". Super forward compatible.

bzfbd:
The link to bugzilla for iwlwifi.4 is in.  Thank you for the idea!

I believe the other bits are either overcome by time or too much for here.

PR #1546, opened 2024-12-12
savecore: add missing call to cap_openlog when in capabilities mode

stephane-rochoy-stormshield:
When run in capabilities mode, `savecore` seems to lack a call to `cap_openlog`. Notably it break the use of `LOG_PERROR`.

With the original one:

    $ /sbin/savecore -vC
    checking for kernel dump on device /dev/gpt/swapfs
    No dump exists

With the patched one:

    $ ./savecore -vC
    checking for kernel dump on device /dev/gpt/swapfs
    savecore 1424 - - /dev/gpt/swapfs: Permission denied
    No dump exists

### DIFF
--- a/sbin/savecore/tests/log_test.sh
+++ b/sbin/savecore/tests/log_test.sh
@@ -1,0 +1,25 @@
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+# Copyright (c) 2025 Stéphane Rochoy <stephane.rochoy@stormshield.eu>
+#
+
+atf_test_case log_perror
+log_perror_head()
+{
+	atf_set "descr" "Test LOG_PERROR behavior"
+}
+log_perror_body()
+{
+	atf_check -s exit:1 \
+	          -o ignore \
+	          -e save:savecore.err \
+	    savecore -vC /dev/missing
+	grep -qE 'savecore [0-9]+ - - /dev/missing: No such file or directory' savecore.err \
+	    || atf_fail "missing/invalid error output"
+}
+
+atf_init_test_cases()
+{
+	atf_add_test_case log_perror
+}
--- a/sbin/savecore/savecore.c
+++ b/sbin/savecore/savecore.c
@@ -101,6 +101,9 @@
 #define	STATUS_GOOD	1
 #define	STATUS_UNKNOWN	2
 
+#define LOG_OPTIONS  LOG_PERROR
+#define LOG_FACILITY LOG_DAEMON
+
 static cap_channel_t *capsyslog;
 static fileargs_t *capfa;
 static bool checkfor, compress, uncompress, clear, force, keep;	/* flags */
@@ -1409,6 +1412,7 @@ init_caps(int argc, char **argv)
 		exit(EXIT_FAILURE);
 	}
 	cap_close(capcas);
+	cap_openlog(capsyslog, "savecore", LOG_OPTIONS, LOG_FACILITY);
 }
 
 static void
@@ -1436,7 +1440,7 @@ main(int argc, char **argv)
 	savedir = ".";
 	comp_desired = KERNELDUMP_COMP_NONE;
 
-	openlog("savecore", LOG_PERROR, LOG_DAEMON);
+	openlog("savecore", LOG_OPTIONS, LOG_FACILITY);
 	signal(SIGINFO, infohandler);
 
 	argc = xo_parse_args(argc, argv);
--- a/sbin/savecore/tests/Makefile
+++ b/sbin/savecore/tests/Makefile
@@ -1,3 +1,3 @@
-ATF_TESTS_SH=	livedump_test
+ATF_TESTS_SH=	livedump_test log_test
 
 .include <bsd.test.mk>


stephane-rochoy-stormshield:
I'm looking to add a test as it should be trivial.

emaste:
Looks good. CC @oshogbo.

Also @bsdimp do we have advice on rebasing/squashing fixup commits in pull requests?

stephane-rochoy-stormshield:
> Looks good. CC @oshogbo.
> 
> Also @bsdimp do we have advice on rebasing/squashing fixup commits in pull requests?

Yes you have ;) I've put this PR in draft mode for now. I'll squash once ready. 
I'm discovering atf(7) and kyua(1) to add at least a basic test.

emaste:
Oh yes I didn't notice the Draft tag. The C change itself looks good :)

bsdimp:
git rebase -i main
squash all the tests into one commit and all the stuff in the actual binary into one. I think that's just the first one and all the rest squashed.

sbin/savecore/tests/log_test.sh
```diff
@@ -0,0 +1,26 @@
...
+
+atf_test_case log_perror
+log_perror_head()
+{
+	atf_set "descr" "Test LOG_PERROR behavior"
+}
+log_perror_body()
+{
+	atf_check -s exit:1 -o ignore -e save:savecore.err savecore -vC /dev/missing
+  grep -qE 'savecore [0-9]+ - - /dev/missing: No such file or directory' savecore.err
```
oshogbo, line 15: I guess the indent is wrong here.

sbin/savecore/tests/log_test.sh
```diff
@@ -0,0 +1,26 @@
...
+  [ $? -eq 0 ] && atf_pass || atf_fail "missing/invalid error output"
```
oshogbo, line 16: Don't we want to use a regular if statement here to increase readability?

stephane-rochoy-stormshield:
Sorry for the initial noise. I cleaned the commits.

But one question remain about the test I propose to add: what is the proper way to test the instance of `savecore` from the installed build, i.e., not the one on the system? I find it convenient to `make world DESTDIR=…` but the tests don't seems to be aware of `$DESTDIR`.

stephane-rochoy-stormshield:
> But one question remain about the test I propose to add: what is the proper way to test the instance of `savecore` from the installed build, i.e., not the one on the system? I find it convenient to `make world DESTDIR=…` but the tests don't seems to be aware of `$DESTDIR`.

@oshogbo Do you have an opinion on this? Do you want me to just ignore `$DESTDIR`?

stephane-rochoy-stormshield:
Any news on this topic? Should I just call `savecore` instead of `${DESTDIR:-''}/sbin/savecore`?

asomers:
> Any news on this topic? Should I just call `savecore` instead of `${DESTDIR:-''}/sbin/savecore`?

Yes, you should do that.  The test suite is intended to run against an installed system.  There aren't any other tests tweaked to work with DESTDIR.  However, you can always alter PATH when running this test yourself.

oshogbo (approved):

oshogbo:
Sorry, for late response. LGTM.

stephane-rochoy-stormshield:
I adjusted copyright to 2025.

sbin/savecore/tests/log_test.sh
```diff
@@ -0,0 +1,29 @@
...
+	          -o ignore \
+	          -e save:savecore.err \
+	    savecore -vC /dev/missing
+	grep -qE 'savecore [0-9]+ - - /dev/missing: No such file or directory' savecore.err \
+	    || atf_fail "missing/invalid error output"
+}
+log_perror_cleanup()
+{
+	rm savecore.err
+}
```
asomers, line 24: You don't need this cleanup, because Kyua always deletes the entire temporary PWD.  Also, this cleanup section wasn't even being invoked, since you don't have the cleanup keyword on line 7.
```suggestion
```

asomers (changes requested):
While I don't dispute the results of your testing, I don't understand why it works.  Shouldn't it be sufficient to run openlog before cap_enter?  I don't understand why the program even needs cap_syslog in the first place.  IIUC, cap_syslog is only needed if you must open a syslog connection _after_ cap_enter.

sbin/savecore/tests/log_test.sh
stephane-rochoy-stormshield: Sorry, my bad. I removed the cleanup routine as suggested.

stephane-rochoy-stormshield:
> While I don't dispute the results of your testing, …

I'm just starting to discover this API so don't hesitate to dispute anything ;) 

> …  I don't understand why it works. Shouldn't it be sufficient to run openlog before cap_enter? I don't understand why the program even needs cap_syslog in the first place. IIUC, cap_syslog is only needed if you must open a syslog connection _after_ cap_enter.

I have to admit I didn't even try to move the call to `openlog` somewhere after `init_caps` because there's some calls to `logmsg` in between. IIUC, it would means calling `vsyslog` without `LOG_PERROR` which is not the intended behavior.

asomers:
> I have to admit I didn't even try to move the call to `openlog` somewhere after `init_caps` because there's some calls to `logmsg` in between. IIUC, it would means calling `vsyslog` without `LOG_PERROR` which is not the intended behavior.

Not after, before.  `openlog` ought to be called before `init_caps`, which it is.  However, `openlog` only affects future calls to `syslog` and `vsyslog`, not calls to `cap_syslog` and `cap_vsyslog`.  I'm wondering if logging will work perfectly if all of the `cap_syslog` stuff is removed, since `openlog` is already called early enough.

asomers:
It sounds like we can elminate cap_openlog/cap_syslog and go back to plain openlog/syslog, then.  Could you please do that?

asomers:
> @asomers Sorry, I'm just jumping in. It's not like syslog might want to reopen the file or connect again to `syslogd z(for example when the daemon was restarted)? Or `savecore`always log only to`stderr`?

Since savecore is a short-lived program, running for only a few seconds, I don't think we need to worry about syslogd restarts.  Are there any other reasons why syslog(3) might reconnect?

oshogbo:
Yeah, that's a valid point.
On the other hand, saving the core might take some time, as dumping a couple of GBs can be a time-consuming task.

I don't have a strong opinion on this.

I'm not sure what happens when the log file is rotated.

markjdb:
syslogd handles log rotation transparently to loggers. syslog(3) users don't need to reopen sockets after a log rotation. I can't see any reason not to use the plain syslog() and openlog() functions.

stephane-rochoy-stormshield:
> syslogd handles log rotation transparently to loggers. syslog(3) users don't need to reopen sockets after a log rotation. I can't see any reason not to use the plain syslog() and openlog() functions.

So the question is why do we have `cap_` variants for them exactly?

asomers:
> So the question is why do we have `cap_` variants for them exactly?

I think it was due to an oversight in d7fffd0689f1b18d1a3a589f113e41e53d72b547 .  The author replaced several other services with their capsicumized versions, where that's really needed.  I think he just didn't realize that the non-capsicum syslog would work fine.

markjdb:
> > So the question is why do we have `cap_` variants for them exactly?
> 
> I think it was due to an oversight in [d7fffd0](https://github.com/freebsd/freebsd-src/commit/d7fffd0689f1b18d1a3a589f113e41e53d72b547) . The author replaced several other services with their capsicumized versions, where that's really needed. I think he just didn't realize that the non-capsicum syslog would work fine.

Yes, I think it was probably me being overlay paranoid.

As was already noted, savecore is a short-lived program, it's unlikely that it'll hit the conditions under which syslog(3) needs to reconnect (syslogd restarts, or its logging buffer is full). So, plain openlog()+syslog() should work fine in practice.

OTOH, I suspect the patch in this PR is also fine as-is. Maybe let's merge it and consider removing cap_syslog separately, given that we're just trying to fix a simple bug here? Either approach is ok with me.

stephane-rochoy-stormshield:
So the short-term plan is to add the missing call to `cap_openlog` to fix `savecore`'s behavior regarding `LOG_PERROR`.

But I'm not sure to understand the long-term plan: is it to just drop the use of `cap_openlog` and `cap_vsyslog`? Is looks like dropping the Casper's `system.syslog` service so could someone explain the purpose of this service in the first place?

markjdb:
> So the short-term plan is to add the missing call to `cap_openlog` to fix `savecore`'s behavior regarding `LOG_PERROR`.
> 
> But I'm not sure to understand the long-term plan: is it to just drop the use of `cap_openlog` and `cap_vsyslog`? Is looks like dropping the Casper's `system.syslog` service so could someone explain the purpose of this service in the first place?

It's to handle the rare case where syslog(3) has to reconnect to syslogd's socket. In capability mode, connect() is not allowed, so we use a casper service. This is mostly suitable for long-running programs, where the likelihood of needing to reconnect at some point is high.

I tend to think that syslog(3) could instead just open /var/run (limiting rights to CAP_CONNECTAT, CAP_LOOKUP, CAP_WRITE, ...?) and use connectat(2) when needed. @oshogbo do you see any problem with this alternative? It means that we hold an extra capability for /var/run, but I think that's a small downside.

markjdb:
Merged, thank you.

PR #1547, opened 2024-12-13
newvers: Set explicit git revision length

patmaddox:
The --short flag is configurable. Setting an explicit length supports reproducible builds.


bsdimp (approved):
Love it

emaste (approved):

emaste:
Committed as a1097094c4c5

PR #1548, opened 2024-12-13
build: Sort distributekernel METALOG when using -DNO_ROOT

patmaddox:
The metalog is produced by install -M, which is not sorted. This results in non-deterministic file ordering in kernel.txz. Order the files in kernel.txz to support reproducible builds.


patmaddox:
Closing this, will push out one branch with a patch series

PR #1549, opened 2024-12-13
MFC style.9 to FreeBSD 14

brooksdavis:
This rolls up all the changes to style 9 not yet merged. Style should generally be the same across all branches.

The SCCS and branch predictor commits are both somewhat manual partial merges altered to only touch style.9.


emaste (approved):

bsdimp (approved):
Looks good. The sccs commits have had a lot of annoying consequences.  Tha ks for fixing.

PR #1550, opened 2024-12-13
Reproducible release tarballs

patmaddox:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=283214

picking up the work from https://reviews.freebsd.org/D48010

TODO:

- [x] fix `usr/share/man/mandoc.db` non-determinism

### DIFF
--- a/Makefile.inc1
+++ b/Makefile.inc1
@@ -1441,7 +1441,11 @@ DEBUG_DISTRIBUTIONS=
 DEBUG_DISTRIBUTIONS+=	base ${EXTRA_DISTRIBUTIONS:S,tests,,}
 .endif
 
-MTREE_MAGIC?=	mtree 2.0
+MTREE_MAGIC?=		mtree 2.0
+.if ${MK_REPRODUCIBLE_BUILD} == "yes"
+MTREE_TIME=		time=${SOURCE_DATE_EPOCH}.000000000
+SED_REPLACE_TIME=	-e "s/ time=[[:digit:].]*/ ${MTREE_TIME}/"
+.endif
 
 distributeworld installworld stageworld: _installcheck_world .PHONY
 	mkdir -p ${INSTALLTMP}
@@ -1470,6 +1474,7 @@ distributeworld installworld stageworld: _installcheck_world .PHONY
 .if defined(NO_ROOT)
 	-mkdir -p ${METALOG:H}
 	echo "#${MTREE_MAGIC}" > ${METALOG}
+	echo "/set ${MTREE_TIME}" >> ${METALOG}
 .endif
 .if make(distributeworld)
 .for dist in ${EXTRA_DISTRIBUTIONS}
@@ -1548,7 +1553,7 @@ distributeworld installworld stageworld: _installcheck_world .PHONY
 	@# the relevant mtree line.
 	cd ${DESTDIR}/${DISTDIR}; \
 	    find ./${dist}${path} | ${METALOG_SORT_CMD} -u ${METALOG} - | \
-	    awk 'BEGIN { print "#${MTREE_MAGIC}" } !/ type=/ { file = $$1 } / type=/ { if ($$1 == file) { sub(/^\.\/${dist}/, "."); print } }' > \
+	    awk 'BEGIN { print "#${MTREE_MAGIC}"; print "/set ${MTREE_TIME}" } !/ type=/ { file = $$1 } / type=/ { if ($$1 == file) { sub(/^\.\/${dist}/, "."); print } }' > \
 	    ${DESTDIR}/${DISTDIR}/${dist}${suffix}
 .endfor
 .endfor
@@ -1912,7 +1917,8 @@ distributekernel distributekernel.debug: .PHONY
 	    ${.TARGET:S/distributekernel/install/}
 .if defined(NO_ROOT)
 	echo "#${MTREE_MAGIC}" > ${DESTDIR}/${DISTDIR}/kernel.meta
-	sed -e 's|^./kernel|.|' ${DESTDIR}/${DISTDIR}/kernel.premeta | \
+	echo "/set ${MTREE_TIME}" >> ${DESTDIR}/${DISTDIR}/kernel.meta
+	sed -e 's|^./kernel|.|' ${SED_REPLACE_TIME} ${DESTDIR}/${DISTDIR}/kernel.premeta | \
 	    ${METALOG_SORT_CMD} >> ${DESTDIR}/${DISTDIR}/kernel.meta
 .endif
 .endif
@@ -1928,7 +1934,8 @@ distributekernel distributekernel.debug: .PHONY
 	    ${.TARGET:S/distributekernel/install/}
 .if defined(NO_ROOT)
 	echo "#${MTREE_MAGIC}" > ${DESTDIR}/${DISTDIR}/kernel.${_kernel}.meta
-	sed -e "s|^./kernel.${_kernel}|.|" \
+	echo "/set ${MTREE_TIME}" >> ${DESTDIR}/${DISTDIR}/kernel.${_kernel}.meta
+	sed -e "s|^./kernel.${_kernel}|.|" ${SED_REPLACE_TIME} \
 	    ${DESTDIR}/${DISTDIR}/kernel.${_kernel}.premeta | \
 	    ${METALOG_SORT_CMD} >> ${DESTDIR}/${DISTDIR}/kernel.${_kernel}.meta
 .endif
--- a/Makefile.inc1
+++ b/Makefile.inc1
@@ -462,7 +462,7 @@ SUBDIR+=	${_DIR}
 # by calling 'makedb' in share/man.  This is only relevant for
 # install/distribute so they build the whatis file after every manpage is
 # installed.
-.if make(installworld) || make(install)
+.if make(installworld) || make(install) || make(distribute)
 SUBDIR+=.WAIT
 .endif
 SUBDIR+=etc
--- a/Makefile.inc1
+++ b/Makefile.inc1
@@ -1903,9 +1903,7 @@ distributekernel distributekernel.debug: .PHONY
 	false
 .endif
 	mkdir -p ${DESTDIR}/${DISTDIR}
-.if defined(NO_ROOT)
-	@echo "#${MTREE_MAGIC}" > ${DESTDIR}/${DISTDIR}/kernel.premeta
-.endif
+	rm -f ${DESTDIR}/${DISTDIR}/kernel.premeta
 	${_+_}cd ${KRNLOBJDIR}/${INSTALLKERNEL}; \
 	    ${IMAKEENV} ${IMAKE_INSTALL:S/METALOG/kernel.premeta/} \
 	    ${IMAKE_MTREE} PATH=${TMPPATH:Q} ${MAKE} KERNEL=${INSTKERNNAME} \
@@ -1913,15 +1911,14 @@ distributekernel distributekernel.debug: .PHONY
 	    METALOG=${METALOG:S/METALOG/kernel.premeta/} \
 	    ${.TARGET:S/distributekernel/install/}
 .if defined(NO_ROOT)
-	@sed -e 's|^./kernel|.|' ${DESTDIR}/${DISTDIR}/kernel.premeta > \
-	    ${DESTDIR}/${DISTDIR}/kernel.meta
+	echo "#${MTREE_MAGIC}" > ${DESTDIR}/${DISTDIR}/kernel.meta
+	sed -e 's|^./kernel|.|' ${DESTDIR}/${DISTDIR}/kernel.premeta | \
+	    ${METALOG_SORT_CMD} >> ${DESTDIR}/${DISTDIR}/kernel.meta
 .endif
 .endif
 .if ${BUILDKERNELS:[#]} > 1 && ${NO_INSTALLEXTRAKERNELS} != "yes"
 .for _kernel in ${BUILDKERNELS:[2..-1]}
-.if defined(NO_ROOT)
-	@echo "#${MTREE_MAGIC}" > ${DESTDIR}/${DISTDIR}/kernel.${_kernel}.premeta
-.endif
+	rm -f ${DESTDIR}/${DISTDIR}/kernel.${_kernel}.premeta
 	${_+_}cd ${KRNLOBJDIR}/${_kernel}; \
 	    ${IMAKEENV} ${IMAKE_INSTALL:S/METALOG/kernel.${_kernel}.premeta/} \
 	    ${IMAKE_MTREE} PATH=${TMPPATH:Q} ${MAKE} \
@@ -1930,9 +1927,10 @@ distributekernel distributekernel.debug: .PHONY
 	    METALOG=${METALOG:S/METALOG/kernel.${_kernel}.premeta/} \
 	    ${.TARGET:S/distributekernel/install/}
 .if defined(NO_ROOT)
-	@sed -e "s|^./kernel.${_kernel}|.|" \
-	    ${DESTDIR}/${DISTDIR}/kernel.${_kernel}.premeta > \
-	    ${DESTDIR}/${DISTDIR}/kernel.${_kernel}.meta
+	echo "#${MTREE_MAGIC}" > ${DESTDIR}/${DISTDIR}/kernel.${_kernel}.meta
+	sed -e "s|^./kernel.${_kernel}|.|" \
+	    ${DESTDIR}/${DISTDIR}/kernel.${_kernel}.premeta | \
+	    ${METALOG_SORT_CMD} >> ${DESTDIR}/${DISTDIR}/kernel.${_kernel}.meta
 .endif
 .endfor
 .endif
--- a/share/man/man7/release.7
+++ b/share/man/man7/release.7
@@ -283,6 +283,16 @@ This is intended for use only when
 .Fa /usr/ports
 is expected to exist by alternative means.
 .El
+.Sh REPRODUCIBLE BUILDS
+Different builders can produce bit-identical release tarballs using
+.Va PKG_TIMESTAMP
+and
+.Va REVISION :
+.Bd -literal -offset indent
+git checkout <REVISION>
+make -DWITH_REPRODUCIBLE_BUILD buildworld buildkernel
+make -C release -DWITH_REPRODUCIBLE_BUILD PKG_TIMESTAMP=<TIMESTAMP> packagesystem
+.Ed
 .Sh EMBEDDED BUILDS
 The following
 .Fa release.conf


Makefile.inc1
```diff
@@ -1904,25 +1904,24 @@ distributekernel distributekernel.debug: .PHONY
...
+	@sed -e 's|^./kernel|.|' ${DESTDIR}/${DISTDIR}/kernel.premeta | \
+	    sort >> ${DESTDIR}/${DISTDIR}/kernel.meta
 .endif
 .endif
 .if ${BUILDKERNELS:[#]} > 1 && ${NO_INSTALLEXTRAKERNELS} != "yes"
 .for _kernel in ${BUILDKERNELS:[2..-1]}
 .if defined(NO_ROOT)
-	@echo "#${MTREE_MAGIC}" > ${DESTDIR}/${DISTDIR}/kernel.${_kernel}.premeta
 .endif
+	rm -f ${DESTDIR}/${DISTDIR}/kernel.${_kernel}.meta
```
brooksdavis, line 1930: I think this is unnecessary as it's truncated below?

Makefile.inc1
```diff
@@ -1904,25 +1904,24 @@ distributekernel distributekernel.debug: .PHONY
...
+	@echo "#${MTREE_MAGIC}" > ${DESTDIR}/${DISTDIR}/kernel.meta
+	@sed -e 's|^./kernel|.|' ${DESTDIR}/${DISTDIR}/kernel.premeta | \
+	    sort >> ${DESTDIR}/${DISTDIR}/kernel.meta
 .endif
 .endif
 .if ${BUILDKERNELS:[#]} > 1 && ${NO_INSTALLEXTRAKERNELS} != "yes"
 .for _kernel in ${BUILDKERNELS:[2..-1]}
 .if defined(NO_ROOT)
-	@echo "#${MTREE_MAGIC}" > ${DESTDIR}/${DISTDIR}/kernel.${_kernel}.premeta
 .endif
```
brooksdavis, line 1929: This `.if` is now empty

Makefile.inc1
```diff
@@ -1904,25 +1904,24 @@ distributekernel distributekernel.debug: .PHONY
...
 	    ${IMAKE_MTREE} PATH=${TMPPATH:Q} ${MAKE} KERNEL=${INSTKERNNAME} \
 	    DISTBASE=/kernel DESTDIR=${INSTALL_DDIR}/kernel \
 	    METALOG=${METALOG:S/METALOG/kernel.premeta/} \
 	    ${.TARGET:S/distributekernel/install/}
 .if defined(NO_ROOT)
-	@sed -e 's|^./kernel|.|' ${DESTDIR}/${DISTDIR}/kernel.premeta > \
-	    ${DESTDIR}/${DISTDIR}/kernel.meta
+	@echo "#${MTREE_MAGIC}" > ${DESTDIR}/${DISTDIR}/kernel.meta
+	@sed -e 's|^./kernel|.|' ${DESTDIR}/${DISTDIR}/kernel.premeta | \
+	    sort >> ${DESTDIR}/${DISTDIR}/kernel.meta
```
brooksdavis, line 1923: I think you'll want to use `${METALOG_SORT_CMD}` instead of a bare `sort`.

release/Makefile
```diff
@@ -140,7 +140,7 @@ base.txz:
 	    -L /dev/null -N
 	echo "./var/db/etcupdate type=dir uname=root gname=wheel mode=0755" >> ${.OBJDIR}/${DISTDIR}/base.meta
 	sed -n 's,^\.,./var/db/etcupdate/current,p' ${.OBJDIR}/${DISTDIR}/base/var/db/etcupdate/current/METALOG \
-	    >> ${.OBJDIR}/${DISTDIR}/base.meta
+	    | sort >> ${.OBJDIR}/${DISTDIR}/base.meta
```
brooksdavis, line 143: The pipe should be on the previous line. I'd probably rewrap it with a break after the substitution pattern.

brooksdavis (commented):
Looks pretty good. A few comments.

The existing `@` prefixes are IMO bugs so you can drop them. I'm not sure why there are there when the surrounding code doesn't use them.

jrtc27:
I can't add myself as a reviewer but would like to review this properly in the coming days

Makefile.inc1
patmaddox: Thanks, that was supposed to be `.premeta`. Fixed

jrtc27:
> build: Default METALOG ownership to root/wheel
>
> Release calls etcupdate.sh, which calls certctl. When building as an
> unprivileged user, this resulted in the certs with uid of the building
> user.
> 
> PR: 283214
> 
> Signed-off-by: Pat Maddox <pat@patmaddox.com>

This isn't the right fix. The correct thing to do is to fix certctl, not paper over it and any other potential buggy build steps.

jrtc27:
> build: Explicitly select debug files when distributing with -DNO_ROOT
> Most of the debug files live in usr/lib/debug, but 
> usr/tests/sys/vm/stack/.debug/libsoxstack.so.debug
> does not and was incorrectly being included in base.txz.
> 
> PR: 283214
> 
> Signed-off-by: Pat Maddox <pat@patmaddox.com>

I feel like that's more of a sign the library is being installed to the wrong place and should be in /usr/tests/lib/sys/... or similar.

patmaddox:
@jrtc27 you're right about the owner, and it's not necessary - it was a false positive from some build cruft.

I'll see if there's a better spot for the test - but maybe someone more familiar with it knows off-hand?

@emaste @bsdimp I've submitted this single PR with a series of patches, but they're mostly independent. I didn't know if I should do one PR and let you cherry pick acceptable commits, or open separate PRs for each one. Let me know if I should split it up - or any general guidance you have, as this is my first time contributing to src.

patmaddox:
on amd64

1. `git checkout ac493dbe5ca707dba0ccc8c0f778e087f85ac32b`
2. (as root) `make -s -j$(sysctl -n hw.ncpu) -DWITH_REPRODUCIBLE_BUILD buildworld buildkernel && make -s -C release -j$(sysctl -n hw.ncpu) -DNOSRC -DNOPORTS -DWITH_REPRODUCIBLE_BUILD PKG_TIMESTAMP=1733729386 packagesystem`

should produce kernel `FreeBSD 15.0-CURRENT n274203-ac493dbe5ca7 GENERIC` and the following MANIFEST:

```
base-dbg.txz    6adcb45f8e91fa84ae0888682df68407a495cf439483f0156cd052618d757493        1104    base_dbg        "Base system (Debugging)"       off
base.txz        ce100ab1671dc6bd754b92743e0bde96a39628a0b607955263c1cdb60f988735        29557   base    "Base system (MANDATORY)"       on
kernel-dbg.txz  60748997e8190fdeb230fa6e3dd9579c5fcf75cf515203d4c730ce4d9777436f        878     kernel_dbg      "Kernel (Debugging)"    on
kernel.txz      ef8d6e947d98f96c5caad22b7ca9d113cd63cec3e87f12649aa9c19eedbc6a4a        881     kernel  "Kernel (MANDATORY)"    on
lib32-dbg.txz   e7d18d478f38aa56c2550f9e26af6b9b1d972cda8b90a55fd78f7ea359ebaadf        243     lib32_dbg       "32-bit compatibility libraries (Debugging)"    off
lib32.txz       758203d29602a9aa37037850d8633ebf1b6ad93afdc0ba1bd8b048b9a781b30e        730     lib32   "32-bit compatibility libraries"        on
tests.txz       f1e3040026122424d58473f72a26f99f6777b16e5b4c1d6f639a44d31fd577ff        7350    tests   "Test suite"    off
```

patmaddox:
Added certctl fix (thanks @jrtc27). Building and packaging as a non-privileged user reproduces the root build.

526efd6c1d79ad4fe3103bd8d97e722caa6b9563 with `PKG_TIMESTAMP=1733729386` should produce `FreeBSD 15.0-CURRENT n274204-526efd6c1d79 GENERIC` with MANIFEST:

```
base-dbg.txz    6adcb45f8e91fa84ae0888682df68407a495cf439483f0156cd052618d757493        1104    base_dbg        "Base system (Debugging)"       off
base.txz        e3e2809ec6399a069cd25c7817c5a3d60d68f73acd4907f568db91d90fc8f6e2        29557   base    "Base system (MANDATORY)"       on
kernel-dbg.txz  9824cb629b5e991e171e08e01bf20bcebda9a9dd3c84e0e8fb9051bccfafe723        878     kernel_dbg      "Kernel (Debugging)"    on
kernel.txz      0453f33e8a14e7762ac0a20dd3fefff3d0ed058e4d0325cc3ef531ef572b02d1        881     kernel  "Kernel (MANDATORY)"    on
lib32-dbg.txz   e7d18d478f38aa56c2550f9e26af6b9b1d972cda8b90a55fd78f7ea359ebaadf        243     lib32_dbg       "32-bit compatibility libraries (Debugging)"    off
lib32.txz       758203d29602a9aa37037850d8633ebf1b6ad93afdc0ba1bd8b048b9a781b30e        730     lib32   "32-bit compatibility libraries"        on
tests.txz       f1e3040026122424d58473f72a26f99f6777b16e5b4c1d6f639a44d31fd577ff        7350    tests   "Test suite"    off
```

jrtc27:
> Added certctl fix (thanks @jrtc27). Building and packaging as a non-privileged user reproduces the root build.
> 
> [526efd6](https://github.com/freebsd/freebsd-src/commit/526efd6c1d79ad4fe3103bd8d97e722caa6b9563) with `PKG_TIMESTAMP=1733729386` should produce `FreeBSD 15.0-CURRENT n274204-526efd6c1d79 GENERIC` with MANIFEST:
> 
> ```
> base-dbg.txz    6adcb45f8e91fa84ae0888682df68407a495cf439483f0156cd052618d757493        1104    base_dbg        "Base system (Debugging)"       off
> base.txz        e3e2809ec6399a069cd25c7817c5a3d60d68f73acd4907f568db91d90fc8f6e2        29557   base    "Base system (MANDATORY)"       on
> kernel-dbg.txz  9824cb629b5e991e171e08e01bf20bcebda9a9dd3c84e0e8fb9051bccfafe723        878     kernel_dbg      "Kernel (Debugging)"    on
> kernel.txz      0453f33e8a14e7762ac0a20dd3fefff3d0ed058e4d0325cc3ef531ef572b02d1        881     kernel  "Kernel (MANDATORY)"    on
> lib32-dbg.txz   e7d18d478f38aa56c2550f9e26af6b9b1d972cda8b90a55fd78f7ea359ebaadf        243     lib32_dbg       "32-bit compatibility libraries (Debugging)"    off
> lib32.txz       758203d29602a9aa37037850d8633ebf1b6ad93afdc0ba1bd8b048b9a781b30e        730     lib32   "32-bit compatibility libraries"        on
> tests.txz       f1e3040026122424d58473f72a26f99f6777b16e5b4c1d6f639a44d31fd577ff        7350    tests   "Test suite"    off
> ```

This will break building on Linux though as there's no wheel group there. This is what DB_FROM_SRC is for, certctl needs to pass -N /path/to/etc for wheel to have meaning.

patmaddox:
> This will break building on Linux though as there's no wheel group there. This is what DB_FROM_SRC is for, certctl needs to pass -N /path/to/etc for wheel to have meaning.

On FreeBSD, `install -U` simply writes the given values to the METALOG:

```sh
#!/bin/sh
set -eu

rm -rf METALOG foo bar installed

echo "this is foo" > foo
echo "this is bar" > bar

mkdir installed
install -U -M METALOG -o missing-user -g missing-group foo bar installed/
cat METALOG
```

This produces:

```
./installed/foo type=file uname=missing-user gname=missing-group mode=0755 size=12
./installed/bar type=file uname=missing-user gname=missing-group mode=0755 size=12
```

Are you saying that on Linux, `install` validates the user and group name before adding it to METALOG, so the above example would error?

patmaddox:
> Are you saying that on Linux, `install` validates the user and group name before adding it to METALOG, so the above example would error?

I just tested on debian, and it doesn't even have `-U` or `-M`.

So... check if `install` supports those flags / check if we're on FreeBSD, and only add them if it does? Because it already doesn't do the right thing on Linux (I assume). Thus the behavior would be improved on FreeBSD, and remain unchanged on Linux for the time being.

jrtc27:
> > This will break building on Linux though as there's no wheel group there. This is what DB_FROM_SRC is for, certctl needs to pass -N /path/to/etc for wheel to have meaning.
> 
> On FreeBSD, `install -U` simply writes the given values to the METALOG:
> 
> ```shell
> #!/bin/sh
> set -eu
> 
> rm -rf METALOG foo bar installed
> 
> echo "this is foo" > foo
> echo "this is bar" > bar
> 
> mkdir installed
> install -U -M METALOG -o missing-user -g missing-group foo bar installed/
> cat METALOG
> ```
> 
> This produces:
> 
> ```
> ./installed/foo type=file uname=missing-user gname=missing-group mode=0755 size=12
> ./installed/bar type=file uname=missing-user gname=missing-group mode=0755 size=12
> ```
> 
> Are you saying that on Linux, `install` validates the user and group name before adding it to METALOG, so the above example would error?

Hm, it seems you're right that our install(1) (which will still be the one used on non-FreeBSD during the build) doesn't validate user/group names for -U. It's not documented though, and seems to be a bad idea to me, we should be validating them at install time, not at makefs time nor, in the case of distribution tarballs, untar, i.e. bsdinstall, time (since our tarballs use symbolic user/group names, and bsdtar doesn't get a -N equivalent yet packageworld works on Linux, so it can't possibly be validating it).

patmaddox:
> install(1) (which will still be the one used on non-FreeBSD during the build)

That's good to know - so build behavior is equivalent on both OSes.

> we should be validating them at install time, not at makefs time nor, in the case of distribution tarballs, untar

Maybe... but I think there would need to be a flag to let you validate or not. It's plausible for a builder to configure a METALOG for users it doesn't have. In fact that's exactly something I'm going to do now that I know about `install` / `metalog` / `tar` combo. The builder tars up files with a METALOG assigning ownership to `www`, `postgres`, `acme`, etc - but the builder doesn't have any of those users defined locally. It also won't have access to a passwd db that defines them.

That may be a useful enhancement to `install(1)`, but seems to me to be out of scope for this particular PR.

jrtc27:
> > install(1) (which will still be the one used on non-FreeBSD during the build)
> 
> That's good to know - so build behavior is equivalent on both OSes.
> 
> > we should be validating them at install time, not at makefs time nor, in the case of distribution tarballs, untar
> 
> Maybe... but I think there would need to be a flag to let you validate or not. It's plausible for a builder to configure a METALOG for users it doesn't have. In fact that's exactly something I'm going to do now that I know about `install` / `metalog` / `tar` combo. The builder tars up files with a METALOG assigning ownership to `www`, `postgres`, `acme`, etc - but the builder doesn't have any of those users defined locally. It also won't have access to a passwd db that defines them.
> 
> That may be a useful enhancement to `install(1)`, but seems to me to be out of scope for this particular PR.

My point is, throughout the build system, we, at least intend to, pass -N /path/to/etc to install(1), mtree(8) and makefs(8) whenever DB_FROM_SRC is set (in some cases even unconditionally), which is implied to NO_ROOT. So, just as we have `INSTALLFLAGS+=  -N ${.CURDIR}/etc` in Makefile.inc1, we should ensure that certctl does the same during the build. It happens to not matter today, but that's an undocumented detail we're relying on, and is inconsistent with the rest of the build.

emaste:
> @emaste @bsdimp I've submitted this single PR with a series of patches, but they're mostly independent.

This is fine - I'm happy to grab individual commits out as I or others review them.

emaste:
CC @brooksdavis and @kevans91 for certctl - this originated in 48e9fb855b021c07642b5393d9eff5704adfd98e.

I think the patch in https://github.com/freebsd/freebsd-src/pull/1550/commits/526efd6c1d79ad4fe3103bd8d97e722caa6b9563 is a good step in the right direction, even if we should also have support for `-N` and fix PR283355.

emaste:
Although that said I wonder if the `-o` and `-u` should be on each install invocation instead.

Makefile.inc1
```diff
@@ -1904,25 +1904,22 @@ distributekernel distributekernel.debug: .PHONY
...
 	    ${IMAKEENV} ${IMAKE_INSTALL:S/METALOG/kernel.premeta/} \
 	    ${IMAKE_MTREE} PATH=${TMPPATH:Q} ${MAKE} KERNEL=${INSTKERNNAME} \
 	    DISTBASE=/kernel DESTDIR=${INSTALL_DDIR}/kernel \
 	    METALOG=${METALOG:S/METALOG/kernel.premeta/} \
 	    ${.TARGET:S/distributekernel/install/}
 .if defined(NO_ROOT)
-	@sed -e 's|^./kernel|.|' ${DESTDIR}/${DISTDIR}/kernel.premeta > \
-	    ${DESTDIR}/${DISTDIR}/kernel.meta
+	echo "#${MTREE_MAGIC}" > ${DESTDIR}/${DISTDIR}/kernel.meta
+	sed -e 's|^./kernel|.|' ${DESTDIR}/${DISTDIR}/kernel.premeta | \
```
emaste: does the unsortedness happen only with -j? I'd be a bit surprised by nondeterminism without -j.

patmaddox: I'm not sure - I've never had the patience to run without `-j`.

One potential source of non-determinism is when file lists are generated rather than declared. For example, some of the items returned by `grep -R '=.*find' --include 'Makefile*' sys` could potentially be non-determinstic. But I can't say for sure.

emaste: It looks like those are all in sys/contrib (and so not actually used in our build), but indeed anything that returns files in filesystem order could be an issue.

I expect `-j` will result in nondeterminism and so the sorting is necessary, mainly just wondering if this is the only source of nondeterminism here.

release/Makefile
```diff
@@ -139,8 +139,9 @@ base.txz:
 	    -s ${WORLDDIR} -d "${.OBJDIR}/${DISTDIR}/base/var/db/etcupdate" \
 	    -L /dev/null -N
 	echo "./var/db/etcupdate type=dir uname=root gname=wheel mode=0755" >> ${.OBJDIR}/${DISTDIR}/base.meta
-	sed -n 's,^\.,./var/db/etcupdate/current,p' ${.OBJDIR}/${DISTDIR}/base/var/db/etcupdate/current/METALOG \
-	    >> ${.OBJDIR}/${DISTDIR}/base.meta
+	sed -n 's,^\.,./var/db/etcupdate/current,p' \
+	    ${.OBJDIR}/${DISTDIR}/base/var/db/etcupdate/current/METALOG | \
+	    ${METALOG_SORT_CMD} >> ${.OBJDIR}/${DISTDIR}/base.meta
```
emaste, line 144: Wait, where is `METALOG_SORT_CMD` defined for relelase/* Makefiles?

patmaddox: oof... it is not.

And the reason this "worked" for removing the non-determinism is because `echo "foo" | > file` is valid shell, and results in empty `file`. So the etc entries aren't being sorted... they're just not being added at all.

patmaddox: fixed in 9da5a0f1d057 and confirmed it packages the same on three systems

brooksdavis: You still have to do what METALOG_SORT_CMD does or this won't be sound on systems with LC_COLLATE!=C.

patmaddox: Updated in 0d3b7348

patmaddox:
> [0d3b734](https://github.com/freebsd/freebsd-src/commit/0d3b7348ac094b1be1fe385e02337666cf6192d2) looks good except for the chicken and metalog files, will commit without those :)

patmaddox:
> just as we have `INSTALLFLAGS+= -N ${.CURDIR}/etc` in Makefile.inc1

Updated in e48266727c2e9bf5b64ecd9b44050ba4363ece1c

#1552 adds the actual validation. I've tested them together.

emaste:
CC @allanjude @kevans91 if you'd like to look at the certctl change in https://github.com/freebsd/freebsd-src/pull/1550/commits/e48266727c2e9bf5b64ecd9b44050ba4363ece1c

allanjude (approved):
Reviewed by: allanjude

emaste:
@allanjude did you review all of the changes or just certctl? (AFAIK there's no way to indicate review of a partial set of commits in a pull request.)

Makefile.inc1
```diff
@@ -1903,25 +1903,22 @@ distributekernel distributekernel.debug: .PHONY
...
 	${_+_}cd ${KRNLOBJDIR}/${INSTALLKERNEL}; \
 	    ${IMAKEENV} ${IMAKE_INSTALL:S/METALOG/kernel.premeta/} \
 	    ${IMAKE_MTREE} PATH=${TMPPATH:Q} ${MAKE} KERNEL=${INSTKERNNAME} \
 	    DISTBASE=/kernel DESTDIR=${INSTALL_DDIR}/kernel \
 	    METALOG=${METALOG:S/METALOG/kernel.premeta/} \
 	    ${.TARGET:S/distributekernel/install/}
 .if defined(NO_ROOT)
-	@sed -e 's|^./kernel|.|' ${DESTDIR}/${DISTDIR}/kernel.premeta > \
-	    ${DESTDIR}/${DISTDIR}/kernel.meta
+	echo "#${MTREE_MAGIC}" > ${DESTDIR}/${DISTDIR}/kernel.meta
```
emaste, line 1919: We want to keep the `@`s presumably

Makefile.inc1
```diff
@@ -462,7 +462,7 @@ SUBDIR+=	${_DIR}
 # by calling 'makedb' in share/man.  This is only relevant for
 # install/distribute so they build the whatis file after every manpage is
 # installed.
-.if make(installworld) || make(install)
+.if make(installworld) || make(install) || make(distribute)
```
emaste, line 465: Should this be `make(distributeworld)`? I'm not sure, but I don't fully understand the `distribute` pseduo-target.

Makefile.inc1
brooksdavis: IMO the `@`s are bugs. These aren't steps with should be hiding from the caller.

emaste:
Sorting pushed in 58610d1e0fbbd1a49927559ee3970d8e4594cc86.

I think `build: Set METALOG time when using -DWITH_REPRODUCIBLE_BUILD` is fine, although we ought to give install the ability to directly emit a reproducible `time=` entry. (I'd do that as a subsequent change.)

` build: Wait to build mandoc.db when making distributeworld ` has an open question.

I've got some comments for the doc update in `release: Document reproducible release tarballs` but that change can wait until the others are solved anyway

share/man/man7/release.7
```diff
@@ -283,6 +283,16 @@ This is intended for use only when
 .Fa /usr/ports
 is expected to exist by alternative means.
 .El
+.Sh REPRODUCIBLE BUILDS
+Different builders can produce bit-identical release tarballs using
+.Va PKG_TIMESTAMP
+and
+.Va REVISION :
```
emaste, line 290: We probably want to differentiate `PKG_TIMESTAMP` from `REVISION` -- the former is an actual (literal) env variable used by Makefile*, but the latter is a placeholder in the example command below. I'm not sure off hand what the markup for that should be.

I wonder if we should have newvers.sh collect the commit time (e.g. from `git show -s --format=%ct`) and have the default SOURCE_DATE_EPOCH derived from that.

PR #1552, opened 2024-12-17
install: validate -o, -g, and -f values when run with -U

patmaddox:
When run as root, install will error if the values given to -o, -g, or -f are invalid. Prior to this change, running with -U (unprivileged) silently accepted any values given to these flags.

Test that the values are correctly recorded to METALOG when called with -M.

[PR: 283355](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=283355)

### DIFF
--- a/usr.bin/xinstall/tests/install_test.sh
+++ b/usr.bin/xinstall/tests/install_test.sh
@@ -497,6 +497,30 @@ set_owner_group_mode_unpriv_body() {
 	atf_check_equal "$u:$g:10$cM" "$(stat -f"%u:%g:%p" testc)"
 }
 
+atf_test_case metalog_unpriv
+metalog_unpriv_head() {
+	atf_set "require.user" "unprivileged"
+}
+metalog_unpriv_body() {
+	printf "test" >testf
+
+	atf_check install -U -M METALOG -o root -g wheel -f arch testf testc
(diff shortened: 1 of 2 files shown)


usr.bin/xinstall/tests/install_test.sh
```diff
@@ -497,6 +497,30 @@ set_owner_group_mode_unpriv_body() {
...
+	atf_set "require.user" "unprivileged"
+}
+metalog_unpriv_body() {
+	printf "test" >testf
+
+	atf_check install -U -M METALOG -o root -g wheel -f arch testf testc
+	atf_check grep -q "testc .*uname=root gname=wheel .*flags=arch" METALOG
+
+	rm -f METALOG testc
+	atf_check -s exit:1 -e match:"unknown user no-such-user" install -U -o no-such-user testf testc
```
jlduran, line 511: I thought this was a feature, especially if used together with `-M`. That is, `-U` and `-o` without `-M` should be a NOP?

bsdimp:
Why close this? We're a little behind in landing, to be true, but we'll get to it soonish...

patmaddox:
I'm not really using GitHub anymore, a cleanup script deleted the branch and I guess that closes the PR.

PR #1554, opened 2024-12-17
Bug 282306: Update spelling of "non-zero" to "nonzero" for C23 compliance

kfv:
C23 ([ISO/IEC 9899:2024](https://www.iso.org/standard/82075.html)) standard specifies the spelling "nonzero" instead of "non-zero." This PR updates all occurrences of "non-zero" in the codebase (excluding man pages originating outside of FreeBSD itself) to align with the standard.

**PR**: [282306](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=282306)


concussious (approved):
Additional context: https://gcc.gnu.org/legacy-ml/gcc/2001-10/msg00610.html

"Society of Industrial and Applied Mathematics recommends that "non" as a prefix not be hyphenated".

I read over all of these and made sure they looked reasonable.

If this is good to you, please ammend your commit message with `reviewed by:      ziaee`. Thank you, this is a very desirable change, to me!

kfv:
Done.

concussious:
Sorry for my oversight, it should have `Reported by:     kargl` in the commit log too.

kfv:
Done.

concussious (approved):

asomers (approved):

asomers:
@concussious you have commit access, right?  Or do you need somebody else to commit this?

bsdimp:
Im pretty sure most of these are an incorrect application of this spelling

concussious:
@asomers If my mentors @sergio-carlavilla and/or @mhorne consent, then yes I can. However, I don't understand if @bsdimp is saying that we could but maybe shouldn't do this.

bsdimp:
I'm saying that the style I've seen is that non-zero is hyphenated only when it modifies a noun, but not when used as a noun or on it's own. That means about some significant number of the changes are wrong by that style.

However, other places I see other advice. So I'd punt the question to doceng@ to make a ruling on which style to use.

C23 uses nonzero consistently. POSIX.1-2024 uses both non-zero and nonzero.

bsdimp:
I'm going to just close this. I think we should do any such sweeps at a smaller scale per iteration so we can get it right.

PR #1555, opened 2024-12-18
make check: Eliminate "operation not permitted" errors

patmaddox:
aftercheck calls clean. ${DESTDIR}/var/empty has schg flag set. Prior to this change, every run of `make check` ended with:

rm: checkdir/var/empty: Operation not permitted
rm: checkdir/var: Directory not empty
rm: checkdir: Directory not empty
*** Error code 1 (ignored)

Remove the flags from DESTDIR before running clean, to bypass the noisy output.

### DIFF
--- a/share/mk/suite.test.mk
+++ b/share/mk/suite.test.mk
@@ -118,8 +118,7 @@ beforecheck:
 #       times. "aftercheck" won't be run if "make check" fails, is interrupted,
 #       etc.
 aftercheck:
-	@cd ${.CURDIR} && ${MAKE} clean
 	@test ! -e ${DESTDIR} || chflags -R 0 "${DESTDIR}"
-	@rm -Rf "${DESTDIR}"
+	@cd ${.CURDIR} && ${MAKE} clean
 
 .endif


share/mk/suite.test.mk
```diff
@@ -118,8 +118,7 @@ beforecheck:
 #       times. "aftercheck" won't be run if "make check" fails, is interrupted,
 #       etc.
```
patmaddox, line 119: It's not clear to me if the "this is intentional" line means that it's intended to set flags after `clean`, and letting these errors through. That seems strange to me.

I addressed this because when running the tests, it was unclear to me if something was wrong with my setup. Error messages shouldn't come up as a part of a normal test run, unless they're expected.

Also: I don't know if this is the sort of thing I should make a Problem Report first for? So a Bugzilla PR "make check gives noisy operation not permitted errors" so there's a place to track / discuss it, and then a github PR to fix it?

Seemed small enough to me that I could just make a GH PR.

brooksdavis: The commit message in 38f8fddf054fd has some backstory. I've added @ngie-eign as a reviewer.

patmaddox:
I'm not liking the github workflow, so I'm closing this. I made a PR and a diff.

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=286025
https://reviews.freebsd.org/D49775

bsdimp:
It's up to you, but those input sources are even slower to the project. Also, duplicates waste develope time...

PR #1556, opened 2024-12-18
Fix some locking violations in fusefs

asomers:
Fix several related bugs in fusefs, all relating to activity that ought to require an exclusive vnode lock but was being performed with merely a shared lock.

* Updating a fuse file's atime on read requires an exclusive lock.  This could result in torn writes to the timestamp structure.
* Attempting to clear a file's attribute cache in response to the server returning an error during FUSE_READ requires an exclusive lock.  This could trigger a panic when DEBUG_VFS_LOCKS is enabled.
* fuse_vnop_rename was accessing the source vnode's cached attributes without a vnode lock.  That might result in an incorrect access determination, if it raced with chmod.

Fix these problems by being much more defensive about checking vnode locks within fusefs, upgrading shared locks as needed, and disallowing shared reads when mounted with atime.

### DIFF
--- a/sys/fs/fuse/fuse_node.c
+++ b/sys/fs/fuse/fuse_node.c
@@ -367,7 +367,7 @@ fuse_vnode_savesize(struct vnode *vp, struct ucred *cred, pid_t pid)
 	struct fuse_setattr_in *fsai;
 	int err = 0;
 
-	ASSERT_VOP_ELOCKED(vp, "fuse_io_extend");
+	ASSERT_VOP_ELOCKED(vp, __func__);
 
 	if (fuse_isdeadfs(vp)) {
 		return EBADF;
@@ -423,7 +423,7 @@ fuse_vnode_setsize(struct vnode *vp, off_t newsize, bool from_server)
 	struct buf *bp = NULL;
 	int err = 0;
 
-	ASSERT_VOP_ELOCKED(vp, "fuse_vnode_setsize");
+	ASSERT_VOP_ELOCKED(vp, __func__); /* For fvdat->cached_attrs */
 
 	iosize = fuse_iosize(vp);
 	oldsize = fvdat->cached_attrs.va_size;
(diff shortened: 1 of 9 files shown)


sys/fs/fuse/fuse_node.c
```diff
@@ -522,14 +528,17 @@ fuse_vnode_update(struct vnode *vp, int flags)
 	if (mp->mnt_flag & MNT_NOATIME)
 		flags &= ~FN_ATIMECHANGE;
 
+	if (flags) {
+		ASSERT_VOP_ELOCKED(vp, __func__); /* For fvdat->cached_attrs */
```
markjdb, line 532: Some other filesystems use the vnode interlock to serialize atime updates. Why can't that be done here? Disabling shared locking entirely seems like a heavy hammer.

asomers: Probably, but I'm hazy on how the interlock works.  None of the man pages describe it.  How might I educate myself?

markjdb: For most purposes it's just a mutex(9) embedded in the vnode that can be used as an alternative to the vnode lock, especially for short code sections.

Within individual filesystems it's mostly up to the implementor to decide how to use it. There's a comment above the definition of struct vnode which explains how individual fields are locked; that's about all I'm aware of as far as documentation goes.

markjdb: More generically there is an interlock associated with lockmgr(9) locks, and the vnode interlock serves that purpose as well. Though, I can't find a general description of the "theory of operation" behind it.

asomers: I'm trying that method, @markjdb .  However, I've also found another bug.  So I'm going to close this PR and split it up into 2-3 separate ones.

asomers:
FTR this bug was finally fixed by 7e68af7ce2c1b892954df415774fe59fd2f1b62f .

PR #1558, opened 2024-12-19
acpi_gpiobus: assume GPIO_PIN_INPUT for interrupt pins

VexedUXR:
Most drivers do nothing when neither GPIO_PIN_INPUT or GPIO_PIN_OUTPUT are set. However, some return `EINVAL` while others (bytgpio is the only example I could find of this) give bad flags to the GPIO controller.

In the case of bytgpio, it causes both the input and output of the GPIO pin to be disabled which the controller doesn't seem to like as it causes the entire machine to lose power shortly after.

To fix this, simply assume GPIO_PIN_INPUT for interrupt pins.

I'll CC everyone from the original revision where this change was suggested (see 0ffd7d4d15d468c60ef274f8368a5722e3f7f6a1)

CC @zxombie @jrtc27 @cperciva

### DIFF
--- a/sys/dev/gpio/acpi_gpiobus.c
+++ b/sys/dev/gpio/acpi_gpiobus.c
@@ -77,6 +77,7 @@ acpi_gpiobus_convflags(ACPI_RESOURCE_GPIO *gpio_res)
 			break;
 		}
 
+		flags |= GPIO_PIN_INPUT;
 #ifdef NOT_YET
 		/* This is not currently implemented. */
 		if (gpio_res->Shareable == ACPI_SHARED)


bsdimp (approved):
Seems ok

jrtc27:
Well what does Linux do for dodgy ACPI tables? We should follow whatever they do if we want the best chance of working.

VexedUXR:
They seem to be doing the same: https://github.com/torvalds/linux/blob/eabcdba/drivers/gpio/gpiolib-acpi.c#L292

wulf7:
Committed [1]. Thanks!

[1] https://cgit.freebsd.org/src/commit/?id=6aab381a498f1a43ba372f4ae9f57ef88b4767b5

VexedUXR:
Thank you.

PR #1560, opened 2024-12-22
Add __aeabi_ symbols to linker script

VoxSciurorum:
This fixes bug [bug 271087](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=271087).


clausecker:
It would be great to finally have this addressed.

PR #1564, opened 2025-01-03
open.2: update description for O_PATH

CismonX:
- Add `fstatfs()`, `fchdir()`, `fchroot()`, `extattr_*_fd()`, `cap_*_get()`, `cap_*_limit()` to the list of syscalls that can take an `O_PATH` fd.
- Remove `readlinkat()` from the list, since it is already discussed in the first few lines of the paragraph.  It was originally added to the list when `readlinkat()` adds support for non-dir fd with an empty relative path (as if with `AT_EMPTY_PATH`), however, such use case is also discussed in the next paragraph.
- Add `funlinkat()` to the list, since it accepts an extra fd (of the file to be unlinked), which is worth extra mentioning.
- Fix a syntax issue which causes a bogus space to be rendered before a closing parentheses.

### DIFF
--- a/lib/libsys/open.2
+++ b/lib/libsys/open.2
@@ -393,7 +393,12 @@ but advisory locking is not allowed
 .It Xr dup 2
 .It Xr close 2
 .It Xr fstat 2
+.It Xr fstatfs 2
+.It Xr fchdir 2
+.It Xr fchroot 2
 .It Xr fexecve 2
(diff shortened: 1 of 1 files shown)


markjdb (approved):
Looks good to me.

bsdjhb (approved):

markjdb:
Merged, thank you.

PR #1565, opened 2025-01-03
dot.shrc: words to words, should be singular

grahamperrin:
Jump from one word to another. Not words to words.

### DIFF
--- a/bin/sh/dot.shrc
+++ b/bin/sh/dot.shrc
@@ -34,9 +34,11 @@
 bind ^[[A ed-search-prev-history
 bind ^[[B ed-search-next-history
 
-# # ctrl+arrow allow to jump from words to words
+# # Ctrl+right arrow: go to the next word
+# # Ctrl+left arrow: go to the previous word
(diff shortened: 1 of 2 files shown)


grahamperrin:
Suggested reviewer: @bapt (the original author). 2021:

- https://github.com/freebsd/freebsd-src/commit/ef0d94a3d34c880bd9f86cd842ee01b6075bc1d8
- https://github.com/freebsd/freebsd-src/commit/5fe97373486619373e3eeecb25582b5c937a26c5

concussious:
These files have been and will be many people's first impressions of configuration files. We should carefully consider the style of everything in this directory and make it consistent.

bsdimp (approved):

bsdimp:
While I agree we should be more thorough, this is a good improvement, so I'll land it.
When someone else comes along with more wide-ranging changes, we can evaluate those then.

bsdimp:

grahamperrin:
Thanks

PR #1566, opened 2025-01-05
release: install multiple kernels if specified

asomers:
If release.conf sets KERNEL to a space-separate list, release.sh will now install all of them into the release media, building separate .txz archives for each.
    
Sponsored by:   ConnectWise


emaste:
Eliminating the double negative LGTM, will look at the other change when I have a moment

asomers:
@markjdb this is what I was talking about in the video chat today.  This PR gets both kernels onto the install media.  One wart is that they also both get installed into /boot, bloating the CD image.  IMHO that shouldn't be a barrier to merge, because the 2nd kernel is disabled by default.
Any freebsd-update changes will come separately.

brooksdavis:
Hmm, on CheriBSD we've added individual kernels to release/scripts/make-manifest.sh so they are in the distribution menu in the installer (this is decidedly not generalized). I think that's somewhat related, but not quite the same thing. I don't think we had to make any changes to Makefile.inc1 for this, but these change might well be simplifying?

CC @jrtc27

asomers:
@brooksdavis the changes that I made to Makefile.inc1 are entirely about renaming the variable to eliminate a double negative.  That change is segregated to the first commit.  The second is the actual functional change.  And yes it does result in the extra kernels being in the install menu.

jrtc27:
Are you sure this works? distributekernel / packagekernel is what creates the tarballs, but you've gone and messed with installworld/installkernel/distribution, which is solely for building the live environment the installer runs in. Which is the exact opposite of what needs to change.

asomers:
> Are you sure this works? distributekernel / packagekernel is what creates the tarballs, but you've gone and messed with installworld/installkernel/distribution, which is solely for building the live environment the installer runs in. Which is the exact opposite of what needs to change.

I can verify that this change works.  It results in the alternate kernel being added to the installation menu in the CD.  And it also results in the alternate kernel being added to the CD's own /boot, which is not ideal.

jrtc27:
> Hmm, on CheriBSD we've added individual kernels to release/scripts/make-manifest.sh so they are in the distribution menu in the installer (this is decidedly not generalized).

There are defaults in there. Without modifying the script it'll default to a blank description and selected for install by default. Including the debug symbols tarballs, which is unlikely to be desired, even if you want the kernels.

jrtc27:
> > Are you sure this works? distributekernel / packagekernel is what creates the tarballs, but you've gone and messed with installworld/installkernel/distribution, which is solely for building the live environment the installer runs in. Which is the exact opposite of what needs to change.
> 
> I can verify that this change works. It results in the alternate kernel being added to the installation menu in the CD. And it also results in the alternate kernel being added to the CD's own /boot, which is not ideal.

Something's gone wrong then, because it should not do so.

asomers:
> Hmm, on CheriBSD we've added individual kernels to release/scripts/make-manifest.sh

@brooksdavis could you please share that change?  It seems to me that the natural way to add more kernels is in the _invocation_of make-manifest.sh, which is in release/Makefile, not within make-manifest.sh itself.

jrtc27:
> > > Are you sure this works? distributekernel / packagekernel is what creates the tarballs, but you've gone and messed with installworld/installkernel/distribution, which is solely for building the live environment the installer runs in. Which is the exact opposite of what needs to change.
> > 
> > 
> > I can verify that this change works. It results in the alternate kernel being added to the installation menu in the CD. And it also results in the alternate kernel being added to the CD's own /boot, which is not ideal.
> 
> Something's gone wrong then, because it should not do so.

I wonder if the use of `${MAKE}` in release/Makefile ends up passing down all the random flags passed by the parent process to the child process, and so it's your modification of RELEASE_RMAKEFLAGS that effectively messes with everything. I'd guess that, were you to just change INSTALLEXTRAKERNELS's default in release/Makefile, *without* changing release/Makefile, it would be broken. Probably also the changes you've made to release/Makefile itself don't matter when called from release.sh...

Anyway, regardless, this is quite broken.

jrtc27:
> > Hmm, on CheriBSD we've added individual kernels to release/scripts/make-manifest.sh
> 
> @brooksdavis could you please share that change? It seems to me that the natural way to add more kernels is in the _invocation_of make-manifest.sh, which is in release/Makefile, not within make-manifest.sh itself.

https://github.com/CTSRD-CHERI/cheribsd/commits/main/release/scripts/make-manifest.sh

asomers:
> > > Hmm, on CheriBSD we've added individual kernels to release/scripts/make-manifest.sh
> > 
> > 
> > @brooksdavis could you please share that change? It seems to me that the natural way to add more kernels is in the _invocation_of make-manifest.sh, which is in release/Makefile, not within make-manifest.sh itself.
> 
> https://github.com/CTSRD-CHERI/cheribsd/commits/main/release/scripts/make-manifest.sh

Oh, that just provides descriptions for the alternate kernels in the installer.  It's the invocation of make-manifest.sh, in release/Makefile, that puts the alternate kernels there in the first place.  By default, without customized descriptions.

asomers:
> I wonder if the use of `${MAKE}` in release/Makefile ends up passing down all the random flags passed by the parent process to the child process, and so it's your modification of RELEASE_RMAKEFLAGS that effectively messes with everything.

Yes, I think that's exactly what's happening.  The change to RELEASE_RMAKEFLAGS is the key part.  Would you have done it differently?

jrtc27:
> > I wonder if the use of `${MAKE}` in release/Makefile ends up passing down all the random flags passed by the parent process to the child process, and so it's your modification of RELEASE_RMAKEFLAGS that effectively messes with everything.
> 
> Yes, I think that's exactly what's happening. The change to RELEASE_RMAKEFLAGS is the key part. Would you have done it differently?

Use a different name in release/Makefile to capture that it's only for the tarballs? Then change the distributekernel invocation to pass that as INSTALLEXTRAKERNELS, and leave the other recursive makes alone.

As it stands the changes to release/Makefile are completely useless, since INSTALLEXTRAKERNELS already gets passed down if provided. So there's no point making them unless it's to enable different behaviour to what you can already get with that.

asomers:
ping @jrtc27 how does it look to you now?

bsdimp:
@asomers I think we do one final pint to @jrtc27 and commit if there's no reply in a couple of days.

Makefile.inc1
```diff
@@ -1843,13 +1843,13 @@ buildkernel: .MAKE .PHONY
 	@echo "--------------------------------------------------------------"
 
 .if !make(packages) && !make(update-packages)
-NO_INSTALLEXTRAKERNELS?=	yes
+INSTALLEXTRAKERNELS?=	no
```
bsdimp, line 1846: Hmmm, this is an old NO_xxx hack. It shouldn't be changaed like this, but instead turned into a real option WITH_INSTALL_EXTRA_KERNELS and set MK_INSTALL_EXTRA_KERNELS here.

Makefile.inc1
```diff
@@ -1843,13 +1843,13 @@ buildkernel: .MAKE .PHONY
...
 .else
 # packages/update-packages installs kernels to a staging directory then builds
 # packages from the result to be installed, typically to other systems.  It is
 # less surprising for these targets to honor KERNCONF if multiple kernels are
 # specified.
-NO_INSTALLEXTRAKERNELS?=	no
+INSTALLEXTRAKERNELS?=	yes
```
ngie-eign, line 1852:
```suggestion
MK_INSTALLEXTRAKERNELS?=	yes
```

Makefile.inc1
ngie-eign, line 1846:
```suggestion
MK_INSTALLEXTRAKERNELS?=	no
```

Makefile.inc1
```diff
@@ -1878,7 +1878,7 @@ reinstallkernel reinstallkernel.debug: _installcheck_kernel .PHONY
 	  echo ">>> Install kernel(s) ${INSTALLKERNEL} completed in $$seconds seconds, ncpu: $$(${_ncpu_cmd})${.MAKE.JOBS:S/^/, make -j/}"
 	@echo "--------------------------------------------------------------"
 .endif
-.if ${BUILDKERNELS:[#]} > 1 && ${NO_INSTALLEXTRAKERNELS} != "yes"
+.if ${BUILDKERNELS:[#]} > 1 && ${INSTALLEXTRAKERNELS} != "no"
```
ngie-eign, line 1881:
```suggestion
.if ${BUILDKERNELS:[#]} > 1 && ${MK_INSTALLEXTRAKERNELS} != "no"
```

Makefile.inc1
```diff
@@ -1917,7 +1917,7 @@ distributekernel distributekernel.debug: .PHONY
 	    ${DESTDIR}/${DISTDIR}/kernel.meta
 .endif
 .endif
-.if ${BUILDKERNELS:[#]} > 1 && ${NO_INSTALLEXTRAKERNELS} != "yes"
+.if ${BUILDKERNELS:[#]} > 1 && ${INSTALLEXTRAKERNELS} != "no"
```
ngie-eign, line 1920:
```suggestion
.if ${BUILDKERNELS:[#]} > 1 && ${MK_INSTALLEXTRAKERNELS} != "no"
```

Makefile.inc1
```diff
@@ -1951,7 +1951,7 @@ packagekernel: .PHONY
 	    @${DESTDIR}/${DISTDIR}/kernel.meta | \
 	    ${XZ_CMD} > ${DESTDIR}/${DISTDIR}/kernel-dbg.txz
 .endif
-.if ${BUILDKERNELS:[#]} > 1 && ${NO_INSTALLEXTRAKERNELS} != "yes"
+.if ${BUILDKERNELS:[#]} > 1 && ${INSTALLEXTRAKERNELS} != "no"
```
ngie-eign, line 1954:
```suggestion
.if ${BUILDKERNELS:[#]} > 1 && ${MK_INSTALLEXTRAKERNELS} != "no"
```

Makefile.inc1
```diff
@@ -1976,7 +1976,7 @@ packagekernel: .PHONY
 	    ${TAR_CMD} cvf - --include '*/*/*.debug' $$(eval find .) | \
 	    ${XZ_CMD} > ${DESTDIR}/${DISTDIR}/kernel-dbg.txz
 .endif
-.if ${BUILDKERNELS:[#]} > 1 && ${NO_INSTALLEXTRAKERNELS} != "yes"
+.if ${BUILDKERNELS:[#]} > 1 && ${INSTALLEXTRAKERNELS} != "no"
```
ngie-eign, line 1979:
```suggestion
.if ${BUILDKERNELS:[#]} > 1 && ${MK_INSTALLEXTRAKERNELS} != "no"
```

Makefile.inc1
```diff
@@ -2059,7 +2059,7 @@ BRANCH_EXT_FROM=
 .endif
 
 PKGMAKEARGS+=	PKG_VERSION=${PKG_VERSION} \
-		NO_INSTALLEXTRAKERNELS=${NO_INSTALLEXTRAKERNELS}
+		INSTALLEXTRAKERNELS=${INSTALLEXTRAKERNELS}
```
ngie-eign, line 2062:
```suggestion
		MK_INSTALLEXTRAKERNELS=${MK_INSTALLEXTRAKERNELS}
```

Makefile.inc1
```diff
@@ -2307,7 +2307,7 @@ create-kernel-packages-flavor${flavor:C,^""$,${_default_flavor},}: _pkgbootstrap
 		-o ${REPODIR}/${PKG_ABI}/${PKG_OUTPUT_DIR}
 . endfor
 .endif
-.if ${BUILDKERNELS:[#]} > 1 && ${NO_INSTALLEXTRAKERNELS} != "yes"
+.if ${BUILDKERNELS:[#]} > 1 && ${INSTALLEXTRAKERNELS} != "no"
```
ngie-eign, line 2310:
```suggestion
.if ${BUILDKERNELS:[#]} > 1 && ${MK_INSTALLEXTRAKERNELS} != "no"
```

release/release.sh
```diff
@@ -196,6 +196,7 @@ env_check() {
 	RELEASE_KMAKEFLAGS="${MAKE_FLAGS} ${KERNEL_FLAGS} \
 		KERNCONF=\"${KERNEL}\" ${ARCH_FLAGS} ${CONF_FILES}"
 	RELEASE_RMAKEFLAGS="${ARCH_FLAGS} ${RELEASE_FLAGS} \
+		INSTALLEXTRAKERNELS=\"yes\" \
```
ngie-eign, line 199:
```suggestion
		MK_INSTALLEXTRAKERNELS=\"yes\" \
```

release/tools/vmimage.subr
```diff
@@ -53,6 +53,7 @@ vm_install_base() {
 
 	cd ${WORLDDIR} && \
 		make DESTDIR=${DESTDIR} ${INSTALLOPTS} \
+		INSTALLEXTRAKERNELS=${INSTALLEXTRAKERNELS} \
```
ngie-eign, line 56:
```suggestion
		MK_INSTALLEXTRAKERNELS=${INSTALLEXTRAKERNELS} \
```

bsdimp:
This is showing conflicts. I don't think it's still relevant. Can you closer or rebase + fix?

bsdimp:
OK. Closing this as inactive. If you have changes reflecting the review, please feel free to resubmit and/or commit directly.

PR #1567, opened 2025-01-05
hda: Remove duplicate macros

JohnAZoidberg:
Both are defined twice.


emaste:
These were removed in 4004e1752ec7992eeace7f0318d2ca3fc7ba9e03 - it looks like your tree is a bit behind

PR #1570, opened 2025-01-12
github: add new checklist workflow

VexedUXR:
Add a new 'checklist' workflow that checks the commit messages on pull
requests. Currently, the workflow creates a comment on the pull request
if any of these conditions are hit:
- Missing Signed-off-by
- Malformed Signed-off-by
- Bad email (i.e \*noreply\*)

A test can be found here: https://github.com/VexedUXR/freebsd-src/pull/6
The comment is edited as the pull request is updated.
The comment get updated to:
```
Thank you for taking the time to contribute to FreeBSD!
All issues resolved.
```
When all issues are resolved.
If no issues are found in the first place, then no comment is added.

### DIFF
--- a/.github/workflows/checklist.yml
+++ b/.github/workflows/checklist.yml
@@ -1,0 +1,106 @@
+name: Checklist
+
+# Produce a list of things that need to be changed
+# for the submission to align with CONTRIBUTING.md
+
+on:
+  pull_request:
+    types: [ opened, reopened, edited, synchronize ]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  checklist:
+    name: commit
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v7
+        with:
+          # An asynchronous javascript function
+          script: |
+            /*
+             * Github's API returns the results in pages of 30, so
+             * pass the function we want, along with it's arguments,
+             * to paginate() which will handle gathering all the results.
+             */
+            const comments = await github.paginate(github.rest.issues.listComments, {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number
+            });
+
+            const commits = await github.paginate(github.rest.pulls.listCommits, {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.issue.number
+            });
+
+            let checklist = {};
+            let checklist_len = 0;
+            let comment_id = -1;
+
+            const msg_prefix = "Thank you for taking the time to contribute to FreeBSD!\n";
+            const addToChecklist = (msg, sha) => {
+              if (!checklist[msg]) {
+                checklist[msg] = [];
+                checklist_len++;
+              }
+              checklist[msg].push(sha);
+            }
+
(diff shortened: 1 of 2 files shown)


bsdimp (approved):
I think I really like these changes. Thanks! Sorry it took me a bit to get around to looking at them.

emaste:
We should check that the (a) Signed-off-by matches the author's email too, no?

But indeed I like this. There are existing actions we could use too (e.g. https://github.com/dcoapp/app) but it's simple enough that a small bespoke GH action seems reasonable (and it's nice to capture all of the various things we want to include, like noreply emails)

.github/workflows/checklist.yml
```diff
@@ -0,0 +1,99 @@
...
+  pull_request:
+    types: [ opened, reopened, edited, synchronize ]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  checklist:
+    name: commit
+    runs-on: ubuntu-latest
```
emaste, line 16: `# XXX would prefer to run this on FreeBSD-latest` :)

brooksdavis:
> We should check that the (or, a) Signed-off-by matches the author's email too, no?

Not necessarily, but it's probably fine to warn if they aren't the same. For example, for FreeBSD committers it wouldn't be unreasonable to have a Signed-off-by with your work email while Author is @freebsd.org.

bsdimp:
> > We should check that the (or, a) Signed-off-by matches the author's email too, no?
> 
> Not necessarily, but it's probably fine to warn if they aren't the same. For example, for FreeBSD committers it wouldn't be unreasonable to have a Signed-off-by with your work email while Author is @freebsd.org.

Other projects warn when they aren't the same. Maybe we could steal their code, but convert it to a warning.
Plus Signed-off-by can be generated by multiple people in the chain between production of the code and its contribution, so we should take that into account. It's not a unique field.

.github/workflows/checklist.yml
VexedUXR: Yes, hopefully one day

VexedUXR:
Do you mean output a warning like the style checker does using https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-a-warning-message?
Or should it be appended to the comment, e.g:
```
There are a few issues that need to be fixed:
...
Additionally, there are a few things that you might want to change:
...
```

The former seems to have limited visibility, there is no indication the job outputted any warnings unless you click on it.

> Plus Signed-off-by can be generated by multiple people in the chain between production of the code and its contribution, so we should take that into account. It's not a unique field.

The current code already loops over all the sob lines :)

VexedUXR:
\- Print a warning if author's email isn't included in the sob lines
\- Resolved merge conflicts
\- Made the "freebsd.org" check case-insensitive to match behavior of checkstyle9.pl
\- Also made the noreply check case-insensitive

I opted for logging the warning instead of appending it to the comment.
The test run can be found here: https://github.com/VexedUXR/freebsd-src/actions/runs/12949254004/job/36119600320?pr=6

VexedUXR:
Thanks!

PR #1571, opened 2025-01-13
release/amd64: cleanup code duplication

VexedUXR:
Improve the release script by checking `MK_LOADER_IA32` instead of the janky `if -f "${BASEBITSDIR}/boot/loader_ia32.efi"`

I'm hesitant about putting this in install-boot.sh, it feels out of place.
Maybe a new script in release/amd64 that gets sourced by both `make-memstick` and `mkisoimages` would be better? I'm not sure..

This also makes the script complain if `MK_LOADER_IA32` is `yes` but the file is missing, instead of silently not copying it.
FWIW it also allows the user to build the ia32 loader, but omit it from the img, though I'm not sure what the use case would be.

I've been meaning to do this since 599273f landed but forgot, so here it is.

### DIFF
--- a/release/amd64/make-memstick.sh
+++ b/release/amd64/make-memstick.sh
@@ -61,11 +61,9 @@ fi
 # Make an ESP in a file.
 espfilename=$(mktemp /tmp/efiboot.XXXXXX)
 if [ -f "${BASEBITSDIR}/boot/loader_ia32.efi" ]; then
-	make_esp_file ${espfilename} ${fat32min} ${BASEBITSDIR}/boot/loader.efi bootx64 \
-	    ${BASEBITSDIR}/boot/loader_ia32.efi bootia32
-else
-	make_esp_file ${espfilename} ${fat32min} ${BASEBITSDIR}/boot/loader.efi
+	extra_args="${BASEBITSDIR}/boot/loader_ia32.efi bootia32"
 fi
+make_esp_file ${espfilename} ${fat32min} ${BASEBITSDIR}/boot/loader.efi bootx64 ${extra_args}
 
 mkimg -s mbr \
     -b ${BASEBITSDIR}/boot/mbr \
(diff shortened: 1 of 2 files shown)


release/amd64/make-memstick.sh
```diff
@@ -60,12 +60,10 @@ fi
 
 # Make an ESP in a file.
 espfilename=$(mktemp /tmp/efiboot.XXXXXX)
-if [ -f "${BASEBITSDIR}/boot/loader_ia32.efi" ]; then
-	make_esp_file ${espfilename} ${fat32min} ${BASEBITSDIR}/boot/loader.efi bootx64 \
-	    ${BASEBITSDIR}/boot/loader_ia32.efi bootia32
-else
-	make_esp_file ${espfilename} ${fat32min} ${BASEBITSDIR}/boot/loader.efi
+if [ ${MK_LOADER_IA32} = "yes" ]; then
+	extra_args="${BASEBITSDIR}/boot/loader_ia32.efi bootia32"
```
bsdjhb, line 64: Should "bootx64" be in the extra_args case here?  It's not present in the existing code for the !LOADER_IA32 case.

bsdjhb (commented):
I think checking for the existence of the file is fine vs checking MK_LOADER_IA32.  That said, I think the cleanup to use a helper extra_args variable to reduce code duplication is useful in its own right (and probably worthy of being its own commit).  I do think bootx64 belongs in extra_args though.

release/amd64/make-memstick.sh
VexedUXR: It shouldn't matter. If `make_esp_file` isn't given the "bootx64", it automatically determines it using `uname -m`.
I just thought this would make more sense since the arguments are given in pairs, and the condition was for the ia32 pair.

VexedUXR:
Ok, went back to simply checking if the loader files exist. :)

bsdimp:
@VexedUXR Can you commit or close? I think we should do this before 15.

VexedUXR:
Ah, I forgot about this. I'll commit it.

PR #1572, opened 2025-01-16
fwget: add needed firmware for AMD Granite Ridge GPUs

SchrodingerZhu:
This is from my 9950X:
```
❯ sudo pciconf -lBbcevV | grep vga -A7
vgapci0@pci0:6:0:0:     class=0x030000 rev=0xc1 hdr=0x00 vendor=0x1002 device=0x13c0 subvendor=0x1002 subdevice=0x13c0
    vendor     = 'Advanced Micro Devices, Inc. [AMD/ATI]'
    device     = 'Granite Ridge [Radeon Graphics]'
    class      = display
    subclass   = VGA
    bar   [10] = type Prefetchable Memory, range 64, base 0xfcf0000000, size 268435456, enabled
    bar   [18] = type Prefetchable Memory, range 64, base 0xf6400000, size 2097152, enabled
    bar   [20] = type I/O Port, range 32, base 0xd000, size 256, enabled
```

### DIFF
--- a/usr.sbin/fwget/pci/pci_video_amd
+++ b/usr.sbin/fwget/pci/pci_video_amd
@@ -160,7 +160,7 @@ pci_video_amd()
 			addpkg "gpu-firmware-amd-kmod-sdma-6-0-1"
 			addpkg "gpu-firmware-amd-kmod-vcn-4-0-2"
 			;;
-		0x164e)
+		0x164e|0x13c0)
 			addpkg "gpu-firmware-amd-kmod-gc-10-3-6"
 			addpkg "gpu-firmware-amd-kmod-psp-13-0-5"
 			addpkg "gpu-firmware-amd-kmod-dcn-3-1-5"


SchrodingerZhu:
A gentle ping~

usr.sbin/fwget/pci/pci_video_amd
```diff
@@ -160,7 +160,7 @@ pci_video_amd()
 			addpkg "gpu-firmware-amd-kmod-sdma-6-0-1"
 			addpkg "gpu-firmware-amd-kmod-vcn-4-0-2"
 			;;
-		0x164e)
+		0x164e|0x13c0)
```
markjdb, line 163: This will get matched by the 0x13*) case on line 48, so this change has no effect, I believe.

SchrodingerZhu: https://github.com/freebsd/freebsd-src/blob/2fa507442cf18e667426e19367a5f1e108354ebf/usr.sbin/fwget/pci/pci_video_amd#L135

But we already have this line? Should that line also be fixed?

markjdb: Yes, that seems like a bug. cc @evadot 

Really, sh switch statements are a poor way to handle this logic. It would be better to use some set of lua tables to map devices to packages. That'd make it easier to variously permit multiple matches, test whether a device has more than one match, etc..

PR #1573, opened 2025-01-17
ice(4): Add host SR-IOV support

kgalazka:
Allows creating Virtual Functions on E800 adapters with up to 16 queues.
Changing MAC address and using promiscuous mode for each VF
may be enabled in configuration.

Please be advised that this is still work in progress and may be unstable.

Authored-by: Eric Joyner <erj@FreeBSD.org>
Signed-off-by: Krzysztof Galazka <krzysztof.galazka@intel.com>

### DIFF
--- a/sys/dev/ice/ice_iov.c
+++ b/sys/dev/ice/ice_iov.c
@@ -1139,17 +1139,34 @@ ice_vc_set_rss_hena_msg(struct ice_softc *sc, struct ice_vf *vf, u8 *msg_buf)
 	enum virtchnl_status_code v_status = VIRTCHNL_STATUS_SUCCESS;
 	struct ice_vsi *vsi = vf->vsi;
 
+	MPASS(vsi != NULL);
+
 	vrh = (struct virtchnl_rss_hena *)msg_buf;
 
-	status = ice_add_avf_rss_cfg(hw, vsi->idx, vrh->hena);
-	if (status) {
-		device_printf(sc->dev,
-			      "ice_add_avf_rss_cfg status %s, error %s\n",
-			      ice_status_str(status),
-			      ice_aq_str(hw->adminq.sq_last_status));
-		v_status = ice_iov_err_to_virt_err(status);
+	/*
+	 * Remove existing configuration to make sure only requested
+	 * config is applied and allow VFs to disable RSS completly.
+	 */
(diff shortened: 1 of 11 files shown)


concussious:
> Changing MAC address and using promiscuous mode for each VF may be enabled in configuration.

Would you be willing to update the manual page for this change?

ricera:
If I was sufficiently motivated, I could maybe take a crack at it.

@kgalazka have you tried this on E830 devices? The last time I tried it didn't work; I don't know if it's worth calling out somewhere that it doesn't work on those since E830 hasn't been released yet, but I imagine once they are people will expect this to work on those and will be surprised that it mysteriously does not.

sys/dev/ice/ice_iov.c
```diff
@@ -369,6 +369,14 @@ ice_iov_add_vf(struct ice_softc *sc, uint16_t vfnum, const nvlist_t *params)
 		goto release_imap;
 	}
 
+	/* Add the broadcast address */
+	error = ice_add_vsi_mac_filter(vsi, broadcastaddr);
```
ricera, line 373: Doing this here might mean that this doesn't get re-added after a reset, right? Though, I want to swear that broadcast promiscuous mode should be getting enabled on this VF VSI already.

kgalazka:
> @kgalazka have you tried this on E830 devices? The last time I tried it didn't work; I don't know if it's worth calling out somewhere that it doesn't work on those since E830 hasn't been released yet, but I imagine once they are people will expect this to work on those and will be surprised that it mysteriously does not.

Not there yet. I'm trying to make it ready for full blown testing on E810 first.

sys/dev/ice/ice_iov.c
kgalazka: > Doing this here might mean that this doesn't get re-added after a reset, right?

Yeah, it's not the best place and I intend to move it.

>Though, I want to swear that broadcast promiscuous mode should be getting enabled on this VF VSI already.

Not in ICE as far as I know. We do that in IXL though.

bsdimp:
looks like this applies, compiles and is up to date.
I'm going to commit it, but keep an eye out for ice(4) complaints.

bsdimp:

PR #1574, opened 2025-01-17
Unsupported hash algorithm should not be considered a success

Huwyler:
Honestly I haven't looked deeply into it, it just seems to me that this line is missing here


emaste:
cc @sgerraty

sgerraty:
LGTM - sorry no idea what to do with pull requests though

emaste:
I pushed as caaeab697bf98bf96e2fa8cb4a1e22240511fbcc

sgerraty:
Feel free thanks, though I should expand that to actually allow for sha384 and sha512 - though I don't know if gpg will actually generate them.

PR #1575, opened 2025-01-17
makesyscalls: restore support for cpp in input

brooksdavis:
Implement functionality largely identical to that in makesyscalls.lua on 14-STABLE. I've excluded prolog and epilogue support as it's not obviously very useful. Could restore the original function if someone is using conditional includes or the like.

Deprecate this functionality (no plan to remove it without replacement, but putting a line in the sand.)


brooksdavis:
CC @agge3

brooksdavis:
I suspect we also need ifdefs in syscalls.map, but I wanted to start with replicating what's in the old version.

brooksdavis:
Pulled in @kevans91's test script and updated for main (I should probably MFC the diff reduction changes.)

Fixed the way the epilogue is handled.

Added syscalls.map cpp support and added an unrelated minor improvement to the syscall.mk code.

agge3:
On the topic of adding tests, I've also added tests: [3c7e9e9](https://github.com/freebsd/freebsd-src/commit/3c7e9e97fee37ff01732a30abc6c80263296ef86). The tests are directed at the parser. They confirm correct parsing by tokenizing both the data that `makesyscalls` has processed and `syscalls.master` (again) to confirm that we can regenerate `syscalls.master` from the library. This helps further refactors/additions specifically targeting the parser/data storage matched against `syscalls.master`, allowing a quick test run. It would also be useful for CI/CD.

Please let me know if the tests, `tests/syscalls_tests.lua`, are something that you would like to see more polished. There's a hacked `tests/README.md` for a run demonstration.

The other parts of the commit are for applications to [qemu-bsd-user](https://github.com/qemu-bsd-user/qemu-bsd-user/tree/blitz) and likely should be a different PR (which Warner is helping me with). Not to derail the topic though, just more context.

I wanted to bring up the tests, as a lot of tests have now been introduced in this PR. I could pursue my tests to more refinement to add onto that theme (i.e., all ABIs and not printing to stdout).

kevans91:
LGTM, thanks!

brooksdavis:
I've merged this with the changes of moving the "test" to an "examples/cpp" to stay out of the way of @agge3's tests and with the idea that we might accumulate have multiple small examples of things we don't do in tree over time.

PR #1576, opened 2025-01-20
[PowerPC] Fix POWER9 power save (stop state) routines

madscientist159:
This series fixes latency spikes observed when the cores are allowed to enter stop states, and also adds a tunable to configure the maximum stop state enabled on the machine.  The latency spikes are due to non-deterministic wakeup interrupts, and are highly observable when using a lightly loaded OPNsense firewall as random ping response spikes from ~0.2ms to 200ms+

### DIFF
--- a/sys/powerpc/powerpc/cpu.c
+++ b/sys/powerpc/powerpc/cpu.c
@@ -95,6 +95,9 @@ static void	cpu_idle_booke(sbintime_t);
 static void	cpu_idle_e500mc(sbintime_t sbt);
 #endif
 #if defined(__powerpc64__) && defined(AIM)
+static int      cpu_idle_max_stop_state = 2;
+SYSCTL_INT(_machdep, OID_AUTO, cpu_idle_max_stop_state,
+	CTLFLAG_RW, &cpu_idle_max_stop_state, 0, "");
 static void	cpu_idle_powerx(sbintime_t);
 static void	cpu_idle_power9(sbintime_t);
 #endif
@@ -815,6 +818,22 @@ cpu_idle_booke(sbintime_t sbt)
 static void
 cpu_idle_powerx(sbintime_t sbt)
 {
+	int max_stop_state = cpu_idle_max_stop_state;
+
+	/* Limit maximum stop state to valid values */
+	if (max_stop_state < 0) {
+		/* Don't nap at all, busy wait instead */
+		cpu_idle_max_stop_state = -1;
+		return;
+	}
+	if (max_stop_state > 1) {
+		/* POWER8 and below only support the one stop state,
+		 * i.e. 'nap'
+		 */
+		max_stop_state = 1;
+		cpu_idle_max_stop_state = max_stop_state;
+	}
+
 	/* Sleeping when running on one cpu gives no advantages - avoid it */
 	if (smp_started == 0)
 		return;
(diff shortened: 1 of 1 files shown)


madscientist159:
@chmeeedalf This is a follow on to your original stop state support for POWER9.  Comments welcome!

sys/powerpc/powerpc/cpu.c
```diff
@@ -836,23 +855,51 @@ cpu_idle_powerx(sbintime_t sbt)
...
+		 */
+		max_stop_state = 3;
+		cpu_idle_max_stop_state = max_stop_state;
+	}
+
+	/*
+	 * Enter spinlock and suspend external interrupts until the stop
+	 * instruction completes.
+	 */
+	spinlock_enter();
```
chmeeedalf, line 883: Are you sure you want a spinlock_enter() here?  It sets thread priority to higher than normal, but we're about to go to sleep.

chmeeedalf:
@madscientist159 thanks for looking at this, I hope it improves performance!  Just the one question, otherwise looks good to me.

sys/powerpc/powerpc/cpu.c
madscientist159: Yes, I think it does make sense.  First, this is what the `powerx` (i.e. POWER7/POWER8) idle task already does, and second it does conceptually make sense that we want to signal the kernel and drain the task queue as fast as possible to determine if we should actually idle or switch to the next task instead.

For what it's worth, the change is extremely effective.  Here's just one ping plot from before and after this patch was applied, no other changes to the system:
ping_response_time_example

chmeeedalf: Okay, thanks.  I did some digging, and it looks like pseries and x86 just disable interrupts, not the full spinlock dance.  arm64, arm, and riscv all use spinlock_enter()/exit().  spinlock_enter()/exit() seems a bit heavy-handed, a lot of extra code running in that time frame.

madscientist159: Just disabling the interrupts doesn't work -- it gives variable wakeup latency for the thread that I've seen spike as high as 1 to 2 seconds (!).   True pSeries boxes (POWER7 and older or "scale up" systems) didn't really have stop state support of this nature, so I suspect the issue was just never seen, though it would probably still manifest in POWER9 guests running in pSeries mode.

Also note that POWER7/POWER8 do the `spinlock` dance too :wink: 
https://github.com/freebsd/freebsd-src/blob/6d48eafa334f102009dbca9d191c6bbf52855551/sys/powerpc/powerpc/cpu.c#L822

chmeeedalf: Okay, thanks for explaining it.  I thought that the spinlock was just to guard the sched_runnable(), and maybe it is; the syncs done in the spinlock_enter() would be enough to let any writes to the thread state complete before checking.

As for power7/power8, I thought the spinlock was taken just to make things easier, or copied from arm/arm64.  Thanks for your explanation.  I just pushed it as 6f0c2938e.

chmeeedalf:
Merged as 6f0c2938e895bd3ddb14f44acf36dba3c6156a3f thanks!

sys/powerpc/powerpc/cpu.c
madscientist159: No problem, glad to help!  Thanks for the fast merge.  FWIW I do know the C1/C1e states on x86 are idiosyncratic compared to most other architectures, so it's possible x86 can get away without the spinlocks due to some kind of backward compatibility feature.

I have always strongly suspected that the thread state is the underlying problem, and that the kernel tries to schedule threads on the stopped core without issuing a doorbell.  The embedded state syncs would be a reasonable explaination for the `spinlock` requirement; could easily see the core stopping without updating kernel state.  It's possible some sequence of `mb()` could also work, but that's a lot of complexity and work just to be different than arm/riscv/etc. :wink:

PR #1577, opened 2025-01-21
checkstyle9.pl: don't require Signed-off-by for committers

brooksdavis:
Silence the warning about Signed-off-by: lines if the Author: is using a FreeBSD.org address.  We don't require the lines so it's just noise in this case.


bsdimp (approved):

emaste (approved):
I think in the fullness of time it will make sense to have a s-o-b on all commits, but it certainly is not required, expected, or usual today and so this makes sense

PR #1582, opened 2025-01-26
Optionally silence devmatch, routing rc scripts operations

sgharms:
I am interested in helping the Laptop working group (a focus of 2025) and was
trying to help smooth any rough edges that might spook a newer installer/user.
One of the items overwhelming is the chattiness of startup. While boot_mute=YES
catches a stage's worth of messages, the rc startup scripts can be daunting
or...well, there's just a lot of them.  The forums and other development
efforts document this. It's a recurrent forum question and [helloSystem] has
done some...extraordinary work to hide them

Looking at the scripts, there didn't seem to be a unified way to lower their
verbosity. Additionally, what's emitting could be the binary where a
more-silent flag is not used e.g. [routing] OR the rc script [devmatch].
Lacking a single "master control" to silence things, I just want after the
chattiest things that irritated, uh, me. ;)

[routing]: https://github.com/sgharms/freebsd-src/blob/main/libexec/rc/rc.d/routing#L22
[devmatch]: https://github.com/sgharms/freebsd-src/blob/main/libexec/rc/rc.d/devmatch#L78
[helloSystem]: https://hellosystem.github.io/docs/developer/boot.html#boot-mute

### DIFF
--- a/libexec/rc/rc.conf
+++ b/libexec/rc/rc.conf
@@ -45,6 +45,8 @@ devd_enable="YES" 	# Run devd, to trigger programs on device tree changes.
 devd_flags=""		# Additional flags for devd(8).
 devmatch_enable="YES"	# Demand load kernel modules based on device ids.
 devmatch_blocklist=""	# List of modules (w/o .ko) to exclude from devmatch.
+devmatch_autoloading_msgs="NO"	# Do not log "Autoloading" to console from devmatch rc.
+
 #kld_list="" 		# Kernel modules to load after local disks are mounted
 kldxref_enable="YES"	# Build linker.hints files with kldxref(8).
 kldxref_clobber="NO"	# Overwrite old linker.hints at boot.
--- a/libexec/rc/rc.conf
+++ b/libexec/rc/rc.conf
@@ -465,6 +465,7 @@ routed_flags="-q"		# Flags for routing daemon.
 arpproxy_all="NO"		# replaces obsolete kernel option ARP_PROXYALL.
 forward_sourceroute="NO"	# do source routing (only if gateway_enable is set to "YES")
 accept_sourceroute="NO"		# accept source routed packets to us
+routing_flags=""		# Flags for adjusting calls to route(8) from routing
 
 ### Bluetooth ###
 hcsecd_enable="NO"		# Enable hcsecd(8) (or NO)
--- a/libexec/rc/rc.d/devmatch
+++ b/libexec/rc/rc.d/devmatch
@@ -74,8 +74,12 @@ devmatch_start()
 		case "${x}" in
 		*"#${m}#"*) continue ;;
 		esac
-		kldstat -q -n ${m} || \
-		    (echo "Autoloading module: ${m}"; kldload -n ${m})
+		if checkyesno devmatch_autoloading_msgs; then
+			kldstat -q -n ${m} || \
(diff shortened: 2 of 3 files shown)


Gulcantaysun1 (approved):

jlduran:
Thank you for your interest regarding this topic.
I tried silencing a few rc scripts as well a while back (https://github.com/jlduran/freebsd-src/pull/90).
The reason why I never submitted it, was because `check_startmsgs` should _just_ silence those services that print "Starting \<service\>".  If we start silencing all echoed messages behind this check, a change in the documentation is required (that can be a bit more delicate than what it sounds).
There's also `rc_quiet`, and as you mentioned `boot_mute`, even using all of them combined won't get us a "clean" boot.
This is my personal opinion on the subject, hopefully this pull request sparks the interest of the rc working group, I'll keep subscribed, as I also find it appealing to have an easy, out-of-the-box "clean" boot option.

Gulcantaysun1:
Tecrübesizim bu tür konularda , cihazımda sürekli olarak değiştirildiğini hissettiğim birseyler var. Değiştirilmiş birşeyi nasıl bulabilirim

cederom:
Thanks @sgharms :-) I am against the change in current form because:
* The concept may be useful for someone but it shall be a manually selectable option like `rc_mute` (in analog to `boot_mute`) and not a default. Maybe `rc_loglevel` that can get `silent, error, warning, info, verbose, debug` levels to rc scripts and services? Change does not provide a choice.
* Boot log messages are extremely useful, they contain important diagnostic messages, these need to be logged even if silenced, these cannot be just sent to `/dev/null`. Change alters (breaks) logging.
* Why not just put a boot logo image if you really do not want to see the boot messages? It will provide better user experience as expected but will not break the diagnostic information that are usually useful.

cederom (changes requested):
Thanks @sgharms :-)
* There is already `rc_info` that may be used? https://github.com/freebsd/freebsd-src/blob/50c962d773705f60d32c29f8e4bb2743ab55733b/libexec/rc/rc.conf#L29
* There is already `rc_startmsg` that may be used? https://github.com/freebsd/freebsd-src/blob/50c962d773705f60d32c29f8e4bb2743ab55733b/libexec/rc/rc.conf#L30
* If really necessary `_msgs` can be replaced with `_verbose` suffix? https://github.com/freebsd/freebsd-src/blob/50c962d773705f60d32c29f8e4bb2743ab55733b/libexec/rc/rc.conf#L568
* Adding `routing_flags` seems okay as `-q` may be set that way :-)
* I guess the idea is to silence the terminal? How many flags need to be put to silent instead just one? Why not just put a boot logo image if you really do not want to see the boot messages? It may provide expected user experience and will not impact diagnostic information that are usually useful :-)

sgharms:
To close the loop here (in parallel to the mailing list): I think the optimal solution is for boot_mute=YES to extend the overlay until the point that the login menu is ready (probably after doing a screen reset so that the login prompt is the only thing on screen).

Presently boot_mute seems to stop the overlay after the kernel gets a hold of the VTs (explains the phenomenon I'm seeing here). Extending that overlay's duration would fix things IMHO.

cederom:
Thank you @sgharms for a good discussion !! :-)

Gulcantaysun1 (approved):

PR #1583, opened 2025-01-26
acpi_iicbus: install the address space handler by default

VexedUXR:
This was previously left off since it wasn't tested.
I've double checked the implementation of both the address space handler itself and the various smbus commands we issue, everything looks good.

Also tested on my Acer one 10 s1002.


wulf7:
LGTM

wulf7 (approved):

wulf7 (approved):

markjdb:
Merged, thank you.

PR #1584, opened 2025-01-27
Fixed EFI loader for UGA protocol

frarohi:
Since `FreeBSD Rel. 13.0`, EFI boot on machines with UGA graphics ends with a black screen, after display of the 'Start @' message.

Apparently `framebuffer.c/efi_find_framebuffer` returned with error when no GOP graphics was found instead of continuing and probing UGA graphics.

Furthermore, switching to the now found UGA graphics blanked the screen until the kernel took over. UGA graphics code in the loader seems to be buggy and needs to be fixed.
Until then, a patch is introduced in `efi_console.c/cons_update_mode` to use text mode in this case.  

CHANGES:
 
stand/efi/loader/framebuffer.c:
- Fixed getting EFI framebuffer settings for UGA protocol.
- Added Apple MacBookPro3,1 framebuffer parameters for 17" model with 1680x1050 screen.
stand/efi/libefi/efi_console.c:
- Disabled switching to UGA graphics since UGA graphics protocol is broken.
stand/efi/loader/bootinfo.c:
- Added 'No framebuffer found' message

### DIFF
--- a/stand/efi/loader/bootinfo.c
+++ b/stand/efi/loader/bootinfo.c
@@ -207,6 +207,8 @@ bi_load_efi_data(struct preloaded_file *kfp, bool exit_bs)
 		    efifb.fb_mask_reserved);
 
 		file_addmetadata(kfp, MODINFOMD_EFI_FB, sizeof(efifb), &efifb);
+	} else {
+		printf("No EFI framebuffer found!\n");
 	}
 #endif
 
--- a/stand/efi/loader/framebuffer.c
+++ b/stand/efi/loader/framebuffer.c
@@ -379,6 +379,7 @@ efifb_from_uga(struct efi_fb *efifb)
 			/* These are the missing bits. */
 			offset = 0x10000;
 			stride = 1728;
+			printf("Model found: %s\n", ev);
 		} else if (ev != NULL && !strcmp(ev, "MacBook3,1")) {
 			/* These are the expected values we should have. */
 			horiz = 1280;
@@ -387,6 +388,23 @@ efifb_from_uga(struct efi_fb *efifb)
 			/* These are the missing bits. */
 			offset = 0x0;
 			stride = 2048;
+			printf("Model found: %s\n", ev);
+		} else if (ev != NULL && !strcmp(ev, "MacBookPro3,1")) {
+			/*
+			 * Valid for MacBookPro 17" with standard resolution.
+			 * Other Models are:
+			 *   MaBookPro 15" with horiz=1440
+			 *   MaBookPro 17" with horiz=1920
+			 * ToDo: differentiate the variants
+			 */
+			/* These are the expected values we should have. */
+			horiz = 1680;
+			vert = 1050;
+			fbaddr = 0xc0000000;
+			/* These are the missing bits. */
+			stride = 2048;
+			offset = stride * 4 * 24; // 24 scan lines down
+			printf("Model found: %s\n", ev);
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


ImAwsumm (approved):

stand/efi/loader/framebuffer.c
```diff
@@ -387,6 +388,23 @@ efifb_from_uga(struct efi_fb *efifb)
...
 			offset = 0x0;
 			stride = 2048;
+			printf("Model found: %s\n", ev);
+		} else if (ev != NULL && !strcmp(ev, "MacBookPro3,1")) {
+			/*
+			 * Valid for MacBookPro 17" with standard resolution.
+			 * Other Models are:
+			 *   MaBookPro 15" with horiz=1440
+			 *   MaBookPro 17" with horiz=1920
+			 * ToDo: differentiate the variants
```
VexedUXR, line 398: Do the other variants have the same offset, stride and fb address?
If so, we could just leave `horiz` and `vert` as reported by `uga->GetMode`.

stand/efi/loader/bootinfo.c
```diff
@@ -207,6 +207,8 @@ bi_load_efi_data(struct preloaded_file *kfp, bool exit_bs)
 		    efifb.fb_mask_reserved);
 
 		file_addmetadata(kfp, MODINFOMD_EFI_FB, sizeof(efifb), &efifb);
+	} else {
+		printf("No EFI framebuffer found!\n");
```
VexedUXR, line 211: Don't see why not. But it should be a separate commit.

VexedUXR (commented):
Hello,

Sorry this went so long without any feedback. This issue was recently brought up in a bug report [1], which led to me stumbling across this. I have two patches up for review [2], [3]. The second one should fix the blank screen issue.

Thanks!

[1] https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=291935
[2] https://reviews.freebsd.org/D54431
[3] https://reviews.freebsd.org/D54432

frarohi:
Hello,great to see that you fixed this issue thoroughly, so that UGA is going to be usable in EFI boot.


> In stand/efi/loader/framebuffer.c:
> @@ -387,6 +388,23 @@ efifb_from_uga(struct efi_fb *efifb)
> /* These are the missing bits. */
> offset = 0x0;
> stride = 2048;
>+			printf("Model found: %s\n", ev);
>+		} else if (ev != NULL && !strcmp(ev, "MacBookPro3,1")) {
>+			/*
>+			 * Valid for MacBookPro 17" with standard resolution.
>+			 * Other Models are:
>+			 *   MaBookPro 15" with horiz=1440
>+			 *   MaBookPro 17" with horiz=1920
>+			 * ToDo: differentiate the variants

>Do the other variants have the same offset, stride and fb address?
>If so, we could just leave horiz and vert as reported by uga->GetMode.

ACK. I think offset and stride might need adjustments for the other variants. Unfortunately, I don‘t have access to any of those machines, so I left it up to be set by other people.



>In stand/efi/loader/bootinfo.c:
> @@ -207,6 +207,8 @@ bi_load_efi_data(struct preloaded_file *kfp, bool exit_bs)
> 		    efifb.fb_mask_reserved);
> 		file_addmetadata(kfp, MODINFOMD_EFI_FB, sizeof(efifb), &efifb);
>+	} else {
>+		printf("No EFI framebuffer found!\n");

>Don't see why not. But it should be a separate commit.


This debug message was helpful for me to trace down the bug.As soon as possible I will be testing the fixed routine on my device.Regards, Frank

stand/efi/loader/framebuffer.c
VexedUXR: > ACK. I think offset and stride might need adjustments for the other variants. Unfortunately, I don‘t have access to any of those machines, so I left it up to be set by other people.

Hmm, ok. I think we should only set them if `horiz` is 1680 then. That should (?) be a reliable way to detect the model.

stand/efi/loader/framebuffer.c
```diff
@@ -387,6 +388,23 @@ efifb_from_uga(struct efi_fb *efifb)
 			/* These are the missing bits. */
 			offset = 0x0;
 			stride = 2048;
+			printf("Model found: %s\n", ev);
+		} else if (ev != NULL && !strcmp(ev, "MacBookPro3,1")) {
+			/*
+			 * Valid for MacBookPro 17" with standard resolution.
+			 * Other Models are:
+			 *   MaBookPro 15" with horiz=1440
+			 *   MaBookPro 17" with horiz=1920
```
VexedUXR, line 397: Shouldn't horiz be 1680 on the 17" model?

frarohi: Yes, 1680 is the standard horiz resolution. There is another variant with 1920.
Good idea to restrict the offset and stride settings to the 1680 variant.

Would you include the MacBookpro3,1 settings into your framebuffer.c change?
Then I would just close this merge request.

Regards, Frank

VexedUXR: >Yes, 1680 is the standard horiz resolution. There is another variant with 1920.

Ah, I see.

>Would you include the MacBookpro3,1 settings into your framebuffer.c change?

Yes, this LGTM. I'll merge shortly.

VexedUXR:
Rewrote the commit message and merged. Sorry again that this sat here for so long.

Thanks!

frarohi:
Hi,
tested on MacbookPro3,1 horiz 1680 
works like a charm!
Thank you very much for your work,
I appreciate it.

PR #1588, opened 2025-02-02
exec_copyin_args: remove the parameter 'segflg'

wy-chung:
The argument 'segflg' passed to exec_copyin_args are always UIO_USERSPACE so this argument can be removed.

### DIFF
--- a/sys/kern/kern_exec.c
+++ b/sys/kern/kern_exec.c
@@ -225,8 +225,7 @@ sys_execve(struct thread *td, struct execve_args *uap)
 	error = pre_execve(td, &oldvmspace);
 	if (error != 0)
 		return (error);
-	error = exec_copyin_args(&args, uap->fname, UIO_USERSPACE,
-	    uap->argv, uap->envv);
+	error = exec_copyin_args(&args, uap->fname, uap->argv, uap->envv);
 	if (error == 0)
 		error = kern_execve(td, &args, NULL, oldvmspace);
 	post_execve(td, error, oldvmspace);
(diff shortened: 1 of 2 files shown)


brooksdavis:
A cursory read of the diff shows the premise is false. (Line 254 of kern_exec.c).

wy-chung:
The calling path sys_fexecve() -> exec_copyin_args() -> exec_args_add_fname() shows that when passing 'fname' as NULL to exec_copyin_args, the 'segflg' will not be used in exec_args_add_fname. That is the 'segflg' is a don't care parameter for exec_copyin_args when the 'fname' is NULL. Since it is a don't care, it can be replaced with UIO_USERSPACE. So all the functions that call exec_copyin_args have passed 'segflg' as UIO_USERSPACE.

brooksdavis:
> The calling path sys_fexecve() -> exec_copyin_args() -> exec_args_add_fname() shows that when passing 'fname' as NULL to exec_copyin_args, the 'segflg' will not be used in exec_args_add_fname. That is the 'segflg' is a don't care parameter for exec_copyin_args when the 'fname' is NULL. Since it is a don't care, it can be replaced with UIO_USERSPACE. So all the functions that call exec_copyin_args have passed 'segflg' as UIO_USERSPACE.

This would have been excellent analysis to include in the commit rather than hoping people would puzzle it out.

I would recommend splitting this change into two commits. One which removes the use of UIO_SYSSPACE with a commit message like this one (there's some argument that UIO_NOCOPY should be used here, but code doesn't handle it as a rule) and another which removes the argument. Make sure to make the same changes to freebsd32_exec_copyin_args and linux_exec_copyin_args.

That being said, is there some larger plan behind this change?

wy-chung:
I did make a mistake for this pull request. I missed the case in sys_fexecve. Out of curisity I retrace the source code again and find that there is really a case here.

wy-chung:
I will close this pull request and open another one.

wy-chung:
BTW, there is no larger plan behind this change.

PR #1590, opened 2025-02-05
Remove parameter 'segflg' from the function *exec_copyin_args

wy-chung:
In kernel "copyin" means copy data from user address space to kernel address space. But in the function *exec_copyin_args there is a parameter 'segflg' that is used to specify the address space of the parameter 'fname'. In the source code there are two places where 'segflg' are not UIO_USERSPACE. In both cases the 'fname' argument are NULL so the argument 'segflg' are not important there. So it is safe to remove the parameter 'segflg' from the function *exec_copyin_args.

### DIFF
--- a/sys/compat/freebsd32/freebsd32_misc.c
+++ b/sys/compat/freebsd32/freebsd32_misc.c
@@ -401,7 +401,7 @@ freebsd32_sigaltstack(struct thread *td,
  */
 int
 freebsd32_exec_copyin_args(struct image_args *args, const char *fname,
-    enum uio_seg segflg, uint32_t *argv, uint32_t *envv)
+    uint32_t *argv, uint32_t *envv)
 {
 	char *argp, *envp;
 	uint32_t *p32, arg;
@@ -422,7 +422,7 @@ freebsd32_exec_copyin_args(struct image_args *args, const char *fname,
 	/*
 	 * Copy the file name.
 	 */
-	error = exec_args_add_fname(args, fname, segflg);
+	error = exec_args_add_fname(args, fname, UIO_USERSPACE);
 	if (error != 0)
 		goto err_exit;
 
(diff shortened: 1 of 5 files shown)


markjdb (commented):
This looks ok to me.

I don't think it makes sense to have the extra commit to wrap a long line - this can be done in the main commit since you are modifying that line already.

bsdjhb (approved):
Hmmm, we do have a UIO_SYSSPACE `fname` when starting the init process (pid 1), but I think there we don't call exec_copyin_args() but construct the args directly.  Hmm, seems even before that change (commit 3a325dec32eba3ae6e9e4a7ec91df52ba5fe16d7) we were still getting the `fname` from userspace in that case.

I think this is fine.

wy-chung:
> This looks ok to me.
> 
> I don't think it makes sense to have the extra commit to wrap a long line - this can be done in the main commit since you are modifying that line already.

I send the pull request and then I receive a email from the github. It says that a line is too long so I send a second commit. I don't know how to change the oroginal pull request. This is the only way I know how to fix this error.

markjdb:
> > This looks ok to me.
> > I don't think it makes sense to have the extra commit to wrap a long line - this can be done in the main commit since you are modifying that line already.
> 
> I send the pull request and then I receive a email from the github. It says that a line is too long so I send a second commit. I don't know how to change the oroginal pull request. This is the only way I know how to fix this error.

To modify the most recent commit, you can stage changes as normal and run `git commit --amend`. More generally, you can use `git rebase` to "squash" commits together into one. Then, force-push your branch to update this pull request.

Using `git rebase` to do this requires some practice. I can squash the commits for this PR if you like, but for next time, it's worth learning how to amend and squash commits.

lwhsu:
@wy-chung why close this?

wy-chung:
It's been there for a long time. I think maybe nobody think this none-bug issue is important so I closed it.
This issue is more like a typo not a bug. The function 'freebsd32_exec_copyin_args' already implis copying from user space so the parameter 'segflg' is redundant.

markjdb:
Merged, thanks. Sorry it took a while to get this landed.

bsdimp:
Landed

PR #1591, opened 2025-02-05
if_infininband: Support BPF write for broadcast frames

ndance-vdura:
This change is predominantly focused at being able to support DHCP for IPoIB network interfaces utilizing the if_infiniband.c driver.

To do this, DHCP clients and servers require the ability to transmit link-layer broadcasts on the IB interfaces using BPF. In the current driver implementation, this isn't possible as frames transmitted via BPF become mangled (whether using IB or Ethernet headers).

This change updates the if_infiniband driver to be capable of accepting link-layer broadcast requests via BPF using Ethernet formatted frames as this honors the DLT_EN10MB type the driver uses to bind to BPF. Only Broadcast frames can reliably be interpreted using the Ethernet header format so unicast and multicast frames are rejected if passed in using the Ethernet header format. This is, however, sufficient to support the primary DHCP use case this commit aims to address. The ability to support unicast, broadcast or multicast frames should not be impacted if native infiniband header support is added to BPF at a later date.

### DIFF
--- a/sys/net/if_infiniband.c
+++ b/sys/net/if_infiniband.c
@@ -149,6 +149,43 @@ infiniband_bpf_mtap(struct ifnet *ifp, struct mbuf *mb)
 	mb->m_pkthdr.len += sizeof(*ibh);
 }
 
+/*
+ * For clients using BPF to send broadcasts.
+ *
+ * This driver binds to BPF as an EN10MB (Ethernet) device type. As such, it is
+ * expected BPF and BPF users will send frames with Ethernet headers, which
+ * we'll do our best to handle. We can't resolve non-native unicast or multicast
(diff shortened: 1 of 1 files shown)


rpokala-freebsd:
@glebius and @chmeeedalf: When I looked around the IB / IP-over-IB code, it looks like you two have made the only non-trivial changes since we lost Hans Petter.

This change looks okay to us within VDura (formerly known as Panasas), but I was hoping to get some expert eyes on it, rather than just committing it on my own. Can either of you take a look?

glebius:
I'm not expert. All my changes to infiniband are related to large network stack sweeping cleanups. I never touched infiniband hardware. If you use it in VDura, you probably should consider becoming maintainers :)

PR #1592, opened 2025-02-06
linprocfs: Continue on error when reading process limits

ricardobranco777:
Continue on error when reading process limits.

Fixes https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=283726

Most text utilities with capsicum(4) enabled fail when reading `/compat/linux/proc/*/limits` because both `LINUX_RLIMIT_SIGPENDING`  & `LINUX_RLIMIT_MSGQUEUE` call `kernel_sysctlbyname`.

The alternative would be to make the variables accessed by the `kern.sigqueue.max_pending_per_proc` & `kern.ipc.msgmnb` sysctl's public.  This was done for reading /proc/cpuinfo in https://github.com/freebsd/freebsd-src/commit/d74a742704eb81f0c6f4aa83e4cb0de26a81c400

### DIFF
--- a/sys/compat/linprocfs/linprocfs.c
+++ b/sys/compat/linprocfs/linprocfs.c
@@ -1915,7 +1915,7 @@ linprocfs_doproclimits(PFS_FILL_ARGS)
 			    "kern.sigqueue.max_pending_per_proc",
 			    &res, &size, 0, 0, 0, 0);
 			if (error != 0)
-				goto out;
+				continue;
 			rl.rlim_cur = res;
 			rl.rlim_max = res;
 			break;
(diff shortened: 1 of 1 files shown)


ricardobranco777:
@bsdimp can we merge this?  I've been using successfully without issues.

bsdimp (approved):

bsdimp:
@lemul What do you think? I think this is good, but a second set of eyes would be good.

lemul:
Well, I'm not object it, but what sense to hide permission error from the user?

ricardobranco777:
> Well, I'm not object it, but what sense to hide permission error from the user?

The intention is not to hide an error but to fetch the rest of the information because the sysctl fails with Capsicum enabled.  We can either ignore the error or make the variables public to avoid making a sysctl call like it was done for cpuinfo here:

https://github.com/freebsd/freebsd-src/commit/d74a742704eb81f0c6f4aa83e4cb0de26a81c400

ricardobranco777:
ping

bsdimp:
Ooops. Missed the exchange back in June.
I'll land this now. In the future, the commit message should have "PR:  <bug-number>" in it.
As well as an explanation for why, as you've provided in the comments.
Since it was my bad to miss this in June, I've done those things and will push this in shortly.

bsdimp:

PR #1593, opened 2025-02-07
pci(4): Always get MSI-X count from config register

kgalazka:
MSI-X vectors count of a device may change after it was attached. This happens e.g. when an Ethernet driver assigns number of Rx/Tx queues, which was selected by an user, to a Virtual Function. In such case cached value returned by pci_msix_count could be incorrect. Instead always get the current number from the config register.

### DIFF
--- a/sys/dev/pci/pci.c
+++ b/sys/dev/pci/pci.c
@@ -2172,9 +2172,13 @@ pci_msix_count_method(device_t dev, device_t child)
 	struct pci_devinfo *dinfo = device_get_ivars(child);
 	struct pcicfg_msix *msix = &dinfo->cfg.msix;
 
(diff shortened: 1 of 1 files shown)


bsdjhb (commented):
Oh, I just uploaded a version of this for review this morning as well, but I avoid caching until the alloc methods are called, and I also do it for plain MSI as well as MSI-X.  Review is at https://reviews.freebsd.org/D48890.

kgalazka:
> Oh, I just uploaded a version of this for review this morning as well, but I avoid caching until the alloc methods are called, and I also do it for plain MSI as well as MSI-X. Review is at https://reviews.freebsd.org/D48890.

It looks much better than mine, thanks!

PR #1594, opened 2025-02-07
bsdinstall: fix ps3 regression

aomsin2526:
b6644f5 introduces FREEBSD_BOOTLABEL="FreeBSD" as default value but cc42ef5 removed this line, expecting it to be set by update_uefi_bootentry()

But ps3 don't use this function so its broken completely.

So we add this line back.

Example /boot/etc/kboot.conf

Bad:

`default=`
`='/kernel/kernel kernelname=/boot/kernel/kernel vfs.root.mountfrom=ufs:/dev/da1s2a'`

Good:

`default=FreeBSD`
`FreeBSD='/kernel/kernel kernelname=/boot/kernel/kernel vfs.root.mountfrom=ufs:/dev/da1s2a'`


aomsin2526:
> Why not using:
> 
> ```shell
> FREEBSD_BOOTLABEL=$OSNAME
> ```
> 
> But after sourcing `$BSDCFG_SHARE/common.subr`:
> 
> https://github.com/freebsd/freebsd-src/blob/2be8b121bf82d409773ee5b404e4e678506beffc/usr.sbin/bsdinstall/scripts/bootconfig#L31

Fixed thanks!

jlduran:
That should work right?

Aesthetically, I would have placed it closer to `TMPDIR`, but that is just nitpicking.

jlduran (approved):

aomsin2526:
I think it should work. I tested my original code and it work fine, I will test this new one tomorrow.

bsdimp:

PR #1595, opened 2025-02-07
Reduce pollution in stddef.h and stdarg.h

brooksdavis:
This set of changes aims to decrease the number of things declared in stddef.h and stdarg.h.

The motivation is that software shouldn't be depending on incidental definitions of standard headers (someone recently found that some headers depended on __BEGIN_DECLS leaking in which caused an issue with stddef.h from clang.)

On the whole I've taken the view that it's OK to use bare `__builtin_` functions when our cdefs.h abstraction is unconditional.  I don't propose a sweeping migration to them, but IMO the hygiene benefits are worth it here.

The more impactful changes will require an exp-run and we may not end up wanting them all, but I've put them up together for ease of commenting.


bsdimp:
Love visible splitting off. Many others call it "features" but I'm agnostic. Will look closely later.

bsdimp:
I generally like this, and the bits I looked at looked fine, but it's too huge to sanely review.

bsdimp:
All the structural changes look good.

brooksdavis:
I've requested an exp-run for these changes https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=286274

bsdimp:
Since an exp-run is in progress, I'm going to close this pull request as OBE. You know the drill: if you need to, reopen / file a new one to get the QA run.

PR #1597, opened 2025-02-12
wg.4: document kernel config option

llfw:
wg(4) can be compiled into the kernel (device wg), but the wg.4 manpage does not document this.  adjust it to mention this like other drivers do.

### DIFF
--- a/share/man/man4/wg.4
+++ b/share/man/man4/wg.4
@@ -23,14 +23,21 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd June 12, 2023
+.Dd February 12, 2025
 .Dt WG 4
 .Os
 .Sh NAME
 .Nm wg
 .Nd "WireGuard protocol driver"
 .Sh SYNOPSIS
-To load the driver as a module at boot time, place the following line in
+To compile this driver into the kernel,
+place the following lines in your kernel configuration file:
+.Bd -ragged -offset indent
+.Cd "device wg"
+.Ed
+.Pp
+Alternatively, to load the driver as a module at boot time,
+place the following line in
 .Xr loader.conf 5 :
 .Bd -literal -offset indent
 if_wg_load="YES"


concussious (approved):
Can wg be loaded with kld_list? We're recommending rc.conf lately for things that don't need to be at early boot.

I really adore the style of vt(4). You might be interested in what I did here for another if_module [^0].

Could we doing something like that here? I think this style is more consistent with how synopsis is in other sections of the manual and other BSDs.
[^0]: https://reviews.freebsd.org/D48905

llfw:
i don't have a strong preference for wording here, but i was following the example of other drivers, e.g. `ix(4)`, `cxgbe(4)`, `epair(4)` and `tun(4)`, which all have the same wording i've used here referencing `loader.conf`.

if we want to recommend using `kld_list` instead, then i think we need to:

- decide if this should be recommended for all network interfaces (e.g. `ix`, `cxl`, ...) or only software interfaces (`wg`, `epair`, ...)
- agree on a standard wording for the SYNOPSIS section
- update all the affected manpages at once

since hardware drivers can sometimes be required before the root filesystem is mounted (e.g., for NFS root) my feeling is we should only recommend `kld_list` for the software drivers.  

alternatively, we would need to document both `loader.conf` and `kld_list` options for the hardware drivers, which i'm not opposed to, but it would make the synopsis quite wordy.

does that sound reasonable?  what do you think?

llfw:
possible sample wording based on `vt(4)`:

```
NAME
     wg – WireGuard protocol driver

SYNOPSIS
    device wg

    In rc.conf(5):
    kld_list="${kld_list} if_wg"

    In loader.conf(5):
    if_wg_load="YES"
```

although i like how concise this is, i'm a bit concerned that a new FreeBSD user will type `man 4 wg` and not know what `device wg` means or where it should go, especially since the vast majority of users do not compile a custom kernel nowadays.

concussious:
The way you have written it is good by me. If @sergio-carlavilla and/or @mhorne consent, I can commit this as is. It follows what we've generally been doing the last decade or so perfectly.

I am jealous of the other BSDs elegant synopsis style, and think vt(4) sets the example to achieve this elegance for FreeBSD.

I think explaining what "device wg" means should be part of a new intro(4). The other BSDs just have a kernel configuration line in synopsis with no explanation for all drivers, as do commands in FreeBSD.

As for loader/rc.conf, I really like that both because it's showing the ways it can be used, but I think it would need to go in intro(4) something about "kernel interfaces should be enabled via one of three methods". The redundancy could easily confuse people.

concussious:
> I think it would need to go in intro(4)

This kind of stuff historically was in section 0, and I heard there are fundamental design issues which make bringing back section 0 unrealistic.

llfw:
i closed this by accident, but it looks like it's already been merged, so i'll leave it closed unless you want to reopen it for some reason.

PR #1600, opened 2025-02-18
Separate console from starting tty

sgharms:


concussious:
Don't we have a sysctl for this?

sgharms:
@concussious I would have thought that something like this should certainly be configurable through `/boot/loader.conf` on the theory that it's happening around the same point as the `boot_mute` overlay or `boot_verbose`, but this seems to be locked away in the `#define` magics. I'd be elated to find otherwise though!

bsdimp:
So I like this concept. However, I have some problems with the implementation. A kernel config option likely isn't the best way to do this.

But it also brings up a much bigger issue. There's no way to set the console in any meaningfully useful way that's universal. If I want a serial console there's several different parameters I might set in the boot loader that will result in the boot loader telling the kernel, through a variable that could be overloaded but is obscure, how to do that. There's similar ways for video console. But there's no really good, universal way to specify a specific console to use. There's several serial ports available, sometimes multiple screens or video devices or in this instance, a particular sub-device of the main screen.

So, the standard feedback I'd normally give on this is 'make it a sysctl / tuneable using this or that method' would make the situation worse. It would add another weird knob on the side to configure things that may or may not interact well with the other weird knobs we have. This is an interesting idea (I'd never do it on my systems, but I can see how others might want to w/o going the full splash screen approach), but it points out a bigger picture problem.

Put another way, is there some variable that we can set in the boot loader that would cut through the mess and tell the kernel exactly what device(s) to use for its console? Is there a good syntax for this? Are there ways we can have good "DWIM" defaults that take into account things like the EFI ConIn variable, the SPCR table, the DBG2 table and whatever else and do it in a space efficient way? Can we unify some of the "purpose built" things for communicating other attributes of the console that we currently do in an ad-hoc manner?  Is there some way to build this "new thing" without having to have all the "old things" converted over instantly? What's a good, thoughtful, extensible design that would allow us to break free from the bitfield interface we inherited from 386BSD 30+ years ago?

sgharms:
Well, I'll certainly not claim to have the ability to describe the one, clean interface to rule them all, but I _fully_ agree that mangling this through kernel definitions and rebuilding seems...excessively complicated. `loader.conf` seems like the right place to me, personally.

I also want to put a quick caveat here, the splash screen (`boot_mute="YES"`) ***does not actually obscure diagnostic messages to login***. It blocks _kernel_ messages, but then the user gets a snootful of RC script noise. FWIW, I consider making `boot_mute` leave a splash screen until login to be its rightful behavior and , therefore, as of 15.0-CURRENT, it is broken.

sys/dev/vt/vt.h
bsdimp, line 70: None of this should be in vt.h. The override can be done in vt_core.c.
We try hard not to have various .h files in the kernel dependent on opt_*.h being included correctly.

bsdimp, line 417: This can be static to vt_core.c

sys/dev/vt/vt_core.c
bsdimp, line 59: this then isn't needed.

sgharms:
@bsdimp You asked a number of good questions in your last comment. Obviously we're now in 15.1-RELEASE land now, so this wasn't/isn't going to land there :).

But the fundamental question I think is phrased:

> is there some variable that we can set in the boot loader that ... tell the kernel exactly what device(s) to use for its console? 

I don't know but this idea's simplicity gives it the ring of correctness. Being able to give directives like this seems like a good idea. You closed with a number of sage questions, but I have to ask (speaking of things from the 90's like 386BSD), in the words of Pearl Jam, "Who answers? Who answers?" Nearly 18 months later, we're still unresolved.

This problem is still there. FreeBSD is doing wonderful things for laptop access and GPU support, but here, at this moment that the whole system starts, we can't find a path in to make a simple thing manifest.

PR #1609, opened 2025-03-05
jail: add JID, JNAME and JPATH to environment for exec.* commands

Defenso-QTH:
Although variable substitution is available in the jail configuration file, the jail identifier is often not since it is dynamically attributed at run time.

In order to facilitate scripting of exec.* commands executed on the system, this change sets the JID, JNAME and JPATH environment variables.

These variables are not added when using exec.clean. Neither are they for commands executed inside jails, to avoid disclosing information about the host system.

### DIFF
--- a/usr.sbin/jail/jail.8
+++ b/usr.sbin/jail/jail.8
@@ -819,6 +819,22 @@ commands in sequence.
 All commands must succeed (return a zero exit status), or the jail will
 not be created or removed, as appropriate.
 .Pp
+The following variables are added to the environment:
+.Bl -tag -width indent -offset indent
+.It Ev JID
+The
+.Va jid
+, or jail identifier.
+.It Ev JNAME
+The
+.Va name
+of the jail.
+.It Ev JPATH
+The
+.Va path
+of the jail.
+.El
+.Pp
 The pseudo-parameters are:
 .Bl -tag -width indent
 .It Va exec.prepare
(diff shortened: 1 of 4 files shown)


Crest:
You can write this to export the jail id and jail name to the shell script, but the quoting gets ugly:

```
exec.start = "
    export JID='$jid' JNAME='$name'
    echo \"Hello \$JNAME from \$JID.\"
";
```

Defenso-QTH:
That could work. It seems my issue comes from using `iocage`, which does not rely on `jail` to exec the `prestart`/`poststart` commands but does it itself, so we don't get replacement, and it does not set environment variables either.

https://github.com/freebsd/iocage/blob/c55e05aebb305b4013554ec80298113c50d1c74c/iocage_lib/ioc_start.py#L873
```python
        # Running exec_poststart now
        poststart_success, poststart_error = \
            iocage_lib.ioc_common.runscript(
                exec_poststart
            )
```

Maybe it makes more sense patching `iocage` then.

ImAwsumm (approved):

usr.sbin/jail/jail.8
```diff
@@ -819,6 +819,22 @@ commands in sequence.
 All commands must succeed (return a zero exit status), or the jail will
 not be created or removed, as appropriate.
 .Pp
+The following variables are added to the environment:
+.Bl -tag -width indent -offset indent
+.It Ev JID
+The
+.Va jid
+, or jail identifier.
```
bsdimp, line 827: The , here should be on the end of the prior line.

kevans91:
JFYI- I have staged some follow-up commits here:
 - https://reviews.freebsd.org/D51501 ("jail: tests: cleanup the commands test a bit")
 - https://reviews.freebsd.org/D51502 ("jail: consistently populate the KP_JID and KP_NAME parameters")

The latter fixes a new crash introduced here, but it's kinda-sorta a pre-existing issue.

PR #1610, opened 2025-03-10
procctl.2: Fix names of PROC_LOGSIGEXIT_CTL constants

endrift:
The headers contain constants that start with PROC_LOGSIGEXIT_CTL_*, but the man page elided the _CTL portion.


jlduran (approved):
Fixes: dabf006a638f ("Add per-process flag to disable logsigexit")

bsdimp:
Unsure why this wasn't autoclosed, but close it now.

PR #1613, opened 2025-03-13
Memory leak and minor security fixes in mdo.c

thesynthax:
Bug 1: Potential Out-of-Bounds Read in getgrouplist()
Fix: Check return value and validate ngroups

Bug 2: Memory leak in groups
Fix: free(groups);

Bug 3: getenv() may be unsafe
Fix: Use secure_getenv()

### DIFF
--- a/usr.bin/mdo/mdo.c
+++ b/usr.bin/mdo/mdo.c
@@ -80,20 +80,24 @@ main(int argc, char **argv)
 		if (groups == NULL)
 			err(EXIT_FAILURE, "cannot allocate memory for groups");
 
-		getgrouplist(pw->pw_name, pw->pw_gid, groups, &ngroups);
+		if (getgrouplist(pw->pw_name, pw->pw_gid, groups, &ngroups) == -1) {
+            		err(EXIT_FAILURE, "failed to get group list for user %s", username);
+        	}
 
 		wcred.sc_gid = wcred.sc_rgid = wcred.sc_svgid = pw->pw_gid;
 		wcred.sc_supp_groups = groups + 1;
 		wcred.sc_supp_groups_nb = ngroups - 1;
 		setcred_flags |= SETCREDF_GID | SETCREDF_RGID | SETCREDF_SVGID |
 		    SETCREDF_SUPP_GROUPS;
-	}
+
+ 	       free(groups);
+    }
 
 	if (setcred(setcred_flags, &wcred, sizeof(wcred)) != 0)
 		err(EXIT_FAILURE, "calling setcred() failed");
 
 	if (*argv == NULL) {
-		const char *sh = getenv("SHELL");
+        	const char *sh = secure_getenv("SHELL");
 		if (sh == NULL)
 			sh = _PATH_BSHELL;
 		execlp(sh, sh, "-i", NULL);


OlCe2:
Hi Kushagra,

Thanks for your interest!

> Bug 1: Potential Out-of-Bounds Read in getgrouplist() Fix: Check return value and validate ngroups

There is in fact no practical bug here, but indeed we can see a conceptual one, and I agree we should change the code so that it feels correct, although probably not as you suggested.

Let me first recall some facts:
1. The maximum number of groups cannot change after boot (it is only a tunable).
2. The groups array is sized to be one more than the maximum number of groups (including the primary group, hence the +2).
3. It may be that getgrouplist() wants to return even more groups (in such cases, it returns -1), but in any case it will have filled the available slots and the number of supplementary groups.
4. setcred() bails out if the number of supplementary groups exceeds the configured number, before trying to copy these into kernel space (which would cause an out-of-bounds read, without catastrophic consequences).

The reason I wrote this code like that is to have `setcred()` itself reject the credentials change in case there are too many groups (see the herald comment in the relevant code block), with the intent that at some point such attempts will be logged somewhere by the kernel.  We can certainly consider changing that as part of devising how we should log attempts/failures from mdo(1).  In the meantime, though, I'd prefer to keep this invariant.

The most disturbing part here I think is that we pass to `setcred()` a number of groups that doesn't necessarily correspond to the number of slots filled in `groups` (iff `getgrouplist()` returned -1).

So I think this should be fixed, but rather by testing if `getgrouplist()` returned -1, and if that's the case just adjusting `ngroups` accordingly (and continuing up to the `setcred()` call).

I'll then probably expand the comment as I don't think it correctly conveys what I've just explained here.

> Bug 2: Memory leak in groups Fix: free(groups);

There is no `free()` indeed, but given that what mdo(1) then immediately does is call `exec()` or exit on error, there is no real leak.  In such contexts, it is even a quite common pattern, because calling cleanup functions there can be seen as a waste of time as the process image is going to be entirely thrown out (and then rebuilt in the case of `exec()`).

I do not object to add a matching call to `free()`.  However, its position in your proposal is wrong and would trigger a use-after-free, it has to be moved after the call to `setcred()`.

> Bug 3: getenv() may be unsafe Fix: Use secure_getenv()

mdo(1) is not installed setuid, so using `secure_getenv()` makes no practical difference.  Additionally, shell restriction is only meaningful if the program intends to run scripts (because, to do so, it has to run a shell, and controlling which shell is run can completely hijack the program's intent), but here, we give that shell back to the user anyway, with which he can launch any command he/she wants (including another shell or program).  So I don't see how this change can make sense and would just drop it.

usr.bin/mdo/mdo.c
```diff
@@ -80,20 +80,24 @@ main(int argc, char **argv)
 		if (groups == NULL)
 			err(EXIT_FAILURE, "cannot allocate memory for groups");
 
-		getgrouplist(pw->pw_name, pw->pw_gid, groups, &ngroups);
+		if (getgrouplist(pw->pw_name, pw->pw_gid, groups, &ngroups) == -1) {
```
OlCe2, line 83: Please see my main comment.

usr.bin/mdo/mdo.c
```diff
@@ -80,20 +80,24 @@ main(int argc, char **argv)
...
+        	}
 
 		wcred.sc_gid = wcred.sc_rgid = wcred.sc_svgid = pw->pw_gid;
 		wcred.sc_supp_groups = groups + 1;
 		wcred.sc_supp_groups_nb = ngroups - 1;
 		setcred_flags |= SETCREDF_GID | SETCREDF_RGID | SETCREDF_SVGID |
 		    SETCREDF_SUPP_GROUPS;
-	}
+
+ 	       free(groups);
```
OlCe2, line 93: Please see my main comment.

usr.bin/mdo/mdo.c
```diff
@@ -80,20 +80,24 @@ main(int argc, char **argv)
...
+    }
 
 	if (setcred(setcred_flags, &wcred, sizeof(wcred)) != 0)
 		err(EXIT_FAILURE, "calling setcred() failed");
 
 	if (*argv == NULL) {
-		const char *sh = getenv("SHELL");
```
OlCe2, line 96: Please see my main comment.

OlCe2 (changes requested):
See my main comment above (should have been put in this GitHub's review; I'm just familiarizing myself with these).

ImAwsumm (approved):

llfw:
per review, the change from `getenv` to `secure_getenv` is useless, please fix this.

can you also confirm you create these code changes yourself, and didn't use a LLM (“AI”)?

PR #1615, opened 2025-03-14
igmp: update IGMP version based on net.inet.igmp.default_version

etihwo:
IGMP requires hosts to use the lowest version they have seen on the network. When net.inet.igmp.default_version is changed, we do not change IGMP version used by interface even if the interface uses an higher version than version set. So we could send IGMPv3 even if the user has requested IGMPv2 or IGMPv1 via the sysctl.

Change IGMP version for each interface when V_igmp_default_version is smaller than the version used by the interface

Signed-off-by: Olivier BLANC etihwo@outlook.com

### DIFF
--- a/sys/netinet/igmp.c
+++ b/sys/netinet/igmp.c
@@ -404,6 +404,7 @@ sysctl_igmp_default_version(SYSCTL_HANDLER_ARGS)
 {
 	int	 error;
 	int	 new;
+	struct igmp_ifsoftc *igi;
 
 	error = sysctl_wire_old_buffer(req, sizeof(int));
 	if (error)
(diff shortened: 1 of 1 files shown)


ImAwsumm (approved):

etihwo:
@kprovost , any opinion on the proposed change ?

kprovost:
igmp_set_version() (indirectly, via igmp_v3_cancel_link_timers()) expects to be run within NET_EPOCH and with the IN_MULTI_LIST_LOCK held.
This explodes as soon as it's run on a WITNESS/INVARIANTS-enabled kernel.

On the bright side, I can reproduce the problem this tries to fix, and absent the locking issues it does fix it.

kprovost:
I've pushed my test case and a reworked version of your patch here: https://github.com/kprovost/freebsd-src/tree/pr-1615

It reads V_igmp_default_version outside of IGMP_LOCK, which isn't ideal, but does fix another locking problem. IGMP_LOCK is a sleepable lock, and sysctl_handle_int() may sleep (when copying from userspace).

kprovost:
https://cgit.freebsd.org/src/commit/?id=d2c2d6d6b09e8a059dffd6b6c7513d0925a8753d

PR #1618, opened 2025-03-20
netinet6: allow binding to anycast addresses

llfw:
the restriction on sending packets from anycast source addresses was removed in RFC4291, so there's no reason to forbid binding to such addresses.  this allows anycast services (e.g., DNS) to actually use anycast addresses, which was previously impossible.

RFC4291 also removes the restriction that only routers may configure anycast addresses; this was never enforced in code but was documented in ifconfig.8.  update ifconfig.8 to document both changes.


ImAwsumm (approved):

ImAwsumm (approved):

llfw:
closing this as i intend to commit it myself: https://reviews.freebsd.org/D49905

PR #1621, opened 2025-03-23
rc.subr: add ${svc}_svcj_ipaddrs option

llfw:
setting ${svc}_svcj_ipaddrs to a list of IP addresses causes rc.subr to set ip6.addr and/or ip4.addr when starting the jail, restricting it to those IP addresses rather than inheriting all IP addresses.

for example:

```
inetd_enable=YES
inetd_svcj=YES
inetd_svcj_options="net_basic"
inetd_svcj_ipaddrs="::1 127.0.0.1 2001:db8::1"
```

if not specified, the default value is unchanged (inherit all addresses if networking is enabled).


ImAwsumm (approved):

netchild:
Merged to FreeBSD.

PR #1622, opened 2025-03-26
ice.4: Update man page with feature documentation

ricera:
This is eventually supposed to document the functionality from #1573, but for now it contains information ported from the out-of-tree driver's man pages in order to bring the current man page up to date with features that are already in the upstream driver.

### DIFF
--- a/share/man/man4/ice.4
+++ b/share/man/man4/ice.4
@@ -73,6 +73,16 @@ Intel\(rg Ethernet Connection E822\-L
 Intel\(rg Ethernet Connection E823\-C
 .It
 Intel\(rg Ethernet Connection E823\-L
+.It
+Intel\(rg Ethernet Connection E825\-C
+.It
+Intel\(rg Ethernet Connection E830\-C
+.It
+Intel\(rg Ethernet Connection E830\-CC
+.It
+Intel\(rg Ethernet Connection E830\-L
+.It
+Intel\(rg Ethernet Connection E830\-XXV
 .El
 .Pp
 For questions related to hardware requirements, refer to the documentation
@@ -182,14 +192,38 @@ See the
 .Xr iflib 4
 man page for more information on using iflib sysctl variables as tunables.
 .Bl -tag -width indent
-.It Va hw.ice.#.enable_health_events
-TBW
-.It Va hw.ice.#.debug.enable_tx_fc_filter
-TBW
-.It Va hw.ice.#.debug.enable_tx_lldp_filter
-TBW
-.It Va hw.ice.#.debug.enable_health_events
-TBW
+.It Va hw.ice.enable_health_events
+Set to 1 to enable firmware health event reporting across all devices. Enabled by default.
+.Pp
+If enabled, when the driver receives a firmware health event message, it will
+print out a description of the event to the kernel message buffer and if
+applicable, possible actions to take to remedy it.
+.It Va hw.ice.irdma
+Set to 1 to enable the RDMA client interface, required by the
+.Xr irdma 4
+driver.
+Enabled by default.
+.It Va hw.ice.rdma_max_msix
+Set the maximum number of per-device MSI-X vectors that are allocated for use by the
+.Xr irdma 4
+driver.
+Set to 64 by default.
+.It Va hw.ice.debug.enable_tx_fc_filter
+Set to 1 to enable the TX Flow Control filter across all devices. Enabled by default.
+.Pp
+If enabled, the hardware will drop any transmitted Ethertype 0x8808 control
+frames that do not originate from the hardware.
+.It Va hw.ice.debug.enable_tx_lldp_filter
+Set to 1 to enable the TX LLDP filter across all devices. Enabled by default.
+.Pp
+If enabled, the hardware will drop any transmitted Ethertype 0x88cc LLDP frames
+that do not originate from the hardware.
+This must be disabled in order to use LLDP daemon software such as
+.Xr lldpd 8 .
+.It Va hw.ice.debug.ice_tx_balance_en
+Set to 1 to allow the driver to use the 5-layer Tx Scheduler tree topology if configured by the DDP package.
+.Pp
+Enabled by default. 
 .El
 .Sh SYSCTL PROCEDURES
 .Bl -tag -width indent
--- a/share/man/man4/ice.4
+++ b/share/man/man4/ice.4
@@ -167,7 +167,7 @@ This mirroring may impact performance.
 .Pp
 To use RDMA monitoring, you may need to reserve more MSI\-X interrupts.
 Before the
-.Nm ice
+.Nm
 driver loads, configure the following tunable:
 .Bd -literal -offset indent
 dev.ice.0.iflib.use_extra_msix_vectors=4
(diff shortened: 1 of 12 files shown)


concussious:
Not all heroes wear capes.

ricera:
I have some additional concerns and questions that I hope someone can answer:

 * Should "information on enabling VLANs, see the README" be removed or add
   another link to ifconfig(8)? Or the VLAN man page? The README isn't shipped
   upstream, so I don't think it's sensible to direct users to the README. Plus,
   basic vlan features aren't very complicated.

 * I don't know if ".Em" is the correct way to highlight the names of the
   utilities like below:
``` .Em Ethernet Port Configuration Tool - FreeBSD ```

 * Not sure how to format "NOTE:" lines; for now I just left their formatting as standard text.

 * I don't think ice_ddp has a man page or maybe should even get one. How do I
   refer to it without using `.Xr`?

 * @kgalazka Need to add 200G modules to the list in the HARDWARE section, but I don't know any of their SKUs or descriptions

 * @kgalazka, for "mirror-src-vsi" iovctl parameter, how do you find the vsi number for the
   PF VSI? I cannot remember -- is it indexed to the driver (so it would be 0) or is it indexed to a value from the FW/HW?

**Comment by concussious** [comment]
> Should "information on enabling VLANs, see the README" be removed or add
another link to ifconfig(8)? Or the VLAN man page? The README isn't shipped
upstream, so I don't think it's sensible to direct users to the README. Plus,
basic vlan features aren't very complicated.

If we don't have the readme, ifconfig sounds good to me.

> I don't know if ".Em" is the correct way to highlight the names of the
utilities like below:
.Em Ethernet Port Configuration Tool - FreeBSD

"If such a stand-alone command is the topic of the current manual page, use Nm; if it is documented in another manual page, use Xr; otherwise, if it is not documented, for example because it is only available in foreign or historical operating systems, use Sy." ~https://mandoc.bsd.lv/mdoc/style/commands.html

But what is Ethernet port configuration tool FreeBSD? If we can get that in ports, it should be something like:
```
.Sy portcfgtoolname
.Pq Pa ports/net/intel_ethernet_conftool
```

> Not sure how to format "NOTE:" lines; for now I just left their formatting as standard text.

That's usually what I do, and I usually say "Note that..." to reinforce that I'm intending it as standard text.

> I don't think ice_ddp has a man page or maybe should even get one. How do I refer to it without using .Xr?

.Sy

Thanks again. This is a big deal. The doc is what sells the users on it, I think.

**Comment by kgalazka** [comment]
image

**Comment by ricera** [comment]
Since #1573 was recently merged, I've rebased this. Any other suggestions or fixes that I should make?

File: share/man/man4/ice.4 [hunk 1]
```diff
@@ -32,18 +32,18 @@
...
 .\" * Other names and brands may be claimed as the property of others.
 .\"
-.Dd May 20, 2024
+.Dd March 28, 2025
 .Dt ICE 4
 .Os
 .Sh NAME
 .Nm ice
-.Nd "Intel Ethernet 800 Series Driver"
+.Nd "Intel\(rg Ethernet 800 Series Driver"
```
concussious, line 40:
```suggestion
.Nd "Intel 800 Series Ethernet driver"
```

share/man/man4/ice.4
```diff
@@ -182,42 +940,62 @@ See the
...
+that do not originate from the hardware.
+This must be disabled in order to use LLDP daemon software such as
+.Xr lldpd 8 .
+.It Va hw.ice.debug.ice_tx_balance_en
+Set to 1 to allow the driver to use the 5-layer Tx Scheduler tree topology if
+configured by the DDP package.
+.Pp
+Enabled by default.
 .El
 .Sh SYSCTL PROCEDURES
```
concussious, line 981:
```suggestion
.Sh SYSCTL VARIABLES
```

share/man/man4/ice.4
```diff
@@ -83,41 +93,757 @@ Selecting an MTU larger than 1500 bytes with the
...
-package.
+To change the behavior of the QSFP28 ports on E810-C adapters, use the Intel
+.Sy Ethernet Port Configuration Tool - FreeBSD .
 .It
-To update the firmware on an adapter, use the Intel Non-Volatile Memory (NVM)
-Update Utility for Intel Network Adapter 800 series; installed by the
-.Em sysutils/intel-nvmupdate-100g
-package.
+To update the firmware on an adapter, use the Intel
+.Sy Non-Volatile Memory (NVM) Update Utility for Intel Ethernet Network Adapters E810 series - FreeBSD
```
concussious, line 840: Is this not in ports anymore?

share/man/man4/ice.4
```diff
@@ -83,41 +93,757 @@ Selecting an MTU larger than 1500 bytes with the
...
-by searching for their names, or by installing certain packages:
+by searching for their names:
 .Bl -bullet
 .It
-To change the behavior of the QSFP28 ports on E810-C adapters, use the
-Intel EPCT (Ethernet Port configuration tool); installed by the
-.Em sysutils/intel-epct
-package.
+To change the behavior of the QSFP28 ports on E810-C adapters, use the Intel
+.Sy Ethernet Port Configuration Tool - FreeBSD .
```
concussious, line 837: Is this not in ports anymore?

share/man/man4/ice.4
```diff
@@ -83,41 +93,757 @@ Selecting an MTU larger than 1500 bytes with the
...
+.Sx Jumbo Frames
+section.
 .Pp
 This driver version supports VLANs.
-For information on enabling VLANs, see the
-.Pa README .
+For information on enabling VLANs, see
+.Xr vlan 4 .
+For additional information on configuring VLANs, see
+.Xr ifconfig 8 Ns 's
```
concussious, line 104:
```suggestion
.Xr ifconfig 8 Ns Ap s
```
(Ap might imply Ns, you'd have to check mandoc -t lint)

share/man/man4/ice.4
```diff
@@ -226,21 +1004,77 @@ a storm condition in the kernel.
...
 .Aq Mt freebsd@intel.com .
 .Sh SEE ALSO
-.Xr arp 4 ,
 .Xr iflib 4 ,
-.Xr netintro 4 ,
-.Xr ng_ether 4 ,
 .Xr vlan 4 ,
-.Xr ifconfig 8
+.Xr ifconfig 8 ,
+.Xr sysctl 8
```
concussious, line 1077: Why removing these?

concussious (approved):
Couple of nits, but let's get this in the tree! After it's in the tree I will refactor it to have HARDWARE and the more consistent style of synopsis I presented at BSDCan.

Again, Eric, this is really just amazing. I believe you are really coming through for the future of FreeBSD in its core use case by writing this.

share/man/man4/ice.4
ricera: Is there a reason for this one? The "Intel(R) Ethernet" is supposed to be consistent with how Intel refers to its current ethernet products, like how you can see here: https://www.intel.com/content/www/us/en/products/details/ethernet.html

share/man/man4/ice.4
ricera: I'm not sure if it is, but @kgalazka has said that Intel will stop maintaining or supporting these ports, so I don't want to refer people to them anymore.

share/man/man4/ice.4
ricera: same as above

share/man/man4/ice.4
ricera: I don't think they're particularly relevant or useful to this driver specifically, but I dunno, I could leave them in.

ricera: e.g. you're not going to learn any more about this driver by looking at arp(4) or ng_ether(4), and I don't know how useful netintro(4) is for readers of this page.

ricera:
@concussious I included your other suggestions, and pushed this into main. You're now free to do your refactor on it!

PR #1623, opened 2025-03-27
style.9: relax "return (value)" requirement for C++

llfw:
consider the following C++ code:

```
	struct S { int a, b; };
	S f() { return {1, 2}; }
```

according to style(9), the return statement should be formatted as:

```
	return ({1, 2});
```

however, this is not valid C++ code and will not compile.

add an exception to style(9) to cover this case.

---

i am aware there's no C++ code in the kernel, but it's often said that userland code should also conform to style(9).

### DIFF
--- a/share/man/man9/style.9
+++ b/share/man/man9/style.9
@@ -22,7 +22,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd February 18, 2025
+.Dd March 27, 2025
 .Dt STYLE 9
 .Os
 .Sh NAME
@@ -766,7 +766,8 @@ to any pointer type.
 .Pp
 Values in
 .Ic return
-statements should be enclosed in parentheses.
+statements should be enclosed in parentheses where possible.
+For example, parentheses cannot be used if the value is a C++ braced-init-list.
 .Pp
 Use
 .Xr err 3


share/man/man9/style.9
```diff
@@ -766,7 +766,8 @@ to any pointer type.
 .Pp
 Values in
 .Ic return
-statements should be enclosed in parentheses.
+statements should be enclosed in parentheses, unless the value is a C++
```
bsdimp, line 769: Don't C aggregate initializers have that problem?
Or better ", unless the resulting expression is a syntax error."

emaste: How about "enclosed in parentheses where possible. For example, parentheses cannot be used if the value is a C++ aggregate initializer."

llfw: i like the suggestion of @emaste and i prefer that to "unless the resulting expression is a syntax error".

if no one objects, i will update the PR to use that phrasing.

llfw:
damnit, sorry, i didn't mean to close this -- automation gone wrong on my side again.

llfw:
updated to use the wording suggested by @emaste, and replace "aggregate initializer" with "braced-init-list", which is the more accurate term for this syntax in C++.

concussious (approved):
I don't know C++, but from manpages standpoint, the language you used here is the same language in the C++ spec and I think that's good!

bsdimp:
This is ready to go, but please let me land it. jhb is also working on some C++ style things and I'll coordinate with him.

llfw:
can i delete/close this branch or do you need it around to merge it later?

i don't mind keeping it, i'm just trying to delete some old branches in my local git repo.

bsdimp:

PR #1630, opened 2025-03-29
pfctl.8: fix typo in reference to rc.conf variable

Defenso-QTH:
The `pfctl(8)` manpage refers to a `pf` variable in `rc.conf(5)` when it should be `pf_enable`.

### DIFF
--- a/sbin/pfctl/pfctl.8
+++ b/sbin/pfctl/pfctl.8
@@ -82,7 +82,7 @@ Translation rules are described in
 .Xr pf.conf 5 .
 .Pp
 When the variable
-.Va pf
+.Va pf_enable
 is set to
 .Dv YES
 in


concussious (approved):

concussious:
Oops, I spoke too soon. The change is not good. The rc.conf syntax is pf_enable. You can enable or disable it with:

```
service pf enable
```
Or
```
service pf disable
```
Thus the text is correct, the variable is `pf`.

concussious (changes requested):
~~Oops, I spoke too soon. The change is not good. The rc.conf syntax is pf_enable. You can enable or disable it with:~~

```
service pf enable
```
Or
```
service pf disable
```
~~Thus the text is correct, the variable is `pf`.~~

jlduran:
> Oops, I spoke too soon. The change is not good. The rc.conf syntax is pf_enable. You can enable or disable it with:
> 
> ```
> service pf enable
> ```
> 
> Or
> 
> ```
> service pf disable
> ```
> 
> Thus the text is correct, the variable is `pf`.

It can be confusing, but the GHPR is valid.

The variable is indeed `pf_enable`, in rc.conf it can be set as `pf_enable="YES"`:

https://github.com/freebsd/freebsd-src/blob/cd46e980134f6fc765b28ee9c8bf41e8fc1b0261/libexec/rc/rc.conf#L233

The case for service(8) is that it works on the script itself, which is named pf, where we can confirm that the `rcvar` is indeed `pf_enable`:

https://github.com/freebsd/freebsd-src/blob/cd46e980134f6fc765b28ee9c8bf41e8fc1b0261/libexec/rc/rc.d/pf#L13

The original form comes from upstream OpenBSD, where in rc.conf it can be configured `pf="YES"`, so it is correct on OpenBSD, but not on FreeBSD (https://github.com/openbsd/src/commit/d7688132f25c7479b231a8fc9bad0b961ea86e40).

concussious (approved):
My mistake, thank you for excellent explanation, and sorry for the misunderstanding.

ImAwsumm (approved):

Defenso-QTH:
Thank you all for the review and commit.

PR #1631, opened 2025-03-31
Fix failed testcase `same_ip_multiple_ifaces_fib0`.

Rin0913:
https://ci.freebsd.org/job/FreeBSD-main-amd64-test/26177/testReport/sys.netinet/fibs_test/same_ip_multiple_ifaces_fib0/

The comment here seems to be attempting to trigger a kernel panic: using ifconfig to assign two identical IP addresses with different masks to the same network interface. If we want to test that it doesn't panic, we should properly handle the error result from the command.

### DIFF
--- a/tests/sys/netinet/fibs_test.sh
+++ b/tests/sys/netinet/fibs_test.sh
@@ -311,8 +311,11 @@ same_ip_multiple_ifaces_fib0_head()
 same_ip_multiple_ifaces_fib0_body()
 {
 	ADDR="192.0.2.2"
+	TMP_ADDR="192.0.2.3"
 	MASK0="24"
+	SUBMASK0="255.255.255.0"
 	MASK1="32"
+	SUBMASK1="255.255.255.255"
 
 	# Unlike most of the tests in this file, this is applicable regardless
 	# of net.add_addr_allfibs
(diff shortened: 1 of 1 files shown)


jlduran:
For some context:

https://reviews.freebsd.org/D48650

At any rate, I think the commit message could benefit from a little more explanation, as it may not be trivial to understand the "why".

cc/ @asomers

glebius (changes requested):
The suggested changes will make the test do something different to what it originally did. The panic triggered in the past happened on deletion of IP address. Now there is no deletion of address, only addition.

IMHO, this test should be just deleted, since condition required to trigger the bug is not reachable itself.

P.S. I'm working on enabling disabled same_ip_multiple_ifaces (note no fib9 suffix) test. I think its existence should be sufficient.

Rin0913:
@glebius How about to add the alias deletion part back?

It is also okay to test in such way.

glebius:
That would make things better, but we need to understand what exactly are we testing against. Also, it it seems strange to me, that test is broken because creation of two same IP addresses is now disallowed, but not fully disallowed - you can still create the combo with help of a temporary address. This looks weird to me. So I would prefer not to adopt the test to reality but rather get opinion from experts what is the expected behavior here. @bu7cher @AlexanderChernikov are invited :)

Rin0913:
@glebius You persuaded me. I missed the words "Then remove the address", so I believed the panic occurs when the IP is assigned. However, it seems that this test is no longer necessary.

PR #1375, opened 2024-08-10
ip6addrctl: Reference RFC 6724 instead

jlduran:
## ip6addrctl.8: Reference RFC 6724 instead

Commit e695500d3cb945e62b4591cb68792601ed5c4b46 updated the policy table to match [RFC 6724], which obsoletes [RFC 3484].

Add a reference to [RFC 6724], and mark it up as a technical report (`%R`).

[RFC 3484]: https://www.rfc-editor.org/rfc/rfc3484
[RFC 6724]: https://www.rfc-editor.org/rfc/rfc6724

---

## ip6addrctl: Update the sample configuration file

Update the sample ip6addrctl.conf.sample file to match the default policy, currently based on [RFC 6724].

---

## getaddrinfo.{1,3}: Cross-reference ip6addrctl(8)

PR #1633, opened 2025-04-01
Fix errors for getaddrinfo_test.

Rin0913:
[getaddrinfo_test](https://ci.freebsd.org/job/FreeBSD-main-amd64-test/26191/testReport/lib.libc.net.getaddrinfo/getaddrinfo_test/)

The same test works on NetBSD. Therefore, I looked the code of NetBSD.

截圖 2025-04-02 下午2 26 15

I think the reason our test is failing is the orders of address information output are different between FreeBSD and NetBSD.

Hence, I add a sort function so that the result would be the same each time no matter what the orders were.

---

UPD1:

The orders only differ by ip6addrctl_policy. It would be better if we could also test the correctness of the sequences of address information. Hence, sorting has less value.

Now, we have three scenarios with their own tests:
1. IPv4 Only;
2. IPv4_prefered;
3. IPv6_prefered.


lwhsu:
Please avoid doing whitespace change (indent changing here) in the same commit for easier review.

Rin0913:
> Please avoid doing whitespace change (indent changing here) in the same commit for easier review.

Done.

glebius:
I don't understand the goal. The existing code allows to run the NetBSD test from contrib unmodified, without making a copy of it. There are no changes required to keep it running.

If you want to add some new tests, just start a new file with a new test set, completely independent from the NetBSD test.

Rin0913:
@glebius 

You said that we try not to modify files in contrib,

However, we have to do it, because only with the sort part added the tests could pass.

Rin0913:
Sorry, let me be clear.

The reason I add the sort function is that once the output order of addrinfo gets changed in the futures, I hope this test could still work.

glebius:
I can't understand what exactly are you fixing with this change. At the current state of the main branch the test passes, the and contrib is not modified, neither the test is copied over to be modified.
```
>kyua test -k /usr/tests/lib/libc/net/getaddrinfo/Kyuafile getaddrinfo_test
getaddrinfo_test:basic  ->  passed  [0.045s]
getaddrinfo_test:empty_hostname  ->  passed  [0.044s]
getaddrinfo_test:empty_servname  ->  passed  [0.032s]
getaddrinfo_test:scopeaddr  ->  passed  [0.022s]
getaddrinfo_test:sock_raw  ->  passed  [0.033s]
getaddrinfo_test:specific  ->  passed  [0.029s]
getaddrinfo_test:unsupported_family  ->  passed  [0.024s]
```

Rin0913:
截圖 2025-04-03 清晨6 28 24

At least, it doesn't pass on our Jenkins, also on my environment.

glebius:
In the Jenkins it fails due to lack of Internet. The NetBSD version doesn't need to be modified due to that. The test either needs to be heavily augmented to build a fake DNS server or just disabled in the certain environments. @ihoro and @ocochard are already working on that.

glebius:
Sorry, I was wrong. It looks like there indeed exist sorting problem.

ocochard:
@Rin0913 Does your host configured with non-standard `ip6addrctl_policy="ipv4_prefer"` or equivalent ?

ocochard (changes requested):
This problem is due to non-default network stack configuration (here something like `service ip6addrctl prefer_ipv4`).
This non-default configuration could be due to the VM configuration or a previous tests that didn’t restore it.

Rin0913:
@ocochard Sorry, no.

My BSD environment is new installed and clean, without any extra configuration.
Actually, I just installed it the day before yesterday.

glebius:
The test fails if machine has IPv6 address for localhost in /etc/hosts, but doesn't have IPv6 activated, e.g. doesn't have a single IPv6 address configured in rc.conf neither have local IPv6 addresses autoconfigured with ipv6_activate_all_interfaces="YES" in rc.conf.

glebius:
With this knowledge as a stopgap we can disable test in certain environments.

Ideally it would be really great to augment the test to check both cases with IPv4 preferred and with IPv6 preferred. That, of course, would require modifications to the NetBSD test and probably creating an extra set of expected results files.

Rin0913:
@glebius Could you please post the test result?

There are 7 tests. Are they all failed?

Rin0913:
@ocochard 

Sorry, I don't know what you want to me change.

ocochard:
> @ocochard
> 
> Sorry, I don't know what you want to me change.

No it is fine, because the problem was more complex than a simple `ip6addrctl prefer_ipv4` config.
So the more stable solution is your proposal.

glebius:
We actually want to test that with prefer_ipv4 we got one sequence and with prefer_ipv6 we got other sequence. That's important feature and sorting results reduces the value of the test. So, please forget about sorting.

Looks like we indeed need to extract the test out of contrib and start hacking on it.

Let me suggest this plan:
* Copy out the test out of contrib unmodified, just change the Makefile. This must be a separate commit. This is important to review further changes.
* Make a copy of expected files for prefer_ipv4 configuration. Every test that behaves differently (3 of them) should be forked into two separate tests: one with prefer_ipv4 set explicitly and other with prefer_ipv6 set explicitly.

Rin0913:
@glebius Sorry, I am not familiar with kernel behavior.

I would like to make sure if the only thing affects orders is ip6addrctl prefer_ipv4 or prefer_ipv6?

If not, we might still have to sort.

glebius:
> I would like to make sure if the only thing affects orders is ip6addrctl prefer_ipv4 or prefer_ipv6?

Right. That should be the only thing that affects order. And the test should validate both cases.

Rin0913:
@glebius Done.

glebius:
Sorry, this is not exactly what I suggested. Here is the plan:

* Copy the test out of contrib unmodified, just modify the Makefile. There is no need to copy h_gai.c. This must be a separate commit.
* Inside the copied out test file add `ip6addrctl  prefer_v6` to existing tests, so that they stop failing. Better make that a separate commit.
* Add more test cases inside the copied out test file for prefer_v4 configuration. Don't create a second copy. Add new expected results files. Don't modify existing expected results file. This also should be a separate commit.

P.S. In a FreeBSD installation there is no sudo. But tests are expected to be run from root, so just remove sudo.

Rin0913:
@glebius  Sorry, I misunderstood you.

I think it is a little bit too late to keep the existing expected result file names because they are generated by myself instead of  using netbsd's.

How about to do merge getaddrinfo_test_v6.sh and getaddr_test.sh only?

Rin0913:
@glebius 

Very sorry to bother you. I am afraid that you miss my message above.

jlduran:
For those of us who arrived late to the party:

NetBSD's /etc/rc.d/ip6addrctl default `auto` configures `prefer_ipv6` if there is no configuration file explicitly, whereas FreeBSD's would also configure `prefer_ipv4` if there are no `ifconfig_*_ipv6` entries in rc.conf.

The tests assume that `prefer_ipv6` is set. When it is not, some of them fail. The proposal here is not only to fix the failing tests, but also to extend coverage in order to test all possibilities (i.e., when `prefer_ipv4` is set).

It would be worth the shot trying to upstream the "fixed" tests.

My only two cents would be not to assume that a user will be OK with a standard[^1] `prefer_ipv6` policy after running the tests, so my suggestion, to make matters worse, would be to backup the current state of the policy. For example:

```diff
--- a/contrib/netbsd-tests/lib/libc/net/getaddrinfo/t_getaddrinfo.sh
+++ b/contrib/netbsd-tests/lib/libc/net/getaddrinfo/t_getaddrinfo.sh
@@ -29,6 +29,28 @@
 # SUCH DAMAGE.
 #

+IP6ADDRCTL_CMD="/usr/sbin/ip6addrctl"
+CURRENT_POLICY_FILE="current.conf"
+
+policy_backup()
+{
+       ${IP6ADDRCTL_CMD} > ${CURRENT_POLICY_FILE}
+}
+
+policy_prefer_ipv6()
+{
+       /etc/rc.d/ip6addrctl prefer_ipv6
+}
+
+policy_cleanup()
+{
+       if [ -f ${CURRENT_POLICY_FILE} ]; then
+               ${IP6ADDRCTL_CMD} flush
+               cat ${CURRENT_POLICY_FILE} | tail -n +2 |
+                   ${IP6ADDRCTL_CMD} install /dev/stdin
+       fi
+}
+
 check_output()
 {
        if [ "$2" = "none" ] ; then
@@ -59,7 +81,7 @@ check_output()
        atf_fail "Actual output does not match expected output"
 }

-atf_test_case basic
+atf_test_case basic cleanup
 basic_head()
 {
        atf_set "descr" "Testing basic ones"
@@ -68,6 +90,9 @@ basic_body()
 {
        TEST=$(atf_get_srcdir)/h_gai

+       policy_backup
+       policy_prefer_ipv6
+
        ( $TEST ::1 http
          $TEST 127.0.0.1 http
          $TEST localhost http
@@ -80,6 +105,10 @@ basic_body()

        check_output basics hosts
 }
+basic_cleanup()
+{
+       policy_cleanup
+}

 atf_test_case specific
 specific_head()
@@ -119,7 +148,7 @@ empty_hostname_body()
        check_output no_host ifconfig
 }

-atf_test_case empty_servname
+atf_test_case empty_servname cleanup
 empty_servname_head()
 {
        atf_set "descr" "Testing empty service name"
@@ -128,6 +157,9 @@ empty_servname_body()
 {
        TEST=$(atf_get_srcdir)/h_gai

+       policy_backup
+       policy_prefer_ipv6
+
        ( $TEST ::1 ''
          $TEST 127.0.0.1 ''
          $TEST localhost ''
@@ -135,8 +167,12 @@ empty_servname_body()

        check_output no_serv hosts
 }
+empty_servname_cleanup()
+{
+       policy_cleanup
+}

-atf_test_case sock_raw
+atf_test_case sock_raw cleanup
 sock_raw_head()
 {
        atf_set "descr" "Testing raw socket"
@@ -145,6 +181,9 @@ sock_raw_body()
 {
        TEST=$(atf_get_srcdir)/h_gai

+       policy_backup
+       policy_prefer_ipv6
+
        ( $TEST -R -p 0 localhost ''
          $TEST -R -p 59 localhost ''
          $TEST -R -p 59 localhost 80
@@ -153,6 +192,10 @@ sock_raw_body()

        check_output sock_raw hosts
 }
+sock_raw_cleanup()
+{
+       policy_cleanup
+}

 atf_test_case unsupported_family
 unsupported_family_head()
```

[^1]: https://github.com/freebsd/freebsd-src/pull/1375#issuecomment-2307155305

jlduran:
Perhaps a "safer bet" would be to skip those tests when IPv4-mapped addresses have a higher precedence than IPv6's default.

Rin0913:
@jlduran Done.

jlduran:
I'm not sure if this is what @glebius had in mind.
I'm still a little uneasy about changing the policies on the machine running the tests (though I feel a bit better returning them to their original state). If there are no plans to upstream these changes, we could use VNET jails.

P.S.: You need to add `atf_set "require.user" "root"` to the head section of the tests that require root privileges.

Rin0913:
@jlduran Done.

Also, 
I’ve organized the commit history a bit — I hope this makes it easier to review.

jlduran:
Let's wait for more input, but I think this is what was requested.

Sorry to insist on the subject, but I would feel much better using VNET jails if we go down this route. Something along the lines of:

```diff
--- a/contrib/netbsd-tests/lib/libc/net/getaddrinfo/t_getaddrinfo.sh
+++ b/contrib/netbsd-tests/lib/libc/net/getaddrinfo/t_getaddrinfo.sh
@@ -29,6 +29,23 @@
 # SUCH DAMAGE.
 #

+freebsd_enter_ipv6()
+{
+       jailname=$1
+
+       jail -c name=${jailname} persist vnet
+       ifconfig -j ${jailname} lo0 inet 127.0.0.1/8
+
+       jexec ${jailname} /etc/rc.d/ip6addrctl prefer_ipv6
+}
+
+freebsd_exit()
+{
+       jailname=$1
+
+       jail -r ${jailname}
+}
+
 check_output()
 {
        if [ "$2" = "none" ] ; then
@@ -59,14 +76,15 @@ check_output()
        atf_fail "Actual output does not match expected output"
 }

-atf_test_case basic
+atf_test_case basic cleanup
 basic_head()
 {
        atf_set "descr" "Testing basic ones"
 }
 basic_body()
 {
-       TEST=$(atf_get_srcdir)/h_gai
+       freebsd_enter_ipv6 basic
+       TEST="jexec basic $(atf_get_srcdir)/h_gai"

        ( $TEST ::1 http
          $TEST 127.0.0.1 http
@@ -80,6 +98,10 @@ basic_body()

        check_output basics hosts
 }
+basic_cleanup()
+{
+       freebsd_exit basic
+}

 atf_test_case specific
 specific_head()
@@ -119,14 +141,15 @@ empty_hostname_body()
        check_output no_host ifconfig
 }

-atf_test_case empty_servname
+atf_test_case empty_servname cleanup
 empty_servname_head()
 {
        atf_set "descr" "Testing empty service name"
 }
 empty_servname_body()
 {
-       TEST=$(atf_get_srcdir)/h_gai
+       freebsd_enter_ipv6 empty_servname
+       TEST="jexec empty_servname $(atf_get_srcdir)/h_gai"

        ( $TEST ::1 ''
          $TEST 127.0.0.1 ''
@@ -135,15 +158,20 @@ empty_servname_body()

        check_output no_serv hosts
 }
+empty_servname_cleanup()
+{
+       freebsd_exit empty_servname
+}

-atf_test_case sock_raw
+atf_test_case sock_raw cleanup
 sock_raw_head()
 {
        atf_set "descr" "Testing raw socket"
 }
 sock_raw_body()
 {
-       TEST=$(atf_get_srcdir)/h_gai
+       freebsd_enter_ipv6 sock_raw
+       TEST="jexec sock_raw $(atf_get_srcdir)/h_gai"

        ( $TEST -R -p 0 localhost ''
          $TEST -R -p 59 localhost ''
@@ -153,6 +181,10 @@ sock_raw_body()

        check_output sock_raw hosts
 }
+sock_raw_cleanup()
+{
+       freebsd_exit sock_raw
+}

 atf_test_case unsupported_family
 unsupported_family_head()
```

Thank you!

Rin0913:
@jlduran Done.

Thank you for the templates.

jlduran:
I'm sold! Thank you for working on fixing the tests.

One minor nit I forgot:

The function that creates the jails should contain the following stanza at the top:

```sh
if [ "$(sysctl -i -n kern.features.vimage)" != 1 ]; then
	atf_skip "This test requires VIMAGE"
fi
```

This ensures that the tests are skipped if the kernel is not compiled with `VIMAGE`.

Also, I know I suggested those names initially, but I think it would be more descriptive/familiar if:

`freebsd_enter_jail` -> `jail_create` (or `vnet_mkjail` if we want to be consistent)
`freebsd_exit` -> `jail_remove` (or `vnet_cleanup` if we want to be consistent)

Rin0913:
@jlduran Done. Thank you.

jlduran:
Landed as:
- 65f6c19efa82e5988092d0c71befd10147a1ede3
- 4b990663ff63de8db700dc0e386421c201e43412
- a18476bd7118a28cc402bda52898999b86f15e58
- a709d15529b3dd8479f2a642b4d0e65e09d0515e
- fe44dee0ef5ca77315c30ad0c819a8b3dabc85a2
- fbd2bea100af71c49b6ba3f9874718424411704b
- 423a21a5b65cc7b7809d29c4d40a8ca4a224a3f5

Thank you!

PR #1634, opened 2025-04-02
bridge(4): allow a member's native vlan id to be configured

llfw:
* add support for a new ifconfig option, `ifpvid <interface> <vlanid>`, which can be used to set the native vlan ID on a bridge member.
* in bridge(4), filter outgoing frames so that they are only sent to interfaces whose native vlan ID is the same as the frame's VLAN ID
* if the interface's VLAN ID is zero (the default) it receives all frames, for backward compatibility.

this is NOT "vlan filtering" (in the Linux sense) and it's not proper access/trunk port support for bridge(4).  however, it's a necessary step to enable proper VLAN support in bridge in the future, and in the mean time, it's a useful self-contained feature.

this change is 100% backward compatible, since if an interface has no VLAN ID configured, it will receive all frames, just like it did before.

i've used `uint32_t` for VLAN ID in bridge(4) so we can support VXLANs in the future, but i haven't changed the userland ABI to avoid breaking the ABI for no reason (yet).

### DIFF
--- a/sbin/ifconfig/ifbridge.c
+++ b/sbin/ifconfig/ifbridge.c
@@ -211,6 +211,8 @@ bridge_status(if_ctx *ctx)
 			else
 				printf(" <unknown state %d>", state);
 		}
+		if (member->ifbr_pvid != 0)
+			printf(" pvid %u", (unsigned)member->ifbr_pvid);
 		printf("\n");
 	}
 
@@ -577,6 +579,24 @@ setbridge_ifpathcost(if_ctx *ctx, const char *ifn, const char *cost)
 }
 
 static void
+setbridge_ifpvid(if_ctx *ctx, const char *ifn, const char *vlanid)
+{
+	struct ifbreq req;
+	u_long val;
+
+	memset(&req, 0, sizeof(req));
+
+	if (get_val(vlanid, &val) < 0)
+		errx(1, "invalid value: %s", vlanid);
+
+	strlcpy(req.ifbr_ifsname, ifn, sizeof(req.ifbr_ifsname));
+	req.ifbr_pvid = val;
+
+	if (do_cmd(ctx, BRDGSIFPVID, &req, sizeof(req), 1) < 0)
+		err(1, "BRDGSIFPVID %s", vlanid);
+}
+
+static void
 setbridge_ifmaxaddr(if_ctx *ctx, const char *ifn, const char *arg)
 {
 	struct ifbreq req;
@@ -659,6 +679,7 @@ static struct cmd bridge_cmds[] = {
 	DEF_CMD_ARG2("ifpriority",	setbridge_ifpriority),
 	DEF_CMD_ARG2("ifpathcost",	setbridge_ifpathcost),
 	DEF_CMD_ARG2("ifmaxaddr",	setbridge_ifmaxaddr),
+	DEF_CMD_ARG2("ifpvid",		setbridge_ifpvid),
 	DEF_CMD_ARG("timeout",		setbridge_timeout),
 	DEF_CMD_ARG("private",		setbridge_private),
 	DEF_CMD_ARG("-private",		unsetbridge_private),
--- a/share/man/man4/bridge.4
+++ b/share/man/man4/bridge.4
@@ -36,7 +36,7 @@
 .\" ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 .\" POSSIBILITY OF SUCH DAMAGE.
 .\"
-.Dd April 10, 2024
+.Dd April 2, 2025
 .Dt IF_BRIDGE 4
 .Os
 .Sh NAME
@@ -50,8 +50,8 @@ kernel configuration file:
 .Cd "device if_bridge"
 .Ed
 .Pp
-Alternatively, to load the driver as a
-module at boot time, place the following lines in
+Alternatively, to load the driver as a module at boot time, place the
+following lines in
 .Xr loader.conf 5 :
 .Bd -literal -offset indent
 if_bridge_load="YES"
@@ -60,18 +60,18 @@ bridgestp_load="YES"
 .Sh DESCRIPTION
 The
 .Nm
-driver creates a logical link between two or more IEEE 802 networks
-that use the same (or
+driver creates a logical link between two or more IEEE 802 networks that
+use the same (or
 .Dq "similar enough" )
 framing format.
-For example, it is possible to bridge Ethernet and 802.11 networks together,
-but it is not possible to bridge Ethernet and Token Ring together.
+For example, it is possible to bridge Ethernet and 802.11 networks
+together, but it is not possible to bridge Ethernet and Token Ring
+together.
 .Pp
 Each
 .Nm
 interface is created at runtime using interface cloning.
-This is
-most easily done with the
+This is most easily done with the
 .Xr ifconfig 8
 .Cm create
 command or using the
@@ -81,8 +81,8 @@ variable in
 .Pp
 The
 .Nm
-interface randomly chooses a link (MAC) address in the range reserved for
-locally administered addresses when it is created.
+interface randomly chooses a link (MAC) address in the range reserved
+for locally administered addresses when it is created.
 This address is guaranteed to be unique
 .Em only
 across all
@@ -112,12 +112,13 @@ by default.
 A bridge can be used to provide several services, such as a simple
 802.11-to-Ethernet bridge for wireless hosts, or traffic isolation.
 .Pp
-A bridge works like a switch, forwarding traffic from one interface
-to another.
-Multicast and broadcast packets are always forwarded to all
-interfaces that are part of the bridge.
-For unicast traffic, the bridge learns which MAC addresses are associated
-with which interfaces and will forward the traffic selectively.
+A bridge works like a switch, forwarding traffic from one interface to
+another.
+Multicast and broadcast packets are always forwarded to all interfaces
+that are part of the bridge.
+For unicast traffic, the bridge learns which MAC addresses are
+associated with which interfaces and will forward the traffic
+selectively.
 .Pp
 By default the bridge logs MAC address port flapping to
 .Xr syslog 3 .
@@ -128,8 +129,8 @@ variable
 to
 .Li 0 .
 .Pp
-All the bridged member interfaces need to be up
-in order to pass network traffic.
+All the bridged member interfaces need to be up in order to pass network
+traffic.
 These can be enabled using
 .Xr ifconfig 8
 or
@@ -137,15 +138,18 @@ or
 in
 .Xr rc.conf 5 .
 .Pp
-The MTU of the first member interface to be added is used as the bridge MTU.
+The MTU of the first member interface to be added is used as the bridge
+MTU.
 All additional members will have their MTU changed to match.
-If the MTU of a bridge is changed after its creation, the MTU of all member
-interfaces is also changed to match.
+If the MTU of a bridge is changed after its creation, the MTU of all
+member interfaces is also changed to match.
 .Pp
-The TOE, TSO, TXCSUM and TXCSUM6 capabilities on all interfaces added to the
-bridge are disabled if any of the interfaces do not support/enable them.
+The TOE, TSO, TXCSUM and TXCSUM6 capabilities on all interfaces added to
+the bridge are disabled if any of the interfaces do not support/enable
+them.
 The LRO capability is always disabled.
-All the capabilities are restored when the interface is removed from the bridge.
+All the capabilities are restored when the interface is removed from the
+bridge.
 Changing capabilities at run-time may cause NIC reinit and a link flap.
 .Pp
 The bridge supports
@@ -153,11 +157,12 @@ The bridge supports
 where the packets are discarded after
 .Xr bpf 4
 processing, and are not processed or forwarded further.
-This can be used to multiplex the input of two or more interfaces into a single
+This can be used to multiplex the input of two or more interfaces into a
+single
 .Xr bpf 4
 stream.
-This is useful for reconstructing the traffic for network taps
-that transmit the RX/TX signals out through two separate interfaces.
+This is useful for reconstructing the traffic for network taps that
+transmit the RX/TX signals out through two separate interfaces.
 .Sh IPV6 SUPPORT
 .Nm
 supports the
@@ -176,8 +181,8 @@ However, the
 .Li AF_INET6
 address family has a concept of scope zone.
 Bridging multiple interfaces changes the zone configuration because
-multiple links are merged to each other and form a new single link
-while the member interfaces still work individually.
+multiple links are merged to each other and form a new single link while
+the member interfaces still work individually.
 This means each member interface still has a separate link-local scope
 zone and the
 .Nm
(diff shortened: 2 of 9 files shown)


llfw:
cc @concussious for docs changes

share/man/man4/bridge.4
```diff
@@ -250,6 +250,27 @@ by setting the
 .Va net.link.bridge.log_stp
 node using
 .Xr sysctl 8 .
+.Sh VIRTUAL LANS (VLANS)
+The
+.Nm
+driver has limited support for virtual LANs (VLANs).
```
concussious, line 256:
```suggestion
driver has limited support for
.Xr vlan 4
interfaces.
```
Maybe?

share/man/man4/bridge.4
```diff
@@ -250,6 +250,27 @@ by setting the
 .Va net.link.bridge.log_stp
 node using
 .Xr sysctl 8 .
+.Sh VIRTUAL LANS (VLANS)
```
concussious, line 253:
```suggestion
.Sh VIRTUAL LANS
```
Parenthesis in a heading is not so good. If the reader searches for vlan they will see it immediately below.

share/man/man4/bridge.4
```diff
@@ -250,6 +250,27 @@ by setting the
...
+An interface without a VLAN ID configured will receive frames for all VLANs.
+An interface with a VLAN ID configured will only receive frames for its
+configured VLAN.
+In either case, if the incoming frame has an 802.1Q tag, the tag will be
+preserved in the output frame; if not, no tag will be added to the output
+frame.
+.Pp
+There is no support for adding or removing 802.1Q tags on outgoing frames,
+and for interfaces without a VLAN ID configured, there is no way to restrict
+which VLANs the interface can send or receive frames for.
```
concussious, line 273:
```suggestion
An incoming packet with an 802.1Q tag
will be assigned to the appropriate VLAN.
An interface's native VLAN ID may be configured using the
.Xr ifconfig 8
.Cm ifvlan
option; incoming packets with no 802.1Q tag,
or where the VLAN ID in the tag is zero,
will be assigned to the interface's VLAN.
.Pp
An interface without a VLAN ID configured will receive frames for all VLANs.
An interface with a VLAN ID configured will only receive frames for its
configured VLAN.
In either case, if the incoming frame has an 802.1Q tag, the tag will be
preserved in the output frame; if not, no tag will be added to the
output frame.
.Pp
There is no support for adding or removing 802.1Q tags on outgoing frames,
and for interfaces without a VLAN ID configured, there is no way to restrict
which VLANs the interface can send or receive frames for.
```
Would you please wrap at 72 for compatibility with the BSD text editor on the BSD standard console, a reference, in base configuration we've preserved as a standard for 50 years?

concussious (commented):
This is a much requested feature, accompanied by great documentation which seems to me to be put in exactly the right places.

share/man/man4/bridge.4
llfw: i don't think this is right: although it's true you can add vlan(4) interfaces to a bridge (and possibly this should be documented as well), that's not what's being discussed here; from bridge's point of view, a vlan(4) interface is just a normal untagged port, not a VLAN.

this section is talking about adding a base ethernet port like `ix0` to a bridge, and then receiving tagged frames on that port, which the bridge will handle internally without passing them to vlan(4).

share/man/man4/bridge.4
llfw: in that case i think i'd rather make the section header "Bridge VLAN support" or something like that, since no one actually says "Virtual LANs" in real life :-)

i will change this to `.Sh BRIDGE VLAN SUPPORT` unless you prefer something else.

share/man/man4/bridge.4
llfw: the rest of the manpage is mostly wrapped at 80 characters, not 72.  should i rewrap the entire thing (as a separate commit)?

llfw:
i've been doing more work on bridge VLANs (based on this PR) on a different branch, but in the interests of actually getting this committed, i think it makes sense to divide this into managable chunks of discrete functionality, so i won't submit the additional changes until this is committed.

sys/net/if_bridge.c
```diff
@@ -2784,6 +2821,10 @@ bridge_broadcast(struct bridge_softc *sc, struct ifnet *src_if,
 		if (sbif && (sbif->bif_flags & dbif->bif_flags & IFBIF_PRIVATE))
 			continue;
 
+		/* VLAN filtering for interfaces with non-zero VLAN ID */
+		if ((dbif->bif_vlan != DOT1Q_VID_NULL) && (vlan != dbif->bif_vlan))
```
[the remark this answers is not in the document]
llfw: fixed

share/man/man4/bridge.4
concussious: I wasn't going to go that far as to ask you to do the whole thing, a bit out of scope. I'd take that though! When I work on manuals, any cleanup commit also includes wrapping at 72 because I use vi on vt.

llfw: done

llfw:
cc @markjdb and @gmshake for code review (only because you reviewed my previous bridge PRs).

share/man/man4/bridge.4
```diff
@@ -128,36 +129,40 @@ variable
...
-The TOE, TSO, TXCSUM and TXCSUM6 capabilities on all interfaces added to the
-bridge are disabled if any of the interfaces do not support/enable them.
+The TOE, TSO, TXCSUM and TXCSUM6 capabilities on all interfaces added to
+the bridge are disabled if any of the interfaces do not support/enable
+them.
 The LRO capability is always disabled.
-All the capabilities are restored when the interface is removed from the bridge.
-Changing capabilities at run-time may cause NIC reinit and a link flap.
+All the capabilities are restored when the interface is removed from the
+bridge.  Changing capabilities at run-time may cause NIC reinit and a
```
concussious, line 152: Roff requires one sentence per line.

llfw: fixed (as a force push)

llfw:
thinking about the terminology here, it occurs to me that `ifpvid` would be a better name for the ifconfig option than `ifvlan`, since PVID is the standard terminology for this and is more specific.  unless anyone has an objection, i'll change this.

llfw:
added a couple of tests (suggested by kp@): one to check that two ports on the same vlan can communicate, and one to check that two ports on different vlans can't communicate.

sbin/ifconfig/ifbridge.c
```diff
@@ -211,6 +211,8 @@ bridge_status(if_ctx *ctx)
 			else
 				printf(" <unknown state %d>", state);
 		}
+		if (member->ifbr_vlan != 0)
+			printf(" vlan %u", (unsigned)member->ifbr_vlan);
```
kprovost: Do we need that cast here? ifbr_vlan is already unsigned.

sys/net/if_bridge.c
```diff
@@ -251,6 +251,7 @@ struct bridge_iflist {
 	uint32_t		bif_addrcnt;	/* cur. # of addresses */
 	uint32_t		bif_addrexceeded;/* # of address violations */
 	struct epoch_context	bif_epoch_ctx;
+	uint32_t		bif_vlan;	/* native vlan id */
```
kprovost: Why uint32_t and not uint16_t as we do elsewhere for vlan id?

sys/net/if_bridge.c
```diff
@@ -1796,6 +1801,23 @@ bridge_ioctl_sifmaxaddr(struct bridge_softc *sc, void *arg)
...
+bridge_ioctl_sifvlan(struct bridge_softc *sc, void *arg)
+{
+	struct ifbreq *req = arg;
+	struct bridge_iflist *bif;
+
+	bif = bridge_lookup_member(sc, req->ifbr_ifsname);
+	if (bif == NULL)
+		return (ENOENT);
+
+	if (req->ifbr_vlan >= 4096)
```
kprovost: Maybe `> DOT1Q_VID_RSVD_IMPL` instead of `>= 4096`?

Or maybe we need to introduce DOT1Q_VID_MAX or something.

kprovost (commented):
I've not yet looked at the other commits in this series in depth.
It does looks like a number of them are fixes to previous commits that can be squashed together. Can you do that? It'll make reviewing a bit easier.

sys/net/if_bridge.c
llfw: i've used uint32_t for the new vlan code (without changing the existing code that uses uint16_t) in case we want to add vxlan support to if_bridge later[0].  i don't mind changing these into uint16_t instead though.

[0] obviously .1q doesn't support vxlan ids, but you might have an access port with a pvid in the vxlan range.

llfw: actually, i think it's better to replace both uint32_t and uint16_t with a new typedef, like bridge_vlanid_t.

sys/net/if_bridge.c
llfw: i think `DOT1Q_VID_RSVD_IMPL` is wrong because that's not guaranteed to be 4096, it just is by chance.

i'll add new macro for min (1) and max (4095) vlan id.

sys/net/if_bridge.c
kprovost: I like that idea. Or just plain vlanid_t?

I'm not a huge fan of choosing a datatype now on the theory that we might someday maybe implement more features. It's not that hard to change it later. It's even easier if we follow your typedef idea.

llfw:
> It does look like a number of them are fixes to previous commits that can be squashed together. Can you do that?

i'll squash all the commits currently on ths branch, but i won't squash the new commits to address review comments unless you specifically want me to do that.

incidentally the rest of my bridge changes are here: https://github.com/freebsd/freebsd-src/compare/main...llfw:freebsd-src:lf/dev/bridge-1q; if you prefer i can submit all of these as a single large PR.  i have no real preference there, it's just about which is easier to review.

kprovost:
> > It does look like a number of them are fixes to previous commits that can be squashed together. Can you do that?
> 
> i'll squash all the commits currently on ths branch, but i won't squash the new commits to address review comments unless you specifically want me to do that.
> 
I'm quite happy for logically distinct changes to be separate commits (e.g. the tests can be their own commit, as can the man page changes). It's just that there are a few commits that are obviously fixes for (mostly minor) mistakes in a previous commit, and those don't really add any value. Let's make it look as if you got it all right the first time ;)

> incidentally the rest of my bridge changes are here: [main...llfw:freebsd-src:lf/dev/bridge-1q](https://github.com/freebsd/freebsd-src/compare/main...llfw:freebsd-src:lf/dev/bridge-1q); if you prefer i can submit all of these as a single large PR. i have no real preference there, it's just about which is easier to review.

I'll try to look at that branch too. It doesn't matter too much where they are. It's as you say, about making it easy to review.

sys/net/if_bridge.c
llfw: as the new type needs to be accessible from userland (for ioctls) and isn't specific to bridge, i suggest calling it `ether_vlanid_t` and putting it in `net/ethernet.h`.  if we want to support vxlans in future then we would need a new type like `vxlanid_t`, but for now let's be clear about what we're doing.

i think the `DOT1Q_*` constants should also move to ethernet.h at some point (they're currently in `if_vlan_var.h`) but that can wait.

llfw:
i've rebased and force pushed this to fix an issue with the padding in `struct ifbreq` being wrong, make sure to take the latest version if this is committed (otherwise the size of the struct will change).

llfw:
someone in another forum has raised some questions about the specific terminology/command syntax we want to use for this which i don't think are unreasonable.

@kprovost do you think it would make sense to close this PR and re-submit it as a small number of commits (probably 4-5) that implement the entire feature (rather than just this one part of it) so it can be reviewed as a whole?  that would mean we could consider the syntax for all the new features at once.

(i can put this in phab if you prefer, i only used GitHub because it's the recommended place for non-committers to submit changes.)

andershagman:
As a long time network guy I have worked with a lot of different switch and bridge types. Naming conventions regarding different functionalities in the switch can easily be misunderstood.
We have trunk, access port, PVID, native VLAN, allowed VLANs, tagged and untagged.

Trunk means a p2p connection that can handle multiple users at the same time. This goes back to the time of telephony. So it can mean multiplexing on the same wire or multiplexing by link aggregetion.
On Ethernet we normally mean multiplexing by VLAN tagging. Att the same time we allow for one VLAN to be untagged.
This because we can have multipoint Ethernet with a hub on the link. Who does that today?

Trunk = tagged frames on the wire. Except one some times.

Access port, PVID or native VLAN are just different ways to express untagged frames. Access port often also disallow tagged frames.

Access, PVID, native = untagged frames on the wire. But only one VLAN.

Allowed VLANs is often a list of VLANs not filtered and can therefore pass in or out of a switch port.

To sum up we want to control how to handle tagged and untagged frames and how to filter them.

So my simplistic suggestion is:

```
ifconfig bridge0 ifvlans xl0 tag 2-5,7,10 untag 15
```
Allow VLANs 2-5,7 and 10 as tagged and 15 as untagged on port xl0. This is a typical trunk.

```
ifconfig bridge0 ifvlans tap3 untag 15
```
Allow VLAN 15 as untagged on port tap3. This is an access port.

```
ifconfig bridge0 ifvlans xl0 +tag 12-13
```
Add VLAN 12 and 13 as allowed tagged VLANs

```
ifconfig bridge0 ifvlans xl0 -tag 10
```
Delete VLAN 10 from allowed VLANs.

```
ifconfig bridge0 ifvlans xl0 untag 12
```
Swap tagging and untagging on VLAN 12 and VLAN 15. Or error message ”untagged used by VLAN 15”

```
ifconfig bridge0 ifvlans xl0 tag 2-4094 untag 1
```
Default configuration for a bridge member.

```
ifconfig bridge0 -ifvlans xl0
```
Put the bridge port back to default

```
ifconfig bridge0 ifvlans xl0
```
Show the VLAN configuration

llfw:
i've thought about this and i prefer your suggested syntax, but due to limitations in how ifconfig currently does argument parsing, the syntax `ifconfig br0 ifvlans intf0 tag X untag Y` is rather difficult to support because every command (like `ifvlans`) has to have a fixed number of arguments.

this touches on a related issue which is that currently it's impossible to add a bridge member and its VLAN configuration at the same time, meaning every interface added will briefly be an unrestricted trunk port.  fixing this in the kernel API is trivial, but making it work in ifconfig is more difficult.

i will sit down with ifconfig/ifbridge.c and see if i can come up with a better way to handle this, preferably without rewriting the entire thing.

andershagman:
I see. Ifconfig only supports one or two arguments.
The simplest solution is use DEF_CMD_ARG2 with a string:
ifconfig bridge0 ifvlans vtnet0 "tag 1-4 untag 12"
ifconfig bridge0 ifvlans vtnet0 "untag 12"
ifconfig bridge0 ifvlans vtnet0 "+tag 5"
ifconfig bridge0 ifvlans vtnet0 show ; don´t need ""

llfw:
no, this syntax with "" is terrible and i will never do this.  either we can fix ifconfig properly or we'll find anyway way to do it.

andershagman:
19 apr. 2025 kl. 08:33 skrev Lexi ***@***.***>:﻿
  no, this syntax with "" is terrible and i will never do this.  either we can fix ifconfig properly or we'll find anyway way to do it.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you commented.Message ID: ***@***.***>



llfw left a comment (freebsd/freebsd-src#1634)
no, this syntax with "" is terrible and i will never do this.  either we can fix ifconfig properly or we'll find anyway way to do it.


Totally agree/Anders

bsdimp:
Tagged needs-review since that's how I read the history here. Finish here or phab, per discord, etc etc

llfw:
https://reviews.freebsd.org/D49993

2Belette:
@llfw what you propose is very nice, and for me one of the key things missing compared to something like openvswitch, it will improve the flexibility and security a lot with bridge ! thanks for that.

Do you know if we can expect this into 14.3 lifecycle or this would not arrive 15.0 ?

ghost:
> Do you know if we can expect this into 14.3 lifecycle or this would not arrive 15.0 ?

this will almost certainly not land in 14.x due to kernel ABI changes that can't be MFC'd, so it'll be in 15.0.

note that the design has changed a bit since this PR, see the Phabricator stack at https://reviews.freebsd.org/D49993 for the current version.

2Belette:
yes thanks I already read it attentively, I don't have an account so I posted here to thank you ! I hope this will be merged and accepted !

PR #1635, opened 2025-04-03
timeout(1): Sync from DragonFly BSD

liweitianux:
The timeout(1) utility in DragonFly BSD has recently been updated and improved to conform to the [POSIX.1-2024 Standard](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/timeout.html).  Sync the changes over.

@bapt

### DIFF
--- a/bin/timeout/timeout.1
+++ b/bin/timeout/timeout.1
@@ -24,7 +24,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd April 2, 2025
+.Dd April 3, 2025
 .Dt TIMEOUT 1
 .Os
 .Sh NAME
@@ -64,6 +64,14 @@ zero, signifies no limit.
 Therefore, a signal is never sent if
 .Ar duration
 is 0.
+.Pp
+The signal dispositions inherited by the
+.Ar command
+are the same as the dispositions that
+.Nm
+inherited, except for the signal that will be sent upon timeout,
+which is reset to take the default action and should terminate
+the process.
 .Pp
 The options are as follows:
 .Bl -tag -width indent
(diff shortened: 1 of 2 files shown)


bsdimp (changes requested):
The commit message for _exit needs to explain why that's a good thing and not just reareanging the deck chairs. What is the benefit from not producing an error message and using _exit >instead of exit. Seems like a bad change to me to not flush the stdio buffers and to not tell the user why something went wrong. How is that more standards compliant?

bsdimp:
Github makes it stupidly hard to critique commit messages so i only did one. Others likely need to include the why in addition to the what.

liweitianux:
@bsdimp, thank you for the quick review.

> What is the benefit from not producing an error message and using _exit >instead of exit.

The [commit](https://github.com/freebsd/freebsd-src/pull/1635/commits/eb99088bcf7eaace66661260304343d8ecf6ec85) didn't drop the error message. It was replacing the `err()` with `warn() + _exit()`.

> Seems like a bad change to me to not flush the stdio buffers and to not tell the user why something went wrong.

The `execvp()` is performed by the *forked child* of timeout(1), so it's actually correct for the child to exit with `_exit()` instead of `exit()`.

> How is that more standards compliant?

This `err()` -> `warn() + _exit()` change didn't mean for POSIX-compliant. It's the other updates (e.g., adding `-f` and `-p` short options; changing the signal inheritance; catching all signals; killing self with the same signal).

bsdimp:
> timeout(1): Use _exit(2) instead of err() in child if exec failed

This commit message needs a why. Why does it need to call _exit. I know, but we should document that we're avoiding the normal exit() stdio flush (or is there a different reason?)

bsdimp:
> timeout(1): sig_atomic_t variables must also be 'volatile'

Again, this needs a why in the commit message. The change is also good.

[[ I'm doing these one at a time because github doesn't have a good interface to suggesting commit message changes ]]

bsdimp (approved):
I like these changes. I've suggested a couple of commit messages that need to be improved a little. We could land with the messages as is, but I think the details on those are important. There's other messages that could be better, honestly, but github's suggetion mechanism sucks and using it is more hassle than the benefit from the couple of wording tweaks I thought of.

bapt:
I had to adapt the test suite in 2 cases (minor), other than merged! thanks a lot !

liweitianux:
Thank you. I was expecting some comments from @bapt and then would do a bunch of updates, including improvements to the commit messages to explain a bit more on "why".

PR #1636, opened 2025-04-03
Remove testcase sys.netinet.fibs_test.same_ip_multiple_ifaces_fib0.

Rin0913:
https://ci.freebsd.org/job/FreeBSD-main-amd64-test/26201/testReport/sys.netinet/fibs_test/same_ip_multiple_ifaces_fib0/

The original test behavior: Create two tap interfaces and assign them both the same IP address with different netmasks on the default FIB, and then remove one's IP. Test if the machine panic or not.

However, it is not allowed to assign same ip address to two interfaces with different netmasks.

https://github.com/freebsd/freebsd-src/pull/1631

### DIFF
--- a/tests/sys/netinet/fibs_test.sh
+++ b/tests/sys/netinet/fibs_test.sh
@@ -297,45 +297,6 @@ default_route_with_multiple_fibs_on_same_subnet_inet6_cleanup()
 	cleanup_ifaces
 }
 
-
-# Regression test for PR kern/189089
-# Create two tap interfaces and assign them both the same IP address but with
-# different netmasks, and both on the default FIB.  Then remove one's IP
-# address.  Hopefully the machine won't panic.
-atf_test_case same_ip_multiple_ifaces_fib0 cleanup
-same_ip_multiple_ifaces_fib0_head()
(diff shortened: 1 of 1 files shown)


glebius:
Me and ae@ are still unsure why can't we add a prefix directly, but can do that with a workaround, that K has found.

Rin0913:
@glebius I think it is because of https://github.com/freebsd/freebsd-src/blob/b2f4d9b67f317f31deeba62cf448d295c29b3f07/sys/netinet/in.c#L534 that we can't add a prefix directly.

However, I didn't find a workaround so that we can do it?

#1631 is nothing but ignoring the error code.

Rin0913:
@glebius 

Could you kindly help me review the PR?

Thank you very much!

glebius:
Sorry, running short on free time. I really want to fully understand the problem: why it was possible to set such configuration before and not possible now, before going forward with any changes. I was really hoping that experts such as @bu7cher and @AlexanderChernikov will help.

Please don't rush with pushing the change in, neither abandon the pull request. It is a problem that needs to be investigated.

PR #1637, opened 2025-04-04
net/if_bridgevar.h: add include guard

llfw:


llfw:
committed in f6aedb956ef154828c4bfaddaa5d5eb2dda5225c

PR #1639, opened 2025-04-04
bridge: define VLANTAGOF correctly

llfw:
to avoid strange precedence errors, enclose the macro body in parentheses.  this fixes constructs like: if (VLANTAGOF(m) == i)


llfw:
committed in f36292f010fcc6391605182f2973fdc12a3bd15f

PR #1641, opened 2025-04-05
bridge: allow IP addresses on members to be disabled

llfw:
add a new sysctl, net.link.bridge.member_ifaddrs, which defaults to 1.

if it is set to 1, bridge behaviour is unchanged.

if it is set to 0:

- an interface which has AF_INET6 or AF_INET addresses assigned cannot be added to a bridge.
- an interface in a bridge cannot have an AF_INET6 or AF_INET address assigned to it.
- the bridge will no longer consider the lladdrs on bridge members to be local addresses, i.e. frames sent to member lladdrs will not be processed by the host.

update bridge.4 to document this behaviour, as well as the existing recommendation that IP addresses should not be configured on bridge members anyway, even if it currently partially works.

in testing, setting this to 0 on a bridge with 50 member interfaces improved throughput by 22% (4.61Gb/s -> 5.67Gb/s) across two member epairs due to eliding the bridge member list walk in GRAB_OUR_PACKETS.

### DIFF
--- a/share/man/man4/bridge.4
+++ b/share/man/man4/bridge.4
@@ -36,7 +36,7 @@
 .\" ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 .\" POSSIBILITY OF SUCH DAMAGE.
 .\"
-.Dd April 10, 2024
+.Dd April 5, 2025
 .Dt IF_BRIDGE 4
 .Os
 .Sh NAME
@@ -158,6 +158,19 @@ This can be used to multiplex the input of two or more interfaces into a single
 stream.
 This is useful for reconstructing the traffic for network taps
 that transmit the RX/TX signals out through two separate interfaces.
+.Pp
+To allow the host to communicate with bridge members, IP addresses
+should be assigned to the
+.Nm
+interface itself, not to the bridge's member interfaces.
+Assigning IP addresses to bridge member interfaces is unsupported, but
+for backward compatibility, it is permitted if the
+.Xr sysctl 8
+variable
+.Va net.link.bridge.member_ifaddrs
+is set to 1, which is the default.
+In a future release, this sysctl may be set to 0 by default, or may be
+removed entirely.
 .Sh IPV6 SUPPORT
 .Nm
 supports the
--- a/tests/sys/net/if_bridge_test.sh
+++ b/tests/sys/net/if_bridge_test.sh
@@ -703,6 +703,82 @@ many_bridge_members_cleanup()
 	vnet_cleanup
 }
 
+atf_test_case "member_ifaddrs_enabled" "cleanup"
+member_ifaddrs_enabled_head()
+{
+	atf_set descr 'bridge with member_ifaddrs=1'
+	atf_set require.user root
+}
+
+member_ifaddrs_enabled_body()
+{
+	vnet_init
+	vnet_init_bridge
+
+	ep=$(vnet_mkepair)
+	ifconfig ${ep}a inet 192.0.2.1/24 up
+
+	vnet_mkjail one ${ep}b
+	jexec one sysctl net.link.bridge.member_ifaddrs=1
+	jexec one ifconfig ${ep}b inet 192.0.2.2/24 up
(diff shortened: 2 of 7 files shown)


llfw:
i added tests for this functionality, but the tests will fail unless #1645 is committed as well.

bsdimp:
Looks like it needs review, per discord do here or phab, etc

kprovost (commented):
Please squash this commit into the main 'bridge: allow IP addresses on members to be disabled' commit.

share/man/man4/bridge.4
```diff
@@ -158,6 +158,19 @@ This can be used to multiplex the input of two or more interfaces into a single
...
+should be assigned to the
+.Nm
+interface itself, not to the bridge's member interfaces.
+Assigning IP addresses to bridge member interfaces is unsupported, but
+for backward compatibility, it is permitted if the
+.Xr sysctl 8
+variable
+.Va net.link.bridge.member_ifaddrs
+is set to 1, which is the default.
+In a future release, this sysctl may be set to 0 by default, or may be
```
kprovost, line 172: I'm tempted to say we should immediately announce this will be default in 16.

tests/sys/net/if_bridge_test.sh
```diff
@@ -703,6 +703,77 @@ many_bridge_members_cleanup()
...
+member_ifaddrs_enabled_body()
+{
+	vnet_init
+	vnet_init_bridge
+
+	ep=$(vnet_mkepair)
+	ifconfig ${ep}a inet 192.0.2.1/24 up
+
+	vnet_mkjail one ${ep}b
+	jexec one sysctl net.link.bridge.member_ifaddrs=1
```
kprovost, line 722: I approve of setting the sysctl explicitly, despise '1' being default, because now the test will pass even when we change the default.

tests/sys/net/if_bridge_test.sh
```diff
@@ -703,6 +703,77 @@ many_bridge_members_cleanup()
...
+	bridge=$(jexec one ifconfig bridge create)
+
+	# adding an interface with an IPv4 address
+	ep=$(jexec one ifconfig epair create)
+	jexec one ifconfig ${ep} 192.0.2.1/32
+	atf_check -s exit:1 -e ignore jexec one ifconfig ${bridge} addm ${ep}
+
+	# adding an interface with an IPv6 address
+	ep=$(jexec one ifconfig epair create)
+	jexec one ifconfig ${ep} inet6 2001:db8::1/128
```
kprovost, line 758: What happens with an interface that only has a link-local IPv6 address? I'd assume that also fails to be added, but it might be worth extending the test case to cover that scenario too.

kprovost (approved):
LGTM. I'll let you negotiate the commit with your mentors.

share/man/man4/bridge.4
llfw: my preference is to make this the default in 15 and remove it in 16.  since i posted this, i met another user on IRC who was running into a strange problem caused by assigning an IP address to a member.  it just doesn't work properly and we should remove it rather than fixing it.

this might be a bit too aggressive though, so i'm okay with making it the default in 16 and removing it in 17.

tests/sys/net/if_bridge_test.sh
llfw: yes, it fails to be added.  this causes a bit of a negative UX experience in one case where you create an interface, add an IPv6 GUA, remove the GUA, and the LLA is still there so you can't put it in a bridge.  i wonder if it's worth having ifconfig check for this to print a better warning.

i didn't want to have bridge(4) itself remove the addresses automatically because i think this will silently break existing setups.

either way i'll add a test for this though.

llfw:
i requested review of this on github, but i'm wondering if it makes sense to put it in phab just so more people from #network have a chance to see it.

llfw:
rebased, squashed and added a new test.  i'll bump Dd before committing.

kprovost:
> i requested review of this on github, but i'm wondering if it makes sense to put it in phab just so more people from #network have a chance to see it.

That's probably a good idea, yes. I know I spend more time looking at pharicator than at GitHub and I suspect I'm not the only one.

PR #1642, opened 2025-04-05
bridge: store a bridge_iflist pointer in ifnet

llfw:
currently, bridge stores a pointer to its softc in ifnet.  this means that when processing a packet, it needs to walk the bridge's member list to find the bridge_iflist for the interface the packet came from.

instead, store a pointer to the bridge_iflist in ifnet, and add a pointer from bridge_iflist back to the bridge_softc.  this means given an ifnet, we always have both the softc and the bridge_iflist without a list walk.

there are two places outside if_bridge that treat ifnet->if_bridge as something other than an opaque pointer: bridgestp, and netinet.  add two function pointers exported from if_bridge to handle those cases.

bump __FreeBSD_version as this is technically a KABI break.

### DIFF
--- a/sys/net/bridgestp.c
+++ b/sys/net/bridgestp.c
@@ -57,6 +57,7 @@
 #include <net/if_types.h>
 #include <net/if_llc.h>
 #include <net/if_media.h>
+#include <net/if_bridgevar.h>
 #include <net/vnet.h>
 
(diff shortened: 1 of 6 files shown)


llfw:
closing as this was committed in 85967694b4590a436c56a061b397620e342784ae.  thanks!

PR #1643, opened 2025-04-06
netstat(1): add description for option -o and -O.

Rin0913:
They are added in 
- https://github.com/freebsd/freebsd-src/commit/fedeb08b6a58e708e1153224d37ad26bdc1062a2
- https://github.com/freebsd/freebsd-src/commit/a666325282eaed4b044459d121f339b2d6d0224b

However, these options are not introduced in the manual.

### DIFF
--- a/usr.bin/netstat/netstat.1
+++ b/usr.bin/netstat/netstat.1
@@ -191,6 +191,10 @@ and the third count is the maximum number of queued connections.
 Do not resolve numeric addresses and port numbers to names.
 See
 .Sx GENERAL OPTIONS .
+.It Fl o
+Print nexthop (nhops) information associated with routing entries.
+.It Fl O
+Print nexthop groups (nhgrp) information associated with routing entries.
 .It Fl P
 Display the log ID for each socket.
 .It Fl R


concussious (commented):
IIUC these should also be added to SYNOPSIS and the program's usage: output.

Rin0913:
> IIUC these should also be added to SYNOPSIS and the program's usage: output.

Added.

concussious:
I see they're also missing from the program's usage statement.

llfw:
since you've added a separate `SYNOPSIS` section for the new options (-o, -O) i think these also need to be in a new `DESCRIPTION` section.

Rin0913:
@llfw Added.

usr.bin/netstat/netstat.1
```diff
@@ -118,7 +124,7 @@ command uses.
 .It Xo
 .Bk -words
 .Nm
-.Op Fl 46AaCLnRSTWx
+.Op Fl 46AaCLnoORSTWx
```
llfw, line 127: you don't need -o or -O here now since they're in their own section

usr.bin/netstat/netstat.1
```diff
@@ -191,6 +197,10 @@ and the third count is the maximum number of queued connections.
 Do not resolve numeric addresses and port numbers to names.
 See
 .Sx GENERAL OPTIONS .
+.It Fl o
```
llfw, line 200: as above, no need to document -o/-O here since they're covered elsewhere

llfw (changes requested):

usr.bin/netstat/netstat.1
Rin0913: Done.

usr.bin/netstat/netstat.1
```diff
@@ -401,6 +407,35 @@ See
...
+.Ek
+.Xc
+Print nexthop (nhops) information associated with routing entries.
+When used with
+.Fl 4
+or
+.Fl 6 ,
+limit the output to IPv4 or IPv6 routes respectively.
+This option provides details about individual nexthop addresses used in routing decisions.
+
```
llfw, line 422: sorry, i missed this in the previous review, but you have an extraneous blank line here which causes the manpage to render badly.

llfw (changes requested):
one more change (sorry, my bad) and could you also change the commit message from `netstat(0)` to `netstat (1)`?

usr.bin/netstat/netstat.1
Rin0913: Done.

llfw:
lgtm, requesting a mentor review to commit.

@ziaee do you want to re-review this?  i think your comments are addressed.

usr.bin/netstat/netstat.1
```diff
@@ -105,6 +105,12 @@
 .It Nm Fl Q
 .Op Fl j Ar jail
 .Op Fl -libxo
+.It Nm Fl o
+.Op Fl 4
+.Op Fl 6
+.It Nm Fl O
+.Op Fl 4
+.Op Fl 6
```
concussious, line 113:
```suggestion
.It Nm Fl o
.Op Fl 4 | Fl 6
.It Nm Fl O
.Op Fl 4 | Fl 6
```
In the description you show these as mutually exclusive. If they are this should be listed here the same way.

concussious (changes requested):
Cool, thanks! Few small nits.

> They are added in:

This is extremely important information, please put this in two `Fixes:` lines the commit message trailer.

Also these don't seem to do anything on client machine? On my laptop I get:
```
netstat: sysctl: net.route.0.0.nhdump.0 estimate: Address family not supported by protocol family
```

usr.bin/netstat/netstat.1
concussious: Oh, in that case, `.Op` is the optional macro, so it should be removed here and in the description.

Rin0913: Sorry, what should I use to describe required parameter?

Rin0913:
@concussious 


截圖 2025-04-24 晚上9 47 43

I am not sure if it is good. It looks a little bit wired.
How should I modify this to follow the mandoc convention?

concussious:
I think it looks exactly correct, but this is contested. Some people use `Ns | Ns` to remove the space. The FreeBSD manpages style guide says that we do that as of 2023, but the reality is that we almost never do that, and other projects do not do that. You can do that if you want.

Rin0913:
I love space here :)
Who doesn't?

usr.bin/netstat/netstat.1
```diff
@@ -401,6 +405,35 @@ See
 .El
 .It Xo
 .Bk -words
+.Nm netstat
+.Fl o
+.Fl 4 | Fl 6
+.Ek
+.Xc
+Either -4 (IPv4) or -6(IPv6) should be set. Print nexthop (nhops) information associated with routing entries.
```
concussious, line 413: The introductory sentence should explain what it does. Switch the first two sentences, and combine the new second sentence with the third sentence.

usr.bin/netstat/netstat.1
```diff
@@ -401,6 +405,36 @@ See
...
+.Fl 4 | Fl 6
+.Ek
+.Xc
+Print nexthop (nhops) information associated with routing entries.
+When used with
+.Fl 4
+or
+.Fl 6 ,
+limit the output to IPv4 or IPv6 routes respectively.
+Either -4 (IPv4) or -6(IPv6) should be set.
```
concussious, line 419: I think you can remove this line.
```suggestion
```

usr.bin/netstat/netstat.1
```diff
@@ -401,6 +405,36 @@ See
...
+.Fl 4 | Fl 6
+.Ek
+.Xc
+Print nexthop groups (nhgrp) information associated with routing entries.
+When used with
+.Fl 4
+or
+.Fl 6 ,
+restrict the output to IPv4 or IPv6 nexthop groups respectively.
+Either -4 (IPv4) or -6(IPv6) should be set.
```
concussious, line 434:
```suggestion
```
I think you can remove this line.

concussious (approved):

usr.bin/netstat/netstat.1
kevans91, line 419: I'm not sure this sentence is really necessary; presumably this is like the RFC "should" rather than trying to express that one of -[46] are mandatory?  I'm not sure that we usually make these kinds of recommendations on output filtering once we've described how the filtering works (the sentence before).  If we keep it, these should probably be converted to `.Fl 4` and `.Fl 6` and the spacing made consistent between the two.

Ditto below for -O.

kevans91: Ah, sorry, same comment was made before; I guess I don't like github's diff view.

usr.bin/netstat/netstat.1
Rin0913: Done!

kevans91 (approved):

llfw:
afa30c5dec5afc8496172a062c93e96d78f5b9b9 with Fixes: note and Dd bump

thanks!

PR #1645, opened 2025-04-07
in6_control_ioctl: correctly report errors from SIOCAIFADDR_IN6

llfw:
we have to use 'goto out' here rather than 'break' because otherwise error is set to 0, which means the error is not propagated back to the caller.

### DIFF
--- a/sys/netinet6/in6.c
+++ b/sys/netinet6/in6.c
@@ -587,7 +587,7 @@ in6_control_ioctl(u_long cmd, void *data,
 #endif
 		error = in6_addifaddr(ifp, ifra, ia);
 		ia = NULL;
-		break;
+		goto out;
 
 	case SIOCDIFADDR_IN6:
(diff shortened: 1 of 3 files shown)


kprovost:
I'm running local tests, but I expect to commit this later today.

That said, it'd be very nice to get a test case for this as well. You know how to reproduce it, so you've done the hard work there already.

llfw:
ok, i will add a test case.

llfw:
i couldn't find an obvious place to put the new test case, so i've added a new test, `sys/netinet6/addr6.sh` on the assumption that there are probably other things related to configuring inet6 addresses that could/should be tested.

llfw:
closing as committed in 1408670c90ae38fd51ea6aaabe46757bc880608e & 50725e585de77c20f7c96b836dd68ba954756ce7, thanks.

PR #1646, opened 2025-04-07
Fix typo in the test causing errors in sys.netinet6.redirect.valid_redirect.

Rin0913:
Bug: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=247729
CI: https://ci.freebsd.org/job/FreeBSD-main-amd64-test/26219/testReport/sys.netinet6/redirect/valid_redirect/

I think it is a typo.

However, this is not the original bug. The original bug was fixed somehow.

### DIFF
--- a/tests/sys/netinet6/redirect.py
+++ b/tests/sys/netinet6/redirect.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # -
 # SPDX-License-Identifier: BSD-2-Clause
 #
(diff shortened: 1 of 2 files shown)


tests/sys/netinet6/redirect.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
```
jlduran, line 1: I see `#!/usr/bin/env python3` wins by a slight margin when counting occurrences.
I would think using `env` is better, do you have any opinions?

jlduran (approved):
Fixes: eddfb2e86fbc ("Fix IPv6 regression introduced by r362900.")

tests/sys/netinet6/redirect.py
Rin0913: I don't have strong preference.
I've modified it.

kprovost:
Excellent catch. Pushed.

PR #1647, opened 2025-04-08
merge traceroute6 into traceroute

llfw:
* extend traceroute(8) to support IPv6 (by merging the existing traceroute6.c code).
* usr.sbin/traceroute6 is deleted.
* traceroute6 and traceroute6.8 are now links to traceroute.
* when invoked as traceroute6, traceroute is backward-compatible with traceroute6 syntax.
* traceroute now has new options -4 and -6.
* the manual page is updated to document new/changed options.

this is the minimal viable diff to make the merged traceroute work; there is a lot of code that could be merged or improved, and several IPv4-only options that should be supported for IPv6 as well, but for now i think it's better to get this in and tested and make other improvements in followup commits.

this commit should be squashed before merging; i didn't do this so it's not a huge diff as a single commit.  however **do not squash the first commit** unless you're sure this won't lose the history of traceroute6.c.

there is one downside to this change which is that, due to how the traceroute6 code works, we need to parse command-line args before we drop privileges.  this isn't ideal, but it's a relatively small amount of code.


llfw:
> How about just doing something very basic like:

although this works and requires less code change i'm personally not very fond of it: there's a lot of duplicated code between traceroute and traceroute6, including in option processing, and i think it makes sense to reduce that duplication as much as possible.

or to put it another way, i think if we started with your version, we'd essentially end up with my version after a few iterations anyway -- which is fine, but my version already exists.  fwiw, i intend to immediately start work on fixing the "option foo is not supported for IPv6" cases (especially obvious ones like -P and -z), but that won't be part of this PR.

regarding MFC, i'm not sure this change should be MFC'd because regardless of how you implement it, it's a breaking change for people who expect `traceroute` to do an IPv4 traceroute.  i don't really have a strong opinion on that though.

jlduran:
> > How about just doing something very basic like:
> 
> although this works and requires less code change i'm personally not very fond of it: there's a lot of duplicated code between traceroute and traceroute6, including in option processing, and i think it makes sense to reduce that duplication as much as possible.

Yes, of course! Code improvements should come afterwards, using individual commits, not at once.

> or to put it another way, i think if we started with your version, we'd essentially end up with my version after a few iterations anyway -- which is fine, but my version already exists. fwiw, i intend to immediately start work on fixing the "option foo is not supported for IPv6" cases (especially obvious ones like -P and -z), but that won't be part of this PR.

Probably yes, but it should be easier to review.

> regarding MFC, i'm not sure this change should be MFC'd because regardless of how you implement it, it's a breaking change for people who expect `traceroute` to do an IPv4 traceroute. i don't really have a strong opinion on that though.

OK, fair point!

llfw:
i've fixed the `-4 -6` issue (hopefully).

---

if i'm understanding your code right, it has one problem that if you do something like:

```
% traceroute -Picmp 2001:db8::1
```

... you will get a usage error from `traceroute6()` saying that `-P` is an unknown option. i think this is a bit confusing for users, because it's not an unknown option for `traceroute`, it's just not implemented for IPv6.

however, if you prefer to commit your version then i will redo my changes on top of it as a new PR.

jlduran:
> however, if you prefer to commit your version then i will redo my changes on top of it as a new PR.

Let's wait a bit to give other reviewers time to chime in. Thank you!

bsdimp:
Looks like this is ready to go, but I tagaged as 'needs review' based on @jlduran 's comment. Review here, or move to Phab. If you move to phab, please close this but include a link here.

llfw:
i'm closing this as it seems there is significant resistance to committing it as-is.  i'll take a different approach and redo this in phab later.

thanks @jlduran for reviewing this!

PR #1648, opened 2025-04-08
du: Fix threshold on non-directory files and regression tests

michal-os:
Fixes issue where file size threshold does not get checked for non-directory type files from PR 265200: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=265200.

Commit summary:
1. Add regression tests.
2. Fix PR 265200.
3. Cleanup to code and style. There was some duplication of code around the threshold checking and printing output for directory and non-directory cases that I refactored out.

I can squash this into a single commit if preferred.

### DIFF
--- a/usr.bin/du/du.c
+++ b/usr.bin/du/du.c
@@ -322,7 +322,9 @@ main(int argc, char *argv[])
 			    howmany(p->fts_statp->st_size, cblocksize) :
 			    howmany(p->fts_statp->st_blocks, cblocksize);
 
-			if (aflag || p->fts_level == 0) {
+			if ((aflag || p->fts_level == 0) && threshold <=
+			    threshold_sign * howmany(curblocks * cblocksize,
+			    blocksize)) {
 				xo_open_instance("paths");
 				if (hflag > 0) {
 					prthumanval("{:blocks/%4s}", curblocks);
--- a/usr.bin/du/du.c
+++ b/usr.bin/du/du.c
@@ -45,6 +45,7 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <stdbool.h>
 #include <string.h>
 #include <sysexits.h>
 #include <unistd.h>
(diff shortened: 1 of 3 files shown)


usr.bin/du/du.c
```diff
@@ -557,16 +526,51 @@ ignorep(FTSENT *ent)
...
 	info = 1;
 }
+
+/*
+ * Record the total disk/block size of the file or directory. The fts_number
+ * variable provided in FTSENT is used for keeping track of the total size.
+ * See FTS(3).
+ */
+static void
+record_file_size(FTSENT *p) {
```
markjdb, line 548: The opening brace should be on its own line.

usr.bin/du/du.c
```diff
@@ -310,37 +296,20 @@ main(int argc, char *argv[])
...
-					    (intmax_t)howmany(curblocks *
-					    cblocksize, blocksize),
-					    p->fts_path);
-				}
-				xo_close_instance("paths");
-			}
+			record_file_size(p);
 
-			p->fts_parent->fts_bignum += curblocks;
+			if ((aflag || p->fts_level == 0) && check_threshold(p))
```
markjdb, line 309: It's hard for me to quickly see that the `check_threshold(p)` check here, which uses `p->fts_number` is equivalent to the old code. Is it the case that we visit each file exactly once, so that `p->fts_number` is only modified once by record_file_size()? (And if so, why does record_file_size() use `+=` for instead of direct assignment?)

markjdb (commented):
Overall looks good to me, thank you for adding tests.

I had one question about the refactoring.

usr.bin/du/du.c
michal-os: I will fix this. The checkstyle9.pl script did not catch this. I'm not sure if I force-push a change now if it will mess up the github review. I'll push once we're happy with everything.

usr.bin/du/du.c
michal-os: > It's hard for me to quickly see that the `check_threshold(p)` check here, which uses `p->fts_number`, is equivalent to the old code.

I think the key to my refactoring is finding that if you replace `curblocks` with `p->fts_number`, then the code in the `default` case and the `FTS_DP` case start looking a lot more similar.

> Is it the case that we visit each file exactly once, so that `p->fts_number` is only modified once by record_file_size()?

Yes, the main while loop and switch statement logic makes sure we visit each file/directory only once.

Each file will call `record_file_size()` once, and for each non-directory type file (default case), it will modify it's `p->fts_number` only once. However each file and directory also adds their size to their containing parent directory's size, `p->fts_parent->fts_number`.

> And if so, why does record_file_size() use `+=` for instead of direct assignment?

The `+=` is necessary for when `record_file_size()` is called by a directory file (`case FTS_DP`), as the files inside that directory would have previously added their sizes to their parent directory `p->fts_parent->fts_number`, which for the current directory file is `p->fts_number`. So for deeper file hierarchies, the `p->fts_number` of a directory will be non-zero by the time it gets visited in the loop, unless it was an empty directory.

usr.bin/du/du.c
markjdb: In general it's ok to force-push during a review. I believe github will still attach discussions to the original diff so that it's easy what the discussion is referring to, even if it's outdated. (Phabricator doesn't do this and it makes for confusing reading.)

usr.bin/du/du.c
markjdb: Thanks for the explanation, this makes perfect sense to me now.

markjdb (approved):
Looks good to me. Once you push the style fix I'll merge the PR.

usr.bin/du/du.c
michal-os: Thanks, that is know helpful to know. And I can see now github has just labelled this as "outdated" after the force-push, that's neat.

michal-os:
I pushed the final update, thanks for reviewing!

markjdb:
Merged, thanks!

PR #1649, opened 2025-04-08
ure.4: mention RTL8156BG as a compatible controller

kenrap:
I recently purchased a [USB adapter](https://www.amazon.com/dp/B0CWV2Q6HJ) using the RTL8156BG chip and it works beautifully with the ure(4) driver. So, I thought it could be safely added in the manpage since I've been using it for the last few days without even noticing.


lwhsu:
`.Dd` needs to be bumped.

emaste (approved):

concussious (approved):
This is a fine bandaid, but would you have time or interest to fix this instead?

This driver manual is currently not conformant with style.mdoc(5) and mdoc(7). Supported hardware should be listed in a HARDWARE section that then gets pulled automatically into the [hardware release notes](https://www.freebsd.org/releases/14.2R/hardware/) so that people can search a for the hardware they have or want to buy.

As such, it is currently absent from that list. It's a small project, but it makes a good impact. I would be happy to teach you how to do this if you are interested. If not, approved anyway.

kenrap:
Sure, sign me up. I would like to do some feel-good work. Where is the best place to reach you?

concussious:
This is probably the best place.

`man mdoc` has the structure of manual pages near the top, there we see that `.Sh` (section heading) `HARDWARE` goes immediately after the end of the DESCRIPTION section. Looking in `man style.mdoc`, we see the format for the body of `.Sh HARDWARE`.

The second step is to make that section. You can do it in this branch and then it will update this pull request. I like to use `git commit -a --amend` so I don't have to fuss so much. Let me know if you have any questions!

After we iron this side out, the final step is a pull request in the doc tree to add one line to the hardware release notes template and upcoming 14.3 hardware release notes telling them to pull in the contents of the page.

kenrap:
Should the second paragraph detailing about the hardware capabilities be moved the new `.Sh HARDWARE` body as well?

concussious:
Great question! Yes. That can all be condensed into a very clean table. [Here's an example](https://reviews.freebsd.org/D49721).

The `.Bl -column` (begin column list) construct uses the length of the longest rendered item in each column in quotes to set the width each column. Each line has each column entry delimited by the `Ta` (tab) macro.

kenrap:
I think I got most of the requirements down. I appreciate the tutoring. It was a great learning exercise for me. Thanks!

Let me know if I need to adjust anything.

share/man/man4/ure.4
```diff
@@ -116,6 +108,16 @@ Force half-duplex operation.
 .Pp
 For more information on configuring this device, see
 .Xr ifconfig 8 .
+.Sh HARDWARE
+The
+.Nm
+driver provides support for the following RealTek-based USB Ethernet adapters:
+.Bl -column "Chipset" "Speed in Mbps"
```
concussious, line 115: Close! But this is considered an archaic language [^1] and even architects get column lists wrong. The "Model:" and "Speed:" entries should actually be the first line of the table, and that goes on a line beginning with `.It`(item). What should be in the `.Bl` line is the longest entry in the list, for spacing.

Even though it's duplicated, put RealTek in front of the models instead so that people can CTRL-F for the string in the hardware notes.

Reverse the order of the list so that the first entry is the current generation that the reader should buy.

You can check your work by specifying a path name to `man(1)`, e.g. `man share/man/man4/ure.4`. You can and should always pass it through the linter as well with `mandoc -Tlint /here/is/the/manual`.

[^1]: the oldest known code on file anywhere is actually RUNOFF. Roff was a rewrite for UNIX and allegedly inherited the semantics.

kenrap:
Alright, I think I'm good. Looking forward to your response.

share/man/man4/ure.4
```diff
@@ -116,6 +108,17 @@ Force half-duplex operation.
 .Pp
 For more information on configuring this device, see
 .Xr ifconfig 8 .
+.Sh HARDWARE
+The
+.Nm
+driver provides support for the following RealTek-based USB Ethernet adapters:
+.Bl -column "RealTek RTL8168/8169/8110/8211 via Xr rgephy 4" "10, 100, 1000, and 2500"
+.It Chipset: Ta Speed (Mbps):
```
concussious, line 116:
```suggestion
driver supports the following USB Ethernet controllers:
.Bl -column "Realtek RTL8168/8169/8110/8211 via rgephy(4)" "10, 100, 1000, and 2500"
.It Model: Ta Speed (Mbps):
```
Another nuance of this language is that you don't want to use macros in a list width. That has undefined behavior, even rendering as -1 space in some implementations!

Chipset is a specific term for chips that only operate in pairs. These USB Ethernet controllers are all standalone. Also, this list contains no adapters, these are the controllers that power them. Very tedious distinctions, but in this type of writing, precise accuracy is essential for credibility.

Semantic density is an equally critical part of this dry artform, being as terse as possible and not more so. If it's terse people will spend the same amount of time reading it, but more seriously, skipping over less instead of skipping too far and having to re-read, and having to keep less words in the memory buffer in their minds and consoles.

share/man/man4/ure.4
```diff
@@ -53,16 +53,8 @@ if_ure_load="YES"
...
-as well as USB Ethernet PHYs provided by
-.Xr rgephy 4 .
-.Pp
-NICs based on the RTL8152 are capable of 10 and 100Mbps.
-NICs based on the RTL8153 or provided by
-.Xr rgephy 4
-are capable of 10, 100, and 1000Mbps.
-NICs based on the RTL8156 are capable of 10, 100, 1000, and 2500Mbps operation.
+driver provides support for various RealTek-based USB Ethernet adapters listed
+in the HARDWARE section below.
```
concussious, line 57:
```suggestion
driver supports the Realtek USB Ethernet Controller family.
```
Digging deeper and looking at how Realtek (which I just realized is not spelled with a capital T) describes them [^1], maybe we should specify the actual family with even more specificity and terseness?

The FreeBSD manual pages spec says that we list hardware in the hardware section, so we probably shouldn't mention that to improve semantic density.

[^1]: https://www.realtek.com/Download/List?cate_id=585

concussious (commented):
This is shaping up really nice! Just some fairly nuanced nits left, to me.

concussious (approved):
That looks great. Since @lwhsu assigned himself, we should give him a few weeks to follow up.

Meanwhile, over in freebsd-doc, to add the driver to the hardware release notes: you're going to edit documentation/release/archetypes.adoc and /documentation/content/en/releases/14.3/hardware.adoc . Start with archetypes, that's the template. At the appropriate place near the bottom of the Ethernet Interfaces section and add a line following the pattern. Just @mention me in the review and we'll pick back up there!

kenrap:
I'm not finding either files in `documentation/release/archetypes.adoc` (there is no release directory) and `documentation/content/en/releases/14.3/hardware.adoc` (but it exists within the `website` root directory). So I'm kind of confused.

concussious:
Oops!

doc/website/archetypes/release/hardware.adoc

doc/website/content/en/releases/14.3R/hardware.adoc

bsdimp:
Looks ready to me, tagged as such, but will coordinate land with @lwhsu

lwhsu:
Oh, I don't mean to block this. I grab this because of @kenrap contacted me in the beginning. Anyway this generally looks good to me. We probably want to mention in the commit message that the layout has been reorganized.

@concussious thanks for taking care of it. Please land this, or let me know if you want me to do so.

bsdimp:
I'm doing a PR landing run, so I'll land this.

PR #1650, opened 2025-04-09
We should not skip sys.net.if_lagg_test.lacp_linkstate_destroy_stress anymore.

Rin0913:


emaste:
The associated PR states:
> I mention this PR in my commit r364220 but note that it is not 100% reliable
> at fixing the tests.

But it is probably worth re-enabling the test and see if it panics.

kprovost:
> The associated PR states:
> 
> > I mention this PR in my commit r364220 but note that it is not 100% reliable
> > at fixing the tests.
> 
That was going to be my remark too, but the entire if_lagg_test has been running in a loop for more than an hour on my test box without a panic, so I'll second your next thought. I'll do the push now.

> But it is probably worth re-enabling the test and see if it panics.

PR #1651, opened 2025-04-09
packages: add a new iputils package for ping and traceroute

llfw:
this moves them out of runtime, where they definitely don't belong. since both ping and traceroute are setuid, there is a (minor) security benefit to putting them in their own package so minimal systems don't need to install them.


llfw:
cc @bapt

bsdimp:
As per our discord conversation: get review here or close and move to Phabricator.

llfw:
i am not really sure this about one to begin with, i'm going to close it for now and i might revisit it on phab.

PR #1652, opened 2025-04-09
traceroute: add tests

llfw:
three basic tests for usr.sbin/traceroute:
- traceroute between two hosts
- ICMP traceroute between two hosts
- traceroute between two hosts with an explicit source address (-s)

to run the tests, we create three jails: one for the source host, one for the destination host, and one to route packets betweem them.  this ensures we're actually testing traceroute across a routed network and not just sending probe packets to a directly connected host.

no tests for traceroute6 are in this commit since the traceroute6 merge into traceroute is in progress.

### DIFF
--- a/usr.sbin/traceroute/tests/Makefile
+++ b/usr.sbin/traceroute/tests/Makefile
@@ -1,0 +1,7 @@
+ATF_TESTS_SH+=	traceroute_test
+
+# Allow tests to run in parallel in their own jails
+TEST_METADATA+=	execenv="jail"
+TEST_METADATA+=	execenv_jail_params="vnet allow.raw_sockets"
+
+.include <bsd.test.mk>
--- a/usr.sbin/traceroute/tests/traceroute_test.sh
+++ b/usr.sbin/traceroute/tests/traceroute_test.sh
@@ -439,7 +439,7 @@ ipv4_srcaddr_cleanup()
 atf_test_case "ipv4_srcinterface" "cleanup"
 ipv4_srcinterface_head()
 {
-	atf_set descr "IPv4 traceroute with explicit source address"
+	atf_set descr "IPv4 traceroute with explicit source interface"
 	atf_set require.user root
 }
 
--- a/usr.sbin/traceroute/tests/traceroute_test.sh
+++ b/usr.sbin/traceroute/tests/traceroute_test.sh
@@ -24,7 +24,7 @@
 # -E (detect ECN bleaching)
 # -g (source routing)
 # -n (or rather, we enable -n by default and don't test without it)
-# -P (for UDP-lite and GRE)
+# -P (for UDP-lite)
 # -r (set SO_DONTROUTE)
 # -S (print per-hop packet loss)
 # -t (set outgoing IP ToS)
@@ -586,6 +586,43 @@ ipv4_baseport_cleanup()
 }
 
 ##
+# test: ipv4_gre
+#
+
+atf_test_case "ipv4_gre" "cleanup"
+ipv4_gre_head()
+{
+	atf_set descr "IPv4 GRE traceroute"
+	atf_set require.user root
+}
+
+ipv4_gre_body()
+{
+	setup_network
+
+	start_tcpdump
+
+	# We expect the second hop to be a failure since the remote host will
+	# ignore the GRE packet.
+	atf_check -s exit:0					\
+	    -e match:"^traceroute to ${LINK_TRDST_TRDST}"	\
+	    -o match:"^ 1  ${LINK_TRSRC_TRRTR}"			\
+	    -o match:"^ 2  \\*"					\
+	    jexec trsrc traceroute $TR_FLAGS -Pgre ${LINK_TRDST_TRDST}
+
+	stop_tcpdump
+	atf_check -s exit:0 -e ignore 				\
+	    -o match:"IP \\(tos 0x0, ttl 1, .*, proto GRE .*\\).* ${LINK_TRSRC_TRSRC} > ${LINK_TRDST_TRDST}: GREv1" \
+	    -o match:"IP \\(tos 0x0, ttl 2, .*, proto GRE .*\\).* ${LINK_TRSRC_TRSRC} > ${LINK_TRDST_TRDST}: GREv1" \
+	    cat tcpdump.output
+}
+
+ipv4_gre_cleanup()
+{
+	vnet_cleanup
+}
+
+##
 # test case declarations
 
 atf_init_test_cases()
@@ -594,6 +631,7 @@ atf_init_test_cases()
 	atf_add_test_case ipv4_icmp
 	atf_add_test_case ipv4_tcp
 	atf_add_test_case ipv4_sctp
+	atf_add_test_case ipv4_gre
 	atf_add_test_case ipv4_srcaddr
 	atf_add_test_case ipv4_srcinterface
 	atf_add_test_case ipv4_maxhops
--- a/usr.sbin/traceroute/tests/traceroute_test.sh
+++ b/usr.sbin/traceroute/tests/traceroute_test.sh
@@ -24,7 +24,6 @@
 # -E (detect ECN bleaching)
 # -g (source routing)
 # -n (or rather, we enable -n by default and don't test without it)
-# -p (specify base port number)
 # -P (for UDP-lite and GRE)
 # -r (set SO_DONTROUTE)
 # -S (print per-hop packet loss)
@@ -549,6 +548,44 @@ ipv4_nprobes_cleanup()
 }
 
 ##
+# test: ipv4_baseport
+#
+
+atf_test_case "ipv4_baseport" "cleanup"
+ipv4_baseport_head()
+{
+	atf_set descr "IPv4 traceroute with non-default base port"
+	atf_set require.user root
+}
+
+ipv4_baseport_body()
+{
+	setup_network
+
+	start_tcpdump
+
+	atf_check -s exit:0					\
+	    -e match:"^traceroute to ${LINK_TRDST_TRDST}"	\
+	    -o match:"^ 1  ${LINK_TRSRC_TRRTR}"			\
+	    -o match:"^ 2  ${LINK_TRDST_TRDST}"			\
+	    -o not-match:"^ 3"					\
+	    jexec trsrc traceroute $TR_FLAGS -p 40000		\
+	    ${LINK_TRDST_TRDST}
+
+	stop_tcpdump
+
+	atf_check -s exit:0 -e ignore 				\
+	    -o match:"IP \\(tos 0x0, ttl 1, .*, proto UDP.*\\).* ${LINK_TRSRC_TRSRC}.[0-9]+ > ${LINK_TRDST_TRDST}.40001: UDP" \
+	    -o match:"IP \\(tos 0x0, ttl 2, .*, proto UDP.*\\).* ${LINK_TRSRC_TRSRC}.[0-9]+ > ${LINK_TRDST_TRDST}.40002: UDP" \
+	    cat tcpdump.output
+}
+
+ipv4_baseport_cleanup()
+{
+	vnet_cleanup
+}
+
+##
 # test case declarations
 
 atf_init_test_cases()
@@ -564,4 +601,5 @@ atf_init_test_cases()
 	atf_add_test_case ipv4_hugepacket
 	atf_add_test_case ipv4_firsthop
 	atf_add_test_case ipv4_nprobes
+	atf_add_test_case ipv4_baseport
 }
--- a/usr.sbin/traceroute/tests/traceroute_test.sh
+++ b/usr.sbin/traceroute/tests/traceroute_test.sh
@@ -125,6 +125,8 @@ start_tcpdump()
 		interface="${epsrc}b"
 	fi
 
+	rm -f "${PWD}/traceroute.pcap"
+
 	jexec trrtr daemon -p "${PWD}/tcpdump.pid" \
 	    tcpdump --immediate-mode -w "${PWD}/traceroute.pcap" -nv \
 	    -i $interface
@@ -290,6 +292,8 @@ ipv4_tcp_body()
 {
 	setup_network
 
+	start_tcpdump
+
 	# We expect the second hop to be a failure since traceroute doesn't
 	# know how to capture the RST packet.
 	atf_check -s exit:0					\
@@ -297,6 +301,12 @@ ipv4_tcp_body()
 	    -o match:"^ 1  ${LINK_TRSRC_TRRTR}"			\
 	    -o match:"^ 2  \\*"					\
 	    jexec trsrc traceroute $TR_FLAGS -Ptcp ${LINK_TRDST_TRDST}
+
+	stop_tcpdump
+	atf_check -s exit:0 -e ignore 				\
+	    -o match:"IP \\(tos 0x0, ttl 1, .*, proto TCP.*\\).* ${LINK_TRSRC_TRSRC}.[0-9]+ > ${LINK_TRDST_TRDST}.33435: Flags \[S\]" \
+	    -o match:"IP \\(tos 0x0, ttl 2, .*, proto TCP.*\\).* ${LINK_TRSRC_TRSRC}.[0-9]+ > ${LINK_TRDST_TRDST}.33436: Flags \[S\]" \
+	    cat tcpdump.output
 }
 
 ipv4_tcp_cleanup()
@@ -319,6 +329,8 @@ ipv4_srcaddr_body()
 {
 	setup_network
 
+	start_tcpdump
+
 	atf_check -s exit:0				\
 	    -e match:"^traceroute to ${LINK_TRDST_TRDST} \\($LINK_TRDST_TRDST\\) from ${LINK_TRSRC2_TRSRC}" \
 	    -o match:"^ 1  ${LINK_TRSRC2_TRRTR}"	\
@@ -326,6 +338,12 @@ ipv4_srcaddr_body()
 	    -o not-match:"^ 3"				\
 	    jexec trsrc traceroute $TR_FLAGS		\
 	        -s ${LINK_TRSRC2_TRSRC} ${LINK_TRDST_TRDST}
+
+	stop_tcpdump
+	atf_check -s exit:0 -e ignore 				\
+	    -o match:"IP \\(tos 0x0, ttl 1, .*, proto UDP.*\\).* ${LINK_TRSRC2_TRSRC}.[0-9]+ > ${LINK_TRDST_TRDST}.33435: UDP" \
+	    -o match:"IP \\(tos 0x0, ttl 2, .*, proto UDP.*\\).* ${LINK_TRSRC2_TRSRC}.[0-9]+ > ${LINK_TRDST_TRDST}.33436: UDP" \
+	    cat tcpdump.output
 }
 
 ipv4_srcaddr_cleanup()
@@ -348,6 +366,8 @@ ipv4_srcinterface_body()
 {
 	setup_network
 
+	start_tcpdump
+
 	# Unlike -s, traceroute doesn't print 'from ...' when using -i.
 	atf_check -s exit:0					\
 	    -e match:"^traceroute to ${LINK_TRDST_TRDST}"	\
@@ -356,6 +376,12 @@ ipv4_srcinterface_body()
 	    -o not-match:"^ 3"					\
 	    jexec trsrc traceroute $TR_FLAGS			\
 	        -i ${epsrc2}a ${LINK_TRDST_TRDST}
+
+	stop_tcpdump
+	atf_check -s exit:0 -e ignore 				\
+	    -o match:"IP \\(tos 0x0, ttl 1, .*, proto UDP.*\\).* ${LINK_TRSRC2_TRSRC}.[0-9]+ > ${LINK_TRDST_TRDST}.33435: UDP" \
+	    -o match:"IP \\(tos 0x0, ttl 2, .*, proto UDP.*\\).* ${LINK_TRSRC2_TRSRC}.[0-9]+ > ${LINK_TRDST_TRDST}.33436: UDP" \
+	    cat tcpdump.output
 }
 
 ipv4_srcinterface_cleanup()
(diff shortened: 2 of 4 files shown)


usr.sbin/traceroute/tests/traceroute4_test.sh
```diff
@@ -0,0 +1,128 @@
...
+
+setup_network()
+{
+	# Create 3 jails: one to be the source host, one to be the router,
+	# and one to be the destination host.
+
+	vnet_init
+
+	# src jail
+	epsrc=$(vnet_mkepair)
```
jlduran, line 29: These names are a bit cryptic, but we'll get used to them:
- `trrtr`
- `trsrc`
- `trdst`
- `epsrc`
- `epdst`

usr.sbin/traceroute/tests/traceroute4_test.sh
```diff
@@ -0,0 +1,128 @@
...
+
+	# dst jail
+	epdst=$(vnet_mkepair)
+	vnet_mkjail trdst ${epdst}a
+
+	# router jail
+	vnet_mkjail trrtr ${epsrc}b ${epdst}b
+
+	# Configure IPv4 addresses and routes on each jail
+	jexec trsrc ifconfig ${epsrc}a inet 192.0.2.2/31 up
```
jlduran, line 40: There is no need for such scarcity, how about using a `/25`? If there is some strange issue with the local subnet/broadcast address we have a better chance catching it. Also is probably closer to a standard setup.

usr.sbin/traceroute/tests/traceroute4_test.sh
```diff
@@ -0,0 +1,128 @@
...
+	jexec trrtr ifconfig ${epsrc}b inet 192.0.2.3/31 up
+	jexec trsrc route add -inet 192.0.2.0/24 192.0.2.3
+
+	jexec trdst ifconfig ${epdst}a inet 192.0.2.4/31 up
+	jexec trrtr ifconfig ${epdst}b inet 192.0.2.5/31 up
+	jexec trdst route add -inet 192.0.2.0/24 192.0.2.5
+
+	jexec trrtr sysctl -w net.inet.ip.forwarding=1
```
jlduran, line 48: The `-w` option is deprecated, and it is silently ignored.

jlduran (changes requested):
Thank you for adding tests!

Sorry if I sound a bit too picky, remember everything is open to discussion:

1. The tests need an mtree entry in etc/mtree/BSD.tests.dist.
2. Given you are using a common script to create jails, `setup_network` will use the same jail names on each test, preventing them from running in parallel. Recently, a nice addition was committed (257e70f1d5ee61037c8c59b116538d3b6b1427a2) that could help with this issue.

usr.sbin/traceroute/tests/Makefile
```diff
@@ -0,0 +1,3 @@
+ATF_TESTS_SH+=		traceroute4_test
+
```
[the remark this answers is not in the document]
llfw: is this necessary?  i was basing my changes on `usr.sbin/chown/tests` which doesn't include `PACKAGE=` but the tests still end up in the right package.

usr.sbin/traceroute/tests/traceroute4_test.sh
llfw: i am happy to rename these, i basically picked the current names to save too much typing :-)

usr.sbin/traceroute/tests/traceroute4_test.sh
llfw: i picked a /31 since that's what i use for link nets in real life, but i don't mind changing it.  a /30 seems more reasonable though, unless there's a particular reason to make it larger?

usr.sbin/traceroute/tests/traceroute4_test.sh
```diff
@@ -0,0 +1,128 @@
...
+	atf_set descr "Basic IPv4 ICMP traceroute across a router"
+	atf_set require.user root
+}
+
+icmp_body()
+{
+	setup_network
+
+	atf_check -s exit:0 -o ignore -e ignore \
+		jexec trsrc $TRACEROUTE -I -n -s 192.0.2.2 192.0.2.4
```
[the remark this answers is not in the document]
llfw: this `-s` was meant to be in the `srcaddr` test, i'll fix that.

actually, i will change the srcaddr test to use a non-default source address since it's not actually testing anything right now exception getopt.

usr.sbin/traceroute/tests/traceroute4_test.sh
jlduran: Do not worry, we'll get used to them!

usr.sbin/traceroute/tests/traceroute4_test.sh
jlduran: A /30 would be just fine, thank you!

usr.sbin/traceroute/tests/Makefile
```diff
@@ -0,0 +1,3 @@
+ATF_TESTS_SH+=		traceroute4_test
```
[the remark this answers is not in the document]
llfw: okay, i'll redo it this way (for now only with `traceroute_test:ipv4_*`).

usr.sbin/traceroute/tests/traceroute4_test.sh
jlduran: Even better would be to use 198.51.100.0/24 for the destination network (make it a bit easier non the eyes).

llfw: i've changed this to a /30.  i could change it again to a /24 but i don't think there's any point, is there? i'm sure anyone interested in traceroute knows how /30s work.

usr.sbin/traceroute/tests/traceroute4_test.sh
llfw: fixed

usr.sbin/traceroute/tests/traceroute4_test.sh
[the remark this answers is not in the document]
llfw: fixed, and i improved the srcaddr test as i described

usr.sbin/traceroute/tests/traceroute4_test.sh
```diff
@@ -0,0 +1,128 @@
...
+	atf_set descr "IPv4 traceroute with explicit source address"
+	atf_set require.user root
+}
+
+srcaddr_body()
+{
+	setup_network
+
+	atf_check -s exit:0 -o ignore -e ignore \
+		jexec trsrc $TRACEROUTE -n 192.0.2.4
```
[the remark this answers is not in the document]
llfw: fixed

llfw:
i've added test metadata to make the tests run in a jail.  i'm not entirely sure how to test this, but `kyua test -k /usr/tests/usr.sbin/traceroute/Kyuafile` still works fine.  let me know if any changes are needed here.

jlduran:
Very nice, thank you!

There are a few nits that still bother me:

1. The repetition of the test, essentially every test runs twice, the second time just to parse it. `atf_check` accepts `-o save:<name of file>`. We can leverage this to avoid such repetition.
2. The `-s` test, I have a hard time swallowing it. I would prefer if `trsrc` would have a secondary interface, with a "real" network. As a side effect, I believe the test result that comes out is much more evident.

Something along these lines:
```diff
diff --git a/usr.sbin/traceroute/tests/traceroute_test.sh b/usr.sbin/traceroute/tests/traceroute_test.sh
index f1afa38a8bcd..2887109f357f 100644
--- a/usr.sbin/traceroute/tests/traceroute_test.sh
+++ b/usr.sbin/traceroute/tests/traceroute_test.sh
@@ -21,9 +21,6 @@ TRACEROUTE="traceroute -w 1 -q 1 -m 5"
 # The prefix our test networks are in.
 TEST_PREFIX="192.0.2.0/24"
 
-# The loopback address of trsrc (used for the -s test).
-TRSRC_LOOPBACK="192.0.2.255"
-
 # The IPv4 addresses of the link net between trsrc and trrtr.
 LINK_TRSRC_TRSRC="192.0.2.5"
 LINK_TRSRC_TRRTR="192.0.2.6"
@@ -34,6 +31,16 @@ LINK_TRDST_TRDST="192.0.2.9"
 LINK_TRDST_TRRTR="192.0.2.10"
 LINK_TRDST_PREFIXLEN="30"
 
+# The IPv4 addresses of the second link net between trsrc and trrtr.
+LINK_TRSRC_TRSRC2="192.0.2.253"
+LINK_TRSRC_TRRTR2="192.0.2.254"
+LINK_TRSRC_PREFIXLEN2="30"
+
+parsed_traceroute_stdout()
+{
+	cat stdout | awk -vORS=, '/^ *([0-9]+)/ { print $2 }'
+}
+
 setup_network()
 {
 	# Create 3 jails: one to be the source host, one to be the router,
@@ -43,23 +50,26 @@ setup_network()
 
 	# src jail
 	epsrc=$(vnet_mkepair)
-	vnet_mkjail trsrc ${epsrc}a
+	epsrc2=$(vnet_mkepair)
+	vnet_mkjail trsrc ${epsrc}a ${epsrc2}a
 
 	# dst jail
 	epdst=$(vnet_mkepair)
 	vnet_mkjail trdst ${epdst}a
 
 	# router jail
-	vnet_mkjail trrtr ${epsrc}b ${epdst}b
+	vnet_mkjail trrtr ${epsrc}b ${epsrc2}b ${epdst}b
 
 	# Configure IPv4 addresses and routes on each jail
-	jexec trsrc ifconfig lo0 ${TRSRC_LOOPBACK}/32 alias
 	jexec trsrc ifconfig ${epsrc}a inet \
 		${LINK_TRSRC_TRSRC}/${LINK_TRSRC_PREFIXLEN}
+	jexec trsrc ifconfig ${epsrc2}a inet \
+		${LINK_TRSRC_TRSRC2}/${LINK_TRSRC_PREFIXLEN2}
 	jexec trrtr ifconfig ${epsrc}b inet \
 		${LINK_TRSRC_TRRTR}/${LINK_TRSRC_PREFIXLEN}
+	jexec trrtr ifconfig ${epsrc2}b inet \
+		${LINK_TRSRC_TRRTR2}/${LINK_TRSRC_PREFIXLEN2}
 	jexec trsrc route add -inet ${TEST_PREFIX} ${LINK_TRSRC_TRRTR}
-	jexec trrtr route add -inet ${TRSRC_LOOPBACK}/32 ${LINK_TRSRC_TRSRC}
 
 	jexec trdst ifconfig ${epdst}a inet \
 		${LINK_TRDST_TRDST}/${LINK_TRDST_PREFIXLEN}
@@ -81,12 +91,10 @@ ipv4_basic_body()
 {
 	setup_network
 
-	atf_check -s exit:0 -o ignore -e ignore jexec trsrc \
+	atf_check -s exit:0 -o save:stdout -e ignore jexec trsrc \
 		$TRACEROUTE -n ${LINK_TRDST_TRDST}
 
-	path=$(jexec trsrc $TRACEROUTE -n ${LINK_TRDST_TRDST} 2>&1 \
-		|awk -vORS=, '/^ *([0-9]+)/ { print $2 }')
-	atf_check_equal "$path" "${LINK_TRSRC_TRRTR},${LINK_TRDST_TRDST},"
+	atf_check_equal $(parsed_traceroute_stdout) "${LINK_TRSRC_TRRTR},${LINK_TRDST_TRDST},"
 }
 
 ipv4_basic_cleanup()
@@ -105,12 +113,10 @@ ipv4_icmp_body()
 {
 	setup_network
 
-	atf_check -s exit:0 -o ignore -e ignore jexec trsrc	\
+	atf_check -s exit:0 -o save:stdout -e ignore jexec trsrc	\
 		$TRACEROUTE -I -n ${LINK_TRDST_TRDST}
 
-	path=$(jexec trsrc $TRACEROUTE -I -n ${LINK_TRDST_TRDST} 2>&1 	\
-		| awk -vORS=, '/^ *([0-9]+)/ { print $2 }')
-	atf_check_equal "$path" "${LINK_TRSRC_TRRTR},${LINK_TRDST_TRDST},"
+	atf_check_equal $(parsed_traceroute_stdout) "${LINK_TRSRC_TRRTR},${LINK_TRDST_TRDST},"
 }
 
 ipv4_icmp_cleanup()
@@ -129,13 +135,10 @@ ipv4_srcaddr_body()
 {
 	setup_network
 
-	atf_check -s exit:0 -o ignore -e ignore jexec trsrc \
-		$TRACEROUTE -n -s ${TRSRC_LOOPBACK} ${LINK_TRDST_TRDST}
+	atf_check -s exit:0 -o save:stdout -e ignore jexec trsrc \
+		$TRACEROUTE -n -s ${LINK_TRSRC_TRSRC2} ${LINK_TRDST_TRDST}
 
-	path=$(jexec trsrc $TRACEROUTE -n -s ${TRSRC_LOOPBACK}	\
-			${LINK_TRDST_TRDST} 2>&1		\
-		| awk -vORS=, '/^ *([0-9]+)/ { print $2 }')
-	atf_check_equal "$path" "${LINK_TRSRC_TRRTR},${LINK_TRDST_TRDST},"
+	atf_check_equal $(parsed_traceroute_stdout) "${LINK_TRSRC_TRRTR2},${LINK_TRDST_TRDST},"
 }
 
 ipv4_srcaddr_cleanup()
```

P.S.: Out of ignorance, what are the advantages of releasing your code under the ISC vs BSD-2-clause?

usr.sbin/traceroute/tests/Makefile
```diff
@@ -0,0 +1,5 @@
+ATF_TESTS_SH+=		traceroute_test
+
+TEST_METADATA+= execenv="jail" execenv_jail_params="vnet allow.raw_sockets"
```
jlduran, line 3: Technically correct, but it is clearer if each one is on a separate line:
```make
# Allow tests to run in parallel in their own jails
TEST_METADATA+= execenv="jail"
TEST_METADATA+= execenv_jail_params="vnet allow.raw_sockets"
```

jlduran (commented):
Let's wait some time for others to review as well.
Feel free to keep adding tests.
I appreciate the effort!

usr.sbin/traceroute/tests/traceroute4_test.sh
```diff
@@ -0,0 +1,128 @@
...
+# WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+# ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+# WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+# ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+# OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+. $(atf_get_srcdir)/../../sys/common/vnet.subr
+
+TRACEROUTE="traceroute -w 1 -q 1 -m 5"
```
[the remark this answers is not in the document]
jlduran: Maybe something like?:
```
TR_FAST_FLAGS="-w 1 -q 1 -m 5"
```

usr.sbin/traceroute/tests/Makefile
llfw: fixed

usr.sbin/traceroute/tests/traceroute4_test.sh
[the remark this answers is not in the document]
llfw: i've added `$TR_FLAGS` and put `-n` there; we'll just not use `$TR_FLAGS` for the test that doesn't want `-n`.

i think this is a more reasonable approach because in other tests we'll want to drop `-m` (for example) and it's better to be consistent about that for all the flags.

jlduran (commented):
I think most (if not all) of the `\` at the end of the lines are not needed.
Per style(9), to indicate continuation on the next line, we use 4 spaces instead of a tab. For a shell script, I don't mind, as long as it's consistent.

llfw:
> I think most (if not all) of the `\` at the end of the lines are not needed.

sorry, which `\` are you talking about here?  the only places i've used `\` is to split commands across multiple lines; i think this is more readable than trying to put the entire command on a single line.

> Per style(9), to indicate continuation on the next line, we use 4 spaces instead of a tab. For a shell script, I don't mind, as long as it's consistent.

okay, i can change this.

jlduran:
> i can't add a test for hostname resolution because there seems to be no way to override the location of `/etc/hosts` via an environment variable, and we can't modify it because these vnet jails use `path=/`.
> 
> @jlduran any ideas here?
> 
> the best thing i can come up with is to create a chroot environment for that test and provide a custom `/etc/hosts`, which would probably work, but doesn't seem like the ideal approach.

At this point shell scripts start becoming convoluted. If you still want to keep using atf-sh, then a Sacpy-based python script helper, faking a DNS server is probably the way to go, otherwise use it directly with an atf-python script.

llfw:
> If you still want to keep using atf-sh, then a Sacpy-based python script helper, faking a DNS server is probably the way to go

how would you go about faking a DNS resolver - is there a way to override the location of `/etc/resolv.conf`?

jlduran:
> > If you still want to keep using atf-sh, then a Sacpy-based python script helper, faking a DNS server is probably the way to go
> 
> how would you go about faking a DNS resolver - is there a way to override the location of `/etc/resolv.conf`?

You can intercept the packet and doctor it. I did something similar with test_ping.py. Or fake everything, as we are not interested in testing name resolution. But, let me think about it.

usr.sbin/traceroute/tests/traceroute_test.sh
```diff
@@ -0,0 +1,365 @@
...
+LINK_TRSRC2_TRRTR="192.0.2.14"
+LINK_TRSRC2_PREFIXLEN="30"
+
+# The IPv4 addresses of the link net between trdst and trrtr.
+LINK_TRDST_TRDST="192.0.2.9"
+LINK_TRDST_TRRTR="192.0.2.10"
+LINK_TRDST_PREFIXLEN="30"
+
+# This is an address inside $TEST_PREFIX which is not routed anywhere.
+UNREACHABLE_ADDR="192.0.2.255"
```
jlduran, line 46: Would you mind not using a potential broadcast address here?

llfw: classful addressing has been obsolete for 30 years... we know this isn't a broadcast address since we configured all the test subnets ourselves.  (although we add a route for 192.0.2.0/24, it doesn't have a broadcast address since it's a prefix, not a subnet.)

i will change this if you feel strongly about it, but under protest :-)

jlduran: > classful addressing has been obsolete for 30 years... we know this isn't a broadcast address since we configured all the test subnets ourselves. (although we add a route for 192.0.2.0/24, it doesn't have a broadcast address since it's a prefix, not a subnet.)
> 
> i will change this if you feel strongly about it, but under protest :-)

I am not that old 😅! What is the broadcast address of the subnet 192.0.2.252/30?

llfw: we don't use 192.0.2.252/30 anywhere so i don't see any need to avoids its network/broadcast addresses.  following this logic, we'd have to arbitrarily avoid half the available address, which is surely far more confusing:

- 192.0.2.255 is bad
- 192.0.2.254 is okay
- 192.0.2.253 is okay
- 192.0.2.252 is bad
- 192.0.2.251 is bad
- 192.0.2.250 is okay
- 192.0.2.249 is okay
- 192.0.2.248 is bad
- ...

this seems a lot more complicated than just starting at 192.0.2.255/32 and working backwards if we need to allocate more /32s.  there's no risk of conflict unless we end up using so many addresses that the two meet in the middle, and if that happens, we're going to have other problems unrelated to (re)using someone's broadcast address.

as a compromise, i would be willing to change this to allocate /32s from another prefix like 198.18.0.0/23 (the benchmarking prefix) but i think this is wrong, since we're doing testing, not benchmarking.

i am inclined to wait to see what reviewer 2 says about this.

jlduran: I guess the thing is, I would prefer to be explicit.
Have a network specifically designed for this purpose, with no hosts in it and try to traceroute to one of them, so the last hop replies with an ICMP host unreachable.

---

Do you have more tests coming in (besides those requiring name resolution)?
Let's put a deadline for reviewer 2, otherwise, I'll commit this.
From my part, everything is OK, maybe still having doubts about the cryptic names, but those can be changed afterwards if deemed necessary.

jlduran: There is no rush. I'll wait a week(ish) then. Thank you!

llfw:
whenever you are ready to commit this, let me know and i'll squash the branch into a single commit with an appropriate commit message; i don't see any reason we need more than 1 commit here.

jlduran:
> whenever you are ready to commit this, let me know and i'll squash the branch into a single commit with an appropriate commit message; i don't see any reason we need more than 1 commit here.

I'll be committing in a week(ish) or so if nobody comments. Please do squash whenever you consider it ready.

llfw:
i've squashed this since i don't think we can usefully test the rest of the options without inspecting the packets we're sending.  as noted in the commit message, i've put a list of things we don't test in `traceroute_test.sh`.

if/when the traceroute6 merge is done i'll open a new PR to add IPV6 tests.

llfw:
i'm not sure i like that, and it also makes the tests non-parallelisable again.  but i'll have a think... in either case i'll submit that as a separate PR.  i'll probably do the other flags first though (with packet inspection) since those are more likely to break without being noticed than DNS.

llfw:
i've improved the ipv4_icmp test to use tcpdump to verify traceroute actually sent the packets it should do.  if you approve of this approach, i'll modify the other tests to do the same thing, otherwise we can talk about other ways to do this.  (i am aware of scapy, but i'm not sure it's easier here?)

jlduran:
It's good!
As a side note, the env variable `$PWD` can be used directly.

usr.sbin/traceroute/tests/traceroute_test.sh
```diff
@@ -0,0 +1,880 @@
...
+}
+
+##
+# test: ipv4_srcinterface
+#
+
+atf_test_case "ipv4_srcinterface" "cleanup"
+ipv4_srcinterface_head()
+{
+	atf_set descr "IPv4 traceroute with explicit source address"
```
jlduran, line 442: I believe it should be: "... explicit source **interface**"

jlduran:
Hey! I've noticed you are now submitting patches over Phabricator, would you prefer to move this one over there? (tagging network and tests). Visibility should be higher. Thank you!

llfw:
sure, i'll do that now.  i've had a Phab account for ages, i just default to GitHub since the contributing docs say to use that.

jlduran:
> sure, i'll do that now. i've had a Phab account for ages, i just default to GitHub since the contributing docs say to use that.

Thank you! It's subjective, depending on the size/complexity of the patch. See you there!

llfw:
closing this as moved to phab: https://reviews.freebsd.org/D49838

PR #1654, opened 2025-04-09
column(1): add tests

llfw:
note that the `colonic_table` test will fail until #1599 is committed, due to an existing bug in column.


llfw:
moved to phabricator: https://reviews.freebsd.org/D49911

PR #1656, opened 2025-04-10
libc/tests: Enable fpsetround_basic which was never triggered since ported.

Rin0913:
ieeefp library is still used for freebsd i386.

However, the tests require a definition named `_FLOAT_IEEE754`, which is not ported from NetBSD into FreeBSD source tree.

Therefore, I created this pr to enable the tests for i386.

Due to the difference usage of APIs, I localized the tests.


lib/libc/tests/gen/fpclassify_test.c
```diff
@@ -0,0 +1,209 @@
...
+
+#ifndef __i386__
+
+ATF_TC(no_test);
+ATF_TC_HEAD(no_test, tc)
+{
+	atf_tc_set_md_var(tc, "descr", "Dummy test");
+}
+
+ATF_TC_BODY(no_test,tc)
```
bsdimp, line 44: A space here would be better.

Rin0913: It's copied from the netbsd-test.
I prefer to leave it as it is.
However, if you insist, I will fix this.

bsdimp: That's fine.

bsdimp:

ngie-eign:
@Rin0913 : what is the reason for copying the test wholesale out of contrib/netbsd-tests ? `share/mk/netbsd-tests.test.mk` handles the name conversion automatically.

Rin0913:
@ngie-eign As I was confirmed, files under `contrib/` should not be modified. If needed, to copy it out for further modification.

PR #1658, opened 2025-04-11
C++ style discussion

bsdjhb:
I do not plan to merge this as-is but instead want to use this to discuss how we might want to augment style(9) with some C++-specific extensions.  I figure it might be easier to do that by commenting on some sample code and later turning that into prose and examples added to style(9).

### DIFF
--- a/contrib/libucl/include/ucl++.h
+++ b/contrib/libucl/include/ucl++.h
@@ -396,6 +396,11 @@ public:
 		return default_val;
 	}
 
+	std::string forced_string_value () const
+	{
+		return ucl_object_tostring_forced(obj.get());
+	}
+
 	size_t size () const
 	{
 		if (type () == UCL_ARRAY) {
--- a/lib/libiscsiutil/log.c
+++ b/lib/libiscsiutil/log.c
@@ -207,6 +207,12 @@ log_warnx(const char *fmt, ...)
 }
 
 void
+log_vwarnx(const char *fmt, va_list ap)
+{
+	log_common(LOG_WARNING, -1, fmt, ap);
+}
+
+void
 log_debugx(const char *fmt, ...)
 {
 	va_list ap;
--- a/lib/libutil++/libutil++
+++ b/lib/libutil++/libutil++
@@ -9,6 +9,7 @@
 #define	__LIBUTILPP__
 
 #include <sys/cdefs.h>
+#include <netdb.h>
 #include <stdarg.h>
 
 #include <memory>
@@ -26,6 +27,20 @@ namespace freebsd {
 	};
 
 	typedef std::unique_ptr<FILE, fclose_deleter> FILE_up;
+
+	/*
+	 * addrinfo_up is a std::unique_ptr<> which uses
+	 * freeaddrinfo() to destroy the wrapped pointer.  It is
+	 * intended to wrap arrays allocated by getaddrinfo().
+	 */
+	struct freeaddrinfo_deleter {
+		void operator() (struct addrinfo *ai) const
+		{
+			freeaddrinfo(ai);
+		}
+	};
+
+	typedef std::unique_ptr<addrinfo, freeaddrinfo_deleter> addrinfo_up;
 
 	/*
 	 * malloc_up<T> is a std::unique_ptr<> which uses free() to
--- a/lib/libutil++/libutil++
+++ b/lib/libutil++/libutil++
@@ -9,6 +9,7 @@
 #define	__LIBUTILPP__
 
 #include <sys/cdefs.h>
+#include <sys/nv.h>
 #include <netdb.h>
 #include <stdarg.h>
 #include <unistd.h>
@@ -135,6 +136,19 @@ namespace freebsd {
 
 	template <class T>
 	using malloc_up = std::unique_ptr<T, free_deleter<T>>;
+
+	/*
+	 * nvlist_up is a std::unique_ptr<> for nvlist_t objects which
+	 * uses nvlist_destroy() to destroy the wrapped pointer.
+	 */
+	struct nvlist_deleter {
+		void operator() (nvlist_t *nvl) const
+		{
+			nvlist_destroy(nvl);
+		}
+	};
+
+	typedef std::unique_ptr<nvlist_t, nvlist_deleter> nvlist_up;
 
 	/*
 	 * Returns a std::string containing the same output as
--- a/lib/libutil++/libutil++
+++ b/lib/libutil++/libutil++
@@ -10,6 +10,7 @@
 
 #include <sys/cdefs.h>
 #include <sys/nv.h>
+#include <libutil.h>
 #include <netdb.h>
 #include <stdarg.h>
 #include <unistd.h>
@@ -149,6 +150,84 @@ namespace freebsd {
 	};
 
 	typedef std::unique_ptr<nvlist_t, nvlist_deleter> nvlist_up;
+
+	/*
+	 * A wrapper class for the pidfile_* API.  The destructor
+	 * calls pidfile_remove() when an object is destroyed.  This
+	 * class is similar to std::unique_ptr<> in that it retains
+	 * exclusive ownership of the pidfh object.
+	 *
+	 * In addition to release() and reset methods(), write(),
+	 * close(), and fileno() methods are provided as wrappers for
+	 * pidfile_*.
+	 */
+	class pidfile {
+	public:
+		pidfile() : _pfh(nullptr) {}
+		pidfile(struct pidfh *pfh) : _pfh(pfh) {}
+		pidfile(pidfile &&other) : _pfh(other.release()) {}
+		pidfile(pidfile const &) = delete;
+
+		~pidfile() { reset(); }
+
+		struct pidfh *
+		release()
+		{
+			struct pidfh *pfh = _pfh;
+
+			_pfh = nullptr;
+			return (pfh);
+		}
+
+		void
+		reset(struct pidfh *pfh = nullptr)
+		{
+			if (_pfh != nullptr)
+				pidfile_remove(_pfh);
+			_pfh = pfh;
+		}
+
+		int
+		write()
+		{
+			return (pidfile_write(_pfh));
+		}
+
+		int
+		close()
+		{
+			int rv = pidfile_close(_pfh);
+			if (rv == 0)
+				_pfh = nullptr;
+			return (rv);
+		}
+
+		int
+		fileno()
+		{
+			return (pidfile_fileno(_pfh));
+		}
+
+		pidfile &operator=(pidfile &&other) noexcept
+		{
+			if (this == &other)
+				return *this;
+			reset(other.release());
+			return *this;
+		}
+
+		pidfile &operator=(pidfile const &) = delete;
+
+		pidfile &operator=(struct pidfh *pfh)
+		{
+			reset(pfh);
+			return *this;
+		}
+
+		explicit operator bool() const { return _pfh != nullptr; }
+	private:
+		struct pidfh *_pfh = nullptr;
+	};
 
 	/*
 	 * Returns a std::string containing the same output as
--- a/lib/libutil++/libutil++
+++ b/lib/libutil++/libutil++
@@ -1,0 +1,55 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2025 Chelsio Communications, Inc.
+ * Written by: John Baldwin <jhb@FreeBSD.org>
+ */
+
+#ifndef __LIBUTILPP__
+#define	__LIBUTILPP__
+
+#include <sys/cdefs.h>
+#include <stdarg.h>
+
+#include <memory>
+
+namespace freebsd {
+	/*
+	 * FILE_up is a std::unique_ptr<> for FILE objects which uses
+	 * fclose() to destroy the wrapped pointer.
+	 */
+	struct fclose_deleter {
+		void operator() (FILE *fp) const
+		{
+			fclose(fp);
+		}
+	};
+
+	typedef std::unique_ptr<FILE, fclose_deleter> FILE_up;
+
+	/*
+	 * malloc_up<T> is a std::unique_ptr<> which uses free() to
+	 * destroy the wrapped pointer.  This can be used to wrap
+	 * pointers allocated implicitly by malloc() such as those
+	 * returned by strdup().
+	 */
+	template <class T>
+	struct free_deleter {
+		void operator() (T *p) const
+		{
+			free(p);
+		}
+	};
+
+	template <class T>
+	using malloc_up = std::unique_ptr<T, free_deleter<T>>;
+
+	/*
+	 * Returns a std::string containing the same output as
+	 * sprintf().  Throws std::bad_alloc if an error occurs.
+	 */
+	std::string stringf(const char *fmt, ...) __printflike(1, 2);
+	std::string stringf(const char *fmt, va_list ap);
+}
+
+#endif /* !__LIBUTILPP__ */
--- a/lib/libutil++/libutil++
+++ b/lib/libutil++/libutil++
@@ -11,6 +11,7 @@
 #include <sys/cdefs.h>
 #include <netdb.h>
 #include <stdarg.h>
+#include <unistd.h>
 
 #include <memory>
 
@@ -41,6 +42,82 @@ namespace freebsd {
 	};
 
 	typedef std::unique_ptr<addrinfo, freeaddrinfo_deleter> addrinfo_up;
+
+	/*
+	 * This class is intended to function similar to unique_ptr<>,
+	 * but it contains a file descriptor rather than a pointer to
+	 * an object.  On destruction the descriptor is closed via
+	 * close(2).
+	 *
+	 * Similar to unique_ptr<>, release() returns ownership of the
+	 * file descriptor to the caller.  reset() closes the current
+	 * file descriptor and takes ownership of a new one.  A move
+	 * constructor permits ownership to be transferred via
+	 * std::move().  An integer file descriptor can be assigned
+	 * directly which is equivalent to calling reset().
+	 *
+	 * An explicit bool conversion operator permits testing this
+	 * class in logical expressions.  It returns true if it
+	 * contains a valid descriptor.
+	 *
+	 * An implicit int conversion operator returns the underlying
+	 * file descriptor allowing objects of this type to be passed
+	 * directly to APIs such as connect(), listen(), etc.
+	 */
+	class fd_up {
+	public:
+		fd_up() : _fd(-1) {}
+		fd_up(int fd) : _fd(fd) {}
+		fd_up(fd_up &&other) : _fd(other.release()) {}
+		fd_up(fd_up const &) = delete;
+
+		~fd_up() { reset(); }
+
+		int
+		get() const
+		{
+			return (_fd);
+		}
+
+		int
+		release()
+		{
+			int fd = _fd;
+
+			_fd = -1;
+			return (fd);
+		}
+
+		void
+		reset(int newfd = -1)
+		{
+			if (_fd >= 0)
+				close(_fd);
+			_fd = newfd;
+		}
+
+		fd_up &operator=(fd_up &&other) noexcept
+		{
+			if (this == &other)
+				return *this;
+
+			reset(other.release());
+			return *this;
+		}
+
+		fd_up &operator=(fd_up const &) = delete;
+
+		fd_up &operator=(int fd)
+		{
+			reset(fd);
+			return *this;
+		}
+
+		explicit operator bool() const { return _fd >= 0; }
+		operator int() const { return _fd; }
+	private:
+		int	_fd;
+	};
 
 	/*
 	 * malloc_up<T> is a std::unique_ptr<> which uses free() to
--- a/lib/libutil++/stringf.cc
+++ b/lib/libutil++/stringf.cc
@@ -1,0 +1,57 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2025 Chelsio Communications, Inc.
+ * Written by: John Baldwin <jhb@FreeBSD.org>
+ */
+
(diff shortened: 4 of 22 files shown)


lib/libutil++/libutil++
```diff
@@ -0,0 +1,240 @@
...
+#include <sys/cdefs.h>
+#include <sys/nv.h>
+#include <libutil.h>
+#include <netdb.h>
+#include <stdarg.h>
+#include <unistd.h>
+
+#include <memory>
+
+namespace freebsd {
```
bsdjhb, line 20: Using cuddle-braces for namespaces seems consistent

lib/libutil++/libutil++
```diff
@@ -0,0 +1,240 @@
...
+	/*
+	 * FILE_up is a std::unique_ptr<> for FILE objects which uses
+	 * fclose() to destroy the wrapped pointer.
+	 */
+	struct fclose_deleter {
+		void operator() (FILE *fp) const
```
bsdjhb, line 26: This is an interesting case.  This style is commonly used in other C++ code that I see, but I think my suggestion is probably more correct?
```suggestion
		void
		operator() (FILE *fp) const
```

lib/libutil++/libutil++
```diff
@@ -0,0 +1,240 @@
...
+	 * directly to APIs such as connect(), listen(), etc.
+	 */
+	class fd_up {
+	public:
+		fd_up() : _fd(-1) {}
+		fd_up(int fd) : _fd(fd) {}
+		fd_up(fd_up &&other) : _fd(other.release()) {}
+		fd_up(fd_up const &) = delete;
+
+		~fd_up() { reset(); }
```
bsdjhb, line 76: What do we think about single-line things like this that have the entire declaration in one line?   Or should this be more like the get() method a few lines below?

lib/libutil++/libutil++
```diff
@@ -0,0 +1,240 @@
...
+
+		void
+		reset(int newfd = -1)
+		{
+			if (_fd >= 0)
+				close(_fd);
+			_fd = newfd;
+		}
+
+		fd_up &operator=(fd_up &&other) noexcept
```
bsdjhb, line 101:
```suggestion
		fd_up &
		operator=(fd_up &&other) noexcept
```

??

lib/libutil++/libutil++
```diff
@@ -0,0 +1,240 @@
...
+		{
+			if (this == &other)
+				return *this;
+
+			reset(other.release());
+			return *this;
+		}
+
+		fd_up &operator=(fd_up const &) = delete;
```
bsdjhb, line 110: I would not like requiring a newline here for deleting default things.  That is, do we want this to be:

```
      fd_up &
      operator=(fd_up const &) = delete;
```

Seems kind of silly to turn that into multiple lines?

usr.sbin/ctld/ctld.h
```diff
@@ -49,25 +57,34 @@
...
-	struct auth_group		*a_auth_group;
-	char				*a_user;
-	char				*a_secret;
-	char				*a_mutual_user;
-	char				*a_mutual_secret;
-};
+	auth(const char *secret) : a_secret(secret) {}
+	auth(const char *secret, const char *mutual_user,
+	    const char *mutual_secret) : a_secret(secret),
+	    a_mutual_user(mutual_user), a_mutual_secret(mutual_secret) {}
```
bsdjhb, line 67: A better example to think about how to handle initializers in constructors along with constructs with an empty body.  One question is do we want to allow function parameters and initializers to mix in the same line, or force a new line always when initializers start?  Forcing newlines would be:

```
        auth(const char *secret) :
            a_secret(secret) {}
        auth(const char *secret, const char *mutual_user,
            const char *mutual_secret) :
            a_secret(secret), a_mutual_user(mutual_user),
            a_mutual_secret(mutual_secret) {}
```

Another question is where to put the colon.  Until I started typing the above, I had always imagined it being at the start of the new line, but I think I actually line it being at the end of the function declaration if we do a forced line split here.

Another question is whether we require a newline before the empty body, and if so, do we permit the empty body as `{}` or do we expand that out as:

```
     auth(const char *secret) :
         a_secret(secret)
     {
     }
```

usr.sbin/ctld/ctld.h
```diff
@@ -49,25 +57,34 @@
...
+
+	bool mutual() const { return !a_mutual_user.empty(); }
 
-struct auth_name {
-	TAILQ_ENTRY(auth_name)		an_next;
-	struct auth_group		*an_auth_group;
-	char				*an_initiator_name;
+	const char *secret() const { return a_secret.c_str(); }
```
bsdjhb, line 71: More examples of single-line vs do we split them with new lines after the return type, before the opening brace, before the statement, and before the closing brace, e.g.:

```
          const char *
          secret() const
          {
                   return (a_secret.c_str());
          }

          ...
```

usr.sbin/ctld/ctld.h
```diff
@@ -49,25 +57,34 @@
...
 };
 
 struct auth_portal {
-	TAILQ_ENTRY(auth_portal)	ap_next;
-	struct auth_group		*ap_auth_group;
-	char				*ap_initiator_portal;
+	auth_portal(const struct sockaddr_storage *ss, int mask) : ap_sa(*ss),
+	    ap_mask(mask) {}
+
+	bool matches(const struct sockaddr_storage *ss) const;
```
bsdjhb, line 85: I think for method prototypes we should follow the same rule as standalone prototypes and data members in structs in C where if the return type name is short enough, a tab is used to separate the return type from the method name (so in this case it would change to use a tab here after `bool`)

usr.sbin/ctld/ctld.h
```diff
@@ -79,147 +96,426 @@ struct auth_portal {
...
+private:
+	struct pport			*p_pport;
 };
 
 struct lun {
-	TAILQ_ENTRY(lun)		l_next;
+	lun(struct conf *conf, const char *name);
+
+	const char *name() const { return l_name.c_str(); }
+	const std::string &path() const { return l_path; }
```
bsdjhb, line 301: I think `foo &bar` for references (rather than `foo& bar`) is more consistent with our existing style of `foo *bar` for pointers.

usr.sbin/ctld/ctld.cc
```diff
@@ -93,231 +96,236 @@ usage(void)
...
+{
+	assert(ag_type == AG_TYPE_UNKNOWN);
+
+	ag_type = type;
+}
+
+const struct auth *
+auth_group::find_auth(const char *user) const
+{
+	auto it = ag_auths.find(user);
```
bsdjhb, line 227: I think `auto` should be permitted (even encouraged!) for iterator types.  I am not a fan of <mumble>::iterator or <mumble>::const_iterator.  Too hard to read.

usr.sbin/ctld/ctld.cc
```diff
@@ -93,231 +96,236 @@ usage(void)
...
 		return (false);
 	}
 
-	auth_check_secret_length(ag, user, secret, "secret");
+	check_secret_length(user, secret, "secret");
 
-	auth = auth_new(ag);
-	auth->a_user = checked_strdup(user);
-	auth->a_secret = checked_strdup(secret);
+	const auto &pair = ag_auths.try_emplace(user, secret);
```
bsdjhb, line 265: I prefer `const auto` over `auto const` (not sure why, just "looks right").  I think `auto` is also preferred in this case when you are getting a std::pair<> back (especially when the type of the pair<> is a pain.. here one template type is an iterator).

usr.sbin/ctld/ctld.cc
```diff
@@ -93,231 +96,236 @@ usage(void)
...
 
-	ap = reinterpret_cast<struct auth_portal *>(calloc(1, sizeof(*ap)));
-	if (ap == NULL)
-		log_err(1, "calloc");
-	ap->ap_auth_group = ag;
-	ap->ap_initiator_portal = checked_strdup(portal);
-	mask = str = checked_strdup(portal);
+	freebsd::malloc_up<char> str(strdup(portal));
+	if (str == nullptr)
+		throw std::bad_alloc();
```
bsdjhb, line 326: While it might not be consistent with `return` I think it's better to not require ()'s for the operand to `throw`.

BTW, there are apparently some weird edge cases with `decltype` for return types of functions where ()'s around return values causes breakage.  We may wish to relax the requirement for ()'s around return values for C++.

usr.sbin/ctld/ctld.cc
```diff
@@ -654,626 +562,1113 @@ parse_addr_port(char *arg, const char *def_port, struct addrinfo **ai)
...
+	if (pg_foreign)
+		return (0);
+
+	if (!pg_assigned) {
+		log_debugx("not listening on portal-group \"%s\", "
+		    "not assigned to any target", name());
+		return (0);
+	}
+
+	for (portal_up &portal : pg_portals) {
```
bsdjhb, line 783: When the type is "simple" I prefer using the real type for ranged-for.  If the type is complicated (e.g. iterating over a map which uses an iterator type), I prefer to use `auto`.

lib/libutil++/libutil++
llfw: you might want to be explicit about how to open and close nested namespaces, e.g.

```c++
namespace foo::bar {
```

or

```c++
namespace foo { namespace bar {
```

and comments on closing braces, e.g.

```c++
} // namespace foo
} // namespace bar
```

lib/libutil++/libutil++
llfw: not directly related to your comment, but i would write `std::FILE` here, although i know some people will argue that's not technically necessary on the FreeBSD implementation.  either way, how to access ISO C identifiers seems worth standardising.

usr.sbin/ctld/ctld.cc
llfw: more generally, as a lot of C++ is moving towards AAA it might be worth enumerating all the cases where `auto` can/should be used.

```c++
auto v = std::vector<int>(); // ?
auto it = begin(v); // fine
auto i = std::size_t{0}; // ? (explicit type)
auto j = v[0]; // ? (implicit type)
```

and also in function return types: only where necessary, or where it makes the return type easier to write, or always, and when it's okay to elide the return type instead of using a trailing return type.

usr.sbin/ctld/ctld.cc
llfw: cf https://github.com/freebsd/freebsd-src/pull/1623

bsdjhb: Ah, I think one of the things I'm planning to come up with as a result of this discussion is a new C++ section for style(9), and I think we will want to subsume your text into that.

lib/libutil++/libutil++
bsdjhb: Note that the reason for this style for normal C functions and out-of-line methods is to that `grep ^<function name>` works to find a function's definition.  That won't be applicable for in-line methods in classes.

lib/libutil++/libutil++
bsdjhb: Hmm, that I am less certain of.  In particular when converting an existing C application to C++ I think using <stdio.h> rather than <cstdio> will result in fewer diffs.

lib/libutil++/libutil++
bsdjhb: In particular, it might be better to permit the current style as it makes the formatting of in-line vs out-of-line methods similar, in-line would just have a body below the prototype?  I think I might prefer having the body always require a preceding newline?  Not sure if we want to permit single-line bodies however?  `{}` is a special case body, and simple functions that are a single statement might be another special case?

lib/libutil++/libutil++
emaste: To me
```
void
operator()
```
feels perhaps like we're translating the original C style without necessarily having a reason to do so. We don't need to split the type for `^` searching. I guess there is some argument that it's visually consistent / recognizable by the reader.

As for stdio.h I think it's OK to have a different preference when migrating existing code vs writing new.

lib/libutil++/libutil++
emaste: I'm OK with these - I think they're a reasonable exception, similar to single-line `case foo: return bar;` in existing C code.

lib/libutil++/libutil++
bsdjhb: Yes, upon further thought I prefer not splitting after the return type for inline declarations.

contrib/libucl/include/ucl++.h
```diff
@@ -396,6 +396,11 @@ class Ucl final {
 		return default_val;
 	}
 
+	std::string forced_string_value () const
+	{
```
bsdimp, line 400: Looks like we'd need to update the style checker for C++

lib/libiscsiutil/log.c
```diff
@@ -206,6 +206,12 @@ log_warnx(const char *fmt, ...)
 	va_end(ap);
 }
 
+void
+log_vwarnx(const char *fmt, va_list ap)
+{
+	log_common(LOG_WARNING, -1, fmt, ap);
+}
+
```
bsdimp, line 214: This change looks good on its own.

lib/libutil++/libutil++
bsdimp: Yes, inlines in C++ classes shouldn't be split. Not sure about inlines in namespaces, but likely not either.

lib/libutil++/libutil++
```diff
@@ -0,0 +1,240 @@
...
+
+	typedef std::unique_ptr<FILE, fclose_deleter> FILE_up;
+
+	/*
+	 * addrinfo_up is a std::unique_ptr<> which uses
+	 * freeaddrinfo() to destroy the wrapped pointer.  It is
+	 * intended to wrap arrays allocated by getaddrinfo().
+	 */
+	struct freeaddrinfo_deleter {
+		void operator() (struct addrinfo *ai) const
```
bsdimp, line 40: Style(9) would eat the space after the first ). But This style is more consistent with C++ programs more generally, so we should leave it.

lib/libutil++/libutil++
bsdimp: These are fine. I think devd does this (or maybe did before review). It's fine and very common C++ idiom. In general I'm in agreement with ed who said, I think, in the srcmgr meeting that we should follow C++ common practice as much as possible, and when it conflicts with style(9), lean towards common practice unless it's a very identical situation. Though I'd also favor ditching () in return in C++.

lib/libutil++/libutil++
```diff
@@ -0,0 +1,240 @@
...
+		fd_up(int fd) : _fd(fd) {}
+		fd_up(fd_up &&other) : _fd(other.release()) {}
+		fd_up(fd_up const &) = delete;
+
+		~fd_up() { reset(); }
+
+		int
+		get() const
+		{
+			return (_fd);
```
bsdimp, line 81: I think this is too verbose. You can do it one one line. And w/o the () in the return

lib/libutil++/libutil++
bsdimp: I lean towards one line. That's more common style in the C++ world (at least of the c++ world of the 90s and 00s). I'd lean towards allowing the oneliner on one line or two if the function decl part is too long.

contrib/libucl/include/ucl++.h
markpeek: It would be nice if these changes were expressible in a .clang-format configuration file to allow for easy conformance formatting.

lib/libutil++/libutil++
bsdimp: this looks good to me. I didn't know this C++ idiom.

lib/libutil++/libutil++
```diff
@@ -0,0 +1,240 @@
...
+		fd_up &operator=(int fd)
+		{
+			reset(fd);
+			return *this;
+		}
+
+		explicit operator bool() const { return _fd >= 0; }
+		operator int() const { return _fd; }
+	private:
+		int	_fd;
```
bsdimp, line 121: Ii'm not going to get into the can-of-worms debate on m_fd vs _fd and worrying about _fd being in the implementation name space. Pedantically, yes. Practically, we avoid a dozen variables like _i386 and we're good.

lib/libutil++/libutil++
```diff
@@ -0,0 +1,240 @@
...
+	template <class T>
+	struct free_deleter {
+		void operator() (T *p) const
+		{
+			free(p);
+		}
+	};
+
+	template <class T>
+	using malloc_up = std::unique_ptr<T, free_deleter<T>>;
```
bsdimp, line 139: will also have to look up this idiom. Might want to articulate a rule here of newline vs no-newline.
I find this style easy to read, though given my (I guess now distant) past C++ experience.

lib/libutil++/stringf.cc
```diff
@@ -0,0 +1,57 @@
...
+#include <stdio.h>
+#include <string>
+
+#include "libutil++"
+
+static int
+stringf_write(void *cookie, const char *buf, int len)
+{
+	std::string *str = reinterpret_cast<std::string *>(cookie);
+	try {
```
bsdimp, line 18: style checker is wrong

bsdimp:
I only commented on style.
I do generally think that C++ should have it's own style that's more appropriate and in better fashion than style(9) slavisihly followed for C++, even when it doesn't make sense. I see you mixed return foo and return (foo), as well as sometimes having a one line prototype definition, while other times two. Is there a rule there? Or are both of these permitted variations?

lib/libutil++/libutil++
llfw: nitpick: `_fd` is not in the implementation namespaces.  the names reserved to the implementation are those starting with `__` and those starting with `_` followed by a capital letter.  so `__fd` and `_Fd`, but not `_fd`.

usr.sbin/ctld/ctld.cc
dstolfa: > BTW, there are apparently some weird edge cases with `decltype` for return types of functions where ()'s around return values causes breakage. We may wish to relax the requirement for ()'s around return values for C++.

https://godbolt.org/z/zhcv8edea

This is the only edge case I'm aware of. The C++ style guide should either talk about `return (...);` or `decltype(auto)` in some way.

dstolfa: The reasoning for this comes from https://en.cppreference.com/w/cpp/language/decltype, specifically the:
> Note that if the name of an object is parenthesized, it is treated as an ordinary lvalue expression, thus decltype(x) and decltype((x)) are often different types.

part. My 2 cents would be to be permissive with `()` in C++ instead of trying to talk about `decltype` in the style guide, because who knows what future versions of C++ might bring.

lib/libutil++/libutil++
dstolfa: Alternatively, `fd_` could be used. This is what Google's style guide does. I don't really mind either way, however.

dstolfa:
One thing that is quite common to do in C++ is to have class wrappers around flag setting/unsetting and similar things to work with RAII, and then writing code similar to:

```
...
{
  UseFoo temp(...);
  DoStuff();
  // temp's destructor runs here
}
```

It might be worth considering that usecase when thinking about using tab indentation, because one could very quickly run into the 80col limit by adding two loops, an if or two and then using scopes for RAII purposes.

usr.sbin/ctld/ctld.cc
emaste: I'd not require `()` for C++ returns, and omit them from C++ examples.

I don't think we have a great argument for requiring them in C, but they also don't have any material downside and it doesn't really matter that they persist as a historical quirk of our style. No need for that to carry over to C++.

contrib/libucl/include/ucl++.h
emaste: Can clang-format apply different settings for C and C++, from the same .clang-format file?

lib/libutil++/libutil++
emaste: i.e. `int get() const { return _fd; }`?

contrib/libucl/include/ucl++.h
dstolfa: @emaste I think the the `Language:` key should do that. See https://clang.llvm.org/docs/ClangFormatStyleOptions.html

bsdimp:
Great discussion. jhb will try to create a commit to style(9) with this and other experience.

bsdjhb:
> One thing that is quite common to do in C++ is to have class wrappers around flag setting/unsetting and similar things to work with RAII, and then writing code similar to:
> 
> ```
> ...
> {
>   UseFoo temp(...);
>   DoStuff();
>   // temp's destructor runs here
> }
> ```
> 
> It might be worth considering that usecase when thinking about using tab indentation, because one could very quickly run into the 80col limit by adding two loops, an if or two and then using scopes for RAII purposes.

Hmm, what I tend to see now is that unique_ptr<>-like classes have a release() method and that the code will just call release() to drop the RAII object's reference on a resource when needed without resorting to nested blocks.  In particular, this  paradigm permits more complicated cases where you have reference lifetimes that are not strict subsets of each other.  For example if you acquire a reference to some object from a container, you can drop the reference to the container once you have obtained a local reference to the contained object.

usr.sbin/ctld/ctld.cc
bsdjhb: Hmm, it seems that AAA is still somewhat debatable.  I might be inclined to be a bit conservative initially, though I do think we should encourage make_unique<> instead of `new`.

lib/libutil++/libutil++
bsdjhb: Hmmm, I might make this an open question in the style review.

lib/libutil++/libutil++
bsdjhb: `using` is kind of a template version of `typedef`.  I think the question though is about how to deal with the `template` keyword.  I often see a newline after `template <args>` before the rest of the declaration.

bsdjhb:
style(9) review opened at https://reviews.freebsd.org/D50983 for those who are interested

PR #1660, opened 2025-04-12
netlink/test_snl: turn off the NETLINK_EXT_ACK flag during the test.

Rin0913:
This test was disabled due to the extra 32-byte data in the `rx_hdr->nlmsg_len`.

It's not we expected, that should be `sizeof(struct nlmsghdr) + sizeof(struct nlmsgerr)`.

However, that's because the NETLINK_EXT_ACK is enabled, if we disable it, the length will be correct.

https://github.com/freebsd/freebsd-src/commit/04dacd5691fb3de5fddd0868836c5a7ff9888370


spmzt:
@lwhsu ping

spmzt:

PR #1661, opened 2025-04-13
sifive: dtb: add unmatched dtb

RevySR:


jlduran (approved):

bsdimp (approved):

bsdimp (approved):
Looks ready to roll.

jrtc27:
Why? Any non-ancient firmware has a complete and stable device tree that’s provided to the kernel. If you need this I’d like to know what you’re doing that’s different to how I’ve booted mine for years.

RevySR:
> Why? Any non-ancient firmware has a complete and stable device tree that’s provided to the kernel.

I'm not sure what freebsd's recommended best practices are for dtb.
From what I've seen, dtb seems to have stabilized for unmatched.

>  If you need this I’d like to know what you’re doing that’s different to how I’ve booted mine for years.

Is there a guide to the operation here? I don't seem to have found one.

I'd like to ask an additional question.
How do you go about getting the adapted device tree for a new model such as sifive's p550.
Do I need to wait for the linux merge and then synchronize it to the freebsd source code?

jrtc27:
> > Why? Any non-ancient firmware has a complete and stable device tree that’s provided to the kernel.
> 
> I'm not sure what freebsd's recommended best practices are for dtb. From what I've seen, dtb seems to have stabilized for unmatched.
> 
> > If you need this I’d like to know what you’re doing that’s different to how I’ve booted mine for years.
> 
> Is there a guide to the operation here? I don't seem to have found one.

You just boot it? Whether by letting it automatically pick the right media or by running, say, `usb_boot`. All the various commands to boot from different media using bootefi under the hood will pass U-Boot's device tree onto FreeBSD's loader, and in turn kernel.

> I'd like to ask an additional question. How do you go about getting the adapted device tree for a new model such as sifive's p550. Do I need to wait for the linux merge and then synchronize it to the freebsd source code?

Yes, but the firmware-provided one may well suffice. I am not so familiar with the status of the P550.

jrtc27:
> > > > Why? Any non-ancient firmware has a complete and stable device tree that’s provided to the kernel.
> > > 
> > > 
> > > I'm not sure what freebsd's recommended best practices are for dtb. From what I've seen, dtb seems to have stabilized for unmatched.
> > > > If you need this I’d like to know what you’re doing that’s different to how I’ve booted mine for years.
> > > 
> > > 
> > > Is there a guide to the operation here? I don't seem to have found one.
> > 
> > 
> > You just boot it? Whether by letting it automatically pick the right media or by running, say, `usb_boot`. All the various commands to boot from different media using bootefi under the hood will pass U-Boot's device tree onto FreeBSD's loader, and in turn kernel.
> > > I'd like to ask an additional question. How do you go about getting the adapted device tree for a new model such as sifive's p550. Do I need to wait for the linux merge and then synchronize it to the freebsd source code?
> > 
> > 
> > Yes, but the firmware-provided one may well suffice. I am not so familiar with the status of the P550.
> 
> There may also be a strange situation here, that is, the linux kernel dtb upstream will be rewritten later and the vendor uboot (there is no upstream uboot at this time) The two uboot dtbs are inconsistent.
> 
> However, the kernel driver follows the device tree of the Linux kernel. Uboot also needs to correspond to the device tree driver before it can enter the Linux kernel.
> 
> At this time, the freebsd kernel provides a correct device tree to load the corresponding dtb before the bootefi stage. I think it is necessary.

For the Unmatched I solved this problem by just making the drivers in question (in that case, for the PCIe controller) able to work with both vendor and upstreamed device trees.

PR #1662, opened 2025-04-14
ahci.4: make HARDWARE section more conformant

kenrap:
@concussious

Also, I'm not sure how to conform the next paragraph about nvme(4).

### DIFF
--- a/share/man/man4/ahci.4
+++ b/share/man/man4/ahci.4
@@ -22,7 +22,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd December 17, 2021
+.Dd April 13, 2025
 .Dt AHCI 4
 .Os
 .Sh NAME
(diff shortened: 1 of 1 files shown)


concussious:
This is kind of a hard one, and it's pretty fine the way it is. I'm not really sure how to improve it, rather.

A good choice would be like, whatever starts with "The currently supported ATA/SATA controller chips are:" below it, because we have no idea. Many drivers need just the first sentence fixed so it's clear in the HW Relnotes what's going on there.

Some others:
+ uipaq has all its information in the HW Relnotes, those need to be removed and put in a hardware section in the manual
+ Same with puc

kenrap:
> * uipaq has all its information in the HW Relnotes, those need to be removed and put in a hardware section in the manual

It seems that uipaq(4) already had all the items from the release notes in its HARDWARE section. But in the release notes, it's not properly formatted with asterisk bullets.

concussious:
Oh I missed it because it's sorted wrong

kenrap:
I'm going to make a puc(4) PR soon that adds a HARDWARE section for it.

share/man/man4/ahci.4
```diff
@@ -149,12 +149,21 @@ statuses in some way (IBPI standard) to report them using present indicators.
...
 .Nm
-driver supports AHCI compatible controllers having PCI class 1 (mass storage),
-subclass 6 (SATA) and programming interface 1 (AHCI).
+driver supports the following AHCI compatible controllers:
+.Bl -column "Legacy-PATA + AHCI-SATA via ata(4)" "Programming Interface 0x01"
+.It Type: Ta Code:
+.It Mass Storage Ta PCI Class 0x01
+.It SATA Ta Subclass 0x06
+.It AHCI Ta Programming Interface 0x01
+.It Legacy-PATA + AHCI-SATA via ata(4) Ta Programming Interface 0x01
```
amotin, line 158: This is not a list of compatible controllers.  Those are requisites that must be met same time for the driver to attach.

concussious: Yeah, as I said above I don't think we can really improve this one.

kenrap:
Should I close this or would there be value into keeping this open?

concussious:
I would close this because we want to keep the pending patches as small and manageable as possible. Since we don't have a finished patch, or even an idea about how to obtain one, this is a good one to close.

PR #1663, opened 2025-04-14
puc.4: Import HARDWARE from HW Relnotes

kenrap:
@concussious

### DIFF
--- a/share/man/man4/puc.4
+++ b/share/man/man4/puc.4
@@ -22,7 +22,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd January 24, 2008
+.Dd April 13, 2025
 .Dt PUC 4
 .Os
 .Sh NAME
@@ -36,7 +36,10 @@ Communications driver
 .Cd "device uart"
 .Cd "device ppc"
 .Sh DESCRIPTION
-This driver acts as a shim to connect PCI serial and parallel ports to the
+The
+.Nm
+driver acts as a shim to connect
+PCI multi-port serial and parallel adapters to the
 .Xr uart 4
 and
 .Xr ppc 4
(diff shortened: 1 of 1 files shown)


share/man/man4/puc.4
```diff
@@ -45,6 +45,35 @@ driver.
 The list of supported devices is in
 .Pa sys/dev/puc/pucdata.c .
 Support for new cards should be added there.
+.Sh HARDWARE
+The
+.Nm
+driver supports the following PCI serial and parallel ports:
```
concussious, line 51:
```suggestion
driver supports the following
PCI multi-port serial and parallel port controllers:
```
Since this one is a bit long, we have to make a decision about wrapping. The Roff language limits lines to 80 characters, and courtesy limits them to 72 when possible, so we want to be considerate of grouping semantic constructs together for grep.

Make sure we aren't loosing information while distilling. "multi-port" is an interesting detail.

Also consider that people tend to search for parallel ports as "parallel ports".

Even though it says so above, these aren't really ports. Controllers maybe?

share/man/man4/puc.4
```diff
@@ -45,6 +45,35 @@ driver.
...
+.Pp
+.Bl -tag -width Ds -offset indent -compact
```
concussious, line 53:
```suggestion
.Bl -bullet -compact
```
You'll want to use a bullet list so it makes a subsection, this is one of the low hanging fruits of the hardware release notes.

Bullet lists require all the `.It`s on their own line.

You'll want to not use offset since some entries in this list is pretty wide already. MANWIDTH=80 is the reference standard, and we try to consider 59.

concussious (commented):
Awesome! Few details:

bsdimp:
A quick survey of eBay has the vast majority of sellers calling these PCI or PCIe cards.

But after that there's more variation "Serial and Parallel PCI Card" "Quad-Serial 4-port PCIe Card adapter" "serial adapter card"  "two port serial controller adapter card" etc. However, a lot of that is maximizing search hits.

The most common is "PCI card" the second most common appears to be a variation on "Adapter". Rather than "Ports" or "Controller" I'd suggest "Adapter" is the best choice. It's more precise than PCI or PCIe Card and is also close to something people would search for. Controller is less likely to be a search phrase. Adapter also aparallels trends in storage where you have HBA or Host Bus Adapters or some variation on the word Adapter.

concussious (approved):

kenrap:
@concussious
Would it make sense for uipaq(4) to use bullets for its adapter list? I borrowed that `.Bl -tag -width Ds -offset indent -compact` line from its man page. Could that same line there be adjusted to `.Bl -bullet -compact`?

concussious:
Yes.

concussious:
Maybe a better commit title is "puc.4: Import HARDWARE from HW Relnotes".

concussious (approved):

sergio-carlavilla:
> Cc @sergio-carlavilla @mhorne
> 
> May I commit this with MFC after 3 days?

yes!

bsdimp:
Where did this list come from. It seems to short from what I remember about the driver...

concussious:
It came from the 14.2 HW Relnotes

kenrap:
From here: https://github.com/freebsd/freebsd-doc/blob/f8006b944cae79a909bc9d647ede5f5d8b7d7175/website/content/en/releases/14.3R/hardware.adoc?plain=1#L3489-L3510

bsdimp:
The device list in the driver is rather more extensive than this. Any chance we could amend what we had in the release notes to include all we match?

kenrap:
I found an extensive resource which I can use to amend what is already here:
https://github.com/freebsd/freebsd-src/blob/41b30bbc1a57b60afee9acdd6ad240c92ef13790/sys/dev/puc/pucdata.c#L70-L1709

kenrap:
Done a lot of work here, but basically I used the device data from the file above to revamp the entire HARDWARE section. Since there was so much listed redundancy for devices with different models, I decided to go formatting sub-bullets for their model numbers to make it easier on the eyes.

concussious:
Please open this up in the manpages and look at it.

kenrap:
Via `man share/man/man4/puc.4`? I did...

kenrap:
I understand that this is a big change and I was just trying to help @bsdimp. I'll be happy to revert this to its agreeable state.

concussious:
No, you didn't do anything wrong, please excuse my tone earlier. Bsdimp is very smart and has been doing this longer than I've been alive, and it's very good to learn from him. I just don't know how to proceed from here.

This is eleven pages on standard console, with many lines having only three meaningful characters on them. To me, that is not really acceptable, but this is a complex issue. It's very good to list the supported hardware, that's the entire point of what we're trying to do. And on your part, it's a lot of hard work to arrive here, as a volunteer, which is very noble, and appreciated!

I don't know how the HW Relnotes will handle the sub lists. This form is also extremely wasteful of horizontal space. Also, it has no sorting whatsoever, which is very hard when you have a list this big.

This is why I do them one at a time by hand and try to not get ahead of myself, because there be dragons. I need some time to reflect on how to proceed from my end. Hopefully one of the senior developers like imp will come and just teach us what to do.

kenrap:
I gave thought to what you've said yesterday and allow me to express myself.

I can appreciate that this was unexpected and that I unintentionally exposed uncharted waters in man page development. Sadly, these guideline restrictions actually inhibits the way I work and I really can't function under them.

Also, I'm very sorry I caused you a lot of stress. I tend to "go big" when I'm enthusiastic about solving problems with my work. Given these experiences, I'm feeling incompatible and unable to help with this work. The "feel-goodness" of this work had vanished, as if a Dementor sucked it right of me. :(

concussious:
Hey, thanks for your contributions. You didn't cause me any stress, chewing on these problems is interesting for me, but I understand it's not for everyone.

kenrap:
Appreciate it, thanks.

bsdimp:

PR #1664, opened 2025-04-14
sysrc: Fix check flag logic for append and subtract

michal-os:
This is a fix for PR 279200, https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=279200.

- Commit 1: sysrc regression tests, including the c_flag test which will fail on the current sysrc
- Commit 2: sysrc PR fix, c_flag test will pass

Because sysrc modifies rc.conf files, I tried to make the tests cover some additional edge cases.


markjdb:
This looks reasonable to me, but I had to read the bugzilla PR to understand the change -- could you please amend the commit log to explain the bug and how it's being solved?

michal-os:
Thanks for reviewing. I have updated the commit message. Please let me know if it still needs some improvement.

markjdb (approved):
Looks good, thank you!

markjdb:
Merged, thank you.

PR #1665, opened 2025-04-14
uipaq.4: misc conformance improvements

kenrap:
@concussious

### DIFF
--- a/share/man/man4/uipaq.4
+++ b/share/man/man4/uipaq.4
@@ -50,6 +50,16 @@ module at boot time, place the following line in
 .Bd -literal -offset indent
 uipaq_load="YES"
 .Ed
+.Sh DESCRIPTION
+The
+.Nm
+driver provides support for the USB serial emulation provided
+by the iPAQ devices.
+.Pp
+The device is accessed through the
+.Xr ucom 4
+driver which makes it behave like a
+.Xr tty 4 .
 .Sh HARDWARE
 The
 .Nm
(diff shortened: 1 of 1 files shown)


share/man/man4/uipaq.4
```diff
@@ -55,17 +55,27 @@ The
 .Nm
 driver supports the following adapters:
```
concussious, line 56: This line is the bigger issue for the hardware release notes.

kenrap: I noticed both uipaq(4) man pages from OpenBSD and NetBSD included DESCRIPTION sections. Should I derived from theirs to be added into ours?

kenrap: Also, it seems like this list is more device-oriented. Should they be called "devices" instead of "adapters"?

concussious: I think that sounds nice.

bsdimp (approved):
Though given that windows ce phones haven't been relevant in 15 or 20 years, maybe we just retire this device.. 

But in the mean time, go for it.

kenrap:
If there are reasons to not retire mention of Windows CE phones, I can restore them back. I removed them now because I thought the point about the relevancy was reasonable and for committer convenience.

sergio-carlavilla:
For me is ok too

bsdimp (approved):

concussious:
Merged, thanks!

bsdimp:
hmmm, was landing this and think I unwound it, but this looks like I might not have... @concussious can you let me know if I messed up anything?

concussious:
https://cgit.freebsd.org/src/log/ LGTM!

PR #1666, opened 2025-04-16
Add POSIX psiginfo(3) call

ricardobranco777:
Add POSIX psiginfo(3) call.

Fixes https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=286133

For now, use the same implementation used by NetBSD & DragonflyBSD.

References:
- https://pubs.opengroup.org/onlinepubs/9699919799/functions/psiginfo.html
- https://github.com/NetBSD/src/blob/trunk/lib/libc/gen/psignal.c
- https://github.com/DragonFlyBSD/DragonFlyBSD/blob/master/lib/libc/gen/psignal.c

### DIFF
--- a/lib/libc/gen/Symbol.map
+++ b/lib/libc/gen/Symbol.map
@@ -458,6 +458,7 @@ FBSD_1.8 {
 	aio_read2;
 	aio_write2;
 	execvpe;
+	psiginfo;
 	rtld_get_var;
 	rtld_set_var;
 };
--- a/include/signal.h
+++ b/include/signal.h
@@ -115,6 +115,7 @@ int	siginterrupt(int, int);
 #endif
 
 #if __POSIX_VISIBLE >= 200809
+void	psiginfo(const siginfo_t *, const char *);
 void	psignal(int, const char *);
 #endif
 
--- a/lib/libc/gen/Makefile.inc
+++ b/lib/libc/gen/Makefile.inc
@@ -465,7 +465,8 @@ MLINKS+=posix_spawn.3 posix_spawnp.3 \
 	posix_spawnattr_getsigdefault.3 posix_spawnattr_setsigdefault.3 \
 	posix_spawnattr_getsigmask.3 posix_spawnattr_setsigmask.3 \
 	posix_spawnattr_init.3 posix_spawnattr_destroy.3
-MLINKS+=psignal.3 strsignal.3 \
+MLINKS+=psignal.3 psiginfo.3 \
+	psignal.3 strsignal.3 \
 	psignal.3 sys_siglist.3 \
 	psignal.3 sys_signame.3
 MLINKS+=pwcache.3 gid_from_group.3 \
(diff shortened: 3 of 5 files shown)


lib/libc/gen/Symbol.map
```diff
@@ -210,6 +210,7 @@ FBSD_1.0 {
 	posix_madvise;
 	popen;
 	pclose;
+	psiginfo;
```
kostikbel, line 213: The symbol should have FBSD_1.8 version.

ricardobranco777:
> It seems that glibc version does more: `As well as the signal description, psig‐ info() displays information about the origin of the signal, and other information relevant to the signal (e.g., the relevant memory address for hardware-generated signals, the child process ID for SIGCHLD, and the user ID and process ID of the sender, for signals set using kill(2) or sigqueue(3)).`

Yes.  I can add this behaviour, but wanted to test the waters with the simple behaviour in other BSD's first.  Will add this in another commit.

lib/libc/gen/Symbol.map
ricardobranco777: Fixed.

kostikbel:
Please send me the git-format-patch by email, with all the metadata added.

bsdimp:
> Please send me the git-format-patch by email, with all the metadata added.

You can generate that from the pull request... yhe power of hit 😀 or just cherry pick them... it's kinda the point of github: we have all the data and can commit.

emaste:
Yeah it can be obtained by adding _.patch_ i.e. _https://github.com/freebsd/freebsd-src/commit/2f492d91b2a76b63984449d8e29244cdfbe87c13.patch_

It can be directly applied locally via `fetch -o - https://github.com/freebsd/freebsd-src/commit/2f492d91b2a76b63984449d8e29244cdfbe87c13.patch | git am`

But kib's asking to make sure the metadata (any Sponsored-by, Signed-off-by, etc.) is as desired.

bsdimp:
I get it with 'git fetch' :). But I'm pretty sure thid isn't the first submission for this author. If it isn't @github.com i assume it's good. For pull requests i never double check if the commit message is otherwise good.

kenrap:
> For pull requests i never double check if the commit message is otherwise good.

Would that mean Github PR authors wouldn't need to worry about the quality of their commit messages?

ricardobranco777:
I was looking for an automated way of adding the codes and string descriptions from [signal.h](https://github.com/freebsd/freebsd-src/blob/main/sys/sys/signal.h#L290) from the Makefile but the comments on definitions can't be fetched with `echo '#include <signal.h>' | cc -C -dD -E -`.  Glibc works around this by first declaring those constants as `enum`, a comment and then adding a `#define`.

Will look over this over the long weekend.

concussious:
> > For pull requests i never double check if the commit message is otherwise good.
> 
> Would that mean Github PR authors wouldn't need to worry about the quality of their commit messages?

Comitters can fix it, but 1. That makes extra work for them and 2. commit messages are an extremely important skill, you don't want to rob yourself of the practice.

bsdimp:
> > For pull requests i never double check if the commit message is otherwise good.
> 
> Would that mean Github PR authors wouldn't need to worry about the quality of their commit messages?

I routinely ask for fixes to commit messages.  I never check the metadata if the commit message is otherwise good.

ricardobranco777:
Merged. Thanks!

PR #1667, opened 2025-04-16
gpio: misc fixes for acpi_gpiobus & gpioaei

VexedUXR:
Some misc fixes for acpi_gpiobus and gpioaei

About getting the pin flags via an accessor instead of `pin->flags`:
We shouldn't be putting pin flags from acpi_gpiobus into `pin->flags` since the rest of the gpiobus code expects it to contain FDT flags. e.g `gpio_pin_is_active` checks for `GPIO_ACTIVE_LOW`, an FDT-specific flag. If we want to mirror this with ACPI, we would have to: 1. Put the ACPI-specific flags there (currently we convert the ACPI flags to the ones in `sys/gpio.h`). 2. Add a way for gpiobus to differentiate between ACPI and FDT pins.

### DIFF
--- a/sys/dev/gpio/acpi_gpiobus.c
+++ b/sys/dev/gpio/acpi_gpiobus.c
@@ -357,7 +357,7 @@ acpi_gpiobus_attach(device_t dev)
 	    &ctx);
 
 	if (ACPI_FAILURE(status))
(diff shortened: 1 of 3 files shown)


bsdimp (approved):
these look good.

PR #1668, opened 2025-04-16
Reject providers with too small a size for metadata

AZero13:
Otherwise we will end up with potential vulnerabilities.


sys/geom/raid/md_sii.c
```diff
@@ -332,6 +332,13 @@ sii_meta_write(struct g_consumer *cp, struct sii_raid_conf *meta)
 
 	pp = cp->provider;
 
+	/* Verify that the sector size is large enough for the metadata */
+	if (pp->sectorsize < sizeof(*meta)) {
+		G_RAID_DEBUG(1, "Sector size %u too small for metadata on %s",
+		    pp->sectorsize, pp->name);
+		return (EINVAL);
+	}
+
```
bsdimp, line 341: I think this hunk is bogus. We'll have already checked in taste. It can be dropped.

bsdimp:
commit message is better.

sys/geom/raid/md_sii.c
AZero13: Anything else I should know or should do?

bsdimp (commented):
This looks good, but for one (last?) minor quibble (sorry I've not noticed these all at once). The sii_raid_conf structure is 256 bytes, not 512. Can you update the commit message please?

bsdimp (approved):

PR #1669, opened 2025-04-17
vtnet: Option to force vtnet to compile the multiqueue capable driver, even if ALTQ is enabled

jakobsen-lrz:
The Problem is that when compiling the FreeBSD Kernel with ALTQ, it forces the VTNET driver to compile a non-multiqueue capable driver. But i need a kernel with ALTQ enabled, and a multiqueue capable VTNET driver. 

So i added a check, so that when building the kernel `-DVTNET_NO_ALTQ` can be specified, to get the VTNET driver, that is multiqueue capable, instead of the ALTQ capable driver,

### DIFF
--- a/sys/dev/virtio/network/if_vtnetvar.h
+++ b/sys/dev/virtio/network/if_vtnetvar.h
@@ -29,7 +29,7 @@
 #ifndef _IF_VTNETVAR_H
 #define _IF_VTNETVAR_H
 
-#ifdef ALTQ
+#if defined(ALTQ) && !defined(VTNET_NO_ALTQ)
 #define	VTNET_LEGACY_TX
 #endif
 
--- a/sys/dev/virtio/network/if_vtnetvar.h
+++ b/sys/dev/virtio/network/if_vtnetvar.h
@@ -374,7 +374,7 @@ CTASSERT(((VTNET_TX_SEGS_MAX - 1) * MCLBYTES) >= VTNET_MAX_MTU);
  */
 #define VTNET_DEFAULT_BUFRING_SIZE	4096
 
-#define VTNET_CORE_MTX(_sc)		&(_sc)->vtnet_mtx
+#define VTNET_CORE_MTX(_sc)		(&(_sc)->vtnet_mtx)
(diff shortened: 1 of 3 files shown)


sys/dev/virtio/network/if_vtnetvar.h
```diff
@@ -29,7 +29,7 @@
 #ifndef _IF_VTNETVAR_H
 #define _IF_VTNETVAR_H
 
-#ifdef ALTQ
+#if defined ALTQ && !defined VTNET_NO_ALTQ
```
bsdimp, line 32: You need to add this to sys/conf/options. Add it to the file 'opt_global.h' for now. I'll take care of moving it if something srcmgr is talking about doesn't get off the ground.

sys/dev/virtio/network/if_vtnetvar.h
```diff
@@ -374,7 +374,7 @@ CTASSERT(((VTNET_TX_SEGS_MAX - 1) * MCLBYTES) >= VTNET_MAX_MTU);
  */
 #define VTNET_DEFAULT_BUFRING_SIZE	4096
 
-#define VTNET_CORE_MTX(_sc)		&(_sc)->vtnet_mtx
+#define VTNET_CORE_MTX(_sc)		(&(_sc)->vtnet_mtx)
```
bsdimp, line 377: This is good, but can you please make this as a second commit.

sys/dev/virtio/network/if_vtnetvar.h
jakobsen-lrz: I hope this is fine. Thanks.

bsdimp: You'll also need parens around ALTQ and VTNET_NO_ALTQ. I'm not sure this would even compile.

jakobsen-lrz: I changed it, but it does also work without them.

jakobsen-lrz:
Hello, i created another pull request ( #1679 ) , which would make the changes in this pull request obsolete. Can we please put this pull request on halt, until we see if the other one is mergeable?

PR #1671, opened 2025-04-20
bin/cat: add test for reading data from stdin

Rin0913:
I was working on testing FreeBSD component on MacOS, and the tested component is bin/cat.

I found that the cat program binary generated on MacOS are unable to read from stdin, however, which could pass all the tests.

Therefore, I hope I can add a test for testing stdin reading.

### DIFF
--- a/bin/cat/tests/Makefile
+++ b/bin/cat/tests/Makefile
@@ -12,6 +12,7 @@ ${PACKAGE}FILES+=		d_s_output.in
 ${PACKAGE}FILES+=		d_s_output.out
 ${PACKAGE}FILES+=		d_vt_output.in
 ${PACKAGE}FILES+=		d_vt_output.out
+${PACKAGE}FILES+=		stdin_test.out
 
 .include <netbsd-tests.test.mk>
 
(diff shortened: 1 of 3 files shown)


jlduran:
> I found that the cat program binary generated on MacOS are unable to read from stdin, however, which could pass all the tests.

Out of curiosity, which version of macOS fails? I could not reproduce it locally:
https://github.com/apple-oss-distributions/text_cmds/blob/main/cat/cat.c

Also, would you mind if after this test is accepted and committed, I (or you) submit it upstream? The less `# Begin FreeBSD`/`# End FreeBSD`s we have, the better.

Rin0913:
> > I found that the cat program binary generated on MacOS are unable to read from stdin, however, which could pass all the tests.
> 
> 
> 
> Out of curiosity, which version of macOS fails? I could not reproduce it locally:
> 
> https://github.com/apple-oss-distributions/text_cmds/blob/main/cat/cat.c
> 
> 
> 
> Also, would you mind if after this test is accepted and committed, I (or you) submit it upstream? The less `# Begin FreeBSD`/`# End FreeBSD`s we have, the better.

I actually built from the source code of FreeBSD-src, instead of that you posted.

You can check out the repo: https://github.com/Rin0913/freebsd-test-on-macos

After all, you can certainly upstream the test if you consider it meaningful.

bsdimp:

PR #1673, opened 2025-04-21
Make a space between the password prompt

0x16000:
So i made it sync a bit better with the login prompt so basically the expected changes are:
password:(no space) -> password: (with space)

### DIFF
--- a/usr.bin/login/login.c
+++ b/usr.bin/login/login.c
@@ -95,7 +95,7 @@ static void		 usage(void);
 #define	DEFAULT_BACKOFF		3
 #define	DEFAULT_RETRIES		10
 #define	DEFAULT_PROMPT		"login: "
-#define	DEFAULT_PASSWD_PROMPT	"Password:"
+#define	DEFAULT_PASSWD_PROMPT	"Password: "
 #define	TERM_UNKNOWN		"su"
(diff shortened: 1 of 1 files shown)


concussious:
Why?

0x16000:
Like i said i think it just looks better when its synced with the space in between the prompt

0x16000:
i wouldnt have to think so

concussious:
I do not like this. The way it is is the way it's been. When I first started trying to contribute, I thought "first appeared" looks better than "appeared". I was wrong. The way it has been is culturally significant. It's the language used in a 50 year traditional. I think people let me do it to create engagement, because it didn't break anything.

We don't know if this patch will break anything, but it doesn't fix anything. I don't like it.

jrtc27:
sudo uses exactly "Password:" too as its prompt, no space.

bsdimp:
The not having a space after the ':' is part of the interface we can't change. There's thousands and thousands of scripts out there to interact with login, and changing it will break some portion of them. (closed while typing this, thanks @0x16000 )

jrtc27:
Also, this is entirely configurable via passwd_prompt=, so if you don't like it you can set it in your local login.conf

concussious:
> Also, this is entirely configurable via passwd_prompt=, so if you don't like it you can set it in your local login.conf

-- and this is documented in the manual.

PR #1675, opened 2025-04-22
wordexp(3): Handle ECHILD from waitpid

kennylevinsen:
If the calling process has used SIG_IGN as handler or set the SA_NOCLDWAIT flag for SIGCHLD, processes will be automatically reaped on exit and calls to waitpid(3) will therefore fail with ECHILD.

We waitpid primarily to reap our child so that the caller does not have to worry about it. ECHILD indicates that there is no child to reap, so we can just treat that as a success and move on.

### DIFF
--- a/lib/libc/gen/wordexp.c
+++ b/lib/libc/gen/wordexp.c
@@ -265,7 +265,15 @@ cleanup:
 		errno = serrno;
 		return (error);
 	}
-	if (wpid < 0 || !WIFEXITED(status) || WEXITSTATUS(status) != 0)
+	/*
+	 * Check process exit status, but ignore ECHILD as the child may have
+	 * been automatically reaped if the process had set SIG_IGN or
+	 * SA_NOCLDWAIT for SIGCHLD, and our reason for waitpid was just to
+	 * reap our own child on behalf of the calling process.
+	 */
+	if (wpid < 0 && errno != ECHILD)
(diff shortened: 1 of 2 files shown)


kennylevinsen:
Gentle ping

/cc @jbeich

jillest:
Merged as da9e73e5d483c47e67b3094356dd4b640749849e, thanks.

jbeich:
@jillest, do you plan to MFC to stable/14? I'm trying to figure out *how* to phase out [Sway workaround](https://github.com/freebsd/freebsd-ports/blob/main/x11-wm/sway/files/patch-common_stringop.c).

- stable/13 reaches EOL on [2026-04-30](https://www.freebsd.org/security/#sup). If Sway 1.12 is out before the workaround may need to be rebased. Alternatively, the port/package can be left partially broken near EOL.
- stable/14 reaches EOL on 2028-11-30. Until then Sway may release 1.12, 1.13, 1.14, so rebasing the workaround may become tricky.
- stable/15 is [scheduled](https://www.freebsd.org/releases/15.0R/schedule/) for 2025-09-05 (plus 2 months to release). No need for the workaround. Unfortunately, many users may delay upgrade from 14.* releases until 15.1 is out.

jillest:
> @jillest, do you plan to MFC to stable/14?

Yes, to stable/14 and stable/13.

PR #1676, opened 2025-04-24
ath: fix all -wunused-but-set-variable warnings

svmhdvn:
Patch set attached in the following manner:

[PATCH 1/8] ath: fix DEBUG print varargs macros to handle unused parameters
  This clears out a lot of unused-but-set warnings cleanly while avoiding the `#ifdef DEBUG` pattern

[PATCH 2..6/8]
  These are non-trivial changes that need a closer review to ensure no breaking changes.

[PATCH 7/8] ath_hal: fix all remaining trivial -wunused-but-set-variable warnings
  These are all the remaining (trivial) changes.

[PATCH 8/8] build: ath: re-enable -wunused-but-set-variable in ATH_C
  Once all warnings are cleaned up, the ${NO_WUNUSED_BUT_SET_VARIABLE} can be removed safely.

### DIFF
--- a/sys/conf/kern.pre.mk
+++ b/sys/conf/kern.pre.mk
@@ -231,7 +231,7 @@ ZFS_RPC_C=	${CC} -c ${ZFS_CFLAGS} -DHAVE_RPC_TYPES ${WERROR} ${.IMPSRC}
 ZFS_S=		${CC} -c ${ZFS_ASM_CFLAGS} ${WERROR} ${.IMPSRC}
 
 # ATH driver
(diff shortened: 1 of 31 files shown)


erikarn:
hi,

some of the bits in here are good, but a lot of them remove diagnostics that are useful. I'll spend some time cherry-picking from this diff when I'm finished with my current 802.11 diff stack (as I also have some other ath diffs to land!)

bsdimp:
Any progress on reply to @erikarn 's comments of 3 weeks ago?

svmhdvn:
@erikarn which diagnostics are you referring to? I've tried to maintain as much as possible, but I'll make the necessary changes.

svmhdvn:
Abandoning this work for now, I might revisit warnings cleanup some other time.

PR #1677, opened 2025-04-24
netpfil/ipfilter: fix remaining compiler warnings

svmhdvn:


sys/netpfil/ipfilter/netinet/ip_htable.c
```diff
@@ -603,8 +603,7 @@ ipf_htent_remove(ipf_main_softc_t *softc, void *arg, iphtable_t *iph,
 	switch (iph->iph_type & ~IPHASH_ANON)
 	{
 	case IPHASH_GROUPMAP :
-		if (ipe->ipe_group != NULL)
-			ipf_group_del(softc, ipe->ipe_ptr, NULL);
+		ipf_group_del(softc, ipe->ipe_ptr, NULL);
```
bsdimp: This one I think is bogus. But I'm not sure if the `ipe->ipe_ptr` should be `ipe->ipe_group` in the call, or if the test should be against `ipe->ipe_ptr` since `ipf_group_del` will fault if the second arg is NULL.

sys/netpfil/ipfilter/netinet/ip_nat.c
```diff
@@ -7352,30 +7347,18 @@ ipf_nat_nextaddr(fr_info_t *fin, nat_addr_t *na, u_32_t *old, u_32_t *dst)
...
 	case FRI_DYNAMIC :
 	case FRI_NORMAL :
-		/*
-		 * Compute the maximum address by adding the inverse of the
-		 * netmask to the minimum address.
-		 */
-		amax = ~na->na_addr[1].in4.s_addr;
-		amax |= amin;
-		break;
-
```
bsdimp, line 7378: This is indeed unused. I don't know if this indicates a problem or not. Why was this computed before? Can we trace back to where its use was removed? The tangled history of this port / software isn't hepful. this appeared from an upstream import of
`Import IP-Filter 5.1.2 into vendor branches using the existing license that`
from 2013. It wasn't in any prior version, so I think it's likely safe. It won't break anything that's not already broken and this code will still be in git if someone discovers that it should have been used and needs to recover it.

sys/netpfil/ipfilter/netinet/ip_nat6.c
```diff
@@ -3802,32 +3791,19 @@ ipf_nat6_nextaddr(fr_info_t *fin, nat_addr_t *na, i6addr_t *old, i6addr_t *dst)
...
 	case FRI_DYNAMIC :
 	case FRI_NORMAL :
-		/*
-		 * Compute the maximum address by adding the inverse of the
-		 * netmask to the minimum address.
-		 */
-		amax = ~na->na_addr[1].in4.s_addr;
-		amax |= amin;
-		break;
-
```
bsdimp, line 3830: Same comments about amin/amax as above. tl;dr: likely fine.

sys/netpfil/ipfilter/netinet/ip_proxy.c
```diff
@@ -685,7 +685,7 @@ ipf_proxy_ioctl(ipf_main_softc_t *softc, caddr_t data, ioctlcmd_t cmd,
 	caddr_t ptr;
 	int error;
 
-	mode = mode;	/* LINT */
+	(void)mode;	/* UNUSED */
```
bsdimp: There is an __unused ctypes.h annoation that we COULD put on the args, but given the age of this code, this construct is fine, imho.

bsdimp (commented):
There's only one of these that I had a real question about. The other comments are nice to know items, but for code that was imported 12 years ago from an external repo that I had trouble finding in a quick google search suggests that this will be quite hard (though I spent only a few minutes looking and might have missed something)

sys/netpfil/ipfilter/netinet/ip_htable.c
cschuber: The code is valid ipe->ipe_group is a char * pointing to the group name. Patch is incorrect. Please remove this patch segment.

sys/netpfil/ipfilter/netinet/ip_nat6.c
cschuber: Agreed. This can be removed.

sys/netpfil/ipfilter/netinet/fil.c
```diff
@@ -2597,7 +2595,7 @@ ipf_acctpkt(fr_info_t *fin, u_32_t *passp)
 	frentry_t *fr, *frsave;
 	u_32_t pass, rulen;
 
-	passp = passp;
+	(void)passp; /* UNUSED */
```
cschuber: Let's just flag the argument __unused instead.

sys/netpfil/ipfilter/netinet/fil.c
```diff
@@ -7318,7 +7314,7 @@ ipf_resolvenic(ipf_main_softc_t *softc, char *name, int v)
 {
 	void *nic;
 
-	softc = softc;	/* gcc -Wextra */
+	(void)softc;	/* UNUSED */
```
cschuber: Here too. __unused in args list.

sys/netpfil/ipfilter/netinet/ip_ftp_pxy.c
```diff
@@ -1715,7 +1715,7 @@ ipf_p_ftp_eprt4(ipf_ftp_softc_t *softf, fr_info_t *fin, ip_t *ip, nat_t *nat,
 		return (0);
 	if (c != delim)
 		return (0);
-	addr |= addr;
+	/* addr |= (addr << 0); */
```
cschuber: Commenting out statements can be prone to undesirable results. How about a #if 0 block instead?

sys/netpfil/ipfilter/netinet/ip_ipsec_pxy.c
```diff
@@ -348,7 +348,7 @@ ipf_p_ipsec_match(fr_info_t *fin, ap_session_t *aps, nat_t *nat)
 	mb_t *m;
 	int off;
 
-	nat = nat;	/* LINT */
+	(void)nat;	/* UNUSED */
```
cschuber: __unused in args list.

sys/netpfil/ipfilter/netinet/ip_irc_pxy.c
```diff
@@ -232,7 +232,7 @@ ipf_p_irc_new(void *arg, fr_info_t *fin, ap_session_t *aps, nat_t *nat)
 	if (irc == NULL)
 		return (-1);
 
-	nat = nat;	/* LINT */
+	(void)nat;	/* UNUSED */
```
cschuber: ditto

sys/netpfil/ipfilter/netinet/ip_irc_pxy.c
```diff
@@ -424,6 +424,6 @@ ipf_p_irc_send(fr_info_t *fin, nat_t *nat)
 int
 ipf_p_irc_out(void *arg, fr_info_t *fin, ap_session_t *aps, nat_t *nat)
 {
-	aps = aps;	/* LINT */
+	(void)aps;	/* UNUSED */
```
cschuber: ditto.

sys/netpfil/ipfilter/netinet/ip_lookup.c
```diff
@@ -235,7 +235,7 @@ ipf_lookup_ioctl(ipf_main_softc_t *softc, caddr_t data, ioctlcmd_t cmd,
 	int err;
 	SPL_INT(s);
 
-	mode = mode;	/* LINT */
+	(void)mode;	/* UNUSED */
```
cschuber: here too.

sys/netpfil/ipfilter/netinet/ip_raudio_pxy.c
```diff
@@ -53,7 +53,7 @@ ipf_p_raudio_new(void *arg, fr_info_t *fin, ap_session_t *aps, nat_t *nat)
 {
 	raudio_t *rap;
 
-	nat = nat;	/* LINT */
+	(void)nat;	/* UNUSED */
```
cschuber: __unused

sys/netpfil/ipfilter/netinet/ip_raudio_pxy.c
```diff
@@ -82,7 +82,7 @@ ipf_p_raudio_out(void *arg, fr_info_t *fin, ap_session_t *aps, nat_t *nat)
 	int len = 0;
 	mb_t *m;
 
-	nat = nat;	/* LINT */
+	(void)nat;	/* UNUSED */
```
cschuber: same

sys/netpfil/ipfilter/netinet/ip_state.c
```diff
@@ -2207,20 +2206,6 @@ ipf_state_tcpinwindow(fr_info_t *fin, tcpdata_t  *fdata, tcpdata_t *tdata,
...
-#if 0
-		if (is_pkts[i]0 == 0) {
-			/*
-			 * Picking up a connection in the middle, the "next"
-			 * packet seen from a direction that is new should be
-			 * accepted, even if it appears out of sequence.
-			 */
-			inseq = 1;
-		} else
-#endif
```
cschuber, line 2223: Good catch. This was removed by FreeBSD in IPF 4.2.28 but crept back in with IPF 5.1.2 as a mismerge.

sys/netpfil/ipfilter/netinet/ip_nat.c
cschuber: Agreed.

sys/netpfil/ipfilter/netinet/ip_netbios_pxy.c
```diff
@@ -75,8 +75,8 @@ ipf_p_netbios_out(void *arg, fr_info_t *fin, ap_session_t *aps, nat_t *nat)
 	ip_t *ip;
 	mb_t *m;
 
-	aps = aps;	/* LINT */
-	nat = nat;	/* LINT */
+	(void)aps;	/* UNUSED */
+	(void)nat;	/* UNUSED */
```
cschuber: Again, let's use __unused in args list.

sys/netpfil/ipfilter/netinet/ip_proxy.c
cschuber: We should use __unused.

I converted it from K&R a number of years ago. __unused is the next progression.

cschuber (changes requested):
The test for the character string pointer ipe->ipe_group (line ip_htable.c:606) is valid. It must remain.

Let's use __unused in arg lists wherever possible.

sys/netpfil/ipfilter/netinet/ip_htable.c
svmhdvn: This seems like either it should be a test for `ipe->ipe_ptr != NULL`, OR a test for `ipe->ipe_value != 0`, due to the union definition here: https://github.com/freebsd/freebsd-src/blob/3b9b51fe464ebb91e894742a6a0e6417e256f03a/sys/netpfil/ipfilter/netinet/ip_htable.h#L6-L30. The first case seems likely because of @bsdimp's comment that `ipf_group_del` assumes that the second argument pointer is non-null. What do you think?

svmhdvn:
Handled all other feedback in v2.

sys/netpfil/ipfilter/netinet/ip_htable.c
cschuber: Removing the test entirely will result in a panic. Either test for the group's existence before calling ipf_group_del or ipf_group_del() must test for a NULL *group. In this case it makes sense to test it here rather than going through the expense of calling ipf_group_del().

cschuber (changes requested):
We must not pass a NULL ipe_group to ipf_group_del(). Some kind of test must be performed before it is called.

sys/netpfil/ipfilter/netinet/ip_htable.c
cschuber: Test for ipe->ipe_ptr != NULL.

cschuber (changes requested):
Make the ipe->ipe_ptr != NULL test.

After that I will give it a spin here.

cschuber:
Something is not quite right with this group of commits. ippool -l displays:

50040:walking hash nodes: could not find token for lookup iterator

And hash pools that should be populated are not.

cschuber:
Investigating this patch further, it tickled a bug in ip_hasthable.c. I will fix that bug and try this patch again. Sorry for the alarm.

sys/netpfil/ipfilter/netinet/ip_htable.c
svmhdvn: Sorry about the delay, made that change now in the latest revision.

cschuber:
Committed. Thanks for the patches.

PR #1679, opened 2025-04-28
vtnet(4): Add Tunable to deactivate ALTQ support

jakobsen-lrz:
This adds a new tunable for the VirtIO Networking driver (VTNET), called `hw.vtnet.altq_disable`. Which deactivates ALTQ support, so that Multiqueue can be used instead. It does this be changing all compiletime checks, that use the preprocessor, into runtime checks, which depend on the tunable.

If the kernel is compiled without ALTQ included this should have no effect. And if the kernel is compiled with ALTQ, it can npw be deactivated using the tunable as described above.

I tested this on a freebsd vm running under KVM. With ALTQ disabled multiple queues can be used and ALTQ Rules get rejected by pfctl. With ALTQ not disabled, multiqueue cant be used, but pfctl accepts ALTQ Rules.

### DIFF
--- a/share/man/man4/vtnet.4
+++ b/share/man/man4/vtnet.4
@@ -91,6 +91,9 @@ The number of queue pairs used is the lesser of the maximum supported by the
 driver and the hypervisor, the number of CPUs present in the guest, and this
 tunable if not zero.
 The default value is 0.
+.It Va hw.vtnet.altq_disable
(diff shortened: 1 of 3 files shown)


bsdimp (approved):
oh nice! Even documents it, well done on that!

bsdimp:

jakobsen-lrz:
Thanks very much for your time and effort, looking into and merging my Changes.

PR #1680, opened 2025-04-28
graid.8: fix typo

mcimerman:


bsdimp:

PR #1681, opened 2025-04-28
fusefs: Upgrade FUSE protocol to version 7.33.

rmscotty:
This commit upgrades the FUSE API to protocol 7.33, it doesn't implement any of protocol 7.33's new features, setxattr is tested for binary compatibility with protocol version 7.32.

### DIFF
--- a/sys/fs/fuse/fuse_kernel.h
+++ b/sys/fs/fuse/fuse_kernel.h
@@ -176,6 +176,12 @@
  *
  *  7.32
  *  - add flags to fuse_attr, add FUSE_ATTR_SUBMOUNT, add FUSE_SUBMOUNTS
+ *
+ *  7.33
+ *  - add FUSE_HANDLE_KILLPRIV_V2, FUSE_WRITE_KILL_SUIDGID, FATTR_KILL_SUIDGID
+ *  - add FUSE_OPEN_KILL_SUIDGID
+ *  - extend fuse_setxattr_in, add FUSE_SETXATTR_EXT
+ *  - add FUSE_SETXATTR_ACL_KILL_SGID
  */
 
 #ifndef _FUSE_FUSE_KERNEL_H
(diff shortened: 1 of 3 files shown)


sys/fs/fuse/fuse_kernel.h
```diff
@@ -435,11 +453,22 @@ struct fuse_file_lock {
 
 /**
  * fuse_attr flags
- *
```
asomers, line 438: Why delete this line?

sys/fs/fuse/fuse_vnops.c
```diff
@@ -2744,17 +2745,26 @@ fuse_vnop_setextattr(struct vop_setextattr_args *ap)
 	len = strlen(prefix) + sizeof(extattr_namespace_separator) +
 	    strlen(ap->a_name) + 1;
 
-	fdisp_init(&fdi, len + sizeof(*set_xattr_in) + uio->uio_resid);
+	/* we want to be binary compatible */
```
asomers, line 2748: "binary compatible with what"?  I suggest this wording instead.
```suggestion
	/* older FUSE servers  use a smaller fuse_setxattr_in struct*/
```

tests/sys/fs/fusefs/xattr.cc
```diff
@@ -119,6 +121,33 @@ void expect_setxattr(uint64_t ino, const char *attr, const char *value,
...
+{
+	EXPECT_CALL(*m_mock, process(
+		ResultOf([=](auto in) {
+			const char *a = (const char*)in.body.bytes +
+				FUSE_COMPAT_SETXATTR_IN_SIZE;
+			const char *v = a + strlen(a) + 1;
+			return (in.header.opcode == FUSE_SETXATTR &&
+				in.header.nodeid == ino &&
+				in.body.setxattr.size == (strlen(value) + 1) &&
+				in.body.setxattr.flags == 0 &&
```
asomers, line 142: You shouldn't check setxattr.flags right here, because it's undefined.  It actually overlaps with attr.

asomers (changes requested):

sys/fs/fuse/fuse_kernel.h
rmscotty: It should have been there, I didn't notice, thanks for pointing out.

asomers (approved):

asomers:
Committed as d5e3cf41e89400e75da87aad0cc9bde108e2573b .

PR #1684, opened 2025-05-01
Fix comments about cbrt iterations being Halley and not Newton

ubitux:
Since we're inverting a cube, we have:
    f(Tₙ)=Tₙ³-x    (1)

Its first and second derivatives are:
    f'(Tₙ)=3Tₙ²    (2)
    f"(Tₙ)=6Tₙ     (3)

Halley iteration¹ uses:
    Tₙ₊₁=Tₙ-2f(Tₙ)f'(Tₙ)/(2f'(Tₙ)²-f(Tₙ)f"(Tₙ))    (4)

Replacing the terms of (4) using (1), (2) and (3):

Tₙ₊₁ = Tₙ-2f(Tₙ)f'(Tₙ)/(2f'(Tₙ)²-f(Tₙ)f"(Tₙ))
     = Tₙ-2(Tₙ³-x)3Tₙ²/(2(3Tₙ²)²-(Tₙ³-x)6Tₙ)
     = <snip, see WolframAlpha² alternate forms>
     = Tₙ(2x+Tₙ³)/(x+2Tₙ³)

This formula corresponds to the exact expression used in the code.

Newton formula is Tₙ-f(Tₙ)/f'(Tₙ) which would have simplified to (2Tₙ³+x)/(3Tₙ²) instead.

¹ https://en.wikipedia.org/wiki/Halley's_method
² https://www.wolframalpha.com/input?i=T-2%28T%5E3-x%293T%5E2%2F%282%283T%5E2%29%5E2-%28T%5E3-x%296T%29

### DIFF
--- a/lib/msun/src/s_cbrt.c
+++ b/lib/msun/src/s_cbrt.c
@@ -90,7 +90,7 @@ cbrt(double x)
      * the result is larger in magnitude than cbrt(x) but not much more than
      * 2 23-bit ulps larger).  With rounding towards zero, the error bound
      * would be ~5/6 instead of ~4/6.  With a maximum error of 2 23-bit ulps
-     * in the rounded t, the infinite-precision error in the Newton
+     * in the rounded t, the infinite-precision error in the Halley
      * approximation barely affects third digit in the final error
      * 0.667; the error in the rounded t can be up to about 3 23-bit ulps
      * before the final error is larger than 0.667 ulps.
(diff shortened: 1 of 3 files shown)


bsdimp (approved):
This looks sane to me, and seems to track following the links.

bsdimp:
I'm a little torn on the UTF8 in the commit message. However, that's allowed and it serves a good purpose. It's only in the headline it's not allowed. I tweaked the message a little none-the-less to make the footnotes more like how we normally do them and add a note about UTF-8. Primarily because people will think I didn't do it on purpose. git handles it just fine.

bsdimp:

PR #1685, opened 2025-05-02
Fix incorrect version introduced in manual pages

tomhukins:
Several manual pages for releng/14.3 incorrectly claim that features were first introduced in FreeBSD 15.0.

I discovered these by running:

```
  git checkout origin/releng/14.3
  git grep -F '.Fx 15.0'
```

I suggest whoever commits these changes also merges them to the `stable/14` and `releng/14.3` branches.

### DIFF
--- a/lib/libc/gen/psignal.3
+++ b/lib/libc/gen/psignal.3
@@ -120,7 +120,7 @@ function appeared in
 The
 .Fn psiginfo
 function appeared in
-.Fx 15.0 ,
+.Fx 14.3 ,
 .Nx 6.0 ,
 and
 .Dx 4.1 .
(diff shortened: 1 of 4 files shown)


concussious (approved):
> I suggest whoever commits these changes also merges them to the stable/14 and releng/14.3 branches.

In the future you can declare this with a commit message trailer:
```
MFC after: 3 days
MFC to: 14.3
```

tomhukins:
> In the future you can declare this with a commit message trailer

Thank you for suggesting that: I had always assumed that committers, not contributors, decided when/if to MFC commits and so it made no sense for contributors to include such details.

bsdimp:

ricardobranco777:
Thank you for fixing this. It was lingering in my head when it was merged to 14-stable.

PR #1686, opened 2025-05-02
tcp(4): the default congestion algorithm changed

tomhukins:
cc_cubic(4) replaced cc_newreno(4) as the default in FreeBSD 14 with commit bb1d472d79f7.

### DIFF
--- a/share/man/man4/tcp.4
+++ b/share/man/man4/tcp.4
@@ -447,7 +447,7 @@ and the reverse source route is used in responding.
 The default congestion control algorithm for
 .Tn TCP
 is
-.Xr cc_newreno 4 .
+.Xr cc_cubic 4 .
 Other congestion control algorithms can be made available using the
 .Xr mod_cc 4
 framework.


concussious (approved):
Nice catch!
Nits:
+ Need a `Fixes:` tag
+ s/tcp\(4/)/tcp.4 to clarify it's only the manual

tomhukins:
Thank you for your quick, helpful review.

> * Need a `Fixes:` tag

I disagree.  Commit 29f269dc1f1c0 introduced this documentation detail in 2011, when the statement was true.

[The Committer's Guide](https://docs.freebsd.org/en/articles/committers-guide/) says to use `Fixes:` to show a _Git short hash and the title line of a commit that is fixed by this change_ but I don't believe my work fixes any mistake; it just updates the documentation.

I will include this line in the commit message if you find my reasoning unconvincing.  Please let me know whether you believe my commit fixes bb1d472d79f71, afa30c5dec5afc, or something else.

> * s/tcp(4/)/tcp.4 to clarify it's only the manual

Interesting.  I have made this change as requested.

Many committers use parentheses for manual changes, such as for afa30c5dec5afc849, 1a7c8b9739331e0, and 995b690d13980.  Can you point me at where this is documented, and would you consider encouraging other committers to follow the convention?

concussious:
> Thank you for your quick, helpful review.
> 
> > * Need a `Fixes:` tag
> 
> I disagree. Commit [29f269d](https://github.com/freebsd/freebsd-src/commit/29f269dc1f1c0225e4a11cf8938ae36617000607) introduced this documentation detail in 2011, when the statement was true.

> [The Committer's Guide](https://docs.freebsd.org/en/articles/committers-guide/) says to use `Fixes:` to show a _Git short hash and the title line of a commit that is fixed by this change_ but I don't believe my work fixes any mistake; it just updates the documentation.

It `Fixes:` bb1d472. Fixes is not an accusation, it's so we can link them together. If someone is taking bb1d472 for something, they're going to also want this commit. Although, bb1d472 introduced a bug in the documentation by changing behavior without changing the documentation, which this commit fixes.

> I will include this line in the commit message if you find my reasoning unconvincing. Please let me know whether you believe my commit fixes [bb1d472](https://github.com/freebsd/freebsd-src/commit/bb1d472d79f718296a2e9487dd0c219a9b67b2ff), [afa30c5](https://github.com/freebsd/freebsd-src/commit/afa30c5dec5afc8496172a062c93e96d78f5b9b9), or something else.
> 
> > * s/tcp(4/)/tcp.4 to clarify it's only the manual
> 
> Interesting. I have made this change as requested.
> 
> Many committers use parentheses for manual changes, such as for [afa30c5](https://github.com/freebsd/freebsd-src/commit/afa30c5dec5afc8496172a062c93e96d78f5b9b9), [1a7c8b9](https://github.com/freebsd/freebsd-src/commit/1a7c8b9739331e031fce531ca82d04cae6733001), and [995b690](https://github.com/freebsd/freebsd-src/commit/995b690d1398044dc9d85a6d86ec550cda30b2ac). Can you point me at where this is documented, and would you consider encouraging other committers to follow the convention?

`tcp(4)` could be referring to the interface as a whole, but tcp.4 could only be referring to the manual, so it is more specific, also in less characters. I have been encouraging everyone to do so because it makes it easier to sift through the log. I don't know that it is documented.

tomhukins:
> It `Fixes:` [bb1d472](https://github.com/freebsd/freebsd-src/commit/bb1d472d79f718296a2e9487dd0c219a9b67b2ff). Fixes is not an accusation, it's so we can link them together. If someone is taking [bb1d472](https://github.com/freebsd/freebsd-src/commit/bb1d472d79f718296a2e9487dd0c219a9b67b2ff) for something, they're going to also want this commit.

I understand this and hadn't appreciated the "if you want X you will also want Y" point before.

> Although, [bb1d472](https://github.com/freebsd/freebsd-src/commit/bb1d472d79f718296a2e9487dd0c219a9b67b2ff) introduced a bug in the documentation by changing behavior without changing the documentation, which this commit fixes.

This makes sense too.

> `tcp(4)` could be referring to the interface as a whole, but tcp.4 could only be referring to the manual, so it is more specific, also in less characters. I have been encouraging everyone to do so because it makes it easier to sift through the log. I don't know that it is documented.

Again, thank you for explaining this.  I like your reasoning.

I believe the most recent version of my commit message fixes both the problems you mentioned and includes a `Fixes: ` tag following [the documented format](https://docs.freebsd.org/en/articles/committers-guide/#_include_appropriate_metadata_in_a_footer).

Finally, I have noticed the excellent work you have done on FreeBSD's documentation over the past few months and appreciate it greatly.

bsdimp:

PR #1687, opened 2025-05-02
zfsd: don't try to fix an OFFLINE condition

asomers:
If the system administrator does "zpool offline", he's doing it for a reason.  zfsd shouldn't consider an offline disk to be an event that requires automatic healing.  Don't online it in response to a GEOM event, and don't try to activate a hotspare to take over from it.

MFC after:	2 weeks
Sponsored by:	ConnectWise


bsdimp:
Hey @asomers what's your plans for this? Why's it still open after so long?

bsdimp:
oh! n/m.

PR #1688, opened 2025-05-02
fr.kbd and fr.acc.kbd: Added alias to tilde on french layouts for consistency

celesteornato:
While not specified by the AZERTY standard employed by the `fr.kbd` and `fr.acc.kbd` layouts, the keycode `041 | shift` (Shift-²) is assigned to the tilde/home symbol on Xorg, the Linux TTY, and all popular wayland compositors; it is much easier to type than the standard way of writing a tilde with AZERTY.

The only context that I know of where this is not the case is the FreeBSD TTY, this addition aims to improve the quality-of-life to the french-typing FreeBSD experience.


celesteornato:
Oops, disregard that last warning, I synced the forks slightly too hastily

bsdimp:

PR #1691, opened 2025-05-05
sys/intpm: fix SMBus Read Block

ckoehne:
The Read Block function missed saving the data into the preallocated buffer. This causes it to return some garbage when called.

MFC after:		1 week
Sponsored by:		Beckhoff Automation GmbH & Co. KG

### DIFF
--- a/sys/dev/intpm/intpm.c
+++ b/sys/dev/intpm/intpm.c
@@ -879,7 +879,7 @@ intsmb_bread(device_t dev, u_char slave, char cmd, u_char *count, char *buf)
 		if (nread != 0 && nread <= SMBBLOCKTRANS_MAX) {
 			*count = nread;
 			for (i = 0; i < nread; i++)
-				bus_read_1(sc->io_res, PIIX4_SMBBLKDAT);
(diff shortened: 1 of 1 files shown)


ckoehne:
@brian90013 @kostikbel @avg-I @bsdjhb @bsdimp @cemeyer I've found you in the git log. Could you please take a look at this fix?

avg-I:
LGTM

sys/dev/intpm/intpm.c
```diff
@@ -879,7 +879,7 @@ intsmb_bread(device_t dev, u_char slave, char cmd, u_char *count, char *buf)
 		if (nread != 0 && nread <= SMBBLOCKTRANS_MAX) {
 			*count = nread;
```
cemeyer, line 880: Do we need to do any validation that `nread <= *count` from entry to make sure we aren't overrunning `buf`?  (I'm unfamiliar with the smbus_bread API.)

avg-I: I think that `count` is an output parameter.
Callers are expected to handle up to `SMBBLOCKTRANS_MAX`  bytes of data.

PR #1695, opened 2025-05-08
Update atomic.h, 64 bit compare exchange on ARM 32

EyeAndTea:
Hello everyone.

Forgive me if I am not following the rules. This is a quick fix, and I like to stay private. 

Assembly code pertaining to ARM 32bit processors acting on 64bit wide data in what amounts to a compare and exchange operation has an apparent bug where the current value in memory that is returned to the caller is the same suspected current value that was passed by the caller even when the exchange fails.

Please be warned that this apparent bug was caught by code analysis, not testing.

### DIFF
--- a/sys/arm/include/atomic.h
+++ b/sys/arm/include/atomic.h
@@ -213,7 +213,7 @@ ATOMIC_ACQ_REL_LONG(clear)
 	    "   teqeq    %R[tmp], %R[cmp]                     \n"  \
 	    "   ittee    ne                                   \n"  \
 	    "   movne    %[ret], #0                           \n"  \
-	    "   strdne   %[cmp], [%[oldv]]                    \n"  \
+	    "   strdne   %[tmp], [%[oldv]]                    \n"  \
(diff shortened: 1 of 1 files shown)


eyestorefivem:
Very clean code with good test coverage. This is a great example of best practices.

fivemservereyestore:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

bsdimp:
The commit message is missing a why: What is the bug and why does this fix that bug?

bsdimp:
No feedback in 6 months, closing. If this is really a problem, please answer my question and reopen.

EyeAndTea:
If I am understanding the question, there was no more to be said. All is already explained in the original comment above. You can let things be, which I will not dispute because this was done by code analysis, or you can let some one, who is familiar with that code and assembly, analyze the change to confirm what my original comment said.

emaste:
The commit message should explain _why_, not what. The commit message `Update atomic.h, 64 bit compare exchange on ARM 32` says nothing about why the change is being made.

PR #1696, opened 2025-05-10
Add missing POSIX sig2str(3) & str2sig(3) calls

ricardobranco777:
Add `sig2str(3)` & `str2sig(3)` calls defined in POSIX 1003.1-2024:
- https://pubs.opengroup.org/onlinepubs/9799919799/functions/sig2str.html

Also add needed `SIG2STR_MAX` definition:
- https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/signal.h.html

Alternative implementations:
- Illumos:: https://github.com/illumos/illumos-gate/blob/master/usr/src/lib/libc/port/gen/str2sig.c
- Gnulib: https://github.com/coreutils/gnulib/blob/master/lib/sig2str.c

Notes:
- Except for manpage date, this patch applies cleanly on FreeBSD 14
- Had to remove a smaller definition for `SIG2STR_MAX` [here](https://github.com/freebsd/freebsd-src/blob/main/lib/libproc/libproc.h#L108) to use `32` as [defined by Illumos](https://github.com/illumos/illumos-gate/blob/master/usr/src/head/signal.h#L95)

Next steps could be:
- Modifying `kill`, `fuser` and other tools that need translation between signal names & numbers.
- Add POSIX definition for `NSIG_MAX` in `<limits.h>` 

Fixes https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=286714


ricardobranco777:
Tested with:

```
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <signal.h>
#include <ctype.h>

static void
test(const char *arg)
{
	char buf[SIG2STR_MAX];
	int signum;
	char *end;

	if (isdigit((unsigned char)arg[0])) {
		signum = strtol(arg, &end, 10);
		if (sig2str(signum, buf) == 0)
			printf("sig2str(%d) = \"%s\"\n", signum, buf);
		else
			printf("sig2str(%d) = error\n", signum);
	} else {
		if (str2sig(arg, &signum) == 0)
			printf("str2sig(\"%s\") = %d\n", arg, signum);
		else
			printf("str2sig(\"%s\") = error\n", arg);
	}
}

int
main(int argc, char *argv[])
{
	for (int i = 1; i < argc; i++)
		test(argv[i]);

	return (0);
}
```


```
# Shall work
./str2sig HUP 1 USR2 31 RTMIN 65 RTMIN+1 66 RTMAX 126 RTMAX-1 125 RTMIN+30 95 RTMAX-30 96
str2sig("HUP") = 1
sig2str(1) = "HUP"
str2sig("USR2") = 31
sig2str(31) = "USR2"
str2sig("RTMIN") = 65
sig2str(65) = "RTMIN"
str2sig("RTMIN+1") = 66
sig2str(66) = "RTMIN+1"
str2sig("RTMAX") = 126
sig2str(126) = "RTMAX"
str2sig("RTMAX-1") = 125
sig2str(125) = "RTMAX-1"
str2sig("RTMIN+30") = 95
sig2str(95) = "RTMIN+30"
str2sig("RTMAX-30") = 96
sig2str(96) = "RTMAX-30"

# Shall fail
./str2sig XXX 0 -1 SIGXXX RTMIN-1 RTMAX+1
str2sig("XXX") = error
sig2str(0) = error
str2sig("-1") = error
str2sig("SIGXXX") = error
str2sig("RTMIN-1") = error
str2sig("RTMAX+1") = error
```

sys/sys/signal.h
```diff
@@ -385,6 +385,10 @@ struct sigaction {
 #define	NSIG		32	/* number of old signals (counting 0) */
 #endif
 
+#if __POSIX_VISIBLE >= 202405
+#define SIG2STR_MAX	32	/* size of buffer required for sig2str() */
```
kostikbel, line 389: Why is this symbol in sys/signal.h?  It is not needed by the kernel, why cannot it be placed into include/signal.h?

lib/libc/gen/psignal.c
```diff
@@ -61,3 +62,91 @@ psiginfo(const siginfo_t *si, const char *s)
...
+		n = signum - SIGRTMIN;
+	} else {
+		strcpy(str, "RTMAX");
+		n = signum - SIGRTMAX;
+	}
+
+	if (n != 0) {
+		/*
+		 * This block does the equivalent of
+		 * sprintf(str + 5, "%+d", n);
```
kostikbel, line 92: Then why cannot the code use s(n)printf() directly, and needs to hand-roll the printf?

lib/libc/gen/psignal.c
```diff
@@ -61,3 +62,91 @@ psiginfo(const siginfo_t *si, const char *s)
...
+	char tmp[16];
+	char *t, *p;
+	int n;
+
+	if (signum > 0 && signum < sys_nsig) {
+		strcpy(str, sys_signame[signum]);
+		return (0);
+	}
+
+	if (signum < SIGRTMIN || signum > SIGRTMAX)
```
kostikbel, line 78: This leaves SIGTHR, SIGRTLIB, and [SIGRTLIB + 1, SIGRTMIN - 1] with error, instead of following the requirement
```
66907
If signum is a valid, supported signal number, is either less than SIGRTMIN or greater than
66908
SIGRTMAX, and is not equal to one of the symbolic constants listed in the table of signal
66909
numbers in <signal.h>, the stored string shall uniquely identify the signal number signum in an
66910
unspecified manner.
```

lib/libc/gen/psignal.c
```diff
@@ -61,3 +62,91 @@ psiginfo(const siginfo_t *si, const char *s)
...
+	}
+
+	for (sig = 1; sig < sys_nsig; sig++) {
+		if (strcmp(sys_signame[sig], str) == 0) {
+			*pnum = sig;
+			return (0);
+		}
+	}
+
+	return (-1);
```
kostikbel, line 151: This does not handle signals above SIGUSR2 down to SIGRTMIN - 1.

lib/libc/gen/psignal.c
ricardobranco777: > Then why cannot the code use s(n)printf() directly, and needs to hand-roll the printf?

For the same reason it's done in strsignal.c:

> /* Do this by hand, so we don't link to stdio(3). */

https://github.com/freebsd/freebsd-src/commit/78121e79aa390b2e94fb8aca2d0477d81aa20cf2#diff-4615e09983d3d0fad76f04357ef5f26d3ee008e7f3f21e71dfd1a85035d04279R56

I can change it, of course.

lib/libc/gen/psignal.c
ricardobranco777: > This leaves SIGTHR, SIGRTLIB, and [SIGRTLIB + 1, SIGRTMIN - 1] with error, instead of following the requirement

Will look into this. Beware, though, that `strsignal(SIGTHR)` returns `Unknown signal: 32`.

sys/sys/signal.h
ricardobranco777: > Why is this symbol in sys/signal.h? It is not needed by the kernel, why cannot it be placed into include/signal.h?

I put it there because there are no SIGxxx definitions in include/signal.h but I can change it.

lib/libc/gen/psignal.c
ricardobranco777: Fixed.

lib/libc/gen/psignal.c
ricardobranco777: > This does not handle signals above SIGUSR2 down to SIGRTMIN - 1.

Fixed it so `./str2sig 32` to `./str2sig 64` now return the numeric string.

No special handling for `str2sig("SIGTHR", ...)` though, as no special handling is done in `kill` for aliases such as `SIGIOT`.

sys/sys/signal.h
kostikbel: Still not moved

lib/libc/gen/psignal.c
```diff
@@ -61,3 +63,70 @@ psiginfo(const siginfo_t *si, const char *s)
...
+	return (0);
+}
+
+int
+str2sig(const char * restrict str, int * restrict pnum)
+{
+	int n, sig;
+	char *end;
+
+	if (strncmp(str, "RTMIN", 5) == 0 || strncmp(str, "RTMAX", 5) == 0) {
```
kostikbel, line 100: I suggest to add
```
static const char rtmin_str[] = "RTMIN";
static const char rtmax_str[] = "RTMAX";
```
and use them and sizeof(rtmin_str) etc systematically instead of the directly placed literals and the constant '5'.

lib/libc/gen/psignal.c
```diff
@@ -61,3 +63,70 @@ psiginfo(const siginfo_t *si, const char *s)
...
+int
+sig2str(int signum, char *str)
+{
+	int n;
+
+	if (signum <= 0 || signum > SIGRTMAX)
+		return (-1);
+
+	if (signum < sys_nsig)
+		(void) strcpy(str, sys_signame[signum]);
```
kostikbel, line 76: No space between ')' and 's'.  There and everywhere you do the cast to ignore result.

lib/libc/gen/psignal.c
```diff
@@ -61,3 +63,70 @@ psiginfo(const siginfo_t *si, const char *s)
...
+		} else if (str[5] != '\0')
+			return (-1);
+		sig += n;
+		if (sig < SIGRTMIN || sig > SIGRTMAX)
+			return (-1);
+		*pnum = sig;
+		return (0);
+	}
+
+	if (str[0] >= '0' && str[0] <= '9') {
```
kostikbel, line 116: if (isdigit(srt[0]))

kostikbel (commented):
I think we should put the new functions into a new source file, in particular due to stdio use, as you noted.

sys/sys/signal.h
ricardobranco777: Fixed.

lib/libc/gen/psignal.c
ricardobranco777: Hopefully fixed.

lib/libc/gen/sig2str.c
```diff
@@ -0,0 +1,109 @@
...
+	if (signum <= 0 || signum > SIGRTMAX)
+		return (-1);
+
+	if (signum < sys_nsig)
+		(void)strcpy(str, sys_signame[signum]);
+	else if (signum < SIGRTMIN)
+		(void)sprintf(str, "%d", signum);
+	else {
+		if (signum <= (SIGRTMIN + SIGRTMAX) / 2) {
+			(void)strcpy(str, rtmin_str);
```
kostikbel, line 56: isn't it simpler to do
```
sprintf(str, "%s%d", rtmin_str, signum - SIGRTMIN);
```

ricardobranco777: > isn't it simpler to do
> 
> ```
> sprintf(str, "%s%d", rtmin_str, signum - SIGRTMIN);
> ```

No.  POSIX requires the lower half to be `RTMIN+n` and the upper half to be `RTMAX-n`.

> If signum is between SIGRTMIN+1 and (SIGRTMIN+SIGRTMAX)/2 inclusive, the stored string shall be of the form "RTMIN+n", where n is the shortest decimal representation of the value of signum-SIGRTMIN.
> If signum is between (SIGRTMIN+SIGRTMAX)/2 + 1 and SIGRTMAX-1 inclusive, the stored string shall be either of the form "RTMIN+n" or of the form "RTMAX-m", where n is the shortest decimal representation of the value of signum-SIGRTMIN and m is the shortest decimal representation of the value of SIGRTMAX-signum.

kostikbel: I might make a mistake in the exact string content, I did not re-checked, but this was not the point.  I mean, replace the construction of the resulting string by the series of strcpy/strcat/sprintf() by single sprintf().  For lower half it seems to be
```
sprintf(str, "%s+%d", rtmin_str, signum - SIGSTRMIN);
```

ricardobranco777: Fixed.

kostikbel (approved):
@bsdimp please review and push

igs-raiderler:
The documentation in this repository is excellent! Very clear and comprehensive.

dag-erling:
Could I ask you to please rewrite your test to use libatf-c, expand it to cover as many cases (including errors) as possible, and include it in the PR? (give it a name that ends in `_test`, put it in `lib/libc/tests/gen/`, don't forget to update the Makefile)

lib/libc/gen/psignal.3
```diff
@@ -108,6 +114,27 @@ contains a count of the strings in
 .Va sys_siglist
 and
 .Va sys_signame .
+.Pp
+The
+.Fn sig2str
+translates the signal number
```
dag-erling, line 120: missing “function”

lib/libc/gen/psignal.3
```diff
@@ -108,6 +114,27 @@ contains a count of the strings in
...
+.Fa signum
+to the signal name, without the "SIG" prefix, and stores it at the location specified by
```
dag-erling, line 122: use `.Dq`

lib/libc/gen/psignal.3
```diff
@@ -108,6 +114,27 @@ contains a count of the strings in
...
+.Fa str ,
+which should be large enough to hold the name and the terminating NUL byte.
+The symbol
+.Dv SIG2STR_MAX
+gives the maximum size in bytes required.
+.Pp
+The
+.Fn str2sig
+translates the signal name
```
dag-erling, line 131: missing “function”

lib/libc/gen/psignal.3
```diff
@@ -108,6 +114,27 @@ contains a count of the strings in
...
+.Fa str
+to a signal number, and stores it in the location referenced by
```
dag-erling, line 133: redundant comma

lib/libc/gen/psignal.3
```diff
@@ -108,6 +114,27 @@ contains a count of the strings in
...
+.Fa pnum .
+The name in
+.Fa str
+can be either the name of the signal, without the "SIG" prefix, or a decimal number.
```
dag-erling, line 137: use `.Dq`

lib/libc/gen/psignal.3
```diff
@@ -108,6 +114,27 @@ contains a count of the strings in
...
 .Sh SEE ALSO
```
dag-erling, line 138: Consider adding a `STANDARDS` section between `SEE ALSO` and `HISTORY`.

lib/libc/gen/sig2str.c
```diff
@@ -0,0 +1,104 @@
...
+static const char rtmax_str[] = "RTMAX";
+
+int
+sig2str(int signum, char *str)
+{
+	if (signum <= 0 || signum > SIGRTMAX)
+		return (-1);
+
+	if (signum < sys_nsig)
+		(void)strcpy(str, sys_signame[signum]);
```
dag-erling, line 49: Please use `strlcpy()`

lib/libc/gen/sig2str.c
```diff
@@ -0,0 +1,104 @@
...
+	else if (signum < SIGRTMIN)
+		(void)sprintf(str, "%d", signum);
```
dag-erling, line 51: Please use `snprintf()`

lib/libc/gen/sig2str.c
```diff
@@ -0,0 +1,104 @@
...
+
+	if (strncmp(str, rtmin_str, sizeof(rtmin_str) - 1) == 0 ||
+	    strncmp(str, rtmax_str, sizeof(rtmax_str) - 1) == 0) {
+		sig = (str[4] == 'X') ? SIGRTMAX : SIGRTMIN;
+		n = 0;
+		if (str[rtend] == '+' || str[rtend] == '-') {
+			n = (int) strtol(str + rtend, &end, 10);
+			if (*end != '\0' || n == 0)
+				return (-1);
+		} else if (str[rtend] != '\0')
```
dag-erling, line 79: Please add braces

lib/libc/gen/sig2str.c
```diff
@@ -0,0 +1,104 @@
...
+			return (-1);
+		sig += n;
+		if (sig < SIGRTMIN || sig > SIGRTMAX)
+			return (-1);
+		*pnum = sig;
+		return (0);
+	}
+
+	if (isdigit(str[0])) {
+		sig = (int)strtol(str, &end, 10);
```
dag-erling, line 89: Please use an appropriately-sized variable instead of a cast, and check for overflow (see `strtol(3)` for details).

ricardobranco777:
> Could I ask you to please rewrite your test to use libatf-c, expand it to cover as many cases (including errors) as possible, and include it in the PR? (give it a name that ends in `_test`, put it in `lib/libc/tests/gen/`, don't forget to update the Makefile)

Sure, I could do that. I have some experience with ATF on NetBSD.

dag-erling (changes requested):
Please also consider adding SSP instrumentation, see for instance d0b7445904f59 which added instrumentation for `realpath()` which, like `sig2str()`, has an implied buffer size.

lib/libproc/libproc.h
```diff
@@ -105,8 +105,6 @@ typedef enum {
 	REG_RVAL2
 } proc_reg_t;
 
-#define SIG2STR_MAX	8
```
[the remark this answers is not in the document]
dag-erling: Actually, never mind, libproc doesn't even use `SIG2STR_MAX`, I have no idea why it's defined here.

dag-erling:
Additionally, in `bin/kill/kill.c`, `signame_to_signum()` can be replaced with `str2sig()` and the `-l` logic can use `sig2str()` instead of accessing `sys_signame[]` directly.

BTW, both times `strtol()` is used in that file to parse a signal number, it assigns a `long` to an `int` without a range check, and the subsequent error check is incorrect. It would probably be better to use `strtonum()` instead. Also, it is bad form to use “illegal” in an error message; it should be replaced with something more neutral like “invalid” or “unrecognized”.

dag-erling:
In `bin/sh/trap.c`, `sigstring_to_signum()` can be replaced with `str2sig()`. It still needs to check for `EXIT`, but since the function is only used once, that check can be moved to the caller. The other uses of `sys_signame` in `bin/sh` can be left as-is.

usr.bin/fstat/fuser.c
```diff
@@ -353,7 +353,7 @@ dofiles(struct procstat *procstat, struct kinfo_proc *kp,
  * Returns signal number for it's string representation.
  */
 static int
-str2sig(const char *str)
+xstr2sig(const char *str)
```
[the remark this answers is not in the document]
ricardobranco777: > This can simply be discarded if the libc version is modified to accept a `SIG` prefix.

The implementation used by `fuser` allows for lowercase strings.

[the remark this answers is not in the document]
dag-erling: Ah, I didn't realize that. POSIX allows us to accept lowercase strings in `str2sig()` too.

lib/libc/gen/sig2str.c
```diff
@@ -0,0 +1,110 @@
...
+
+	if (strncasecmp(str, "SIG", 3) == 0)
+		str += 3;
+
+	if (strncasecmp(str, rtmin_str, sizeof(rtmin_str) - 1) == 0 ||
+	    strncasecmp(str, rtmax_str, sizeof(rtmax_str) - 1) == 0) {
+		sig = (toupper(str[4]) == 'X') ? SIGRTMAX : SIGRTMIN;
+		n = 0;
+		if (str[rtend] == '+' || str[rtend] == '-') {
+			n = strtonum(str + rtend, 1, SIGRTMAX, &errstr);
```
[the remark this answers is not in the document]
ricardobranco777: strtonum is better as we have to do the checks anyway.

lib/libc/gen/psignal.3
```diff
@@ -108,10 +114,48 @@ contains a count of the strings in
...
+and
+.Fn psiginfo
+functions are defined by
+.St -p1003.1-2008
+, while the
+.Fn sig2str
+and
+.Fn str2sig
+functions are defined by
+.St -p1003.1-2024
```
dag-erling, line 158:
```suggestion
.St -p1003.1-2024 .
```

lib/libc/gen/psignal.3
```diff
@@ -124,3 +168,9 @@ function appeared in
 .Nx 6.0 ,
 and
 .Dx 4.1 .
+The
+.Fn sig2str
+and
+.Fn str2sig
+functions appeared in
+.Fx 15.0
```
dag-erling, line 176:
```suggestion
functions first appeared in
.Fx 15.0 .
```

lib/libc/gen/psignal.3
```diff
@@ -108,10 +114,48 @@ contains a count of the strings in
...
+.Pp
+The
+.Fn str2sig
+function translates the signal name
+.Fa str
+to a signal number and stores it in the location referenced by
+.Fa pnum .
+The name in
+.Fa str
+can be either the name of the signal, without the
```
dag-erling, line 139:
```suggestion
can be either the name of the signal, with or without the
```

usr.bin/fstat/fuser.c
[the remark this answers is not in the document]
dag-erling: Now that `str2sig()` accepts lowercase strings, this can be dropped.

dag-erling (changes requested):

usr.bin/fstat/fuser.c
```diff
@@ -199,17 +197,8 @@ do_fuser(int argc, char *argv[])
...
-				if (*ep != '\0' || sig < 0 || sig >= sys_nsig)
-					errx(EX_USAGE, "illegal signal number" ": %s",
-					    optarg);
-			} else {
-				sig = str2sig(optarg);
-				if (sig < 0)
-					errx(EX_USAGE, "illegal signal name: "
-					    "%s", optarg);
-			}
+			if (str2sig(optarg, &sig) < 0)
```
dag-erling, line 200: I prefer `!= 0` but that's entirely a matter of taste (and of consistency with surrounding code).

dag-erling (commented):
This is great! All that's missing is tests and fixing `proc_signame()`. Thank you for your hard work!

ricardobranco777:
> This is great! All that's missing is tests and fixing `proc_signame()`. Thank you for your hard work!

For fixing `proc_signame()` I mostly adapted the existing code from Illumos but using `strlcpy`:

https://github.com/illumos/illumos-gate/blob/master/usr/src/lib/libproc/common/proc_names.c#L95

lib/libproc/proc_util.c
```diff
@@ -180,8 +181,20 @@ proc_getwstat(struct proc_handle *phdl)
...
+	size_t len;
+
+	/* sig2str() omits the leading "SIG" */
+	(void)strlcpy(buf, "SIG", sizeof(buf));
+
+	if (sig2str(sig, buf + 3) == 0) {
+		(void)strlcpy(name, buf, namesz);
+	} else {
+		len = snprintf(name, namesz, "SIG#%d", sig);
+		/* ensure null-termination */
```
dag-erling, line 194: That's not needed as `snprintf()` already takes care of it. Since libproc already has ATF tests, I suggest you add a test case which verifies that `proc_signame()` not only produces the correct result but also correctly terminates the string if the buffer is too short.

ricardobranco777: > That's not needed as `snprintf()` already takes care of it. Since libproc already has ATF tests, I suggest you add a test case which verifies that `proc_signame()` not only produces the correct result but also correctly terminates the string if the buffer is too short.

You're right. Fixed it.  I think this function is not worth creating a new test since it only has 2 consumers.

bsdimp:
My plan is to review and commit monday. The feedback so far has made it better.

ricardobranco777:
> Please also consider adding SSP instrumentation, see for instance [d0b7445](https://github.com/freebsd/freebsd-src/commit/d0b7445904f592bf379d64932fb6f1fdc29c4aa7) which added instrumentation for `realpath()` which, like `sig2str()`, has an implied buffer size.

Added it in the last commit.  Only failing:

```
/usr/obj/usr/src/amd64.amd64/lib/libc/tests/secure/fortify_signal_test sig2str_heap_after_end
fortify_signal_test: WARNING: Running test cases outside of kyua(1) is unsupported
fortify_signal_test: WARNING: No isolation nor timeout control is being applied; you may get unexpected failures; see atf-test-case(4)
failed: FORTIFY_SOURCE failed to abort
```

dag-erling:
This looks good to me, just waiting for confirmation from our SSP expert. But please drop `plockstat.c` from this PR, it's third-party code and will be dealt with separately.

ricardobranco777:
> This looks good to me, just waiting for confirmation from our SSP expert. But please drop `plockstat.c` from this PR, it's third-party code and will be dealt with separately.

Fixed. Thanks!

lib/libproc/proc_util.c
```diff
@@ -180,8 +181,15 @@ proc_getwstat(struct proc_handle *phdl)
...
 proc_signame(int sig, char *name, size_t namesz)
 {
+	char buf[SIG2STR_MAX + 3];
 
-	strlcpy(name, strsignal(sig), namesz);
+	/* sig2str() omits the leading "SIG" */
+	(void)strlcpy(buf, "SIG", sizeof(buf));
+
+	if (sig2str(sig, buf + 3) == 0)
+		(void)strlcpy(name, buf, namesz);
```
dag-erling, line 190: Wouldn't it be simpler to just `snprintf(name, namesz, "SIG%s", buf)` here... no need for `+ 3` or `strlcpy()` above.

ricardobranco777: > Wouldn't it be simpler to just `snprintf(name, namesz, "SIG%s", buf)` here... no need for `+ 3` or `strlcpy()` above.

Agree.  It makes no sense to follow their style here.  Fixed.

lib/libc/gen/sig2str.c
```diff
@@ -0,0 +1,114 @@
...
+			return (-1);
+		}
+		sig += (int)n;
+		if (sig < SIGRTMIN || sig > SIGRTMAX)
+			return (-1);
+		*pnum = sig;
+		return (0);
+	}
+
+	if (isdigit(str[0])) {
```
jillest, line 98: `isdigit` needs an unsigned char.
```suggestion
	if (isdigit((unsigned char)str[0])) {
```

bin/kill/kill.c
```diff
@@ -72,12 +71,7 @@ main(int argc, char *argv[])
...
 				usage();
-			numsig = strtol(*argv, &ep, 10);
-			if (!**argv || *ep)
-				errx(2, "illegal signal number: %s", *argv);
-			if (numsig >= 128)
-				numsig -= 128;
-			if (numsig <= 0 || numsig >= sys_nsig)
+			if (str2sig(*argv, &numsig) < 0)
 				nosig(*argv);
 			printf("%s\n", sys_signame[numsig]);
```
jillest, line 76: This change seems incorrect. Things like `kill -l INT` aren't supposed to work, while it is supposed to be possible to do something like `kill -l 139` and see `SEGV` (this is so that the signal name can be obtained from a value from `$?`). This change is likely to break various tests from bin/sh including builtins/kill2.0.

A change that could be made instead would be to use `sig2str` instead of `sys_signame` access.

An issue in the existing code is that the `long` return value from `strtol` is assigned to `int numsig`.

lib/libc/gen/psignal.3
```diff
@@ -108,10 +114,48 @@ contains a count of the strings in
...
+.Pp
+The
+.Fn sig2str
+function translates the signal number
+.Fa signum
+to the signal name, without the
+.Dq SIG
+prefix, and stores it at the location specified by
+.Fa str ,
+which should be large enough to hold the name and the terminating NUL byte.
```
jillest, line 126: Existing man pages tend to mark up `NUL`.
```suggestion
which should be large enough to hold the name and the terminating
.Dv NUL
byte.
```

lib/libc/tests/gen/sig2str_test.c
```diff
@@ -0,0 +1,189 @@
...
+	ATF_REQUIRE(str2sig("RTMIN-777", &sig) != 0);
+	ATF_REQUIRE(str2sig("RTMIN+777", &sig) != 0);
+	ATF_REQUIRE(str2sig("RTMAX-777", &sig) != 0);
+	ATF_REQUIRE(str2sig("RTMAX+777", &sig) != 0);
+	ATF_REQUIRE(str2sig("RTMIN-", &sig) != 0);
+	ATF_REQUIRE(str2sig("RTMAX-", &sig) != 0);
+	ATF_REQUIRE(str2sig("RTMIN0", &sig) != 0);
+	ATF_REQUIRE(str2sig("RTMAX1", &sig) != 0);
+	ATF_REQUIRE(str2sig("RTMIN+abc", &sig) != 0);
+	ATF_REQUIRE(str2sig("RTMIN-abc", &sig) != 0);
```
jillest, line 101: `ATF_CHECK` might be better here so more information can be obtained from a test failure.

lib/libc/tests/gen/sig2str_test.c
```diff
@@ -0,0 +1,189 @@
...
+ATF_TC_WITHOUT_HEAD(str2sig_fullname);
+ATF_TC_BODY(str2sig_fullname, tc)
+{
+	char fullname[SIG2STR_MAX + 3];
+	int n, sig;
+
+	for (sig = 1; sig < sys_nsig; sig++) {
+		snprintf(fullname, sizeof(fullname), "SIG%s", sys_signame[sig]);
+
+		ATF_REQUIRE_MSG(str2sig(fullname, &n) == 0,
```
jillest, line 113: `ATF_REQUIRE_INTEQ_MSG` might be better so that a failing test reports the error number.

With a little more code (skip the check on `n` on failure), `ATF_CHECK_INTEQ_MSG` could be used so the other signal numbers can be checked after a failure.

lib/libc/tests/gen/sig2str_test.c
```diff
@@ -0,0 +1,189 @@
...
+		    "str2sig(\"%s\") failed", fullname);
+
+		ATF_REQUIRE_MSG(n == sig,
```
jillest, line 116: `ATF_CHECK_INTEQ_MSG` please to make failures more informative and simplify code.

jillest (commented):
Very nice to have this; having to use non-portable `sys_nsig` and `sys_signame` in bin/sh has annoyed me for a long time.

kevans91:
> > Please also consider adding SSP instrumentation, see for instance [d0b7445](https://github.com/freebsd/freebsd-src/commit/d0b7445904f592bf379d64932fb6f1fdc29c4aa7) which added instrumentation for `realpath()` which, like `sig2str()`, has an implied buffer size.
> 
> Added it in the last commit. All tests passing.

Thanks for taking the time- current version of the SSP bits + associated test LGTM.

bin/kill/kill.c
ricardobranco777: > An issue in the existing code is that the `long` return value from `strtol` is assigned to `int numsig`.

For this I need to use strtonum as intended before.

lib/libc/gen/sig2str.c
ricardobranco777: Fixed. That's why my initial implementation lacked isdigit().

bin/kill/kill.c
ricardobranco777: > This change seems incorrect. Things like `kill -l INT` aren't supposed to work

Why not?  It works in Bash at least.

> while it is supposed to be possible to do something like `kill -l 139` and see `SEGV`

This is working.

> A change that could be made instead would be to use `sig2str` instead of `sys_signame` access.

Will look into this.

ricardobranco777: Restored the functionality with numeric values above 128 being special in `kill`.

jillest (approved):

dag-erling (approved):

ricardobranco777:
I've been using this on production with no issues.

bsdimp:
I think this is ready to go and all the feedback has been addressed. Is that assessment correct?

ricardobranco777:
ping @bsdimp

bsdimp:

svmhdvn:
This seems to break `bin/pwait/tests/pwait_test:basic` and `bin/pwait/tests/pwait_test:or_flag`, due to the usage of `kill -0`. This implementation of `sig2str` seems to return `-1` for sig 0. Could someone take a look? POSIX states:

> If signum is equal to 0, the behavior is unspecified.

ricardobranco777:
> This seems to break `bin/pwait/tests/pwait_test:basic` and `bin/pwait/tests/pwait_test:or_flag`, due to the usage of `kill -0`. This implementation of `sig2str` seems to return `-1` for sig 0. Could someone take a look? POSIX states:
> 
> > If signum is equal to 0, the behavior is unspecified.

Fixed in https://github.com/freebsd/freebsd-src/pull/1733

Thanks!

PR #1697, opened 2025-05-15
racct_add_racct and racct_sub_racct should be restrict

AZero13:
It probably will be inlined anyway, but if the compiler knows they do not overlap, then it can help especially when dealing with array math.


bsdimp:
So the commit message needs some work to explain why restrict is needed.

bsdimp:
No explanation proffered.  Closing for lack of feedback.

PR #1699, opened 2025-05-17
If the hostname is not set for the host, the value "Amnesiac" should be written to rc.conf

ykla:
see also https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=286847

see also https://docs.freebsd.org/en/books/handbook/network/#hostname

If no hostname has been set for the host FreeBSD will assign the value Amnesiac.

This operation does not actually work because the hostname setting in `rc.conf` is `hostname=""`. As a result, all software, including the output of the `hostname` command, still shows an empty value. For Xorg, a setting like `hostname=""` will prevent the GUI from starting. If the hostname is not set for the host, the value `"Amnesiac"` should be written to `rc.conf`.

Why is the hostname empty (`hostname=""`): because the user simply pressed the Enter key at the hostname setup step during the FreeBSD installation.

Signed-off-by: ykla <yklaxds@gmail.com>

### DIFF
--- a/usr.sbin/bsdinstall/scripts/hostname
+++ b/usr.sbin/bsdinstall/scripts/hostname
@@ -111,15 +111,15 @@ HOSTNAME=$( dialog_hostname "$HOSTNAME" )
 # Store the user's choice
 #
 f_eval_catch "$pgm" echo "$ECHO_OVERWRITE" \
-    'hostname=\"$HOSTNAME\"' "$HOSTNAMEFILE"
+	'hostname=\"$HOSTNAME\"' "$HOSTNAMEFILE"
 retval=$?
 
 #
 # Activate entry if configured
 #
 if [ "$BSDINSTALL_CONFIGCURRENT" ]; then
-    f_eval_catch "$pgm" hostname "$SET_HOSTNAME" "$HOSTNAME"
-    retval=$?
+	f_eval_catch "$pgm" hostname "$SET_HOSTNAME" "$HOSTNAME"
+	retval=$?
 fi
 
 exit $retval
--- a/usr.sbin/bsdinstall/scripts/hostname
+++ b/usr.sbin/bsdinstall/scripts/hostname
@@ -103,19 +103,16 @@ HOSTNAME=$( dialog_hostname "$HOSTNAME" )
 [ $? -eq $DIALOG_CANCEL ] && exit 1
 
 #
+# Set hostname to Amnesiac if empty
+#
+[ -z "$HOSTNAME" ] && HOSTNAME="Amnesiac"
+
+#
 # Store the user's choice
 #
 f_eval_catch "$pgm" echo "$ECHO_OVERWRITE" \
-	'hostname=\"$HOSTNAME\"' "$HOSTNAMEFILE"
+    'hostname=\"$HOSTNAME\"' "$HOSTNAMEFILE"
 retval=$?
-
-#
-# Activate entry if configured
-#
-if [ "$BSDINSTALL_CONFIGCURRENT" ]; then
-	f_eval_catch "$pgm" hostname "$SET_HOSTNAME" "$HOSTNAME"
-	retval=$?
-fi
 
 exit $retval
 
--- a/usr.sbin/bsdinstall/scripts/hostname
+++ b/usr.sbin/bsdinstall/scripts/hostname
@@ -110,9 +110,16 @@ HOSTNAME=$( dialog_hostname "$HOSTNAME" )
 #
 # Store the user's choice
 #
+f_eval_catch "$pgm" echo "$ECHO_OVERWRITE" \
+    'hostname=\"$HOSTNAME\"' "$HOSTNAMEFILE"
+retval=$?
+
+#
+# Activate entry if configured
+#
 if [ "$BSDINSTALL_CONFIGCURRENT" ]; then
-	f_eval_catch "$pgm" hostname "$SET_HOSTNAME" "$HOSTNAME"
-	retval=$?
+    f_eval_catch "$pgm" hostname "$SET_HOSTNAME" "$HOSTNAME"
+    retval=$?
 fi
 
 exit $retval
--- a/usr.sbin/bsdinstall/scripts/hostname
+++ b/usr.sbin/bsdinstall/scripts/hostname
@@ -110,9 +110,10 @@ HOSTNAME=$( dialog_hostname "$HOSTNAME" )
 #
 # Store the user's choice
 #
-f_eval_catch "$pgm" echo "$ECHO_OVERWRITE" \
-    'hostname=\"$HOSTNAME\"' "$HOSTNAMEFILE"
-retval=$?
+if [ "$BSDINSTALL_CONFIGCURRENT" ]; then
+	f_eval_catch "$pgm" hostname "$SET_HOSTNAME" "$HOSTNAME"
+	retval=$?
+fi
 
 exit $retval
 


jrtc27:
Amnesiac isn't the actual hostname that gets set if you omit hostname from rc.conf, though. That's just what getty substitutes if hostname is empty.

Alex6357:
Unless there are specific use cases I'm unaware of, hostname should always be configured. Therefore, setting a default value during installation is reasonable. A more optimal solution might be to implement a yesno dialog.

jrtc27:
Banning an empty hostname in the dialog would seem reasonable to me. As would giving it a default value (in CheriBSD we use the default of "cheribsd"; FreeBSD may wish to default to "freebsd", like Debian IIRC defaults to "debian").

jrtc27:
Having said that, there is a legitimate use case for leaving it blank if the hostname is to come from a DHCP server.

ykla:
> Banning an empty hostname in the dialog would seem reasonable to me. As would giving it a default value (in CheriBSD we use the default of "cheribsd"; FreeBSD may wish to default to "freebsd", like Debian IIRC defaults to "debian").

The current documentation and the code shown in the previous login section are both Amnesiac. It should not be difficult for users to set it to empty after installing the system.

jrtc27:
> > Banning an empty hostname in the dialog would seem reasonable to me. As would giving it a default value (in CheriBSD we use the default of "cheribsd"; FreeBSD may wish to default to "freebsd", like Debian IIRC defaults to "debian").
> 
> The current documentation and the code shown in the previous login section are both Amnesiac.

What documentation and code shown in the previous login section?

concussious:
> It should not be difficult for users to set it to empty after installing the system.

That would be a regression for that use case.

ceri:
There's already a warning in rc.d/hostname if hostname is not set and an explicit default of "" in defaults/rc.conf which would be the appropriate place to fix this if it is deemed incorrect.  Setting a default value of "Amnesiac" would obviate all the code in rc.d/hostname that tries to default to the DHCP given name, which I would suggest is a better default along with the very clear existing warning that is issued if $hostname is indeed still unset after those checks.

ceri:
If different behaviour is needed in the installer then it should be a check for an empty hostname and a warning at that point, in my opinion.

bsdimp:
Thank you for this submission. However, this problem should be solved elsewhere. If you do solve it elsewhere, please open a new pull request.

NatUni:
This is a real Chesterton's Fence, if i've ever seen one.

I think this is as good a place as any for this reference: [Joe Bauers becomes Not Sure](https://youtu.be/sVyRkl5qNb8)

@ykla i sincerely thank you kindly for your suggestion and your participation; i do not want you to take this the wrong way. Yes. Bring this energy, and don't slow down on account of me or this. I take seriously the responsibility to constructively educate my juniors and all readers...but i am not the most tactful or diplomatic, i know. I don't contribute much, and i've never been one to bike-shed. This is a teachable moment; i really do not mean to condescend.

You know `NULL !== "NULL"` and you wouldn't force it to be so, i believe. Empty string has a particular and undeniably singular identity, one that any other chosen value simply cannot. There is no way to know what it could mean in certain contexts or what state of a system it might correspond with.

Nag the user, nag them hard `while [ $hostname = '' ]` if you feel you must (given that we live in an ever-connected always online world, for better or worse)...but do not override the empty string _way_ up **here** without clear and unmistakable intent from the end-user.

We don't want to inconvenience our users but the surest way to do that with the greatest consequence is to guess what the user wants or tell them what they should want.

If nginx or xorg or anything the heck else, refuses to play ball because they've been naively constructed to assume that everyone's parents loved them enough to name them, or, that everyone should want to talk to everyone these days, then those presumptions and deficiencies need to be addressed particularly and individually where they are. Yes, we will find no shortage of software, which has zero business worrying about network connectivity, stubbornly actin' a fool because hostname is empty. We can populate an arbitrary string (preferably anything *but* "Amnesiac") at _those_ points, to flow _down_stream, as prudent or necessary. You have proposed a [seemingly trivial, obvious and admittedly desirable] *perfectly clever* change at the...Very. Top...of the mountain. It would solve so many issues with so little effort.

We have to think of the `$LUSER` (legitimate user). This has been this way for...Over. Forty. Years. The string "Amnesiac" is in the _**presentation** layer_ (getty/main.c) and it's become a word we're all familiar with...but we were never meant to become _too_ familiar. Amnesiacs like their solitude, their isolation.

I, for one, practice defense in depth just as you do, i'm sure...but there exist deployments, ridiculous, i agree...there are systems out there...where (?...we will never know) hostname equaling empty string is the only 'firewall' and only security layer some even have! Consider, if you will, automatic this and AI-powered that...now that Bluetooth and WiFi and NFC are two pennies worth of every single device outnumbering life itself on Earth...where tech such as Rendezvous has become the norm and entirely taken for granted...or dismissed, overlooked. Should those folks be encouraged to extract their craniums from their posteriors and get with the program? Of course. Chastised and shamed? Probably.

If i'm designing a pacemaker with BSD at its core i should already know how to compile a kernel entirely lacking a networking stack...but people don't; they're too busy keeping satellites in orbit over our heads or using this tech to make applications, implementations, deployments. Things we can't imagine.

I realize we're talking about a hostname and it _should_ have nothing to influence regarding obtaining an identifier at half a dozen lower layers. Please don't squabble over the stupidity of my example because admittedly it's only a contrived one. Nobody wants their infrastructure or operating system to "change" yet somehow everyone still demands it get "better"...and we have no shortage of ways we can improve FreeBSD.

The most important way, most impactful way, is communicating with each other, respectfully and tenderly. I hope you feel i take you seriously and considerately.

This is BSD; we don't demand users make their implementation details public nor stay in our consult. However, many an obscure and esoteric system will still run `freebsd-update` from time to time. POLA applies. Backward compatibility and legacy support, respect for the way things were and still may be, is a cardinal reason that i'm a BSD user, not shedding my blood, sweat and tears into that...other comparable community.

I'm not trying to ridicule you and i don't want you to feel bad. I hope i've conveyed some of the sentiments that others have already felt and struggled to articulate clearly and sufficiently. I pray you appreciate my insights as informative, with the respect i intend, and they contribute to your unique ongoing journey of discovery and expression. You may feel honored that you've awaken a sleeping giant by your quite reasonable...Very. Sensible. Suggestion. It's not my place to feed others crow. I pray the energy i have dispensed in addressing this issue, late as i may be, exhibits that i, we all, care not only about this community but very much so in _your_ sustained and enthusiastic involvement. You are warmly welcomed to the inner circle of the elite whose merge requests to Resource Control over Planet Earth have sparked very serious consideration and interesting discussion.

@bsdimp thank *Gawd* you're still kickin', good sir. Glory to you and your haaaus!

ykla:
> This is a real Chesterton's Fence, if i've ever seen one.
> 
> I think this is as good a place as any for this reference: [Joe Bauers becomes Not Sure](https://youtu.be/sVyRkl5qNb8)
> 
> @ykla i sincerely thank you kindly for your suggestion and your participation; i do not want you to take this the wrong way. Yes. Bring this energy, and don't slow down on account of me or this. I take seriously the responsibility to constructively educate my juniors and all readers...but i am not the most tactful or diplomatic, i know. I don't contribute much, and i've never been one to bike-shed. This is a teachable moment; i really do not mean to condescend.
> 
> You know `NULL !== "NULL"` and you wouldn't force it to be so, i believe. Empty string has a particular and undeniably singular identity, one that any other chosen value simply cannot. There is no way to know what it could mean in certain contexts or what state of a system it might correspond with.
> 
> Nag the user, nag them hard `while [ $hostname = '' ]` if you feel you must (given that we live in an ever-connected always online world, for better or worse)...but do not override the empty string _way_ up **here** without clear and unmistakable intent from the end-user.
> 
> We don't want to inconvenience our users but the surest way to do that with the greatest consequence is to guess what the user wants or tell them what they should want.
> 
> If nginx or xorg or anything the heck else, refuses to play ball because they've been naively constructed to assume that everyone's parents loved them enough to name them, or, that everyone should want to talk to everyone these days, then those presumptions and deficiencies need to be addressed particularly and individually where they are. Yes, we will find no shortage of software, which has zero business worrying about network connectivity, stubbornly actin' a fool because hostname is empty. We can populate an arbitrary string (preferably anything *but* "Amnesiac") at _those_ points, to flow _down_stream, as prudent or necessary. You have proposed a [seemingly trivial, obvious and admittedly desirable] *perfectly clever* change at the...Very. Top...of the mountain. It would solve so many issues with so little effort.
> 
> We have to think of the `$LUSER` (legitimate user). This has been this way for...Over. Forty. Years. The string "Amnesiac" is in the _**presentation** layer_ (getty/main.c) and it's become a word we're all familiar with...but we were never meant to become _too_ familiar. Amnesiacs like their solitude, their isolation.
> 
> I, for one, practice defense in depth just as you do, i'm sure...but there exist deployments, ridiculous, i agree...there are systems out there...where (?...we will never know) hostname equaling empty string is the only 'firewall' and only security layer some even have! Consider, if you will, automatic this and AI-powered that...now that Bluetooth and WiFi and NFC are two pennies worth of every single device outnumbering life itself on Earth...where tech such as Rendezvous has become the norm and entirely taken for granted...or dismissed, overlooked. Should those folks be encouraged to extract their craniums from their posteriors and get with the program? Of course. Chastised and shamed? Probably.
> 
> If i'm designing a pacemaker with BSD at its core i should already know how to compile a kernel entirely lacking a networking stack...but people don't; they're too busy keeping satellites in orbit over our heads or using this tech to make applications, implementations, deployments. Things we can't imagine.
> 
> I realize we're talking about a hostname and it _should_ have nothing to influence regarding obtaining an identifier at half a dozen lower layers. Please don't squabble over the stupidity of my example because admittedly it's only a contrived one. Nobody wants their infrastructure or operating system to "change" yet somehow everyone still demands it get "better"...and we have no shortage of ways we can improve FreeBSD.
> 
> The most important way, most impactful way, is communicating with each other, respectfully and tenderly. I hope you feel i take you seriously and considerately.
> 
> This is BSD; we don't demand users make their implementation details public nor stay in our consult. However, many an obscure and esoteric system will still run `freebsd-update` from time to time. POLA applies. Backward compatibility and legacy support, respect for the way things were and still may be, is a cardinal reason that i'm a BSD user, not shedding my blood, sweat and tears into that...other comparable community.
> 
> I'm not trying to ridicule you and i don't want you to feel bad. I hope i've conveyed some of the sentiments that others have already felt and struggled to articulate clearly and sufficiently. I pray you appreciate my insights as informative, with the respect i intend, and they contribute to your unique ongoing journey of discovery and expression. You may feel honored that you've awaken a sleeping giant by your quite reasonable...Very. Sensible. Suggestion. It's not my place to feed others crow. I pray the energy i have dispensed in addressing this issue, late as i may be, exhibits that i, we all, care not only about this community but very much so in _your_ sustained and enthusiastic involvement. You are warmly welcomed to the inner circle of the elite whose merge requests to Resource Control over Planet Earth have sparked very serious consideration and interesting discussion.
> 
> @bsdimp thank *Gawd* you're still kickin', good sir. Glory to you and your haaaus!
> 

Thanks for your response. I've submitted a new PR that adds a warning when the hostname is empty. See also <https://github.com/freebsd/freebsd-src/pull/1700>

PR #1703, opened 2025-05-19
newsyslog: Add tests for rotating files based on size

bdesham:
The tests for newsyslog(8) did not previously test size-based rotation, e.g., “rotate `foo.log` once it becomes larger than 8 KB.” This pull request adds a few tests for this functionality.

### DIFF
--- a/usr.sbin/newsyslog/tests/legacy_test.sh
+++ b/usr.sbin/newsyslog/tests/legacy_test.sh
@@ -205,8 +205,7 @@ tmpdir_clean()
 
 run_newsyslog()
 {
-
-	newsyslog -f ../newsyslog.conf -F -r "$@"
+	newsyslog -f ../newsyslog.conf -r "$@"
 }
 
 tests_normal_rotate() {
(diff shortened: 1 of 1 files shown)


asomers (changes requested):
I would rather not add additional steps to a pre-ATF test.  Could you instead please write a new file, using atf-c, and put the tests in there?

bdesham:
@asomers I’m happy to revise this, but I’m not very familiar with FreeBSD’s testing infrastructure. Are you requesting that I rewrite the tests in C specifically? I would feel a lot more comfortable rewriting them as an ATF shell script (like [usr.sbin/sysrc/tests/sysrc_test.sh](https://github.com/freebsd/freebsd-src/blob/main/usr.sbin/sysrc/tests/sysrc_test.sh), just to pick a concrete example).

markjdb:
> @asomers I’m happy to revise this, but I’m not very familiar with FreeBSD’s testing infrastructure. Are you requesting that I rewrite the tests in C specifically? I would feel a lot more comfortable rewriting them as an ATF shell script (like [usr.sbin/sysrc/tests/sysrc_test.sh](https://github.com/freebsd/freebsd-src/blob/main/usr.sbin/sysrc/tests/sysrc_test.sh), just to pick a concrete example).

I think Alan is just asking for new tests to use the ATF framework, so atf-sh is perfectly reasonable. The existing newsyslog tests don't use ATF only because they predate it.

bsdimp:
Is there any update on this? It conflicts and is older. I'd like to see this update so it can be committed or closed.

PR #1704, opened 2025-05-20
sed: Fix usage guidelines

AndrewC243:
When viewing the usage guidelines of sed, whether by invoking it with an invalid argument or through viewing its manual page, it is suggested that [-i extension] or [-I extension] can or should be used after supplying a command, when in fact they can only be used before supplying a command.

Additionally, the manual pages and usage guidelines use inconsistent language by using "script" and "command" interchangeably. This change commits to using "command."

### DIFF
--- a/usr.bin/sed/main.c
+++ b/usr.bin/sed/main.c
@@ -198,8 +198,8 @@ static void
 usage(void)
 {
 	(void)fprintf(stderr,
-	    "usage: %s script [-Ealnru] [-i extension] [file ...]\n"
-	    "\t%s [-Ealnu] [-i extension] [-e script] ... [-f script_file]"
+	    "usage: %s [-Ealnru] [-i extension] command [file ...]\n"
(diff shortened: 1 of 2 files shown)


usr.bin/sed/main.c
```diff
@@ -198,8 +198,8 @@ static void
 usage(void)
 {
 	(void)fprintf(stderr,
-	    "usage: %s script [-Ealnru] [-i extension] [file ...]\n"
-	    "\t%s [-Ealnu] [-i extension] [-e script] ... [-f script_file]"
+	    "usage: %s [-Ealnru] [-i extension] command [file ...]\n"
```
bsdimp, line 201: So I'm not sure that 'command' is better than 'script' here. Since all the places that have 'script' in them are a series of commands... I almost think we should change the man page... Lemme see what other systems use here...

bsdimp:
I'm sorry we can't use this. POSIX standard uses 'script' rather than 'command' so I'm going to NAK this patch. We do not want to make this change since we try to use POSIX language in our man pages. Sorry.

PR #1706, opened 2025-05-23
Fix full memory dumps on amd64

VexedUXR:
Currently, libkvm can't read full memory dumps correctly on amd64 since it assumes the kernel was loaded at 2M physical.   Fix this by adding the kernel's physical address in an auxiliary ELF section, similar to what arm does.

This seems to have been broken since 13.1 (see 8ca493ffb44691e70ae92300b8de1c1d30134ef4 and 1b33aa1f5f99e1270d526ffa5b652250ec80a7ef), so document that.

From my testing, this also stops kgdb and lldb from crashing when trying to open full memory dumps.

### DIFF
--- a/lib/libkvm/kvm_amd64.c
+++ b/lib/libkvm/kvm_amd64.c
@@ -109,8 +109,9 @@ _amd64_initvtop(kvm_t *kd)
 {
 	struct kvm_nlist nl[2];
 	amd64_physaddr_t pa;
-	kvaddr_t kernbase;
+	kvaddr_t kernbase, kernphys;
 	amd64_pml4e_t *PML4;
+	int found = 0;
 
 	kd->vmst = (struct vmstate *)_kvm_malloc(kd, sizeof(*kd->vmst));
 	if (kd->vmst == NULL) {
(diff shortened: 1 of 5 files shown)


lib/libkvm/kvm_amd64.c
```diff
@@ -123,16 +124,43 @@ _amd64_initvtop(kvm_t *kd)
...
+		if (kvm_nlist2(kd, nl) != 0) {
+			_kvm_err(kd, kd->program, "bad namelist - no kernbase");
+			return (-1);
+		}
+
+		nl[0].n_name = "kernphys";
+		nl[1].n_name = 0;
+
+		/* XXX
+		 * Relocatable kernels can still be loaded at 2M.
```
VexedUXR, line 154: Kernels loaded with copy_staging enabled in the loader (which isn't the default) will still be at 2M. There doesn't seem to be a way to detect if they were actually loaded there though.

The only way to read such dumps seems to be to read the dump anyways, at the risk of reading garbage.

lib/libkvm/kvm_open.3
```diff
@@ -245,6 +245,14 @@ The value passed via
...
 .Dv NULL .
 .El
+.Sh NOTE
+Full memory dumps taken on 13.x (excluding 13.0) and 14.x amd64 kernels
+will cause both
+.Fn kvm_open
+and
+.Fn kvm_open2
+to fail since they do not provide sufficient information to figure out
+where in physical memory the kernel was loaded.
```
markjdb, line 255: I might further comment that minidumps are the default (`sysctl debug.minidump`), so this is unlikely to bite people.

We really ought to have a kerneldump.4 man page which documents these kinds of things (what's a "full dump", what's a minidump, what do they look like at a high level, etc. etc.)...

markjdb (approved):
This looks ok to me. I'm not sure the note in the man page is really necessary given that this PR also fixes the bug, but I think it's ok, we should just explain the problem a bit further. (Do folks still use full memory dumps?)

lib/libkvm/kvm_open.3
VexedUXR: > We really ought to have a kerneldump.4 man page which documents these kinds of things (what's a "full dump", what's a minidump, what do they look like at a high level, etc. etc.)...

I think dumpon(8) partly explains this, but yea they should be explained more thoroughly in their own man page.

VexedUXR:
Mention that full memory dumps have to be explicitly enabled.

> (Do folks still use full memory dumps?)

Given that no one seems to have noticed this in the last 4 or so years, I'd say probably not.

lib/libkvm/kvm_open.3
markjdb: Ah, right, I forgot about dumpon(8). You might consider adding an `.Xr dumpon 8` too, then.

VexedUXR:
Added dumpon(8) to the `SEE ALSO` section.

VexedUXR:
@bsdimp I think this is ready now

sys/x86/x86/dump_machdep.c
```diff
@@ -52,3 +59,22 @@ dumpsys_map_chunk(vm_paddr_t pa, size_t chunk, void **va)
...
+	phdr.p_type = PT_DUMP_DELTA;
+	phdr.p_flags = PF_R;
+	phdr.p_offset = 0;
+	phdr.p_vaddr = KERNBASE;
+	phdr.p_paddr = kernphys;
+	phdr.p_filesz = 0;
+	phdr.p_memsz = 0;
+	phdr.p_align = KERNLOAD;
+
+	return (dumpsys_buf_write(di, (char *)&phdr, sizeof(phdr)));
```
bsdimp, line 78: Do we need to call the generic on too?

VexedUXR: I don't think so. It's currently a stub and the existing arm code doesn't call it.

markjdb: I agree.

markjdb:
I tested this in a VM and it appears to work fine: kgdb opens the core dump and enumerates threads without problems. `kernphys` is 0x33c00000.

markmi:
> Mention that full memory dumps have to be explicitly enabled.
> 
> > (Do folks still use full memory dumps?)
> 
> Given that no one seems to have noticed this in the last 4 or so years, I'd say probably not.

I was aware of there being a problem that I’d never investigated.

PR #1707, opened 2025-05-24
Apply libxo to a part of nvmecontrol

kenrap:
@bsdimp

### DIFF
--- a/sbin/nvmecontrol/modules/wdc/wdc.c
+++ b/sbin/nvmecontrol/modules/wdc/wdc.c
@@ -30,6 +30,7 @@
 #include <ctype.h>
 #include <err.h>
 #include <fcntl.h>
+#include <libxo/xo.h>
 #include <stddef.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -202,7 +203,7 @@ wdc_do_dump_e6(int fd, char *tmpl, const char *suffix, uint32_t opcode,
 			if (len == 0)
 				errx(EX_PROTOCOL, "No data for %s", suffix);
 
-			printf("Dumping %d bytes of version %d.%d log to %s\n", len,
+			xo_emit("Dumping {:length/%d} bytes of version {:buffer8/%d}.{:buffer9/%d} log to {:template/%s}\n", len,
 			    buf[8], buf[9], tmpl);
 			/*
 			 * Adjust amount to dump if total dump < 1MB,
@@ -347,7 +348,7 @@ wdc_do_dump_dui(int fd, char *tmpl, uint8_t data_area,
 		if (first) {
 			hdr_ver = ((buf[len_off] & 0xF) != 0) ?
 			    (buf[len_off]) : (le16dec(buf + len_off));
-			printf("Dumping %jd bytes of version %d log to %s\n",
+			xo_emit("Dumping {:log-length/%jd} bytes of version {:hdr-version/%d} log to {:template/%s}\n",
 			    (uintmax_t)log_len, hdr_ver, tmpl);
 			first = 0;
 		}
@@ -469,7 +470,7 @@ print_hgst_info_subpage_gen(void *buf, uint16_t subtype __unused, uint32_t size,
 		param = 0;
 		for (i = 0; i < plen && wsp < esp; i++)
 			param |= (uint64_t)*wsp++ << (i * 8);
-		printf("  %-30s: %jd\n", kv_lookup(kv, kv_count, ptype), (uintmax_t)param);
+		xo_emit("{P:  }{:kv-lookup/%-30s}: {:parameter/%jd}\n", kv_lookup(kv, kv_count, ptype), (uintmax_t)param);
 	}
 }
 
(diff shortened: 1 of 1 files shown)


concussious:
Does this need an update to nvmecontrol(8)? If so, I can help!

kenrap:
I'm not sure. The changes here just give it the ability to change outputs for different uses.

concussious:
So the user facing behavior will change?

bsdimp:
We'd need do document `--libxo` command line argument. Generally, we don't document the schema used, but maybe we should start. I don't want to gate this commit, though, on that.

bsdimp:
appplogies for the delay. Do you get any warnings with --libxo:XPW or --libxo:JPW?

kenrap:
It doesn't recognize both as options for the `logpage` and `devlist` commands... 😅

kenrap:
So, I didn't do all of nvmecontrol because it would be a huge diff count and didn't want to create extra review work. I could spend time getting all of nvmecontrol libxo-ified, but imagining the diff count being at least in the 4 digits was what made me submit this work much earlier.

sbin/nvmecontrol/logpage.c
```diff
@@ -491,47 +492,47 @@ print_log_sanitize_status(const struct nvme_controller_data *cdata __unused,
...
 		break;
 	default:
-		printf("Unknown 0x%x", sstat);
+		xo_emit("Unknown 0x{:sstat/%x}", sstat);
 		break;
 	}
 	p = NVMEV(NVME_SS_PAGE_SSTAT_PASSES, sstat);
 	if (p > 0)
-		printf(", %d passes", p);
+		xo_emit(", {:p/%d}{N:passes}", p);
```
philshafer: Use tag names like "passes" instead of the shorter and less meaningful "p".

There a lot of information about picking tag names in:

https://libxo.readthedocs.io/en/latest/faq.html#what-makes-a-good-field-name

sbin/nvmecontrol/logpage.c
```diff
@@ -491,47 +492,47 @@ print_log_sanitize_status(const struct nvme_controller_data *cdata __unused,
...
-	printf("Time For Crypto Erase No-Deallocate: %u sec\n", letoh(ss->etfcewnd));
+		xo_emit(", Global Data Erased");
+	xo_emit("\n");
+	xo_emit("{Lc:Sanitize Command Dword 10}{P:           }0x%x\n", letoh(ss->scdw10));
+	xo_emit("{Lc:Time For Overwrite}{P:                  }%u sec\n", letoh(ss->etfo));
+	xo_emit("{Lc:Time For Block Erase}{P:                }%u sec\n", letoh(ss->etfbe));
+	xo_emit("{Lc:Time For Crypto Erase}{P:               }%u sec\n", letoh(ss->etfce));
+	xo_emit("{Lc:Time For Overwrite No-Deallocate}{P:    }%u sec\n", letoh(ss->etfownd));
+	xo_emit("{Lc:Time For Block Erase No-Deallocate}{P:  }%u sec\n", letoh(ss->etfbewnd));
+	xo_emit("{Lc:Time For Crypto Erase No-Deallocate}{P: }%u sec\n", letoh(ss->etfcewnd));
```
philshafer: The values here aren't in a field.  The "%u" won't be formatted; you end up with just "%u" in your output.

% xo "meaning %u\n" 42
meaning %u
% xo "meaning {:answer/%u}\n" 42
meaning 42

Always test your output to make sure it makes sense in text, xml, and json.

sbin/nvmecontrol/logpage.c
```diff
@@ -558,31 +559,31 @@ print_log_self_test_status(const struct nvme_controller_data *cdata __unused,
...
+		xo_emit("{d:page-test-name/Vendor specific}{e:page-test/vendor-specific}\n");
 		break;
 	default:
-		printf("Reserved (0x%x)\n", letoh(dst->curr_operation));
+		xo_emit("Reserved (0x{:curr-operation/%x})\n", letoh(dst->curr_operation));
 	}
 
 	if (letoh(dst->curr_operation) != 0)
-		printf("Current Completion: %u%%\n", letoh(dst->curr_compl) & 0x7f);
+		xo_emit("Current Completion: {:current-completion/%u}%%\n", letoh(dst->curr_compl) & 0x7f);
```
philshafer: You only need a single "%" outside of fields, since all text outside the braces is considered "plain":

% xo "working: {:complete/%u}%\n" 55
working: 55%

xo(1) is useful for testing your understanding, though admittedly all the arguments are really strings.
Also be sure to use the warning flag (--libxo:W) to see problems that libxo detects and handles (as much as possible).

sbin/nvmecontrol/logpage.c
```diff
@@ -233,11 +234,11 @@ print_log_error(const struct nvme_controller_data *cdata __unused, void *buf, ui
 	uint8_t					p, sc, sct, m, dnr;
 	struct nvme_error_information_entry	*entry = buf;
 
-	printf("Error Information Log\n");
-	printf("=====================\n");
+	xo_emit("{T:Error Information Log}\n");
+	xo_emit("{T:=====================}\n");
```
philshafer, line 238: I know this is "a part of ....", but there are no "organizational" xo calls in this code.  Take a look at xo_open/close_container/list() functions, which are used to make hierarchy in the "encoding" style (xml, json, etc) encodings.  Perhaps those calls are handled elsewhere, but I wanted to point this out.

bsdimp:
@kenrap have you updated based on feedback?

kenrap:
I applied all feedback except for the tip on the `xo_open/close_container/list()` functions, since I don't know how to go about that yet. I also rebased and fixed a merge conflict with `sbin/nvmecontrol/Makefile`.

@philshafer Thanks for the tips and help! Using `xo` as a checker tool is very useful. 😄

kenrap:
Closing in favor of https://github.com/freebsd/freebsd-src/pull/1972

PR #1708, opened 2025-05-24
[DRAFT] devd: rewrite in C

ngortheone:
WIP


jrtc27:
I don't think this is an improvement, and would be against accepting this into the tree, unless there is a convincing argument put forward for why it must be in C.

arrowd:
I'd love to see at least a short rationale for this change.

ngortheone:
after talking to @bsdimp more we decided to scratch this.

PR #1713, opened 2025-05-29
/etc/pkg/FreeBSD.conf latest, quarterly: comments

grahamperrin:
Since base 29f076ae0cb5, neither of these files is for a single repository.

In each file, update the comments to give an example for one of the two repos.

MFC after: 3 days.


grahamperrin:
@cperciva

grahamperrin:
https://github.com/freebsd/freebsd-src/commit/955ab1a8f21ec1a662a87fc5e76e6d5690e09085

PR #1714, opened 2025-06-02
ipfw: add nomatch_pass option for nat64lsn

anavapeur:
Option to continue rule search if nat64lsn did not match a state.


ghost:
ae@ in freebsd_ru (Telegram) pointed out issues with this, so I am closing this until I understand the problem better.

PR #1715, opened 2025-06-05
Add USB quirk for Creative Stage SE mini

janbiedermann:
usb_msc_auto_quirk() would otherwise crash it.


janbiedermann:
see https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=287333

PR #1716, opened 2025-06-06
FIX: usb_msc_auto_quirk() called when UQ_MSC_IGNORE is set

janbiedermann:
usb_msc_auto_quirk() potentially crashes usb devices with a hidden storage (see bug 287333). These devices may still operate normally if usb_msc_auto_quirk() would not be called all and the hidden storage is ignored. This patch makes sure, that usb_msc_auto_quirk() is not called when the UQ_MSC_IGNORE quirk is set for a device. It shouldn't be called anyway if the hidden storage supposed to be ignored. This gives users a chance to get their devices working by using 'usbconfig add_dev_quirk_vplh <vid> <pid> <lo_rev> <hi_rev> UQ_MSC_IGNORE'.


janbiedermann:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=287343

bsdimp:
This is good for stable/14, but I think in `main` we never call this routine.... Though there's an ifdef there that might not be bad to MFC...  I'll look to land this soon.

bsdimp:
So this can go in as is. It's a nop for 15 though

bsdimp:

PR #1718, opened 2025-06-12
Update legacy.sh to do both MBR and EFI booting, keep data partition

tickerguy:
Recompute the partition layout and set up for an EFI partition on 3, a BSD one on 4 containing both cfg and data, thus preserving the two-code partition capability (for "hot" updates.)  Result boots on both legacy (e.g. pcEngines apu2) and EFI-only "small box" modern devices (tested on two I have here that will not boot legacy USB media at all along with the pcEngines units I have here)

### DIFF
--- a/tools/tools/nanobsd/legacy.sh
+++ b/tools/tools/nanobsd/legacy.sh
@@ -2,6 +2,7 @@
 #
 # Copyright (c) 2005 Poul-Henning Kamp All rights reserved.
 # Copyright (c) 2016 M. Warner Losh <imp@FreeBSD.org>
+# Copyright (c) 2025 Karl Denninger <karl@denninger.net>
 #
 #
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


bsdimp (approved):
This does break really old CHS systems, but I'm pretty sure the kernel hasn't got those right since around FreeBSD 9 or so.

jlduran:
@bsdimp I'm thinking about testing this series of patches. I like the overall idea, but I would prefer an approach that can be easily reconciled with poudriere image -t firmware.
Initially my plan was to do it at EuroBSDcon, but if things cannot wait that long, please proceed.
Thank you!

bsdimp:
@jlduran so what's your current state of play?

jlduran:
> @jlduran so what's your current state of play?

I'm currently in the rabbit hole step 2:

1. mtree (almost done, 1 revert/doc missing)
2. makefs `<- I am here`
3. mkimg
4. NanoBSD `NO_ROOT` (already approved)

Regarding this particular Pull Request, obviously I like the addition of an EFI partition, what I don't like is:

1. The current default size (40MB), that does not match the current 260MB default:
https://github.com/freebsd/freebsd-src/blob/3ae56f659aad8621a9d26b5ad0caeeaf1092c4ad/usr.sbin/bsdinstall/partedit/partedit_x86.c#L36

2. I would prefer to switch to GPT once we create an EFI partition.
3. The other inline observations.

If you prefer to commit this Pull Request, by all means, please do. If further down the line we need to make some adjustments, we'll do.
I think I'll also go ahead and commit all approved NanoBSD reviews, it's better if they're committed, so people can start testing it. At the moment it should not break current setups.

Thank you!

PR #1719, opened 2025-06-12
nanobsd: Update the updatep1 script to edit the EFI device to boot from

tickerguy:
Part of the nanobsd legacy.sh change. The only fly in the ointment is that its not possible to reliably know what disk number the BIOS EFI loader "saw"; I've seen USB media come up as disk1 when there was another device (non-bootable) in the machine, yet if you look at rootdev or similar after it comes up it shows as disk0.  If this is wrong in loader.env it won't boot at all, where if the file is missing then it will boot (but not your updated code slice!)

This requires more investigation in the case where other disk media is in the unit; if you have only the boot media in then it works as expected.

### DIFF
--- a/tools/tools/nanobsd/Files/root/updatep1
+++ b/tools/tools/nanobsd/Files/root/updatep1
@@ -59,7 +59,7 @@ EFI=`gpart show ${NANO_DRIVE}|grep 'efi'`
 if [ $EFI != '' ] ;
 then
 	PART=`echo $EFI|awk '{ print $3 }'`
-	mount -t msdosfs /dev/${NANO_DRIVE}s$EFI /mnt
+	mount -t msdosfs /dev/${NANO_DRIVE}s$PART /mnt
 	echo 'rootdev=disk0s1a' >/mnt/EFI/FreeBSD/loader.env
 	umount /mnt
 fi


bsdimp (approved):

tools/tools/nanobsd/Files/root/updatep1
```diff
@@ -51,3 +52,14 @@ dd of=/dev/${NANO_DRIVE}s1 obs=64k
 fsck_ffs -n /dev/${NANO_DRIVE}s1a
 
 gpart set -a active -i 1 ${NANO_DRIVE}
+#
+# KD - If the partition has an EFI partition then set "loader.env"
```
bsdimp: Noticed afer I approved...
But 'KD - ' should be deleted. We don't tag our comments with initials...

tools/tools/nanobsd/Files/root/updatep1
```diff
@@ -51,3 +52,14 @@ dd of=/dev/${NANO_DRIVE}s1 obs=64k
...
+#
+EFI=`gpart show ${NANO_DRIVE}|grep 'efi'`
```
bsdimp, line 58: since we're running on the system in question, you can look at `sysctl -n machdep.bootmethod` to know if we've booted with UEFI.

tools/tools/nanobsd/Files/root/updatep1
```diff
@@ -51,3 +52,14 @@ dd of=/dev/${NANO_DRIVE}s1 obs=64k
...
+if [ $EFI != '' ] ;
+then
+	PART=`echo $EFI|awk '{ print $3 }'`
+	mount -t msdosfs /dev/${NANO_DRIVE}s$EFI /mnt
+	echo 'rootdev=disk0s1a' >/mnt/EFI/FreeBSD/loader.env
```
bsdimp, line 63: Also, is there a reason we've not switched entirely GPT partitions? Or is this just a small, incremental change?

tools/tools/nanobsd/Files/root/updatep1
```diff
@@ -51,3 +52,14 @@ dd of=/dev/${NANO_DRIVE}s1 obs=64k
...
 
 gpart set -a active -i 1 ${NANO_DRIVE}
+#
+# KD - If the partition has an EFI partition then set "loader.env"
+#
+EFI=`gpart show ${NANO_DRIVE}|grep 'efi'`
+if [ $EFI != '' ] ;
+then
+	PART=`echo $EFI|awk '{ print $3 }'`
+	mount -t msdosfs /dev/${NANO_DRIVE}s$EFI /mnt
```
bsdimp: and shouldn' t this be s$PART not s$EFI?

bsdimp (changes requested):
Some minor things that need cleanup

tools/tools/nanobsd/Files/root/updatep1
tickerguy: No worries; want me to do that or just do it?

tools/tools/nanobsd/Files/root/updatep1
tickerguy: I've got older devices that barf on the idea of booting from a USB stick that is formatted in GPT; they're old but they still work so my goal was to find an answer that would boot both on those and on devices with no CSM boot capability at all (a LOT of 'small form factor' multi-interface things you can find today that are nicely suitable for firewalls and such are in this group; they have no compatibility mode option in their BIOS)

tools/tools/nanobsd/Files/root/updatep1
tickerguy: True but I need the partition anyway and if its not on the drive for some reason (the /etc/nanobsd.conf file is wrong) it won't work, or if you use the media in two boxes, and the one you're on right now isn't EFI, it still should set it correctly.  That's actually relevant for me in that I have a higher-performance EFI-boot-only firewall but if it blows up my older pcEngines box will work (at reduced performance) and being able to yank the stick out of one and shove it in the other and have it come up and run is nice.

tools/tools/nanobsd/Files/root/updatep1
bsdimp: fair

tools/tools/nanobsd/Files/root/updatep1
bsdimp: So. I'm not quite sure what you're saying.  The device that we booted from can be found with
```
sudo efivar -N --utf8 cfee69ad-a0de-47a9-93a8-f63106f8ae99-LoaderDev
PciRoot(0x1)/Pci(0x1,0x1)/Pci(0x0,0x0)/NVMe(0x1,00-00-00-00-00-00-00-00)/HD(1,GPT,B05DF68B-625D-11EB-81AA-E0D55E1E73BD,0x28,0x96000)
```
and you can use the efimedia from geom list to find the actual partition that booted... That will be robust, but only for EFI booted system.

But if you want to always update the EFI partition, in a constrained environment, then this will work...

tools/tools/nanobsd/Files/root/updatep1
bsdimp: Yes.  You also need to fix the commit messages on these pull requests to comply with
https://docs.freebsd.org/en/articles/committers-guide/#commit-log-message

tools/tools/nanobsd/Files/root/updatep1
tickerguy: No, that does not work -- here's what it shows on a running machine that I have this on:
root@IpGw:/data/karl # efivar -N --utf8 cfee69ad-a0de-47a9-93a8-f63106f8ae99-LoaderDev
PciRoot(0x0)/Pci(0xD,0x0)/USB(0x2,0x0)/HD(3,MBR,0xA8A8A8A8,0x1578D77,0x14000)

Problem: Its actually disk1!  The reason is that there is a non-bootable nVME drive in there (currently has a GPT partition table on it, but no bootcode, no EFI partition, etc.) and the EFI BIOS finds it first.  If I stick the loader.env file in /EFI/FreeBSD then the EFI loader has a fit because there's nothing bootable on disk0.  I've spent a fair bit of time rooting around in the sysctl and efi data and haven't been able to find a reliable way to know in what order the BIOS enumerated the devices.  Gpart returns them in the same order but that's unhelpful because there's no device number that works (e.g. nVME comes up as ndax .vs. dax, etc.) and I'm a long way from certain that its enumeration order will match that of the BIOS.

For many people this won't matter since there won't be any second device in there, but if someone wants to build, for example, a firewall that is *also* a local storage repository for inside things then it does because the power protection of "boot from stick" is near-absolute, where putting the image on the inside drive exposes you to risk, particularly with an SSD, even if the filesystems in question are mounted read-only.

What would be **nice** would be to have a notation in the loader.env file that is denotes the slice or partition _but uses the current device from which the EFI loader was executed_; I'm unaware of a syntax that works for that or if its even reasonable to implement in the FreeBSD loader.efi code.  It would solve that problem decisively in that then you can change which partition you boot from but it will always be the device from which the loader was read which, in an environment like this (presuming you have the EFI BIOS set to boot the correct device if more than one is bootable) will always be correct.

tools/tools/nanobsd/Files/root/updatep1
tickerguy: I believe that (in the last edit) now is and also fixed the typo ($PART where was erroneously $EFI)

PR #1720, opened 2025-06-12
sockstat: Add automatic column sizing and remove -w option

daminrido139:
Refactor sockstat to dynamically size table columns based on content. This change eliminates the need for the -w option, which is now ignored for backwards compatibility. Column widths are calculated per flag, improving accuracy and overall performance.

Sponsored by: Google, LLC (GSoC 2025)
MFC after: 2 weeks

### DIFF
--- a/usr.bin/sockstat/sockstat.c
+++ b/usr.bin/sockstat/sockstat.c
@@ -193,20 +193,6 @@ static cap_channel_t *capnetdb;
 static cap_channel_t *capsysctl;
 static cap_channel_t *cappwd;
 
-static int
-xprintf(const char *fmt, ...)
-{
-	va_list ap;
-	int len;
-
-	va_start(ap, fmt);
-	len = vprintf(fmt, ap);
-	va_end(ap);
-	if (len < 0)
-		err(1, "printf()");
-	return (len);
-}
-
 static bool
 _check_ksize(size_t received_size, size_t expected_size, const char *struct_name)
 {
@@ -941,7 +927,7 @@ getfiles(void)
 }
 
 static int
-printaddr(struct sockaddr_storage *ss)
+formataddr(struct sockaddr_storage *ss, char *buf, size_t bufsize)
 {
 	struct sockaddr_un *sun;
 	char addrstr[NI_MAXHOST] = { '\0', '\0' };
@@ -961,7 +947,8 @@ printaddr(struct sockaddr_storage *ss)
 	case AF_UNIX:
 		sun = sstosun(ss);
 		off = (int)((char *)&sun->sun_path - (char *)sun);
-		return (xprintf("%.*s", sun->sun_len - off, sun->sun_path));
+		return snprintf(buf, bufsize, "%.*s",
+				sun->sun_len - off, sun->sun_path);
 	}
 	if (addrstr[0] == '\0') {
 		error = cap_getnameinfo(capnet, sstosa(ss), ss->ss_len,
@@ -970,9 +957,8 @@ printaddr(struct sockaddr_storage *ss)
 			errx(1, "cap_getnameinfo()");
 	}
 	if (port == 0)
-		return xprintf("%s:*", addrstr);
-	else
-		return xprintf("%s:%d", addrstr, port);
+		return snprintf(buf, bufsize, "%s:*", addrstr);
+	return snprintf(buf, bufsize, "%s:%d", addrstr, port);
 }
 
 static const char *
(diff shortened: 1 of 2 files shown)


asomers:
The automatic column sizing works pretty well.  But it isn't perfect.  On my machine, with almost any options, sockets whose protocol is "stream (not connected)" will have misaligned columns.  This is easiest to see with the "-l" option.  It looks like the "(not connected)" part isn't contributing to the calculation of the width of the PROTO column.  For example:
```
USER   COMMAND    PID   FD PROTO  LOCAL ADDRESS                FOREIGN ADDRESS
somers sshd-sessi 56573 7  stream (not connected)              ??                           ??             
root   sshd-sessi 56571 7  stream (not connected)              ??                           ??             
somers sshd-sessi 4084  7  stream (not connected)              ??                           ??             
root   sshd-sessi 4082  7  stream (not connected)              ??                           ??             
root   sshd       3974  7  tcp6   *:22                         *:*            
root   sshd       3974  8  tcp4   *:22                         *:*            
root   syslogd    2679  6  udp6   *:514                        *:*            
root   syslogd    2679  7  udp4   *:514                        *:*            
root   syslogd    2679  8  dgram  /var/run/log                  <-            
root   syslogd    2679  10 dgram  /var/run/logpriv             ??             
root   nfsuserd   2080  3  udp6   *:633                        *:*            
root   nfsuserd   2079  3  udp6   *:633                        *:*            
root   nfsuserd   2078  3  udp6   *:633                        *:*            
root   nfsuserd   2077  3  udp6   *:633                        *:*            
root   nfsuserd   2076  3  udp6   *:633                        *:*            
root   devd       1593  4  stream /var/run/devd.pipe           ??             
root   devd       1593  5  seqpac /var/run/devd.seqpacket.pipe ??             
??     ??         ??    ?? stream (not connected)              ??                           ??             
```
Also, there's another change that wasn't mentioned in the commit message.  Is this deliberate?
* Fields that were previously "" are now printed as "??".
* Fields that were previously "?" for the USER, COMMAND, PID, and FD columns are now "??".  Such sockets are most easily seen if NFS is in use.

usr.bin/sockstat/sockstat.c
```diff
@@ -1108,234 +1094,428 @@ sctp_path_state(int state)
...
+	while (laddr != NULL || faddr != NULL) {
+		if (s->family == AF_UNIX) {
+			if ((laddr == NULL) || (faddr == NULL))
+				errx(1, "laddr = %p or faddr = %p is NULL",
+					(void *)laddr, (void *)faddr);
+			if (laddr->address.ss_len == 0 && faddr->conn == 0)
+				printf(" %-*s", cw->local_addr, "(not connected)");
+			strcpy(buf,"??");
+			if (laddr->address.ss_len > 0)
+				formataddr(&laddr->address, buf, sizeof(buf));
```
asomers: This line will overwrite the "??" from line 1399.  If you meant for line 1399 to be more like a default initializer, I suggest you move it below, with `else`.  There are similar patterns elsewhere, like on line 1407-1409.

daminrido139:
Thanks, @asomers,

You're absolutely right, `"stream (not connected)"` should be fully inside the PROTO column. I initially misunderstood and treated `(not connected)` as part of the LOCAL ADDRESS, but I’ll fix that and ensure the width calculation includes the full string.

Regarding the `"??"` change - yes, that was intentional. I wanted to make the output more uniform by using `"??"` for all unknown fields instead of mixing `""` and `"?"`. Sorry for not mentioning that in the commit message. happy to amend the commit if you prefer.

asomers:
> Thanks, @asomers,
> 
> You're absolutely right, `"stream (not connected)"` should be fully inside the PROTO column. I initially misunderstood and treated `(not connected)` as part of the LOCAL ADDRESS, but I’ll fix that and ensure the width calculation includes the full string.
> 
> Regarding the `"??"` change - yes, that was intentional. I wanted to make the output more uniform by using `"??"` for all unknown fields instead of mixing `""` and `"?"`. Sorry for not mentioning that in the commit message. happy to amend the commit if you prefer.

Using "??" uniformly is ok.  The only risk is that it upsets some scripts that people use to parse sockstat's output with awk.  But those scripts are fragile anyway.  Using libxo would be much more reliable!

usr.bin/sockstat/sockstat.c
```diff
@@ -1344,33 +1524,38 @@ display(void)
 	struct passwd *pwd;
 	struct file *xf;
 	struct sock *s;
-	int n, pos;
+	int n;
+    struct col_widths cw;
```
asomers: Some spaces snuck in on these two lines.  You should consistently indent with tabs.

usr.bin/sockstat/sockstat.c
```diff
@@ -1108,234 +1094,428 @@ sctp_path_state(int state)
...
+		if (faddr != NULL)
+			faddr = faddr->next;
+		first = false;
+	}
+}
+
+static void
+calculate_column_widths(struct col_widths *cw)
+{
+    cw->user         	= 4;
```
asomers: Also here you indented with spaces.  Please change them all to tabs.  See style(9) for a detailed description of how to format C code.

usr.bin/sockstat/sockstat.c
```diff
@@ -1108,234 +1094,428 @@ sctp_path_state(int state)
...
+    cw->fib 			= 3;
+    cw->splice_address	= 14;
+    cw->inp_gencnt		= 2;
+    cw->encaps			= 6;
+    cw->path_state 		= 10;
+    cw->conn_state		= 10;
+    cw->stack			= TCP_FUNCTION_NAME_LEN_MAX;
+    cw->cc				= TCP_CA_NAME_MAX;
+
+    char buf[512];
```
asomers: This is a lot of memory to put on the stack.  Better to use malloc().  But in this case, you don't have to!  It's possible to eliminate the need for buf entirely, but taking advantage of the fact that `snprintf` always returns the number of characters that it _would've_ written if the buffer had been big enough.  So just call `len = snprintf(NULL, 0, ...)` and remove all of the `strlen` lines.

usr.bin/sockstat/sockstat.c
```diff
@@ -1108,234 +1094,428 @@ sctp_path_state(int state)
 }
 
 static void
-displaysock(struct sock *s, int pos)
+format_unix_faddr(struct addr *faddr, char *buf, size_t bufsize) {
+	int pos = 0;
+	/* Remote peer we connect(2) to, if any. */
+	if (faddr->conn != 0) {
+		struct sock *p;
+		strcpy(buf, "-> ");
```
asomers: `strcpy` should almost always be avoided.  Use `strncpy`, or better yet `strlcpy`.

usr.bin/sockstat/sockstat.c
daminrido139: fixed!

usr.bin/sockstat/sockstat.c
daminrido139: fixed!

usr.bin/sockstat/sockstat.c
daminrido139: fixed! Thanks for the reference.

usr.bin/sockstat/sockstat.c
daminrido139: Removed buf and strlen wherever possible by using `len = snprintf(NULL, 0, ...) `. Used malloc() instead of stack memory.

usr.bin/sockstat/sockstat.c
daminrido139: Replaced `strcpy` with `strlcpy`.

asomers (commented):
Pro tips:
* Once a reviewer first looks at a PR, never force-push it.  If you must amend
  it, leave your changes in a separate commit.  That makes it easy for the
  reviewer to see what's changed since his first review.
* Never merge the main branch into your PR branch.  That can confuse the
  history and doesn't really solve any problems.  Instead, better to leave the
  PR out-of-date unless it conflicts with the main branch.  If you must resolve
  conflicts, it's better to rebase than to merge.

I'll come back tomorrow for a lengthier review.  I'll have to re-review the entire thing.

daminrido139:
Thanks for the advice, @asomers. I will make sure to follow this workflow and keep changes in separate commits after reviews begin.

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
 		if (faddr != NULL)
 			faddr = faddr->next;
-		if ((laddr != NULL) || (faddr != NULL)) {
-			xprintf("\n");
-			pos = 0;
-		}
-		first = 0;
+		if (laddr != NULL || faddr != NULL)
+			printf("%-*s %-*s %-*s %-*s %-*s",
+				cw->user,		"",
```
asomers: Why use tabs before the "" ?  You can use space instead.  Tabs are only necessary for the beginning of the line.  Also, make sure you set your tabstop to 8 in your editor.  I suspect you're using 4.

usr.bin/sockstat/sockstat.c
```diff
@@ -1484,7 +1661,7 @@ static void
 usage(void)
 {
 	errx(1,
-    "usage: sockstat [-46ACcfIiLlnqSsUuvw] [-j jid] [-p ports] [-P protocols]");
+	"usage: sockstat [-46ACcfIiLlnqSsUuvw] [-j jid] [-p ports] [-P protocols]");
```
asomers: This line runs over 80 characters.  It should be split up.  Or, just revert your change to this line.  It was previously indented with 4 spaces instead of 1 tab probably just for that reason.  See 051a2132f41.

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
+								buf, bufsize);
+			}
+			else
+				strlcpy(buf, "??", bufsize);
+			printf(" %-*s", cw->splice_address, buf);
+		}
+		if (opt_i) {
+			if (s->proto == IPPROTO_TCP ||
+				s->proto == IPPROTO_UDP) {
+				printf(" %*" PRIu64, cw->inp_gencnt ,s->inp_gencnt);
```
asomers: Style: line longer than 80 chars

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
+				if (sp != NULL)
+					formataddr(&sp->laddr->address,
+								buf, bufsize);
+			}
+			else
+				strlcpy(buf, "??", bufsize);
+			printf(" %-*s", cw->splice_address, buf);
+		}
+		if (opt_i) {
+			if (s->proto == IPPROTO_TCP ||
```
asomers: These two lines would be longer than 80 chars if you joined them.  However, it's acceptable to join them and then move the `{` to the next line.  That's more readable, IMHO.

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
+				s->proto == IPPROTO_UDP) {
+				printf(" %*" PRIu64, cw->inp_gencnt ,s->inp_gencnt);
+			}
+			else
```
asomers: style: cuddle the `else`
```suggestion
			} else
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
+		if (opt_I) {
+			if (s->splice_socket != 0) {
+				struct sock *sp;
+				sp = RB_FIND(socks_t, &socks, &(struct sock)
+					{ .socket = s->splice_socket });
+				if (sp != NULL)
+					formataddr(&sp->laddr->address,
+								buf, bufsize);
+			}
+			else
```
asomers: style: cuddle the else
```suggestion
			} else
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
+				((s->proto == IPPROTO_SCTP &&
+					s->state != SCTP_CLOSED &&
+					s->state != SCTP_BOUND &&
+					s->state != SCTP_LISTEN) ||
+					(s->proto == IPPROTO_TCP &&
+					s->state != TCPS_CLOSED &&
+					s->state != TCPS_LISTEN))) {
+				printf(" %*u", cw->encaps, ntohs(faddr->encaps_port));
+			}
+			else
```
asomers:
```suggestion
			} else
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
+		if (opt_s) {
+			if (faddr != NULL &&
+				s->proto == IPPROTO_SCTP &&
+				s->state != SCTP_CLOSED &&
+				s->state != SCTP_BOUND &&
+				s->state != SCTP_LISTEN) {
+				printf(" %-*s", cw->path_state,
+					sctp_path_state(faddr->state));
+			}
+			else
```
asomers:
```suggestion
			} else
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
+							printf(" %-*s", cw->conn_state,
+								tcpstates[s->state]);
 						else
-							pos += xprintf("?");
+							printf(" %-*s", cw->conn_state, "??");
 						break;
 					}
 				}
-				offset += 13;
+				else
```
asomers: style: cuddle the `else`

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
-					pos += xprintf("%.*s",
-					    TCP_FUNCTION_NAME_LEN_MAX,
-					    s->stack);
-				}
-				offset += TCP_FUNCTION_NAME_LEN_MAX + 1;
+				if (s->proto == IPPROTO_TCP)
+					printf(" %-*.*s", TCP_FUNCTION_NAME_LEN_MAX,
+						TCP_FUNCTION_NAME_LEN_MAX, s->stack);
+				else
+					printf(" %-*s", cw->stack, "??");
```
asomers: For both brevity and consistency
```suggestion
					printf(" %-*.*s", cw->stack, cw->stack, s->stack);
				else
					printf(" %-*s", cw->stack, "??");
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
-				if (s->proto == IPPROTO_TCP) {
-					do
-						pos += xprintf(" ");
-					while (pos < offset);
-					xprintf("%.*s", TCP_CA_NAME_MAX, s->cc);
-				}
-				offset += TCP_CA_NAME_MAX + 1;
+				if (s->proto == IPPROTO_TCP)
+					printf(" %-*.*s", TCP_CA_NAME_MAX,
+						TCP_CA_NAME_MAX, s->cc);
```
asomers:
```suggestion
					printf(" %-*.*s", cw->cc, cw->cc, s->cc);
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
+			printf("%-*s %-*s %-*s %-*s %-*s",
+				cw->user,		"",
+				cw->command,	"",
+				cw->pid,		"",
+				cw->fd,			"",
+				cw->proto,		"");
+		first = false;
 	}
-	xprintf("\n");
+	printf("\n");
```
asomers, line 1509: Pro tip: using `printf` repeatedly is inefficient, because every call results in a `write(2)` syscall.  Using `fprintf` is much more efficient if you plan to call it over and over, because it buffers the output in the C library, and only calls `write(2)` when the buffer is full or you intentionally flush it.  HOWEVER, don't make that change now, because in a future commit you'll replace all of these with `xo_emit` instead.

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
 }
 
+static int
+format_unix_faddr(struct addr *faddr, char *buf, size_t bufsize) {
+	size_t pos = 0;
+	/* Remote peer we connect(2) to, if any. */
+	if (faddr->conn != 0) {
+		struct sock *p;
+		strlcpy(buf, "-> ", bufsize);
+		pos += 3;
```
asomers: Better to rely on strlcpy returning the right value than to hard-code it.
```suggestion
		pos += strlcpy(buf, "-> ", bufsize);
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
+				buf == NULL ? 0 : bufsize-pos, "??");
+		} else if (p->laddr->address.ss_len == 0) {
+			struct file *f;
+			f = RB_FIND(files_t, &ftree,
+				&(struct file){ .xf_data =
+				p->socket });
+			if (f != NULL) {
+				pos += snprintf(buf == NULL ? NULL : buf+pos,
+					buf == NULL ? 0 : bufsize-pos,
+					"[%lu %d]", (u_long)f->xf_pid, f->xf_fd);
```
asomers: Instead of trying to keep track of `pos`, you could print lines like this into a stack buffer, then use `strlcat` to copy that to `buf`.  Which technique you prefer is your choice.

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
+	first = true;
+
+	len = strlen(s->protoname);
+	if (s->vflag & (INP_IPV4 | INP_IPV6))
+		len += 1;
+	if (laddr != NULL && faddr != NULL && s->family == AF_UNIX &&
+		laddr->address.ss_len == 0 && faddr->conn == 0)
+		len += 16;
+	if (len > cw->proto)
+		cw->proto = len;
```
asomers:
```suggestion
	cw->proto = MAX(cw->proto, len);
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
 	faddr = s->faddr;
-	first = 1;
+	first = true;
+
+	len = strlen(s->protoname);
+	if (s->vflag & (INP_IPV4 | INP_IPV6))
+		len += 1;
+	if (laddr != NULL && faddr != NULL && s->family == AF_UNIX &&
+		laddr->address.ss_len == 0 && faddr->conn == 0)
+		len += 16;
```
asomers: Where does the magic number 16 come from?

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
+			if (len > cw->local_addr)
+				cw->local_addr = len;
+			len = format_unix_faddr(faddr, NULL, 0);
+			if (len > cw->foreign_addr)
+				cw->foreign_addr = len;
+		} else {
+			if (laddr != NULL) {
+				len = formataddr(&laddr->address, NULL, 0);
+				if (len > cw->local_addr)
+					cw->local_addr = len;
```
asomers:
```suggestion
				cw->local_addr = MAX(cw->local_addr, len);
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
-						    (u_long)f->xf_pid,
-						    f->xf_fd);
-					}
-					ref = p->faddr->nextref;
-					fref = false;
-				}
+			if (faddr != NULL) {
+				len = formataddr(&faddr->address, NULL, 0);
+				if (len > cw->foreign_addr)
+					cw->foreign_addr = len;
```
asomers:
```suggestion
				cw->foreign_addr = MAX(cw->foreign_addr, len);
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
-		if (opt_A) {
-			pos += xprintf("0x%16lx", s->pcb);
-			offset += 18;
 		}
 		if (opt_f) {
-			pos += xprintf("%d", s->fibnum);
-			offset += 7;
+			len = snprintf(NULL, 0, "%d", s->fibnum);
+			if (len > cw->fib)
+				cw->fib = len;
```
asomers:
```suggestion
			cw->fib = MAX(cw->fib, len);
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
-					pos += printaddr(&sp->laddr->address);
-				} else {
-					do
-						pos += xprintf(" ");
-					while (pos < offset);
-					pos += xprintf("??");
-					offset += opt_w ? 46 : 22;
+					len = formataddr(&sp->laddr->address, NULL, 0);
+					if (len > cw->splice_address)
+						cw->splice_address = len;
```
asomers:
```suggestion
					cw->splice_address = MAX(cw->splice_addres, len);
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
-			    s->proto == IPPROTO_UDP) {
-				do
-					pos += xprintf(" ");
-				while (pos < offset);
-				pos += xprintf("%" PRIu64, s->inp_gencnt);
+				s->proto == IPPROTO_UDP) {
+				len = snprintf(NULL, 0,
+					"%" PRIu64, s->inp_gencnt);
+				if (len > cw->inp_gencnt)
+					cw->inp_gencnt = len;
```
asomers:
```suggestion
				cw->inp_gencnt = MAX(cw->inp_gencnt, len);
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
+					s->state != SCTP_CLOSED &&
+					s->state != SCTP_BOUND &&
+					s->state != SCTP_LISTEN) ||
+					(s->proto == IPPROTO_TCP &&
+					s->state != TCPS_CLOSED &&
+					s->state != TCPS_LISTEN))) {
+				len = snprintf(NULL, 0, "%u",
+					ntohs(faddr->encaps_port));
+				if (len > cw->encaps)
+					cw->encaps = len;
```
asomers:
```suggestion
				cw->encaps = MAX(cw->encaps, len);
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
-				pos += xprintf("%s",
-				    sctp_path_state(faddr->state));
+				s->proto == IPPROTO_SCTP &&
+				s->state != SCTP_CLOSED &&
+				s->state != SCTP_BOUND &&
+				s->state != SCTP_LISTEN) {
+				len = snprintf(NULL, 0, "%s",
+					sctp_path_state(faddr->state));
+				if (len > cw->path_state)
+					cw->path_state = len;
```
asomers:
```suggestion
				cw->path_state = MAX(cw->path_state, len);
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
-					while (pos < offset);
+					s->proto == IPPROTO_TCP) {
 					switch (s->proto) {
 					case IPPROTO_SCTP:
-						pos += xprintf("%s",
-						    sctp_conn_state(s->state));
+						len = snprintf(NULL, 0, "%s",
+							sctp_conn_state(s->state));
+						if (len > cw->conn_state)
+							cw->conn_state = len;
```
asomers:
```suggestion
						cw->conn_state = MAX(cw->conn_state, len);
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
 					case IPPROTO_TCP:
 						if (s->state >= 0 &&
-						    s->state < TCP_NSTATES)
-							pos += xprintf("%s",
-							    tcpstates[s->state]);
+							s->state < TCP_NSTATES) {
+							len = snprintf(NULL, 0, "%s",
+								tcpstates[s->state]);
+							if (len > cw->conn_state)
+								cw->conn_state = len;
```
asomers:
```suggestion
							cw->conn_state = MAX(cw->conn_state, len);
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
+	cw->foreign_addr	= 15;
+	cw->pcb_kva		= 18;
+	cw->fib			= 3;
+	cw->splice_address	= 14;
+	cw->inp_gencnt	= 2;
+	cw->encaps		= 6;
+	cw->path_state	= 10;
+	cw->conn_state	= 10;
+	cw->stack		= TCP_FUNCTION_NAME_LEN_MAX;
+	cw->cc			= TCP_CA_NAME_MAX;
```
asomers: I suggest not trying to align the parts after the `=`.  But if you do, make sure you use spaces, or use tabstop=8.  As it is, they aren't actually aligned.

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
+		if (s == NULL || (!check_ports(s)))
+			continue;
+		s->shown = 1;
+		if (opt_n || 
+			(pwd = cap_getpwuid(cappwd, xf->xf_uid)) == NULL)
+			len = snprintf(NULL, 0, "%lu", (u_long)xf->xf_uid);
+		else
+			len = snprintf(NULL, 0, "%s", pwd->pw_name);
+		if (len > cw->user)
+			cw->user = len;
```
asomers:
```suggestion
		cw->user = MAX(cw->user, len);
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
+		len = snprintf(NULL, 0, "%lu", (u_long)xf->xf_pid);
+		if (len > cw->pid)
+			cw->pid = len;
```
asomers:
```suggestion
		cw->pid = MAX(cw->pid, len);
```

usr.bin/sockstat/sockstat.c
```diff
@@ -1107,235 +1093,420 @@ sctp_path_state(int state)
...
+		len = snprintf(NULL, 0, "%d", xf->xf_fd);
+		if (len > cw->fd)
+			cw->fd = len;
```
asomers:
```suggestion
		cw->fd = MAX(cw->fd, len);
```

asomers (changes requested):
Numeric columns like "FIB" "ENCAPS", and "FD" are now right-aligned, whereas they used to be left-aligned.  Is that intentional?

The "STACK" column is far wider than it needs to be.
That's probably because you use TCP_FUNCTION_NAME_LEN_MAX instead of calculating how much space is actually needed.  But that's overkill.  The constant is 32, whereas the longest stackname currently defined only uses 7 characters.  The same problem applies to the "CC" column.

The "opt_w" variable is now write-only.  You can simply remove it, but of course getopt() should still accept it.

daminrido139:
Thanks a lot for the detailed review, @asomers !

1. The right-alignment of numeric columns was intentional for readability (similar to ps and netstat).
2. Great point on the STACK and CC column widths. I now compute them based on the actual observed lengths instead of using constant maximums.
3. Removed the opt_w variable while retaining it in the getopt() call.
4. Fixed the mid-line tab usage. my tabstop wasn’t set to 8 earlier; it is now corrected.
5. Reverted the usage() line to stay within the 80-character limit, as originally intended.
6. Applied all the suggested style changes: cuddled else, MAX() usage, line length limits, alignment fixes, etc.
7. Added a comment to clarify the use of the magic number 16.

As per your suggestion, I initially made the changes in a separate commit and pushed normally. However, I noticed a few unintended changes had crept in, so I had to amend and force-push to revert them. I apologize for that and will be more careful next time.

Please review the updated changes whenever you get a chance. Thanks again!

usr.bin/sockstat/sockstat.c
asomers: Better to literally use `len += strlen(" (not connected)");`  Magic numbers are to be avoided wherever possible.

usr.bin/sockstat/sockstat.c
```diff
@@ -1271,23 +1260,31 @@ calculate_sock_column_widths(struct col_widths *cw, struct sock *s)
...
 					case IPPROTO_TCP:
 						if (s->state >= 0 &&
-							s->state < TCP_NSTATES) {
-							len = snprintf(NULL, 0, "%s",
-								tcpstates[s->state]);
-							if (len > cw->conn_state)
-								cw->conn_state = len;
+						    s->state < TCP_NSTATES) {
+						    len = strlen(
+						    tcpstates[s->state]);
```
asomers: This line should be indented more than the one above.

usr.bin/sockstat/sockstat.c
```diff
@@ -1300,22 +1297,22 @@ calculate_sock_column_widths(struct col_widths *cw, struct sock *s)
...
-	cw->splice_address	= 14;
-	cw->inp_gencnt	= 2;
-	cw->encaps		= 6;
-	cw->path_state	= 10;
-	cw->conn_state	= 10;
-	cw->stack		= TCP_FUNCTION_NAME_LEN_MAX;
-	cw->cc			= TCP_CA_NAME_MAX;
+	cw->user = 4;
+	cw->command = 10;
+	cw->pid	= 3;
```
asomers: Here and on the "fib" line there are still tabs before `=`.  Better to make them all spaces.

usr.bin/sockstat/sockstat.c
```diff
@@ -1498,12 +1488,9 @@ display_sock(struct sock *s, struct col_widths *cw, char *buf, size_t bufsize)
...
 			faddr = faddr->next;
 		if (laddr != NULL || faddr != NULL)
-			printf("%-*s %-*s %-*s %-*s %-*s",
-				cw->user,		"",
-				cw->command,	"",
-				cw->pid,		"",
-				cw->fd,			"",
-				cw->proto,		"");
+			printf("%-*s %-*s %-*s %-*s %-*s", cw->user, "",
+				cw->command, "", cw->pid, "", cw->fd,	"",
```
asomers: Likewise, there's still a tab character on this line, before the final pair of quotes.

asomers (changes requested):
This is looking pretty good now!  I just have a few more comments regarding style.  Also, please ensure that every user-visible change is mentioned in the commit message.  That includes changing the alignment of numeric fields.

daminrido139:
Hi @asomers ,
I’ve fixed those style issues and included the alignment change of the numeric fields in the commit message. Thank you for your valuable advice! Kindly review the changes when you have time and let me know the next steps.

asomers (changes requested):
Just one more small change: your commit message should include this line.  Other than that, it LGTM.

```
Pull Request:   https://github.com/freebsd/freebsd-src/pull/1720
```

daminrido139:
@asomers done!!

asomers (approved):
Done.
At this point, if you were a committer, you should:

1. Squash your commits
2. Rebase on main
3. Force push to update the PR
4. push to main
5. If you get a conflict pushing to main, return to step 2.

However, since you aren't a committer, I will do those things for you.  And you can get back to work on libxo support.  Thanks for your contribution!

asomers (changes requested):
Testing revealed one more problem.  Your changes don't compile on i386 due to printf format specifier details.  It's a very common mistake.  It's surprising that they don't, because you used the exact same printf format specifier as the old code.  However, you switched from using vprintf to printf, which enables more compile-time checking.
```
/usr/home/somers/src/freebsd.org/src/usr.bin/sockstat/sockstat.c:1402:23: error: format specifies type 'unsigned long' but the argument has type 'kvaddr_t' (aka 'unsigned long long') [-Werror,-Wformat]
 1402 |                         printf(" 0x%16lx", s->pcb);
      |                                    ~~~~~   ^~~~~~
      |                                    %16llx
```
To reproduce, you can do `make TARGET_ARCH=i386 -j(nproc) buildworld` from the top of the git tree.

usr.bin/sockstat/sockstat.c
```diff
@@ -1384,7 +1384,8 @@ display_sock(struct sock *s, struct col_widths *cw, char *buf, size_t bufsize)
 			else
 				strlcpy(buf, "??", bufsize);
 			printf(" %-*s", cw->local_addr, buf);
-			format_unix_faddr(faddr, buf, bufsize);
+			if (format_unix_faddr(faddr, buf, bufsize) == 0)
+				strlcpy(buf, "??", bufsize);
```
daminrido139, line 1388: The above 2 lines ensures that the foreign address is set to '??' when no foreign address exists.

daminrido139:
Hi @asomers ,
Thanks for pointing that out, I didn’t expect it either. I’ve updated the format specifier to ensure it compiles correctly on i386. Let me know if there’s anything else I should adjust!

usr.bin/sockstat/sockstat.c
```diff
@@ -1399,7 +1400,8 @@ display_sock(struct sock *s, struct col_widths *cw, char *buf, size_t bufsize)
 			printf(" %-*s", cw->foreign_addr, buf);
 		}
 		if (opt_A)
-			printf(" 0x%16lx", s->pcb);
+			printf(" 0x%*lx", cw->pcb_kva,
+				(u_long)(uintptr_t)s->pcb);
```
asomers: You might consider doing it this way instead.  There is already precedent for using PRIu64 in this file.  Also, though I know you didn't introduce the "0x", it's more idiomatically represented using "#".
```suggestion
			printf(" %#*" PRIx64, cw->pcb_kva, s->pcb);
```

asomers (commented):
I'll approve it as-is, or with my suggestion.  Just let me know which you prefer.

Also, I just noticed that "-A" is undocumented.  We should probably fix that.  Not for this PR, though.

usr.bin/sockstat/sockstat.c
daminrido139: Yes, I noticed the use of the PRIu64 macro earlier. I went with your suggestion..  it's cleaner and more consistent now.

asomers:
@daminrido139 in order to close this PR automatically when I push your changes to the main branch, I first need permission to push to your gsock-sockstat-libxo branch.  Can you please grant it?  I think you need to check the box that says "allow edits and access to secrets by maintainers".

daminrido139:
Hi @asomers ,
The "Allow edits by maintainers" checkbox is already ticked, so I’m not exactly sure what went wrong. I've sent you an invitation to be a collaborator on my fork, maybe that will help. Could you please check your GitHub inbox?

asomers:
> Hi @asomers , The "Allow edits by maintainers" checkbox is already ticked, so I’m not exactly sure what went wrong. I've sent you an invitation to be a collaborator on my fork, maybe that will help. Could you please check your GitHub inbox?

That helped.  "Allow edits by maintainers" might not be working because my github account technically doesn't have permission to push to the freebsd-src repo on github.  This repo is really just a read-only mirror of the official FreeBSD git repo.

PR #1721, opened 2025-06-12
LoongArch: Clang, lld add support

yushanwei:
LoongArch: Clang, lld add support


bsdimp:
@DimitryAndric hope it's OK that I tagged you for this. It seems good to my eye, but I'm no expert. Also, hoped you could opine on the best way to get this into the tree... Do we need to commit it first to get it upstreamed? Or is upstreaming easier?

DimitryAndric:
Yes, this needs to be submitted to https://github.com/llvm/llvm-project first. If it is integrated into main, we can attempt backporting it. I would rather not apply such a big patch if it isn't vetted by upstream.

bsdimp:
Thanks @DimitryAndric , @yushanwei you can point to this to show FreeBSD wants this feature in llvm, but would prefer it going into llvm first so we don't have to carry it. I understand there can be resistance for including targets that FreeBSD doesn't support, so this should serve to show the FreeBSD project definitely has an interest.

Once we've got the changes upstreamed, tag me here and I'll coordinate with @DimitryAndric about how to get it into the tree.

yushanwei:
When we started to transplant FreeBSD to LoongArch, the llvm version was 17, which was the first version to support LoongArch. [ [clang][LoongArch] Add FreeBSD targets #119191 ](https://github.com/llvm/llvm-project/pull/119191) was done by the three of us.

haowuge:
> Thanks @DimitryAndric , @yushanwei you can point to this to show FreeBSD wants this feature in llvm, but would prefer it going into llvm first so we don't have to carry it. I understand there can be resistance for including targets that FreeBSD doesn't support, so this should serve to show the FreeBSD project definitely has an interest.

OK, we'll try to submit the patch upstream first.
Thank you.

haowuge:
> Yes, this needs to be submitted to https://github.com/llvm/llvm-project first. If it is integrated into main, we can attempt backporting it. I would rather not apply such a big patch if it isn't vetted by upstream.

Hi, @DimitryAndric
I have a question: if the llvm20 branch is merged in, can we update the relevant content in the FreeBSD base system?
Because I noticed that changes in the contrib/llvm-project/* part have already been made in 20.x .
Or in other words, which parts of the FreeBSD base system need to be updated by the upstream branch first?

thank you.😃

DimitryAndric:
> > Yes, this needs to be submitted to https://github.com/llvm/llvm-project first. If it is integrated into main, we can attempt backporting it. I would rather not apply such a big patch if it isn't vetted by upstream.
> 
> I have a question: if the llvm20 branch is merged in, can we update the relevant content in the FreeBSD base system? Because I noticed that changes in the contrib/llvm-project/* part have already been made in 20.x, thank you.😃

We will attempt to merge the patch into the 19.1.7 version we have in the FreeBSD base system. I don't think it should be too difficult.

haowuge:
> We will attempt to merge the patch into the 19.1.7 version we have in the FreeBSD base system. I don't think it should be too difficult.

You may want to pay attention to the fact that the upstream has already removed the loongarch32 code. 🤔

[[Driver][FreeBSD][NFC] Remove some code duplication](https://github.com/llvm/llvm-project/commit/bcd32ef0ba6b33a4aa3a5d0a96526f86634035da)
[[Driver][FreeBSD] Remove FreeBSD/loongarch32 support](https://github.com/llvm/llvm-project/pull/122515)

yushanwei:
[freebsd ports/devel/llvm19 backport it ](https://cgit.freebsd.org/ports/commit/devel/llvm19?id=983003a6153f1727cc46921197a8b0718f2b4e3e) on 2024-12-18

yushanwei:
waiting for llvm21

PR #1729, opened 2025-06-20
fusefs: only search for FREAD fufh in readdir

CismonX:
The extra search for an `FEXEC` fufh shall be removed, since readdir is only supposed to be called on a directory opened with `FREAD`.

Also remove the `fuse_filehandle_get_dir()` function, since it's not used anywhere else.

### DIFF
--- a/sys/fs/fuse/fuse_vnops.c
+++ b/sys/fs/fuse/fuse_vnops.c
@@ -263,16 +263,6 @@ fuse_extattr_check_cred(struct vnode *vp, int ns, struct ucred *cred,
 	}
 }
 
-/* Get a filehandle for a directory */
-static int
-fuse_filehandle_get_dir(struct vnode *vp, struct fuse_filehandle **fufhp,
-	struct ucred *cred, pid_t pid)
-{
-	if (fuse_filehandle_get(vp, FREAD, fufhp, cred, pid) == 0)
-		return 0;
-	return fuse_filehandle_get(vp, FEXEC, fufhp, cred, pid);
-}
-
 /* Send FUSE_FLUSH for this vnode */
 static int
 fuse_flush(struct vnode *vp, struct ucred *cred, pid_t pid, int fflag)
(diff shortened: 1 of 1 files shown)


CismonX:
More background on this patch:

The `fuse_filehandle_get_dir()` function was introduced in commit f8d4af104b7d.
This commit (unintentionally?) added an extra search for an `FEXEC` file handle
in `fuse_vnop_readdir()`, which is not a valid access mode for `getdirentries()`,
but should be harmless, since such syscalls fail with `EBADF` before `VOP_READDIR()`.

`fuse_filehandle_get_dir()` was also called in `fuse_vnop_close()`, however,
this usage was removed in commit 35cf0e7e56ca.
Thus it should be okay to remove the function completely, since there's no longer
any legitimate usage for getting a "`FREAD` or `FEXEC`" directory file handle.

CismonX:
@asomers

Please review this patch, thanks!

bsdimp:
@asomers can you look at this please?

asomers:
`kern_getdirentries` does indeed require that the file be opened with `FREAD`.  But that's not the only caller of `VOP_READDIR`.  NFS calls it.  So does autofs, and a few other places.  I don't think we can assume that `FREAD` will always be set in `fuse_vnop_readdir`.

CismonX:
> `kern_getdirentries` does indeed require that the file be opened with `FREAD`. But that's not the only caller of `VOP_READDIR`. NFS calls it. So does autofs, and a few other places. I don't think we can assume that `FREAD` will always be set in `fuse_vnop_readdir`.

- For those who call `VOP_READDIR()` without `VOP_OPEN()`, it fails, with or without this patch.
  * This includes autofs and a few others, notably `mount` with `MNT_EMPTYDIR`.
  * For NFS, there's already an implicit open in `fuse_vnop_readdir()`.
    Perhaps this could be applied beyond NFS? (provided that the server does not implement `FUSE_OPENDIR`)
- For (potential) callers who call `VOP_READDIR()` on a directory _explicitly_ opened with `FEXEC`,
  failing with `EBADF` should be the correct thing to do.
  * Actually, I'm yet to find an example for this case.
    The only remaining callers to `VOP_READDIR()` that may call it on a fusefs are unionfs and `vop_stdvptocnp()`.
    They both properly open the directory with `FREAD` before doing so.

bsdimp:
@asomers can you check this out again please?

asomers (approved):
Ok, you convinced me.  Sorry I took so long to review it.

asomers:
Committed as 4179f1d9deed83977f159c8afea204293ef4c7d7 .

PR #1735, opened 2025-06-22
bsdinstall(8): Fix GPT label conflicts with disks not managed by us

tuffnatty:
When doing a ZFS install on a machine having another installation made by bsdinstall(8), with GPT labels kept as `efiboot0`, `gptboot0`, `boot0`, `zfs0`, or `swap0`, bsdinstall(8) creates conflicting GPT labels and calls newfs_msdos(8) on existing `gpt/efiboot0` device from unmanaged disk.

This patch checks for potential GPT label conflicts and prevents them by adjusting the numeric index to an unused value.

### DIFF
--- a/usr.sbin/bsdinstall/scripts/zfsboot
+++ b/usr.sbin/bsdinstall/scripts/zfsboot
@@ -241,6 +241,7 @@ ZPOOL_SET='zpool set %s "%s"'
 hline_alnum_arrows_punc_tab_enter="Use alnum, arrows, punctuation, TAB or ENTER"
 hline_arrows_space_tab_enter="Use arrows, SPACE, TAB or ENTER"
 hline_arrows_tab_enter="Press arrows, TAB or ENTER"
+msg_all_possible_gpt_labels_already_taken="All possible GPT labels already taken"
 msg_an_unknown_error_occurred="An unknown error occurred"
 msg_back="Back"
 msg_cancel="Cancel"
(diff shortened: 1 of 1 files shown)


usr.sbin/bsdinstall/scripts/zfsboot
```diff
@@ -841,6 +841,18 @@ zfs_create_diskpart()
 
 	case "$ZFSBOOT_PARTITION_SCHEME" in
 	""|GPT*) f_dprintf "$funcname: Creating GPT layout..."
+		#
+		# 0. Check for potential GPT label conflicts with disks not
+		# managed by us and adjust $index accordingly
+		#
+		while [ -e /dev/gpt/efiboot$index ] ||
+			[ -e /dev/gpt/gptboot$index ] ||
```
bsdimp, line 850: This needs to likely be in the caller so that we catch it for UFS installs as well.
Also, there needs to be a limit so we don't loop forever.

tuffnatty: Thank you for the review!

AFAICT, the UFS installation code does not generate any GPT labels (other than the ones explicitly set by user) and does try to use them to refer to partitions (/dev/gpt/...) (where this bug surfaces). So here's the only place where the problem happens.

As for setting a limit for the loop, probably you can't have an indefinite number of GPT labels, so IMHO the loop will definitely finish. Still, for the sake of readability, a hard limit could be added. What about 4096? Googling the drive count limits on FreeBSD, I've found a report of a kernel panic with more than 4096 drives: https://forums.FreeBSD.org/threads/maximum-number-of-disks.87168/post-588524 , so we could use it as guidance ;-)

bsdimp: OK. That makes sense... I'll admit I didn't go look, but relied on my memory that the ufs path does this too.

4096 is likely a good limit. 10 likely is a good limit too, but you never know who might have huge 'imaging farms' or something.  There's no real hard limit, but I'd like some protection from the installer hanging for reasons that are opaque to the user.

bsdimp: OK. That makes sense... I'll admit I didn't go look, but relied on my memory that the ufs path does this too.

4096 is likely a good limit. 10 likely is a good limit too, but you never know who might have huge 'imaging farms' or something.  There's no real hard limit, but I'd like some protection from the installer hanging for reasons that are opaque to the user.

tuffnatty: Added a limit.

bsdimp:

PR #1737, opened 2025-06-26
sbin/md5/tests: avoid skipping missing tests

svmhdvn:
As seen here https://ci.freebsd.org/view/Test/job/FreeBSD-main-amd64-test/lastCompletedBuild/testReport/sbin.md5/ and the overall tests status here https://ci.freebsd.org/view/Test/job/FreeBSD-main-amd64-test/, this pollutes the test results with unnecessarily "skipped" tests. For functionality that doesn't exist, we should not create empty "skipped" tests.

Tested locally to ensure that all skipped tests are gone, and that all previously passing tests are passing now.


dag-erling:
I don't agree that skipped tests “pollute” the test results. What problem are you actually trying to solve?

svmhdvn:
Tests ideally should be skipped only when there are missing requirements, or when something in the test environment prevents it from working correctly. In this case, they are testing nothing at all since the functionality doesn't exist.

The problem is that the optics of a large number of "skipped tests" in a CI environment de-incentivizes people from looking into why they are skipped. It also gives the wrong impression that the larger number of skipped tests are reducing coverage of the codebase.

It doesn't matter too much in this case, but why add empty skips instead of just removing those empty test cases from the list?

svmhdvn:
Any feedback on my previous comment? I honestly think it's pretty harmless and will just make things cleaner for people viewing kyua test reports or https://ci.freebsd.org

svmhdvn:
Closing as unnecessary.

PR #1738, opened 2025-06-26
cap_dns/tests/dns_test: mark tests as needing network access

svmhdvn:
See https://ci.freebsd.org/job/FreeBSD-main-amd64-test/26672/testReport/lib.libcasper.services.cap_dns/dns_test/


markjdb:
In runtest() there is a check for failure which skips the test if we think the failure occurred for lack of internet access. Can we get rid of that check now?

svmhdvn:
> In runtest() there is a check for failure which skips the test if we think the failure occurred for lack of internet access. Can we get rid of that check now?

Thanks! Applied that change now.

PR #1740, opened 2025-06-27
libsa: Add isprint()

kpowkitty:
Add `isprint()` alongside other `isfoo()` functions in `libsa`.
Motivations for change:
1) Future `ACPICA` work in loader needs it.
2) `libsa/hexdump.c` already uses it.

### DIFF
--- a/stand/libsa/stand.h
+++ b/stand/libsa/stand.h
@@ -275,6 +275,11 @@ static __inline int ispunct(int c)
 	    (c >= '[' && c <= '`') || (c >= '{' && c <= '~');
 }
 
+static __inline int isprint(int c)
+{
+	return (c >= ' ') && (c <= '~');
+}
+
 static __inline int toupper(int c)
 {
     return islower(c) ? c - 'a' + 'A' : c;
--- a/stand/libsa/hexdump.c
+++ b/stand/libsa/hexdump.c
@@ -61,7 +61,7 @@ hexdump(caddr_t region, size_t len)
 	for (x = 0; x < 16; x++) {
 	    if ((line + x) < (region + len)) {
 		c = *(uint8_t *)(line + x);
-		if ((c < ' ') || (c > '~'))	/* !isprint(c) */
+		if (!isprint(c))
 		    c = '.';
 		emit("%c", c);
 	    } else {


bsdimp:
Things look good.

I'm kinda on the fence about including the following:
"libsa/hexdump.c uses it raw, will be replaced in a separate commit."
in the first commit message. since you're going to immediately follow it up with a separate commit, I'm not sure that saying so and then having that commit be the next one adds value. If there were a rework that was coming some weeks or months in the future, it would make sense though. Can you remove it? Or do you have a reason I'm not seeing to include it?

So the rest looks great.

stand/libsa/stand.h
```diff
@@ -275,6 +275,11 @@ static __inline int ispunct(int c)
 	    (c >= '[' && c <= '`') || (c >= '{' && c <= '~');
 }
 
+static __inline int isprint(int c)
+{
+	return ((c >= ' ') && (c <= '~'));
```
bsdimp: Minor nit. I know it's taught that it's a best practices to have the extra parens, all the other isfoo() functions don't use this style. It's a subtle part of our style guide that we try to match the existing style of the file. Now, it's a horrible mix of return foo; and return (foo);, but the latter is what style(9) proscribes, so that's another lesson too: when in doubt or it's ambiguous, favor style(9) when it offers an opinion on something.

kpowkitty: No, that makes complete sense---I knew this and got too focused on following style(9) general fundamentals that I forgot that tidbit. Thank you for reminding me, and now I know that for certain moving forward. Fixed.

concussious:
Does this need an update to libsa.3 (the manual page?). If so, I can help!

bsdimp:
> Does this need an update to libsa.3 (the manual page?). If so, I can help!

We do need it, but libsa.3 is also somewhat out of date and needs a lot of attention with someone that can read the code and see what's missing. It doesn't need to be with this commit, which I think is otherwise ready. Would you feel comfortable putting it into your list @concussious and looping me and @kpowkitty into the loop when you do? You can I can work on what's missing in libsa.4 as well, if you're wanting to really help a lot :).

bsdimp (approved):
OK. I think this is ready to go. I'll land the change soon.

bsdimp:

PR #1745, opened 2025-06-28
p9fs: MFC into stable/14

arrowd:
This PR contains an MFC of https://reviews.freebsd.org/D41844 with follow-up fixes to 14-STABLE. Most of the commits applied cleanly, but here are the changes that need attention:

1. I had to apply additional change
```
diff --git a/sys/dev/virtio/p9fs/virtio_p9fs.c b/sys/dev/virtio/p9fs/virtio_p9fs.c
index fb5b80258c1b..67426a88bceb 100644
--- a/sys/dev/virtio/p9fs/virtio_p9fs.c
+++ b/sys/dev/virtio/p9fs/virtio_p9fs.c
@@ -246,7 +246,7 @@ vt9p_alloc_virtqueue(struct vt9p_softc *sc)
            vt9p_intr_complete, sc, &sc->vt9p_vq,
            "%s request", device_get_nameunit(dev));
 
-       return (virtio_alloc_virtqueues(dev, 1, &vq_info));
+       return (virtio_alloc_virtqueues(dev, 0, 1, &vq_info));
 }
 
 /* Probe for existence of 9P virtio channels */
```
to the initial commit. This is because 180c02405b12e08bbb69bb693a38077db0c85e9b was never MFCed.

2. I'm not sure if ObsoleteFiles.inc part of 5b62dc782d4891845b47f2c0dfbb7717cbfe3897 should be picked. Since this is the first time the change lands on stable/14, we can skip this part?

3. I only picked p9fs part of 0d224af399a66f00a5b33e5512fc018062cabf1d . It seems that all other parts except unionfs are already picked.

4. I skipped picking e453e498cbb88570a3ff7b3679de65c88707da95 as it causes a lot of churn, but doesn't seem important.

My testing was `make buildkernel`, but I did not run-test this.

### DIFF
--- a/sys/fs/p9fs/p9fs_vfsops.c
+++ b/sys/fs/p9fs/p9fs_vfsops.c
@@ -345,6 +345,8 @@ p9fs_vget_common(struct mount *mp, struct p9fs_node *np, int flags,
 	P9FS_SET_LINKS(inode);
 
 	lockmgr(vp->v_vnlock, LK_EXCLUSIVE, NULL);
+	if (vp->v_type != VFIFO)
+		VN_LOCK_ASHARE(vp);
 	error = insmntque(vp, mp);
 	if (error != 0) {
 		/*


arrowd:
https://reviews.freebsd.org/D50760 is also related, but is not landed yet.

bapt:
this looks good to me, @dfr can you have a look? @arrowd if you have run tested it, than please give some time for @dfr  to step on it then please push

bapt (approved):

arrowd:
I did not run-test it. Let me know if I should do it, because that'd require setting up a 14-stable VM for me.

dfr:
I can test it, hopefully later this afternoon after I make a fresh
stable/14 build.

Doug.

dfr:
I tested this on a bhyve vm with a p9fs root filesystem and it works as far as I can tell with basic testing.

I agree that the ObsoleteFiles.inc fix isn't needed for stable/15 but it would be harmless if it got merged anyway. Thanks for taking the time to merge all this over to stable/14.

bsdimp:
So are there any actual differences / resolutions that are needed? Or can we cherry-pick the long list of commits like we do with everything else?

arrowd:
Pushed, thanks everyone.

PR #1746, opened 2025-06-30
sockstat: Reintroduce -w flag to automatically size the columns

daminrido139:
The -w flag was previously removed when automatic column sizing was introduced. However, that approach was not compact on small terminal windows, such as 80-column widths.

In this commit, sockstat now uses fixed-width columns by default for cleaner and more compact output. The -w flag has been reintroduced to automatically size the columns, ensuring they are wide enough to fit the longest entry in each column.

Sponsored by:  Google, LLC (GSoC 2025)
MFC after:  2 weeks

### DIFF
--- a/usr.bin/sockstat/sockstat.c
+++ b/usr.bin/sockstat/sockstat.c
@@ -97,6 +97,7 @@ static bool	 opt_s;		/* Show protocol state if applicable */
 static bool	 opt_U;		/* Show remote UDP encapsulation port number */
 static bool	 opt_u;		/* Show Unix domain sockets */
 static u_int	 opt_v;		/* Verbose mode */
+static bool	 opt_w;		/* Automatically size the columns */
 
 /*
  * Default protocols to use if no -P was defined.
@@ -1101,7 +1102,7 @@ format_unix_faddr(struct addr *faddr, char *buf, size_t bufsize) {
 	/* Remote peer we connect(2) to, if any. */
 	if (faddr->conn != 0) {
 		struct sock *p;
-		pos += strlcpy(buf, "-> ", bufsize);
+		pos += strlcpy(SAFEBUF, "-> ", SAFESIZE);
 		p = RB_FIND(pcbs_t, &pcbs,
 			&(struct sock){ .pcb = faddr->conn });
 		if (__predict_false(p == NULL)) {
@@ -1132,8 +1133,7 @@ format_unix_faddr(struct addr *faddr, char *buf, size_t bufsize) {
 		while ((p = RB_FIND(pcbs_t, &pcbs,
 			&(struct sock){ .pcb = ref })) != 0) {
 			f = RB_FIND(files_t, &ftree,
-				&(struct file){ .xf_data =
-				p->socket });
+				&(struct file){ .xf_data = p->socket });
 			if (f != NULL) {
 				pos += snprintf(SAFEBUF, SAFESIZE,
 					"%s[%lu %d]", fref ? "" : ",",
@@ -1178,13 +1178,10 @@ calculate_sock_column_widths(struct col_widths *cw, struct sock *s)
 	len = strlen(s->protoname);
 	if (s->vflag & (INP_IPV4 | INP_IPV6))
 		len += 1;
-	if (laddr != NULL && faddr != NULL && s->family == AF_UNIX &&
-		laddr->address.ss_len == 0 && faddr->conn == 0)
-		len += strlen(" (not connected)");
 	cw->proto = MAX(cw->proto, len);
 
 	while (laddr != NULL || faddr != NULL) {
-		if (s->family == AF_UNIX) {
+		if (opt_w && s->family == AF_UNIX) {
 			if ((laddr == NULL) || (faddr == NULL))
 				errx(1, "laddr = %p or faddr = %p is NULL",
 					(void *)laddr, (void *)faddr);
@@ -1193,7 +1190,7 @@ calculate_sock_column_widths(struct col_widths *cw, struct sock *s)
 			cw->local_addr = MAX(cw->local_addr, len);
 			len = format_unix_faddr(faddr, NULL, 0);
 			cw->foreign_addr = MAX(cw->foreign_addr, len);
-		} else {
+		} else if (opt_w) {
 			if (laddr != NULL) {
 				len = formataddr(&laddr->address, NULL, 0);
 				cw->local_addr = MAX(cw->local_addr, len);
@@ -1296,23 +1293,6 @@ calculate_sock_column_widths(struct col_widths *cw, struct sock *s)
 static void
 calculate_column_widths(struct col_widths *cw)
 {
-	cw->user = 4;
-	cw->command = 10;
-	cw->pid = 3;
-	cw->fd = 2;
-	cw->proto = 5;
-	cw->local_addr = 13;
-	cw->foreign_addr = 15;
-	cw->pcb_kva = 18;
-	cw->fib = 3;
-	cw->splice_address = 14;
-	cw->inp_gencnt = 2;
-	cw->encaps = 6;
-	cw->path_state = 10;
-	cw->conn_state = 10;
-	cw->stack = 5;
-	cw->cc = 2;
-
 	int n, len;
 	struct file *xf;
 	struct sock *s;
@@ -1366,13 +1346,10 @@ display_sock(struct sock *s, struct col_widths *cw, char *buf, size_t bufsize)
 	faddr = s->faddr;
 	first = true;
 
-	snprintf(buf, bufsize, "%s%s%s%s",
+	snprintf(buf, bufsize, "%s%s%s",
 		s->protoname,
 		s->vflag & INP_IPV4 ? "4" : "",
-		s->vflag & INP_IPV6 ? "6" : "",
-		(laddr != NULL && faddr != NULL &&
-		s->family == AF_UNIX && laddr->address.ss_len == 0 &&
-		faddr->conn == 0) ? " (not connected)" : "");
+		s->vflag & INP_IPV6 ? "6" : "");
 	printf(" %-*s", cw->proto, buf);
 	while (laddr != NULL || faddr != NULL) {
 		if (s->family == AF_UNIX) {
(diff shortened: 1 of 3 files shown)


asomers:
Could you please adjust your commit message to emphasize how the new behavior of "-w" differs from the original behavior?

daminrido139:
done @asomers !

daminrido139:
Hi @asomers , just wanted to kindly follow up on this. I've updated the commit message. please let me know if there's anything else needs to be changed. Thanks!

usr.bin/sockstat/sockstat.c
```diff
@@ -1510,7 +1492,27 @@ display(void)
...
+		.local_addr = opt_w ? strlen("LOCAL ADDRESS") : 21,
+		.foreign_addr = opt_w ? strlen("FOREIGN ADDRESS") : 21,
+		.pcb_kva = 18,
+		.fib = opt_w ? strlen("FIB") : 6,
+		.splice_address = opt_w ? strlen("SPLICE ADDRESS") : 21,
+		.inp_gencnt = opt_w ? strlen("ID") : 8,
+		.encaps = opt_w ? strlen("ENCAPS") : 6,
+		.path_state = opt_w ? strlen("PATH STATE") : 12,
+		.conn_state = opt_w ? strlen("CONN STATE") : 12,
+		.stack = opt_w ? strlen("STACK") : TCP_FUNCTION_NAME_LEN_MAX,
```
asomers: `TCP_FUNCTION_NAME_LEN_MAX` is much too wide.  No existing stack comes close to using the max width.

usr.bin/sockstat/sockstat.c
```diff
@@ -1510,7 +1492,27 @@ display(void)
...
+		.cc = opt_w ? strlen("CC") : TCP_CA_NAME_MAX,
```
asomers: `TCP_CA_NAME_MAX` is also too wide.  The values for these two widths in the previous revision, 5 and 2, were better.

daminrido139:
Hi @asomers ,
Thanks for the detailed review! You’re absolutely right. I’ve now updated the code to always compute column widths dynamically (except for the address-like columns), and I’ve fixed the miscalculation in the PROTO column as well. Let me know if this looks better now!

asomers (changes requested):
Testing still shows some problems:
1) Without "-w", the LOCAL ADDRESS column can overflow
2) Without "-w", the FOREIGN ADDRESS column can overflow and exceed 80 cols.
Both problems are visible in this screenshot:
```
root   nfsuserd    2076  3 udp6   *:618                 *:*                  
root   zfsd        1691  9 seqpac ??                    -> /var/run/devd.seqpacket.pipe
root   zfsd        1691 10 dgram  ??                    -> /var/run/log      
root   devd        1593  4 stream /var/run/devd.pipe    ??                   
root   devd        1593  5 seqpac /var/run/devd.seqpacket.pipe ??                   
root   devd        1593  8 seqpac /var/run/devd.seqpacket.pipe -> [1691 9]          
root   devd        1593  9 dgram  ??                    -> /var/run/log      
```
If you're going to limit a column's width to 80 cols, you should truncate entries within it.  I think sockstat already does truncate for the USER column.
3) Without "-w", the SPLICE ADDRESS column is too wide, when `-I` is used.

daminrido139:
Hi @asomers ,
Thanks for pointing those out! I've addressed all three issues: added truncation for LOCAL ADDRESS and FOREIGN ADDRESS when `-w` not used, and always calculate SPLICE ADDRESS width when `-I` is used. Let me know if this looks good now!

daminrido139:
Hi @asomers ,

> One more problem. The "SPLICE ADDRESS" column can overflow, whether or not "-w" is in use. In order to create some spliced sockets to view, you can cd to /usr/tests/sys/kern and run kyua test socket_splice

I am looking into it.

> Also, I see that the Author of your three commits in this PR is [daminrido139@gmail.com](mailto:daminrido139@gmail.com) rather than [rido@FreeBSD.org](mailto:rido@FreeBSD.org) . You can use either, but was the switch intentional?

Yes, it was intentional. Earlier, when I mistakenly used FreeBSD.com, you advised me to use the real email address. I thought you meant [daminrido139@gmail.com](mailto:daminrido139@gmail.com) rather than the other. If you'd prefer I switch to [rido@FreeBSD.org](mailto:rido@FreeBSD.org), I will do so starting with the next commit.

>  Also, weirdly, using -w can make the "LOCAL ADDRESS" column narrower, especially when -4 is in use. While that's not a bug, it is certainly unexpected.

Yes, it's not a bug. When -4 is used (regardless of whether -w is used), it shows only IPv4 sockets, which have a maximum width of 21. So, it can accommodate all sockets within that width. That's why the "LOCAL ADDRESS" and "FOREIGN ADDRESS" columns appear narrower.

daminrido139:
Hi @asomers ,
Regarding the "SPLICE ADDRESS" column, its width is currently calculated dynamically, regardless of whether  `-w` is used. So it shouldn’t overflow into other columns. I also tested it as you suggested by creating some spliced sockets, and the output looks fine on my machine.

Could you confirm if you're still seeing the overflow? If so, a specific example or test case would be really helpful.

asomers:
> Hi @asomers , Regarding the "SPLICE ADDRESS" column, its width is currently calculated dynamically, regardless of whether `-w` is used. So it shouldn’t overflow into other columns. I also tested it as you suggested by creating some spliced sockets, and the output looks fine on my machine.
> 
> Could you confirm if you're still seeing the overflow? If so, a specific example or test case would be really helpful.

I ran the tests that I mentioned, and now I see that column overflowing by one character, like this:
```
> /usr/obj/usr/home/somers/src/freebsd.org/src/amd64.amd64/usr.bin/sockstat/sockstat -I4C
USER   COMMAND     PID FD PROTO LOCAL ADDRESS    FOREIGN ADDRESS    SPLICE ADDRESS CC   
somers sshd-sessi 6372  4 tcp4  192.168.0.27:22  192.168.0.16:29256 ??             cubic
root   sshd-sessi 6370  4 tcp4  192.168.0.27:22  192.168.0.16:29256 ??             cubic
somers sshd-sessi 4091  4 tcp4  192.168.0.27:22  192.168.0.16:33428 ??             cubic
root   sshd-sessi 4089  4 tcp4  192.168.0.27:22  192.168.0.16:33428 ??             cubic
somers sshd-sessi 4082  4 tcp4  192.168.0.27:22  192.168.0.16:57422 ??             cubic
root   sshd-sessi 4080  4 tcp4  192.168.0.27:22  192.168.0.16:57422 ??             cubic
root   sshd       3974  8 tcp4  *:22             *:*                ??             cubic
root   syslogd    2679  7 udp4  *:514            *:*                ??             ??   
??     ??           ?? ?? tcp4  127.0.0.1:36039  127.0.0.1:21469    127.0.0.1:21469 cubic
??     ??           ?? ?? tcp4  127.0.0.1:25059  127.0.0.1:29789    127.0.0.1:29789 cubic
??     ??           ?? ?? tcp4  127.0.0.1:65417  127.0.0.1:64376    127.0.0.1:64376 cubic
??     ??           ?? ?? tcp4  127.0.0.1:29789  127.0.0.1:25059    127.0.0.1:25059 cubic
??     ??           ?? ?? tcp4  127.0.0.1:63584  127.0.0.1:43323    127.0.0.1:43323 cubic
??     ??           ?? ?? tcp4  127.0.0.1:31661  127.0.0.1:30747    127.0.0.1:30747 cubic
??     ??           ?? ?? tcp4  192.168.0.27:890 192.168.0.2:2049   ??             cubic
??     ??           ?? ?? tcp4  127.0.0.1:64376  127.0.0.1:65417    127.0.0.1:65417 cubic
??     ??           ?? ?? tcp4  127.0.0.1:21469  127.0.0.1:36039    127.0.0.1:36039 cubic
??     ??           ?? ?? tcp4  127.0.0.1:11437  127.0.0.1:26663    127.0.0.1:26663 cubic
```

asomers:
> Yes, it was intentional. Earlier, when I mistakenly used FreeBSD.com, you advised me to use the real email address. I thought you meant [daminrido139@gmail.com](mailto:daminrido139@gmail.com) rather than the other. If you'd prefer I switch to [rido@FreeBSD.org](mailto:rido@FreeBSD.org), I will do so starting with the next commit.

Well, rido@FreeBSD.org is the best, if you plan to stick around ;) .

daminrido139:
Hi @asomers ,
After some debugging, I found the issue. Previously, in some cases, the SPLICE ADDRESS was the same as the FOREIGN ADDRESS. This has now been fixed. Kindly review it again. Thanks!

asomers (approved):

bsdimp:
@glebius what do you think of this?

daminrido139:
Hi @asomers ,
Thanks again for the approval! I noticed the PR hasn’t been merged yet and saw that @glebius  was mentioned for input. Just wanted to check in - should I wait for his response, or is there anything else I should do from my side?

glebius:
@daminrido139 thanks for tagging! I've been damn busy with an urgent issue at work and postpone reading FreeBSD emails since Friday. I will try to get back tomorrow or later this week. Sorry for that!

glebius (approved):
Thanks a lot for restoring the classic behavior by default! The wide mode is definitely very useful, and I'm going to use it. Thanks!

One suggestion for the wide mode: trim the whitespace at the end of lines. A single long FOREIGN ADDRESS makes the line not fitting into almost any terminal window. That's fine if terminal wraps this line. But we don't want to have every other line trailing whitespace to be also printed, effectively creating an absolutely empty lines.

daminrido139:
Thanks for your feedback, @glebius ! That’s a great point regarding the trailing whitespace in wide mode. I'll definitely address it in a future commit, either during the GSoC period or right after.

@asomers , would it be okay to merge this PR now so I can start working on the libxo conversion in a separate set of changes?

asomers:
> Thanks for your feedback, @glebius ! That’s a great point regarding the trailing whitespace in wide mode. I'll definitely address it in a future commit, either during the GSoC period or right after.
> 
> @asomers , would it be okay to merge this PR now so I can start working on the libxo conversion in a separate set of changes?

It would be OK, yes. But I'm travelling right now, so I can't do it until Saturday.  I suggest you create a separate branch to work on the libxo changes. You can include this pr's change in that branch for now, and then later rebase that branch after I've merged this one.

daminrido139:
Thanks, @asomers ! I'll proceed as you suggested. Safe travels!

PR #1748, opened 2025-06-30
Fix processor accounting on non-x86

ehem:
The underlying problem with processor accounting on non-x86 appears `intr_event_handle()` was designed for large numbers of assembly-language to C bridges and few cascaded PICs (similar to x86 with 5 bridges and only 2 cascades).  Whereas all other architectures have a single bridge and many cascading PICs.

Issue is if a cascading PIC calls `intr_event_handle()` from an interrupt handle, `curthread->td_intr_nesting_level` will be 2 and the timeslice will be accounted as being consumed by interrupt handling.  Due to the speed of modern processors the error in time accounting may not be huge, but it is lurking.

This resolves the issue by moving `++curthread->td_intr_nesting_level`, pushing of `curthread->td_intr_frame` and the `critical_enter()` out of `intr_event_handle()` and into the bridge points.

### DIFF
--- a/sys/kern/kern_intr.c
+++ b/sys/kern/kern_intr.c
@@ -111,7 +111,7 @@ static struct mtx event_lock;
 MTX_SYSINIT(intr_event_list, &event_lock, "intr event list", MTX_DEF);
 
 static void	intr_event_update(struct intr_event *ie);
-static int	intr_event_schedule_thread(struct intr_event *ie, struct trapframe *frame);
+static int	intr_event_schedule_thread(struct intr_event *ie);
 static struct intr_thread *ithread_create(const char *name);
 static void	ithread_destroy(struct intr_thread *ithread);
 static void	ithread_execute_handlers(struct proc *p, 
@@ -810,8 +810,15 @@ intr_handler_barrier(struct intr_handler *handler)
 		return;
 	}
 	if ((handler->ih_flags & IH_CHANGED) == 0) {
+#ifdef HWPMC_HOOKS
+		struct trapframe *oldframe = curthread->td_intr_frame;
+		curthread->td_intr_frame = NULL;
+#endif
 		handler->ih_flags |= IH_CHANGED;
-		intr_event_schedule_thread(ie, NULL);
+		intr_event_schedule_thread(ie);
+#ifdef HWPMC_HOOKS
+		curthread->td_intr_frame = oldframe;
+#endif
 	}
 	while ((handler->ih_flags & IH_CHANGED) != 0)
 		msleep(handler, &ie->ie_lock, 0, "ih_barr", 0);
@@ -894,6 +901,10 @@ intr_event_remove_handler(void *cookie)
 		CK_SLIST_REMOVE_PREVPTR(prevptr, ih, ih_next);
 		intr_event_barrier(ie);
 	} else {
+#ifdef HWPMC_HOOKS
+		struct trapframe *oldframe = curthread->td_intr_frame;
+		curthread->td_intr_frame = NULL;
+#endif
 		/*
 		 * Let the interrupt thread do the job.  The interrupt source is
 		 * disabled when the interrupt thread is running, so it does not
@@ -902,7 +913,10 @@ intr_event_remove_handler(void *cookie)
 		KASSERT((handler->ih_flags & IH_DEAD) == 0,
 		    ("duplicate handle remove"));
 		handler->ih_flags |= IH_DEAD;
-		intr_event_schedule_thread(ie, NULL);
+		intr_event_schedule_thread(ie);
+#ifdef HWPMC_HOOKS
+		curthread->td_intr_frame = oldframe;
+#endif
 		while (handler->ih_flags & IH_DEAD)
 			msleep(handler, &ie->ie_lock, 0, "iev_rmh", 0);
 	}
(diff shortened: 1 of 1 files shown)


sys/x86/x86/local_apic.c
```diff
@@ -1272,6 +1272,9 @@ lapic_handle_intr(int vector, struct trapframe *frame)
 {
 	struct intsrc *isrc;
 
+	critical_enter();
+	++curthread->td_intr_nesting_level;
+
 	kasan_mark(frame, sizeof(*frame), sizeof(*frame), 0);
 	kmsan_mark(&vector, sizeof(vector), KMSAN_STATE_INITED);
 	kmsan_mark(frame, sizeof(*frame), KMSAN_STATE_INITED);
```
ehem: I'm unsure of the ordering here.  Should these instead be in the opposite order?  (similar issue above in `atpic.c`)

sys/x86/isa/atpic.c
```diff
@@ -526,6 +527,9 @@ atpic_handle_intr(u_int vector, struct trapframe *frame)
 {
 	struct intsrc *isrc;
 
+	critical_enter();
+	++curthread->td_intr_nesting_level;
+
 	kasan_mark(frame, sizeof(*frame), sizeof(*frame), 0);
 	kmsan_mark(frame, sizeof(*frame), KMSAN_STATE_INITED);
 	trap_check_kstack();
```
ehem: I'm unsure of the ordering here.  Should these instead be in the opposite order?  (similar issue below in `local_apic.c`)

sys/kern/kern_intr.c
```diff
@@ -810,8 +810,15 @@ intr_handler_barrier(struct intr_handler *handler)
...
+#ifdef HWPMC_HOOKS
+		struct trapframe *oldframe = curthread->td_intr_frame;
+		curthread->td_intr_frame = NULL;
+#endif
 		handler->ih_flags |= IH_CHANGED;
-		intr_event_schedule_thread(ie, NULL);
+		intr_event_schedule_thread(ie);
+#ifdef HWPMC_HOOKS
+		curthread->td_intr_frame = oldframe;
+#endif
```
ehem, line 821: I'm unsure here.  This keeps functionality unchanged.  Issue is most other points, passing `NULL` as a `struct trapframe *` was due to the author not knowing what to do.  Could be `curthread->td_intr_frame` should simply be left alone here, but I'm unsure.

sys/kern/kern_intr.c
```diff
@@ -894,6 +901,10 @@ intr_event_remove_handler(void *cookie)
 		CK_SLIST_REMOVE_PREVPTR(prevptr, ih, ih_next);
 		intr_event_barrier(ie);
 	} else {
+#ifdef HWPMC_HOOKS
+		struct trapframe *oldframe = curthread->td_intr_frame;
+		curthread->td_intr_frame = NULL;
+#endif
```
ehem, line 907: Same issue as above.  Should `curthread->td_intr_frame` simply be left alone here?  Was there a deliberate need to simulate an empty trap frame?

sys/kern/kern_clock.c
```diff
@@ -533,8 +535,11 @@ hardclock(int cnt, int usermode)
 
 		critical_enter();
 		curthread->td_intr_nesting_level++;
-		intr_event_handle(clk_intr_event, NULL);
+		oldframe = curthread->td_intr_frame;
+		curthread->td_intr_frame = NULL;
+		intr_event_handle(clk_intr_event);
 		curthread->td_intr_nesting_level--;
+		curthread->td_intr_frame = oldframe;
 		critical_exit();
```
ehem: This is another _odd_ spot.  Following the inclination to keep functionality unchanged, but this seems an odd choice.

sys/kern/kern_intr.c
```diff
@@ -1105,8 +1122,15 @@ swi_sched(void *cookie, int flags)
...
 		VM_CNT_INC(v_soft);
-		error = intr_event_schedule_thread(ie, NULL);
+		error = intr_event_schedule_thread(ie);
+#ifdef HWPMC_HOOKS
+		curthread->td_intr_frame = oldframe;
+#endif
```
ehem, line 1133: Yet another instance of this.  Alas here I'm again wondering about correct course of action.

ehem:
Unlike many of my patch series, this series is mostly independent of one-another.  They're a bit interesting to reorder, but should be doable in most any order.  Is notable the first is the highest value (since it is a bugfix), whereas the rest are somewhere closer to cleanup/reduction in stack size.

sys/arm/mv/gpio.c
```diff
@@ -537,7 +537,7 @@ mv_gpio_intr_handler(device_t dev, int pin)
 	    CK_SLIST_EMPTY(&isrc.isrc_event->ie_handlers))
 		return;
 
-	intr_isrc_dispatch(&isrc, NULL);
+	intr_isrc_dispatch(&isrc);
```
ehem, line 541: If someone wants to cherry-pick b42bfb832b82ffd1559b066d4daa062b8b84db09 from #1457 I would be a bit happy.  This source file has become rather troublesome and I'm pretty sure the driver is now unused.

ehem:
@chmeeedalf #1748 should be of interest to you.  I notice several PowerPC PICs have issues due to what this is pointing at.

ehem:
Looks like OpenSSL fun is now killing builds.

Ping to @bsdjhb (x86), @strejda (INTRNG),and @chmeeedalf (PowerPC) as likely reviewers.  The impact might not be all that big, but has been causing errors in time accounting for a long time.

PR #1751, opened 2025-07-02
tests/ci: fix qemu-user-static existence check

svmhdvn:


tests/ci/Makefile
```diff
@@ -144,10 +144,10 @@ portinstall-pkg: .PHONY
 .endif
 
 portinstall-qemu: portinstall-pkg .PHONY
-.if !exists(/usr/local/bin/qemu-${TARGET_ARCH}-static)
+.if !exists(/usr/local/bin/qemu-${QEMU_ARCH}-static)
```
lwhsu, line 147: Not sure if we can use ${QEMUSTATIC} directly.

svmhdvn: Not easily, since `QEMUSTATIC` is conditionally defined. I didn't want to mess around with too much logic for this simple fix, it can be cleaned up as style later.

lwhsu: Yeah, I also noticed that and I think `QEMUSTATIC` is only useful when cross-build and on non-x86. We can of course clean up that later.

lwhsu (approved):

lwhsu:
Merged as ed4c64bc792fd2509d8fd6805f8e174c2db62338

PR #1752, opened 2025-07-03
fmod: avoid undefined left shift of int32_t

phoddie:
For certain input values `__ieee754_fmod` does a shift left of an `int32_t` value (variable `i`) greater than `0x3FFFFFFF`. In C, this is an undefined behavior so the subsequent check if the resulting value is less than 0 may fail.

```c
for (iy = -1022,i=(hy<<11); i>0; i<<=1) iy -=1;
```

The following calls to `fmod` trigger an invalid shift:

```c
__ieee754_fmod(2, 1.0578376737312813e-308);
__ieee754_fmod(2, 3.1735130211938439e-309);
__ieee754_fmod(2, 9.5205390635815563e-310);
__ieee754_fmod(2, 2.8561617190744471e-310);
__ieee754_fmod(2, 8.5684851572235885e-311);
__ieee754_fmod(2, 2.5705455471671259e-311);
__ieee754_fmod(2, 7.7116366415008838e-312);
__ieee754_fmod(2, 2.3134909924497711e-312);
__ieee754_fmod(2, 6.9404729773295506e-313);
__ieee754_fmod(2, 2.0821418932038058e-313);
__ieee754_fmod(2, 6.2464256797596372e-314);
```

This PR includes a simple change to avoid the undefined behavior at the expense of an additional compare. Note that there are three other loops in this function that use the same pattern; we have not observed them triggering an undefined behavior and so have left them unchanged.

This issue was discovered while testing using [ClusterFuzz](https://github.com/google/clusterfuzz) via oss-fuzz with [UndefinedBehaviorSanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) enabled.

Signed-off-by: Peter Hoddie <peter@moddable.com>


phoddie:
The following calls also trigger an invalid shift:

```c
__ieee754_fmod(2, 5.6217831096591912e-315);
__ieee754_fmod(2, 1.6865349343799543e-315);
__ieee754_fmod(2, 5.0596047883178935e-316);
__ieee754_fmod(2, 1.5178814216733987e-316);
__ieee754_fmod(2, 4.5536444626464544e-317);
__ieee754_fmod(2, 1.3660934870136301e-317);
__ieee754_fmod(2, 4.0982794729095985e-318);
__ieee754_fmod(2, 1.229482359675942e-318);
__ieee754_fmod(2, 3.6884470790278261e-319);
__ieee754_fmod(2, 1.1065588269906399e-319);
__ieee754_fmod(2, 3.3196270744073355e-320);
__ieee754_fmod(2, 9.9603634201595303e-321);
__ieee754_fmod(2, 2.9890971573395416e-321);
__ieee754_fmod(2, 8.9425881897265624e-322);
__ieee754_fmod(2, 2.6679544875427313e-322);
__ieee754_fmod(2, 7.9050503334599447e-323);
__ieee754_fmod(2, 2.4703282292062327e-323);
__ieee754_fmod(2, 4.9406564584124654e-324);
```

kargl:
See https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=288778

phoddie:
@kargl, thanks for taking this on. Much appreciated. Your changes are (obviously) far more thorough than my quick work-around.

I'll get this merged into our engine, so oss-fuzz can do its usual stress testing on the changes.

> a bit of code spelunking shows that the issue raised at https://github.com/freebsd/freebsd-src/pull/1752 is simply to appease a sanitizer

I tried to be clear about that above. Apologies if it was insufficiently explicit.

kargl:
The problem with libm/msun sources is that there is a rat's nest of left-shift of signed integers.
For as long as I have hacked on libm, there has been an effort to reduce the diff between the float and double versions of a function.  The algorithm should be the same between both with only magic numbers and maybe functions names differing.  I fixed the issue in src/e_fmod.c, src/s_ilogb.c, and src/s_remquo.c that you raised.  This issue also exists in the float version of these functions, so now I need fix the float versions.  I do this work as a hobby, but do feel a commitment to make thing correct.

phoddie:
@kargl – That all makes good sense. Your volunteer efforts here to pull this important code base forward are invaluable.

FWIW – Moddable is using fdlibm in our embedded JavaScript engine, XS, to provide consistent floating point results across a wide variety of hosts. This is important in JavaScript as variations in the result of math operations have been used to identify/fingerprint users. Variations are also a small source of entropy that can lead to non-deterministic execution, which is a problem for blockchain uses of JavaScript like Agoric does. That's why the potential non-determinism of undefined behaviors is important to address. My understanding is that other JavaScript engines, including some used in web browsers, also rely on fdlibm, primarily to avoid user fingerprinting.

Overnight, oss-fuzz turned up another undefined left shift in `fmod` on this line:

```c
lx = (hx<<(32-n))|(lx>>n); hx = sx;
```

I haven't isolated the values it is using yet. Once I do, where is the best place to report those?

kargl:
> Overnight, oss-fuzz turned up another undefined left shift in `fmod` on this line:
> 
> ```c
> lx = (hx<<(32-n))|(lx>>n); hx = sx;
> ```

It would be quicker to just grep.

% grep -R "hx<<" src
src/e_fmod.c:           hx = (hx<<n)|(lx>>(32-n));
src/e_fmod.c:           lx = (lx>>n)|((u_int32_t)hx<<(32-n));
src/e_fmod.c:           lx = (hx<<(32-n))|(lx>>n); hx = sx;
src/e_fmodf.c:      hx = hx<<n;
src/s_remquo.c:         hx = (hx<<n)|(lx>>(32-n));
src/s_remquo.c:         lx = (lx>>n)|((u_int32_t)hx<<(32-n));
src/s_remquo.c:         lx = (hx<<(32-n))|(lx>>n); hx = 0;
src/e_fmodf.c:      for (ix = -126,i=(hx<<8); i>0; i<<=1) ix -=1;
src/s_ilogbf.c:         for (ix = -126,hx<<=8; hx>0; hx<<=1) ix -=1;
src/s_remquof.c:            for (ix = -126,i=(hx<<8); i>0; i<<=1) ix -=1;

I've submitted a patch to fix the last three.
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=288850

I haven't analyzed the other left shifts.  Others that need to be
looked at are


% grep -R "hx <<" src
src/s_remquof.c:            hx <<= n;
src/s_remquof.c:            if(hz<0) hx = hx << 1;
src/s_remquof.c:            hx <<= 1;

% grep -R "lx<<" src
src/e_fmod.c:           hx = lx<<(n-32);
src/s_remquo.c:         hx = lx<<(n-32);

% grep -R "lx <<" src
src/e_fmod.c:           lx <<= n;
src/s_remquo.c:         lx <<= n;

phoddie:
> It would be quicker to just grep.

Of course. ;)

PR #1754, opened 2025-07-04
tests/ci: run ci-full kyua tests in parallel

svmhdvn:
By default, use all available cpus given to the VM. This can be controlled with the already available PARALLEL_JOBS make variable.

### DIFF
--- a/tests/ci/tools/freebsdci
+++ b/tests/ci/tools/freebsdci
@@ -34,6 +34,7 @@ rcvar=freebsdci_enable
 start_cmd="firstboot_ci_run"
 stop_cmd=":"
 os_arch=$(uname -p)
+parallelism=$(nproc)
 tardev=/dev/vtbd1
 metadir=/meta
 istar=$(file -s ${tardev} | grep "POSIX tar archive" | wc -l)
@@ -74,7 +75,7 @@ full_tests()
 		tar xvf ${tardev} -C ${metadir}
 		cd /usr/tests
 		set +e
-		kyua test
+		kyua -v parallelism=${parallelism} test
 		rc=$?
 		set -e
 		if [ ${rc} -ne 0 ] && [ ${rc} -ne 1 ]; then


lwhsu:
I like this, and I believe this is the direction we want in the long term. For now I'm a bit hesitated of it because from the experience some test cases may conflict and cause the result not stable.  (This is of course a bug and we need to fix.  Either fixing the cases cannot run in parallel and providing an exclusive annotation for test case use.) Before that, do you think it's ok to make this feature opt-in and we can see if it works fine?

svmhdvn:
> Before that, do you think it's ok to make this feature opt-in and we can see if it works fine?

Since the `tests/ci` Makefile CI infra is not the default right now, then I think we should keep this feature *opt-out* instead of opt-in. Currently, Cirrus CI is only set to run smoke tests, so this PR should not affect that. However, when people are interested in running local CI and trying things out, they ideally SHOULD notice that badly parallelized tests are failing. If we keep it opt-out, then they will not notice until they opt-in.

lwhsu:
> > Before that, do you think it's ok to make this feature opt-in and we can see if it works fine?
> 
> Since the `tests/ci` Makefile CI infra is not the default right now, then I think we should keep this feature _opt-out_ instead of opt-in. Currently, Cirrus CI is only set to run smoke tests, so this PR should not affect that. However, when people are interested in running local CI and trying things out, they ideally SHOULD notice that badly parallelized tests are failing. If we keep it opt-out, then they will not notice until they opt-in.

Ok, that sounds a plan to me.  I was worried that people see too many failures and think this is not useful, but we still can change the default if things went bad.

kprovost:
> I like this, and I believe this is the direction we want in the long term. For now I'm a bit hesitated of it because from the experience some test cases may conflict and cause the result not stable. (This is of course a bug and we need to fix. Either fixing the cases cannot run in parallel and providing an exclusive annotation for test case use.) 
>
At least the networking tests do work just fine when run at limited parallelism levels (2 to 4 is usually fine). Higher levels are very likely to have a lot of intermittent failures. A lot of networking functions have timeouts, and when we run many tests in parallel these timeouts will become a factor and break test cases.

lwhsu:
merged as f6e5bcd4716c2974c7ceb1fbbfc0ef8db8e13d8b

PR #1755, opened 2025-07-05
mgb: fix mac tx disable bug in teardown

jefby:
When mgb teardown, mac tx should be disabled instead of remaining enabled

### DIFF
--- a/sys/dev/mgb/if_mgb.c
+++ b/sys/dev/mgb/if_mgb.c
@@ -1435,7 +1435,7 @@ mgb_hw_teardown(struct mgb_softc *sc)
 
 	/* Stop MAC */
 	CSR_CLEAR_REG(sc, MGB_MAC_RX, MGB_MAC_ENBL);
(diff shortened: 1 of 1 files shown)


sys/dev/mgb/if_mgb.c
```diff
@@ -1435,7 +1435,7 @@ mgb_hw_teardown(struct mgb_softc *sc)
 
 	/* Stop MAC */
 	CSR_CLEAR_REG(sc, MGB_MAC_RX, MGB_MAC_ENBL);
-	CSR_WRITE_REG(sc, MGB_MAC_TX, MGB_MAC_ENBL);
```
gmshake, line 1438: Not sure if that is intentional or not, but it appears to be a mistake. CC @emaste  .

emaste: Yes appears so. Thanks @jefby for the fix, will push soon.

bsdimp:
this didn't autoclose.

PR #1757, opened 2025-07-07
tests/ci: fix race condition in bhyveload boot

svmhdvn:


tests/ci/Makefile
```diff
@@ -210,8 +210,10 @@ ci-extract-meta: .PHONY
 ci-runtest: ci-buildimage-${TARGET_ARCH:tl} portinstall .PHONY
 .if ${MACHINE} == "amd64" && ( ${TARGET_ARCH} == "amd64" || ${TARGET_ARCH} == "i386" ) && ( !defined(USE_QEMU) || empty(USE_QEMU) )
 	/usr/sbin/bhyvectl --vm=${TEST_VM_NAME} --destroy || true
-	/usr/sbin/bhyveload -c stdio -m ${VM_MEM_SIZE} -d ${CIDISK} ${TEST_VM_NAME}
 	expect -c "set timeout ${TIMEOUT_EXPECT}; \
+		spawn /usr/bin/timeout -k 60 ${TIMEOUT_VM} /usr/sbin/bhyveload \
```
lwhsu: I am a bit worried that `${TIMEOUT_VM}` is too long here, it is used for waiting for the whole test suite. And `-k 60` to timeout(1) is for waiting for VM shutdown. I think bhyveload needs much shorter time. Perhaps `-k 5 30` or even short is fine.

tests/ci/Makefile
```diff
@@ -210,8 +210,10 @@ ci-extract-meta: .PHONY
...
+		-c stdio -m ${VM_MEM_SIZE} -d ${CIDISK} ${TEST_VM_NAME}; \
+		expect { eof }; \
```
lwhsu, line 216: I am not expect expert so this is more a question: will the expect script stops here if bhyveload fails?

lwhsu:
Please add more explanation in the comment message about the race condition, and the fix.

tests/ci/Makefile
svmhdvn: It does not. I have updated this PR to explicitly ensure that failure is handled correctly by making it explicit.

tests/ci/Makefile
svmhdvn: Fixed now.

svmhdvn:
Applied feedback and added a bit more info in the commit message about the fix.

lwhsu:
Merged as 2f50ea54c6e3b15f655e800eb680ddcb80e30b4f

PR #1758, opened 2025-07-07
Minor O_CLOFORK fixes

ricardobranco777:
- [mkostemp(3)](https://man.freebsd.org/cgi/man.cgi?query=mkostemp) should support O_CLOFORK as per https://pubs.opengroup.org/onlinepubs/9799919799/functions/mkdtemp.html
- ~Fix calls that naively set F_SETFD~

~The latter follows the same spirit as this patch suggested in OpenBSD, though the scope of this PR are not the security enhancements that are now possible with this flag~:
https://marc.info/?l=openbsd-tech&m=175055263416829&w=2

Ticket: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=286843

### DIFF
--- a/lib/libc/stdio/mktemp.3
+++ b/lib/libc/stdio/mktemp.3
@@ -25,7 +25,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd July 29, 2019
+.Dd July 07, 2025
 .Dt MKTEMP 3
 .Os
 .Sh NAME
@@ -101,9 +101,10 @@ The permitted flags are
 .Dv O_DIRECT ,
 .Dv O_SHLOCK ,
 .Dv O_EXLOCK ,
-.Dv O_SYNC
+.Dv O_SYNC ,
+.Dv O_CLOEXEC
 and
-.Dv O_CLOEXEC .
+.Dv O_CLOFORK .
 .Pp
 The
 .Fn mkstemps
--- a/lib/libc/stdio/mktemp.c
+++ b/lib/libc/stdio/mktemp.c
@@ -121,7 +121,7 @@ _gettemp(int dfd, char *path, int *doopen, int domkdir, int slen, int oflags)
 
 	if ((doopen != NULL && domkdir) || slen < 0 ||
 	    (oflags & ~(O_APPEND | O_DIRECT | O_SHLOCK | O_EXLOCK | O_SYNC |
-	    O_CLOEXEC)) != 0) {
+	    O_CLOEXEC | O_CLOFORK)) != 0) {
 		errno = EINVAL;
 		return (0);
 	}


lib/libc/gen/fdopendir.c
```diff
@@ -49,8 +49,14 @@
...
 {
+	int flags;
 
-	if (_fcntl(fd, F_SETFD, FD_CLOEXEC) == -1)
+	flags = _fcntl(fd, F_GETFD, 0);
+	if (flags == -1)
+		return (NULL);
+
+	if ((flags & FD_CLOEXEC) == 0 &&
+	    _fcntl(fd, F_SETFD, flags | FD_CLOEXEC) == -1)
```
kostikbel: I believe I already explained why is this not mt-safe.

ricardobranco777: > I believe I already explained why is this not mt-safe.

Yeah, and I thought it was better to have the discussion in another PR.

Isn't this function protected by this lock?

https://github.com/freebsd/freebsd-src/blob/main/lib/libc/stdio/freopen.c#L67

Otherwise I see the creation of a fcntl overkill, as we're trying to fix an edge-case of uncooperative threads.  In the OpenBSD patch suggested to me they didn't see the need for it.  But I may be missing something.

kostikbel: Function might be protected, but the fd flags are not.  If two threads to F_SETFD in parallel, they would override each other' flags.  It did not matter before because FD_CLOEXEC was the only flag that matter.  Now we have at least three flags, and no op to safely manipulate that in mt environment.

This is not limited to the current specific use, but for generic API completeness.

markjdb: I am confused. From my reading F_SETFD is an OR operation for all flags except FD_CLOEXEC. Why exactly is this change necessary?

ricardobranco777: > I am confused. From my reading F_SETFD is an OR operation for all flags except FD_CLOEXEC. Why exactly is this change necessary?

Previously this function was setting FD_CLOEXEC overwriting any previous flags. Now we get the flags and OR them with FD_CLOEXEC.

It's the same change done in the OpenBSD patch:
https://marc.info/?l=openbsd-tech&m=175055263416829&w=2

markjdb: I understand what the patch is trying to do, but look at what `F_SETFD` is doing:

```
 550                 fde = fdeget_noref(fdp, fd);                                                                                                                                                                                                                                                                             
 551                 if (fde != NULL) {                                                                                                                                                                                                                                                                                       
 552                         /*                                                                                                                                                                                                                                                                                               
 553                          * UF_RESOLVE_BENEATH is sticky and cannot be cleared.                                                                                                                                                                                                                                           
 554                          */                                                                                                                                                                                                                                                                                              
 555                         fde->fde_flags = (fde->fde_flags & ~UF_EXCLOSE) |                                                                                                                                                                                                                                                
 556                             ((arg & FD_CLOEXEC) != 0 ? UF_EXCLOSE : 0) |                                                                                                                                                                                                                                                 
 557                             ((arg & FD_CLOFORK) != 0 ? UF_FOCLOSE : 0) |                                                                                                                                                                                                                                                 
 558                             ((arg & FD_RESOLVE_BENEATH) != 0 ?                                                                                                                                                                                                                                                           
 559                             UF_RESOLVE_BENEATH : 0);                                                                                                                                                                                                                                                                     
 560                         error = 0;                                                                                                                                                                                                                                                                                       
 561                 }
```

It ORs the new flags with all the ones already set on the fd, except UF_EXCLOSE. In fact, I think it cannot be any other way, since otherwise the addition of FD_CLOFORK would break backwards compatibility, as this patch demonstrates.

ricardobranco777: > I understand what the patch is trying to do, but look at what `F_SETFD` is doing:
> 
> ```
>  550                 fde = fdeget_noref(fdp, fd);                                                                                                                                                                                                                                                                             
>  551                 if (fde != NULL) {                                                                                                                                                                                                                                                                                       
>  552                         /*                                                                                                                                                                                                                                                                                               
>  553                          * UF_RESOLVE_BENEATH is sticky and cannot be cleared.                                                                                                                                                                                                                                           
>  554                          */                                                                                                                                                                                                                                                                                              
>  555                         fde->fde_flags = (fde->fde_flags & ~UF_EXCLOSE) |                                                                                                                                                                                                                                                
>  556                             ((arg & FD_CLOEXEC) != 0 ? UF_EXCLOSE : 0) |                                                                                                                                                                                                                                                 
>  557                             ((arg & FD_CLOFORK) != 0 ? UF_FOCLOSE : 0) |                                                                                                                                                                                                                                                 
>  558                             ((arg & FD_RESOLVE_BENEATH) != 0 ?                                                                                                                                                                                                                                                           
>  559                             UF_RESOLVE_BENEATH : 0);                                                                                                                                                                                                                                                                     
>  560                         error = 0;                                                                                                                                                                                                                                                                                       
>  561                 }
> ```
> 
> It ORs the new flags with all the ones already set on the fd, except UF_EXCLOSE. In fact, I think it cannot be any other way, since otherwise the addition of FD_CLOFORK would break backwards compatibility, as this patch demonstrates.

You're right.  UF_RESOLVE_BENEATH was added in the middle of the previous PR.  Will drop the second commit as it doesn't make sense.

ricardobranco777: Dropped the second commit.

ricardobranco777: Dropped the second commit.  We can merge this to make mkostemp(3) support O_CLOFORK.

markjdb (approved):

kostikbel (approved):
But I still think (insist even) that we need two more ops on descriptor flags: atomic OR to set some flags, and atomic NAND to clear some flags.

ricardobranco777:
I don't plan to add any further changes to this PR as I want to make the flag available on other BSD's, so feel free to merge as is.

markjdb:
> But I still think (insist even) that we need two more ops on descriptor flags: atomic OR to set some flags, and atomic NAND to clear some flags.

That sounds reasonable to me, though F_SETFD is now a de factor OR operation for everything except FD_CLOEXEC.

markjdb:
Merged, thank you.

kostikbel:
> > But I still think (insist even) that we need two more ops on descriptor flags: atomic OR to set some flags, and atomic NAND to clear some flags.
> 
> That sounds reasonable to me, though F_SETFD is now a de factor OR operation for everything except FD_CLOEXEC.

Which is bug in fact.

ricardobranco777:
I see you guys fixed the case the stickiness in https://reviews.freebsd.org/D51212

Following up in https://github.com/freebsd/freebsd-src/pull/1766

PR #1761, opened 2025-07-09
tests/ci: add /usr/local/{sbin,bin} to freebsdci rc PATH

svmhdvn:
Currently, a lot of tests report 'skipped' due to missing binaries in the PATH. The real issue is that /etc/rc forcibly restricts the PATH to the base system only.

This patch re-enables a large chunk of skipped tests by adding the missing LOCALBASE directories to the PATH so that Kyua can discover third-party packages. It also fixes some minor rc scripting style as per the official freebsd scripting guide[0].

[kyua-reports.tar.gz](https://github.com/user-attachments/files/21132840/kyua-reports.tar.gz)

This attached diff between kyua runs shows that we skip 128 fewer tests with this change (and reveal one new broken test).

[0] https://docs.freebsd.org/en/articles/rc-scripting

### DIFF
--- a/tests/ci/tools/freebsdci
+++ b/tests/ci/tools/freebsdci
@@ -25,9 +25,6 @@
 
 . /etc/rc.subr
 
-: ${freebsdci_enable:="NO"}
-: ${freebsdci_type:="full"}
-
 name="freebsdci"
 desc="Run FreeBSD CI"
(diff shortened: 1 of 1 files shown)


svmhdvn:
Superseded by https://reviews.freebsd.org/D51263

PR #1762, opened 2025-07-09
sbin/pfctl/tests: use require.kmods instead of manual check for pf

svmhdvn:
@kprovost


kprovost:
Committed as 68fe0d9cc03bd80f63a5317a633d2426ae286316

Thanks!

PR #1764, opened 2025-07-12
devd/moused.conf: Stop moused quietly on destroy of a ums device

kostyaz:
If "moused" service is not enabled, devd action related to the "ums" driver will print an error message in the console.
Stopping it quietly avoids that message.

Example message:
```
Cannot 'stop' moused. Set moused_ums0_enable to YES in /etc/rc.conf or use 'onestop' instead of 'stop'.
```


bsdimp:

PR #1765, opened 2025-07-13
GEOM: allocate gp->name immediately after gp

wy-chung:
This is how it is done in g_new_providerf().


bsdimp (approved):
Don't think it matters much either way... 
But this change looks good to me.
and has the removal of g_free(gp->name).

The only reason not to do this would be for accounting for goems and looking for geom leaks...  But that's super hard to do and this won't make it any harder. We don't think there's any today, though...

bsdimp:

PR #1766, opened 2025-07-14
lib: Fix calls that naively set F_SETFD

ricardobranco777:
Fix calls that naively set F_SETFD

Ticket: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=286843


lib/libc/gen/fdopendir.c
```diff
@@ -48,8 +48,14 @@
...
 {
+	int flags;
 
-	if (_fcntl(fd, F_SETFD, FD_CLOEXEC) == -1)
+	flags = _fcntl(fd, F_GETFD, 0);
+	if (flags == -1)
+		return (NULL);
+
+	if ((flags & FD_CLOEXEC) == 0 &&
+	    _fcntl(fd, F_SETFD, flags | FD_CLOEXEC) == -1)
```
kostikbel: Please move actions out of the if () conditions.  I.e. something like
```
  if ((flags & FD_CLOEXEC) == 0) {
      res = _fcntl(fd, F_SETFD, flags | FD_CLOEXEC);
      if (res == -1)
          return (NULL);
   }
   return (__opendir_common(...));
```
and so on for all places you change.

ricardobranco777: Fixed.

lib/libc/stdio/fdopen.c
```diff
@@ -76,9 +76,15 @@ fdopen(int fd, const char *mode)
 	if ((fp = __sfp()) == NULL)
 		return (NULL);
 
-	if ((oflags & O_CLOEXEC) && _fcntl(fd, F_SETFD, FD_CLOEXEC) == -1) {
-		fp->_flags = 0;
-		return (NULL);
+	if ((oflags & O_CLOEXEC) != 0) {
+		tmp = _fcntl(fd, F_GETFD, 0);
+		if (tmp != -1 && (tmp & FD_CLOEXEC) == 0) {
```
kostikbel:
```
if (tmp == -1)
   return (NULL);
if ((tmp & FD_CLOEXEC) == 0) {
  ....
```

lib/libc/stdio/freopen.c
```diff
@@ -113,8 +113,11 @@ freopen(const char * __restrict file, const char * __restrict mode,
 			(void) ftruncate(fp->_file, (off_t)0);
 		if (!(oflags & O_APPEND))
 			(void) _sseek(fp, (fpos_t)0, SEEK_SET);
-		if (oflags & O_CLOEXEC)
-			(void) _fcntl(fp->_file, F_SETFD, FD_CLOEXEC);
+		if ((oflags & O_CLOEXEC) != 0) {
+			fdflags = _fcntl(fp->_file, F_GETFD, 0);
+			if (fdflags != -1 && (fdflags & FD_CLOEXEC) == 0)
+				(void) _fcntl(fp->_file, F_SETFD, fdflags | FD_CLOEXEC);
```
kostikbel: The line is too long.

BTW, why is freopen() ignores errors from anything it uses?

lib/libfetch/common.c
```diff
@@ -277,13 +277,16 @@ conn_t *
...
 
 	/* allocate and fill connection structure */
 	if ((conn = calloc(1, sizeof(*conn))) == NULL)
 		return (NULL);
-	fcntl(sd, F_SETFD, FD_CLOEXEC);
-	setsockopt(sd, SOL_SOCKET, SO_NOSIGPIPE, &opt, sizeof opt);
+	flags = fcntl(sd, F_GETFD);
+	if (flags != -1 && (flags & FD_CLOEXEC) == 0)
+		(void)fcntl(sd, F_SETFD, flags | FD_CLOEXEC);
+	(void)setsockopt(sd, SOL_SOCKET, SO_NOSIGPIPE, &opt, sizeof opt);
```
kostikbel:
```
(void)setsockopt(sd, SOL_SOCKET, SO_NOSIGPIPE, &opt, sizeof(opt));
```
since you are modifying the line anyway

lib/libc/stdio/fdopen.c
ricardobranco777: Fixed.

lib/libc/stdio/freopen.c
ricardobranco777: Fixed.

Good question :)

lib/libfetch/common.c
ricardobranco777: Fixed.

kostikbel (approved):

lib/libc/stdio/fdopen.c
```diff
@@ -76,9 +76,17 @@ fdopen(int fd, const char *mode)
 	if ((fp = __sfp()) == NULL)
 		return (NULL);
 
-	if ((oflags & O_CLOEXEC) && _fcntl(fd, F_SETFD, FD_CLOEXEC) == -1) {
-		fp->_flags = 0;
-		return (NULL);
+	if ((oflags & O_CLOEXEC) != 0) {
+		tmp = _fcntl(fd, F_GETFD, 0);
+		if (tmp == -1)
+			return (NULL);
```
markjdb, line 83: I think you need to set `fp->_flags = 0` in the first error path too.

ricardobranco777: > I think you need to set `fp->_flags = 0` in the first error path too.

Fixed.

markjdb (approved):
The patch is ok, but the commit message needs to provide more detail, i.e., explain why it's now necessary to fetch existing flags first.

ricardobranco777:
> The patch is ok, but the commit message needs to provide more detail, i.e., explain why it's now necessary to fetch existing flags first.

As I see you assigned yourself, I'm assuming you're taking care of it. :)

markjdb:
> > The patch is ok, but the commit message needs to provide more detail, i.e., explain why it's now necessary to fetch existing flags first.
> 
> As I see you assigned yourself, I'm assuming you're taking care of it. :)

I prefer that you update the commit message. I assigned it in anticipation of merging the commit soon, I'm running tests etc..

ricardobranco777:
> > > The patch is ok, but the commit message needs to provide more detail, i.e., explain why it's now necessary to fetch existing flags first.
> > 
> > 
> > As I see you assigned yourself, I'm assuming you're taking care of it. :)
> 
> I prefer that you update the commit message. I assigned it in anticipation of merging the commit soon, I'm running tests etc..

Fixed.

PR #1769, opened 2025-07-16
getfmac.8: add missing cross-references

Defenso-QTH:
Bring getfmac.8 up to par with getpmac.8 by adding cross-references.

This can help orient users that are getting started with the MAC framework and browsing the docs.

### DIFF
--- a/usr.sbin/getfmac/getfmac.8
+++ b/usr.sbin/getfmac/getfmac.8
@@ -51,5 +51,8 @@ specified files.
 .Xr mac 3 ,
 .Xr mac_get_file 3 ,
 .Xr mac 4 ,
+.Xr maclabel 7 ,
+.Xr getpmac 8 ,
 .Xr setfmac 8 ,
+.Xr setpmac 8 ,
 .Xr mac 9


concussious:
The problem with adding more, is that if the user doesn't know then too many options doesn't help.

AIUI, the best way is to for the subtopics to just to the topic overview page or pages that are a reasonable followup from there, and then the topic overview page has all the relevant links.

For this type of intent, I recommend looking at mac manuals holistically, here is an example:

https://github.com/freebsd/freebsd-src/commit/a516440afdd342fef9925680d890117d6d9024ed

Defenso-QTH:
I get that. I missed that perspective. Thanks for the feedback. So maybe it would be better to try and orient the user to mac(4).

mac(9) seems superfluous here, then, since it drives into development considerations that are not really useful for a user. 

And then we should trim down the references in pages of other utilities in section 8.

One beef I have with this set of docs is that they are inconsistent in referencing either users or processes or both when it comes to subjects, making it unnecessarily difficult for beginners to understand how it works and how to achieve what they want.

For instance, in mac(4), processes are left out of the table and not mentioned as subjects, although they are in the MAC Labels sections - where users aren't. maclabel(7) cites users and sockets as subjects but not processes.

Also maclabel(7) does not reference mac(4) although that is probably the top of the pyramid, that should be referenced everywhere.

Would you see value in me proposing a more complete patch that would address all these different points at once?

concussious:
Happily! So (and I mess this up all the time right) the best form is to take one logical change per patch. Then:

1. You don't burn out midway through and ambitious project because you have demonstratable progress
2. It's easier for people to review because they only have one concept to keep track of
3. You don't get into a corner where one idea is controversial and the other idea is not so we're stuck.

So, I would start with just organizing the flow of cross-references. That is already a big task which is going to take a lot of analysis and help from others. Then hopefully a long the way you will discover why the manuals refer to processes vs users. Manuals are usually written by the people who architects the software, so there's always a deep reason why they chose something, even though it often could use some editorial work.

> mac(9) seems superfluous here, then, since it drives into development considerations that are not really useful for a user.

So for reference doc like the system reference manual, it's really important to consider that the manual for *all* the use cases equally, and needs to be structured in a way that's useful for everybody.

Section 9 is the kernel programming manual. So, the reference manual is also for kernel programmers that need to use the Mac framework for whatever they're building.

All in all, there is a lot of room to help with manuals, but this is a 99/1 analysis vs writing activity. So rewarding though!

bsdimp:

PR #1770, opened 2025-07-16
sockstat: Convert to use libxo and document its usage

daminrido139:
Sponsored by:   Google, LLC (GSoC 2025)
MFC after:      2 weeks

### DIFF
--- a/usr.bin/sockstat/sockstat.1
+++ b/usr.bin/sockstat/sockstat.1
@@ -25,7 +25,7 @@
 .\" (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
 .\" THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 .\"
-.Dd June 30, 2025
+.Dd July 17, 2025
 .Dt SOCKSTAT 1
 .Os
 .Sh NAME
@@ -33,6 +33,7 @@
 .Nd list open sockets
 .Sh SYNOPSIS
 .Nm
+.Op Fl -libxo
 .Op Fl 46ACcfIiLlnqSsUuvw
 .Op Fl j Ar jail
 .Op Fl p Ar ports
@@ -46,6 +47,13 @@ domain sockets.
 .Pp
 The following options are available:
 .Bl -tag -width Fl
+.It Fl -libxo
+Generate output via
+.Xr libxo 3
+in a selection of different human and machine readable formats.
+See
+.Xr xo_options 7
+for details on command line arguments.
 .It Fl 4
 Show
 .Dv AF_INET
@@ -229,6 +237,11 @@ Show TCP IPv6 sockets which are listening and connected (default):
 .Bd -literal -offset indent
 $ sockstat -6 -P tcp
 .Ed
+.Pp
+Show all sockets in JSON format with neat alignment:
+.Bd -literal -offset indent
+$ sockstat --libxo json,pretty
+.Ed
 .Sh SEE ALSO
 .Xr fstat 1 ,
 .Xr netstat 1 ,
@@ -237,6 +250,8 @@ $ sockstat -6 -P tcp
 .Xr inet 4 ,
 .Xr inet6 4 ,
 .Xr protocols 5
+.Xr libxo 3 ,
+.Xr xo_options 7
 .Sh HISTORY
 The
 .Nm
--- a/usr.bin/sockstat/sockstat.c
+++ b/usr.bin/sockstat/sockstat.c
@@ -55,7 +55,6 @@
 
 #include <capsicum_helpers.h>
 #include <ctype.h>
-#include <err.h>
 #include <errno.h>
 #include <inttypes.h>
 #include <jail.h>
@@ -67,6 +66,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <libxo/xo.h>
 
 #include <libcasper.h>
 #include <casper/cap_net.h>
@@ -74,6 +74,7 @@
 #include <casper/cap_pwd.h>
 #include <casper/cap_sysctl.h>
 
+#define SOCKSTAT_XO_VERSION "1"
 #define	sstosin(ss)	((struct sockaddr_in *)(ss))
 #define	sstosin6(ss)	((struct sockaddr_in6 *)(ss))
 #define	sstosun(ss)	((struct sockaddr_un *)(ss))
@@ -197,7 +198,7 @@ static bool
 _check_ksize(size_t received_size, size_t expected_size, const char *struct_name)
 {
 	if (received_size != expected_size) {
-		warnx("%s size mismatch: expected %zd, received %zd",
+		xo_warnx("%s size mismatch: expected %zd, received %zd",
 		    struct_name, expected_size, received_size);
 		return false;
 	}
@@ -209,7 +210,7 @@ static void
 _enforce_ksize(size_t received_size, size_t expected_size, const char *struct_name)
 {
 	if (received_size != expected_size) {
-		errx(1, "fatal: struct %s size mismatch: expected %zd, received %zd",
+		xo_errx(1, "fatal: struct %s size mismatch: expected %zd, received %zd",
 		    struct_name, expected_size, received_size);
 	}
 }
@@ -227,7 +228,7 @@ get_proto_type(const char *proto)
 	else
 		pent = getprotobyname(proto);
 	if (pent == NULL) {
-		warn("cap_getprotobyname");
+		xo_warn("cap_getprotobyname");
 		return (-1);
 	}
 	return (pent->p_proto);
@@ -248,7 +249,7 @@ init_protos(int num)
 	}
 
 	if ((protos = malloc(sizeof(int) * proto_count)) == NULL)
-		err(1, "malloc");
+		xo_err(1, "malloc");
 	numprotos = proto_count;
 }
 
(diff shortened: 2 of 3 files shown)


daminrido139:
Kindly review the changes when you get a chance, @asomers .

usr.bin/sockstat/sockstat.1
```diff
@@ -46,6 +47,13 @@ domain sockets.
 .Pp
 The following options are available:
 .Bl -tag -width Fl
+.It Fl -libxo
+Generate output via
+.Xr libxo 3
+in a selection of different human and machine readable formats.
+See
+.Xr xo_parse_args 3
```
asomers: Please link to xo_options.7 instead.  I'm going to make that change globally in https://reviews.freebsd.org/D51354 .

usr.bin/sockstat/sockstat.1
```diff
@@ -237,6 +250,8 @@ $ sockstat -6 -P tcp
 .Xr inet 4 ,
 .Xr inet6 4 ,
 .Xr protocols 5
+.Xr libxo 3 ,
```
asomers, line 253: We probably don't need to link libxo.3 in a section 1 man page.  I think that xo_options.7 should be sufficient.

usr.bin/sockstat/sockstat.c
```diff
@@ -1350,40 +1351,57 @@ display_sock(struct sock *s, struct col_widths *cw, char *buf, size_t bufsize)
...
-				cw->foreign_addr, buf);
+			if (xo_get_style(NULL) == XO_STYLE_TEXT)
+				xo_emit(" {:foreign-address/%-*.*s}",
+				cw->foreign_addr, cw->foreign_addr, buf);
+			else
+				xo_emit("{:foreign-address/%s}", buf);
+		}
+		if (opt_A) {
+			snprintf(buf, bufsize, "%#*" PRIx64,
+				cw->pcb_kva, s->pcb);
```
asomers, line 1451: Will this result in the field being space-padded even in JSON output?

usr.bin/sockstat/sockstat.c
```diff
@@ -1414,69 +1434,75 @@ display_sock(struct sock *s, struct col_widths *cw, char *buf, size_t bufsize)
...
+						xo_emit(" {:path-state/%-*s}",
+							cw->path_state,
+							sctp_path_state(
+								faddr->state));
 						break;
 					case IPPROTO_TCP:
 						if (s->state >= 0 &&
 							s->state < TCP_NSTATES)
-							printf(" %-*s",
+						    xo_emit(" {:conn-state/%-*s}",
```
asomers: Looks like some spaces/tabs problems here.

usr.bin/sockstat/sockstat.c
```diff
@@ -1414,69 +1434,75 @@ display_sock(struct sock *s, struct col_widths *cw, char *buf, size_t bufsize)
...
 		}
 		if (laddr != NULL)
 			laddr = laddr->next;
 		if (faddr != NULL)
 			faddr = faddr->next;
 		if (laddr != NULL || faddr != NULL)
-			printf("%-*s %-*s %-*s %-*s %-*s", cw->user, "",
-				cw->command, "", cw->pid, "", cw->fd, "",
-				cw->proto, "");
+			xo_emit("{:uid/%-*s} {:command/%-*s} {:pid/%*s} "
```
asomers: The "??" fields are nice for human-readable output.  But for machine-readable output I don't think they make sense.  Better to simply not print those fields, I think.

usr.bin/sockstat/sockstat.c
```diff
@@ -1550,37 +1582,50 @@ display(void)
...
 	if (opt_j >= 0)
 		return;
 	SLIST_FOREACH(s, &nosocks, socket_list) {
 		if (!check_ports(s))
 			continue;
-		printf("%-*s %-*s %*s %*s", cw.user, "??", cw.command, "??",
-			cw.pid, "??", cw.fd, "??");
+		xo_open_instance("socket");
+		xo_emit("{:uid/%-*s} {:command/%-*s} {:pid/%*s} {:fd/%*s}",
+			cw.user, "??", cw.command, "??", cw.pid, "??",
```
asomers: Here, too, I think you should omit the "??" for machine-readable output.

asomers (changes requested):
This looks like a good start, though I haven't tested it yet.

daminrido139:
Hi @asomers ,
I've replaced xo_parse_args.3 with xo_options.7, and now the fields with "??" or "(not connected)" are completely omitted for machine-readable output,

usr.bin/sockstat/sockstat.c
daminrido139: No, there won’t be any padding. When xo_get_style(NULL) is JSON, XML, or HTML, no width calculation is performed, and the column width is set to zero.

daminrido139:
Hey @asomers  , I just wanted to follow up in case this may have been missed in notifications. No rush, I’d appreciate your feedback whenever you have a moment.

usr.bin/sockstat/sockstat.c
```diff
@@ -1657,6 +1730,10 @@ main(int argc, char *argv[])
 	int protos_defined = -1;
 	int o, i;
 
+	xo_set_version(SOCKSTAT_XO_VERSION);
```
asomers: xo_set_version needs to come after xo_parse_args.  Otherwise the version number will never be printed.

usr.bin/sockstat/sockstat.c
```diff
@@ -1396,14 +1434,20 @@ display_sock(struct sock *s, struct col_widths *cw, char *buf, size_t bufsize)
...
 			if (s->proto == IPPROTO_TCP || s->proto == IPPROTO_UDP)
-				printf(" %*" PRIu64, cw->inp_gencnt,
+			{
+				snprintf(buf, bufsize, "%" PRIu64,
 					s->inp_gencnt);
-			else
-				printf(" %*s", cw->inp_gencnt, "??");
+				xo_emit(" {:id/%*s}", cw->inp_gencnt, buf);
+			}
+			else if (is_text_style)
```
asomers: The standard style is to cuddle the "else" with the "}" on the previous line.

daminrido139:
Hi @asomers ,
I believe the libxo output is now more machine-readable and consistent. Kindly review it when you're free.

usr.bin/sockstat/sockstat.c
```diff
@@ -1299,6 +1320,7 @@ calculate_column_widths(struct col_widths *cw)
 	struct sock *s;
 	struct passwd *pwd;
 
+	cap_setpassent(cappwd, 1);
```
asomers, line 1335: Why did you add this part?

usr.bin/sockstat/sockstat.c
```diff
@@ -1138,10 +1153,17 @@ format_unix_faddr(struct addr *faddr, char *buf, size_t bufsize) {
 				pos += snprintf(SAFEBUF, SAFESIZE,
```
asomers: All of these snprintf calls aren't actually useful in json mode.  Can you conditionalize them?

usr.bin/sockstat/sockstat.c
```diff
@@ -1116,6 +1126,10 @@ format_unix_faddr(struct addr *faddr, char *buf, size_t bufsize) {
 			if (f != NULL) {
 				pos += snprintf(SAFEBUF, SAFESIZE, "[%lu %d]",
 					(u_long)f->xf_pid, f->xf_fd);
+				if (!is_text_style) {
+				    xo_emit("{:pid/%lu}", (u_long)f->xf_pid);
+				    xo_emit("{:fd/%d}", f->xf_fd);
```
asomers: Style: spaces to tabs

asomers: Also, for consistency's sake you should add a `xo_open_list` here, even though it will always have only one element.  That will make it easier for consumers to parse .

usr.bin/sockstat/sockstat.c
```diff
@@ -1345,65 +1368,104 @@ display_sock(struct sock *s, struct col_widths *cw, char *buf, size_t bufsize)
...
+				if (is_text_style) {
+					xo_emit(" {:/%-*.*s}", cw->local_addr,
+						cw->local_addr, buf);
+				}
+				xo_close_container("local");
+			} else if (is_text_style)
+				xo_emit(" {:/%-*.*s}", cw->local_addr,
+					cw->local_addr, "??");
+			if (faddr != NULL) {
+				xo_open_container("foreign");
```
asomers, line 1437: Also here, for consistency's sake, there should be a `xo_open_container("connections")`.

asomers (changes requested):
It looks a lot better now.  My main concern is now the "connections" container.  You currently only output it for sockets where there is a possibility of more than one connection.  But I think it would be easier for users to parse our json if it were output all of the time.  What do you think?

Also, I just noticed that sequential packet sockets will be displayed as `seqpac` for short.  In machine-readable mode, could you lengthen that to `seqpacket`?

usr.bin/sockstat/sockstat.c
daminrido139: I just realized I had missed adding `cap_setpassent(cappwd, 1);`  in my earlier commits. It's used in the display() function. Looks like it’s used as an optimization for resolving usernames. If you think it’s better to keep this in a separate commit, I can revert this one and add it separately.

daminrido139:
Hi @asomers, I’ve made the changes. Could you please test it?

bsdimp:
Do you get any warnings with --libxo:XPW or --libxo:JPW?

daminrido139:
Hey @bsdimp , thanks for pointing that out! I tested with `--libxo:JPW` and `--libxo:XPW`,  and it did reveal some warnings earlier, caused by emitting fields in the middle of the `connections` list. That’s been fixed in the latest commit. No more warnings now!

usr.bin/sockstat/sockstat.c
```diff
@@ -1136,25 +1135,27 @@ format_unix_faddr(struct addr *faddr, char *buf, size_t bufsize) {
...
+					xo_close_list("connections");
 				}
 			}
 		} else
 			pos += formataddr(&p->laddr->address,
 				SAFEBUF, SAFESIZE);
-	}
-	/* Remote peer(s) connect(2)ed to us, if any. */
-	if (faddr->firstref != 0) {
+	} else if (faddr->firstref != 0) {
```
daminrido139, line 1149: Introduce `else if` here because a socket can either be connected or listening, but not both. This makes the logic clearer and avoids confusion for future readers.

asomers:
One other thing.  When adding new features like this, you should put "Relnotes: yes" in the commit message.  I can do that for you when I rebase.

daminrido139:
@asomers Noted, thanks!

daminrido139:
Hi @asomers , just checking in, would it be possible to merge this when you get a chance? The GSoC final submission period is going to begin soon. Thanks!

asomers:
> Hi @asomers , just checking in, would it be possible to merge this when you get a chance? The GSoC final submission period is going to begin soon. Thanks!

@daminrido139 you caught me traveling again.  I don't think I'll be able to merge this before Wednesday.

daminrido139:
@asomers Got it, no rush. Enjoy your travels and have a safe journey :)

PR #1773, opened 2025-07-17
kldstat: Memory zeroing using memset()

nazarenkomykyta:
- Using memset() instead of the outdated bzero() function when zeroing memory
- bzero() has been removed since the POSIX.1-2001 standard


concussious:
Why?

ghost:
> Why?

Because it's the right thing to do <3

asomers:
> Because it's the right thing to do <3

Again, why?

bsdimp:
Both the bzero and memset interfaces are correct here. While memset is standardized, and bzero has been dropped from POSIX.1, FreeBSD generally doesn't proactively change code like this without some reason to do so: other changes in the code, there's something wrong with the current code, etc.

This is the only instance of bzero in the kld* progams. The commit looks OK, but the message might be better like
```
Prefer the standard memset interface to the deprectated bzero interface.
```
which says the same thing, but focuses more on why with a specific reason. One reason we sometimes allow simpler changes like this is for people to learn the process.

ghost:
> Both the bzero and memset interfaces are correct here. While memset is standardized, and bzero has been dropped from POSIX.1, FreeBSD generally doesn't proactively change code like this without some reason to do so: other changes in the code, there's something wrong with the current code, etc.
> 
> This is the only instance of bzero in the kld* progams. The commit looks OK, but the message might be better like
> 
> ```
> Prefer the standard memset interface to the deprectated bzero interface.
> ```
> 
> which says the same thing, but focuses more on why with a specific reason. One reason we sometimes allow simpler changes like this is for people to learn the process.

I agree that the description could have been better. But the fact remains that besides the fact that memset() is a standardized function, it is more universal and portable, although in this case it is not required.

bsdimp:
> I agree that the description could have been better. But the fact remains that besides the fact that memset() is a standardized function, it is more universal and portable, although in this case it is not required.

Great! If you update the commit message, I'm incline to go ahead and land this under the 'learning how to do it' exception because the code is correct before and after.

concussious:
Please mention dropped from posix.1 for the rest of the class.

PR #1774, opened 2025-07-17
kldstat: Memory zeroing using memset()

nazarenkomykyta:
- Using memset() instead of the outdated bzero() function when zeroing memory
- bzero() has been removed since the POSIX.1-2001 standard

### DIFF
--- a/sbin/kldstat/kldstat.c
+++ b/sbin/kldstat/kldstat.c
@@ -51,7 +51,7 @@ printmod(int modid)
 {
 	struct module_stat stat;
 
-	bzero(&stat, sizeof(stat));
+	memset(&stat, 0, sizeof(stat));
 	stat.version = sizeof(struct module_stat);
 	if (modstat(modid, &stat) < 0) {
 		warn("can't stat module id %d", modid);


concussious:
You don't have to open a new pull, just force push your branch and the pull will update. Then we don't lose the context/reviewers/pull increments/discussion.

ghost:
> You don't have to open a new pull, just force push your branch and the pull will update. Then we don't lose the context/reviewers/pull increments/discussion.

I did so, but accidentally closed the pull request, so I don't think it's a problem.

concussious:
Oops sorry, I hit it on my phone by accident while trying to type >.<

No it's not a problem, everyone does it when they start out. I actually did it 3 times on the same patch. But for the future, you can just reopen it.

bsdimp (approved):

bsdimp:
```
/usr/home/imp/git/head/sbin/kldstat/kldstat.c:54:2: error: call to undeclared library function 'memset' with type 'void *(void *, int, unsigned long)'; ISO C99 and later do not support implicit function declarations [-Werror,-Wimplicit-function-declaration]
   54 |         memset(&stat, 0, sizeof(stat));
      |         ^
/usr/home/imp/git/head/sbin/kldstat/kldstat.c:54:2: note: include the header <string.h> or explicitly provide a declaration for 'memset'
1 error generated
```
when I tried to build it...  So was this tested?

ghost:
> ```
> /usr/home/imp/git/head/sbin/kldstat/kldstat.c:54:2: error: call to undeclared library function 'memset' with type 'void *(void *, int, unsigned long)'; ISO C99 and later do not support implicit function declarations [-Werror,-Wimplicit-function-declaration]
>    54 |         memset(&stat, 0, sizeof(stat));
>       |         ^
> /usr/home/imp/git/head/sbin/kldstat/kldstat.c:54:2: note: include the header <string.h> or explicitly provide a declaration for 'memset'
> 1 error generated
> ```
> 
> when I tried to build it... So was this tested?

As we said earlier, the bzero() function is deprecated, which is why it is in the “strings.h” header. I forgot that I need to include the “string.h” header that contains memset(), and remove "strings.h" header.

bsdimp:
> > ```
> > /usr/home/imp/git/head/sbin/kldstat/kldstat.c:54:2: error: call to undeclared library function 'memset' with type 'void *(void *, int, unsigned long)'; ISO C99 and later do not support implicit function declarations [-Werror,-Wimplicit-function-declaration]
> >    54 |         memset(&stat, 0, sizeof(stat));
> >       |         ^
> > /usr/home/imp/git/head/sbin/kldstat/kldstat.c:54:2: note: include the header <string.h> or explicitly provide a declaration for 'memset'
> > 1 error generated
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > when I tried to build it... So was this tested?
> 
> As we said earlier, the bzero() function is deprecated, which is why it is in the “strings.h” header. I forgot that I need to include the “string.h” header that contains memset(), and remove "strings.h" header.

Yea, I've been burned by that... but it's been long enough I'd forgotten to check that detail.   Please fix 😀

gmshake:
I do not have strong option for this change. If you insist, I'd recommend choosing a better reason to do so, say `Prefer memset() over bzero()`

Quoted from the manpage of bzero(3),

```
    IEEE Std 1003.1-2008 (“POSIX.1”) removes the specification of bzero() and
     it is marked as LEGACY in IEEE Std 1003.1-2004 (“POSIX.1”).  For
     portability with other systems new programs should use memset(3).
```

ghost:
> Please fix 😀

Fixed.

PR #1775, opened 2025-07-18
g_concat_add_disk: should set an error number here

wy-chung:
Otherwise the error will be 0.


bsdimp (approved):
This looks good to my eye. It treats the error like a read error, which I believe is appropriate because we don't use the as if we'd gotten a disk error.

bsdimp:
Oh, I just looked at the commit messagae. It's a bit incomplete. It should include why it should be set:
```
Return EINVAL when the metadata is invalid for an added disk. We don't use the disk and stop using it right afterwards.
The user should get an error indication, just like they would if there had been a disk read error.
```

wy-chung:
The commit message added.

bsdimp:

PR #1776, opened 2025-07-18
add thermal monitoring driver for intel cpu thermal sensors

Phyninite:
sys/dev/thermal/thermal.c (new file)
sys/dev/thermal/thermal.h (new file, empty)
sys/conf/files.amd64 (modified, +1 line)
sys/amd64/conf/GENERIC (modified, +1 line)

total: 2 new files, 2 modified files.


sys/dev/thermal/thermal.c
```diff
@@ -0,0 +1,213 @@
+#include <sys/param.h>
```
bsdimp: This file needs a copyright + license header.

Also, systm.h includes sys/param.h, so you don't need to include both.

sys/dev/thermal/thermal.h
```diff
@@ -0,0 +1 @@
+
```
bsdimp: If there's nothing in this file, there's no need for it.

sys/dev/thermal/thermal.c
```diff
@@ -0,0 +1,213 @@
...
+    len = snprintf(buf, sizeof(buf), "temp: %d throttle: %d\n", 
+                   sc->current_temp, sc->throttle_count);
+    
+    return uiomove(buf, min(len, uio->uio_resid), uio);
+}
+
+static int
+thermal_probe(device_t dev)
+{
+    if (!(cpu_feature & CPUID_ACPI))
```
bsdimp: Is this conclusive? What about AMD processors?

bsdimp:
There should be a man page for this as well, describing what it does, what the sysctl are and what the read interface to it is. Thermal throttling doesn't seem to be connected to anything, but I may have overlooked setting/clearing a bit that's written to some MSR somewhere.

sys/dev/thermal/thermal.c
ghost: Fixed in commit ff705aa: added BSD-2-Clause header and dropped the extra include.

sys/dev/thermal/thermal.h
ghost: File removed in commit a757374; contents folded into thermal.c.

sys/dev/thermal/thermal.c
ghost: Probe now gated on cpu_vendor == CPU_VENDOR_INTEL in commit ff705aa; AMD left for future patch.

bsdimp:
> This functionality is already provided by the coretemp(4) driver. Improvements should go there, there's no need to add another driver.

Yea, looking at that driver suggests that we just close this pull request since it's so much more functional and complete.

bsdimp:
Closing this out. It's not a useful direction for us to go. If you make improvements to coretemp(4), please open a new review with that.

PR #1778, opened 2025-07-20
reboot: Minor changes & refactoring

nazarenkomykyta:
- In add_env() env is checked for NULL before dereferencing.
- In main() function when checking the kernel file, dynamic memory allocation is replaced by a static buffer of size PATH_MAX, as if it will be much more efficient, because all allocation will take place on the stack, the whole thing will be more productive.

### DIFF
--- a/sbin/reboot/reboot.c
+++ b/sbin/reboot/reboot.c
@@ -111,6 +111,12 @@ zfsbootcfg(const char *pool, bool force)
 }
 
 static void
+unlink_tmp(char *tmp) {
+	if (unlink(tmp))
+		warn("unlink %s", tmp);
+}
+
+static void
 write_nextboot(const char *fn, const char *env, bool force)
 {
 	char tmp[PATH_MAX];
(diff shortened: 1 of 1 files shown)


bsdimp:
So there's two problem here, which is all of the change:
(1) The refactoring doesn't have a good basis. It introduces churn for no real benefit. Creating a function for the sake of saving two lines doesn't seem like good ROI and introduces the possibility of an error accidentally creaping in.
(2) The if (x) err; if (y) err -> if (x) err; else if (y) err;  actively goes against FreeBSD style.

ghost:
> So there's two problem here, which is all of the change: (1) The refactoring doesn't have a good basis. It introduces churn for no real benefit. Creating a function for the sake of saving two lines doesn't seem like good ROI and introduces the possibility of an error accidentally creaping in. (2) The if (x) err; if (y) err -> if (x) err; else if (y) err; actively goes against FreeBSD style.

Please reconsider this, thank you.

bsdimp:
So these changes aren't ready for a pull request. Generally, we require tested, mature changes. This review has gone from one thing to another with seemingly more random changes. The changes themselves don't look good (some even introduce regressions). I'm closing this. Please in the future have changes that fix actual bugs and don't just randomly change things to make them better.

ghost:
> So these changes aren't ready for a pull request. Generally, we require tested, mature changes. This review has gone from one thing to another with seemingly more random changes. The changes themselves don't look good (some even introduce regressions). I'm closing this. Please in the future have changes that fix actual bugs and don't just randomly change things to make them better.

I'm amazed at how resilient your nervous system is after my worst pull request.

PR #1780, opened 2025-07-21
tests/*: use require.kmods property instead of ad-hoc checks 

svmhdvn:
This patch series makes the reporting of required kmods uniform across all tests. It helps to identify skipped tests due to missing kmods in a CI test environment and declaratively signals test requirements.

Please see each commit in the series for a smaller per-subdir patch.

### DIFF
--- a/tests/sys/vmm/vmm_cred_jail.sh
+++ b/tests/sys/vmm/vmm_cred_jail.sh
@@ -35,12 +35,10 @@ vmm_cred_jail_host_head()
 {
 	atf_set "descr" "Tests deleting the host's VM from within a jail"
 	atf_set "require.user" "root"
+	atf_set "require.kmods" "vmm"
 }
 vmm_cred_jail_host_body()
 {
-	if ! -c /dev/vmmctl; then
-		atf_skip "vmm is not loaded"
-	fi
 	bhyvectl --vm=testvm --create
 	vmm_mkjail myjail
 	atf_check -s exit:1 -e ignore jexec myjail bhyvectl --vm=testvm --destroy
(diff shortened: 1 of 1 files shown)


tests/sys/aio/aio_test.c
```diff
@@ -469,37 +467,61 @@ aio_file_test(completion comp, struct sigevent *sev, bool vectored)
 	close(fd);
 }
 
-ATF_TC_WITHOUT_HEAD(file_kq);
+ATF_TC(file_kq);
+ATF_TC_HEAD(file_kq, tc)
+{
+	atf_tc_set_md_var(tc, "require.kmods", "aio");
```
asomers: Are there any test cases in this file that _don't_ require the aio kernel module?  It might be easier to add the requirement in the Makefile, if not.

tests/sys/aio/lio_test.c
```diff
@@ -71,7 +74,6 @@ ATF_TC_BODY(lio_listio_eagain_kevent, tc)
 	const char *path="tempfile";
 	void *udata[2];
 
-	ATF_REQUIRE_KERNEL_MODULE("aio");
```
asomers, line 74: This may be the only test case in this file that explicitly required aio.  But all of the others require it too.  The requirement was just missing.  You should put the requirement in the Makefile for the entire program instead of in the test case head.

svmhdvn:
Thanks! Applied feedback in this revision.

tests/sys/netpfil/common/dummynet.sh
```diff
@@ -265,10 +265,6 @@ queue_body()
 {
 	fw=$1
 
-	if [ $fw = "ipfw" ] && [ "$(atf_config_get ci false)" = "true" ]; then
```
kprovost: This (and the other changes like it) doesn't seem to belong in this commit.

I'm not at all convinced it's correct either, but even if it is, it needs to be in a separate commit and justified separately from the introduction of require.kmods.

svmhdvn: Thanks for that, I accidentally added these changes to the wrong branch. These are meant for a new PR which I will be submitting soon.

Removed in the latest revision.

svmhdvn:
Any updates on merging this? Is there anything left over?

bsdimp:

PR #1782, opened 2025-07-22
fusefs: report that all tests require the fusefs kernel module

asomers:
Previously the googletest tests would skip themselves if /dev/fuse could not be found.  But that information would not be passed to Kyua. Instead it would think that they had passed.  Also, the atf-sh test would previously fail if the fusefs module weren't loaded.  Now both tests will correctly report their requirements to Kyua.

Note that fusefs's googletest tests still require that the mac_bsdextended(4) module _not_ be loaded, but Kyua has no way to report such a requirement.

MFC after:	2 weeks
Sponsored by:	ConnectWise

### DIFF
--- a/tests/sys/fs/fusefs/Makefile
+++ b/tests/sys/fs/fusefs/Makefile
@@ -70,7 +70,8 @@ TEST_METADATA.nfs+=	required_user="root"
 TEST_METADATA.ctl+=	is_exclusive="true"
 TEST_METADATA.ctl+=	required_user="root"
 
-TEST_METADATA+= timeout=10
+TEST_METADATA+=	timeout=10
+TEST_METADATA+=	required_kmods="fusefs"
 
 FUSEFS=		${SRCTOP}/sys/fs/fuse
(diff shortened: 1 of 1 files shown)


asomers:
@svmhdvn review, please.

tests/sys/fs/fusefs/Makefile
```diff
@@ -70,7 +70,8 @@ TEST_METADATA.nfs+=	required_user="root"
 TEST_METADATA.ctl+=	is_exclusive="true"
 TEST_METADATA.ctl+=	required_user="root"
 
-TEST_METADATA+= timeout=10
+TEST_METADATA+=	timeout=10
```
emaste, line 73: is this just a whitespace change?

asomers: Yes.  It's a space -> tab, in order to match the rest of the Makefile.

svmhdvn:
LGTM, yes I'll make a similar change on https://github.com/freebsd/freebsd-src/pull/1780 as you mentioned to reduce the size of the changes.

PR #1783, opened 2025-07-22
mac_bsdextended: report that all tests require the kernel module

asomers:
Use the standard required_kmods reporting mechanism to notify Kyua of which kernel modules are required.

MFC after:	2 weeks
Sponsored by:	ConnectWise


asomers:
@svmhdvn review, please.

svmhdvn:
Tested locally with a before and after loading `mac_bsdextended`. LGTM, thanks!

PR #1785, opened 2025-07-23
Misc typos to libc/rpc

pgiffuni:
These are really minor cleanups.
No behavioural change intended.


lib/libc/rpc/getpublickey.c
```diff
@@ -41,8 +41,10 @@
 #include <pwd.h>
 #include <rpc/rpc.h>
 #include <rpc/key_prot.h>
+#ifdef YP
 #include <rpcsvc/yp_prot.h>
 #include <rpcsvc/ypclnt.h>
+#endif
```
bsdimp: How is this a typo?

include/rpc/des.h
```diff
@@ -61,7 +61,7 @@ struct desparams {
 /*
  * These ioctls are only implemented in SunOS. Maybe someday
  * if somebody writes a driver for DES hardware that works
- * with FreeBSD, we can being that back.
+ * with FreeBSD, we can bring that back.
```
bsdimp: DES is deprecated. Why should we fix a typo here as opposed to maybe deleting it entirely? FreeBSD at this point is not at all likely to add these IOCTLs, I don't think.

bsdimp: This needs to be rebased now. emaste removed all this code a few hours ago.

lib/libc/rpc/getpublickey.c
pgiffuni: Yeah not a typo, just a missing condition.
The individual commit had the right description though.

PR #1786, opened 2025-07-23
GEOM: add a new function g_new_geom

wy-chung:
This function accepts a regular string as its input string so the callers calling this function can save the time wasted on format string processing.

### DIFF
--- a/sys/geom/geom_subr.c
+++ b/sys/geom/geom_subr.c
@@ -368,20 +368,15 @@ g_retaste(struct g_class *mp)
 }
 
 struct g_geom *
-g_new_geomf(struct g_class *mp, const char *fmt, ...)
+g_new_geom(struct g_class *mp, const char *name)
 {
+	int len;
 	struct g_geom *gp;
-	va_list ap;
-	struct sbuf *sb;
 
 	g_topology_assert();
 	G_VALID_CLASS(mp);
-	sb = sbuf_new_auto();
-	va_start(ap, fmt);
-	sbuf_vprintf(sb, fmt, ap);
-	va_end(ap);
-	sbuf_finish(sb);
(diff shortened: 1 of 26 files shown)


markjdb:
We should avoid modifying openzfs code, since that comes from its own upstream.

There is a style bug in g_new_geom()--could you please fix it? With that, this commit is fine to merge.

sys/geom/geom_subr.c
```diff
@@ -404,6 +396,29 @@ g_new_geomf(struct g_class *mp, const char *fmt, ...)
...
+	struct g_geom *gp;
+	va_list ap;
+	struct sbuf *sb;
+
+	sb = sbuf_new_auto();
+	va_start(ap, fmt);
+	sbuf_vprintf(sb, fmt, ap);
+	va_end(ap);
+	sbuf_finish(sb);
+	gp = _g_new_geom(mp, sbuf_data(sb), sbuf_len(sb));
```
bsdjhb: I would just call `g_new_geom(mp, sbuf_data(sb))` here.  Avoiding the length check isn't that big of a deal IMO vs having simpler to read code.

sys/geom/geom_subr.c
```diff
@@ -367,30 +367,22 @@ g_retaste(struct g_class *mp)
 	return (error);
 }
 
-struct g_geom *
-g_new_geomf(struct g_class *mp, const char *fmt, ...)
+static struct g_geom *
+_g_new_geom(struct g_class *mp, const char *name, int len)
```
bsdjhb: Following my suggestion below, just have this be `g_new_geom()` without the length and just have `len` be a local variable initialized with `strlen(name)`.

bsdjhb:
The first commit here looks good to me.  Adding a second commit to convert existing callers that didn't need formatting as was done in the original PR would be nice to have as well.

wy-chung:
I have checked that this pull request has been merged to main so I close it.

PR #1787, opened 2025-07-23
GEOM: add a new function g_new_provider

wy-chung:
This is a variant of g_new_providerf. This function accepts a regular string instead of a format string as its input parameter. It can save the time wasted on unnecessary format string processing.


bsdjhb (approved):
Thanks

PR #1790, opened 2025-07-25
tests/mac_portacl: enable is_exclusive for now

svmhdvn:
Both tests rely on ports 77 and 7777 to be available and thus cannot be run concurrently. This is a temporary measure to ensure that they don't conflict with each other.

In the future, these should be rewritten to wait until the necessary ports are available, or deterministically select a free port instead.


lwhsu:
Merged as 5dadfac51c3876b0ba6fde831cb4417594167fed

PR #1793, opened 2025-07-27
GEOM: remove the redundant "if statement"

wy-chung:
The "if statement" executed before calling g_provider_by_name is redundant. That statement is also in g_provider_by_name.

### DIFF
--- a/sys/geom/concat/g_concat.c
+++ b/sys/geom/concat/g_concat.c
@@ -1107,8 +1107,6 @@ g_concat_ctl_append(struct gctl_req *req, struct g_class *mp)
 		gctl_error(req, "No 'arg%u' argument.", 1);
 		goto fail;
 	}
-	if (strncmp(name, "/dev/", strlen("/dev/")) == 0)
-		name += strlen("/dev/");
 	pp = g_provider_by_name(name);
 	if (pp == NULL) {
 		G_CONCAT_DEBUG(1, "Disk %s is invalid.", name);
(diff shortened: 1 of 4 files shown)


sys/geom/concat/g_concat.c
```diff
@@ -1107,8 +1107,6 @@ g_concat_ctl_append(struct gctl_req *req, struct g_class *mp)
 		gctl_error(req, "No 'arg%u' argument.", 1);
 		goto fail;
 	}
-	if (strncmp(name, "/dev/", strlen("/dev/")) == 0)
-		name += strlen("/dev/");
```
ehem, line 1111: This loses the effect of `name` being changed.  Notice `name` is potentially used in several more places and this loses that effect.

ehem (changes requested):
In all of these cases the variable `name` (`pname` in one case) is being modified.  This is mostly used for later error messages, but the effect is lost with this commit.

Likely it is a good idea to merge these duplicate removals of "/dev/" from the strings, but this doesn't work.  I wonder whether all other callers of `g_provider_by_name()` have already stripped "/dev/" from the string.  If so it might be appropriate to have a function which takes a `char **` and modifies the argument to strip the prefix and then remove this from `g_provider_by_name()`.

sys/geom/concat/g_concat.c
wy-chung: I agree. Should I just close this issue?

bsdimp: This change is good. We use name twice. g_provider_by_name does the right thing. As does g_concat_find_disk. The latter should be recoded in the same style as the former, but this code can be removed here. The error message changing as a result are of no consequence.

I'd honestly do the change to g_concat_find_disk in a new pull request, since I'ld like to land this one.

bsdimp:
A useful cleanup may be to centralize this function in one place (likely as a static inline). But that's a lot more churn.

bsdimp:

PR #1794, opened 2025-07-28
ctld: Use modern C++ for many data structures and code paths

bsdjhb:
I have further changes stacked on top of this that also add NVMe support.  However, this is already a pretty big chunk of changes to review on its own.  Given the size of the stack here, I don't think phabricator can review this well, so posting for review here.


bsdjhb:
Some offline feedback I've received is that I should probably write some manpages for libutil++.  It's not clear to me how well mdoc will handle C++ things, but I'll give it a shot.

bsdjhb:
Also, not sure what to do about the private class member names in libutil++.hh.  Haven't really settled that debate.  I might just go with the "no prefix/suffix at all" since there are no public members for now.

bsdjhb:
This now has manpages for libutil++.

bsdjhb:
FYI, I really want to land this in 15.0 so I can then land the follow-on commits to add NVMeoF support and avoid shipping the more limited /usr/sbin/nvmfd binary in any release.

contrib/libucl/include/ucl++.h
```diff
@@ -396,6 +396,11 @@ class Ucl final {
 		return default_val;
 	}
 
+	std::string forced_string_value () const
+	{
```
bsdimp, line 400: Clearly the style checker has much heartburn with C++

lib/libutil++/Makefile
```diff
@@ -0,0 +1,20 @@
...
+SHLIB_MAJOR=	1
+SRCS=		stringf.cc
+
+MAN+=	freebsd::FILE_up.3 \
+	freebsd::addrinfo_up.3 \
+	freebsd::fd_up.3 \
+	freebsd::malloc_up.3 \
+	freebsd::nvlist_up.3 \
+	freebsd::pidfile.3 \
+	freebsd::stringf.3
```
bsdimp, line 13: Colons in filenames are cool?

lib/libutil++/libutil++.hh
```diff
@@ -0,0 +1,229 @@
...
+
+#ifndef __LIBUTILPP_HH__
+#define	__LIBUTILPP_HH__
+
+#include <sys/cdefs.h>
+#include <sys/nv.h>
+#include <libutil.h>
+#include <netdb.h>
+#include <stdarg.h>
+#include <unistd.h>
```
bsdimp, line 14: There'd been talk about using more cXXXX instead of XXXX.h. Can any of these be converted? or is that just more churn at this point.

lib/libutil++/libutil++.hh
```diff
@@ -0,0 +1,229 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2025 Chelsio Communications, Inc.
+ * Written by: John Baldwin <jhb@FreeBSD.org>
+ */
+
+#ifndef __LIBUTILPP_HH__
+#define	__LIBUTILPP_HH__
+
+#include <sys/cdefs.h>
```
bsdimp: Needed?

lib/libutil++/libutil++.hh
```diff
@@ -0,0 +1,229 @@
...
+#include <sys/nv.h>
+#include <libutil.h>
+#include <netdb.h>
+#include <stdarg.h>
+#include <unistd.h>
+
+#include <memory>
+
+namespace freebsd {
```
bsdimp, line 21: MyFutureSelfThanksYouForThis....

bsdimp (commented):
I can't review it all, but I spot checked and it looks good.

lib/libutil++/Makefile
bsdjhb: As others have noted, some ports (such as Perl) already use colons in filenames.  I did have to fix bsd.man.mk (one of the commits in the branch)

lib/libutil++/libutil++.hh
bsdjhb: Hmmm, for new code like libutil++, I probably should prefer using cXXXX.  Some of these can't be done, but I'll do the ones I can.

lib/libutil++/libutil++.hh
bsdjhb: I thought I needed it for `__printflike`, but perhaps not.

bsdimp: ah yes. Keep it for that. I suspect it's included elsewhere by pollution.

contrib/libucl/include/ucl++.h
bsdjhb: I wonder if it's not worth running it on C++ code, but only *.c and *.h?

lib/libutil++/Makefile
nibanks: Hello! While I realize it might not be a priority here, I just wanted to leave a comment to let you know these recent changes to add files with : in the filename have broken git clone/pull on Windows (which doesn't allow these characters in the file names). I had the repo cloned on a Windows laptop so I could read the code, but now can no longer git pull the latest. It would be great if we could maintain that support. Thanks!

asomers: @nibanks please open a ticket at bugs.freebsd.org .

PR #1798, opened 2025-07-31
release/vmimage.subr: don't run newaliases in default setup

svmhdvn:
release/scripts/mk-vmimage.sh calls vm_emulation_setup to perform any emulator-specific setup needed for the desired output VM image. Currently, this script calls newaliases (coming from commit 1e7c1f1) which was originally needed to build the aliases.db file when sendmail(8) was the default MTA.

When running dma(8) with emulators like qemu-user-static (which doesn't yet honour setgid executables on FreeBSD), this step fails with a 'permission denied' error while trying to read dma.conf, which is now chown root:mail and chmod 0640 after commit a3d4ae7.

Since moving to dma(8) as the base MTA in FreeBSD 14+, newaliases is a no-op in the default image configuration. This patch removes this unnecessary command from the default vm_emulation_setup() function. If mk-vmimage.sh users require newaliases for a sendmail(8)-based VM image setup, they can override vm_emulation_setup() in their own configs, since an override will be required anyway to use sendmail(8) instead of dma(8).

### DIFF
--- a/release/tools/vmimage.subr
+++ b/release/tools/vmimage.subr
@@ -118,7 +118,6 @@ vm_emulation_setup() {
 
 	mkdir -p ${DESTDIR}/dev
 	mount -t devfs devfs ${DESTDIR}/dev
-	chroot ${DESTDIR} ${EMULATOR} /usr/bin/newaliases
 	chroot ${DESTDIR} ${EMULATOR} /bin/sh /etc/rc.d/ldconfig forcestart
 	cp /etc/resolv.conf ${DESTDIR}/etc/resolv.conf
 


svmhdvn:
@dag-erling thanks for the pointer from https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=288409, this patch should fix the issue you mentioned.

svmhdvn:
I did look into that, but the `newaliases=1` path in `contrib/dma/dma.c` also flushes the queue. Changing that would be a more complicated change if consumers of dma are relying on that behaviour.

This is a simpler patch since users who do rely on `newaliases` doing something in a release VM image can simply override `vm_emulation_setup`.

dag-erling:
“`newaliases` also flushed the queue” is arguably a bug. The correct fix is to make it a no-op. But also, this problem will eventually crop up with something else that can't be fixed as easily. We should still address the issue of qemu ostensibly dropping privileges and not honoring the setgid bit.

ghost:
nb, @cperciva has landed the proposed changed in 0a3792d5c5764945259333db4c66774385f01ba7.

cperciva:
Heh, clearly I need to learn to check github when I trip over breakage like this.  Sorry!

svmhdvn:
> “newaliases also flushed the queue” is arguably a bug. The correct fix is to make it a no-op.

I agree, but I'm glad for now that release tooling is unbroken. I'll propose that change sometime soon.

> We should still address the issue of qemu ostensibly dropping privileges and not honoring the setgid bit.

That's an issue to file in https://github.com/qemu-bsd-user/qemu-bsd-user and I currently don't have the expertise or time to track that down.

> Heh, clearly I need to learn to check github when I trip over breakage like this. Sorry!

Lol no worries, thanks! Closing this PR now.

PR #1799, opened 2025-08-02
sockstat: revert incorrect use of path-state for SCTP connections

daminrido139:
Revert a change that mistakenly used SCTP path state and the field name `path-state` instead of the correct `conn-state` and `sctp_conn_state()` call. This was introduced in 7b35b4d during the addition of libxo support.

Fixes:		7b35b4d (sockstat: add libxo support)
Reported by:	Phil Shafer <phil@juniper.net>
Sponsored by:	Google, LLC (GSoC 2025)


daminrido139:
@asomers , please review this when you have time.

asomers (changes requested):
This was reported by Phil, right?  If so, you should credit him in the commit message, like this:
```
Reported by:    phil
```

daminrido139:
You are absolutely right, @asomers . I've added that part in the latest commit message. kindly review!

PR #1801, opened 2025-08-02
Remove duplicate header install with MIT krb5 and fix src.conf.5

brooksdavis:
With this CI returns to green.

Commit messages are a bit sketchy and some of the decisions are arbitrary, but at least in CI it work.


bsdimp:
Closing since it looks like this has landed.

PR #1802, opened 2025-08-04
kern: Allow O_CLOFORK to be passed to shm_open(2)

ricardobranco777:
Allow O_CLOFORK to be passed to shm_open(2)

This flag landed on [OpenBSD](https://github.com/openbsd/src/commit/42a7be81bef70c04732f45ec573622effe56b563) and it's also implemented in this function as an extension to POSIX.

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=286843


kostikbel (approved):

ricardobranco777:
cc/ @markjdb

PR #1804, opened 2025-08-05
libutil: Remove crypt_set_format stub

svmhdvn:
Currently, login_setcryptfmt() from libutil is only called in two places:
* lib/libpam/modules/pam_unix/pam_unix.c
* usr.sbin/pw/pw_user.c

This function is a very thin wrapper over crypt_set_format() from libcrypt. To accomodate this in libutil, stub.c contains a stub for crypt_set_format().

After commit 0c381b7f0570cb118b5d82c3d1777f8b1fdeffd1, we see new CI test failures[0] in usr.sbin/pw and libexec/nuageinit with the following message:

pw: setting crypt(3) format: No error: 0

Upon further investigation, pw(8) eventually calls the crypt_set_format stub in stub.c rather than the libcrypt concrete implementation.

This patch removes the need for this stub.c file and inlines login_setcryptfmt, since it is effectively just one functional line of code.

[0] https://ci.freebsd.org/view/Test/job/FreeBSD-main-aarch64-test/1729/testReport/


svmhdvn:
@dag-erling I understand that this might be a larger change than necessary to fix the exact issue caused by your previous commit 0c381b7f0570cb118b5d82c3d1777f8b1fdeffd1, but the end result is cleaner, since there is no need for a stub or any weak symbols. Please advise on whether the deletion can be done from the `Symbol.map`, as I don't have much experience writing these files.

dag-erling:
You can't just drop a published interface like that.

svmhdvn:
Could you at least discuss the alternatives? Because I am trying to fix a regression caused by 0c381b7f0570cb118b5d82c3d1777f8b1fdeffd1.

dag-erling:
https://reviews.freebsd.org/D51746

PR #1808, opened 2025-08-11
bananapi-r64: patch for early printf ns8250 on arm

Martinfx:


sys/dev/uart/uart_dev_ns8250.c
```diff
@@ -103,6 +104,19 @@ uart_ns8250_early_putc(int c)
...
 	outb(tx, c);
 }
+#elif (defined(__arm__) || defined(__aarch64__))
+static void
+uart_ns8250_early_putc(int c)
+{
+	volatile uint32_t *stat = (uint32_t *)(socdev_va + REG_LSR * 4);
+	volatile uint32_t *tx = (uint32_t *)(socdev_va + REG_DATA * 4);
+	while ((*stat & LSR_THRE) == 0)
+		continue;
```
bsdimp: Please add a limit, ala x86, for safety.

Martinfx: ok

Martinfx: I sent fix.

bsdimp (approved):
I think this is ready. There's no socdev_va/pa on other platofrms, so arm and arm64 is the complete list.

Martinfx:
> I think this is ready. There's no socdev_va/pa on other platofrms, so arm and arm64 is the complete list.

Ok. Thank you.

strejda:
Unfortunately, this only works for some UART implementations. It does not take into account the register layout and access restrictions of different UARTs ("reg-shift", "reg-io-width"). These are not available at an early stage, which is why we do not yet have the early_putc() function implemented. It should not be commited in this form.

Martinfx:
> Unfortunately, this only works for some UART implementations. It does not take into account the register layout and access restrictions of different UARTs ("reg-shift", "reg-io-width"). These are not available at an early stage, which is why we do not yet have the early_putc() function implemented. It should not be commited in this form.

Ok. I thought it was okay.

zxombie:
We could add options to set them, e.g. `options EARLY_PRINTF_REG_SHIFT=2`.

bsdimp:
> We could add options to set them, e.g. `options EARLY_PRINTF_REG_SHIFT=2`.

Couldn't we just read the SPCR table to get these values as well?

zxombie:
I would prefer to not depend on the SPCR/FDT so printf works before we parse them, e.g. we may need to call printf from within the parsing code to debug it.

bsdimp:
> I would prefer to not depend on the SPCR/FDT so printf works before we parse them, e.g. we may need to call printf from within the parsing code to debug it.

Since we parse it in the boot loader, maybe it could provide adequate hints? Or do we have the same issue?

strejda:
IMHO, early_putc() should work from the first line of initarm(). I am afraid that the only viable solution is #define at compile time, runtime detection looks impossible.

erikarn:
hi! Would someone please give @Martinfx  a more concrete example of how we did this compile time stuff for other platforms? That way we can get this finished and landed. Thanks!

sys/conf/options.arm64
```diff
@@ -22,6 +22,10 @@ EFIRT				opt_efirt.h
 # Bhyve
 VMM				opt_global.h
 
+# Early printf UART support
+UART_NS8250_EARLY_REGSHIFT       opt_uart.h.h
```
erikarn: Should be opt_uart.h not opt_uart.h.h.

Martinfx: ah, typo mistake. ok i will fix

Martinfx: i fixed the problem

PR #1809, opened 2025-08-12
vkbd: use the real dev in the man page

YusufKhan-gamedev:
I noticed this in vkbd when porting it to Dragonfly then forgot it, thankfully I believe mneumann figured out the issue and patched it in 0885527.

### DIFF
--- a/share/man/man4/vkbd.4
+++ b/share/man/man4/vkbd.4
@@ -101,7 +101,7 @@ main(void)
 {
         int     fd, len;
 
-        fd = open("/dev/vkbdctl0", O_RDWR);
+        fd = open("/dev/vkbdctl", O_RDWR);
         if (fd < 0)
                 err(1, "open");
 


YusufKhan-gamedev:
> I'm not familiar with this interface, how do you test it?

You `write()` scancodes,(google "keyboard scancodes" for a list) to the /dev/vkbdctl file and it returns keystrokes from the virtual keyboard,  there is an example in the man page. If you want to test anything else, modify codes[] with more scancodes but running the example as is should lead to this output:

```
sh#./run

sh#
```
(Note: assuming the terminal takes input from keyboards)
See how a break was inserted so the terminal emulator added a newline?

> If this is true we need to fix it on line 66 as well.

I don't understand, could you tell me whats wrong?

concussious:
> > If this is true we need to fix it on line 66 as well.
> 
> You write() scancodes,(google "keyboard scancodes" for a list) to the /dev/vkbdctl file and it returns keystrokes from the virtual keyboard, there is an example in the man page.

This part makes sense, but kldloading vkbd did not spawn a /dev/vkbdctl for me.

> I don't understand, could you tell me whats wrong?

Sure, it says:
```
The virtual AT keyboard control device, normally
.Pa /dev/vkbdctl Ns Aq Ar N ,
is exclusive-open
(it cannot be opened if it is already open)
and is restricted to the super-user.
```
That is, `.Pa` (path) `/dev/vkbdctl` `Ns` (no space) `Aq` (angle quote) `Ar` (argument) `N` , e.g. `/dev/vkbdctl<N>` as in, `/dev/vkbdctl0`, `/dev/vkbdctl1`, etc. It doesn't render that way because this is an illegal use of the angle quote macro, but perhaps it was legal twenty years ago when it was written :-)

YusufKhan-gamedev:
> This part makes sense, but kldloading vkbd did not spawn a /dev/vkbdctl for me.

Hmmmm this is weird, there is nothing in the MOD_LOAD area that shows to me that at least the /dev/vkbdctl dev shouldn't be made. UHIDD depends on this behavior!

> That is, .Pa (path) /dev/vkbdctl Ns (no space) Aq (angle quote) Ar (argument) N , e.g.

I guess we should clarify in the man page that `/dev/vkbdctl` is the master clone device and /dev/vkbdctl<N> are the clones of it. You can spawn `/dev/vkbdctl<N>` from interacting with the master device `/dev/vkbdctl` but the example code shouldn't work if you don't open `vkbdctl` first.

I am very sorry for my late reply and the fact that this probably isn't a good explanation, I am very bad at explaining these things.

bsdimp:
Cloned devices don't appaer in /dev. devfs doesn't list them. So it's tricky to know what loading will or won't do.

But in this case, the code sends the keys to virtual keyboard 0, unconditionally. Opening the vkdbctl would create a new virtual keyboard, which I don't think is what the example is showing.

I think we can just close this, because the change is wrong.

PR #1812, opened 2025-08-13
tcp: Initial ktest for HPTS

nibanks:
This adds an initial ktest for the HPTS module. The first commit adds just a minimal stub and the second starts the process of extracting out the testable interface from the HPTS code to be tested.

### DIFF
--- a/sys/modules/ktest/ktest_tcphpts/Makefile
+++ b/sys/modules/ktest/ktest_tcphpts/Makefile
@@ -1,0 +1,13 @@
+PACKAGE=	tests
+WARNS?=		6
+
+SYSDIR?=${SRCTOP}/sys
+.include "${SYSDIR}/conf/kern.opts.mk"
+
+.PATH: ${SYSDIR}/netinet
+
+KMOD=	ktest_tcphpts
+SRCS=	tcp_hpts_test.c
+
+.include <bsd.kmod.mk>
+
--- a/sys/tests/ktest.h
+++ b/sys/tests/ktest.h
@@ -104,6 +104,9 @@ MODULE_VERSION(ktest_##_n, 1);						\
 MODULE_DEPEND(ktest_##_n, ktestmod, 1, 1, 1);				\
 MODULE_DEPEND(ktest_##_n, netlink, 1, 1, 1);				\
 
+#define	KTEST_MODULE_DEPEND(_n, _d)		\
+MODULE_DEPEND(ktest_##_n, _d, 1, 1, 1);	\
+
 #endif /* _KERNEL */
 
 /* genetlink definitions */
--- a/sys/tests/ktest.h
+++ b/sys/tests/ktest.h
@@ -57,12 +57,16 @@ struct ktest_test_info {
 	ktest_parse_t	parse;
 };
 
+#define KTEST_FUNC(X) static int test_hpts_##X(struct ktest_test_context *ctx)
+
 struct ktest_module_info {
 	const char			*name;
 	const struct ktest_test_info	*tests;
 	int				num_tests;
 	void				*module_ptr;
 };
+
+#define KTEST_INFO(X) { "test_hpts_" #X, "Test " #X, test_hpts_##X, NULL }
 
 int ktest_default_modevent(module_t mod, int type, void *arg);
 
@@ -83,6 +87,9 @@ void ktest_end_msg(struct ktest_test_context *ctx);
 
 #define	KTEST_LOG(_ctx, _fmt, ...)					\
 	KTEST_LOG_LEVEL(_ctx, LOG_DEBUG, _fmt, ## __VA_ARGS__)
+
+#define	KTEST_ERR(_ctx, _fmt, ...)	\
+	KTEST_LOG_LEVEL(_ctx, LOG_ERR, _fmt, ## __VA_ARGS__)
 
 #define KTEST_MAX_BUF	512
 
--- a/tests/atf_python/ktest.py
+++ b/tests/atf_python/ktest.py
@@ -67,6 +67,10 @@ class KtestLoader(object):
     def __init__(self, module_name: str, autoload: bool):
         self.module_name = module_name
         self.autoload = autoload
+        # Ensure the base ktest.ko module is loaded
+        result = libc.kldload("ktest")
+        if result != 0 and result != 17:  # 17 is EEXIST (already loaded)
+            logger.debug(f"Failed to load base ktest module (error {result})")
         self.helper = NlHelper()
         self.nlsock = Nlsock(NlConst.NETLINK_GENERIC, self.helper)
         self.family_id = self._get_family_id()
@@ -76,7 +80,9 @@ class KtestLoader(object):
             family_id = self.nlsock.get_genl_family_id(NETLINK_FAMILY)
         except ValueError:
             if self.autoload:
-                libc.kldload(self.module_name)
+                result = libc.kldload(self.module_name)
+                if result != 0 and result != 17:  # 17 is EEXIST (already loaded)
+                    raise RuntimeError(f"Failed to load kernel module '{self.module_name}' (error {result})")
                 family_id = self.nlsock.get_genl_family_id(NETLINK_FAMILY)
             else:
                 raise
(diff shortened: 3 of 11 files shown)


tuexen:
I am wondering if the location of the `tcp_hpts_test.c` file is the right choice. I am not that familiar with test code in the FreeBSD repository, but there exists code in `tests/sys/netinet/` for testing code in `sys/netinet/` and there is code in `sys/tests/`. Now we would have a third location: `sys/netinet/`.  Or should it `sys/netinet/tests/`? Could that code from `tests/sys/netinet/` move there?
I have no specific preference. I am just in favor of consistency, if that is possible. If not, we might want to document somewhere why it is the way it is.

nibanks:
> I am wondering if the location of the `tcp_hpts_test.c` file is the right choice. I am not that familiar with test code in the FreeBSD repository, but there exists code in `tests/sys/netinet/` for testing code in `sys/netinet/` and there is code in `sys/tests/`. Now we would have a third location: `sys/netinet/`. Or should it `sys/netinet/tests/`? Could that code from `tests/sys/netinet/` move there? I have no specific preference. I am just in favor of consistency, if that is possible. If not, we might want to document somewhere why it is the way it is.

I originally followed the example of `FreeBSD/sys/netlink/ktest_netlink_message_writer.c`, which has it in the same path as product code. I also prefer this model as it keeps the product and test code right next to each other for easy discovery.

tuexen:
> > I am wondering if the location of the `tcp_hpts_test.c` file is the right choice. I am not that familiar with test code in the FreeBSD repository, but there exists code in `tests/sys/netinet/` for testing code in `sys/netinet/` and there is code in `sys/tests/`. Now we would have a third location: `sys/netinet/`. Or should it `sys/netinet/tests/`? Could that code from `tests/sys/netinet/` move there? I have no specific preference. I am just in favor of consistency, if that is possible. If not, we might want to document somewhere why it is the way it is.
> 
> I originally followed the example of `FreeBSD/sys/netlink/ktest_netlink_message_writer.c`, which has it in the same path as product code. I also prefer this model as it keeps the product and test code right next to each other for easy discovery.

I see. So would it make sense then to move the tests from `tests/sys/netinet/` to `sys/netinet/` or are they different?

nibanks:
> > I originally followed the example of `FreeBSD/sys/netlink/ktest_netlink_message_writer.c`, which has it in the same path as product code. I also prefer this model as it keeps the product and test code right next to each other for easy discovery.
> 
> I see. So would it make sense then to move the tests from `tests/sys/netinet/` to `sys/netinet/` or are they different?

The main difference I see is kernel vs user tests. I don't have an opinion about moving the existing ones, other than I would prefer to make that a separate PR if we do that.

tests/atf_python/ktest.py
```diff
@@ -67,6 +67,10 @@ class KtestLoader(object):
     def __init__(self, module_name: str, autoload: bool):
         self.module_name = module_name
         self.autoload = autoload
+        # Ensure the base ktest.ko module is loaded
```
AlexanderChernikov, line 70: Shouldn't this be handled by the target module dependency?

AlexanderChernikov:
(Sorry, a bit late to the party). So there are multiple approaches on the actual test code location:
1) separate file, with some boilerplate on exposing internal functions and structures (as seen here). It keeps tests relatively close to the main business logic, while not impacting the main code.
1.1) separate file, `included` in the primary one. In this case we keep splitting between tests and business logic, while avoiding writing boilerplate code. However, it's more "ugly" and I guess some tools may be a bit unhappy.
2) same file, with tests embedded. It adds up to the total file size, but still avoids the boilerplate code, simplifying adding new tests.

Personally I'd prefer either 1.1 or 2 approach as those minimise code changes to be "testable" and reduces boilerplate code.

On the tests file location - in the kernel we'll (potentially) mostly thinking of unit tests, so keeping the test code close to the actual code (e.g. same directory) may be beneficial. We'd still have the "glue" ATF file at the standard FreeBSD testing path, so it's sort of "following" the logic.

Neither are hard opinions and I'm fine with any (consistent) approach we choose and land to the tree.

sys/modules/ktest/ktest_tcphpts/Makefile
```diff
@@ -0,0 +1,13 @@
+PACKAGE=	tests
+WARNS?=		6
+
+SYSDIR?=${SRCTOP}/sys
+.include "${SYSDIR}/conf/kern.opts.mk"
+
+.PATH: ${SYSDIR}/netinet
+
+KMOD=	ktest_tcphpts
+SRCS=	tcp_hpts_test.c
```
AlexanderChernikov, line 10: Nit: should we have the `test` prefix, not postfix?

sys/tests/ktest.h
```diff
@@ -57,13 +57,17 @@ struct ktest_test_info {
 	ktest_parse_t	parse;
 };
 
+#define KTEST_FUNC(X) static int test_hpts_##X(struct ktest_test_context *ctx)
```
AlexanderChernikov, line 60: `_hpts_` seem to be not the intended prefix

sys/tests/ktest.h
```diff
@@ -57,13 +57,17 @@ struct ktest_test_info {
...
+
 struct ktest_module_info {
 	const char			*name;
 	const struct ktest_test_info	*tests;
 	int				num_tests;
 	void				*module_ptr;
 };
 
+#define KTEST_INFO(X) { "test_hpts_" #X, "Test " #X, test_hpts_##X, NULL }
```
AlexanderChernikov, line 69: Same here

tuexen:
> (Sorry, a bit late to the party). So there are multiple approaches on the actual test code location:
> 
> 1. separate file, with some boilerplate on exposing internal functions and structures (as seen here). It keeps tests relatively close to the main business logic, while not impacting the main code.
>    1.1) separate file, `included` in the primary one. In this case we keep splitting between tests and business logic, while avoiding writing boilerplate code. However, it's more "ugly" and I guess some tools may be a bit unhappy.
> 2. same file, with tests embedded. It adds up to the total file size, but still avoids the boilerplate code, simplifying adding new tests.
> 
> Personally I'd prefer either 1.1 or 2 approach as those minimise code changes to be "testable" and reduces boilerplate code.
> 
> On the tests file location - in the kernel we'll (potentially) mostly thinking of unit tests, so keeping the test code close to the actual code (e.g. same directory) may be beneficial. We'd still have the "glue" ATF file at the standard FreeBSD testing path, so it's sort of "following" the logic.

I have no hard opinion either. But FreeBSD has tests for functionality implemented in `sys/netinet` in `tests/sys/netinet/` and in `sys/tests`. So you are suggesting to not locate test files somewhere, but put them near the files they are testing. Would you suggest to move the files from `tests/sys/netinet/` around to new locations, or keep them? Then what is the criteria? Or create a `tests` subdirectory in `sys/netinet/` and put all related test files there?
We can also do it in some inconsistent way, just want that we do this intentionally and not by accident.

> 
> Neither are hard opinions and I'm fine with any (consistent) approach we choose and land to the tree.

sys/tests/ktest.h
nibanks: Yes, I've since updated internally (but not pushed externally) this to:
```c
#define KTEST_FUNC(X) static int __ktest_##X(struct ktest_test_context *ctx)
```

sys/tests/ktest.h
nibanks: Internally updated to :
```c
#define KTEST_INFO(X) { "test_" #X, "Test " #X, __ktest_##X, NULL }
```

sys/modules/ktest/ktest_tcphpts/Makefile
nibanks: I don't feel super strongly about it, but when viewing files alphabetically, I feel you figure out a file has tests more easily with a suffix:
```
$ ls -w 1 sys/netinet/tcp*
...
sys/netinet/tcp.h
sys/netinet/tcp_hostcache.c
sys/netinet/tcp_hpts.c
sys/netinet/tcp_hpts.h
sys/netinet/tcp_hpts_internal.h
sys/netinet/tcp_hpts_test.c
sys/netinet/tcp_input.c
...
```

tests/atf_python/ktest.py
nibanks: When debugging, it didn't seem to be working automatically...

PR #1815, opened 2025-08-19
tests/ci: prepare for integrating into ci.freebsd.org

svmhdvn:
See full commit stack for incremental details and explanations.

### DIFF
--- a/tests/ci/Makefile.aarch64
+++ b/tests/ci/Makefile.aarch64
@@ -27,7 +27,7 @@ portinstall-aarch64: portinstall-pkg .PHONY
 	env ASSUME_ALWAYS_YES=yes pkg install sysutils/u-boot-bhyve-arm64
 .endif
 
-ci-runtest-aarch64: ci-buildimage-aarch64 .PHONY
+ci-runtest-aarch64: .PHONY
 	/usr/sbin/bhyvectl --vm=${TEST_VM_NAME} --destroy || true
 	expect -c "set timeout ${TIMEOUT_EXPECT}; \
 		spawn /usr/bin/timeout -k 60 ${TIMEOUT_VM} /usr/sbin/bhyve \
@@ -47,5 +47,5 @@ portinstall-aarch64: portinstall-pkg .PHONY
 	env ASSUME_ALWAYS_YES=yes pkg install sysutils/u-boot-qemu-arm64
 .endif
 
-ci-runtest-aarch64: ci-runtest .PHONY
+ci-runtest-aarch64: ci-runtest-qemu .PHONY
 .endif
--- a/tests/ci/Makefile.aarch64
+++ b/tests/ci/Makefile.aarch64
@@ -13,11 +13,6 @@ QEMU_EXTRA_PARAM=-bios /usr/local/share/u-boot/u-boot-qemu-arm64/u-boot.bin -cpu
 QEMU_MAX_CPU_COUNT=64
 QEMU_MAX_MEM_SIZE=64
 
-portinstall-aarch64: portinstall-pkg .PHONY
-.if !exists(/usr/local/share/u-boot/u-boot-qemu-arm64/u-boot.bin)
-	env ASSUME_ALWAYS_YES=yes pkg install sysutils/u-boot-qemu-arm64
-.endif
-
 # NOTE: Nothing should be changed below this line unless explicitly required.
 
 ci-buildworld-aarch64: ci-buildworld .PHONY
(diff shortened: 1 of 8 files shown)


svmhdvn:
@lwhsu @5u623l20 switching to github for the time being since it's a lot easier to reorganize commits and give full context automatically.

lwhsu:
I'm still reading and thinking. Just a quick response: I think we can drop i386 (`Makefile.i386` file) as it's also disabled in main branch. I don't think we would MFC those test/ci stuff as older branches can still use the old mechanism.

tests/ci/Makefile.aarch64
```diff
@@ -17,6 +17,9 @@ portinstall-aarch64: portinstall-pkg .PHONY
 .if !exists(/usr/local/share/u-boot/u-boot-qemu-arm64/u-boot.bin)
```
lwhsu: Maybe this part also needs to be guarded by `USE_QEMU` ?

lwhsu:
Moving bhyve bits into Makefile.{aarch64,amd64} seems introducing some code duplication, perhaps the difference part like `-o bootrom` can be set as a variable in those files like `BHYVE_ADDITIONAL_PARAMETERS` and the common parts can still be `Makefile`?

svmhdvn:
> Moving bhyve bits into Makefile.{aarch64,amd64} seems introducing some code duplication, perhaps the difference part like `-o bootrom` can be set as a variable in those files like `BHYVE_ADDITIONAL_PARAMETERS` and the common parts can still be `Makefile`?

The difference is quite large:
* `bhyveload` is not supported yet on arm64 (i.e. difference in bootrom loading)
* `lpc` is not supported on arm64 (i.e. difference in devices)
* `-o console` is used on arm64 (i.e. difference in output configuration)

There is not too much code duplication here in the grand scheme of things, and it keeps things simpler by relying on less variable indirection.

tests/ci/Makefile.aarch64
svmhdvn: Applied the feedback in v2.

svmhdvn:
Fixed merge conflict in v3.

lwhsu:
@svmhdvn I finally got time to read this PR and think of it. I haven't had a clear thought of a7aaf60b1b31440c7da2331c7c68c19414aa264b , do you think we can get the other 8 commits merged first? Is there any concern I missed if doing so?

svmhdvn:
> do you think we can get the other 8 commits merged first?

Yes that should be fine. However, since it has been a while since making this PR, let me do a local rebase and properly test with the other 8 commits and report back.

svmhdvn:
Just ran a test with all commits except https://github.com/freebsd/freebsd-src/commit/a7aaf60b1b31440c7da2331c7c68c19414aa264b and everything works just fine. You can go ahead with the rest of the commits after being satisfied with your own testing.

5u623l20:
I am really not sure why we are splitting the bhyve and qemu into 2 parts. If I am not mistaken bhyve support in arm64 will be here in the tree. And that means we will need to duplicate that code block again in Makefile.arm64.

5u623l20:
I am also not upto the point that we should entirely shift to pkgbase for the image creation. We should optionize it with the var `NOPKGBASE`. In case we plan to add these into 14 this will be the sane way forward. And we should get rid of packages  and move to pkgbase only when there are no more release remaining in support that does not support packages.

svmhdvn:
> I am really not sure why we are splitting the bhyve and qemu into 2 parts. If I am not mistaken bhyve support in arm64 will be here in the tree. And that means we will need to duplicate that code block again in Makefile.arm64.

Because the command line options to bhyve are highly platform specific. For example, `-A`, `-H`, and `-P` arguments are not supported on arm64, and the syntax for specifying a console and bootrom are different between amd64 and arm64. In the future when riscv64 hardware starts to support the hypervisor extension and we use bhyve on those platforms, those arguments may be different too.

Trying to abstract these specific arguments out into variables introduces more complexity and makes it harder to follow. I don't think these code blocks are fully "duplicated".

> I am also not upto the point that we should entirely shift to pkgbase for the image creation. We should optionize it with the var NOPKGBASE. In case we plan to add these into 14 this will be the sane way forward.

I agree with this, I'll optionize pkgbase and keep it as the default, since we do want CI to catch pkgbase-specific errors early. When integrating src/tests/ci into jenkins for stable 14, I'll set `NOPKGBASE=1`. Let me post a new revision of that commit shortly.

bsdimp:
> > I am really not sure why we are splitting the bhyve and qemu into 2 parts. If I am not mistaken bhyve support in arm64 will be here in the tree. And that means we will need to duplicate that code block again in Makefile.arm64.
> 
> Because the command line options to bhyve are highly platform specific. For example, `-A`, `-H`, and `-P` arguments are not supported on arm64, and the syntax for specifying a console and bootrom are different between amd64 and arm64. In the future when riscv64 hardware starts to support the hypervisor extension and we use bhyve on those platforms, those arguments may be different too.

Seems like for many of these things, we should have something in common. Why are they different?

5u623l20:
> > I am really not sure why we are splitting the bhyve and qemu into 2 parts. If I am not mistaken bhyve support in arm64 will be here in the tree. And that means we will need to duplicate that code block again in Makefile.arm64.
> 
> Because the command line options to bhyve are highly platform specific. For example, `-A`, `-H`, and `-P` arguments are not supported on arm64, and the syntax for specifying a console and bootrom are different between amd64 and arm64. In the future when riscv64 hardware starts to support the hypervisor extension and we use bhyve on those platforms, those arguments may be different too.
> 
> Trying to abstract these specific arguments out into variables introduces more complexity and makes it harder to follow. I don't think these code blocks are fully "duplicated".

If that is the case then we should still keep the basic blocks in the main `Makefile` and modify the parameters using something like `EXTRA_ARGS_ARCH` or something. Not duplicate the entire thing.

svmhdvn:
> [Siva] When integrating src/tests/ci into jenkins for stable 14, I'll set NOPKGBASE=1. Let me post a new revision of that commit shortly.

In the interest of getting the important commits in, I've decided to set `NOPKGBASE=1` unconditionally for now (thus not using pkgbase at all). I'll submit another revision later on and we can examine that change, since it currently proves to be a bit unstable after some unknown changes that came in the last month.

> [Moin] If that is the case then we should still keep the basic blocks in the main Makefile and modify the parameters using something like EXTRA_ARGS_ARCH or something. Not duplicate the entire thing.

The entire block is not duplicated. On AMD64, we need an extra command to load the image through `bhyveload`:

```sh
expect -c "set timeout ${TIMEOUT_EXPECT}; \
    spawn /usr/bin/timeout -k 5s 30s /usr/sbin/bhyveload \
    -c stdio -m ${VM_MEM_SIZE} -d ${CIDISK} ${TEST_VM_NAME}; \
    expect { eof }; \
    exit [lindex [wait] 3]"
```

This command is not supposed to be run on aarch64 and riscv64.

I personally believe that it would cause _more_ complexity to try and reuse just the main block of this logic, while also overriding the logic specifically for amd64 by adding an extra command before the actual `bhyve` invocation. We not only need a new `BHYVE_EXTRA_ARGS` per architecture, but we'd also need another target in between to handle the amd64 case specially.

svmhdvn:
Is there anything further to discuss from this review? I'd like to start getting in the commits and work on integrating this work to freebsd/freebsd-ci.

svmhdvn:
Disregard the spam, something went wrong in my local VCS.

svmhdvn:
@5u623l20 @bsdimp can we get these commits merged and then later discuss style issues as necessary? The changes have been well tested and are needed to do further work on CI.

svmhdvn:
Issues with `release/Makefile` and `mk-vmimage.sh` are fixed now. 

> I agree with this, I'll optionize pkgbase and keep it as the default, since we do want CI to catch pkgbase-specific errors early. When integrating src/tests/ci into jenkins for stable 14, I'll set NOPKGBASE=1. Let me post a new revision of that commit shortly.

I have made this change in the latest revision of this PR. `NOPKGBASE=1` will use the traditional install method.

ngie-eign (commented):
- I would use `INSTALL ...` instead of `mkdir` to avoid problems with umasks, questionable ownership of paths, etc.
- I would use patterns with default values like this instead:
```
QEMUSTATIC?= /usr/local/bin/qemu-${QEMU_ARCH}-static
.if !exists(${QEMUSTATIC}
```

PR #1817, opened 2025-08-20
Add NUL-terminated output to jail(8)

kraken-jim:
jail(8) has only simple string-quoting logic with regard to the output from `jail -e`.  This pull request adds NUL-delimited output (ala `find(1)`, `xargs(1)`, `grep(1)`, et al.) so that arbitrary strings defined in the `/etc/jail.conf(5)` configuration hierarchy can be unambiguously represented in the `jail -e` output.  This enhancement to `jail(8)` makes it possible for user scripts to leverage this accurate data format to create automated scripting capable of performing detailed jail-related maintenance tasks.  A corresponding edit to the man page is also included.


kraken-jim:
To see an example of a tool that utilizes this functionality:

Download the script:

https://github.com/kraken-jim/jail-info/blob/master/FreeBSD/stage/usr/local/bin/jail-info

Or install the prototype port **sysutils/jail-info**:

```
# cd /usr/ports
# git checkout -b jail-info
# wget ftp:://umpquanet.com/pub/usr-ports-sysutils-jail-info.patch
# git apply usr-ports-sysutils-jail-info.patch
# cd sysutils/jail-info
# make install clean
```

kraken-jim:
Closed in favor of patch against -CURRENT

PR #1818, opened 2025-08-21
[GSoC WIP] efi: create libacpi

kpowkitty:
#### Description:
Refactored ACPICA for partial initialization in the loader.
This work is the preface for Lua bindings.

For our Lua bindings, we want to be able to walk the
ACPI namespace and evaluate objects. So, I've taken what
the kernel does in acpi_Startup() as a starting point,
and added the  ACPI initialization for this. That includes
AcpiInitializeSubsystem, AcpiInitializeTables, 
AcpiEnableSubsystem, and AcpiLoadTables. It is initialized 
in reduced hardware mode with events enabled, to (1) 
match what our Lua interface desires and (2) create a 
smaller subset of ACPICA for the loader environment.

#### Notes:
* acpi_detect() moved into libacpi.
* Libacpi conditionally compiles ACPICA and glue code on amd64.
(arm64 will be addressed in a future update)
* AcpiEnableSubsystem is initialized in reduced hardware mode,
but with events enabled. This is so that we do not need to pull
in all of ACPI_HARDWARE, but can evaluate objects.
* Stubbed out Osd functions required for this limited initialization.
* Conditionally exclude functions to make a minimal subset designed
around loader's/lua binding's needs:
     * Exclude AcpiInitializeObjects, as we do not need this for walking
the namespace and evaluating objects. (utxfinit.c)
     * Exclude notify handlers or fixed event handlers. (evxface.c)
     * Stub out print functions, as the loader does not have standard
output. (osunixxf.c)
     * Stub out AcpiOsSleep and AcpiOsGetTimer, as the loader is not
multi-threaded and does not implement timers. (osunixxf.c)

Sponsored by: Google Summer of Code 2025

### DIFF
--- a/stand/efi/libacpi/README.md
+++ b/stand/efi/libacpi/README.md
@@ -1,0 +1,64 @@
+# Google Summer of Code 2025 @ FreeBSD
+Mentor: Warner Losh @bsdimp <imp@FreeBSD.org>  
+Student: Kayla (Kat) Powell @kpowkitty <kpowkitty@FreeBSD.org>
+---------------
+## ACPI Initialization in Loader with Lua Bindings
+Intel's Advanced Configuration and Power Interface (ACPI) brought power management out of the BIOS
+and into the operating system. We can make it even more powerful by creating an interface
+to script it. Currently, the scripting language of choice in FreeBSD (rather than POSIX `sh`, `bash`, or `awk`) 
+is Lua. Therefore, this project aims to initialize a portion of ACPI in the FreeBSD bootloader, with respect
+to its storage, memory, and stdlib constraints, so we can enumerate and evaluate objects on the device trees to
+provide an interface to Lua.
+
+## Milestones
+[x] `OsdMemory.c` for `amd64` (`arm64` postponed)  
+[x] `AcpiInitializeSubsystem`  
+[x] `AcpiInitializeTables`  
+[x] `AcpiEnableSubsystem` in reduced hardware mode, with events enabled  
+[x] `AcpiLoadTables`  
+[x] `AcpiWalkNamespace`  
+[x] `AcpiEvaluateObject`  
+[x] ACPICA initialized in loader for `amd64`    
+[x] lacpi_object.c  (V1)
+    -- V1: Users can evaluate objects
+[x] lacpi_data.c    (V1)
+    -- V1: Users can attach, get, detach data from nodes
+[x] lacpi_walk.c    (V1)
+    -- V1: Walk and read nodes off the namespace
+    -- V2: Namespace format -- TBR
+    -- V3: Strategies for walking the namespace -- TBR
+[x] Man page update  
+
+**Future goals of this project include arm64 compat and extending the interface.**
+
+### For more information, please consult these resources:
+[ACPI Lua Bindings FreeBSD project wiki page](https://wiki.freebsd.org/SummerOfCode2025Projects/ACPI%20Initialization%20in%20Loader%20With%20Lua%20Bindings)  
+[Write-up](https://kmpow.com/content/gsoc-writeup)
+
+---------------
+
+## Building
+This project requires an amd64 UEFI FreeBSD system.
+
+The current working version of this project is found on branch `acpi_init`.
+
+1. Clone the repo
+```
+$ git clone git@github.com:kpowkitty/freebsd-src.git
+$ git checkout acpi_init
+```
+
+2. Build the necessary dependencies and install the loader image
+```
+$ cd stand/efi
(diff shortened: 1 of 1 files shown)


stand/efi/acpica/acpi_detect.c
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+#include <machine/_inttypes.h>
+
+#include <efi.h>
+
+#include <acpi.h>
+
+#include "acpi_detect.h"
+
```
bsdimp, line 32: style: delete the  bank lines between the includes.

stand/efi/acpica/acpi_detect.c
```diff
@@ -0,0 +1,71 @@
...
+	setenv("acpi.rsdt", buf, 1);
+	if (revision >= 2) {
+		/* XXX extended checksum? */
+		sprintf(buf, "0x%016llx",
+		    (unsigned long long)rsdp->XsdtPhysicalAddress);
+		setenv("acpi.xsdt", buf, 1);
+		sprintf(buf, "%d", rsdp->Length);
+		setenv("acpi.xsdt_length", buf, 1);
+	}
+}
```
bsdimp, line 69: So moving this routine to this file should be its own commit since we're just moving existing code.
And since we're moving existing code, it should retain the copyrights of the original authors. The original author is Ed Maste (emaste@freebsd.org). The code moved once before into stand/efi/loader/main.c from many copies, the oldest of which was in stand/efi/loader/arch/amd64/elf64_freebsd.c which Ed wrote in 2014.

stand/efi/acpica/init_acpi.c
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	/* Create the ACPI namespace from ACPI tables. */
+	if (ACPI_FAILURE(status = AcpiLoadTables())) {
+		printf("ACPI: Load tables failed: %s\n",
+		    AcpiFormatException(status));
+		return_VALUE (status);
+	}
+
+	printf("Successfully initialized ACPI.\n");
```
bsdimp: This is good for debug, but should likely not be printed every time the loader loads.

stand/efi/acpica/init_acpi.c
```diff
@@ -0,0 +1,118 @@
...
+	ACPI_FUNCTION_TRACE((char *)(uintptr_t)__func__);
+
+	if (acpi_inited) {
+		return 0;
+	}
+	
+	/* Initialize root tables. */
+	if (ACPI_FAILURE(status = acpi_Startup())) {
+		printf("ACPI: Startup failed with status %s\n", AcpiFormatException(status));
+		printf("ACPI: Try disabling either ACPI or apic support.\n");
```
bsdimp: this advice needs to be tailored to the loader.
And suggests we want a way to disable looking for ACPI which might be its own can of worms.

stand/efi/loader/Makefile
```diff
@@ -30,6 +30,56 @@ SRCS=	autoload.c \
 	gfx_fb.c \
 	8x16.c
 
+# ACPICA Loader Initialization
```
bsdimp: same set of comments I had for the lua review.

stand/efi/contrib/dev/acpica/os_specific/service_layers/osunixxf.c
```diff
@@ -0,0 +1,1526 @@
+/******************************************************************************
```
bsdimp, line 1: I'll have to snag this diff, github's tool is useless for review.

sys/contrib/dev/acpica/include/acpixf.h
```diff
@@ -261,7 +261,11 @@ ACPI_INIT_GLOBAL (UINT8,            AcpiGbl_AutoSerializeMethods, TRUE);
  * because ACPICA is fully compatible with other ACPI implementations.
  * Changing this will revert ACPICA (and machine ASL) to pre-OSI behavior.
  */
+#ifndef _STANDALONE
```
bsdimp: Why is this false for the loader?

kpowkitty: Hmm.. So, when I first was building, I was consistently getting a null reference on the _OSI node when I was printing the ACPI namespace for debug in one of the ACPI routines. May be my own error. I did test that it compiles and runs without the interjected print statements, so I reverted this change. Let me know if we should keep it.

stand/efi/acpica/init_acpi.c
kpowkitty: I opted to have method headers and minimal error message. Let me know if a more verbose error message is preferred.

PR #1819, opened 2025-08-21
[GSoC WIP] liblua: ACPICA Lua bindings

kpowkitty:
#### Follows #1843 and #1818 in that order.

Description:
Lua bindings for ACPICA in loader to provide an interface for users to
print the ACPI namespace, attach/detach/get data off of device nodes,
and evaluate objects.

Please see man page commit for `acpi.lua` for more information

#### lacpi_object.c
Module for evaluating ACPI objects.
This module also provides the ability to convert pathnames into
their opaque pointer using `get_handle`.

#### lacpi_walk.c
Allows users to walk the namespace in the loader.
Currently works by walking the entire namespace and
storing each node in a 2D table of nodes.

Use case:
nodes = lacpi.walk.namespace()
for _, node in ipairs(nodes) do
	print(node.path, node.lvl, node.hid, node.uid)
end

Future plans:
(1) better representation of nodes for the user when printed
(2) strategy patterns for how to receive the namespace

#### lacpi_data.c
Allows a user to post ACPI_OBJECTs onto namespace nodes.
See actypes.h for all possible ACPI_OBJECTs.

Sponsored by: Google Summer of Code 2025

### DIFF
--- a/stand/lua/acpi.lua.8
+++ b/stand/lua/acpi.lua.8
@@ -1,0 +1,182 @@
+.\"
+.\" Copyright (c) 2025 Google Summer of Code
+.\"
+.\" SPDX-License-Identifier: BSD-2-Clause(XXX?)
+.\"
+.Dd September 28, 2025
+.Dt ACPI.LUA 8
+.Os
+.Sh NAME
+.Nm acpi.lua
+.Nd ACPI Lua binding module
+.Sh DESCRIPTION
+The ACPI Lua bindings are for utilizing ACPICA
+functionality pre-kernel. They are only available to the
+.Fx
+Lua bootloader, via the
+.Ic loader
+table, and require
+.Ic amd64
+architecture. See
+.Xr loader.lua 8
+for more information on the loader table.
+.Pp
+These bindings are in the 
+.Ic lacpi
+namespace, and are divided between three modules:
+.Bl -tag -width 0n
+.It Sy walk 
+The
+.Ic walk
+module is for walking the ACPI namespace under user specified
+callbacks and passing the information back to the interpreter
+through the Lua stack. Currently, users can walk the entire 
+namespace or start the walk at a user specified
+.Ic handle .
+There is currently only one callback, and it retrieves the
+following into a Lua table.
+.Bl -tag -width 0n
+.It Sy pathname
+The full pathname of the current device node
+.It level
+The level of the current device node
+.It hid
+If applicable, the HID of the current device node
+.It uid
+If applicable, the UID of the current device node
+.El
+.It Sy object 
+The
+.Ic object
+module is for evaluating ACPI objects. It provides
+the ability to
+.Bl -tag -width 0n
+.It Fn get_handle pathname
+Converts a
+.Va pathname
+to its corresponding
+.Ic opaque pointer .
+.It Fn evaluate handle abs_path rel_path args
+Evaluates an ACPI object. The arguments are:
+.Bl -enum
+.It 
+The handle of the node.
+.It 
+The relative pathname to the handle. Optional, and only
+provided if the handle is being provided.
+.It 
+An absolute pathname.
+.It 
+An
+.Va OBJECT_LIST
+for the specified objects
+.El
+.Pp 
+This submodule takes either:
+.Bl -bullet
+.It
+A handle, with an optional relative pathname, plus an 
+optional OBJECT_LIST.
+.It 
+An absolute pathname, plus an optional OBJECT_LIST
+.El
+.El
+.It Sy data
+The
+.Ic data
+module allows users to manipulate data on namespace
+nodes. The interface currently provides the abilities:
+.Bl -tag -width 0n
+.It Fn attach handle ACPI_OBJECT
+Attach data onto a namespace node
+.It Fn detach handle
+Detach data from a namespace node
+.It Fn get handle
+Get data off of a namespace node
+.El
+.El
+.Sh NOTES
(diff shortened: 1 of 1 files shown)


stand/efi/loader/Makefile
```diff
@@ -30,6 +30,65 @@ SRCS=	autoload.c \
 	gfx_fb.c \
 	8x16.c
 
+# ACPICA Loader Initialization
+SRCS+= acpi_detect.c
+
+.if ${MACHINE_ARCH} == "amd64"
+SRCS+= OsdMemory.c osunixxf.c
```
bsdimp: So for all these files, I'd be tempted to create a stand/efi/libacpi and build them all and put them all in there. Since they are all listed here, they all build three times. It's fine for development, but it's a lot to build three times. Not an immediate thing, but before commit. Fortunately, moving it a library is pretty straight forward (though I've done it so many times I may be forgetting a gotcha).

stand/liblua/Makefile
```diff
@@ -29,6 +29,11 @@ SRCS+=	lfs.c
 .PATH:	${FLUALIB}/libhash
 SRCS+=	lhash.c
 
+# ACPI
+.PATH: ${LIBLUASRC}/acpi ${EFISRC}/acpica
+SRCS+=  lacpi_object.c init_acpi.c
+CFLAGS+= -I${LIBLUASRC}/acpi -I${EFISRC}/acpica/include -I${SYSDIR}/contrib/dev/acpica/include
+
```
bsdimp, line 45: This needs to be conditional on at least amd64.

stand/liblua/acpi/lacpi.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include <lua.h>
+#include <sys/linker_set.h>
+
+#include <contrib/dev/acpica/include/acpi.h>
+
+typedef int (*lua_module_init_fn)(lua_State *L);
```
bsdimp, line 7: this line triggers a false positive on the style checker that can be ignored.

stand/liblua/acpi/lacpi.h
```diff
@@ -0,0 +1,28 @@
...
+#define LUA_ACPI_COMPILE_SET(name, initfn)			\
+	static struct lua_acpi_module lua_##name =		\
+	{							\
+		.mod_name = #name,				\
+		.init = initfn					\
+	};							\
+	DATA_SET(lua_acpi_modules, lua_##name)
+
+extern struct lacpi_handle {
+	const char*	pathname;
```
bsdimp, line 27: this should be `const char *pathname` to match FreeBSD's style(9).

stand/liblua/acpi/lacpi_object.c
```diff
@@ -0,0 +1,377 @@
...
+	obj->Package.Elements = NULL;
+	obj->Package.Count = 0;
+}
+
+static void
+free_acpi_obj(ACPI_OBJECT *obj)
+{
+	if (obj) {
+		switch (obj->Type) {
+			case ACPI_TYPE_STRING:
```
bsdimp: I'd be tempted to have a ACPI_TYPE_INTEGER here that does nothing. And then below in the default trigger some kind of assert or error (not sure which, but given where I've seen this called from, I think an assert would be better since it's a 'can't happen' programming error.

stand/efi/loader/Makefile
```diff
@@ -30,6 +30,65 @@ SRCS=	autoload.c \
...
+	exsystem.c exnames.c exresolv.c exstore.c exfield.c exconvrt.c \
+	exresnte.c exutils.c extrace.c hwpci.c hwregs.c hwxface.c hwvalid.c \
+	hwacpi.c evmisc.c evregion.c evrgnini.c evxface.c evevent.c evxfevnt.c \
+	evhandler.c evxfregn.c psobject.c psloop.c pstree.c pswalk.c psopinfo.c \
+	psscope.c psutils.c psargs.c psparse.c psxface.c psopcode.c init_acpi.c
+
+# lua
+.PATH: ${.CURDIR}/../../liblua/acpi
+SRCS+= lacpi.c
+CFLAGS+= -I${.CURDIR}/../../liblua/acpi
```
bsdimp: The whole lua section likely needs something like what I do for kboot:
```
# Add our own lua bindings until they are universal                                                                                                                                          
.if ${LOADER_INTERP} == "lua"
SRCS+=          \
                lua_bindings.c
CFLAGS.lua_bindings.c+=  -I${FLUALIB}/libhash
.endif

```
But we really want to bring it all in, and that's a little more complex since we want them to live in the liblua.a, so the way we do that is to put everything in liblua. To make sure that lacpi.c gets pulled in, we have the loader main program reference something in that. For gfx, we define

```
void
gfx_interp_ref(void)
{
}
```
in the liblua files that implement the gfx table, and then call it from main(). We'll need something similar here.

I also don't see lacpi.c in this pull request, but maybe it's in others.

stand/liblua/lutils.c
```diff
@@ -469,6 +484,11 @@ luaopen_loader(lua_State *L)
 	lua_add_features(L);
 	/* Set global printc to loader.printc */
 	lua_register(L, "printc", lua_printc);
+	
+	if (acpi_is_initialized()) {
+		register_lacpi_modules(L);
+	}
+
```
bsdimp, line 494: to go along with the comments above, we'll need to use the gfx mechanisms and/or the kboot mechanisms so we avoid references to acpi in this file. Otherwise, other loaders will fail to load since they don't have the acpi symbols.

PR #1821, opened 2025-08-25
libcrypto: diable KTLS in the bootstrap libcrypto

kevans91:
It won't be supported on, e.g., macOS, and we don't actually need it for what we do with libcrypto in bootstrap tools.

(Just to trigger builds)


markjdb:
I submitted a similar patch on phabricator: https://reviews.freebsd.org/D52340

kevans91:
> I submitted a similar patch on phabricator: https://reviews.freebsd.org/D52340

Ah yes, sorry, I forgot to abandon this.

PR #1823, opened 2025-08-27
Optimizations to the AT keyboard driver to reduce unnecessary delays during the FreeBSD boot process.

bllgg:
### Summary of changes

- Added TSLOGs to identify delay bottlenecks in the boot sequence.
- Introduced a new loader tunable `hw.atkbd.fast_delay` to `/boot/loader.conf`.
  - When enabled, the driver uses atkbd_short_delay, which applies significantly shorter delays in the following routines:
    - `empty_both_buffers()`
    - `reset_aux_dev()`
    - `test_controller()`
    - `reset_kbd()`
  - New short delays are up to 1000× smaller (for buffer resets) and about 40× smaller (for controller and keyboard resets).
  - Default behaviour remains with shorter delays, and `hw.atkbd.short_delay` should be set to 0, if someone needs longer delays for the old, slower hardware

#### Notes:
This patch is part of the ongoing GSoC 2025 work on FreeBSD boot process optimisation.


stand/kshim/bsd_kernel.c
```diff
@@ -25,6 +25,8 @@
 
 #include <bsd_global.h>
 
+#include <sys/tslog.h>
```
bsdimp, line 28: I doubt that these will work with this file. kshim is a special environment and it's hard to test. I doubt this would even build in that environment.

contrib/pnpinfo/pnpinfo.c
```diff
@@ -25,6 +25,7 @@
  */
 
 #include <sys/time.h>
+#include <sys/tslog.h>
```
bsdimp, line 28: This file is userland code, and not used in boot at all. In fact, only really old i386 systems would have any cards this might find and it's likely going to be removed from the system when we remove i386 kernel support. What's the point of profiling it?

bsdimp:
I know @cperciva asked for the ts profiling changes to be in separate commits. I'd suggest that it should go a step further and be a separate pull request since it's so different than the keyboard driver changes. Work with him about when the right time to do that might be...

bllgg:
Hi @bsdimp, Thank you for the comments. I am creating a separate pull request according to @cperciva's instructions. Will send it soon.

PR #1824, opened 2025-08-27
acpi_ibm: register all appropriate sysctls as uint

Defenso-QTH:
All sysctls except thermal sensor readings and handlerevents whitelist-separated string are bitmasks or small integers. This avoids some bitmasks being erroneously displayed as negative values.


kprovost:
Should we also change acpi_ibm_sysctl()'s 'arg' variable to be unsigned?

gmshake:
The function `acpi_ibm_sysctl_get()` which is called by `acpi_ibm_sysctl()` returns `-1` in some cases. I think it needs more revising if change it to return unsigned int.

bsdimp (approved):

bsdimp:

PR #1825, opened 2025-08-27
fusefs: Upgrade FUSE protocol to version 7.36.

rmscotty:
This commit upgrades the FUSE API to protocol 7.36, it doesn't implement any of protocol 7.36's new features,


emaste:
CC @asomers

sys/fs/fuse/fuse_internal.c
asomers: You can't use `fuse_libabi_geq` here because `data` isn't fully initialized yet.  It can't be initialized until the server responds to FUSE_INIT.  Instead, you should always fully initialize the structure.  And you should set `FUSE_INIT_EXT`.  Otherwise the server might be expecting a structure of the wrong size.

asomers (changes requested):
Since this change doesn't have any real functionality at all, I don't think you need to add tests for it.  But I do recommend initializing `out->body.init.flags2` in `MockFS::init` in the test suite.

PR #1826, opened 2025-08-28
Optimizations to the AT keyboard driver to reduce unnecessary delays during the FreeBSD boot process

bllgg:
### Summary of changes

- Added TSLOGs to identify delay bottlenecks in the boot sequence.
- Introduced a new loader tunable `hw.atkbd.fast_delay` to `/boot/loader.conf`.
  - When enabled, the driver uses atkbd_short_delay, which applies significantly shorter delays in the following routines:
    - `empty_both_buffers()`
    - `reset_aux_dev()`
    - `test_controller()`
    - `reset_kbd()`
  - New short delays are up to 1000× smaller (for buffer resets) and about 40× smaller (for controller and keyboard resets).
  - Default behaviour remains with shorter delays, and `hw.atkbd.short_delay` should be set to 0, if someone needs longer delays for the old, slower hardware

#### Notes:
This patch is part of the ongoing GSoC 2025 work on FreeBSD boot process optimisation.


sys/dev/atkbdc/atkbdc.c
```diff
@@ -57,7 +57,10 @@
 
 /* constants */
 
-#define MAXKBDC		1		/* XXX */
+#define MAXKBDC				1	/* XXX */
+#define MILISECOND_MULTIPLIER		1000	/* micro second values will be multiplied to convert to mili seconds */
```
tsoome: typo, 'milli'. also, be mindful about line lengths, in general, we want to fit into 80 chars

sys/dev/atkbdc/atkbdc.c
```diff
@@ -80,6 +83,9 @@
 #define write_command(k, d)	\
 			(bus_space_write_1((k)->iot, (k)->ioh1, 0, (d)))
 
+#define ATKBD_DELAY(x)	(atkbd_short_delay ? (x) : (x) * MILISECOND_MULTIPLIER)
+#define RESET_DELAY(x)	(atkbd_short_delay ? (x) * MILISECOND_MULTIPLIER : (x) * MILISECOND_MULTIPLIER / RESET_DELAY_DIVISION_FACTOR)
```
tsoome: with long macro body we use \ to continue on next line and next line will be indented

sys/dev/atkbdc/atkbdc.c
```diff
@@ -1025,8 +1031,8 @@ reset_kbd(KBDC p)
     if (retry < 0)
         return FALSE;
 
-    int delay_us = 5000;
-    int max_wait_us = KBD_RESETDELAY*1000*KBD_MAXWAIT;
+    int delay_us = RESET_DELAY(KBD_RESETDELAY); /* if the atkbd_short_delay is activated, delay will be shorten */
```
tsoome: line length there too

PR #1827, opened 2025-08-28
ng_device: enable FIONBIO ioctl to set non-blocking mode

Defenso-QTH:
Also replace `vnode.h`'s `IO_NDELAY` by `fnctl.h`'s `O_NONBLOCK` to follow the current convention.

### DIFF
--- a/sys/netgraph/ng_device.c
+++ b/sys/netgraph/ng_device.c
@@ -32,7 +32,7 @@
  */
 
 #if 0
-#define	DBG do { printf("ng_device: %s\n", __func__ ); } while (0)
+#define	DBG do { printf("ng_device: %s\n", __func__); } while (0)
 #else
 #define	DBG do {} while (0)
 #endif
@@ -150,9 +150,11 @@ static struct cdevsw ngd_cdevsw = {
 	.d_name =	NG_DEVICE_DEVNAME,
 };
 
-/******************************************************************************
+/*
+ *****************************************************************************
  *  Netgraph methods
- ******************************************************************************/
+ *****************************************************************************
+ */
 
 /*
  * Handle loading and unloading for this node type.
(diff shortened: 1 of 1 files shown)


bsdimp (approved):
The ioctl function might not be formatted quite right, but that's hard to see on this interface.

sys/netgraph/ng_device.c
```diff
@@ -51,7 +51,7 @@
 #include <sys/syslog.h>
 #include <sys/systm.h>
 #include <sys/uio.h>
-#include <sys/vnode.h>
+#include <sys/fcntl.h>
```
markjdb: The header should be sorted after conf.h.

sys/netgraph/ng_device.c
```diff
@@ -397,6 +394,20 @@ ngdclose(struct cdev *dev, int flag, int mode, struct thread *td)
...
+ * Process IOCTLs
+ *
+ * At this stage we only return success on FIONBIO to allow setting the device
+ * as non-blocking.
+ *
+ */
+static int
+ngdioctl(struct cdev *dev, u_long cmd, caddr_t addr, int flag, struct thread *td)
+{
+    if (cmd == FIONBIO)
```
bsdjhb: This should really use a switch on `cmd`, and if you want to support `fcntl(F_SETFL)` and not just `fcntl(FIONBIO)` you have to support both `FIONBIO` and `FIOASYNC` and reject values that aren't supported.

See the implementation of `F_SETFL` in sys/kern/kern_descrip.c:kern_fcntl() which always invokes both FIONBIO and FIOASYNC (even if the bits didn't change state) and rejects the entire `F_SETFL` operation if either of those fail.

bsdimp:
I tagged this as 'changes required' for the switch() thing. The rest would be nice, but isn't gating this change.

sys/netgraph/ng_device.c
Defenso-QTH: Fixed, along with a couple of others that weren't in order.

sys/netgraph/ng_device.c
Defenso-QTH: Done. Thanks a lot to @bsdjhb and his _Character Device Driver Tutorial_ that helped me understand this better too.

Defenso-QTH:
Thanks everyone for your comments.
@bsdimp it should be fine now.
I also added a commit for style using the `checkstyle9.pl` script.

bsdimp:

bsdimp:
Thanks jhb and @Defenso-QTH . Landed, lemme know if I messed up something.

PR #1828, opened 2025-08-28
ipfw: Add option for firewall_type to be a directory

thegushi:
If specified, this will run rcorder over a directory.  A small subset of examples is also provided for /usr/share/examples

### DIFF
--- a/share/man/man5/rc.conf.5
+++ b/share/man/man5/rc.conf.5
@@ -692,11 +692,13 @@ basic protection for a LAN.
 .El
 .Pp
 If a filename is specified, the full path
-must be given.  If a directory is specified, then
-the directory should contain files containing headers
-compatible with rcorder(8) as they will be sorted
-in this order, and then ipfw(8) will be run on each,
-as though each were a singular rules file.
+must be given.
+If a directory is specified, then the
+directory should contain files containing headers compatible with
+.Xr rcorder 8
+as they will be sorted in this order, and then
+.Xr ipfw 8
+will be run on each, as though each were a singular rules file.
 .Pp
 Most of the predefined rulesets define additional configuration variables.
 These are documented in


cperciva:
IIRC this was discussed a bit at BSDCan with @allanjude .

allanjude:
> IIRC this was discussed a bit at BSDCan with @allanjude .
Indeed it was. ScaleEngine happened to independently develop basically the exact same solution

cperciva:
@allanjude Can you take care of landing this (either Dan's version or the ScaleEngine one)?  As release engineer I'm fine with this going into 15.0 since it won't change the behaviour for existing configurations.

thegushi:
I put in a few files to go into the examples directory, but realistically the whole point of this is that users are going to drop their own in.

But because the stock firewall scripts have some good examples for things that you want to do like Bogan blocking, and the rest of that it felt like that was useful and also an example of how a good way to set up pseudo services is.  Just like the regular RC.D system has pseudo services.

thegushi:
And yes, the two examples I didn't mention, but did in my talk or: if you're using something like puppet to deploy services, it means that one file can come in with the service then it needs, and also that now since you don't have a model at the firewall file, if a rule fails to load, it doesn't crash the entire stack.

libexec/rc/rc.firewall
```diff
@@ -547,7 +547,16 @@ case ${firewall_type} in
...
-		${fwcmd} ${firewall_flags} ${firewall_type}
+		if [ -f "${firewall_type}" ]; then
+			${fwcmd} ${firewall_flags} ${firewall_type}
+		else
+		if [ -d "${firewall_type}" ]; then
+			for fwfile in `rcorder $firewall_type/*`
+			do
+				ipfw -q $fwfile;
+			done
+			fi
```
markjdb, line 558: The indentation looks wonky here, or is that just github rendering it incorrectly somehow?

libexec/rc/rc.firewall
```diff
@@ -547,7 +547,16 @@ case ${firewall_type} in
 	;;
 *)
 	if [ -r "${firewall_type}" ]; then
-		${fwcmd} ${firewall_flags} ${firewall_type}
+		if [ -f "${firewall_type}" ]; then
+			${fwcmd} ${firewall_flags} ${firewall_type}
+		else
+		if [ -d "${firewall_type}" ]; then
```
markjdb, line 553: We need to update share/man/man5/rc.conf.5 to note that `firewall_type` can be a directory.

libexec/rc/rc.firewall
thegushi: I may have messed it up.  The style in this file is odd -- things are indented with hard-tabs inside a function (but only *sometimes*?), but two-spaces inside a for loop?  I had used a tab inside the for

Here's a screenshot in Nano with whitespace display turned on.  (Also look at the function above this one)

image

But go look at the original file, it's just as weird and other than the "for" loop it's all hard tabs.  This file is very inconsistent.

libexec/rc/rc.firewall
thegushi: That I can do -- this change came out of the way we had modded this file to meet our needs, and obviously, we didn't tweak the docs for our own use.  Coming shortly.

libexec/rc/rc.firewall
thegushi: For context, here's the original file.  Also tabs inside an if-block.

image

markjdb: Ok, I think this is fine then.

thegushi:
I've added the manpage changes.  I don't know how to "sign off" on where I synced the branch with upstream, or if I should remove that from the pull, somehow?

share/man/man5/rc.conf.5
```diff
@@ -692,7 +692,11 @@ basic protection for a LAN.
 .El
 .Pp
 If a filename is specified, the full path
-must be given.
+must be given.  If a directory is specified, then
+the directory should contain files containing headers
+compatible with rcorder(8) as they will be sorted
+in this order, and then ipfw(8) will be run on each,
+as though each were a singular rules file.
```
concussious:
```suggestion
must be given.
If a directory is specified, then the
directory should contain files containing headers compatible with
.Xr rcorder 8
as they will be sorted in this order, and then
.Xr ipfw 8
will be run on each, as though each were a singular rules file.
```
+ One sentence per line in roff
+ The Xr macro makes links clickable in graphical manual viewers
+ Also recommend joining line 694/695
+ In the future you can verify your syntax with `mandoc -Tlint /here/is/the/manual`.

libexec/rc/rc.firewall
thegushi: Done (modulo some fighting with github)

thegushi:
At this point, with the suggested man changes, I believe this can be merged, please?

share/man/man5/rc.conf.5
concussious: This is still not done

thegushi: I thought I had applied the suggestion, per github?

concussious: Doesn't seem like it from here. You can test it with `gh pr checkout 1828`

thegushi:
Okay, I've used the Github web UI to accept those changes into my branch.  I couldn't figure out how to view those suggestions with the gh command line, and I was confused about adding a signed-off-by line on those since those lines of code are your work, not mine.  But either way, the corrected man pages should now be part of the source.

concussious:
Approved from manpages, cc @allanjude, @markjdb

share/examples/ipfw.d/setup
```diff
@@ -0,0 +1,47 @@
...
+
+# deny unexpected sources
+add deny ip from table(bogons) to me in // unexpected sources
+
+# deny explicitly disabled (non-persistent) sources
+add deny ip from table(blocked) to me in // emergency (non-persistent) blocklist
+
+# allow bsd-standard-port traceroutes
+add allow udp from me to any 33434-33600 // traceroute in
+add allow udp from any to me 33434-33600 // traceroute out
```
jlduran, line 38: I would prefer:
```
add unreach port udp from any to me dst-port 33434-33626
```
`traceroute` expects the host to return a destination port unreachable (ICMP type 3, code 3), by default starting on port $`2^{15} + 666 = 33434`$ and up to `net.inet.ip.ttl` (64 by default) hops, each with 3 probes by default ($`33434 + 64 \times 3 = 33626`$).

PR #1832, opened 2025-08-30
ntpd: fix rc.d to pass rclint

thegushi:
Minor (mainly cosmetic and line-ordering) fixes such that ntpd's startup script passes muster by devel/rclint.  Still two warnings are emitted about chown possibly being replaced by install, but that change (fixup of a directory in base) doesn't make sense in this context.


thegushi:
Withdrawing this for now, until 15.0 RELEASE is out.

PR #1834, opened 2025-09-01
Update vmimage.subr to create fstab with noatime reduced the OPNsense constant 50kb - 80kb disk writes that wear down the SSD / NVME

Unicorn9x:
This "noatime" tweak, reduced the OPNsense constant 50kb - 80kb disk writes that wear down the SSD / NVME.

before
image

after
image

Signed-off-by: Unicorn9x

### DIFF
--- a/release/tools/vmimage.subr
+++ b/release/tools/vmimage.subr
@@ -82,7 +82,7 @@ vm_install_base() {
 	echo '# Custom /etc/fstab for FreeBSD VM images' \
 		> ${DESTDIR}/etc/fstab
 	if [ "${VMFS}" != zfs ]; then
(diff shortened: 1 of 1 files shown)


bsdimp (approved):
While there is some moderate value to having atime enabled on long running systems (I use it on my firewall test images to see what's used and to prune everything else, for example, before deploying the image), VM images aren't expected to be long-running enough for this to be useful. The default expectation is to treat them as disposable. The default for that use case should be to be as efficient as possible. Nothing will stop someone from removing the ',noatime' in /etc/fstab should their specific use-case require atimes.

markjdb:
I thought this would make UFS and ZFS inconsistent, but we already disable atime for ZFS images it seems (except for /var/mail)...

bsdimp:

PR #1839, opened 2025-09-05
eqos: removed unecessary code

Martinfx:
hi,
I don't find this phy-supply for gmac eth in actually rk356x.dtsi

### DIFF
--- a/sys/dev/eqos/if_eqos_fdt.c
+++ b/sys/dev/eqos/if_eqos_fdt.c
@@ -245,14 +245,6 @@ eqos_fdt_init(device_t dev)
 	    EQOS_GMAC_RXCLK_DLY_ENABLE |
 	    EQOS_GMAC_TXCLK_DLY_ENABLE);
 
-	if (!regulator_get_by_ofw_property(dev, 0, "phy-supply",
-	    &eqos_supply)) {
-		if (regulator_enable(eqos_supply))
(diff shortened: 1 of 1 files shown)


bsdimp:
Three problems that I see here.
(1) The commit messages don't say WHY this is OK. 
(2) The two commits should be one
(3) How do you know some other dtsi doesn't have this property?

@evadot touched this last, maybe he can say?

So at the very least these issues need to be addressed before we can proceed.

Martinfx:
> Three problems that I see here. (1) The commit messages don't say WHY this is OK. (2) The two commits should be one (3) How do you know some other dtsi doesn't have this property?
> 
> @evadot touched this last, maybe he can say?
> 
> So at the very least these issues need to be addressed before we can proceed.

It was in the kernel verbose message on my board bananapi-r2 pro

bsdimp:
> > Three problems that I see here. (1) The commit messages don't say WHY this is OK. (2) The two commits should be one (3) How do you know some other dtsi doesn't have this property?
> > @evadot touched this last, maybe he can say?
> > So at the very least these issues need to be addressed before we can proceed.
> 
> It was in the kernel verbose message on my board bananapi-r2 pro

But what about boards that have this property? Your change means that regulator_enable() never gets called.

Martinfx:
> But what about boards that have this property? Your change means that regulator_enable() never gets called.

Yes. I didn't think of that.... I am sorry

PR #1840, opened 2025-09-08
Update vmimage.subr to create fstab with noatime reduced the OPNsense constant 50kb - 80kb disk writes that wear down the SSD / NVME

Unicorn9x:
This "noatime" tweak reduced the OPNsense (FreeBSD 14.3-RELEASE-p2) constant 50kb - 80kb disk writes that wear down the SSD / NVME.

image

before
image

after
image

Signed-off-by: Unicorn9x


bsdimp:
#1834 has landed and this seems to be a duplicate.

PR #1841, opened 2025-09-09
freebsd-update.sh: improve the pkgbase statement

grahamperrin:
The wiki page for pkgbase is outdated, and quite overwhelming. Instead, direct readers to the FreeBSD Handbook.

Use the commonplace phrase 'pkgbase', not 'packaged base'.

Describe the incompatibility in a slightly different way.

### DIFF
--- a/usr.sbin/freebsd-update/freebsd-update.sh
+++ b/usr.sbin/freebsd-update/freebsd-update.sh
@@ -3616,8 +3616,9 @@ get_params "$@"
 # Disallow use with packaged base.
 if check_pkgbase; then
 	cat <<EOF
-freebsd-update is incompatible with the use of packaged base.  Please see
-https://wiki.freebsd.org/PkgBase for more information.
+Systems that use pkgbase can not use this version of freebsd-update.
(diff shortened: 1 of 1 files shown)


grahamperrin:
Currently: 

> freebsd-update is incompatible with …

I changed that part of the sentence partly because it doesn't quite fit with ongoing discussions about non-pkgbase jails that may be inferior to 15.0.

markmi:
[My note about the handbook's pkgbase /usr/src/ and /usr/src/sys/ material is now in the proper place, instead of here.]

grahamperrin:
Does https://github.com/freebsd/freebsd-src/pull/1841#issuecomment-3268472827 relate to the pkgbase statement in `freebsd-update.sh` in the `src` tree?

If you have not already seen it, there's a separate PR for the FreeBSD Handbook: 

- https://github.com/freebsd/freebsd-doc/pull/542

grahamperrin:
Mark, thanks again 👍

----

Housekeeping: this closure, and a few others.

PR #1842, opened 2025-09-09
sockstat: fix the -j option with piped output after libxo integration

asomers:
The legacy code handling -j in display() was causing xo_finish() to be skipped.  It has also been causing a memory leak since 0726c6574f8 (sockstat: Add automatic column sizing and remove -w option)

Fixes:		7b35b4d1963 (sockstat: add libxo support)
MFC after:	1 week
Reported by:	gleb
Sponsored by:	ConnectWise

### DIFF
--- a/usr.bin/sockstat/main.c
+++ b/usr.bin/sockstat/main.c
@@ -1612,7 +1612,7 @@ display(void)
 		}
 	}
 	if (opt_j >= 0)
-		return;
+		goto out;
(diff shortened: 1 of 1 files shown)


glebius:
Note: gleb@FreeBSD.org is a different person. I'm glebius@FreeBSD.org.

asomers:
Thanks again @glebius .  I fixed the commit message and committed.

PR #1847, opened 2025-09-15
Allow building with latest Clang HEAD

arichardson:
I uploaded many of these to phabricator, but for a lot of these small ones it seems like a lot of overhead, so I'm just pushing the full set of required changes here.


arichardson:
This should be ready now, I've built it locally targeting amd64 and risc64

arichardson:
Rebased, let me know if I should upload these to phabricator or if I can manually land the commits and link to this PR instead?

share/mk/bsd.sys.mk
```diff
@@ -93,6 +93,10 @@ CWARNFLAGS.clang+=	-Wno-error=unused-but-set-parameter
 # Similar to gcc >= 8.1 -Wno-error=cast-function-type below
 CWARNFLAGS.clang+=	-Wno-error=cast-function-type-mismatch
 .endif
+.if ${COMPILER_TYPE} == "clang" && ${COMPILER_VERSION} >= 210000
+# Until gtest is updated: https://github.com/google/googletest/issues/4762
```
arichardson, line 97: Looks like we could also apply https://github.com/google/googletest/commit/fa8438ae6b70c57010177de47a9f13d7041a6328 instead?

arichardson:
Confirmed locally that this allows me to build with LLVM HEAD as of today (only tested amd64).

jrtc27:
ZFS changes need to go upstream and be reviewed, especially when in files shared by Linux and FreeBSD. Also not sure I entirely understand the justification for them; is the problem that the callers are passing format strings that are using FreeBSD kprintf extensions but the attributes only say they follow GNU C? I'd have thought that would be rare, we don't have many extensions, and I'd be very surprised if generic ZFS code was using them given they won't work on Linux.

arichardson:
> ZFS changes need to go upstream and be reviewed, especially when in files shared by Linux and FreeBSD. Also not sure I entirely understand the justification for them; is the problem that the callers are passing format strings that are using FreeBSD kprintf extensions but the attributes only say they follow GNU C? I'd have thought that would be rare, we don't have many extensions, and I'd be very surprised if generic ZFS code was using them given they won't work on Linux.

Yeah the problem is that clang warns about passing a format string with a different format attribute: error:
`passing 'printf' format string where 'freebsd_kprintf' format string is expected [-Werror,-Wformat]`. I'm sure we could hack around that using a different approach but just changing the attribute seemed simpler

jrtc27:
Uh but where's the format string it's warning about coming from? Can you give a specific example?

arichardson:
If you look at e.g. sys/dev/ocs_fc/ocs_utils.h, those functions are all wrappers around the normal `*printf` and clang now warns about the format string using a different attribute. The warning is a bit annoying but it does make sense since they could interpret the arguments differently. We could silence this with `-Dfreebsd_kprintf=printf` but fixing it seems more appropriate.

PR #1848, opened 2025-09-16
man9: add VNET.9 and VIMAGE.9 aliases to vnet.9

Defenso-QTH:
Add capitalized aliases to improve discoverability.


concussious:
Okay why? This clutters apropos results badly.
Edit: the search itself is case sensitive, but it will print a listing for both capitalized and uncapitalized and then force the output to wrap awkwardly on narrow consoles.

Defenso-QTH:
> Okay why? This clutters apropos results badly. Edit: the search itself is case sensitive, but it will print a listing for both capitalized and uncapitalized and then force the output to wrap awkwardly on narrow consoles.

All the man pages referencing that one use uppercase `VNET`:
- `mac_ipacl(4)`: https://github.com/freebsd/freebsd-src/blob/8c9a919abaa3f6f16a17586ca72bab939788bf69/share/man/man4/mac_ipacl.4#L57
- `dummymbuf(4)`: https://github.com/freebsd/freebsd-src/blob/8c9a919abaa3f6f16a17586ca72bab939788bf69/share/man/man4/dummymbuf.4#L87 and https://github.com/freebsd/freebsd-src/blob/8c9a919abaa3f6f16a17586ca72bab939788bf69/share/man/man4/dummymbuf.4#L208
- `netintro(9)`: https://github.com/freebsd/freebsd-src/blob/8c9a919abaa3f6f16a17586ca72bab939788bf69/share/man/man9/intro.9#L221

So the user is likely to just type `man 9 VNET`. It also results in broken links on the online man pages: https://man.freebsd.org/cgi/man.cgi?query=mac_ipacl&apropos=0&sektion=0&manpath=FreeBSD+14.3-RELEASE+and+Ports&arch=default&format=html (click `VNET(9)` to check).

I did not realize about the side effect because I rarely use apropos. The resulting line is 79-character long. Isn´t the standard console width 80?

We could also shorten the title to `VNET, VIMAGE, vimage, vnet(9) - virtual network subsystem infrastructure` if that helps.

I agree the `VIMAGE` one is not as important too. Should I just remove that one?

concussious:
> All the man pages referencing that one use uppercase VNET

Ah, I see. Except `contrib/kyua/kyuafile.5.in`, which we should change to match the others actually.

This is actually a bug, this manual should be renamed to just VNET.9, there should be no vnet.9.

I would certainly accept a pull request with just those two changes.

> I did not realize about the side effect because I rarely use apropos. The resulting line is 79-character long. Isn´t the standard console width 80?

It is, but we should be careful to clutter it unnecessarily. A width of 59 is a lower priority target we also aim for if we can, because of split screen.

> We could also shorten the title to VNET, VIMAGE, vimage, vnet(9) - virtual network subsystem infrastructure if that helps.

Switching the location of `virtual` really changes the meaning here. If it's more accurate, then we should, but we'd have to ask someone from networking. That seems like a separate intent to what is going on here though.

> I agree the `VIMAGE` one is not as important too. Should I just remove that one?

I don't understand where VIMAGE is coming from, I don't see it in SYNOPSIS.

Defenso-QTH:
Thanks for the feedback and precisions on secondary goals. I agree changing the title / short description is probably a different issue.

I will make the two changes.

Regarding `VIMAGE`, it's the name of the kernel configuration option and is the first line of SYNOPSIS, you just missed it:

https://github.com/freebsd/freebsd-src/blob/5eb917426e0aab1e82c7b0e7239724d18bae206a/share/man/man9/vnet.9#L34-L35

concussious:
I did! Yes, it should definitely be `VIMAGE, VNET - network subsystem virtualization infrastructure`. Thank you.

Defenso-QTH:
Done. Ready for review.

Since the file is named `VNET.9` that is the first keyword that comes out in `apropos`. Is that OK? Or should we rename to `VIMAGE.9` and make `VNET.9` the alias?

```
VNET, VIMAGE(9) - network subsystem virtualization infrastructure
```

concussious (approved):

ngie-eign (approved):

ngie-eign:
> Failed to start an instance: FAILED_PRECONDITION: Monthly compute limit exceeded!

It looks like we're over quota even though we're only halfway through the month :/..

concussious:
Merged, thanks @Defenso-QTH and @ngie-eign!

> It looks like we're over quota even though we're only halfway through the month :/..

Afaik we just use the free tier since we have our own CI.

concussious:
Oops, I forgot to add the old manual to ObsoleteFiles.inc so it gets cleaned up. I added that in a later commit, 08d41d8803d9dd49b7.

PR #1849, opened 2025-09-17
atf_pytest: fix xfail detection from pytest report

svmhdvn:
The location of the 'wasxfail' attribute was moved from the 'reason' attribute back to the parent 'report'. This fixes an issue where xfails are wrongly reported to ATF as skipped tests.

### DIFF
--- a/tests/atf_python/atf_pytest.py
+++ b/tests/atf_python/atf_pytest.py
@@ -256,7 +256,7 @@ class ATFHandler(object):
                 # Record failure  & override "skipped" state
                 self.set_report_state(test_name, state, reason)
             elif state == "skipped":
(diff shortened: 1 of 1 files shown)


svmhdvn:
@AlexanderChernikov

kprovost:
Looks sane to me.

emaste:
I had a look for `wasxfail` in the tree and only found the three references in this file. I'm curious where it actually gets set.

svmhdvn:
It gets set in pytest's `pytest_runtest_makereport()` here: https://github.com/pytest-dev/pytest/blob/main/src/_pytest/skipping.py#L271

PR #1850, opened 2025-09-18
lib/libifconfig: Fix pointer-sign warnings

ryan-moeller:
Fix all pointer-sign warnings generated when compiling libifconfig.

While here, fix a small consistency nit in ifconfig_sfp_get_sfp_dump.

This eliminates all warnings in libifconfig, and may be trivially backported.

The server with my keys is packed away in storage, so I'd appreciate someone committing this small fix for me.


lib/libifconfig/libifconfig_carp.c
```diff
@@ -177,7 +177,7 @@ ifconfig_carp_set_info(ifconfig_handle_t *h, const char *name,
 	    &carpr->carpr_addr);
 	snl_add_msg_attr(&nw, CARP_NL_ADDR6, sizeof(carpr->carpr_addr6),
 	    &carpr->carpr_addr6);
-	snl_add_msg_attr_string(&nw, CARP_NL_KEY, carpr->carpr_key);
+	snl_add_msg_attr_string(&nw, CARP_NL_KEY, (caddr_t)carpr->carpr_key);
```
kprovost: That wants to be (const caddr_t).
I'll amend the commit.

kprovost (commented):
I found at least one build failure during make universe.
I'm re-building with a fix and if that works I'll push this.

lib/libifconfig/libifconfig_sfp.c
```diff
@@ -557,7 +557,7 @@ ifconfig_sfp_get_sfp_dump(ifconfig_handle_t *h, const char *name,
 	struct i2c_info ii;
 	uint8_t *buf = dump->data;
 
-	memset(dump->data, 0, sizeof(dump->data));
+	memset(buf, 0, sizeof(dump->data));
```
gmshake, line 560: No, the original one is more clear . I'd rather recommend

```
--- uint8_t *buf = dump->data;
+++ uint8_t *buf = &dump->data[0];
```

so that we have the context, and it is a little easier to read the code.

lib/libifconfig/libifconfig_carp.c
kprovost: (const char *) actually, for some reason llvm was unhappy about (const caddr_t), which doesn't make a lof of sense to me, but (const char *) is pretty clearly correct here so I'd just go with that.

markjdb: `const T`, where `T` is a typedef of `t *` is the same as `t * const`, i.e., a const pointer rather than a pointer to const. We have `c_caddr_t` for this purpose, but yeah it should just be `const char *`. We shouldn't be adding new uses of `caddr_t`.

markjdb: Are we sure that treating `carpr_key` as a string is right? It looks like netlink assumes it's nul-terminated, but as far as I can see it's not.

lib/libifconfig/libifconfig_sfp.c
ryan-moeller: As the author of this file, I prefer the terseness and consistency with the rest of the function and the rest of the file over ultimate explicit verbosity in this case. The way I wrote it is correct, and it minimizes the mental pointer dereferences for the reader. But it's just a nitpick, I'm not opposed to just dropping the change if it is contentious.

lib/libifconfig/libifconfig_carp.c
kprovost: ifconfig's carp.c does actually treat it as a string too, so I think it's correct. Arguably the structure definition should be changed to make it a char rather than uint8_t array.

ryan-moeller: I was concerned about that, too. The kernel treats the key as bytes, but userland is treating it as a string. This should probably be using `snl_add_msg_attr` along with `snl_attr_get_bytes`. It is also treated as a string in ifconfig. I can open another PR for those changes if someone else doesn't tackle it before I get there, to avoid scope creep in this PR.

markjdb: I think handling this in a separate PR is fine, thank you.

ryan-moeller: The key was supposed to be a sha1 hash of the passphrase (20 bytes, may contain NUL), but ifconfig has never done the hashing since CARP's port to FreeBSD, and always handles it as a string:

```
#define     CARP_KEY_LEN            20      /* a sha1 hash of a passphrase */
```

```
+       /* XXX Should hash the password into the key here, perhaps? */
+       strlcpy(carpr.carpr_key, val, CARP_KEY_LEN);
```

Looking at OpenBSD, it was never done there, either. It is a bit late to implement that, though. I'm afraid the ship has sailed, for 22 years.

gmshake: Why not just forget snl_add_msg_attr_string() and directly use snl_add_msg_attr() ?

```
--- snl_add_msg_attr_string(&nw, CARP_NL_KEY, carpr->carpr_key);
+++ snl_add_msg_attr(&nw, CARP_NL_KEY, sizeof(carpr->carpr_key), carpr->carpr_key);
```

So do not bother strlen(carpr->carpr_key), and do not worry  the \0 terminator, as a good effect no cast required.

lib/libifconfig/libifconfig_sfp.c
gmshake: > I prefer the terseness and consistency ..

I'm OK with that. The author knows better how the code is written, but the readers / reviewers do not.

I personally dislike such dereferences
```
uint8_t *buf = dump->data;
```
but always recommend
```
uint8_t *buf = &dump->data[0];
```
if possible.

Until a reader / reviewer grep `ifconfig_sfp_dump` he knows the `data` is an array but not a pointer. We ( the FreeBSD project ) have limited resources. So I'd appreciate if the readability is improved.

ryan-moeller:
I've updated the change in libifconfig_carp.c to treat carpr_key as bytes rather than a string. This fixes the pointer-sign warning and enables future work to properly hash the key in ifconfig, probably using an alternative CARP parameter to avoid breaking the current pass parameter behavior.

gmshake (approved):
LGTM.

bsdimp:

PR #1854, opened 2025-09-25
libc aarch64: fix ELF size of timingsafe_memcmp

paulfloyd:
Looks like a copy and paste error. The ELF size of 0 prevents Valgrind from redirecting this function.

See also
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=289845 and
https://bugs.kde.org/show_bug.cgi?id=509406


clausecker:
Fixed and MFC'ed.

PR #1856, opened 2025-09-25
tests/mac_ipacl: restore sysctls after modification

svmhdvn:
While here, use 'required_kmods' instead of ad-hoc checks.


markjdb:
Merged, thanks!

PR #1857, opened 2025-09-26
libcasper: explicit cast in cap_fileargs header

Defenso-QTH:
The implicit cast is incompatible with standard compiler options in a C++ project, making the library difficult to use.

The explicit cast is already the workaround in `libcasper.h`:
https://github.com/freebsd/freebsd-src/blob/7e8fb7756c3ed89a2141b923e6da1b6fd96f509c/lib/libcasper/libcasper/libcasper.h#L106-L108

Discussed with @kevans91


emaste:
Looks good, @VexedUXR do you want to push it to main? I can otherwise

VexedUXR:
> Looks good, @VexedUXR do you want to push it to main? I can otherwise

You can push it :)

PR #1859, opened 2025-09-28
libfetch, fetch: Improve handling of HTTP request body, method, fields, and status code

ryan-moeller:
Modern APIs often rely on HTTP fields as part of their interface, both
in requests and responses.

[lib/libfetch: Add fetchXReqHTTP(3)](https://github.com/freebsd/freebsd-src/commit/b54ab720456455decacc91ee52e349051797c25d) 

Add a function to libfetch allowing additional request fields to be
specified for an HTTP request and response fields to be returned to the
caller.

This function accepts an optional stream for the body content, as a
nul-terminated C string may be inadequate for some content.  If the
length of the stream can be determined, a Content-Length header is
added.  Otherwise, the body is sent chunk-encoded.

[usr.bin/fetch: Handle custom headers and method for HTTP](https://github.com/freebsd/freebsd-src/commit/af6163d844fce56230d09e91658cf7b72d8aa425) 

Add functionality to fetch allowing HTTP method and headers to be
specified.  The option names align with those used by curl.

[usr.bin/fetch: Add options for HTTP upload data](https://github.com/freebsd/freebsd-src/commit/9ac571a1e4eb4ce03c03d8131e4888c78c86635a) 

Add a -D/--data option to specify a file from which to read HTTP request
body contents, with the special case that "-" will read from stdin.

This option deviates from curl, taking a much more barebones approach
for simplicity.

Add a -C/--chunked option to specify that the HTTP request body contents
are to use chunked transfer-encoding.  Otherwise, it is automatically
used by libfetch when the size of the file cannot be predetermined.

[lib/libfetch: Accept all HTTP 2xx (Successful) status codes](https://github.com/freebsd/freebsd-src/commit/747ae417c56408fe33923ac42073b938e0373977) 

Some HTTP 2xx (Successful) class status codes were being treated as
errors.  Add a definition for the full range of successful status codes
and use it to test for a success response.

This allows libfetch and fetch to receive the response body from more
practical requests, such as POST responding 201 Created with content.

### Demonstration

Here is an example of the new functionality being used to access the GitHub REST API.  The API requires a valid bearer token in the Authorization header, along with a couple other recommended headers described in the documentation.

```sh
$ fetch -q \
    -X GET \
    -H "Accept: application/vnd.github+json" \
    -H "Authorization: Bearer $GITHUB_TOKEN" \
    -H "X-GitHub-Api-Version: 2022-11-28" \
    -o - \
    https://api.github.com/repos/ryan-moeller/kernel-nbd-client/branches
[{"name":"main","commit":{"sha":"d69b0a1c8d2b63ab48726027e6e756ccf9a0f2fb","url":"https://api.github.com/repos/ryan-moeller/kernel-nbd-client/commits/d69b0a1c8d2b63ab48726027e6e756ccf9a0f2fb"},"protected":false},{"name":"structured","commit":{"sha":"4831ee5ece393e596262d76609a0be69d3d408bf","url":"https://api.github.com/repos/ryan-moeller/kernel-nbd-client/commits/4831ee5ece393e596262d76609a0be69d3d408bf"},"protected":false},{"name":"unix/stream","commit":{"sha":"7f020b22ec0e9414a212fe34b01701b23443d825","url":"https://api.github.com/repos/ryan-moeller/kernel-nbd-client/commits/7f020b22ec0e9414a212fe34b01701b23443d825"},"protected":false}]
```

The (redundant, as it is the default) GET method is specified along with a valid token in an Authorization header, and the server responds with the requested data.

Specifying a different HTTP method is allowed by fetch(1) and rejected by the server as expected:

```sh
$ fetch -q \
    -X TEST \
    -H "Accept: application/vnd.github+json" \
    -H "Authorization: Bearer $GITHUB_TOKEN" \
    -H "X-GitHub-Api-Version: 2022-11-28" \
    -o - \
    https://api.github.com/repos/ryan-moeller/kernel-nbd-client/branches
fetch: https://api.github.com/repos/ryan-moeller/kernel-nbd-client/branches: Forbidden
```

The server does not respond to requests with an invalid token:

```sh
$ fetch -q \
    -H "Accept: application/vnd.github+json" \
    -H "Authorization: Bearer not$GITHUB_TOKEN" \
    -H "X-GitHub-Api-Version: 2022-11-28" \
    -o - \
    https://api.github.com/repos/ryan-moeller/kernel-nbd-client/branches
fetch: https://api.github.com/repos/ryan-moeller/kernel-nbd-client/branches:
$ echo $?
1
```

This is not the most compelling example, as this is a read-only use of the GitHub API to access public data that would have been allowed with no token.  There are of course other API endpoints with stricter access controls where the headers are actually necessary.  And this functionality becomes even more useful with support for fetch(1) uploading a request body. This enables use of common HTTP methods like PUT, POST, and PATCH.

For example, uploading a gist to GitHub uses most of the new features:

```sh
$ fetch -q \
    -X POST \
    -H "Authorization: Bearer $GITHUB_TOKEN" \
    -H "Accept: application/vnd.github+json" \
    -H "Content-Type: application/json" \
    -D - \
    -o - \
    https://api.github.com/gists <<EOF
{
  "description": "fetch(1) test gist",
  "public": false,
  "files": {
    "demo.txt": {
      "content": "This gist was created using fetch(1)"
    }
  }
}
EOF
<verbose JSON response omitted>
```

`-X POST` tells fetch to use the POST method for the HTTP request
`-H ...` adds custom headers to the request
`-D -` sends a request body from a file (stdin, in this case)

### Testing

FreeBSD has no tests for libfetch or fetch.  I didn't go so far as to implement a Kyua test suite for inclusion in FreeBSD, but I do have some ancillary testing apparatus used to exercise my changes.

My test harness for this uses a Lua library wrapping libfetch with a Lua script exercising all request and response field sections as well as different request and response body transfer encodings and different success status codes. This can all be found [here](https://github.com/ryan-moeller/flualibs/tree/73b2f602b9dc2b9f889f5f653b60bc458be2bc1d/libfetch).

CI runs: https://cirrus-ci.com/build/5568610090876928
15.0-ALPHA4 test output: https://api.cirrus-ci.com/v1/task/5673940506378240/logs/libfetch_test.log

### DIFF
--- a/lib/libfetch/common.h
+++ b/lib/libfetch/common.h
@@ -175,9 +175,6 @@ int		 fetch_no_proxy_match(const char *);
  */
 FILE		*http_request(struct url *, const char *,
 		     struct url_stat *, struct url *, const char *);
-FILE		*http_request_body(struct url *, const char *,
-		     struct url_stat *, struct url *, const char *,
-		     const char *, const char *);
 FILE		*ftp_request(struct url *, const char *,
 		     struct url_stat *, struct url *, const char *);
 
--- a/lib/libfetch/fetch.h
+++ b/lib/libfetch/fetch.h
@@ -31,7 +31,7 @@
 #ifndef _FETCH_H_INCLUDED
 #define _FETCH_H_INCLUDED
 
-#define _LIBFETCH_VER "libfetch/2.0"
+#define _LIBFETCH_VER "libfetch/2.1"
 
 #define URL_SCHEMELEN 16
 #define URL_USERLEN 256
@@ -60,6 +60,15 @@ struct url_ent {
 	char		 name[PATH_MAX];
 	struct url_stat	 stat;
 };
+
+#ifdef _SYS_QUEUE_H_
+struct http_field {
+	const char	*name;
+	const char	*value;
+	TAILQ_ENTRY(http_field)	fields;
+};
+TAILQ_HEAD(http_fields, http_field);
+#endif
 
 /* Recognized schemes */
 #define SCHEME_FTP	"ftp"
@@ -105,6 +114,13 @@ int		 fetchStatHTTP(struct url *, struct url_stat *, const char *);
 struct url_ent	*fetchListHTTP(struct url *, const char *);
 FILE		*fetchReqHTTP(struct url *, const char *, const char *,
 		    const char *, const char *);
+#ifdef _SYS_QUEUE_H_
+FILE		*fetchXReqHTTP(struct url *, struct url_stat *, const char *,
+		    const char *, const struct http_fields *,
+		    const struct http_fields *, struct http_fields *,
+		    struct http_fields *, FILE *);
+void		 fetchFreeResFields(struct http_fields *);
+#endif
 
 /* FTP-specific functions */
 FILE		*fetchXGetFTP(struct url *, struct url_stat *, const char *);
--- a/usr.bin/fetch/fetch.c
+++ b/usr.bin/fetch/fetch.c
@@ -30,6 +30,7 @@
  */
 
 #include <sys/param.h>
+#include <sys/queue.h>
 #include <sys/socket.h>
 #include <sys/stat.h>
 #include <sys/time.h>
@@ -39,6 +40,7 @@
 #include <errno.h>
 #include <getopt.h>
 #include <signal.h>
+#include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -85,7 +87,10 @@ static int	 v_tty;		/*        stdout is a tty */
 static int	 v_progress;	/*        whether to display progress */
 static pid_t	 pgrp;		/*        our process group */
 static long	 w_secs;	/*    -w: retry delay */
+static char	*method;	/*    -X: HTTP method */
 static int	 family = PF_UNSPEC;	/* -[46]: address family to use */
+static struct http_fields headers = TAILQ_HEAD_INITIALIZER(headers);
+				/*    -H: HTTP headers */
 
 static int	 siginfo;	/* SIGINFO received */
 static int	 sigint;	/* SIGINT received */
@@ -126,6 +131,7 @@ static struct option longopts[] =
 	{ "direct", no_argument, NULL, 'd' },
 	{ "force-restart", no_argument, NULL, 'F' },
 	/* -f not mapped, since it's deprecated */
+	{ "header", required_argument, NULL, 'H' },
 	/* -h not mapped, since it's deprecated */
 	{ "if-modified-since", required_argument, NULL, 'i' },
 	{ "symlink", no_argument, NULL, 'l' },
@@ -145,6 +151,7 @@ static struct option longopts[] =
 	{ "passive-portrange-default", no_argument, NULL, 'T' },
 	{ "verbose", no_argument, NULL, 'v' },
 	{ "retry-delay", required_argument, NULL, 'w' },
+	{ "request", required_argument, NULL, 'X' },
 
 	/* options without a single character equivalent */
 	{ "bind-address", required_argument, NULL, OPTION_BIND_ADDRESS },
@@ -415,10 +422,22 @@ query_auth(struct url *URL)
 }
 
 /*
+ * Decide how to invoke libfetch
+ */
+static FILE *
+invoke(struct url *url, struct url_stat *us, const char *flags, bool is_http)
+{
+	if (is_http)
+		return (fetchXReqHTTP(url, us, method ? method : "GET", flags,
+		    &headers, NULL, NULL, NULL, NULL));
+	return (fetchXGet(url, us, flags));
+}
+
+/*
  * Fetch a file
  */
 static int
-fetch(char *URL, const char *path, int *is_http)
+fetch(char *URL, const char *path, bool *is_http)
 {
 	struct url *url;
 	struct url_stat us;
@@ -565,7 +584,7 @@ again:
 	size_prev = sb.st_size;
 
 	/* start the transfer */
-	f = fetchXGet(url, &us, flags);
+	f = invoke(url, &us, flags, *is_http);
 	if (f == NULL) {
 		if (i_flag && *is_http && fetchLastErrCode == FETCH_OK &&
 		    strcmp(fetchLastErrString, "Not Modified") == 0) {
@@ -693,7 +712,7 @@ again:
 			 * from scratch if we want the whole file
 			 */
 			url->offset = 0;
-			if ((f = fetchXGet(url, &us, flags)) == NULL) {
+			if ((f = invoke(url, &us, flags, *is_http)) == NULL) {
 				warnx("%s: %s", URL, fetchLastErrString);
 				goto failure;
 			}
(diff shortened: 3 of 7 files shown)


lib/libfetch/fetch.h
```diff
@@ -61,6 +61,16 @@ struct url_ent {
 	struct url_stat	 stat;
 };
 
+#ifdef _SYS_QUEUE_H_
```
ryan-moeller, line 64: The `#ifdef _SYS_QUEUE_H_` is a compromise. Currently, fetch.h includes none of the headers it relies on. Rather than changing this, or changing every consumer to include the sys/queue.h header, I added checks for it to only expose the parts that rely on it when the consumer has included it.

I'm open to handling this differently.

lib/libfetch/common.h
```diff
@@ -161,9 +161,6 @@ int		 fetch_no_proxy_match(const char *);
  */
 FILE		*http_request(struct url *, const char *,
 		     struct url_stat *, struct url *, const char *);
-FILE		*http_request_body(struct url *, const char *,
```
ryan-moeller, line 178: This function is not used outside http.c so while changing the parameters I also made it static.

usr.bin/fetch/fetch.c
```diff
@@ -956,8 +994,7 @@ main(int argc, char *argv[])
 			f_filename = optarg;
 			break;
 		case 'H':
-			warnx("the -H option is now implicit, "
```
ryan-moeller, line 945: Seems worth replacing this warning with useful functionality. -H matches curl.

lib/libfetch/fetch.h
```diff
@@ -105,6 +115,12 @@ int		 fetchStatHTTP(struct url *, struct url_stat *, const char *);
 struct url_ent	*fetchListHTTP(struct url *, const char *);
 FILE		*fetchReqHTTP(struct url *, const char *, const char *,
 		    const char *, const char *);
+#ifdef _SYS_QUEUE_H_
+FILE		*fetchXReqHTTP(struct url *, struct url_stat *, const char *,
+		    const char *, const struct http_headers *,
+		    struct http_headers *, struct iovec *, int);
```
ryan-moeller: I'm having second thoughts about this taking an array of iovecs as the body. While it enables functionality not available through the previous interface, what I really want is something that also supports streaming.  A FILE * fits the requirements nicely, as it is relatively easy to construct one from a C string in memory to implement the existing fetchReqHTTP interface, it supports predetermining the content length or failing to fast, and I don't have to reinvent the wheel.

I'll take a stab at this approach and report back.

ryan-moeller: This is working great, except I need to be a little more thorough determining whether to send a content-length header or switch to chunked transfer encoding.  Using stdin as the file causes the seek/tell to succeed and set the length to 0 when it should be chunked and streamed, but at the same time an actual file that is actually 0 bytes should also avoid switching to chunk-encoding for nothing.

Easily fixable, so one more update in the works.

ryan-moeller: I'm happy with the use of `FILE *` for the body now.

ryan-moeller:
* Switched from using a `struct iovec` array for the body to support binary content, to using a `FILE *` and implementing automatic (or forced) chunked transfer-encoding. This enables streaming data uploads (in libfetch, not yet in fetch). Trailing headers remain unimplemented.

ryan-moeller:
* Documented the 'C' flag in fetch.3

ryan-moeller:
* Implemented body upload options for fetch(1)

ryan-moeller:
* Fixed sorting of additions to fetch.3 so the convention of fetchXFoo being listed before fetchFoo is maintained

ryan-moeller:
It would be best to implement support for HTTP trailer fields in libfetch while I'm here. Taking both headers and trailers into consideration, we would prefer to use more general terms like `struct http_field`, `fields`, `struct http_fields`, and `fetchFreeResFields`.  Two additional optional parameters to `fetchXReqHTTP` would provide the request trailers and receive the response trailers.

I have started implementing this change and will update this pull request after I finish testing the idea.

ryan-moeller:
* Revised names to more use generic "field" terms in libfetch
* Implemented request and response trailer fields in libfetch

I have a Lua binding for libfetch and a test server scripted with socat I've been testing with. I'll share more of the test harness  details when I have some time.

ryan-moeller:
> I have a Lua binding for libfetch and a test server scripted with socat I've been testing with. I'll share more of the test harness details when I have some time.

* Brief description of testing apparatus added to the opening comment

ryan-moeller:
* Send "Content-Length: 0" when there is no request message body

I ran into an existing bug in libfetch during testing. When sending a request with no body, libfetch would not send a content-length header. This, combined with libfetch sending "Connection: close" and not sending a transfer-encoding header, means the message is delimited by connection close. However, libfetch intentionally avoids actually shutting down the write half of the connection before waiting for the response, for interoperability with some proxies that forbid half-closed connections.

In practice, many servers use heuristics the work around such misbehavior by clients. But for correctness and for better operation with more servers, we should set the content-length to 0 when we have no body to send.  This lets the server know where the request ends without relying on workarounds (such as a timeout). Otherwise, a server would be waiting for the connection close to terminate the request body, and libfetch would be waiting for a response without ever closing the connection.

ryan-moeller:
On closer inspection, https://www.rfc-editor.org/rfc/rfc9112#section-6.3-4.1 notes that request messages are never close-delimited. The previous commit has been removed.

ryan-moeller:
I've just noticed that the group of HTTP response codes to treat as success in libfetch is missing a few.  I'll be back with another revision to treat e.g. `201 Created` as success and proceed to read the rest of the response. That will inspire some new test cases, as well.

ryan-moeller:
* Allow libfetch and fetch to receive the response body for more status codes, such as 201 Created

ryan-moeller:
* nit: In the first commit, make `fetchReqHTTP` use `fetchXReqHTTP` instead of `http_request_body`, like `fetchGetHTTP` uses `fetchXGetHTTP` instead of `http_request`.

ryan-moeller:
* Rebased

ryan-moeller:
I have another small commit extending the existing handling of *Range* and *Content-Range* headers in libfetch to better support range requests and partial responses according to [RFC 9110 Section 14][1]. However, I'm reluctant to continue expanding the original scope of this pull request. I'll hold off on pushing that commit unless requested, otherwise I can save it for another pull request.

[1]: https://www.rfc-editor.org/rfc/rfc9110#name-range-requests

bsdimp:
This is really cool! My apologies for not noticing it earlier. I saw the comments and assumed others were looking at it. Sadly, this assumption was bad, leading to this being ignored until now. My apologies.

I've wanted many of these features, and often install curl to get them. Our product at work does too, but I fear that curl is a bit too entrenched there to migrate away from (though curl has been a bit of a thorn at times).

I agree that this pull request already is quite full...  I need to look closely to give advice on how to proceed, but wanted to give feedback...

ryan-moeller:
Thanks! Let me know if you need anything.

bsdimp:
Added des for feedback. I thought I'd done that before, but apparently not.

ryan-moeller:
* Rebased

@bapt I came across https://ron-dev.freebsd.org/FreeBSD/src/pulls/10 so you may also be interested in this.

ryan-moeller:
* Rebase

PR #1861, opened 2025-09-28
mediatek 7622 interrupt

priittrees:
Added missing pic methods. 
fix: gic0: unsupported trigger/polarity configuration 0x08
Replaced IRQ type numbers with names.


priittrees:
Sorry! I posted this in the wrong place.

PR #1862, opened 2025-09-29
Update the Github action to use LLVM 15

zxombie:
This is the latest supported version in Ubuntu 22.04 Sponsored by:	Arm Ltd


VexedUXR (approved):
I think it's reasonable to stop caring about clang-14, given that the errors it has been producing recently would affect the quality of the code if fixed.

bsdimp:
I think it's fine to drop the clang-14 stuff, given that it's no longer supported here...

emaste (approved):

PR #1922, opened 2025-12-10
speaker(4): drop NEEDGIANT and modernize the driver

knz:
**Summary:** the calls to `disable_intr()` / `enable_intr()` in `spkr.c` were leftover from previous refactorings and unnecessary for correctness, so this PR removes them. There are also a few other improvements to bring this driver to the 21th century.

My detailed analysis of why this change is correct can be found here:  [spkr.md](https://github.com/user-attachments/files/24087193/spkr.md)

The main related commit referenced in the analysis is this 2008 change by phk: e46598588587b4897f6604489364f83fffd4d033 - where a mutex was added to clock.c and the disable_intr/enable_intr pair could have been removed, but wasn't.

**Motivation:** I am a fervent spkr(4) user and would like it to persist beyond FreeBSD 16, so we needed to do something about its use of the GIANT mutex.

cc @bsdphk @bsdimp

### DIFF
--- a/sys/dev/speaker/spkr.c
+++ b/sys/dev/speaker/spkr.c
@@ -15,7 +15,6 @@
 #include <sys/conf.h>
 #include <sys/ctype.h>
 #include <sys/malloc.h>
-#include <machine/atomic.h>
 #include <machine/clock.h>
 #include <dev/speaker/speaker.h>
 
@@ -51,11 +50,12 @@ static MALLOC_DEFINE(M_SPKR, "spkr", "Speaker buffer");
 #define SPKRPRI PSOCK
 static char endtone, endrest;
 
+struct spkr_state;
 static void tone(unsigned int thz, unsigned int centisecs);
 static void rest(int centisecs);
-static void playinit(void);
-static void playtone(int pitch, int value, int sustain);
-static void playstring(char *cp, size_t slen);
+static void playinit(struct spkr_state *state);
+static void playtone(struct spkr_state *state, int pitch, int value, int sustain);
+static void playstring(struct spkr_state *state, char *cp, size_t slen);
 
 /* 
  * Emit tone of frequency thz for given number of centisecs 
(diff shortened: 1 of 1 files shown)


sys/dev/speaker/spkr.c
```diff
@@ -428,14 +433,16 @@ spkrwrite(struct cdev *dev, struct uio *uio, int ioflag)
 		char *cp;
 		int error;
 
+		mtx_lock(&spkr_op_locked);
```
kevans91: Looking at this again, I don't think this works as-written?  spkr_op_locked should probably be an sx(9), because I imagine uiomove(9) will throw a fit under WITNESS as it may try to sleep on copyin.

knz: Cool, thanks. TIL about sx(9) and how mtx works under the hood.

Done.

knz:
@kevans91 if you're open to it, I'm willing to import this commit into the current PR as well:
https://github.com/knz/freebsd-src/commit/e1d982a4eeab1de00ddf7d8d37208dca091120fb

knz:
> @kevans91 if you're open to it, I'm willing to import this commit into the current PR as well: [knz@e1d982a](https://github.com/knz/freebsd-src/commit/e1d982a4eeab1de00ddf7d8d37208dca091120fb)

> [kevans:  Go ahead and add it to the PR](https://discord.com/channels/727023752348434432/757305573866733680/1460331830795567214)

Done.

sys/dev/speaker/spkr.c
```diff
@@ -390,53 +401,82 @@ playstring(char *cp, size_t slen)
...
+		return error;
+	}
+
+	return 0;
 }
 
 static int
 spkrwrite(struct cdev *dev, struct uio *uio, int ioflag)
 {
+	int error;
```
kevans91: Please flip the ordering around to be more style(9)-ish -- size descending, then alphabetical

sys/dev/speaker/spkr.c
```diff
@@ -390,53 +401,82 @@ playstring(char *cp, size_t slen)
...
+
+	/* initialize playstring state */
+	playinit(state);
+
+	/* store in the current fd private data */
+	error = devfs_set_cdevpriv(state, spkr_dtor);
+	if (error) {
+		free(state->inbuf, M_SPKR);
+		free(state, M_SPKR);
+		return error;
```
kevans91:
```suggestion
		return (error);
```
Pleases apply to the other returns as well, both literal and of non-literal expressions

kevans91 (commented):
I'm going to hand this off to @bsdimp to render a final judgement; I think the multiple-open thing is probably fine and not likely to cause any real problems, and the rest of it seems technically sound.  Just some style nits remaining, I think.

sys/dev/speaker/spkr.c
knz: Done,

sys/dev/speaker/spkr.c
knz: Done.

knz:
@gvnn3 👋

bsdimp:

PR #1955, opened 2026-01-09
pcpu_aux: refine error message and add more checks

mchoo7:
This refines error message and adds more checks.

When padding needs to be increased so the total size of `pcpu` is next factor of `PAGE_SIZE` (4096 bytes), current static assertion doesn't show `sizeof(pcpu)` in number and thus developers doesn't know the next factor of page size unless they print `sizeof(pcpu)`. This check always guarantees that `pcpu` is as small as possible while being a factor page size.

I didn't touch `i386/include/pcpu_aux.h` since i386 is now deprecated and the build is broken.

### DIFF
--- a/sys/amd64/include/pcpu.h
+++ b/sys/amd64/include/pcpu.h
@@ -101,7 +101,7 @@ _Static_assert(sizeof(struct monitorbuf) == 128, "2x cache line");
 	u_int	pc_small_core;						\
 	u_int	pc_pcid_invlpg_workaround;				\
 	struct pmap_pcid pc_kpmap_store;				\
-	char	__pad[2900]		/* pad to UMA_PCPU_ALLOC_SIZE */
+	char	__pad[2928]		/* pad to UMA_PCPU_ALLOC_SIZE */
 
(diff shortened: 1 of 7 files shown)


mchoo7:
With this fix, I could find and fix a bug in commit 85c85a72285b1ba426e2748f866e13b5f316fbe5.

Because of unnecessary padding in aarch64, it used 256 more bytes for `struct pcpu`. Now it only uses 256 bytes (from 512).

bsdimp (commented):
I think this is good, but not 100% seems like if it compiles we're good.

mchoo7:
> I think this is good, but not 100% seems like if it compiles we're good.

I couldn't build kernel on all architectures today. I'll notify you tomorrow when build succeeds on all `TARGET_ARCH`s.

mchoo7:
@bsdimp Build succeeds on all `TARGET_ARCH`s with GENERIC conf.

bsdimp:
OK. I think this is ready...

mchoo7:
@bsdimp Is this ready for merge?

PR #1956, opened 2026-01-10
Remove i386 target (was broken)

mchoo7:
i386 is unsupported since FreeBSD 15 and today I confirmed that buildkernel is broken on `TARGET_ARCH=i386`. We still have compat layers for 32-bit x86, so I'm only removing i386 target, specific Makefiles (`Makefile.i386`), and release scripts here.

`sys/i386` could be removed maybe, but I need to test amd64 buildkernel for that.

Remaining i386 stuff in most Makefiles and source code will take long time to clean up...


bsdimp:
I'm not sure we're ready for this one...  We'll talk about it in srcmgr, but may be a month

mchoo7:
Superseded by #2052

PR #1958, opened 2026-01-10
bhyve: Use getaddrinfo(3) consistently for AF_UNIX.

yuichiro-naito:
The FreeBSD getaddrinfo(3) function also supports the AF_UNIX address family. Use it the same way as inet and inet6 so that we can share the same code. However, the getaddrinfo(3) accepts an absolute path only. The path string that follows 'unix:' will be resolved by the realpath(3) function.

The freeaddrinfo(3) function also supports a NULL argument, as does free(3). No need to check a NULL pointer.

While I'm here, the UNIX domain socket should allow all users to access it as if it were a TCP port. Changes umask to 0 before bind(2), and restores the previous value after the bind(2).

### DIFF
--- a/usr.sbin/bhyve/rfb.c
+++ b/usr.sbin/bhyve/rfb.c
@@ -34,8 +34,8 @@
 #include <sys/endian.h>
 #include <sys/socket.h>
 #include <sys/select.h>
+#include <sys/stat.h>
 #include <sys/time.h>
-#include <sys/un.h>
 #include <arpa/inet.h>
 #include <stdatomic.h>
 #include <machine/cpufunc.h>
@@ -1258,12 +1258,11 @@ int
 rfb_init(sa_family_t family, const char *hostname, int port, int wait,
     const char *password)
 {
-	int e;
+	int e, errno0;
 	char servname[6];
 	struct rfb_softc *rc;
 	struct addrinfo *ai = NULL;
 	struct addrinfo hints;
-	struct sockaddr_un sun;
 	int on = 1;
 	int cnt;
 #ifndef WITHOUT_CAPSICUM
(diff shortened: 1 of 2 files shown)


bsdimp:
@bsdjhb @markjdb Please take a look. Thanks.

yuichiro-naito:
I refactored the `abspath` function.
- add a shortcut if the argument is an absolute path.
- remove strdup(3) and basename(3) call to reduce overhead.
- simplify error logic.

After the path is resolved, its length may be longer than the original. I checked the path length after resolving.

And @hrs-allbsd told me that getaddrinfo(3) doesn't see the ai_flags in case of AF_UNIX. I changed to set the ai_flags unconditionally. 

The unlink(2) of the UNIX domain socket path serves a similar purpose to setsockopt(SO_REUSEADDR) if it were a TCP socket. The unlink(2) and the setsockopt(2) are exclusive among AF_UNIX and other cases. The AF_UNIX condition should separate unlink(2) and setsockopt(2).

I removed the umask(2) call. It affects process-wide change. The bhyve process runs multiple threads; umask(2) may affect other threads, leading to unexpected behaviour. To change the UNIX domain socket permission, I use the chmod(2).

The addrinfo structure is no longer needed after the bind(2) call. I changed freeing it after the bind(2) and removed the freeaddrinfo(3) call from the error: label. The intention is to improve maintainability by narrowing the effective scope of the addrinfo. Because the error: label is far away from the bind(2) line.

yuichiro-naito:
I refactored the `abspath` function more simpliy.
And pass the ai_protocol valuet o the socket(2). It won't affect anything actually, but it is expected by the getaddrinfo(3).

usr.sbin/bhyve/rfb.c
```diff
@@ -1304,46 +1303,43 @@ rfb_init(sa_family_t family, const char *hostname, int port, int wait,
...
+	errno0 = errno;
+	freeaddrinfo(ai);
 	if (e < 0) {
+		errno = errno0;
 		perror("bind");
 		goto error;
 	}
 
+	if (family == AF_UNIX && chmod(hostname,
+		S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH) < 0) {
```
markjdb, line 1338: It's not obvious to me that this is a desirable change: maybe one uses a unix socket specifically because it prevents unprivileged users from connecting to the socket.

markjdb (commented):
Aside from the comment, this looks ok, though I'm not sure that it's really an improvement, as it adds a bit more code than it deletes.

usr.sbin/bhyve/rfb.c
yuichiro-naito: According to the original [commit message](https://github.com/freebsd/freebsd-src/commit/cbc6f7e941e42639a0314cd121b06493cce8e0e6) and the [review](https://reviews.freebsd.org/D53814), a UNIX domain socket is used as an alternative to a network port on a jail that doesn't have network connectivity. Usually, all users can connect to a network port. If a UNIX domain socket is limited to a privileged user, it won't be an alternative. I just honor the original intention.

yuichiro-naito:
The reason why my patch code gets longer is to support a relative path to a UNIX domain socket. If it's not needed, we can remove the `abspath` function.

yuichiro-naito:
I don't have a better idea for the code and cannot make progress anymore.
This patch doesn't include a functional change. It's not serious.
Thanks for reviewing my code. I will close this pull request.

PR #1976, opened 2026-01-17
make check: Add KYUADEBUG to target specific kyua tests

patmaddox:
# make check: Add make vars to list and debug kyua test cases

kyua supports listing tests, optionally including metadata. It also supports targeting a specific test in debug mode. It shows the commands that were run, stdout / stderr, and details of the checks that failed.

Examples:

    make -C bin/ls check -DKYUALIST
    make -C bin/ls check -DKYUALIST_V
    make -C bin/ls check KYUADEBUG=ls_tests:T_flag

---

(looks like I'm unable to edit the PR subject)

### DIFF
--- a/share/mk/suite.test.mk
+++ b/share/mk/suite.test.mk
@@ -46,6 +46,9 @@ KYUAFILE?= auto
 # unqualified TEST_METADATA variable.
 #TEST_METADATA.<test-program>+= key="value"
 
+# Test name to run with kyua debug instead of running all tests.
+#KYUADEBUG
+
 .if ${KYUAFILE:tl} != "no"
 ${PACKAGE}FILES+=	Kyuafile
 ${PACKAGE}FILESDIR_Kyuafile=	${TESTSDIR}
@@ -78,6 +81,18 @@ Kyuafile: Makefile
 
 KYUA?=	kyua
 
+.if defined(KYUALIST)
+KYUA_CMD_VERB=list
+.elif defined(KYUALIST_V)
+KYUA_CMD_VERB=list
+KYUA_CMD_ARGS=-v
+.elif !empty(KYUADEBUG)
+KYUA_CMD_VERB=debug
+KYUA_CMD_ARGS=${KYUADEBUG}
+.endif
+
+KYUA_CMD= ${KYUA} ${KYUA_CMD_VERB:Utest} -k ${DESTDIR}${TESTSDIR}/Kyuafile ${KYUA_CMD_ARGS}
+
 # Definition of the "make check" target and supporting variables.
 #
 # This target, by necessity, can only work for native builds (i.e. a FreeBSD
@@ -98,7 +113,7 @@ realcheck: .PHONY
 		echo "KYUA=\"${LOCALBASE}/bin/kyua\""; \
 		false; \
 	fi
-	@env ${TESTS_ENV:Q} ${KYUA} test -k ${DESTDIR}${TESTSDIR}/Kyuafile
+	@env ${TESTS_ENV:Q} ${KYUA_CMD}
 
 MAKE_CHECK_SANDBOX_DIR=	checkdir
 CLEANDIRS+=	${MAKE_CHECK_SANDBOX_DIR}
--- a/share/man/man7/build.7
+++ b/share/man/man7/build.7
@@ -114,6 +114,12 @@ The default directory used is
 .Pa ${.OBJDIR} ,
 but the check directory can be changed with
 .Pa ${CHECKDIR} .
+List available test cases with
+.Pa -DKYUALIST ,
+or include their metadata with
+.Pa -DKYUALIST_V .
+Run a specific test case in debug mode with
+.Pa KYUADEBUG=test_file:case_name .
 .It Cm checkworld
 Run the
 .Fx


share/mk/suite.test.mk
```diff
@@ -78,6 +81,11 @@ Kyuafile: Makefile
 
 KYUA?=	kyua
 
+KYUA_CMD=	${KYUA} test -k ${DESTDIR}${TESTSDIR}/Kyuafile
```
kevans91: I'd be tempted to move this to an `.else` branch, or construct it differently to highlight the differences better.  Something like:

```
if !empty(KYUADEBUG)
KYUA_CMD_VERB=debug
KYUA_CMD_ARGS=${KYUADEBUG}
.endif

KYUA_CMD= ${KYUA} ${KYUA_CMD_VERB:Utest} -k ${DESTDIR}${TESTSDIR}/Kyuafile ${KYUA_CMD_ARGS}
```

Maybe that's not really an improvement, though.

patmaddox: I like it. I wanted to communicate that `test` was default and I think yours does that more clearly. It also opens up more of kyua API. I added `list` and `list -v`.

Not sure when it makes sense to move from make vars to separate targets... but I get good mileage out of `KYUADEBUG`, and now `KYUALIST_V` looks nice.

Though these sorts of changes make me wonder: how do other people typically debug tests? I guess they must `make install` and then `kyua debug` from there?

ngie-eign:
@patmaddox : I really appreciate and like the concept of being able to run `kyua test` and `kyua debug` with different arguments via `make check` in order to ease development... but the interface (passing variables prefixed with `KYUA`) seems a bit non-ideal because it requires reading the right documentation to understand how to make one's life easier, instead of there being a better way to skin the proverbial cat (I personally like the idea of ZSH completions more, but that's me...).
I'll simmer on this a bit and see if I can offer a counterproposal, incorporating your request :).

ngie-eign:
> @patmaddox : I really appreciate and like the concept of being able to run `kyua test` and `kyua debug` with different arguments via `make check` in order to ease development... but the interface (passing variables prefixed with `KYUA`) seems a bit non-ideal because it requires reading the right documentation to understand how to make one's life easier, instead of there being a better way to skin the proverbial cat (I personally like the idea of ZSH completions more, but that's me...). I'll simmer on this a bit and see if I can offer a counterproposal, incorporating your request :).

Quick question that just came to mind: how do other projects handle parameterizing out things like this with their respective `make check` targets? Maybe there's something we can glean from the community at large in order to make the interface more intuitive.

ihoro:
How ugly would it be to have something like `$(make kyua) test|debug|whatever ...`.
The goal is not to learn new "language" and use Kyua as is. The only technical need here would be to implement addition to make `kyua` understand `-k <path-to-Kyuafile>` option, so that `make kyua` would yield something like `kyua -k <path-to-the-Kyuafile-in-question>`, as currently only sub-commands like `test` or `debug` can accept the `-k` option.

ngie-eign:
> How ugly would it be to have something like `$(make kyua) test|debug|whatever ...`. The goal is not to learn new "language" and use Kyua as is. The only technical need here would be to implement addition to make `kyua` understand `-k <path-to-Kyuafile>` option, so that `make kyua` would yield something like `kyua -k <path-to-the-Kyuafile-in-question>`, as currently only sub-commands like `test` or `debug` can accept the `-k` option.

I'm not sure I follow the suggestion. `$(make kyua)` asks `make` to evaluate the `kyua` target.. this suggestion blends 2 tools together in a way that doesn't seem to make sense to me.

patmaddox:
> @patmaddox : I really appreciate and like the concept of being able to run `kyua test` and `kyua debug` with different arguments via `make check` in order to ease development... but the interface (passing variables prefixed with `KYUA`) seems a bit non-ideal because it requires reading the right documentation to understand how to make one's life easier, instead of there being a better way to skin the proverbial cat (I personally like the idea of ZSH completions more, but that's me...). I'll simmer on this a bit and see if I can offer a counterproposal, incorporating your request :).

I agree it's not a great interface... it's trying to stay within the constraints of the `make` framework we have. Not sure how to avoid reading that documentation - same goes for even knowing that `make check` exists. But I also think a cleaner way would be much nicer. I think it has to be work with a pretty vanilla setup though... zsh completions would be a nice to have, but there should be an ergonomic way to debug tests without having zsh.

> How ugly would it be to have something like `$(make kyua) test|debug|whatever ...`. The goal is not to learn new "language" and use Kyua as is. The only technical need here would be to implement addition to make `kyua` understand `-k <path-to-Kyuafile>` option, so that `make kyua` would yield something like `kyua -k <path-to-the-Kyuafile-in-question>`, as currently only sub-commands like `test` or `debug` can accept the `-k` option.

I really like this suggestion, it's much more in line with what I was hoping to accomplish: which is to be able to call `kyua` directly, but wired up to the local tree.

I wonder if this just needs a `make -V KYUAFILE` and an example in `tests(7)` showing how to use it, e.g. `kyua list -k $(make -V KYUAFILE)`?

---

Stepping back a bit: what is the workflow that people currently use to `list` / `debug` with kyua in freebsd-src? Like say you add a new test to `ls`, and you want to `kyua debug` that test - how would you do it today?

patmaddox:
I'm thinking this may not be necessary... I've been approaching running FreeBSD tests as though you run them directly in the source tree, similar to how you would `mix test` or `cargo test` or whatever. However as I've explored more, my understanding is that tests are primarily meant to be run from `/usr/tests` on an installed system (whether the running system, a jail, a VM, whatever). In that case, I think the real fix is a doc update that shows the development workflow of updating a component and running its tests - which is a build and install cycle, followed by running tests against the installation.

PR #1977, opened 2026-01-17
build.7: Add an example for developing and testing userspace

patmaddox:
Adding new functionality to userspace sometimes requires updating a lib and bin together. This example demonstrates how to build the lib, and use the build result for building the bin and running tests.

This example differs from "Example 3: Build and upgrade a single piece of userspace" because it demonstrates how a developer may quickly iterate on a piece of userspace before installing it to its final destination.

While here, update examples to consistently use /usr/src.

### DIFF
--- a/share/man/man7/build.7
+++ b/share/man/man7/build.7
@@ -1109,7 +1109,22 @@ cd /usr/src/bin/ls
 make clean all
 make install
 .Ed
-.Ss Example 4: Build and upgrade a loadable kernel module
+.Ss Example 4: Develop and test a single piece of userspace with a dependency
+Develop and check a single piece of userspace with a dependency, in this case
+.Xr bectl 1 ,
+which depends on
+.Xr libbe 3 :
+.Bd -literal -offset indent
+cd /usr/src
+make buildworld
+make -C lib/libbe
+LD_LIBRARY_PATH=$(make -C lib/libbe -V .OBJDIR) make -C lib/libbe check
+CFLAGS="-I$(realpath lib/libbe)" make -C sbin/bectl
+PATH=$(make -C sbin/bectl -V .OBJDIR):$PATH \\
+	LD_LIBRARY_PATH=$(make -C lib/libbe -V .OBJDIR) \\
+	make -C sbin/bectl check
+.Ed
+.Ss Example 5: Build and upgrade a loadable kernel module
 Rebuild and reinstall a single loadable kernel module, in this case
 .Xr sound 4 :
 .Bd -literal -offset indent
@@ -1156,7 +1171,7 @@ config -d /home/user/my-kernel-builddir -s /home/user/src/sys/amd64/conf MY-KERN
 cd my-module
 make all KERNBUILDDIR=/home/user/my-kernel-builddir SYSDIR
 .Ed
-.Ss Example 5: Quickly rebuild a kernel in place
+.Ss Example 6: Quickly rebuild a kernel in place
 Quickly rebuild and reinstall the kernel, only recompiling the files
 changed since last build; note that this will only work if the full
 kernel build has been completed in the past, not on a fresh source tree:
@@ -1164,13 +1179,13 @@ kernel build has been completed in the past, not on a fresh source tree:
 cd /usr/src
 make kernel KERNFAST=1
 .Ed
-.Ss Example 6: Cross-compiling for different architectures
+.Ss Example 7: Cross-compiling for different architectures
 To rebuild parts of
 .Fx
 for another CPU architecture,
 first prepare your source tree by building the cross-toolchain:
 .Bd -literal -offset indent
-cd src
+cd /usr/src
 make toolchain TARGET_ARCH=aarch64
 .Ed
 .Pp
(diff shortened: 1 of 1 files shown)


share/man/man7/build.7
```diff
@@ -1119,6 +1119,16 @@ cd src
 make buildenv TARGET_ARCH=aarch64
 make kernel KERNFAST=1 DESTDIR=/armclient
 .Ed
+.Ss Example 7: Build and test a single piece of userspace
```
concussious: Please put this after example 3 and increment subsequent example numbers

share/man/man7/build.7
```diff
@@ -1119,6 +1119,16 @@ cd src
...
+Rebuild and check a single piece of userspace, in this case
+.Xr libbe 3 :
+.Bd -literal -offset indent
+cd /usr/src
```
concussious:
```suggestion
cd src
```
For consistency with the other examples

concussious (approved):
Hi Pat! Thanks so much for this, this is wonderful! I would be happy to take this with the following small tweaks:

share/man/man7/build.7
concussious: Oops, I'm wrong; looking closer I now think the ones that just say src should be aligned to how you wrote it.

bsdimp: yea, this is already consistent with the other examples.  the 'src' in them is a shorthand for the top within the example, but they all start with /usr/src with the understanding that you can have the source anywhere.

share/man/man7/build.7
```diff
@@ -1119,6 +1119,16 @@ cd src
...
+make buildenv
+cd lib/libbe
+DESTDIR=${SYSROOT} make clean all install -DWITHOUT_TESTS
```
bsdimp: why WITHOUT_TESTS if you are doing make check?

share/man/man7/build.7
patmaddox: Done!

share/man/man7/build.7
patmaddox: Based on this feedback, I updated the last example to consistently use `/usr/src` - before the sequence was: src, /usr/src, src/bin/ls, src

share/man/man7/build.7
patmaddox: I did my best to explain in the commit comment, please let me know if I can word things more clearly. As far as I can tell, it only comes up when adding new symbols. Here's how it goes:

1. Add a symbol to `foo.h` and call it from the test program
2. `make all` now fails, because a) `${SYSROOT}/usr/include/foo.h`  does not have that symbol b) the test program uses it and c) the test program is included in `make all`
3. You can't call `make install` to install it

If you check out the code from #1975 you should be able to reproduce this (example below).

So the trick is we need to run `make install` to update the header file _before_ we can build the test program. That's what `make all install -DWITHOUT_TESTS` provides, and then the test program gets built when you run `make check`. 

Another way to do it is setting up CFLAGS and LDFLAGS (e.g. `CFLAGS="-I$(make -C lib/libbe -V .OBJDIR)" LDFLAGS="-L$(make -C lib/libbe -V .OBJDIR)" make -s -C lib/libbe clean all`) but that's pretty painful imo.

Perhaps instead of this, building the test program needs to be changed to use those when run in the source tree? Because my understanding is that tests can be run in two ways:

1. From `/usr/tests` against an installed system
2. From within the source tree, (presumably) against the built object files

```
root@bsdev:~/lab.jj/oss/freebsd-src/main.jj # make buildenv
Entering world for amd64:amd64
root@bsdev:~/lab.jj/oss/freebsd-src/main.jj # make -s -C lib/libbe clean all
===> tests (clean)
===> tests (all)
ld: error: undefined symbol: be_create_empty
>>> referenced by target_prog.c:51 (/root/lab.jj/oss/freebsd-src/main.jj/lib/libbe/tests/target_prog.c:51)
>>>               target_prog.pieo:(main)
cc: error: linker command failed with exit code 1 (use -v to see invocation)
*** Error code 1

Stop.
make[4]: stopped making "all" in /root/lab.jj/oss/freebsd-src/main.jj/lib/libbe/tests
*** Error code 1

Stop.
make[3]: stopped making "all" in /root/lab.jj/oss/freebsd-src/main.jj/lib/libbe/tests
*** Error code 1

Stop.
make[2]: stopped making "clean all" in /root/lab.jj/oss/freebsd-src/main.jj/lib/libbe
```

patmaddox: hrm... I think I jumped the gun on this. After some more testing, it seems that `make check` _does_ require it to be installed to the system and does not actually look in `SYSTEM_ROOT` for the headers and obj. I just happened to have previously installed it and so was getting false positives.

> Perhaps instead of this, building the test program needs to be changed to use those when run in the source tree?

I think this is really it... when I'm running `make check` within the source tree, I expect it to build against the obj files from the source tree, not from the system. I'll poke around a bit more.

patmaddox:
I rebuilt my dev VM so I could walk through the steps from scratch that led me to the place where I needed this... and so far have not been able to reproduce it. So now I'm thinking I just did something dumb, and found a workaround in a busted environment.

Going to close this for now - and if I come across it again, I'll re-open and revise with new info.

patmaddox:
Okay reopened and rewritten. The `-DWITHOUT_TESTS` was a red herring caused by a line missing from `bin/bectl/tests/Makefile`.

My goal was to develop `libbe(3)` and `bectl(1)` together without installing files until they were ready. This example documents how to do that.

patmaddox:
I think this example is fighting the build / test system... the intent of this example is to not have to install the components. But as you can see, it's pretty convoluted. From what I can tell, the idiomatic approach is to install the built components (whether onto the running system, or into a jail), and run the tests from `/usr/tests`.

That's ultimately what I would like to document: how to iterate on components and test them without installing them onto the running system. So the example would explicitly build, install into a jail, and run tests from the jail.

I could use some guidance on the best location for that kind of example. One possibility is `build.7` and just be more clear about it being for development, as opposed to updating the running system (which is what the majority of the examples indicate). Another is `development.7` although that seems to primarily be an index to other documentation - I'm not sure if it's intended to be more self-sufficient in the way I'm describing. Still another would be the developer's handbook.

I am leaning towards `development.7` because I think that's the natural man page to answer questions about "how do I develop changes for FreeBSD?" I don't know if it's intended to be a thin index, or if it just happens to be that way so far and more information would be welcomed.

PR #1978, opened 2026-01-17
Makefile.inc1: Include sbin in buildenv path

patmaddox:
`make buildenv` sets up PATH under SYSROOT to support targeted installation and testing with `make check`. It previously ommitted sbin, leading to confusion and false reports when running `make check` on any sbin programs.


bsdimp (approved):
Normally this isn't needed. But since make check I guess it has for testing things in sbin....

patmaddox:
> Normally this isn't needed. But since make check I guess it has for testing things in sbin....

I wonder if it may not be. This came about because I was doing `make buildenv` and then `DESTDIR=${SYSROOT} make install`. But I think I may have misunderstood the intention of `SYSROOT` in the first place. If I'm going to install stuff there, then I should start with `DESTDIR=${SYSROOT} make hierarchy`.

So I think we can close this.

PR #1993, opened 2026-01-27
release: Support VMCONFIG for vm-image

patmaddox:
# release: Pass optional VM_IMAGE_CONFIG to vm-image

`make vm-image` calls mk-vmimage.sh, which supports `-c CONFFILE`. This file gets sourced before building the image.

One example of how to use it is to define vm_extra_filter_base_packages() to filter the list of packages installed into the VM image:

    # vm-nodbg32.conf
    vm_extra_filter_base_packages() {
        grep -v -E '(-dbg|lib32)'
    }

    $ make VM_IMAGE_CONFIG=path/to/vm-nodbg32.conf \
        VMFORMATS=raw \
        -DWITH_VMIMAGES \
        vm-image

### DIFF
--- a/release/Makefile.vm
+++ b/release/Makefile.vm
@@ -211,6 +211,7 @@ vm-image:	${QEMUTGT} ${PKGBASE_REPO_DIR}
 		PKGBASE_REPO_DIR=${.OBJDIR}/pkgbase-repo-dir \
 		${.CURDIR}/scripts/mk-vmimage.sh \
 		-C ${.CURDIR}/tools/vmimage.subr \
+		${VM_IMAGE_CONFIG:D-c ${VM_IMAGE_CONFIG}} \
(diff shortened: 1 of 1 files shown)


kevans91:
Ideally we get re@ buy-in for things that touch ^/release -- this seems fairly innocuous and obviously correct, though.

cperciva:
I like the concept, but I'd prefer VM_IMAGE_CONFIG rather than VMCONFIG in order to make it clear that this is just for the vm-image target.  Also, the commit message could use some rewording along the lines of "This makes it possible to specify a configuration file to be used for the vm-image target; one example of how this might be used is...".

patmaddox:
@cperciva thanks for the feedback. I've updated the var name, and included more detail of how the config file gets passed around, and an example of how I use it.

cperciva:
LGTM.  @kevans91 can you land this?

PR #1994, opened 2026-01-28
Import spdxtool from pkgconf into the base system

khorben:
[Edited to match the current status]

This series of commits relied on a new vendor import, pkgconf, which has been done at version 2.9.93 by now (version 3.0.6 has been released since). pkgconf into the base system also provides two programs able to generate SBOM artefacts:

* bomtool(1) for SPDX 2.2, and
* spdxtool(1) for SPDX 3.0.1, created with funding from the FreeBSD Foundation.

This import is now the last stretch automating the generation of SBOM artefacts while building the system. Regulation like the European Union's CRA will soon require manufacturers to ship SBOM artefacts in their products. This framework provides this functionality within the base system, not only to allow commercial users of FreeBSD to easily comply with the regulation, but also for any user to benefit from the tooling around it (e.g., for security checks).

Note also that:

* libpkgconf in base has been made private, in order to avoid conflicts with ports.
* `pkgconf(1)` in base automatically selects the correct `LOCALBASE` for the system, as per the `user.localbase` sysctl(8), or per the `LOCALBASE` environment variable.
* (Installations of `pkgconf(1)` from the ports still behave as expected and always stick to their respective build-time location)
* The relevance of SBOM artefacts and role of `bomtool(1)` and `spdxtool(1)` has now been introduced broadly, e.g., at FreeBSD DevSummits, BSD conferences, in quarterly status reports, Phabricator reviews, the FreeBSD Journal...
* The correct version number in the artefacts is obtained automatically for software originating from FreeBSD, but not (yet) for third-party vendor imports.
* The granularity of SBOM artefacts is currently based on the packages listed in `packages`, and on their respective libraries.
* As a consequence, this currently adds about one hundred meta-data files to the release.
* The SBOM artefacts generated are shipped within their corresponding packages when using PKGBASE.


bsdimp:
So with vendor branches, this pull request can't be landed.  But that's a silly process thing: this review looks like it is open for discussion and refinement and for that, this is as good a place as any.

concussious:
So, why should we import spdxtool to the base system?

khorben:
> So, why should we import spdxtool to the base system?

Very good question, I should have started with that!

There is increasing pressure from the broader community to be accountable for their software supply chain, and this can be implemented through the generation of SBOM files (Software Bill of Material). This is not only the case for industrial users but also for anyone shipping  software commercially, as materialised in Europe by its new CRA regulation (Cyber Resilience Act). It may seem like a burden, but it is really an opportunity not only to provide mechanisms for vulnerability tracking in the base system, but also to raise the profile for FreeBSD as a readily compliant, attractive platform for manufacturers considering this system.

In practice, there are mainly two specifications competing as SBOM file formats: CycloneDX and SPDX, and then different versions for them (see also https://sbom.ecosyste.ms). While tooling for "modern" languages already exists, until recently no particular solution emerged for systems written in C. This is when Ariadne Conill (@kaniini), a major author of pkgconf (https://github.com/pkgconf/pkgconf) presented bomtool (https://ariadne.space/2025/02/08/c-sboms-and-how-pkgconf.html) and published it as part of pkgconf. While bomtool only supported version 2.2 of the SPDX file format, Tuukka Pasanen (@illuusio) wrote spdxtool, which supports version 3 (aka JSON-LD) and looks like a better, more future-proof candidate as a file format for SBOM.

Tuukka is funded by Germany's STA (Sovereign Tech Agency, https://www.sovereign.tech) through the FreeBSD Foundation (https://freebsdfoundation.org) to help FreeBSD modernise its infrastructure, and this work is part of the Work Package E of this initiative (https://github.com/FreeBSDFoundation/all-projects/tree/main/Infrastructure%20Modernization%20(STA%20commissioned)). I am involved in this effort as well but from a different avenue, Alpha-Omega's Beach Cleaning project (https://alpha-omega.dev), where I help track and "fix/fork/forego" third-party dependencies of FreeBSD (https://github.com/FreeBSDFoundation/alpha-omega-beach-cleaning).

I anticipate our common solution to eventually look as follows:
* Import pkgconf, or at least bomtool and ideally spdxtool into the base system (at least for use during the build phase)
* Generate a file in the pkg-config format for each of the components built
* Leverage bomtool and/or spdxtool to convert these `.pc` files into SPDX files
* (Optionally?) Ship the `.pc` and SPDX files into e.g., `/usr/lib/sbom`, `/usr/lib/sbom/SPDX-2.2`, `/usr/lib/sbom/SPDX-3` (mimicking pkg-config) or `/usr/share/sbom/...` (TBD)

A number of details remain TBD, like where these SBOM files should be located but also their granularity. I anticipate that in practice, there will be one SBOM file per package in base (not including the `-dev` variants) and that these files will be part of their corresponding package (or set). Other suggestions are welcome evidently.

We are now beginning to work on this latter part, but we want to be transparent and collect feedback on the initiative as we go, increasing the chances for a smooth adoption.

I hope this clarifies!

bms:
> > So, why should we import spdxtool to the base system?
> 
> Very good question, I should have started with that!
...

That's fine, but why didn't you submit it as a port first? SPDX seems to be preferred.

That is far easier for most committers to deal with than raising a PR on GitHub, which is not the preferred workflow; it's offered as a courtesy to encourage 3rd party contributions, but it is not the fast path.

Moreover, FreeBSD is gradually moving towards a packaged base system. This is not the optimal workflow for actively hacking on -CURRENT at the moment, but that is another story.

khorben:
> > > So, why should we import spdxtool to the base system?
> > 
> > 
> > Very good question, I should have started with that!
> > ...
> 
> That's fine, but why didn't you submit it as a port first? SPDX seems to be preferred.

spdxtool is not part of any official release of pkgconf yet; but even if it was, this wouldn't be our preferred path.

Our objective is to allow the creation of SBOM information while building FreeBSD's src tree. We cannot rely on the presence of spdxtool, of the right version, in the build environment, except for having it as part of the src tree directly.

> That is far easier for most committers to deal with than raising a PR on GitHub, which is not the preferred workflow; it's offered as a courtesy to encourage 3rd party contributions, but it is not the fast path.

I am familiar with reviews.freebsd.org, but Phabricator is not convenient for workflows involving a vendor branch. Here, I could push the vendor branch in my own fork on GitHub, base this proposal on it, and easily reference both in a pull-request.

> Moreover, FreeBSD is gradually moving towards a packaged base system. This is not the optimal workflow for actively hacking on -CURRENT at the moment, but that is another story.

👍

HTH!

concussious:
Thanks for the detailed explanation!

>> So, why should we import spdxtool to the base system?

> Our objective is to allow the creation of SBOM information while building FreeBSD's src tree. We cannot rely on the presence of spdxtool, of the right version, in the build environment, except for having it as part of the src tree directly.

This is the part I'm trying to imply would be helpful to be in the commit message.

bms:
> I am familiar with reviews.freebsd.org, but Phabricator is not convenient for workflows involving a vendor branch. Here, I could push the vendor branch in my own fork on GitHub, base this proposal on it, and easily reference both in a pull-request.

I actually partially retract some of my original comment, as I see @bsdjhb raised a PR for a change which is potentially too big to stage in Phrabricator; I may have to do the same in future.

bms:
> > > So, why should we import spdxtool to the base system?
> 
> > Our objective is to allow the creation of SBOM information while building FreeBSD's src tree. We cannot rely on the presence of spdxtool, of the right version, in the build environment, except for having it as part of the src tree directly.
> 
> This is the part I'm trying to imply would be helpful to be in the commit message.

Please do. I am tracking sigstore.dev and this (SBOM etc) carefully, it will be needed in future to support software supply chain assurance in many jurisdictions.

khorben:
One problem remains with the 32-bit version of the library packages; they also include the corresponding SBOM files generated, in the same location, so they will most likely conflict with the native 64-bit versions.

khorben:
> > > > So, why should we import spdxtool to the base system?
> > [...]
> Please do. I am tracking sigstore.dev and this (SBOM etc) carefully, it will be needed in future to support software supply chain assurance in many jurisdictions.

I am glad to report that spdxtool is now in the base system (for the future 16.0).
I will now focus on the integration framework for the generation of SBOM files as part of the build, while completing the integration of pkgconf's own test suite into that of FreeBSD proper.

khorben:
This pull-request is now marked as ready for review, since its latest state matches exactly the patch in https://reviews.freebsd.org/D56474.
I am aware of the inconsistencies noted there within the corpus of meta-data provided; I am going to try to address them, even though this part was not done by myself but is the outcome of (tedious!) work from @illuusio (kudos to him!).
Do not hesitate to report anything we may still have missed, here or in Phabricator.

kaniini:
Hi, pkgconf maintainer here! :smile:

This is incredible work.  I spent time looking through both this and the review on Phabricator.

For what it's worth, instead of maintaining hundreds to thousands of pc(5) files by hand since the hand-generated ones are not usable for describing how to use system components as SDKs, it might be better to treat them as an intermediate artifact of SBOM construction only, e.g. generating them automatically from another source like the pkg(8) database.

I think that pkgbase probably has enough information about the component relationships to generate intermediate pc(5) files to use as a data source to construct the SBOM, but probably does not have relevant copyright information.  Personally, I would suggest extracting that from the source files for each component automatically, which is pretty easy to do from the headers.

In general, it is probably better for the pc(5) files to live with each component and, for libraries, describe how they can be used with compilers alongside the SBOM data.  For tools/programs, it may also be worth annotating the pc(5) with a `Kind: tool` annotation, which will be supported by pkgconf 3.1 to distinguish between SDKs and tools as software components.

khorben:
> Hi, pkgconf maintainer here! 😄

Hi, awesome hearing from you!

> This is incredible work. I spent time looking through both this and the review on Phabricator.

Glad you like it, thank you!

> For what it's worth, instead of maintaining hundreds to thousands of pc(5) files by hand since the hand-generated ones are not usable for describing how to use system components as SDKs, it might be better to treat them as an intermediate artifact of SBOM construction only, e.g. generating them automatically from another source like the pkg(8) database.
> 
> I think that pkgbase probably has enough information about the component relationships to generate intermediate pc(5) files to use as a data source to construct the SBOM, but probably does not have relevant copyright information. Personally, I would suggest extracting that from the source files for each component automatically, which is pretty easy to do from the headers.

In this new version of the SBOM generation framework, I have integrated your feedback here and improved it towards your suggestion here. There is now an intermediate step, generating the `.pc` file, which is then used to generate the 2.2 and 3.0.1 SPDX files.

The trigger for the generation of SBOM files is now the presence of a `Makefile.sbom` file in the respective `lib/???` or `usr.{bin,sbin}/???` directory, containing or correcting the information used for the SBOM output. I believe this will make it easier for maintainers to remember to update information there as relevant (e.g., version number for third-party software), but also that it will help find the right balance for the amount of SBOM artefacts to actually generate and ship.

> In general, it is probably better for the pc(5) files to live with each component and, for libraries, describe how they can be used with compilers alongside the SBOM data. For tools/programs, it may also be worth annotating the pc(5) with a `Kind: tool` annotation, which will be supported by pkgconf 3.1 to distinguish between SDKs and tools as software components.

I have taken this into account, and added this annotation to the output for the intermediate `.pc` files.

Looking forward to your reviews!

PR #2005, opened 2026-02-08
stdio: Support 32-bit file descriptors in FILE objects

svenski123:
FreeBSD's libc stdio implementation to this day still stores file descriptors in a short integer field in FILE objects thus causing failures when more than 32,767 files are open.

This pull request allows FreeBSD libc stdio to support the full range of the larger 32-bit file descriptors that the FreeBSD kernel has supported for many years, while still maintaining ABI compatibility with the current lib stdio implementation.

This work follows discussion in both bugzilla bug report 291610 as well as prior work in phabricator D54354 and D54355; please see the URLs below for further details.

* [Bug 291610](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=291610)
* [Differential D54354](https://reviews.freebsd.org/D54354)
* [Differential D54355](https://reviews.freebsd.org/D54355)

### DIFF
--- a/lib/libc/stdio/fdopen.c
+++ b/lib/libc/stdio/fdopen.c
@@ -106,7 +106,7 @@ fdopen(int fd, const char *mode)
 		fp->_flags2 |= __S2OAP;
 	else if (oflags & O_APPEND)
 		fp->_flags |= __SAPP;
-	fp->_file = fd;
+	__sfileno_set(fp, fd);
 	fp->_cookie = fp;
 	fp->_read = __sread;
 	fp->_write = __swrite;
--- a/lib/libc/stdio/fdopen.c
+++ b/lib/libc/stdio/fdopen.c
@@ -48,6 +48,11 @@ fdopen(int fd, const char *mode)
 	FILE *fp;
 	int flags, oflags, fdflags, rc, tmp;
 
+	if (fd < 0) {
+		errno = EBADF;
+		return (NULL);
+	}
+
 	/*
 	 * File descriptors are a full int, but _file is only a short.
 	 * If we get a valid file descriptor that is greater than
--- a/lib/libc/stdio/fdopen.c
+++ b/lib/libc/stdio/fdopen.c
@@ -39,11 +39,12 @@
 #include <stdio.h>
 #include <errno.h>
 #include <limits.h>
+#include <stdbool.h>
 #include "un-namespace.h"
 #include "local.h"
 
-FILE *
-fdopen(int fd, const char *mode)
+static FILE *
+__sfdopen(int fd, const char * __restrict mode, bool short_only)
 {
 	FILE *fp;
 	int flags, oflags, fdflags, rc, tmp;
@@ -60,7 +61,7 @@ fdopen(int fd, const char *mode)
 	 * invalid file descriptor.  Handle this case by failing the
 	 * open.
 	 */
-	if (fd > SHRT_MAX) {
+	if (__sforce_short_fildes(short_only) && fd > SHRT_MAX) {
 		errno = EMFILE;
 		return (NULL);
 	}
@@ -113,4 +114,17 @@ fdopen(int fd, const char *mode)
 	fp->_seek = __sseek;
 	fp->_close = __sclose;
 	return (fp);
+}
+
+FILE *
+freebsd15_fdopen(int fd, const char *mode)
+{
+	return (__sfdopen(fd, mode, true));
+}
+__sym_compat(fdopen, freebsd15_fdopen, FBSD_1.0);
+
+FILE *
+fdopen(int fd, const char *mode)
+{
+	return (__sfdopen(fd, mode, false));
 }
--- a/lib/libc/stdio/findfp.c
+++ b/lib/libc/stdio/findfp.c
@@ -91,7 +91,6 @@ static struct glue *
 moreglue(int n)
 {
 	struct glue *g;
-	static FILE empty = { ._fl_mutex = PTHREAD_MUTEX_INITIALIZER };
 	FILE *p;
 	size_t align;
 
@@ -104,7 +103,7 @@ moreglue(int n)
 	g->niobs = n;
 	g->iobs = p;
 	while (--n >= 0)
-		*p++ = empty;
+		*p++ = (FILE)FAKE_FILE;
 	return (g);
 }
 
--- a/lib/libc/stdio/findfp.c
+++ b/lib/libc/stdio/findfp.c
@@ -182,17 +182,3 @@ f_prealloc(void)
 		STDIO_THREAD_UNLOCK();
 	}
 }
-
-/*
- * exit() calls _cleanup() through *__cleanup, set via a weak reference
- * at program initialisation.  This chicanery is done so that programs
- * that do not use stdio need not link it all in.
- *
- * The name `_cleanup' is, alas, fairly well known outside stdio.
- */
-void
-_cleanup(void)
-{
-	/* (void) _fwalk(fclose); */
-	(void) _fwalk(__sflush);		/* `cheating' */
-}
--- a/lib/libc/stdio/findfp.c
+++ b/lib/libc/stdio/findfp.c
@@ -46,8 +46,6 @@
 #include "local.h"
 #include "glue.h"
 
-int	__sdidinit;
-
 #define	NDYNAMIC 10		/* add ten more whenever necessary */
 
 #define	std(flags, file) {		\
@@ -117,8 +115,6 @@ __sfp(void)
 	int	n;
 	struct glue *g;
 
-	if (!__sdidinit)
-		__sinit();
 	/*
 	 * The list must be locked because a FILE may be updated.
 	 */
@@ -188,8 +184,8 @@ f_prealloc(void)
 }
 
 /*
- * exit() calls _cleanup() through *__cleanup, set whenever we
- * open or buffer a file.  This chicanery is done so that programs
+ * exit() calls _cleanup() through *__cleanup, set via a weak reference
+ * at program initialisation.  This chicanery is done so that programs
  * that do not use stdio need not link it all in.
  *
  * The name `_cleanup' is, alas, fairly well known outside stdio.
@@ -199,16 +195,4 @@ _cleanup(void)
 {
 	/* (void) _fwalk(fclose); */
 	(void) _fwalk(__sflush);		/* `cheating' */
-}
-
-/*
- * __sinit() is called whenever stdio's internal variables must be set up.
- */
-void
-__sinit(void)
-{
-
-	/* Make sure we clean up on exit. */
-	__cleanup = _cleanup;		/* conservative */
-	__sdidinit = 1;
 }
--- a/lib/libc/stdio/findfp.c
+++ b/lib/libc/stdio/findfp.c
@@ -65,7 +65,7 @@ static struct glue uglue = { NULL, FOPEN_MAX - 3, usual };
 static FILE __sF[3] = {
 	std(__SRD, STDIN_FILENO),
 	std(__SWR, STDOUT_FILENO),
-	std(__SWR|__SNBF, STDERR_FILENO)
+	std(__SWR | __SNBF, STDERR_FILENO)
 };
 
 FILE *__stdinp = &__sF[0];
@@ -75,7 +75,7 @@ FILE *__stderrp = &__sF[2];
 struct glue __sglue = { &uglue, 3, __sF };
 static struct glue *lastglue = &uglue;
 
-static struct glue *	moreglue(int);
+static struct glue *moreglue(int);
 
 spinlock_t __stdio_thread_lock = _SPINLOCK_INITIALIZER;
 
@@ -157,9 +157,8 @@ found:
  * XXX.  Force immediate allocation of internal memory.  Not used by stdio,
  * but documented historically for certain applications.  Bad applications.
  */
-__warn_references(f_prealloc, 
+__warn_references(f_prealloc,
 	"warning: this program uses f_prealloc(), which is not recommended.");
-void f_prealloc(void);
 
 void
 f_prealloc(void)
@@ -174,7 +173,7 @@ f_prealloc(void)
 	 * removed.
 	 */
 	for (g = &__sglue; (n -= g->niobs) > 0 && g->next; g = g->next)
-		/* void */;
+		;	/* nothing */
 	if ((n > 0) && ((g = moreglue(n)) != NULL)) {
 		STDIO_THREAD_LOCK();
 		SET_GLUE_PTR(lastglue->next, g);
--- a/lib/libc/stdio/fopen.3
+++ b/lib/libc/stdio/fopen.3
@@ -260,6 +260,12 @@ is returned and the global variable
 is set to indicate the error.
 .Sh ERRORS
 .Bl -tag -width Er
+.It Bq Er EBADF
+The
+.Fa fildes
+argument to
+.Fn fdopen
+was not a valid file descriptor.
 .It Bq Er EINVAL
 The
 .Fa mode
@@ -271,6 +277,10 @@ to
 or
 .Fn fmemopen
 was invalid.
+.It Bq Er EMFILE
+The file descriptor for the opened
+.Tn FILE
+is greater than 32,767 and legacy compatibility mode is enabled.
 .El
 .Pp
 The
@@ -320,6 +330,18 @@ may also fail and set
 if the
 .Fa size
 argument is 0.
+.Pp
+The
+.Fn fopen ,
+.Fn fdopen
+and
+.Fn freopen
+functions will enforce the legacy limit of at most 32,767 open files with
+binaries linked against previous versions of the standard C library.
+As a debugging aid, this legacy compatibility mode can also be enabled for
+the time being by setting the environment variable
+.Va __STDIO_FORCE_SHORT_FILDES
+prior to executing the program.
 .Sh SEE ALSO
 .Xr open 2 ,
 .Xr fclose 3 ,
--- a/lib/libc/stdio/freopen.c
+++ b/lib/libc/stdio/freopen.c
@@ -66,9 +66,6 @@ freopen(const char * __restrict file, const char * __restrict mode,
 
 	FLOCKFILE_CANCELSAFE(fp);
 
-	if (!__sdidinit)
-		__sinit();
-
 	/*
 	 * If the filename is a NULL pointer, the caller is asking us to
 	 * re-open the same file with a different mode. We allow this only
--- a/lib/libc/stdio/freopen.c
+++ b/lib/libc/stdio/freopen.c
@@ -78,7 +78,8 @@ freopen(const char * __restrict file, const char * __restrict mode,
 			fp = NULL;
 			goto end;
 		}
-		if ((dflags = _fcntl(fp->_file, F_GETFL)) < 0) {
+		f = __sfileno(fp);
+		if ((dflags = _fcntl(f, F_GETFL)) < 0) {
 			sverrno = errno;
 			fclose(fp);
 			errno = sverrno;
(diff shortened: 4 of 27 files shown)


bsdimp:
Adding @bsdjhb since John has been doing something in this area, I think, or at least reviewing it.

lib/libc/stdio/fdopen.c
```diff
@@ -114,3 +115,16 @@ fdopen(int fd, const char *mode)
...
+
+FILE *
+__sfdopen_short_only(int fd, const char * __restrict mode)
+{
+	return __sfdopen(fd, mode, true);
+}
+__sym_compat(fdopen, __sfdopen_short_only, FBSD_1.0);
+
+FILE *
+fdopen(int fd, const char * __restrict mode)
```
brooksdavis: declaring `fdopen` with a restrict qualifier doesn't accomplish anything (there's nothing for `mode` to alias with) and POSIX doesn't do it (it does for `fopen`.)

lib/libc/stdio/freopen.c
```diff
@@ -50,12 +52,15 @@
...
  * all possible, no matter what.
  */
-FILE *
-freopen(const char * __restrict file, const char * __restrict mode,
-    FILE * __restrict fp)
+static FILE *
+__sfreopen(const char * __restrict file,
+	   const char * __restrict mode,
+	   FILE * __restrict fp,
+	   bool short_only)
```
brooksdavis:
```suggestion
__sfreopen(const char * __restrict file, const char * __restrict mode,
    FILE * __restrict fp, bool short_only)
```
Similar style issues below.

lib/libc/stdio/local.h
```diff
@@ -95,6 +96,79 @@ __fgetwc(FILE *fp, locale_t locale)
...
+ * for file descriptors in this range.
+ */
+#define __S2FDX_SHFT				(1)
+#define __S2FDX_EXTRACT(_flags2)		((int)((unsigned)((_flags2) & __S2FDX) >> __S2FDX_SHFT))
+#define __S2FDX_INSERT(_flags2, val)		((_flags2) & (~__S2FDX) | (__S2FDX & ((val) << __S2FDX_SHFT)))
+#define __SFD_TO_LOW(fd)			((short)(fd))
+#define __SFD_TO_HSX(fd)			((unsigned)((fd) ^ (int)(short)(fd)) >> 16)
+#define __SLOW_HSX_TO_FD(low, hsx)		((int)(short)(low) ^ ((int)(hsx) << 16))
+
+_Static_assert(USHRT_MAX << __S2FDX_SHFT	== __S2FDX,	"__S2FDX coding invariant violated");
```
brooksdavis: These are all too long per style(9) and the (annoyingly required) message is uninformative. Maybe shorten it so they fit?

brooksdavis (commented):
Nice work. 

On the last commit, I worry a little that we're losing atomicity (in the sense of not having write tearing) in updates to the file handle. I think software relying on atomic updates would be buggy, but it might exist and be hard to find.

I wonder if we might want a way to control `short_only` on a per-process basis to make it easier to debug problems. An environmental variable queried at first FILE creation would be an option. There's also the option of a more complex approach using AT_BSDFLAGS and elfctl/proccontrol which would allow certain binaries or processes to opt out of the new model.

lib/libc/stdio/freopen.c
```diff
@@ -265,18 +263,16 @@ __sfreopen(const char * __restrict file,
 }
 
 FILE *
-__sfreopen_short_only(const char * __restrict file,
-		      const char * __restrict mode,
-		      FILE * __restrict fp)
+__sfreopen_short_only(const char * __restrict file, const char * __restrict mode,
```
brooksdavis: This line is one column too long

lib/libc/stdio/freopen.c
```diff
@@ -136,6 +136,15 @@ __sfreopen(const char * __restrict file, const char * __restrict mode,
 	 * a descriptor, defer closing it; freopen("/dev/stdin", "r", stdin)
 	 * should work.  This is unnecessary if it was not a Unix file.
 	 */
+
+	/*
+	 * As __sfp() called by fopen, fdopen or funpen in another thread may
+	 * reuse closed FILE objects at any time, lock the glue list and set
+	 * _flags to ensure this doesn't happen. Note there is still a theo-
+	 * retical race condition if __sfp() has just acquired this FILE, as
```
brooksdavis: I don't think I've ever seen mid-word hyphenation for line wrapping in FreeBSD comments.

lib/libc/stdio/findfp.c
```diff
@@ -205,10 +208,29 @@ _cleanup(void)
...
+__sinit_once(void)
 {
+	if (__sinit_done)
+		return;
+
 	/* Make sure we clean up on exit. */
 	__cleanup = _cleanup;		/* conservative */
-	__sdidinit = 1;
+
+	if (getenv("__STDIO_FORCE_SHORT_ONLY__") != NULL)
```
brooksdavis: I'd probably include `FD` or `FD_RANGE` in there somewhere. This should be documented in fopen.3.

lib/libc/stdio/findfp.c
```diff
@@ -205,10 +208,29 @@ _cleanup(void)
...
+void
+__sinit(void)
+{
+	if (__sinit_done)
+		return;
+
+	if (__isthreaded)
+		_pthread_once(&__sinit_once_control, &__sinit_once);
+	else
+		__sinit_once();
```
brooksdavis: This isn't my area of expertise so I'm not 100% confident but I think this whole function can be:
```
	_once(&__sinit_once_control, &__sinit_once);
```
and `__sinit_done` can be GC'd entirely.

svenski123: _once() is nifty and replaces 232-235. Still need __sinit_done IMHO, so inlined __smaybe_init() remains light with no calls in usual case.

_once() is call & always checks __isthreaded as well as the pthread_once_t item, which looks like an int + mutex.

svenski123:
The six commits today apply/fix style edits to all files involved in this pull request, address feedback points regarding qualifiers etc. (thank you!), add an environment variable to force legacy "short mode" and update the manual page.

I've looked at threading / concurrency in stdio and while I don't think splitting the file descriptor in two pieces in two different cache lines poses a problem as fileno() locks the FILE's mutex, there are other possible race conditions that are more concerning; in particular the way __sfp() reuses closed FILE objects without locking them and the interaction with fopen/fdopen/fclose and freopen in particular.

In a multi-threaded context it is not safe to access a FILE object once it has closed as once _flags is cleared any other thread running fopen etc. may get it. Also __sfp() and fopen etc. do not lock the FILE when initialising it so fclose called on it concurrently will race with the open. While this may be overcome by locking the FILE in __sfp() and/or fopen(), there is a real risk of introducing the potential for deadlocking between __stdio_thread_lock and one or more FILE mutexes and so this must considered with extreme care. As such, that is somewhat outside the scope of this pull request and so I've limited myself to concurrency fixes that are clear and unambiguous.

I am also wondering whether _file should be renamed/obfuscated as part of this change to dissuade applications from attempting to use it. The last two ports that accessed FILE->_file directly have now been patched (see bug 292939) so renaming _file should now be possible without breaking anything. However doing so might preclude this feature ever ultimately being backported to a later 14 or 15 release. Perhaps for backporting to 14.x/15.x it might make sense for this feature to be opt-in i.e. require a distinct environment variable be set to enable long file descriptors in stdio in a backport.

As for the ELF flags, I must admit I am rather unfamiliar with them as I've never had need to dig that deep into it before. I gather this would involve elf_aux_info() and defining a new bit in NT_FREEBSD_FEATURE_CTL or adding a new kind of "note" to the ".note.tag" ELF section. The features I see defined in sys/elf_common.h or displayed by 'elfctl -l' seem to affect very basic thngs such as the layout of a process's virtual memory, randomisation of segment load addresses, no execute bit permissions on memory mappings, stack layout, etc. - all stuff that belongs in either the kernel or the dynamic link loader. I am not sure if libc stdio changes are quite that fundamental to warrant that.

svenski123:
There has been other work and efforts over the years that I am aware of (notably by @bsdjhb)  to hide many or even all of the implementation details of struct FILE from userland to then presumably allow for more wholesale changes within stdio to be made without the risk of breaking applications across the system.

It is certainly possible to do this now, moving the current __sFILE definition into stdio/local.h and leaving a public version of it similar size and alignment with most field name redacted, thus presenting a compatible "Swiss cheese FILE" ABI with mostly holes for the remaining unruly applications and libraries that insist on poking into FILE internals; most notably GNUlib and FreeBSD's own /bin/cat to my knowledge, and then over time, removing cheese and add holes to the public ABI as the fingers get prised off of the FILE innards until you are ultimately left with nothing so that struct FILE can then be left undefined for the public. Though I woud think that having at least the most frequently used buffer pointers and count fields visible for inlined versions of putc() and the like would make sense, and in that regard, I think what DragonFlyBSD have done is a good balance. OpenBSD on the other hand have gone fully opaque with their stdio FILE, and there even a putchar() is a function call.

In any event, before embarking down a seemingly well-trodden road, I would ask if there is sufficient interest or demand for this, or if the odds of such an effort succeeding have increased since the last attempts over the years.

svenski123:
Also as far as this pull request goes, I have pushed additional commits so that it was clear what has changed following the review.

I can certainly squash all of the commits back down into three or four i- one for style/formatting, one for bug fixes incidental to the feature, and one with the feature itself - let me know if this would be helpful.

lib/libc/stdio/findfp.c
brooksdavis: Updates to __sinit_done are not guaranteed to be visible to other threads without atomics or memory barriers, at which point, why bother.

Unless you can show this optimization matters in the real world I'd leave it out and perhaps provide an ifdef to allow this whole set of functionality to be disabled (I'd be included to plan to remove it in FreeBSD 17 or 18 once any problematic programs have been shaken out.)

brooksdavis:
Thanks for digging into the concurrency issues.

Before any commit it would be good so squash down all the cleanups and revisions. I might be inclined to split the first commit up a bit more to in case there's a subtle bug or Hyrum's Law issue in one of the fixes.

Re the future of __sFILE, I think we should be aiming to make it completely opaque over time. I believe that was @bsdjhb's overall goal and we have a need for it in our CHERI work where I'd like to make it entirely inaccessible to prevent attacks on _file[0]. Doing so is somewhat likely to impact performance on embedded microarchitectures doing things we currently inline in a tight loop, but probably won't be particularly noticeable in a modern superscalar microarchitecture. If it does turn out to matter, it might be the case that that the right balance is complete opacity with no inlining by default with a compile time option to define `STDIO_UNSTABLE_INLINED_ABI` where you do the swiss cheese API and are willing to break it with little or no notice. I'd suggest driving that decision based on performance analysis of real applications rather than micro benchmarks.

[0] In our compartmentalization work, we've observed that it's somewhat common for the only I/O in a particular library to be of the form `sprintf(stderr, ...)` which is regrettably a "write to any open file descriptor" gadget given enough of a "write-what-where" gadget to trash _file. We'd like to seal FILE*'s so they are truly opaque handles. Ideally we'd do this sooner rather than later.

lib/libc/stdio/fopen.3
```diff
@@ -320,6 +324,14 @@ may also fail and set
 if the
 .Fa size
 argument is 0.
+.Pp
+The
+.Fn fopen ,
+.Fn fdopen
+and
+.Fn freopen
+functions will enforce the legacy limit of at most 32,767 open files with binaries linked against previous versions of the standard C library. As a debugging aid, this legacy compatibility mode can also be enabled for the time being by setting the environment variable
```
brooksdavis:
```suggestion
functions will enforce the legacy limit of at most 32,767 open files with 
binaries linked against previous versions of the standard C library.
As a debugging aid, this legacy compatibility mode can also be enabled
for the time being by setting the environment variable
```
Wrap lines at <80 columns and new line after full stop.

lib/libc/stdio/fopen.3
```diff
@@ -320,6 +324,14 @@ may also fail and set
...
+.Va LIBC_STDIO_FORCE_SHORT_FILDES_ONLY .
```
brooksdavis:
```suggestion
.Va LIBC_STDIO_FORCE_SHORT_FILDES_ONLY
prior to executing the program.
```
People might think you could `setenv("LIBC_STDIO_FORCE_SHORT_FILDES_ONLY", "true")` later on without similar wording.

lib/libc/stdio/fopen.3
svenski123: Lines have been wrapped.
checkstyle9.pl doesn't seem to work on man page source files unfortunately.

lib/libc/stdio/fopen.3
svenski123: Clarified per your suggestion.

lib/libc/stdio/findfp.c
svenski123: Further edits and code comments for clarity on this, but what is now __stdio_initialised (and formerly __sinit_done) was __sdidinit in the beginning (~90s). Prior to this branch, the only setup that was done was to set the function pointer __cleanup to &_cleanup; in some places this was first time only by checking __sdidinit and other places __cleanup was assigned everytime stdio was called. Replacing this setup everywhere it is needed with a static inlined function is a clear improvement

__stdio_initialised is bss initialised to zero by the loader, only ever set true and this is done in one place only, in a function called via pthread_once() (courtesy of _once()) which guarantees it is called once with all side effects visible before it returns in any thread. This is a standard double-checked initialisation pattern which has minimal impact on performance and code size - (i.e. decompile anywhere __stdio_init_if_needed() is called such as ungetc 'objdump -glrC --symbolize-operands --disassemble-symbols=ungetc libc.so.7' and it's just a load/compare/jump-over-call in the common case).

I note old comments in the code base suggest the reason for the lazy runtime initialisation of __cleanup was to avoid linking in stdio for applications that don't use it. However this is irrelevant for the vast majority of applications that are dynamically linked to libc and it also doesn't work for statically linked applications as examining the result of compiling and statically linking "int main(){}" will reveal the presence of fwrite, funopen, fprintf, fclose, fdclose, fflush and many other stdio functions in the resulting executable.

So lazily setting __cleanup at runtime no longer achieves anything useful and so assuming we are happy to continue having a large chunk of stdio code linked into every C application, then it would be simpler to assign __cleanup = &_cleanup at program load time or even have exit()/abort() call _cleanup() directly (though maybe someone somewhere has a justifiable need for hooking the stdio cleanup function at exit).

Then the code to check the force only environment variable can go in an attribute((constructor)) function and the need for __sdidint and the like goes away entirely.

svenski123:
Thanks for the background on FILE and your perspective; I hadn't heard of CHERI before however is certainly an interesting project, efficient hardware enforced guarded/constrained pointers if used properly would make a whole slew of stack smashing/buffer overrun attacks non-viable.

As regards this pull request, I've reshuffled and squashed the commits down to five (or four setting aside the last commit). The cleanup of __cleanup has been put into a separate commit, __cleanup in exit() is now initialised with a weak reference to _cleanup, so stdio will not get linked in purely on account of stdlib. Putting the force-short-fildes mode envvar check in a __attribute__((constructor)) seems appropriate, lib/libstdbuf does it this way as well.

It did occur to me that instead of the envvar, the force-short-fildes symbol could be exported as a weak symbol, and then to switch it on, one would just need to compile an object exporting the symbol set to one and link it into the binary (i.e. using LD_PRELOAD) which would be another way of controlling the feature.

As regards FILE more generally, I've submitted a separate pull request #2025 to prise _mbstate out the cat's claws as it were so that it can be hidden from the public FILE. Encoding/decoding is character driven and I suspect a fully opaque FILE may impose a significant overhead. Of course the inability to access, control or reset the internal mbstate_t object in FILE is a deficiency in the standard (C95 I believe) hence why /bin/cat succumbed to the temptation to directly access it. The irony is though if one could leave the encoding/decoding to the stream itself and read a whole line or buffer, then it would easily be fast enough with an opaque FILE.

svenski123:
I have reworked this branch into nine commits, consisting of six commits addressing bugs, cleanups and similar issues, two commits implementing the feature itself (namely support for 32-bit file descriptors in stdio) followed by a commit for checkstyle9.pl style edits in files touched by this work.

I have taken out the thread locking fixes relating to the FILE glue list to be worked further independently of this branch.

I believe this branch as a self-contained backwards compatible enhancement to stdio allowing stdio to work in the presence of large numbers of other open files is now largely complete and ready for inclusion which I believe would be of value to the wider FreeBSD community.

I would appreciate a final round of review to identify any remaining areas that need improvement.

At this time I would also ask for a review on pull request #2025 which makes /bin/cat better behaved and not access internal FILE members which will help efforts to make the public struct FILE opaque.

lib/libc/stdio/fdopen.c
```diff
@@ -114,3 +115,16 @@ fdopen(int fd, const char *mode)
 	fp->_close = __sclose;
 	return (fp);
 }
+
+FILE *
+__sfdopen_fbsd10(int fd, const char *mode)
```
brooksdavis: With recent compat code, we'd commonly name this something like `freebsd15___sfdopen`

lib/libc/stdio/stdio.c
```diff
@@ -185,3 +187,14 @@ _cleanup(void)
...
 }
+
+/*
+ * This function runs at program load time before main() runs
+ * and initialises stdio based on the environment.
+ */
+__attribute__((constructor)) static void
+__stdio_init(void)
+{
+	if (getenv("__STDIO_FORCE_SHORT_FILDES") != NULL)
```
brooksdavis: I wonder if this should be secure_getenv?

brooksdavis (commented):
Sorry for dropping this. I was hoping to get @bsdjhb to give it a review and then got busy.

A couple comments:
 - New EBADF errnos should be documented in the appropriate manpages
 - The third commit has a typo "desciptor" and might want an "is"

lib/libc/stdio/stdio.c
svenski123: That would preclude 16-bit fildes compatibility mode for setuid executables; as the other way of doing that by overriding the weak definition of __stdio_force_short_fildes with a preload shareded object won't work.

The compatibility mode was prompted by your suggestion - is not having it available for setuid binaries reasonable?

lib/libc/stdio/fdopen.c
svenski123: There are symbols with an fbsdXY suffix such as those below which is where that came from.
Regardless I have renamed __sfdopen_fbsd10 to freebsd15_fdopen and the other similarly as you suggested.
```
# nm /usr/lib/libc.a|grep fbsd
0000000000000030 T __arc4random_addrandom_fbsd11
0000000000000001 b __arc4random_addrandom_fbsd11.warned
0000000000000000 T __arc4random_stir_fbsd11
0000000000000000 b __arc4random_stir_fbsd11.warned
00000000000000d0 T __getlogin_r_fbsd12
0000000000000100 T __rand_fbsd12
0000000000000170 T __srand_fbsd12
0000000000000180 T __sranddev_fbsd12
0000000000000008 b __sranddev_fbsd12.warned
```

lib/libc/stdio/stdio.c
svenski123: I have made that a secure_getenv for now.

svenski123:
I've explicitly documented EBADF in fopen.3 and printf.3 as it is clearer though both man pages already documented return values of EBADF by way of reference to errors returned by fcntl(2) and open(2) respectively.

Thanks for pointing out the typo in the commit comment which I have since corrected.

All fixes have been merged/rebased into the relevant commit; I have also rebased the chain of nine commits against main.

svenski123:
Force pushed two commit to address issues flagged up by the CI Style Checker.

bms:
This is so weird. I had no idea these limits were present. It is of potential concern for a currently back-burnered effort; some (not so special) sauce to go on top of CHERI and capsicum.

lib/libc/stdio/stdio.c
brooksdavis: My thought is that arbitrary users being able to change program behavior around edge cases seems problematic for suid programs. In the short term, this might make things slightly worse, but our goal is certainly to get to the point where we remove this support.

svenski123:
Hi @brooksdavis, if time permits and the opportunity presents itself at BSDCan, could you have a quick chat with @bsdjhb about this PR? I would appreciate it.

PR #2028, opened 2026-02-16
/bin/sh: Handle long input lines (>1024 chars) correctly with libedit

svenski123:
Interactive /bin/sh shells using libedit do not handle long command lines in excess of 1024 characters correctly.

While the shell will execute the input command line correctly initially, attempts to bring the command back from history then fail as it will have chopped it into multiple history entires, each one at most 1024 bytes and only the last one line terminated.

Additionally entering a long command line but without keying return, then going to the previous entry (<up> or ^P) and then back to the current line (<down> or ^N) will return a garbled form the command with trailing nulls and garbage data.

This pull request addresses both of these with edits to bin/sh and contrib/libedit.

The command below will generate a sufficiently long command line that can be selected and pasted into a terminal window to reproduce the problem. Alternatively running a build (i.e. make buildworld) and taking one of the longer compiler invocations will also do the trick.
```
# This one line python will generate a suitably long command line for /bin/sh
$ echo "print(f\"true {''.join(c*1024 for c in 'wxyz')}\")" | python3
true ww[...1020 w's...]wwxx[...1020 x's...]xxyy[...1020 y's...]wyyzz[...1020 z's...]zz

# Paste the preceding into a terminal window and hit return:
$ true www...wwwxxx...xxxyyy...yyyzzz...zzz

# Paste the preceding into a terminal window and press ctrl-P then ctrl-N; see ^@ at right: 
$ true www...www^@^@^@...^@^@^@
```

### DIFF
--- a/bin/sh/input.c
+++ b/bin/sh/input.c
@@ -81,6 +81,7 @@ struct parsefile {
 	int lleft;		/* number of lines left in this buffer */
 	const char *nextc;	/* next char in buffer */
 	char *buf;		/* input buffer */
+	size_t bufsize;		/* input buffer size */
 	struct strpush *strpush; /* for pushing strings at this level */
 	struct strpush basestrpush; /* so pushing one is fast */
 };
@@ -93,7 +94,8 @@ const char *parsenextc;		/* copy of parsefile->nextc */
 static char basebuf[BUFSIZ + 1];/* buffer for top level input file */
 static struct parsefile basepf = {	/* top level input file */
 	.nextc = basebuf,
-	.buf = basebuf
+	.buf = basebuf,
+	.bufsize = sizeof(basebuf),
 };
 static struct parsefile *parsefile = &basepf;	/* current input file */
 int whichprompt;		/* 1 == PS1, 2 == PS2 */
@@ -127,52 +129,61 @@ static int
 preadfd(void)
 {
 	int nr;
-	parsenextc = parsefile->buf;
 
 retry:
 #ifndef NO_HISTORY
 	if (parsefile->fd == 0 && el) {
-		static const char *rl_cp;
-		static int el_len;
+		const char *line;
 
-		if (rl_cp == NULL) {
-			el_resize(el);
-			rl_cp = el_gets(el, &el_len);
+		el_resize(el);
+		line = el_gets(el, &nr);
+		if (nr > 0 && parsefile->bufsize < (size_t)nr + 1) {
+			size_t bufsize;
+
+			INTOFF;
+			if (parsefile->buf != basebuf) {
+				ckfree(parsefile->buf);
+				parsefile->buf = NULL;
+				parsefile->bufsize = 0;
+			}
+			bufsize = (size_t)nr + BUFSIZ + 1;
+			bufsize -= bufsize % BUFSIZ;
+			parsefile->buf = ckmalloc(bufsize);
+			parsefile->bufsize = bufsize;
+			INTON;
 		}
-		if (rl_cp == NULL)
-			nr = el_len == 0 ? 0 : -1;
-		else {
-			nr = el_len;
-			if (nr > BUFSIZ)
-				nr = BUFSIZ;
-			memcpy(parsefile->buf, rl_cp, nr);
-			if (nr != el_len) {
-				el_len -= nr;
-				rl_cp += nr;
-			} else
-				rl_cp = NULL;
-		}
+		if (nr > 0 && line != NULL)
+			memcpy(parsefile->buf, line, nr);
+		else
+			nr = nr ? -1 : 0;
 	} else
 #endif
-		nr = read(parsefile->fd, parsefile->buf, BUFSIZ);
+	nr = read(parsefile->fd, parsefile->buf, parsefile->bufsize - 1);
 
-	if (nr <= 0) {
-                if (nr < 0) {
-                        if (errno == EINTR)
-                                goto retry;
-                        if (parsefile->fd == 0 && errno == EWOULDBLOCK) {
-                                int flags = fcntl(0, F_GETFL, 0);
-                                if (flags >= 0 && flags & O_NONBLOCK) {
-                                        flags &=~ O_NONBLOCK;
-                                        if (fcntl(0, F_SETFL, flags) >= 0) {
-						out2fmt_flush("sh: turning off NDELAY mode\n");
-                                                goto retry;
-                                        }
-                                }
-                        }
+	if (nr < 0)
+		switch (errno) {
+			int flags;
+
+		case EINTR:
+			goto retry;
+		case EWOULDBLOCK:
+			if (parsefile->fd != 0)
+				break;
+			if ((flags = fcntl(0, F_GETFL, 0)) < 0)
+				break;
+			if (!(flags & O_NONBLOCK))
+				break;
+			if (fcntl(0, F_SETFL, flags & ~O_NONBLOCK) < 0)
+				break;
+			out2fmt_flush("sh: turning off NDELAY mode\n");
+			goto retry;
                 }
-                nr = -1;
-	}
+	else if (nr > 0)
+		parsefile->buf[nr] = '\0';
+	else
+		nr = -1;
+
+	parsenextc = parsefile->buf;
 	return nr;
 }
 
@@ -189,7 +200,8 @@ retry:
 int
 preadbuffer(void)
 {
-	char *p, *q, *r, *end;
+	const char *end;
+	char *q, *r;
 	char savec;
 
 	while (parsefile->strpush) {
@@ -208,31 +220,22 @@ preadbuffer(void)
 		return PEOF;
 
 again:
-	if (parselleft <= 0) {
-		if ((parselleft = preadfd()) == -1) {
-			parselleft = parsenleft = EOF_NLEFT;
-			return PEOF;
-		}
+	if (parselleft <= 0 && (parselleft = preadfd()) == -1) {
+		parselleft = parsenleft = EOF_NLEFT;
+		return (PEOF);
 	}
-
-	p = parsefile->buf + (parsenextc - parsefile->buf);
-	end = p + parselleft;
-	*end = '\0';
-	q = strchrnul(p, '\n');
-	if (q != end && *q == '\0') {
+	end = parsenextc + parselleft;
+	q = strchrnul(parsenextc, '\n');
+	if (*q == '\0' && q != end) {
 		/* delete nul characters */
-		for (r = q; q != end; q++) {
+		for (r = q++; q != end; q++)
 			if (*q != '\0')
 				*r++ = *q;
-		}
-		parselleft -= end - r;
-		if (parselleft == 0)
-			goto again;
-		end = p + parselleft;
-		*end = '\0';
-		q = strchrnul(p, '\n');
+		*r = '\0';
+		parselleft = r - parsenextc;
+		goto again;
 	}
-	if (q == end) {
+	if (*q == '\0') {
 		parsenleft = parselleft;
 		parselleft = 0;
 	} else /* *q == '\n' */ {
@@ -307,7 +310,7 @@ pushstring(const char *s, int len, struct alias *ap)
 	INTOFF;
 /*out2fmt_flush("*** calling pushstring: %s, %d\n", s, len);*/
 	if (parsefile->strpush) {
-		sp = ckmalloc(sizeof (struct strpush));
+		sp = ckmalloc(sizeof(struct strpush));
 		sp->prev = parsefile->strpush;
 		parsefile->strpush = sp;
 	} else
@@ -391,15 +394,15 @@ setinputfile(const char *fname, int push, int verify)
 void
 setinputfd(int fd, int push)
 {
-	if (push) {
+	if (push)
 		pushfile();
-		parsefile->buf = ckmalloc(BUFSIZ + 1);
-	}
 	if (parsefile->fd > 0)
 		close(parsefile->fd);
 	parsefile->fd = fd;
-	if (parsefile->buf == NULL)
+	if (parsefile->buf == NULL) {
 		parsefile->buf = ckmalloc(BUFSIZ + 1);
+		parsefile->bufsize = BUFSIZ + 1;
+	}
 	parselleft = parsenleft = 0;
 	plinno = 1;
 }
@@ -410,14 +413,12 @@ setinputfd(int fd, int push)
  */
 
 void
-setinputstring(const char *string, int push)
+setinputstring(const char *string)
 {
 	INTOFF;
-	if (push)
-		pushfile();
+	pushfile();
 	parsenextc = string;
 	parselleft = parsenleft = strlen(string);
-	parsefile->buf = NULL;
 	plinno = 1;
 	INTON;
 }
@@ -434,15 +435,12 @@ pushfile(void)
 {
 	struct parsefile *pf;
 
+	pf = (struct parsefile *)ckmalloc(sizeof(struct parsefile));
+	*pf = (struct parsefile){ .prev = parsefile, .fd = -1 };
 	parsefile->nleft = parsenleft;
 	parsefile->lleft = parselleft;
 	parsefile->nextc = parsenextc;
 	parsefile->linno = plinno;
-	pf = (struct parsefile *)ckmalloc(sizeof (struct parsefile));
-	pf->prev = parsefile;
-	pf->fd = -1;
-	pf->strpush = NULL;
-	pf->basestrpush.prev = NULL;
 	parsefile = pf;
 }
 
--- a/bin/sh/eval.c
+++ b/bin/sh/eval.c
@@ -153,7 +153,7 @@ evalstring(const char *s, int flags)
 	flags &= ~EV_EXIT;
 	any = 0;
 	setstackmark(&smark);
-	setinputstring(s, 1);
+	setinputstring(s);
 	while ((n = parsecmd(0)) != NEOF) {
 		if (n != NULL && !nflag) {
 			if (flags_exit && preadateof())
--- a/bin/sh/input.h
+++ b/bin/sh/input.h
@@ -54,7 +54,7 @@ void pungetc(void);
 void pushstring(const char *, int, struct alias *);
 void setinputfile(const char *, int, int);
 void setinputfd(int, int);
-void setinputstring(const char *, int);
+void setinputstring(const char *);
 void popfile(void);
 struct parsefile *getcurrentfile(void);
 void popfilesupto(struct parsefile *);
--- a/bin/sh/parser.c
+++ b/bin/sh/parser.c
@@ -1167,7 +1167,7 @@ parsebackq(char *out, struct nodelist **pbqlist,
 		INTOFF;
 		ostr = ckmalloc(olen);
 		memcpy(ostr, stackblock(), olen);
-		setinputstring(ostr, 1);
+		setinputstring(ostr);
 		INTON;
         }
 	nlpp = pbqlist;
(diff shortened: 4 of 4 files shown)


bin/sh/input.c
```diff
@@ -244,13 +249,20 @@ preadbuffer(void)
...
+	 */
+	if (parsefile->fd == 0 && hist && rl_cp_last != NULL) {
+		if (rl_cp_last[strspn(rl_cp_last, " \t\n")] != '\0') {
+			HistEvent he;
+			INTOFF;
+			history(hist, &he, whichprompt == 1 ? H_ENTER : H_ADD,
+			    rl_cp_last);
+			INTON;
+		}
+		rl_cp_last = NULL;
```
jillest: Probably nobody cares but right now history still works after `set +o emacs`.

bin/sh/input.c
```diff
@@ -136,7 +139,9 @@ preadfd(void)
 		if (rl_cp == NULL) {
 			el_resize(el);
 			rl_cp = el_gets(el, &el_len);
-		}
+			rl_cp_last = rl_cp;
+		} else
+			rl_cp_last = NULL;
```
jillest: This looks like it should work, but how about skipping parsefile->buf entirely when libedit is being used (which will probably involve calling `el_gets` from `preadbuffer`)?
If it is too hard, please leave it like this.

contrib/libedit/hist.c
```diff
@@ -223,9 +223,13 @@ hist_command(EditLine *el, int argc, const wchar_t **argv)
...
 libedit_private int
 /*ARGSUSED*/
-hist_enlargebuf(EditLine *el, size_t oldsz, size_t newsz)
+hist_enlargebuf(EditLine *el, size_t newsz)
 {
 	wchar_t *newbuf;
+	const size_t oldsz = el->el_history.sz;
+
+	if (newsz <= oldsz)
+		return 1;
```
jillest, line 232: The libedit change should be submitted to the upstream (NetBSD).

bin/sh/input.c
svenski123: Maybe vi mode was still active?
If I do `set +o emacs vi`, then history doesn't work anymore (i.e. ^P/^N and <ESC> + j or k and <up>/<down> just self insert)
Interestingly in vi mode, <ESC>  + j or k work as expected, ^P/^N don't do anything as expected but <up>/<down> still go back and forward in history

bin/sh/input.c
svenski123: Not clear to me that that's safe - if libedit reads in long multi-line input with a `set +VEi` at the beginning, shell could turn off editing, call el_end() which would free the buffer el_gets() returns and the shell could potentially continue parsing the now-freed memory.
Also tried running valgrind on /bin/sh, result wasn't pretty.

contrib/libedit/hist.c
svenski123: I've submitted the following problem report to NetBSD.
https://gnats.netbsd.org/cgi-bin/query-pr-single.pl?number=60050

svenski123: I've also submitted PRs 60052 and 60054 to NetBSD, all three fixes have been committed to the NetBSD CVS repository.

I've included a commit with the latest libedit from NetBSD-CURRENT upstream.

Can you get the vendored contrib/libedit tree in FreeBSD updated or advise on how that should be done?

bin/sh/input.c
svenski123: Latest version of this PR gets rid of all the rl_cp gumpf, the line as returned by libedit is copied in full, parsefile->buf is enlarged as necessary.

svenski123:
I have rebased and reworked this PR, some further refactoring was necessary but the end result is less brittle and more straightforward.

Three bugs in libedit cropped up during this work, fixes for all three have all been submitted upstream to NetBSD and merged there.

The changes to /bin/sh here can be applied before the libedit vendor branch is updated, though the bugs in libedit will cause problems with long lines, much as they currently do.

contrib/libedit/hist.c
concussious: > Can you get the vendored contrib/libedit tree in FreeBSD updated

This can only be done by maintainers due to our workflow. @bapt will do it, or if he's too busy, I will.

bin/sh/input.c
jillest: I mean the history is still tracked, like this:
```
$ set +o emacs +o vi
$ echo a
a
$ echo b
b
$ fc -l
    1 set +o emacs +o vi
    2 echo a
    3 echo b
$ 
```
Of course, the history is only accessible using `fc` or by re-enabling `emacs` or `vi` mode.

svenski123: Ah ok I see what you're saying.
The history mode is actually independent of line editing, it's always enabled for interactive shells and disabled for non-interactive shells.
So if you do `set +i` (or equivalently `set +o interactive`) to turn off interactive mode, history should be disabled as well.
You won't have prompt but you can then turn interactive mode on again with `set -i` or `set -o interactive`.

svenski123: FYI you can run a shell script in interactive mode with either
`sh -i script.sh`
or
`sh -i <script.sh`
In the latter case, the script commands will get added to history (IIRC the code in input.c will only save a line to history if it was read from file descriptor 0 i.e. stdin).

It also looks like if you switch off interactive mode and then turn it on again, it won't reload history from ~/.sh_history but instead will be blank, not sure if that's really the best behaviour.

svenski123:
I noticed contrib/libedit has been updated (thank you!) so I have removed the netbsd libedit commit and rebased the remaining commit against freebsd/main.

With the libedit fixes nulls (^@) should now no longer appear in history however /bin/sh will still truncate lines at 1024 characters - this pull request addresses that last issue.

Incidentally I have also noticed that /bin/sh does not seem to save history if terminated by a signal (i.e. when running `reboot` / `shutdown -r now`); I will look into this and if appropriate will include a fix here otherwise will put it in a separate pull request.

concussious:
> Incidentally I have also noticed that /bin/sh does not seem to save history if terminated by a signal (i.e. when running reboot / shutdown -r now); I will look into this and if appropriate will include a fix here otherwise will put it in a separate pull request.

Wow that would be amazing! I am also affected by this.

svenski123:
On shutdown `/sbin/init` sends every process a SIGTERM and then a SIGKILL for stragglers. `/bin/sh` running in interactive mode ignores SIGTERM and SIGKILL of course can't be caught, so no joy there for history saving.

However when `/sbin/init` kills the `/usr/libexec/sshd-session` parent of `/bin/sh`, it closes the controlling terminal which causes a SIGHUP to be sent to `/bin/sh` which can be caught. (Here I am assuming one is logging in with `sshd` but if you're coming in over a serial line or something else then the same thing should apply to the applicable `getty` process when it closes the controlling `/dev/tty...` device.)

Trapping SIGHUP and having the shell exit normally works insofar as history saving goes. I've dropped the following file in `/etc/profile.d` on my machines so that all interactive `/bin/sh` shells pick it up. Aside from trapping SIGHUP it also increases the default history size from 100 to a more reasonable 4096 entries.
#### /etc/profile.d/history.sh
```
HISTSIZE=4096
trap exit HUP
```
`/bin/bash` and `/bin/zsh` both handle SIGHUP in interactive mode as part of job control - search for `HUP` in `man 1 bash` and `man 1 zshall` - so these shells work out of the box as far as saving history on a SIGHUP goes. `/bin/sh` even in interactive mode with job control enabled doesn't do anything with `SIGHUP` by default and the [POSIX.1-2017](https://pubs.opengroup.org/onlinepubs/9799919799/) standard doesn't really make clear if it needs to or not.

The default action for SIGHUP (for any application mind you not `/bin/sh` in particular) is [Abnormal termination of the process](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/signal.h.html) which means ["...the process is terminated as if by a call to _exit()..."](https://pubs.opengroup.org/onlinepubs/9799919799/functions/V2_chap02.html#tag_16_04_03) and that appears to be exactly what `/bin/sh` does by default when the system goes down. Perhaps not the most useful behaviour for an interactive, job controlling shell and perhaps it could do more by default and still conform to POSIX but that should probably be discussed first with the wider community.

svenski123:
@concussious @jillest
It's been over a month, please kindly merge these fixes into current or advise what needs to be addressed prior to merge and I will address it within 24 hours.

bin/sh/input.c
```diff
@@ -126,53 +128,47 @@ pgetc(void)
...
+
+	if (nr < 0) {
+		if (errno == EINTR)
+			goto retry;
+		if (errno == EWOULDBLOCK && parsefile->fd == 0 && (flags =
+		    fcntl(0, F_GETFL, 0)) >= 0 && (flags & O_NONBLOCK) != 0 &&
+		    fcntl(0, F_SETFL, flags & ~O_NONBLOCK) >= 0) {
+			out2fmt_flush("sh: turning off NDELAY mode\n");
+			goto retry;
+		}
```
jillest: Please keep this block as it was (but re-indented, matching the rest of the new logic). I liked the separate statements better.

bin/sh/input.c
```diff
@@ -208,39 +205,24 @@ preadbuffer(void)
...
-		parsenleft = q - parsenextc;
-		parselleft -= parsenleft;
+		*r = '\0';
+		parselleft = r - parsenextc;
+		goto again;
 	}
-	parsenleft--;
+
+	parsenleft = (*q == '\0' ? parselleft : ++q - parsenextc);
+	parselleft -= parsenleft--;
```
jillest, line 247: I don't really like that this logic now has `++` and `--` operators in larger expressions. My personal preference is that `++` and `--` be used either as standalone statements or with indirection like `*r++` or `a[i++]`.
Can you make it a bit more similar to how used to look?

bin/sh/input.c
```diff
@@ -410,14 +392,12 @@ setinputfd(int fd, int push)
  */
 
 void
-setinputstring(const char *string, int push)
+setinputstring(const char *string)
```
jillest, line 416: :+1:, should have been done earlier

bin/sh/input.c
```diff
@@ -439,10 +419,7 @@ pushfile(void)
 	parsefile->nextc = parsenextc;
 	parsefile->linno = plinno;
 	pf = (struct parsefile *)ckmalloc(sizeof (struct parsefile));
-	pf->prev = parsefile;
-	pf->fd = -1;
-	pf->strpush = NULL;
-	pf->basestrpush.prev = NULL;
+	*pf = (struct parsefile){ .prev = parsefile, .fd = -1 };
```
jillest: 👍

bin/sh/input.c
```diff
@@ -126,53 +128,47 @@ pgetc(void)
...
+
+		el_resize(el);
+		line = el_gets(el, &nr);
+		if (nr > 0 && parsefile->bufsize < (size_t)nr + 1) {
+			INTOFF;
+			if (parsefile->buf != basebuf)
+				ckfree(parsefile->buf);
+			parsefile->bufsize = (size_t)nr + BUFSIZ + 1;
+			parsefile->bufsize -= parsefile->bufsize % BUFSIZ;
+			parsefile->buf = ckmalloc(parsefile->bufsize);
```
jillest: We should do something here to prevent `parsefile->buf` being freed twice if `ckmalloc` throws an exception.

There may be other places with this issue, but they should not be fixed in this pull request.

bin/sh/input.c
```diff
@@ -452,6 +429,8 @@ popfile(void)
 {
 	struct parsefile *pf = parsefile;
 
+	if (pf == &basepf)
+		return;
```
jillest: This feels like it might be hiding bugs, since previously the code avoided calling `popfile` if there was nothing to pop? For which scenario do we need this change?

If it is not supposed to happen but you want to avoid the UB, a conditional message to stderr and `abort()` may be better.

jillest (commented):
I tried the reproduction steps which work (which is nice), but I'm trying to write an automated test and encounter a bug:
```sh
v=1234
v=$v$v$v$v
v=$v$v$v$v
v=$v$v$v$v
v=0$v$v$v$v@
echo ${#v}
result=$(ENV= HISTFILE=/dev/null ${SH} +m -i -o emacs <<EOF
set -x
printf '%s\n' "$v"
fc -l
EOF
)
set -x
: result is "$result"
```
When I save it to `parser/long-line1.0` and run it like `SH=sh sh parser/long-line1.0`, the result ends with `2341    3 23412341234123412341234@"`, while with `SH=bash sh parser/long-line1.0` or `SH=mksh sh parser/long-line1.0` the result is correct. I haven't completed the automated check yet because of this issue. Do you have an idea?

I also have some code issues.

svenski123:
> I tried the reproduction steps which work (which is nice), but I'm trying to write an automated test and encounter a bug:
> 
> ```shell
> v=1234
> v=$v$v$v$v
> v=$v$v$v$v
> v=$v$v$v$v
> v=0$v$v$v$v@
> echo ${#v}
> result=$(ENV= HISTFILE=/dev/null ${SH} +m -i -o emacs <<EOF
> set -x
> printf '%s\n' "$v"
> fc -l
> EOF
> )
> set -x
> : result is "$result"
> ```
> 
> When I save it to `parser/long-line1.0` and run it like `SH=sh sh parser/long-line1.0`, the result ends with `2341 3 23412341234123412341234@"`, while with `SH=bash sh parser/long-line1.0` or `SH=mksh sh parser/long-line1.0` the result is correct. I haven't completed the automated check yet because of this issue. Do you have an idea?

The shell executed in the command substitution has a stdin which is not a tty thus line editing mode isn't enabled, only history so preadfd() in input.c does not use el_gets() from editline but rather a plain read() into the default 1024 byte buffer instead. Thus the long input line gets split into two history lines numbered 2 and 3, where the first one (numbered two) lacks a terminating newline.

If you run the shell with a pseudo tty, then it'll use el_gets() which will read in the whole line. i.e.
```
result=$(ENV= HISTFILE=/dev/null script -qe /dev/null ${SH} +m -i -o emacs <<EOF
```
Of course the output will be raw and have carriage returns etc. but the whole line will be there.

For the purposes of automated testing, you can set HISTFILE to a temporary file, run the sub shell using script with the long command line while discarding output and then examine the temporary history file to ensure the whole command line is there.

I should point out that the behaviour of history when running an interactive shell without a tty on standard input is unchanged from before. I am not aware of many uses for command line history with interactive shells that do not have a controlling tty (though automated testing of interactive shell features would be one). Perhaps POSIX has something to say on it though I would doubt it.

bin/sh/input.c
svenski123: Popping the base parsefile is always an error; I've not come across a scenario that does this however I've not gone out of my to look for one. The pattern of pushfile() / popfile() resulting from a parse could be quite complicated and it's not readily apparent to me that it is impossible for popfile() to be called more times than pushfile() has.

I note popfilesupto() does check for popping &basepf and calls error() if it comes to it.
As error() will perform a longjmp() if a handler is installed, it is conceivable that execution could continue.
Though I would consider popping the base parsefile with either popfile() or popfilesupto() to be a logic error; it should not be possible for this situation to occur when running any script.

I don't know what the userland equivalent of KASSERT in freebsd is but I think that would be appropriate here (i.e. production build it favours uptime, debug build it fails noisily so the problem is exposed.)

In any event I've removed the `if (pf == &basepf) return;` line from popfile() so the status quo there remains unchanged.

bin/sh/input.c
svenski123: parsefile->buf and ->bufsize are now zeroed after ->buf is freed which eliminates the possibility of a double free.

parsefile->buf is then assigned from ckmalloc() and only if that succeeds is ->bufsize assigned.

The only other place that mallocs a parsefile buffer is setinputfd() which only does this if parsefile->buf is NULL; ->bufsize is only assigned if ckmalloc() succeeds.

popfile() is the only other place that frees a parsefile buffer; it does this unconditionally if ->buf is not NULL, however as long as ->buf isn't pointing at &basepf this is fine, and it won't unless popfile() is called when there are no other pushed parsefiles which is the situation discussed at the other comment for popfile().

bin/sh/input.c
svenski123: I've expanded it out a bit with the pre/post increment/decrement operators separated out - I trust it's now clearer and more inline with what you expect.

bin/sh/input.c
svenski123: Done!

svenski123:
I took a further look at history when standard input isn't a tty as having the 1,024 byte limitation there seems a bit off.

I also took a look at the following POSIX standards:
https://pubs.opengroup.org/onlinepubs/9799919799/idx/shell.html
https://pubs.opengroup.org/onlinepubs/9799919799/utilities/sh.html
https://pubs.opengroup.org/onlinepubs/9799919799/utilities/fc.html

Unfortunately there is some ambiguity in the POSIX standards as to what exactly a command is for the purposes of history. The specification for the sh command does state that unlimited line lengths must be supported and the fc command document specifies multiple output formats for the fc -l command, including a format for the display of lines after the first line for a multi-line <command>, which would suggest that a single command for shell history is intended to be something more than just everything up to the next newline.

Currently command insertion into history is very close to the raw read before parsing and lexical analysis with only minimal processing (i.e. lines consisting of only whitespace are not written to history).

Pushing the saving of command line history storage down into the parser when it recognises or accepts a complete command would make the processing consistent and easier in some respects though formatting and escaping from the initial input would likely be lost. On the hand leaving it further up the stack introduces some redundant processing to determine what a command line is; for example libedit appears to do some multi-line processing itself, reading multiple lines of input when a quoted string is unterminated for example and returning it all to the shell as a single input.

To get consistent behaviour for interactive shells when standard input is a tty and when it is not, the multiline input processing of libedit would need to be reproduced for non-tty input.

I threw together the diff below which does pass your testcase without requiring the use of script, however it does leave rather much to be desired.

I'd think work to fix history for interactive non-tty shells should be a separate change from this.

```
diff --git a/bin/sh/input.c b/bin/sh/input.c
index 8e4e68dea6f6..c93801026601 100644
--- a/bin/sh/input.c
+++ b/bin/sh/input.c
@@ -38,2 +38,3 @@
 #include <unistd.h>
+#include <stdbool.h>
 #include <stdlib.h>
@@ -247,9 +248,31 @@ preadbuffer(void)
 #ifndef NO_HISTORY
-       if (parsefile->fd == 0 && hist &&
-           parsenextc[strspn(parsenextc, " \t\n")] != '\0') {
-               HistEvent he;
-               INTOFF;
-               history(hist, &he, whichprompt == 1 ? H_ENTER : H_ADD,
-                   parsenextc);
-               INTON;
+       static bool histlinecont = false;
+
+       if (parsefile->fd != 0 || !hist) {
+               histlinecont = false;
+       } else {
+               char *p, *e;
+
+               for (p = &parsefile->buf[parsenextc - parsefile->buf];
+                    *(e = strchrnul(p, '\n')) == '\n'; p = e)
+                       if (e++ != p + strspn(p, " \t")) {
+                               const char c = *e;
+                               HistEvent he;
+
+                               INTOFF;
+                               *e = '\0';
+                               history(hist, &he, histlinecont ? H_ADD : H_ENTER, p);
+                               *e = c;
+                               histlinecont = false;
+                               INTON;
+                       }
+
+               if (e != p && (histlinecont || e != p + strspn(p, " \t"))) {
+                       HistEvent he;
+
+                       INTOFF;
+                       history(hist, &he, histlinecont ? H_ADD : H_ENTER, p);
+                       histlinecont = true;
+                       INTON;
+               }
        }
```

svenski123:
That should cover the review comments.
There were a number of style complaints thrown up by CI.
I believe most were pre-existing; let me know if you'd like me to fix any of them.

bin/sh/input.c
svenski123: In the end I refactored this into a switch statement to remove the repeated if-then nesting and fix the line-too-long style check error on the out2fmt_flush line. Like the original version, the statements are all separate so should be easy to read.

svenski123:
Fixed remaining style errors and sqashed down to one commit.

jillest (approved):
I will commit soon, along with an automated test using script(1).

jillest:
Merged.

I'm fine with leaving history for interactive non-tty shells as is. It may not be worth the extra code since it is likely to be used in test cases only.

PR #2047, opened 2026-02-24
rc.8: Add rc.suspend as an alias

patmaddox:
Make it consistent with the other lifecycle scripts, thus making it more findable.


concussious:
Hey pat! 

So, why?

patmaddox:
> So, why?

When I learned of its existence, I looked up the man page, and there was none. Had it been consistent with the other lifecycle scripts (resume and shutdown in particular) I may have learned of it sooner. In short, consistency and discoverability.

concussious:
That makes sense! You should put that in the commit message so others can see!

patmaddox:
good thinking :)  updated. thanks

concussious (approved):
Thanks Pat!

PR #2050, opened 2026-02-25
share/examples/kld/cdev: Replace copystr() with copyout()

isoux:
The example cdev module used copystr() to copy data from kernel space to user space. copystr() is intended for kernel-to-kernel copies and expands to strlcpy(), which does not perform user address validation.

Using copystr() in mydev_read() may lead to a kernel page fault when writing to user memory.

Replace copystr() with copyout(), which is the correct API for copying data from kernel space to user space.

The existing formatting and coding style of this example file are intentionally preserved to keep the change minimal and focused solely on correcting the API usage.

### DIFF
--- a/share/examples/kld/cdev/module/cdev.c
+++ b/share/examples/kld/cdev/module/cdev.c
@@ -168,15 +168,20 @@ mydev_write(struct cdev *dev, struct uio *uio, int ioflag)
 int
 mydev_read(struct cdev *dev, struct uio *uio, int ioflag)
 {
-    int err = 0;
+	int err = 0;
+	size_t outlen;
 
-    printf("mydev_read: dev_t=%lu, uio=%p, ioflag=%d\n",
-	dev2udev(dev), uio, ioflag);
+	printf("mydev_read: dev_t=%lu, uio=%p, ioflag=%d\n",
+	    dev2udev(dev), uio, ioflag);
 
-    if (len <= 0) {
-	err = -1;
-    } else {	/* copy buf to userland */
-	copystr(&buf, uio->uio_iov->iov_base, 513, &len);
-    }
-    return(err);
+	if (len <= 0) {
+		err = EINVAL;
+	} else {	/* copy buf to userland */
+		outlen = strlen(buf) + 1;
+		err = copyout(buf, uio->uio_iov->iov_base, outlen);
+		if (err == 0)
+			uio->uio_resid -= outlen;
+	}
+
+	return (err);
 }


share/examples/kld/cdev/module/cdev.c
```diff
@@ -174,9 +174,11 @@ mydev_read(struct cdev *dev, struct uio *uio, int ioflag)
 	dev2udev(dev), uio, ioflag);
 
     if (len <= 0) {
-	err = -1;
+        err = EINVAL;
     } else {	/* copy buf to userland */
-	copystr(&buf, uio->uio_iov->iov_base, 513, &len);
+        err = copyout(buf, uio->uio_iov->iov_base, len);
+        if (err == 0)
+            uio->uio_resid -= len;
```
spmzt: style(9): Indentation is an 8 character tab.
Please use tab instead of spaces.

spmzt (changes requested):

share/examples/kld/cdev/module/cdev.c
bsdimp: I'm still not sure that code is right, since the old code copied only through the first NUL.

isoux: Thanks for the review.

Indentation has been updated to use tabs according to style(9).
The read path now preserves the original copystr() semantics by
copying strlen(buf) + 1 bytes, including the terminating NUL.

The original example file mixes spaces and tabs in several places,
so I initially followed the existing formatting to keep the change
minimal. This has now been corrected in the modified section.

If this change is accepted, I would be happy to submit a separate
cleanup patch to bring the entire example in line with style(9).

share/examples/kld/cdev/module/cdev.c
```diff
@@ -168,15 +168,20 @@ mydev_write(struct cdev *dev, struct uio *uio, int ioflag)
...
+	int err = 0;
+	size_t outlen;
+
+	printf("mydev_read: dev_t=%lu, uio=%p, ioflag=%d\n",
+	    dev2udev(dev), uio, ioflag);
+
+	if (len <= 0) {
+		err = EINVAL;
+	} else {	/* copy buf to userland */
+		outlen = strlen(buf) + 1;
```
kostikbel, line 180: The copy ignores the uio length and uio segments.  Same is true for mydev_write().

The whole example is not mt-safe.

isoux: Thanks for the review.

I see that the current implementation still ignores uio_resid and iov segments, and that the example itself is not fully correct or mt-safe.

Would you prefer that I:

1. adjust this change to properly respect uio_resid (keeping the scope minimal), or
2. rework the example more broadly to address these issues?

My initial goal was to keep the change minimal and focused on replacing copystr() with the correct API.

kostikbel: This should be either significantly reworked, keeping userspace interface for the userspace program (I did not looked at it, there should be surprises), or completely removing. s/copystr/copyin/ is palliative.

isoux: Thanks, that makes sense.

Given that the example has more fundamental issues, would you prefer:

a follow-up change that properly reworks the read/write paths (uio handling, etc.), or
that this example be removed entirely?

I can work on a more complete rewrite if that would be useful, but I’d prefer to align on the intended direction first.

kostikbel: I do not have an opinion.  Generally fixing things is better than dropping them, but it requires qualified and willing person to work on it.

isoux: Thanks, understood.

In that case, I’d like to try improving the example rather than just applying a minimal fix. My goal would be to handle the uio interface correctly while keeping the code simple enough for demonstration purposes.

I’ll post an updated version here — feedback is very welcome if I’m going in the wrong direction.

PR #2053, opened 2026-02-28
arm64: Expose ESR in mcontext

ararslan:
At present, the exception syndrome register (ESR) is not accessible via the user context on FreeBSD AArch64, which makes it difficult to determine the cause of an exception. For example, a signal handler might get a `SIGSEGV` with `SEGV_ACCERR`, but it can't know whether that occurred due to a bad read or write. The change implemented here includes ESR in `struct __mcontext`, thereby allowing access via `ucontext_t` for use by signal handlers, sanitizers, debuggers, etc. Note that this addition consumes one of the seven spare 64-bit slots in `struct __mcontext`.

Apologies if this should have been submitted via Phabricator instead of GitHub; I can resubmit there if needed.

For reference, in case it's useful:
- A freebsd-hackers mailing list [thread](https://lists.freebsd.org/archives/freebsd-hackers/2026-February/005857.html) thread I started around this
- The [shenanigans](https://github.com/freebsd/freebsd-src/blob/178d0b5b8da7480f455273aedf40dd8f1e785d3f/contrib/llvm-project/compiler-rt/lib/sanitizer_common/sanitizer_linux.cpp#L1936-L1957) that compiler-rt sanitizer functions use to get ESR from `ucontext_t` on Linux


bsdimp (approved):
Looks good to me. We'll let @zxombie have a chance to object before committing though.

sys/arm64/arm64/exec_machdep.c
```diff
@@ -854,6 +857,7 @@ sendsig(sig_t catcher, ksiginfo_t *ksi, sigset_t *mask)
 	tf->tf_x[8] = (register_t)catcher;
 	tf->tf_sp = (register_t)fp;
 	tf->tf_elr = (register_t)PROC_SIGCODE(p);
+	tf->tf_esr = (register_t)READ_SPECIALREG(esr_el1);
```
zxombie: Why are you reading `esr_el1` here? It might have changed, e.g. if the thread has migrated to a new CPU.

sys/arm64/include/ucontext.h
```diff
@@ -76,9 +76,11 @@ struct __mcontext {
 	struct fpregs	mc_fpregs;
 	int		mc_flags;
 #define	_MC_FP_VALID	0x1		/* Set when mc_fpregs has valid data */
+#define	_MC_ESR_VALID	0x10		/* Set when mc_esr has valid data */
```
zxombie: Why `0x10`? We can use `0x2`

ararslan: No strong reason. I'll make it `0x2`.

sys/arm64/arm64/exec_machdep.c
ararslan: I only have it there because it's where other fields are being set. Where should it be read instead to ensure it's captured accurately?

ararslan: For context, I'll note that I'm fairly well out of my depth here; if I have things that seem silly, it's much more likely due to ignorance than having any deep reason.

Thinking about this more, I think I understand why `sendsig` would be too late to read `esr_el1`. Should `set_mcontext` read `esr_el1` and write it to both `mc_esr` and `tf_esr`? Currently `set_mcontext` is populating `mc_esr` from `tf_esr`.

zxombie: `esr_el1` is read on exception entry, so the copy in the trap frame is correct.

I don't think `set_mcontext` needs to be changed as `tf_esr` as it's a read-only register so doesn't need to be restoired, and `mcr_esr` is the value read from userspace so doesn't need to be changed.

ararslan: Okay, so if I understand correctly, I should update this PR as follows:

- Do not overwrite `tf_esr` with `mc_esr` in `set_mcontext` because `tf_esr` was already populated with the ESR value when the exception occurred. (It looks like [this happens in `exception.S`](https://github.com/freebsd/freebsd-src/blob/99213b3c352cdf568ea7cf5b4dddb1879f16a601/sys/arm64/arm64/exception.S#L74).)
- Do not re-read ESR in `sendsig` because it's both not correct to do so and also `tf_esr` already has the right value, as noted.

Then the changes this PR will be making will just be to add `mc_esr` to `struct __mcontext` and to populate it based on `tf_esr` in `get_mcontext`. Is that correct?

Relatedly, I'm not sure the `_MC_ESR_VALID` mask is useful, as evidenced primarily by the fact that it's being used to update `mc_flags` but never used when querying `mc_flags`. Should I remove it?

zxombie: That looks correct.

If we don't have a flag then we will need to bump `__FreeBSD_version` in `sys/sys/param.h`. This should be a separate change as the value will be different in the `stable/*` branches.

ararslan: I see, so the thought is that retaining `_MC_ESR_VALID` would allow downstream code to guard accesses to `mc_esr` with e.g. `#ifdef _MC_ESR_VALID`, otherwise they would need a (potentially complex, if this change gets backported to other branches) check on `__FreeBSD_version`?

Do you see value in retaining `_MC_ESR_VALID` as a mask and using it to set `mc_flags` accordingly? If not and it's acting more as a "feature flag" of sorts than as an indication of the validity of a field's value, perhaps a name like `MC_HAS_ESR` would be more appropriate than `_MC_ESR_VALID`? Is there prior art for this kind of flag that I could use as an example to ensure naming is consistent? Querying `__FreeBSD_version` is what I've seen most often in downstream code, but that could be more of a reflection of the kind of code I look at than anything else.

Btw, I really appreciate your help and guidance here!

sys/arm64/include/ucontext.h
ararslan: Done.

sys/arm64/arm64/exec_machdep.c
ararslan: I've updated the PR as suggested and left in `_MC_ESR_VALID`. Let me know whether I should change that as described (or otherwise), or remove.

zxombie:
Having thought about it I think we only want to set esr & `_MC_ESR_VALID` when the signal comes from an exception via `trap.c`.

There isn't currently a good way to signal this. It looks like we could add a new flag to `ksiginfo_t.ksi_flags` for this. The current `KSI_TRAP` is set, but is also used by signals when copying between userspace & the kernel fail & in these cases `tf->tf_esr` won't be valid.

ararslan:
In terms of implementation, would this be something like:
```diff
diff --git a/sys/arm64/arm64/trap.c b/sys/arm64/arm64/trap.c
index 3ce0b29f0..05d7739e1 100644
--- a/sys/arm64/arm64/trap.c
+++ b/sys/arm64/arm64/trap.c
@@ -129,6 +129,7 @@ call_trapsignal(struct thread *td, int sig, int code, void *addr, int trapno)
 	ksi.ksi_code = code;
 	ksi.ksi_addr = addr;
 	ksi.ksi_trapno = trapno;
+	ksi.ksi_flags |= KSI_EXC;
 	trapsignal(td, &ksi);
 }

diff --git a/sys/sys/signalvar.h b/sys/sys/signalvar.h
index 8f181b7be..733284c2e 100644
--- a/sys/sys/signalvar.h
+++ a/sys/sys/signalvar.h
@@ -237,6 +237,7 @@ typedef struct ksiginfo {
 #define	KSI_SIGQ	0x08	/* Generated by sigqueue, might ret EAGAIN. */
 #define	KSI_HEAD	0x10	/* Insert into head, not tail. */
 #define	KSI_PTRACE	0x20	/* Generated by ptrace. */
+#define	KSI_EXC		0x40	/* Generated by an exception. */
 #define	KSI_COPYMASK	(KSI_TRAP | KSI_SIGQ | KSI_PTRACE)
 
 #define	KSI_ONQ(ksi)	((ksi)->ksi_sigq != NULL)

```

We would need to propagate the information from `ksi_flags` into `struct thread` somehow in order to access it from `get_mcontext`, right?

ararslan:
@zxombie, would it be reasonable for `get_mcontext` to retrieve the `ksiginfo_t` from the thread's associated process and do something like this?:
```c
ksiginfo_t *ksi = td->td_proc->p_ksi;
if (ksi != NULL && (ksi->ksi_flags & KSI_EXC) != 0) {
    mcp->mc_esr = tf->esr;
    mcp->mc_flags |= _MC_ESR_VALID;
}
```
(Coupled with the changes in my previous comment that introduce `KSI_EXC`.)

ararslan:
@zxombie, do you think you'd be able to take another look at this?

ararslan:
The changes here have been committed, so I will close this PR. Thanks so much for your patient help and guidance, @zxombie!

@kostikbel, I saw your email reply to the commit. In case you hadn't seen it, this is where the changes were implemented; perhaps the discussion here would provide some context, though I'm not sure it addresses your concern.

For reference, this was the email reply to (what is in this PR) commit 5cc3fa098856cbb8de72d1027c838ef734030445, specifically the addition of the `KSI_EXCEPT` flag:

> This bit is MD, its proper place is in struct trapframe.
> That is, trapframe should grow a flags field if it needs some additional
> information about validity of its members.
>
> This is how x86 arches handle its optional content like segment registers,
> segment bases, and FRED fields in uncommitted changes.
>
> I do not see why the MD bit should leak into the ksi.

(I don't know what "MD" means but I'm assuming from context it's "metadata.")

kostikbel:
MD means machine-depended.

IMO the KSI flag should be removed, and the flag in arm64 mcontext used there.  Or you could use a flag in td_pcb if mcontext is not appropriate.

ararslan:
I'm not sure I understand what you mean about the flag in the arm64 mcontext, could you elaborate?

I added the `KSI_EXCEPT` flag to address https://github.com/freebsd/freebsd-src/pull/2053#issuecomment-4025709937, but looking more closely at my changes now that I have a slightly better understanding of the repository layout than when I started, I can see now that the flag is being defined for all architectures but is only used on arm64, making it MD. That's a good point.

If the goal is to identify when a signal is generated by `trap.c` (given that `KSI_TRAP` is apparently insufficiently specific) so that we know whether we need to populate `mc_esr` in `mcontext`, your suggestion would be to include that information (or at least a designation of whether `tf_esr` is valid) in `struct trapframe` via a new flags field?

kostikbel:
Either into the struct trapframe for arm64, or into the arm64 td_pcb.

But more I thinking about it, the more I have some doubts: tf_esr must be saved by each exception entry from the L3, isn't it?  Then why is it conditionally copied into mcontext, and why userspace needs a flag?

ararslan:
I should note that I'm well outside of my depth here—@zxombie, if you're able to chime in to address some of @kostikbel's questions, that would be great. If you'd both prefer email, we can continue there instead. I have more questions than I have answers, unfortunately. 😅

> tf_esr must be saved by each exception entry from the L3, isn't it?

Sorry, what is L3? If you mean the AArch64 exception level EL3, `tf_esr` is populated from EL1.

> why is it conditionally copied into mcontext

Do you mean why are we checking `KSI_EXCEPT` at all rather than unconditionally setting `mc_esr` to `tf_esr` in `get_mcontext`? It's to avoid populating `mc_esr` if we know `tf_esr` is invalid, which, based on Andrew's earlier comment, I'm guessing would maybe be when `sigfastblock_failed` gets called.

> why userspace needs a flag?

If you're referring to `KSI_EXCEPT`, my assumption is that its utility is largely, if not entirely, internal; I don't think it necessarily needs to be available in userspace. I just didn't know how else to address https://github.com/freebsd/freebsd-src/pull/2053#issuecomment-4025709937 and there were no comments on the approach noted in https://github.com/freebsd/freebsd-src/pull/2053#issuecomment-4040838689.

If you're referring to `_MC_ESR_VALID`, I had a similar question in https://github.com/freebsd/freebsd-src/pull/2053#discussion_r2885561023. It's not currently used anywhere, but I don't know whether downstream consumers of `mc_esr` would be expected to check it. If `mc_esr` is ever not valid when the `mcontext` is populated then I could imagine it being useful.

kostikbel:
It should have been EL0, it clearly shows that I think in terms of x86 rings.  Anyway, I mean that ESR is populated in HW to reflect the state of EL0 when an exception was raised, and as such should be simply present in the mcontext.

But again, I am saying that (if you want to sometimes) indicate that ESR is not present for whatever reasons, this should be marked in the MD (machine-dependent) state of the thread, be it td_pcb or td_frame, instead of mus-using the MI (machine-independent) state in ksi.

ararslan:
Okay, that makes sense. Thank you for the input, @kostikbel, I appreciate it. Perhaps @zxombie can opine on how important it is to indicate when ESR is not populated. If it's not, then I think we could amend the changes from this PR to simply add `mc_esr` to `mcontext` and populate it with `tf_esr` in `get_mcontext`, removing all `ksiginfo_t`-related changes as well as the `_MC_ESR_VALID` flag. And if it is important, capture that in `td_pcb` or `td_frame` (which I think would require adding a flags field to `trapframe`?) and remove all `ksiginfo_t` changes. Is that right?

kostikbel:
Yes, this is one of the options that I see as possible.

PR #2061, opened 2026-03-03
pfctl: consistency

PHSpline:
% pfctl -F ethernet
Ethernet rules cleared

% pfctl -s ethernet
pfctl: Unknown show modifier 'ethernet'

pfctl accepts 'ethernet' (or any prefix of it) in the -F flag but accepts only 'ether' (or any prefix of it) in the -s flag, which seems inconsistent.  This change brings the two to parity while remaining backwards compatible.


kprovost:
@concussious You’re a little late, I already merged this ;)

PR #2068, opened 2026-03-09
Change various kernel APIs to work with pointers instead of vm_offset_t

bsdjhb:
In CHERI kernels, pointers are not the same size as addresses, and vm_offset_t represents a virtual address (ptraddr_t). To handle places that use vm_offset_t to hold pointers that can be dereferenced by the CPU, CheriBSD adds a new vm_pointer_t type (uintptr_t). However, using vm_pointer_t can be a bit fragile, as C compilers will silently accept conversions between uintptr_t and ptraddr_t. The resulting errors can only be found at runtime. However, if pointers are stored as a pointer type such as void *, then mismatches between addresses and pointers can be caught at compile time including when compiling on non-CHERI architectures. As such, this series converts various kernel pmap/virtual memory kernel APIs that currently use vm_offset_t to use pointer types (either void * or char *) instead. Normally void * is used, but char * is preferred if the API or structure member is commonly used in pointer arithmetic. In practice, these changes seem to be cleaner as most kernel consumers for many of these APIs had to cast between pointer types and vm_offset_t anyway and now those casts can be removed.

Note, when landing this series, I will add a commit to bump __FreeBSD_version so that kmods in ports can handle these changes, but it's pointless to add that commit until this is actually landed.  This would also _not_ be a MFC candidate.

Also, this seems too unwieldy to review in phabricator, hence reviewing here.


mchoo7:
The justification makes sense to me. What I worry is mixing `void *` and `char *` part. From the PR description and mailing list I get that `char *` is for pointer arithmetic, but it could be confusing for people who don't know the context (e.g. they might wonder rationale behind `char *vaddr`). This could end up like `unsigned long devid` for linux tasklet API where most people pass pointers instead of integers (I guess CHERI linux WG needs to handle this?)

My suggestion is creating a typedef to `char *`. In that case pointer arithmetic works and the type name makes more sense.

sys/amd64/amd64/pmap.c
```diff
@@ -9683,24 +9685,25 @@ pmap_change_prot(void *va, vm_size_t size, vm_prot_t prot)
...
-	error = pmap_change_props_locked((vm_offset_t)va, size, prot, -1,
+	error = pmap_change_props_locked(va, size, prot, -1,
 	    MAPDEV_ASSERTVALID);
 	PMAP_UNLOCK(kernel_pmap);
 	return (error);
 }
 
 static int
-pmap_change_props_locked(vm_offset_t va, vm_size_t size, vm_prot_t prot,
+pmap_change_props_locked(void *addr, vm_size_t size, vm_prot_t prot,
```
kostikbel, line 9695: I am not sure about this change.  pmap_change_props() operates on address/page table, not on the pointer.

sys/amd64/amd64/pmap.c
```diff
@@ -9683,24 +9685,25 @@ pmap_change_prot(void *va, vm_size_t size, vm_prot_t prot)
...
-	vm_offset_t base, offset, tmpva;
+	vm_offset_t base, offset, tmpva, va;
 	vm_paddr_t pa_start, pa_end, pa_end1;
 	pdp_entry_t *pdpe;
 	pd_entry_t *pde, pde_bits, pde_mask;
 	pt_entry_t *pte, pte_bits, pte_mask;
 	int error;
 	bool changed;
 
+	va = (vm_offset_t)addr;
```
kostikbel, line 9706: And the fact that you have to immediately cast the address to vm_offset_t aligns with my doubts.

sys/amd64/include/vmparam.h
```diff
@@ -255,7 +255,7 @@
 	KASSERT(PHYS_IN_DMAP(x),					\
 	    ("physical address %#jx not covered by the DMAP",		\
 	    (uintmax_t)x));						\
-	(x) + kva_layout.dmap_low; })
+	(void *)((x) + kva_layout.dmap_low); })
```
kostikbel: I think this change is fine.  But I suggest to provide a macro, lets call it PHYS_TO_DMAP_ADDR(), which is identical to the old PHYS_TO_DMAP(), which returns the address.  Then PHYS_TO_DMAP() becomes ((void *)PHYS_TO_DMAP_ADDR()).

PHYS_TO_DMA_ADDR() should be used where you immediately cast PHYS_TO_DMAP() result back to address.

sys/amd64/include/vmparam.h
```diff
@@ -246,8 +246,9 @@
  * vt fb startup needs to be reworked.
  */
 #define	PHYS_IN_DMAP(pa)	(dmaplimit == 0 || (pa) < dmaplimit)
-#define	VIRT_IN_DMAP(va)	\
-    ((va) >= kva_layout.dmap_low && (va) < kva_layout.dmap_low + dmaplimit)
+#define	VIRT_IN_DMAP(va)						\
+	((uintptr_t)(va) >= kva_layout.dmap_low &&			\
```
kostikbel: I would suggest to add a local of the uintptr_t type and evaluate and cast the 'va' arg only once.

sys/amd64/amd64/pmap.c
```diff
@@ -4863,8 +4863,7 @@ pmap_release(pmap_t pmap)
 	pmap_pt_page_count_pinit(pmap, -1);
 
 	if (pmap->pm_pmltopu != NULL) {
-		m = PHYS_TO_VM_PAGE(DMAP_TO_PHYS((vm_offset_t)pmap->
-		    pm_pmltopu));
+		m = PHYS_TO_VM_PAGE(DMAP_TO_PHYS(pmap->pm_pmltopu));
```
kostikbel: Might be we need DMAP_TO_VM_PAGE() ?

sys/amd64/amd64/pmap.c
bsdjhb: Yes, this one was more about a tradeoff of casts in callers vs just one here, and yes it is a bit more squirrely.

sys/amd64/include/vmparam.h
bsdjhb: Ok.

sys/amd64/amd64/pmap.c
bsdjhb: Certainly this is a common pattern.  Such a macro could also be MI as it would be spelled the same on all architectures.

bsdjhb:
> The justification makes sense to me. What I worry is mixing `void *` and `char *` part. From the PR description and mailing list I get that `char *` is for pointer arithmetic, but it could be confusing for people who don't know the context (e.g. they might wonder rationale behind `char *vaddr`). This could end up like `unsigned long devid` for linux tasklet API where most people pass pointers instead of integers (I guess CHERI linux WG needs to handle this?)
> 
> My suggestion is creating a typedef to `char *`. In that case pointer arithmetic works and the type name makes more sense.

BSD has a legacy typedef (caddr_t), but we generally avoid that in new code.  Also, I think the only place that might be char * in this series is td_kstack?

mchoo7:
> > The justification makes sense to me. What I worry is mixing `void *` and `char *` part. From the PR description and mailing list I get that `char *` is for pointer arithmetic, but it could be confusing for people who don't know the context (e.g. they might wonder rationale behind `char *vaddr`). This could end up like `unsigned long devid` for linux tasklet API where most people pass pointers instead of integers (I guess CHERI linux WG needs to handle this?)
> > My suggestion is creating a typedef to `char *`. In that case pointer arithmetic works and the type name makes more sense.
> 
> BSD has a legacy typedef (caddr_t), but we generally avoid that in new code.

I would avoid `caddr_t` for parameters, return values, and struct fields but for local variables `caddr_t` seems ok. I don't mind sticking with `char *` but the name `caddr_t` sounds more intuitive.

>Also, I think the only place that might be char * in this series is td_kstack?

I see `char *` in many places like in `struct sync_list`.

But actually, we can address this by adding a note to style(9). e.g.:

> Usage of caddr_t is avoided and void * is preferred. If there is frequent pointer arithmetic, char * is acceptable.

and use `char *` over `caddr_t` everywhere including local variables. Because new contributors are encouraged to start by reading style(9) they'll get the context before they work on the code.

kostikbel (approved):

kostikbel (commented):
This one seems to be rather arbitrary.  The patch adds more casts than it removes.  Also, the kva is internal member which denotes the address which is used for page table manipulation (pmap_qenter()) and not a pointer that is operated upon.  The single cast in sf_buf_kva() seems more proper to me.

jrtc27:
sf_buf_kva's returned value is used as a pointer, and in the non-DMAP + SFBUF case that pointer comes from the kva member. Yes, it's true that all supported CHERI architectures have a DMAP, but we should be consistent with types.

sys/amd64/amd64/pmap.c
bsdjhb: FWIW, in CheriBSD this has to use a capability (not just an address) because the arm64 version of this flushes the CPU cache when switching a mapping to `VM_MEMATTR_UNCACHEABLE` and the CHERI version of that instruction takes a pointer rather than just an address.   But I can defer that part of the change for now.

sys/amd64/amd64/pmap.c
bsdjhb: Though I also found that once I added PHYS_TO_DMAP_ADDR that several of this pattern ended up using that.  Still I will probably take a stab at adding this.

sys/amd64/amd64/pmap.c
bsdjhb: Hmm, also, PHYS_TO_DMAP is passed as the argument to pmap_change_props_locked several times.  By changing the type to void * and doing a single cast here in the caller, the net change in the overall diff is just the void * change and single vm_offset_t cast inside the function.  If I keep it as vm_offset_t, I need to change all of the callers to pmap_change_props_locked in the net diff to add a cast (including 4 recursive calls in pmap_change_props_locked itself, one in pmap_page_set_memattr_noflush, and one in pmap_mapdev_internal).

sys/amd64/amd64/pmap.c
bsdjhb: Hmm, I misread this as `VM_PAGE_TO_DMAP` at first.  (Possibly that would also be handy to have as a shortcut?)

sys/amd64/include/vmparam.h
bsdjhb: Done

bsdjhb:
I can drop the commit for sf_buf.kva's type.  Right now it is just adding noise.  It is true that in CheriBSD we did change it from `vm_offset_t` to `vm_pointer_t`, but I think we strictly speaking did not need to since the `kva` member should be unused.

I do wonder if we could make the `kva` member conditional on `#ifdef SFBUF`?  I could perhaps try that as an alternate commit to the type change for sf_buf.kva?

sys/vm/vm_page.h
```diff
@@ -462,6 +462,7 @@ extern long first_page;			/* first physical page number */
 vm_page_t PHYS_TO_VM_PAGE(vm_paddr_t pa);
 
 #define	DMAP_TO_VM_PAGE(va)	PHYS_TO_VM_PAGE(DMAP_TO_PHYS((va)))
+#define	VM_PAGE_TO_DMAP(entry)	PHYS_TO_DMAP(VM_PAGE_TO_PHYS((entry)))
```
kostikbel: I think that double pair of `(())` is not needed.

kostikbel (approved):

kostikbel (approved):

sys/vm/vm_page.h
bsdjhb: Yeah, that's probably excessive.

PR #2104, opened 2026-03-27
du(1): document --si option

kovan:
The `--si` option (human-readable output with SI units based on powers of 1000) is implemented in the source code but was missing from both the SYNOPSIS and the options description in the man page.

Verified against the `long_options` table and `SI_OPT`/`UNITS_SI` handling in `usr.bin/du/du.c`.

PR: 265199


bsdimp:

PR #2105, opened 2026-03-27
Add support for ARM pl011 UART driver multiplexing to allow using extra UARTs on Raspberry Pi devices.

Brisk4t:
### The raspberry pi 4 (and other Pi boards although untested) support more than the 2 traditional UART0 and UART1 lines however they require the GPIO multiplexer to set the pin modes to their alternate versions. 

This patch adds the multiplexer config to the pl011 driver and adds a sample overlay for UART3 (enabled with dtoverlay=uart3 in config.txt) and so far attaches GPIO 4 and 5 to TX and RX respectively. The process can simply be duplicated for UART4 and UART5 with the correct device tree node addresses.

**However, due to what seems like an incomplete driver initialization (and my lack of oscilloscope to confirm this) I haven't been able to get legible output at those GPIO pins so further testing is needed.**

For UART3 (named UART1 /dev/ttyu1 | /dev/cuau1 in FreeBSD due to attachment order) this is what the node and its address + config looks like after successfully attaching alongside the existing UART0:
 
image


sys/dev/uart/uart_dev_pl011.c
```diff
@@ -417,6 +418,11 @@ uart_pl011_bus_attach(struct uart_softc *sc)
 	psc = (struct uart_pl011_softc *)sc;
 	bas = &sc->sc_bas;
 
+#ifdef FDT
+    if (IS_FDT)
+        fdt_pinctrl_configure(sc->sc_dev, 0);
+#endif
+
```
bsdimp: The pinctl driver should already do this.... Is this really needed? Also the second arg should be "default," but that should have already happened...

Brisk4t: I did notice the bcm_gpio_set_alternate() function in the bcm2835_gpio.c driver but the overlay doesn't trigger the alt mode without calling the fdt_pinctrl_configure() inside the pl011 driver. I wonder if I need something else in the overlay itself then....

PR #2107, opened 2026-03-30
Add MAINTAINERS.ucl tooling

Humanoid-Human:
Adds a `MAINTAINERS.ucl` file, as an implementation of this suggestion: https://github.com/freebsd/srcmgr/issues/21

This UCL file stores a list of groups, each of which watch some paths, and have some members. It also stores a mapping from people's IDs (probably FreeBSD.org usernames) and their other contact info, such as GitHub or Forgejo usernames. Currently the content of this file is only an example.

Also adds a `MAINTAINERS.lua` script in `tools/lua`, which uses flua and can automatically generate `CODEOWNERS` for GitHub and Forgejo, as well as fetch other data.

### DIFF
--- a/MAINTAINERS.ucl
+++ b/MAINTAINERS.ucl
@@ -1,0 +1,477 @@
+/*
+
+List of maintainers and watch paths.
+
+Each maintainer is represented by their FreeBSD.org username, used as a key in the "people" object.
+This may also be a team of maintainers, in which case the name should be prefixed with "#", i.e. "#bhyve".
+Each value in "people" is an object that may contain the user's GitHub username (key "github"),
+and Forgejo/Codeberg username (key "forgejo"), and possibly other information (all optional).
+
+Each group is a key-value pair in the "groups" object.
+Every group contains:
+- key "members": a single member, or a list of members (FreeBSD.org usernames, or team names)
+- key "watch": a single path, or a list of paths to watch
+- key "except": a single path, a list of paths to ignore (optional)
+Paths to directories must end with trailing separators.
+
+The functionality to ignore paths does not work with Forgejo's CODEOWNERS currently.
+
+After editing this file, run
+
+	./tools/lua/MAINTAINERS.lua update <platform> > path/to/CODEOWNERS
+
+for GitHub and/or Forgejo.
+
+*/
+
+version: 0
+
+groups {
+	"aio(4)" {
+		"members": "asomers",
+		"watch": [ ],
+	},
+
+	"allwinner" {
+		"members": "manu",
+		"watch": "sys/arm/allwinner/",
+	},
+
+	"bhyve" {
+		"members": "#bhyve",
+		"watch": [
+			"lib/libvmmapi/",
+			"sys/*/vmm/",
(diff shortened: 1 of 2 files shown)


MAINTAINERS.ucl
```diff
@@ -0,0 +1,523 @@
...
+	"random" {
+		"members": "csprng",
+		"watch": [
+			"sys/dev/random/",
+			"sys/sys/random.h",
+		],
+	},
+
+	"release.sh" {
+		"members": "re",
```
markjdb: Some of these members, like "re" and "csprng" are not committers but refer to groups of committers. I suspect we want a way to define such groups, so that a "member" can either be a committer/developer or a group of them.

MAINTAINERS.ucl
```diff
@@ -0,0 +1,523 @@
...
+	},
+
+	"linuxkpi" {
+		"members": [
+			"bz",
+			"dumbbell",
+			"emaste",
+			"manu",
+			"minsoochoo0122_proton.me",
+			"rcocker87_gmail.com"
```
markjdb, line 197: These are phabricator IDs, but really we probably want a "name <email>" tuple, e.g., "Mark Johnston <markj@freebsd.org>" (which can abbreviated to markj since it's a @freebsd.org account).

I understand it's hard to get that info out of phabricator, this is a general comment.

markjdb (commented):
I wonder if lua+libucl is really a good way to go... it effectively means that one can only use this script on FreeBSD, but a given developer or contributor might be using a different platform.

MAINTAINERS.ucl
Humanoid-Human: Oh sorry, didn't catch those ones.

I was thinking that a group of committers can just be a group (the existing structure, with a list of members and a list of paths), and all of the necessary paths would be added directly to that item, instead of substituting the group as a member into separate watched paths. For example with the existing "bhyve" group.

MAINTAINERS.ucl
Humanoid-Human: Didn't realize those were Phabricator-specific, will fix when I get the chance.

The intent was for the information in each "group" to be as minimal as possible, so just a single ID for each member. Stuff like names could be added to the "members" object, to avoid duplicating the information unnecessarily.

Humanoid-Human:
> I wonder if lua+libucl is really a good way to go... it effectively means that one can only use this script on FreeBSD, but a given developer or contributor might be using a different platform.

libucl can be installed on non-FreeBSD systems, although last I checked that was a little broken. This script should also work with versions of Lua other than flua.

I can probably make the process of running on non-FreeBSD systems a bit simpler by making `MAINTAINERS.ucl` a valid JSON, and then falling back to a JSON parser if libucl doesn't exist on the system.

MAINTAINERS.ucl
Humanoid-Human: Updated to just use group names prefaced with "#". These are treated essentially as regular users. I'm not sure that specific extra functionality for groups is necessary, but I can add it if need be.

PR #2108, opened 2026-03-31
Add capability rights for timerfd, fcntl, fadvise, and jail…

klheard:
… operations

Add new Capsicum capability rights for fine-grained access control:

- CAP_TIMERFD_GETTIME: Permits timerfd_gettime(2) on timer descriptors
- CAP_TIMERFD_SETTIME: Permits timerfd_settime(2) on timer descriptors
- CAP_FCNTL_READAHEAD: Permits fcntl(F_READAHEAD) and fcntl(F_RDAHEAD)
- CAP_POSIX_FADVISE: Permits posix_fadvise(2)
- CAP_JAIL_ATTACH: Permits jail_attach_jd(2) on jail descriptors
- CAP_JAIL_REMOVE: Permits jail_remove_jd(2) on jail descriptors

Real problem solved:
- Allowing processes to join a jail without the ability to remove it's members.
- Allowing the `gettime` without `settime`.
- Blocking FADVISE - Was in comments so I just fixed it.

Includes ATF tests for all new capability rights and updates rights(4).

### DIFF
--- a/RELNOTES
+++ b/RELNOTES
@@ -10,6 +10,11 @@ newline.  Entries should be separated by a newline.
 
 Changes to this file should not be MFCed.
 
+b92b9da33006, 13d3bd165e22:
+	This patch allows vfs.nfsd.srvmaxio to be set to
+	a value as high as 4Mbytes.  It is usually set
+	via the nfs_server_maxio variable in /etc/rc.conf.
+
 4c486fe40267:
 	The multicast routing modules, ip_mroute and ip6_mroute, are now FIB-aware.
 	That is, there is a multicast routing table per FIB, rather than a global


kevans91:
Can you go ahead and add a right for jail_set(2) as well, please? There's value in any combination of these, including none (to just pass something that one can jail_get(2) on).

This should also likely be two commits, at least, with the jaildesc bits being on its own. (edit: the kernel parts otherwise seem fine at a glance, haven't looked at the test since I'm on mobile at the moment).

klheard:
> Can you go ahead and add a right for jail_set(2) as well, please? There's value in any combination of these, including none (to just pass something that one can jail_get(2) on).
> 
> This should also likely be two commits, at least, with the jaildesc bits being on its own. (edit: the kernel parts otherwise seem fine at a glance, haven't looked at the test since I'm on mobile at the moment).

@kevans91 I broke it into 4 commits and added the jailset capability.

tests/sys/kern/jaildesc_cap_test.c
```diff
@@ -0,0 +1,394 @@
...
+static int
+create_jail_with_desc(const char *name)
+{
+	struct iovec iov[8];
+	char path[] = "/";
+	int persist = 1;
+	int desc_fd = -1;
+	int jid;
+
+	iov[0].iov_base = __DECONST(char *, "name");
```
kevans91: I think this sequence and similar should be fine and a little cleaner looking with jail_setv(3)/jail_getv(3)

tests/sys/kern/fcntl_readahead_cap_test.c
```diff
@@ -0,0 +1,194 @@
...
+}
+ATF_TC_CLEANUP(cap_fcntl_readahead_allowed, tc)
+{
+	unlink("testfile");
+}
+
+/*
+ * Test that missing CAP_FCNTL_READAHEAD denies fcntl(F_READAHEAD).
+ */
+ATF_TC_WITH_CLEANUP(cap_fcntl_readahead_denied);
```
kevans91: In the interest of avoiding too much additional test runtime, you might consider consolidating the _allowed and _denied tests into one for all of these since you can always cap_rights_limit() with an empty or lesser set after testing that the right being granted was successful.

tests/sys/kern/fcntl_readahead_cap_test.c
```diff
@@ -0,0 +1,194 @@
...
+	error = fcntl(fd_limited, F_READAHEAD, 0);
+	ATF_REQUIRE_MSG(error != -1 || errno != ENOTCAPABLE,
+	    "fcntl(F_READAHEAD) should succeed, got %s", strerror(errno));
+
+	close(fd_limited);
+	close(fd);
+}
+ATF_TC_CLEANUP(cap_fcntl_readahead_allowed, tc)
+{
+	unlink("testfile");
```
kevans91: You can also omit cleanup handlers that are just removing files from the testdir -- the standard expectation is that these are run under kyua(1), which will tear it all down and avoid leaving droplets.  You'd mainly need it for, e.g., the jail tests (or anything outside of cwd) since Kyua can't clean those things up.

kevans91 (commented):
This seems more or less fine, just a few suggestions to reduce overhead in the tests  mostly.

kevans91:
I've also floated a pointer to this review to Jamie for another jaildesc-perspective, but the default behavior doesn't change (since all rights are granted by default) and I'd suspect nobody is applying rights(4) to a jaildesc yet to notice that they suddenly can't do these things.

tests/sys/kern/fcntl_readahead_cap_test.c
klheard: I've incorporated this feedback.

tests/sys/kern/fcntl_readahead_cap_test.c
klheard: I've incorporated this feedback.

tests/sys/kern/jaildesc_cap_test.c
klheard: I've incorporated this feedback.

spmzt:
LGTM, but let's wait for @kevans91 feedback.

sys/kern/kern_jail.c
```diff
@@ -1118,7 +1119,8 @@ kern_jail_set(struct thread *td, struct uio *optuio, int flags)
 			 * descriptor's prison.
 			 */
 			prison_free(mypr);
-			error = jaildesc_find(td, jfd_in, &mypr, NULL);
+			error = jaildesc_find(td, jfd_in, &cap_no_rights,
```
kevans91: I think it deserves a comment noting that it's correct to not require jail_set(2) rights here; notably, that using a descriptor as a search base isn't a privileged operation.

sys/kern/kern_jail.c
```diff
@@ -1539,7 +1541,7 @@ kern_jail_set(struct thread *td, struct uio *optuio, int flags)
 	}
 	if (flags & JAIL_USE_DESC) {
 		/* Get the jail from its descriptor. */
-		error = jaildesc_find(td, jfd_in, &pr, &jdcred);
+		error = jaildesc_find(td, jfd_in, &cap_jail_set_rights, &pr, &jdcred);
```
kevans91: Jamie notes that this isn't quite complete- a single jail_set(2) operation can also attach with the `JAIL_ATTACH` flag.  Barring a KPI to do an additional rights checks on a file (which wouldn't be hard to implement, but I'm not sure why it doesn't seeem to exist), I guess we want to use another set that includes both `CAP_JAIL_SET + CAP_JAIL_ATTACH` for the case where `JAIL_ATTACH` was also specified.

We probably also want a different error string for ENOTCAPABLE, I reckon

gritton:
I've started similar work (just the jail part) at [FreeBSD Differential D52516](https://reviews.freebsd.org/D52516).  The main thing stopping it from moving forward has been the request for test cases; somehow that made it less of a priority :-/.  A few notes from that:

- It has CAP_JAIL_GET.  I noticed there are plenty of capabilities in other domains for reading as well as writing things, so it seemed appropriate. 
- It also looks at using jail descriptors to look up other jails.  This requires a parallel set of permissions, like CAP_JAIL_REMOVEAT in addition to CAP_JAIL_REMOVE, and notably adds CAP_JAIL_CREATEAT for creating a new sub-jail of the one you hold a descriptor for.  It would make sense to add all that at a later time, to keep things manageable, since it does add to the complexity quite a bit.
- If you want a capability-mode process to be able to query the jail parameter MIB, the parameter sysctls need to have CTLFLAG_CAPRW set.  I made settable per-parameter, to you could have parameters that aren't settable via capability mode.  But I don't know if that's actually good and/or necessary, and it may be better to have it be a default part of the SYSCTL_JAIL_PARAM define.

tests/sys/kern/Makefile
```diff
@@ -22,6 +24,7 @@ ATF_TESTS_C+=	fdgrowtable_test
 ATF_TESTS_C+=	getdirentries_test
 ATF_TESTS_C+=	jail_lookup_root
 ATF_TESTS_C+=	inotify_test
+ATF_TESTS_C+=	jaildesc_cap_test
```
oshogbo: I think this test should be a part of the tests/sys/capsicum.

spmzt:
Hi @klheard,
just a reminder to re-request review when you're done updating your PR.

runxiyu:
Cleaned up patch at https://lists.freebsd.org/archives/freebsd-hackers/2026-July/006546.html, only for jaildesc. I'll try to submit it to phabricator later.

cc @kevans91

runxiyu:
https://reviews.freebsd.org/D58082

PR #2116, opened 2026-04-03
virtio_balloon: implement STATS_VQ and DEFLATE_ON_OOM support

xz-dev:
## Summary

Implement two virtio balloon features that have been missing from the FreeBSD driver since its introduction in FreeBSD 9.0:

- **VIRTIO_BALLOON_F_STATS_VQ**: Report guest memory statistics (free, total, available, swap, faults, caches) to the hypervisor via the stats virtqueue.
- **VIRTIO_BALLOON_F_DEFLATE_ON_OOM**: Automatically deflate the balloon when the guest is under memory pressure, returning pages to the VM subsystem before processes are killed.

Also update the local virtio balloon definitions needed for the implemented feature bits and stat tags.

The memory statistics intentionally stay within existing FreeBSD VM counters. `VIRTIO_BALLOON_S_AVAIL` is a conservative estimate based on free and inactive pages above VM targets, and `VIRTIO_BALLOON_S_CACHES` uses inactive pages as the closest existing VM-queue approximation. This PR does not add ZFS ARC or other subsystem-specific reclaimable-cache accounting.

## Motivation

FreeBSD guests running under QEMU/KVM cannot report memory statistics to the hypervisor. This means `info balloon` / `virsh dommemstat` only shows `actual` and `max_mem` with no `total_mem`, `free_mem`, or other stats. Platforms like Proxmox that rely on these stats to display guest memory usage and make ballooning decisions see incorrect values for FreeBSD guests.

Without DEFLATE_ON_OOM, an over-inflated balloon cannot automatically shrink under memory pressure, risking process kills while reclaimable balloon pages exist.

## References

- [QEMU: Virtio balloon memory statistics](https://www.qemu.org/docs/master/interop/virtio-balloon-stats.html)
- [OASIS virtio-v1.2 specification §5.5](https://docs.oasis-open.org/virtio/virtio/v1.2/cs01/virtio-v1.2-cs01.html)

## Link bugs
- https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=292570

### DIFF
--- a/sys/dev/virtio/balloon/virtio_balloon.c
+++ b/sys/dev/virtio/balloon/virtio_balloon.c
@@ -127,6 +127,8 @@ static int	vtballoon_sleep(struct vtballoon_softc *);
 static void	vtballoon_thread(void *);
 static void	vtballoon_guest_lowmem(void *, int);
 static void	vtballoon_setup_sysctl(struct vtballoon_softc *);
+static uint64_t	vtballoon_mem_available(struct vtballoon_softc *);
+static uint64_t	vtballoon_mem_caches(void);
 
 #define vtballoon_modern(_sc) \
     (((_sc)->vtballoon_features & VIRTIO_F_VERSION_1) != 0)
@@ -713,6 +715,35 @@ vtballoon_thread(void *xsc)
 	kthread_exit();
 }
 
+/*
+ * Estimate FreeBSD VM headroom without dipping below the page daemon's free
+ * and inactive targets.  This intentionally uses only existing VM counters;
+ * subsystem-specific reclaimable kernel caches such as ZFS ARC are outside
+ * the scope of this virtio driver estimate.
+ */
+static uint64_t
+vtballoon_mem_available(struct vtballoon_softc *sc __unused)
+{
+	int64_t available;
+	int64_t inactive;
+
+	available = ((int64_t)vm_free_count() - vm_cnt.v_free_target) * PAGE_SIZE;
+	inactive = ((int64_t)vm_inactive_count() - vm_cnt.v_inactive_target) *
+	    PAGE_SIZE;
+	if (inactive > 0)
+		available += inactive;
+
+	if (available < 0)
+		available = 0;
+	return ((uint64_t)available);
+}
+
+static uint64_t
+vtballoon_mem_caches(void)
+{
+
+	return ((uint64_t)vm_inactive_count() * PAGE_SIZE);
+}
 
 static int
 vtballoon_update_stats(struct vtballoon_softc *sc)
@@ -736,7 +767,7 @@ vtballoon_update_stats(struct vtballoon_softc *sc)
 	VTBALLOON_SET_STAT(VIRTIO_BALLOON_S_MEMFREE,
 	    (uint64_t)vm_free_count() * PAGE_SIZE);
 	VTBALLOON_SET_STAT(VIRTIO_BALLOON_S_AVAIL,
-	    ((uint64_t)vm_free_count() + vm_inactive_count()) * PAGE_SIZE);
+	    vtballoon_mem_available(sc));
 	VTBALLOON_SET_STAT(VIRTIO_BALLOON_S_SWAP_IN,
 	    VM_CNT_FETCH(v_swappgsin) * PAGE_SIZE);
 	VTBALLOON_SET_STAT(VIRTIO_BALLOON_S_SWAP_OUT,
@@ -746,14 +777,12 @@ vtballoon_update_stats(struct vtballoon_softc *sc)
 	VTBALLOON_SET_STAT(VIRTIO_BALLOON_S_MAJFLT,
 	    VM_CNT_FETCH(v_io_faults));
 	/*
-	 * Map CACHES to the inactive page queue.  The virtio spec defines
-	 * this as "disk caches" (Linux: page cache + reclaimable slab).
-	 * FreeBSD's inactive queue is the closest approximation -- it
-	 * contains pages eligible for reclaim, both file-backed and
-	 * anonymous, similar in spirit to Linux's reclaimable memory.
+	 * FreeBSD has no single Linux-style "disk caches" counter available
+	 * here.  Use inactive pages as the closest existing VM queue
+	 * approximation and avoid subsystem-specific cache accounting.
 	 */
 	VTBALLOON_SET_STAT(VIRTIO_BALLOON_S_CACHES,
-	    (uint64_t)vm_inactive_count() * PAGE_SIZE);
+	    vtballoon_mem_caches());
 
 #undef VTBALLOON_SET_STAT
 
--- a/sys/dev/virtio/balloon/virtio_balloon.c
+++ b/sys/dev/virtio/balloon/virtio_balloon.c
@@ -41,8 +41,11 @@
 #include <sys/mutex.h>
 #include <sys/queue.h>
 
+#include <sys/vmmeter.h>
+
 #include <vm/vm.h>
 #include <vm/vm_page.h>
+#include <vm/vm_extern.h>
 
 #include <machine/bus.h>
 #include <machine/resource.h>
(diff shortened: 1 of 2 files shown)


xz-dev:
I feel no bugs now, let's see how to submit the PR to [FreeBSD technical discussions mailing list](https://lists.freebsd.org/subscription/freebsd-hackers)

xz-dev:
https://lists.freebsd.org/archives/freebsd-virtualization/2026-April/003607.html

sys/dev/virtio/balloon/virtio_balloon.c
```diff
@@ -565,6 +608,87 @@ vtballoon_thread(void *xsc)
...
+	stat[idx].val = modern ? htole64(val_) : (val_);	\
+	idx++;							\
+} while (0)
+
+	VTBALLOON_SET_STAT(VIRTIO_BALLOON_S_MEMTOT,
+	    (uint64_t)vm_cnt.v_page_count * PAGE_SIZE);
+	VTBALLOON_SET_STAT(VIRTIO_BALLOON_S_MEMFREE,
+	    (uint64_t)vm_free_count() * PAGE_SIZE);
+	VTBALLOON_SET_STAT(VIRTIO_BALLOON_S_AVAIL,
+	    ((uint64_t)vm_free_count() + vm_inactive_count()) * PAGE_SIZE);
```
kostikbel: I have no idea about the virtio code.  That said, the counters returned in the patch indeed make the 1:1 mapping between names, and I think that this might be problematic.  The consumers of these statistic on the other side of virtio perhaps expect Linux-like meaning for the fields, while driver provides FreeBSD semantic.

For instance, free pages on FreeBSD are pages without valid content, and on typical machine it would be (almost) zero.  BALOON_S_AVAIL seems to mean 'something that ballon could squeeze out, why don't you count e.g. active pages as well?  On mem pressure, active is moving to laundry/inactive.

I do not object against having this as the initial version, it is definitely much better than not having it at all, but I expect that the report would need some tweaking to get the behavior aligned with expectation of consumers.

sys/dev/virtio/balloon/virtio_balloon.c
```diff
@@ -580,18 +609,57 @@ vtballoon_sleep(struct vtballoon_softc *sc)
...
+	int lowmem;
 
 	sc = xsc;
 
 	for (;;) {
 		if (vtballoon_sleep(sc) != 0)
 			break;
 
+		/* Check and clear the low memory flag. */
+		VTBALLOON_LOCK(sc);
```
kostikbel, line 657: I am confused by which machine low mem situation this refers to.  Is it about host low memory requesting guest to shrink, or guest low mem needed some resources from host?  Can it be at least explained in the comment and naming?

Also, why is this located in the same commit as the statistic?

xz-dev: Thanks for review. This is guest-side low memory. I renamed the path to use guest_lowmem consistently and added comments explaining that vm_lowmem is raised by the FreeBSD guest VM subsystem, not by the host asking the guest to reclaim memory. I also split the STATS_VQ and DEFLATE_ON_OOM implementations into separate commits; the remaining shared commit only adds missing virtio spec definitions.

sys/dev/virtio/balloon/virtio_balloon.c
xz-dev: Thanks, I agree with this concern. A direct 1:1 mapping from FreeBSD VM counters to virtio/Linux-like fields is not ideal.

I updated `VIRTIO_BALLOON_S_AVAIL` to avoid reporting simply `free + inactive`. It is now a FreeBSD VM headroom estimate:
```
    free + inactive - free_target - inactive_target
```
The intent is to report memory that can be consumed without pushing the guest below the page daemon's normal free/inactive targets.

I intentionally did not count active pages for now. While active pages can migrate to inactive/laundry under pressure, they represent the guest's current working set. Also, FreeBSD's exported active queue count is not split into file-backed and anonymous pages. Without adding new VM accounting, or a VM-level helper with documented semantics, the balloon driver cannot tell how much of the active queue is reclaimable file cache versus anonymous/process working set. Counting all active pages would therefore risk reporting anonymous working-set memory as host-available memory.

spmzt:
Hi, @xz-dev
just a reminder to re-request review when you're done updating your PR.

xz-dev:
> Hi, @xz-dev just a reminder to re-request review when you're done updating your PR.

Thank you your reminder :) , I want to test more details before that.

Because it's the first time I've submitted a patch at the kernel level.

xz-dev:
This PR has been open for a while, so I force-pushed a cleaned-up stack and narrowed the scope to the virtio balloon driver itself.

The current stack is split into four commits:

1. implement `VIRTIO_BALLOON_F_STATS_VQ`;
2. implement `VIRTIO_BALLOON_F_DEFLATE_ON_OOM` for guest-side `vm_lowmem` events;
3. harden stats / resize wakeups and virtqueue completion handling;
4. use conservative FreeBSD memory statistics for `S_AVAIL` and `S_CACHES`.

The final diff is now limited to:

- `sys/dev/virtio/balloon/virtio_balloon.c`
- `sys/dev/virtio/balloon/virtio_balloon.h`

I removed the earlier experimental ZFS ARC, VFS buffer-cache, and generic reclaimable-cache accounting work from this PR. The current `VIRTIO_BALLOON_S_AVAIL` value intentionally uses only existing FreeBSD VM counters: free pages above `v_free_target` plus inactive pages above `v_inactive_target`, clamped at zero. `VIRTIO_BALLOON_S_CACHES` uses inactive pages as the closest existing VM-queue approximation for a Linux-style disk-cache statistic.

Validation performed locally with an amd64 FreeBSD guest under QEMU/libvirt:

- `git diff --check`
- focused `sys/modules/virtio/balloon` build
- `GENERIC` `buildkernel`
- runtime stats testing through QEMU/libvirt (`stat-available-memory` / `virsh dommemstat usable`, `stat-disk-caches` / `disk_caches`)
- live balloon resize testing
- `DEFLATE_ON_OOM` / guest-low-memory deflation testing
- a short repeated pressure/resize soak test

There are still two FreeBSD memory-accounting questions that I intentionally did not try to solve in this PR, but I would appreciate guidance on the right direction:

1. Should FreeBSD eventually provide a VM-level equivalent of Linux `si_mem_available()` for kernel consumers such as virtio-balloon, instead of each driver approximating `S_AVAIL` from raw VM counters?
2. Should reclaimable subsystem caches such as ZFS ARC be included in a future `S_AVAIL` / `S_CACHES` implementation? I experimented locally with a small reclaimable-cache provider hook and an ARC provider, but removed it from this PR because it seemed broader than virtio-balloon itself.
3. How should dirty / laundry pages be treated for this purpose? They may become reclaimable after writeback, but they are not immediately available without additional work, so I excluded them from the conservative estimate here.

My intent is for this PR to provide the missing virtio features and a conservative initial stats implementation, while leaving any broader FreeBSD-wide memory-availability accounting to a separate discussion or follow-up change if that is preferred.

sys/dev/virtio/balloon/virtio_balloon.c
```diff
@@ -713,6 +715,35 @@ vtballoon_thread(void *xsc)
...
+ * subsystem-specific reclaimable kernel caches such as ZFS ARC are outside
+ * the scope of this virtio driver estimate.
+ */
+static uint64_t
+vtballoon_mem_available(struct vtballoon_softc *sc __unused)
+{
+	int64_t available;
+	int64_t inactive;
+
+	available = ((int64_t)vm_free_count() - vm_cnt.v_free_target) * PAGE_SIZE;
```
kostikbel: Perhaps use ptoa() instead of multiplying by PAGE_SIZE.  hy do you cast to int64? It probably better to use u_long there.

kostikbel (approved):

sys/dev/virtio/balloon/virtio_balloon.c
```diff
@@ -565,6 +715,149 @@ vtballoon_thread(void *xsc)
...
+	target = vm_cnt.v_free_target;
+	if (count > target)
+		available += count - target;
+
+	count = vm_inactive_count();
+	target = vm_cnt.v_inactive_target;
+	if (count > target)
+		available += count - target;
+
+	return (ptoa((uint64_t)available));
```
kostikbel, line 742: BTW, there is still uint64_t cast.  Why is it needed?

PR #2123, opened 2026-04-05
UPDATING: Remove entries older than stable/13 branch

BigSneakyDuck:
Following "After branch N is created, entries older than the N-2 branch point are removed from this file", remove entries prior to stable/13 branch point since stable/15 exists.

The "only contains entries since" date was 10 years out of date. It's easily missed, so instead describe as "since N-2 branch point".

Also update copyright year and correct date of stable/14 branch to match https://github.com/freebsd/freebsd-src/commit/3efecb71a5eded675f4d0511ebcebf7d4ef98a25 and https://www.freebsd.org/releases/14.0R/schedule/

Have emailed Warner to look at this.


BigSneakyDuck:
I have put a "Submitted by" line in the commit message, but deliberately haven't provided a "Signed-off-by" line since no committer has signed this off yet. I can't see that I would be authorized to sign anything off...

This is my first PR to freebsd-src so let me know if I'm doing this wrong!

cperciva:
Signed-off-by is distinct from Approved-by, so I wouldn't worry about that.  In particular it covers licensing permission.

This sounds reasonable to me but I think it's a srcmgr call to make.

PR #2127, opened 2026-04-07
pci: fix integer overflow in capability parsing

gne:
Change eecp from uint8_t to uint32_t. eecp holds byte offsets for hardware capabilities, which may exceed the range of uint8_t. This overflow causes the offset to wrap, leading to reads from the wrong location and potentially an infinite loop.

Discovered due to a hung boot on AMD USB Controller (PCI 1022:151e) in a Minisforum N5 Pro.

### DIFF
--- a/sys/dev/pci/pci.c
+++ b/sys/dev/pci/pci.c
@@ -3756,7 +3756,7 @@ xhci_early_takeover(device_t self)
 	struct resource *res;
 	uint32_t cparams;
 	uint32_t eec;
-	uint8_t eecp;
+	uint32_t eecp;
 	uint8_t bios_sem;
 	uint8_t offs;
 	int rid;


sys/dev/pci/pci.c
gmshake, line 3717: The macro `EHCI_HCC_EECP` is defined as 
```
#define	EHCI_HCC_EECP(x)	(((x) >> 8) & 0xff)	/* extended ports caps */
```

so `uint8_t` is sufficient to store it. 

While `pci_read_config()` is defined as,
```
static __inline uint32_t
pci_read_config(device_t dev, int reg, int width)
{
    return PCI_READ_CONFIG(device_get_parent(dev), dev, reg, width);
}
```

then `eecp + EHCI_LEGSUP_BIOS_SEM` the `eecp` will be promoted to int automatically.

I do not see a possible integer overflow, am I missing anything ?

gne: You're absolutely correct, I wrongly assumed EHCI_HCC_EECP also was 16 bits. I'll revert both `ehci_early_takeover` and `ehci_pci_take_controller`, and keep only the `xhci_early_takeover` edit.

gne: The mentioned changes were reverted. Do you have any thoughts on the remaining change that fixes the actual bug?

sys/dev/pci/pci.c
```diff
@@ -3756,7 +3756,7 @@ xhci_early_takeover(device_t self)
 	struct resource *res;
 	uint32_t cparams;
 	uint32_t eec;
-	uint8_t eecp;
+	uint32_t eecp;
```
gmshake, line 3759: The previous `uint8_t eecp` looks much like a copy paste error, compared to xhci_pci_take_controller() in `sys/dev/usb/controller/xhci_pci.c` .

gmshake (commented):
I'm not an expert on this. Add @aokblast to confirm.

sys/dev/pci/pci.c
aokblast: 16 bytes is enough, I think.

gne: I think 16 bits would be too small:
- `eecp = XHCI_HCS0_XECP(cparams) << 2`: Here we essentially get a 16-bit int from XHCI_HCS0_XECP and multiply it by 4
- even without the bit-shift, at line 3776: `eecp += XHCI_XECP_NEXT(eec) << 2` could also potentially cause an overflow

Note that the suggested change (`uint32_t`) is exactly what is done in the similar loops `xhci_pci_take_controller` and `xhci_get_xecp`.

aokblast (approved):
You are right. LGTM now! Thanks

PR #2129, opened 2026-04-07
jng: fix mislabeled stats output

kfv:


spmzt (approved):
LGTM, let's wait for Devin input too.

freebsdfrau:
I am installing this change to my stack now and validating it (yes, I know it looks simple, and quite honestly, I do trust the change is good) however, I am still compelled to give it a run in production testing, making sure that downstream stats are good

freebsdfrau:
As a side note -- from merging this into my production stack (which is running jng6 now) -- I had already fixed this upstream, but as I am slow-walking this code into the tree, this still applies to jng2 (recently released).

TL;DR I'll be merging this change, and crediting kfv for the find/fix. Excellent find. It will ultimately reduce the diff on upcoming jng3/jng4/jng5/jng6 that will be dropping in sequence

PR #2131, opened 2026-04-07
hwpmc: Implement IBS process sampling

mashtizadeh:
This change mades process sampling work with IBS by ensuring that read/write only gets or sets the current counter.

Sponsored by: Netflix

### DIFF
--- a/sys/dev/hwpmc/hwpmc_ibs.c
+++ b/sys/dev/hwpmc/hwpmc_ibs.c
@@ -82,13 +82,13 @@ ibs_read_pmc(int cpu, int ri, struct pmc *pm, pmc_value_t *v)
 	KASSERT(ibs_pcpu[cpu],
 	    ("[ibs,%d] null per-cpu, cpu %d", __LINE__, cpu));
 
-	/* read the IBS ctl */
+	/* read the IBS count */
 	switch (ri) {
 	case IBS_PMC_FETCH:
-		*v = rdmsr(IBS_FETCH_CTL);
+		*v = IBS_FETCH_CTL_TO_COUNT(rdmsr(IBS_FETCH_CTL));
 		break;
 	case IBS_PMC_OP:
-		*v = rdmsr(IBS_OP_CTL);
+		*v = IBS_OP_CTL_TO_COUNT(rdmsr(IBS_OP_CTL));
 		break;
 	}
 
@@ -103,11 +103,30 @@ ibs_read_pmc(int cpu, int ri, struct pmc *pm, pmc_value_t *v)
 static int
 ibs_write_pmc(int cpu, int ri, struct pmc *pm, pmc_value_t v)
 {
+	pmc_value_t m;
 
 	KASSERT(cpu >= 0 && cpu < pmc_cpu_max(),
 	    ("[ibs,%d] illegal CPU value %d", __LINE__, cpu));
 	KASSERT(ri >= 0 && ri < IBS_NPMCS,
 	    ("[ibs,%d] illegal row-index %d", __LINE__, ri));
+
+	/* write the IBS count */
+	switch (ri) {
+	case IBS_PMC_FETCH:
+		m = rdmsr(IBS_FETCH_CTL) & ~IBS_FETCH_CTL_CURCNTMASK;
+		/* Setting a count greater than interval is undefined. */
+		if (IBS_FETCH_CTL_TO_INTERVAL(m) > v)
+			m |= IBS_FETCH_COUNT_TO_CTL(v);
+		wrmsr(IBS_FETCH_CTL, m);
+		break;
+	case IBS_PMC_OP:
+		m = rdmsr(IBS_OP_CTL) & ~IBS_OP_CTL_CURCNTMASK;
+		/* Setting a count greater than interval is undefined */
+		if (IBS_OP_CTL_TO_INTERVAL(m) > v)
+			m |= IBS_OP_COUNT_TO_CTL(v);
+		wrmsr(IBS_OP_CTL, m);
+		break;
+	}
 
 	PMCDBG3(MDP, WRI, 1, "ibs-write cpu=%d ri=%d v=%jx", cpu, ri, v);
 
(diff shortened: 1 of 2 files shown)


mashtizadeh:
@bsdimp @mhorne @afscoelho @aalok-agarwal

mashtizadeh:
Diff ready except when I pulled it apart from the others I left out the CURMASK I'll push that #define tomorrow.  Retesting the entire stack after applying it to a clean tree in order.

sys/dev/hwpmc/hwpmc_ibs.c
```diff
@@ -103,12 +103,33 @@ ibs_read_pmc(int cpu, int ri, struct pmc *pm, pmc_value_t *v)
...
+		break;
+	}
+	case IBS_PMC_OP: {
+		m = rdmsr(IBS_OP_CTL) & ~IBS_OP_CTL_CURCNTMASK;
+		/* Setting a count greater than interval is undefined */
+		if (IBS_OP_CTL_TO_INTERVAL(m) > v)
+			m |= IBS_OP_COUNT_TO_CTL(v);
+		wrmsr(IBS_OP_CTL, m);
+		break;
+	}
```
mhorne:
```suggestion
	case IBS_PMC_FETCH:
		m = rdmsr(IBS_FETCH_CTL) & ~IBS_FETCH_CTL_CURCNTMASK;
		/* Setting a count greater than interval is undefined. */
		if (IBS_FETCH_CTL_TO_INTERVAL(m) > v)
			m |= IBS_FETCH_COUNT_TO_CTL(v);
		wrmsr(IBS_FETCH_CTL, m);
		break;
	case IBS_PMC_OP:
		m = rdmsr(IBS_OP_CTL) & ~IBS_OP_CTL_CURCNTMASK;
		/* Setting a count greater than interval is undefined */
		if (IBS_OP_CTL_TO_INTERVAL(m) > v)
			m |= IBS_OP_COUNT_TO_CTL(v);
		wrmsr(IBS_OP_CTL, m);
		break;
```
The braces seem superfluous here.

sys/dev/hwpmc/hwpmc_ibs.h
```diff
@@ -92,9 +92,13 @@
 #define IBS_FETCH_CTL_VALID		(1ULL << 49) /* Valid */
 #define IBS_FETCH_CTL_ENABLE		(1ULL << 48) /* Enable */
 #define IBS_FETCH_CTL_MAXCNTMASK	0x0000FFFFULL
+#define IBS_FETCH_CTL_CURCNTMASK	0xFFFF0000ULL
 
 #define IBS_FETCH_INTERVAL_TO_CTL(_c)	(((_c) >> 4) & 0x0000FFFF)
+#define IBS_FETCH_CTL_TO_INTERVAL(_c)	(((_c) & IBS_FETCH_CTL_MAXCNTMASK) << 4)
 #define IBS_FETCH_CTL_TO_LAT(_c)	(((_c) >> 32) & 0x0000FFFF)
+#define IBS_FETCH_COUNT_TO_CTL(_c)	(((_c) << 12) & IBS_FETCH_CTL_CURCNTMASK)
```
mhorne, line 100: Am I wrong or is there a discrepancy between the shift value of 12 and `IBS_FETCH_CTL_CURCNTMASK`?

0xFFFF0000ULL is a mask covering bits 16-31.

Or is it that some low bits of the count are knowingly discarded?

sys/dev/hwpmc/hwpmc_ibs.c
```diff
@@ -103,12 +103,33 @@ ibs_read_pmc(int cpu, int ri, struct pmc *pm, pmc_value_t *v)
...
 
+	/* write the IBS count */
+	switch (ri) {
+	case IBS_PMC_FETCH: {
+		m = rdmsr(IBS_FETCH_CTL) & ~IBS_FETCH_CTL_CURCNTMASK;
+		/* Setting a count greater than interval is undefined. */
+		if (IBS_FETCH_CTL_TO_INTERVAL(m) > v)
+			m |= IBS_FETCH_COUNT_TO_CTL(v);
+		wrmsr(IBS_FETCH_CTL, m);
+		break;
```
mhorne: I did have to study the block a bit... I think it's good.

It looks like this provides a different/more sophisticated register interface than a lot of other CPU counter registers. The 'interval' value seems to be established at allocation time, from libpmc. This is fine, but was not obvious to me.

sys/dev/hwpmc/hwpmc_ibs.c
```diff
@@ -103,12 +103,33 @@ ibs_read_pmc(int cpu, int ri, struct pmc *pm, pmc_value_t *v)
...
+	}
+	case IBS_PMC_OP: {
+		m = rdmsr(IBS_OP_CTL) & ~IBS_OP_CTL_CURCNTMASK;
+		/* Setting a count greater than interval is undefined */
+		if (IBS_OP_CTL_TO_INTERVAL(m) > v)
```
mhorne, line 125: So in this case where `v` is unexpectedly large, `count` just gets reset to zero, and we get another sample after `interval` ? Note that you can return an error from this function, but I think this behavior is fine in this strange case.

sys/dev/hwpmc/hwpmc_ibs.c
mhorne, line 198: An aside: I did not clue in to the "sampling only" assumption until I saw it checked in libpmc's `ibs_allocate_pmc()`. The verification is more important here on the driver side.

```suggestion

if (!PMC_IS_SAMPLING_MODE(a->pm_mode))
    return (EINVAL);
```

sys/dev/hwpmc/hwpmc_ibs.h
mashtizadeh: Yes, the bottom four bits are zero'ed out the counter corresponds to bits 4:20

sys/dev/hwpmc/hwpmc_ibs.c
mashtizadeh: Yes, I got weird glitches because hwpmc_mod isn't really wanting to assume an arbitrary value.  I also hit the corner case where the counter overflows and stops, which is what lines 118-119 prevent.

sys/dev/hwpmc/hwpmc_ibs.c
mashtizadeh: I think because of the randomization of the lower bits that are hidden you don't want them to be equal even.

sys/dev/hwpmc/hwpmc_ibs.c
mashtizadeh: I always thought it helps readability when it looks long enough I'll fix it

sys/dev/hwpmc/hwpmc_ibs.c
mhorne: I will trust your testing here that it behaves how it is meant to.

mhorne (approved):

PR #2133, opened 2026-04-08
pmc: Implement the feature bits for recent Zen 4/5

mashtizadeh:
This change ensures that the optional MSRs and the user flags that are guarded by cpuid feature flags are properly enforced.  On older AMD processors, certain feature flags may result in a kernel crash.

Sponsored by: Netflix

### DIFF
--- a/lib/libpmc/libpmc.c
+++ b/lib/libpmc/libpmc.c
@@ -31,6 +31,9 @@
 #include <sys/module.h>
 #include <sys/pmc.h>
 #include <sys/syscall.h>
+#if defined(__amd64__) || defined(__i386__)
+#include <machine/cpufunc.h>
+#endif
 
 #include <ctype.h>
 #include <errno.h>
@@ -697,6 +700,8 @@ ibs_allocate_pmc(enum pmc_event pe, char *ctrspec,
 {
 	char *e, *p, *q;
 	uint64_t ctl, ldlat;
+	u_int ibs_features;
+	u_int regs[4];
 
 	pmc_config->pm_caps |=
 	    (PMC_CAP_SYSTEM | PMC_CAP_EDGE | PMC_CAP_PRECISE);
@@ -719,11 +724,22 @@ ibs_allocate_pmc(enum pmc_event pe, char *ctrspec,
 		return (-1);
 	}
 
+	/* Read the ibs feature flags */
+	ibs_features = 0;
+	do_cpuid(0x80000000, regs);
+	if (regs[0] >= CPUID_IBSID) {
+		do_cpuid(CPUID_IBSID, regs);
+		ibs_features = regs[0];
+	}
+
 	/* parse parameters */
 	ctl = 0;
 	if (pe == PMC_EV_IBS_FETCH) {
 		while ((p = strsep(&ctrspec, ",")) != NULL) {
 			if (KWMATCH(p, "l3miss")) {
+				if ((ibs_features & CPUID_IBSID_ZEN4IBSEXTENSIONS) == 0)
+					return (-1);
+
 				ctl |= IBS_FETCH_CTL_L3MISSONLY;
 			} else if (KWMATCH(p, "randomize")) {
 				ctl |= IBS_FETCH_CTL_RANDOMIZE;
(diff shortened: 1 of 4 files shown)


mashtizadeh:
@bsdimp @mhorne @afscoelho @aalok-agarwal

mhorne (approved):
LGTM. It looks like you have appropriate guards on the driver side.

lib/libpmc/pmc.ibs.3
mhorne, line 135: Something is incomplete here. Is `randomize` being removed or `opcount` being added?

mhorne (changes requested):

sys/dev/hwpmc/hwpmc_ibs.c
```diff
@@ -329,9 +329,9 @@ pmc_ibs_process_fetch(struct pmc *pm, struct trapframe *tf, uint64_t config)
 	memset(&mpd, 0, sizeof(mpd));
 
 	mpd.pl_type = PMC_CC_MULTIPART_IBS_FETCH;
-	mpd.pl_length = 4;
+	mpd.pl_length = PMC_MPIDX_FETCH_MAX;
 	mpd.pl_mpdata[PMC_MPIDX_FETCH_CTL] = config;
-	if (ibs_features) {
+	if ((ibs_features & CPUID_IBSID_IBSFETCHCTLEXTD) != 0) {
 		mpd.pl_mpdata[PMC_MPIDX_FETCH_EXTCTL] = rdmsr(IBS_FETCH_EXTCTL);
```
mhorne, line 335: Is it the read of this MSR when unimplemented that causes the kernel crash?

lib/libpmc/libpmc.c
```diff
@@ -719,11 +724,22 @@ ibs_allocate_pmc(enum pmc_event pe, char *ctrspec,
...
+
 	/* parse parameters */
 	ctl = 0;
 	if (pe == PMC_EV_IBS_FETCH) {
 		while ((p = strsep(&ctrspec, ",")) != NULL) {
 			if (KWMATCH(p, "l3miss")) {
+				if ((ibs_features & CPUID_IBSID_ZEN4IBSEXTENSIONS) == 0)
+					return (-1);
+
 				ctl |= IBS_FETCH_CTL_L3MISSONLY;
```
mhorne, line 743: Just looking for a bit more clarity on which cases are at risk of causing crashes.

Can passing this flag crash the kernel if the above feature is not implemented/supported? As discussed in https://github.com/freebsd/freebsd-src/pull/2140, this can be bypassed by a malicious user, if they use the hwpmc(4) ioctl interface directly.

Same question for the other checks added below.

sys/dev/hwpmc/hwpmc_ibs.c
mashtizadeh: Yes, so this change fixes this for those older processors and the removal off one MSR in newer AMDs in the op path.  The rest are permanent based on the bits available.

mashtizadeh: I'm tesitng on Zen2, 4, and 5. which have and don't have all three of the problematic ones.

lib/libpmc/libpmc.c
mashtizadeh: Yes, I have another patch that provides a more general fix to the entire problem across all of PMC.

lib/libpmc/pmc.ibs.3
mashtizadeh: No, the PPR I originally references when I wrote that was incompletely written it just explained the change in randomization when using opcount flag.  In fetch mode there's a randomize flag that enables randomization, in op mode there's only an opcount flag and randomization is always enabled.

mashtizadeh:
@mhorne This diff only makes things better unless you see anything that needs fixing lets land this.  Please take a look at #2157 and if you like that approach I'll write the Intel patch as well. FYI the problem isn't IBS, at least on amd, I hit it on the DF and/or UMC counters prior to my change to redo the field handling in libpmc.

lib/libpmc/pmc.ibs.3
mhorne: I still found something weird about the paragraph, with the remaining reference to `randomize`. I made an adjustment based on my understanding, so that I could proceed with the merge now.

It now looks like:
```
     ibs-op [,l3miss] [,ldlat=ldlat] [,opcount]
             Collect performance samples during instruction execution.  The
             opcount qualifier, upon reaching the maximum count, restarts the
             count with a random value between 1 and 127.
```

If I made a mistake, please let me know the correct wording and I will fix it right away.

PR #2134, opened 2026-04-09
hwpmc_ibs: Add external error handling

afscoelho:
Add EXTERR_CAT_HWPMC_IBS to the external error categories and replace generic EINVAL returns in ibs_allocate_pmc() with EXTERROR() calls that provide detailed error messages. This extended error handling will also serve as the basis for future IBS validation and checks

Sponsored by: AMD
Signed-off-by: Andre Silva <andasilv@amd.com>

### DIFF
--- a/sys/dev/hwpmc/hwpmc_ibs.c
+++ b/sys/dev/hwpmc/hwpmc_ibs.c
@@ -38,6 +38,9 @@
 #include <sys/smp.h>
 #include <sys/systm.h>
 
+#define	EXTERR_CATEGORY	EXTERR_CAT_HWPMC_IBS
+#include <sys/exterrvar.h>
+
 #include <machine/cpu.h>
 #include <machine/cpufunc.h>
 #include <machine/md_var.h>
@@ -166,16 +169,18 @@ ibs_allocate_pmc(int cpu __unused, int ri, struct pmc *pm,
 
 	/* check class match */
 	if (a->pm_class != PMC_CLASS_IBS)
-		return (EINVAL);
+		return (EXTERROR(EINVAL, "PMC class is not IBS"));
 	if (a->pm_md.pm_ibs.ibs_type != ri)
-		return (EINVAL);
+		return (EXTERROR(EINVAL,
+		    "IBS type %ju does not match PMC index %ju",
+		    (uint64_t)a->pm_md.pm_ibs.ibs_type, (uint64_t)ri));
 
 	caps = pm->pm_caps;
 
 	PMCDBG2(MDP, ALL, 1, "ibs-allocate ri=%d caps=0x%x", ri, caps);
 
 	if ((caps & PMC_CAP_SYSTEM) == 0)
-		return (EINVAL);
+		return (EXTERROR(EINVAL, "IBS requires SYSTEM capability"));
 
 	config = a->pm_md.pm_ibs.ibs_ctl;
 	pm->pm_md.pm_ibs.ibs_ctl = config;
--- a/sys/sys/exterr_cat.h
+++ b/sys/sys/exterr_cat.h
@@ -40,6 +40,7 @@
 #define	EXTERR_CAT_FORK		15
 #define	EXTERR_CAT_PROCEXIT	16
 #define	EXTERR_CAT_VMM		17
+#define	EXTERR_CAT_HWPMC_IBS	18
 
 #endif
 
--- a/lib/libc/gen/exterr_cat_filenames.h
+++ b/lib/libc/gen/exterr_cat_filenames.h
@@ -2,6 +2,7 @@
  * Automatically @generated, use
  * tools/build/make_libc_exterr_cat_filenames.sh
  */
+	[EXTERR_CAT_HWPMC_IBS] = "dev/hwpmc/hwpmc_ibs.c",
 	[EXTERR_CAT_VMM] = "dev/vmm/vmm_dev.c",
 	[EXTERR_CAT_FUSE_DEVICE] = "fs/fuse/fuse_device.c",
 	[EXTERR_CAT_FUSE_VFS] = "fs/fuse/fuse_vfsops.c",


afscoelho:
@mashtizadeh  @mhorne

sys/sys/exterr_cat.h
```diff
@@ -40,6 +40,7 @@
 #define	EXTERR_CAT_FORK		15
 #define	EXTERR_CAT_PROCEXIT	16
 #define	EXTERR_CAT_VMM		17
+#define	EXTERR_CAT_HWPMC_IBS	18
```
mhorne, line 43:
```suggestion
#define	EXTERR_CAT_HWPMC     18
```

We will want this category for the whole subsystem.

mhorne (commented):
Thanks for this. It is desirable to make use of the extended error reporting in hwpmc generally. Especially in these allocation paths with a lot of `EINVAL` cases.

sys/sys/exterr_cat.h
afscoelho: I did this way because it seems that the mechanism behaves like "one category -> one source file", not "one category -> one subsystem". Fuse for example has three categories: EXTERR_CAT_FUSE_DEVICE, EXTERR_CAT_FUSE_VFS and EXTERR_CAT_FUSE_VNOPS, each one for a different c file. We could use EXTERR_CAT_HWPMC_MOD for hwpmc_mod.c for example

sys/dev/hwpmc/hwpmc_ibs.c
```diff
@@ -166,16 +169,18 @@ ibs_allocate_pmc(int cpu __unused, int ri, struct pmc *pm,
 
 	/* check class match */
 	if (a->pm_class != PMC_CLASS_IBS)
-		return (EINVAL);
+		return (EXTERROR(EINVAL, "PMC class is not IBS"));
 	if (a->pm_md.pm_ibs.ibs_type != ri)
-		return (EINVAL);
+		return (EXTERROR(EINVAL,
```
mashtizadeh, line 174: Do we have a test case?  What does the user space see if we query the errors?

Because these error paths may be a bit problematic as the hwpmc_mod code will retry multiple times to allocate the counter.  We might be filling the log with innocuous error messages.

afscoelho: There are no automated tests for this change yet, and about filling the log, I think we are safe, because the EXTERROR mechanism doesn't produce any console/dmesg output and it stores errors in a per-thread kernel buffer and each call to EXTERROR() call overwrites the previous one (like the last error wins). So the loop you mentioned, from what I get, will not impact much if EXTERROR called multiple times

sys/sys/exterr_cat.h
afscoelho: @mhorne

mashtizadeh: I think Mitchell is probably right given how few errors trickle up from hwpmc you can just make one code for all hwpmc.

mhorne: Hmmm. @afscoelho you seem to be right that this is part of the design of the EXTERROR interface. It is ugly here, but ultimately unimportant.

sys/dev/hwpmc/hwpmc_ibs.c
```diff
@@ -166,16 +169,18 @@ ibs_allocate_pmc(int cpu __unused, int ri, struct pmc *pm,
...
+		    "IBS type %ju does not match PMC index %ju",
+		    (uint64_t)a->pm_md.pm_ibs.ibs_type, (uint64_t)ri));
 
 	caps = pm->pm_caps;
 
 	PMCDBG2(MDP, ALL, 1, "ibs-allocate ri=%d caps=0x%x", ri, caps);
 
 	if ((caps & PMC_CAP_SYSTEM) == 0)
-		return (EINVAL);
+		return (EXTERROR(EINVAL, "IBS requires SYSTEM capability"));
```
mhorne, line 183: Are we using the `PMC_CAP_SYSTEM` flag correctly here? Its purpose is to indicate that only instructions executing in the kernel should be measured. Likewise, `PMC_CAP_USER` says "please only measure instructions running in userspace".

Are unprivileged/userspace micro-ops measured by IBS, or not?

sys/dev/hwpmc/hwpmc_ibs.c
mhorne: Yeah, it will not pollute any logs. It gets emitted when a program calls `errx()` or similar.

I am not sure how well pmcstat will do with this change alone. But improving the error reporting in the allocation path is much needed.

mhorne (commented):
LGTM, but I am not ready to merge it yet. @afscoelho does it work? Were you able to trigger these messages?

I would like to see this extended to other hwpmc classess...

sys/dev/hwpmc/hwpmc_ibs.c
mhorne: Sorry, this is a drive-by question for @mashtizadeh. Not really the right place to be asking :sweat_smile:

sys/sys/exterr_cat.h
mashtizadeh: Just FYI @mhorne if you look at the header we wouldn't be the only ones not following that.  VMM and GEOM both do this.

sys/dev/hwpmc/hwpmc_ibs.c
mashtizadeh: If you look through the other modules they do weird things like the default for AMD's counters it to have neither set which means both get set by the kernel, rather than requiring both. I think the right thing to do across the board is if you have access to kernel and/or user stacks the corresponding flag should be required, but that's not what I see when I looked through a few architectures.  Ideally I'd like to make it consistent but that subtlely breaks the interface.  So I took it as a permission to access kernel stacks and required it set.

More problematic to me is that the CAP system changes the config in most cases which msotly derived in libpmc.

I suggested that @afscoelho make the Zen6 change and I believe we can publish that patch shortly.

afscoelho:
> LGTM, but I am not ready to merge it yet. @afscoelho does it work? Were you able to trigger these messages?
> 
> I would like to see this extended to other hwpmc classess...

 Yes, I was able to trigger it locally.

  The normal `libpmc` path does not really give me a way to generate these invalid IBS requests, so I used a
  temporary local cheat in `lib/libpmc/libpmc.c` to inject bad values before the request reaches the kernel. 

  The hack lets me set `HWPMC_IBS_TEST=nosys`, `badtype`, or `badclass`. For example:

  ```console
  $ HWPMC_IBS_TEST=badtype sudo -E pmcstat -S ibs-fetch -O /dev/null sleep 1
  pmcstat: ERROR: Cannot allocate system-mode pmc with specification "ibs-fetch": Invalid argument (IBS type 99 does not
  match PMC index 1)
```

  So the EXTERROR string is making it back through pmcstat as expected.

  And yes, agreed on extending this to the other hwpmc classes. I kept this PR scoped to IBS because that was the case I
  was working on, but I can either expand this PR to cover the other obvious allocation-path EINVALs, or send that as a
  follow-up if you prefer to keep this one small.

afscoelho:
> > LGTM, but I am not ready to merge it yet. @afscoelho does it work? Were you able to trigger these messages?
> > I would like to see this extended to other hwpmc classess...
> 
> Yes, I was able to trigger it locally.

@mhorne :- )

mhorne (approved):
Thanks. Let's go with this for now!

afscoelho:
@mashtizadeh

PR #2136, opened 2026-04-10
igc(4): fix Wake-on-LAN with magic packet on I225/I226

ricardobranco777:
WoL via magic packet did not work on igc(4) after poweroff: the link would come back up on the PHY but magic packets never woke the host. Two issues were responsible.

First, the I225/I226 PHY Power Management register (PHPM) has bits that disable 100/1000/2500 Mb/s link negotiation while the device is in D3.  When set, the PHY drops the link in D3 and no magic packet ever reaches the MAC.  Clear DIS_100_D3, DIS_1000_D3 and DIS_2500_D3 in igc_enable_wakeup() so the PHY can keep a link in D3.  This matches what the Linux igc driver does on the WoL configuration path.

Second, igc_enable_wakeup() cleared IGC_WUFC_MAG from sc->wol when IFCAP_WOL_MAGIC was disabled, but never set it again when the capability was (re)enabled via ifconfig.  As a result the subsequent write to WUFC programmed zero and magic-packet detection stayed off. Set IGC_WUFC_MAG in the corresponding else branch.

While here, only advertise IFCAP_WOL_MAGIC in igc_if_attach_pre() when the magic-packet bit is actually present in sc->wol, rather than for any non-zero value.  This matches the "Enable only WOL MAGIC by default" comment immediately above.

PR: 282140
Reported by: Mario Agustin Carranza
Authored by: Hiroshi Nishida
Tested by: Hiroshi Nishida & Ricardo Branco

### DIFF
--- a/sys/dev/igc/if_igc.c
+++ b/sys/dev/igc/if_igc.c
@@ -683,7 +683,7 @@ igc_if_attach_pre(if_ctx_t ctx)
 
 	/* Enable only WOL MAGIC by default */
 	scctx->isc_capenable &= ~IFCAP_WOL;
-	if (sc->wol != 0)
+	if (sc->wol & IGC_WUFC_MAG)
 		scctx->isc_capenable |= IFCAP_WOL_MAGIC;
 
 	iflib_set_mac(ctx, hw->mac.addr);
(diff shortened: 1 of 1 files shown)


kev009:
Closing per PR 282140, a comprehensive WoL cleanup was commited

PR #2137, opened 2026-04-11
asmc: add support for MacBookPro13,1

mgartner:
#### asmc: add support for MacBookPro13,1

This commit adds support for the MacBookPro13,1 (late 2016, 13-inch). The SMC
keys were collected from https://logi.wiki/index.php/SMC_Sensor_Codes. Two
temperature keys are omitted because they fail to be read: TI0P (IO Proximity)
and Ta0P (Ambient Air).

Note that the with this model the `dev.asmc.0.fan.0.minspeed` setting only
applies when the fans have been activated by the system. In my testing, the fans
did not spin up until CPU temperatures hit about 80C. At lower temperatures, the
fans will happily ignore the minimum speed and remain at 0 rpm.

#### asmc: correctly label ASMC_KEY_FANMINSPEED as read-write

### DIFF
--- a/sys/dev/asmc/asmc.c
+++ b/sys/dev/asmc/asmc.c
@@ -334,6 +334,12 @@ static const struct asmc_model asmc_models[] = {
 	  ASMC_MBP115_TEMPS, ASMC_MBP115_TEMPNAMES, ASMC_MBP115_TEMPDESCS
 	},
 
+	{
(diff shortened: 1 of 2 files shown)


bsdimp (approved):
This seems to be correct. w/o hardware it's hard to be 100% sure though.

bsdimp:

PR #2139, opened 2026-04-14
Bugzilla bug 294499 | termcap: fix screen entry standout mode (so/se) capabilities

thegushi:
so=\E[3m (italic) is incorrect, should be so=\E[7m (reverse video). se=\E[23m (italic off) is incorrect, should be se=\E[27m (reverse off).

mr=\E[7m (reverse video) is correctly defined in the same entry. screen-256color inherits from screen via tc=screen and is fixed transitively.

More details: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=294499


cschuber:
Thanks for the patch. I will test it locally.

cschuber:
Commit pushed to FreeBSD repo.

PR #2140, opened 2026-04-14
hwpmc: Add IBS capability control policy

afscoelho:
The IBS driver accepted user-supplied PMC configurations without
validating them against what the CPU actually supports. On CPUs with
the extended op counter or load latency filtering, this could result
in writing unsupported bits to MSRs and incorrect period calculations.
Several CPUID feature flag bit definitions were also wrong (shifted
one nibble too low), causing feature detection to silently fail.

This change adds a capability control policy that initializes allowed
configuration masks and period bounds at driver load time based on
CPUID-detected features, and validates user configurations before
PMC allocation.

Tested on Zen 4 hardware with pmcstat(8) using IBS fetch and op
modes.


afscoelho:
@mhorne @mashtizadeh

sys/dev/hwpmc/hwpmc_ibs.c
```diff
@@ -592,6 +700,10 @@ pmc_ibs_initialize(struct pmc_mdep *pmc_mdep, int ncpus)
 	} else {
 		ibs_features = 0;
 	}
+	if ((ibs_features & CPUID_IBSID_IBSFFV) == 0)
```
RaghavendraKT80: Can we combine with the above else condition with else if ( (ibs_features & CPUID_IBSID_IBSFFV) == 0). OR do you think current code is simple. Not a mandatory thing to fix.

sys/dev/hwpmc/hwpmc_ibs.c
```diff
@@ -68,6 +74,104 @@ struct ibs_cpu {
...
+			return (EINVAL);
+		if ((config & IBS_OP_CTL_L3MISSONLY) == 0)
+			return (EINVAL);
+
+		unsupported &= ~IBS_OP_CTL_LDLATMASK;
+		if ((ibs_features & CPUID_IBSID_ZEN4IBSEXTENSIONS) == 0)
+			unsupported &= ~IBS_OP_CTL_L3MISSONLY;
+
+		ldlat = IBS_OP_CTL_TO_LDLAT(config);
+		if (ldlat < 128 || ldlat > 2048)
```
RaghavendraKT80: 128, 2048 hardcoded, but I hope it is Okay

sys/dev/hwpmc/hwpmc_ibs.c
afscoelho: sure, I folded the IBSFFV check into the CPUID leaf-present path, now stays next to the CPUID read. Thanks.

sys/dev/hwpmc/hwpmc_ibs.c
afscoelho: sure, thanks. I replaced the inline literals with named macros

sys/dev/hwpmc/hwpmc_ibs.h
```diff
@@ -112,10 +117,23 @@
 #define IBS_OP_CTL_VALID		(1ULL << 18) /* Valid */
 #define IBS_OP_CTL_ENABLE		(1ULL << 17) /* Enable */
 #define IBS_OP_CTL_L3MISSONLY		(1ULL << 16) /* L3 Miss Filtering */
-#define IBS_OP_CTL_MAXCNTMASK		0x07F0FFFFULL
-
-#define IBS_OP_CTL_LDLAT_TO_CTL(_c)	((((ldlat) >> 7) - 1) << 59)
-#define IBS_OP_INTERVAL_TO_CTL(_c)	((((_c) >> 4) & 0x0000FFFFULL) | ((_c) & 0x07F00000))
+#define IBS_OP_CTL_MAXCNTMASK		0x07F0FFFFULL /* Max Count */
```
mashtizadeh, line 120: You need should rebase on top of #2133 when we land it.  You'll have a few commit conflicts.

mashtizadeh (commented):
Wouldn't it be a lot simpler to compute the mask rather than check the values?  For example we don't care about the load latency because the bits are shifted in such a way that only those bits are valid.

sys/dev/hwpmc/hwpmc_ibs.h
mashtizadeh: Also you should avoid mixing logic and whitespace changes might just be your editor

mashtizadeh (changes requested):
If you use a mask rather than decode the entire control value then this could be a lot shorter just maybe a dozen lines.  Once you do that here we need the same thing for hwpmc_amd.c

sys/dev/hwpmc/hwpmc_ibs.c
```diff
@@ -54,6 +54,12 @@ struct ibs_descr {
  * Globals
  */
 static uint64_t ibs_features;
+static uint64_t ibs_fetch_allowed_mask;
+static uint64_t ibs_fetch_min_period;
+static uint64_t ibs_fetch_max_period;
+static uint64_t ibs_op_allowed_mask;
+static uint64_t ibs_op_min_period;
+static uint64_t ibs_op_max_period;
```
mhorne, line 62:
```suggestion
static uint64_t ibs_features;
static uint64_t ibs_fetch_allowed_mask;
static uint64_t ibs_op_allowed_mask;
```

Some vars don't appear to be needed; you can just use the named constants.

mhorne (commented):
Okay. In principle, I like the change. We should explore the possible surface of harm...

The general practice of kernel code is to sanitize/validate all user inputs, even redundantly. It's just a good idea.

It is true that some PMCs can be allocated by unprivileged, potentially malicious users. By default, these are limited to process-level PMCs, and subject to resource limits.

The obligation of the hwpmc(4) interface is to prevent misprogramming of PMCs if they would somehow affect the system negatively. The driver is not obligated to prevent "foot-shooting", or nonsense results due to user error. libpmc does try to prevent this, and we recommend its use.

(There are other hwpmc drivers, e.g. hwpmc_amd or hwpmc_core which do very little to validate the counter configuration.)

I am not sure if any of the fields validated here pose a real harm or exploit. That said, I don't think it's wrong to plug the holes.

sys/dev/hwpmc/hwpmc_ibs.c
mashtizadeh: @mhorne Invalid values can cause GPs which is bad so we need to do error checking for some AMD PMC infrastructure.

My problem is that the logic could be way simpler if we instead compute a mask rather than try to decode and verify values.  This should be about a dozen lines of code in the ibs initialization routine

afscoelho: > Some vars don't appear to be needed; you can just use the named constants.

right, I removed those min/max period globals and the unused decode/range helpers. The IBS side now just keeps the CPUID-derived fetch/op masks.

afscoelho:
> If you use a mask rather than decode the entire control value then this could be a lot shorter just maybe a dozen lines. Once you do that here we need the same thing for hwpmc_amd.c

That's right. I changed the IBS validation to build/use the allowed mask and reject the rest. I added the same kind of bit check to the AMD PMU raw config path. It uses a different mask for core, L3, and DF counters

sys/dev/hwpmc/hwpmc_ibs.h
afscoelho: > You need should rebase on top of #2133 when we land it. You'll have a few commit conflicts.

Yep, I'll rebase once #2133 lands. I expect to drop or resolve the conflicts during the rebase.

afscoelho:
> Okay. In principle, I like the change. We should explore the possible surface of harm...
> 
> The general practice of kernel code is to sanitize/validate all user inputs, even redundantly. It's just a good idea.
> 
> It is true that some PMCs can be allocated by unprivileged, potentially malicious users. By default, these are limited to process-level PMCs, and subject to resource limits.
> 
> The obligation of the hwpmc(4) interface is to prevent misprogramming of PMCs if they would somehow affect the system negatively. The driver is not obligated to prevent "foot-shooting", or nonsense results due to user error. libpmc does try to prevent this, and we recommend its use.
> 
> (There are other hwpmc drivers, e.g. hwpmc_amd or hwpmc_core which do very little to validate the counter configuration.)
> 
> I am not sure if any of the fields validated here pose a real harm or exploit. That said, I don't think it's wrong to plug the holes.

Thanks, that makes sense. I reworked to focus on the kernel safety part: don't let unsupported control bits get to wrmsr. For IBS that now means a CPUID-based allow mask, instead of decoding the whole control value in the kernel. I kept the LATFLTEN/L3MISSONLY check

mashtizadeh:
I think this diff looks a lot better I'll take a look at it during the day.  Only high level comment is that we need a sysctl to override it for NDA'ed users, maybe we can just provide a mask of additional bits to allow.

Also take a look at #2157 @RaghavendraKT80 @afscoelho

sys/dev/hwpmc/hwpmc_amd.c
```diff
@@ -178,6 +178,43 @@ struct amd_cpu {
...
+static uint64_t
+amd_config_mask(enum sub_class subclass, uint64_t caps)
+{
+	uint64_t mask;
+	int family;
+
+	family = CPUID_TO_FAMILY(cpu_id);
+	switch (subclass) {
+	case PMC_AMD_SUB_CLASS_CORE:
+		mask = AMD_VALID_BITS;
```
mashtizadeh: I think it would be nice to move a lot of the macros with arguments into the header to make this a bit simpler looking.  Or even go as far to to add AMD_PMC_L3_FAMILY17_MASK so only the PRECISERETIRE requires any computation.

sys/dev/hwpmc/hwpmc_amd.c
```diff
@@ -351,16 +388,20 @@ amd_allocate_pmc(int cpu __unused, int ri, struct pmc *pm,
 	    ((pd->pd_caps & PMC_CAP_PRECISE) == 0))
 		return (EINVAL);
 
-	PMCDBG2(MDP, ALL, 1,"amd-allocate ri=%d caps=0x%x", ri, caps);
+	PMCDBG2(MDP, ALL, 1, "amd-allocate ri=%d caps=0x%x", ri, caps);
```
mashtizadeh: Avoid mixing white space changes with a feature changing diff

sys/dev/hwpmc/hwpmc_amd.c
mashtizadeh: Can we compute the values instead during initialization and store the result in global variables?  Mostly it should be copying the #defined values.

sys/dev/hwpmc/hwpmc_amd.c
```diff
@@ -399,7 +440,7 @@ amd_allocate_pmc(int cpu __unused, int ri, struct pmc *pm,
 	if ((caps & PMC_CAP_SYSTEM) != 0)
 		config |= AMD_PMC_OS;
 	if ((caps & (PMC_CAP_USER | PMC_CAP_SYSTEM)) == 0)
-		config |= (AMD_PMC_USR|AMD_PMC_OS);
+		config |= (AMD_PMC_USR | AMD_PMC_OS);
```
mashtizadeh: Again avoid whitespace changes, it's 22 year old typeo

sys/dev/hwpmc/hwpmc_ibs.c
```diff
@@ -439,7 +490,7 @@ pmc_ibs_intr(struct trapframe *tf)
 	}
 
 	if (retval == 0) {
-		// Lets check for a stray NMI when stopping
+		/* Check for a stray NMI when stopping. */
```
mashtizadeh: Again avoid unrelated changes like this

mashtizadeh (changes requested):
One last change can we add a sysctl for each counter type that allows the user to set a mask of extra  allowable bits.  We need to override this for testing my change and NDA'ed counters.

I would prefer you remove all the excess whitespace changes that are in unrelated parts of the code.

afscoelho:
> I think this diff looks a lot better I'll take a look at it during the day. Only high level comment is that we need a sysctl to override it for NDA'ed users, maybe we can just provide a mask of additional bits to allow.
> 
> Also take a look at #2157 @RaghavendraKT80 @afscoelho

I added five RWTUN uint64 knobs under kern.hwpmc, all default 0. They are OR-ed into the allow mask validation time

sys/dev/hwpmc/hwpmc_amd.c
afscoelho: > I think it would be nice to move a lot of the macros with arguments into the header to make this a bit simpler looking. Or even go as far to add AMD_PMC_L3_FAMILY17_MASK so only the PRECISERETIRE requires any computation.

Done.  AMD_PMC_L3_FAMILY17_MASK, AMD_PMC_L3_FAMILY19_MASK, AMD_PMC_DF_FAMILY17_MASK and AMD_PMC_DF_FAMILY1A_MASK in hwpmc_amd.h

> Can we compute the values instead during initialization and store the result in global variables? Mostly it should be copying the #defined values.

Yep, folded that in too. Please take a look at amd_init_policy(), it now sets amd_core_allowed_mask / amd_l3_allowed_mask / amd_df_allowed_mask once at driver init

mashtizadeh:
Mitchell landed my diff so you can rebase.  Otherwise it looks good to me.

Only change when you rebase, I think you want CTLFLAG_RDTUN not CTLFLAG_RWTUN.  This will ensure that you can only change the variable on boot.

afscoelho:
> Mitchell landed my diff so you can rebase. Otherwise it looks good to me.
> 
> Only change when you rebase, I think you want CTLFLAG_RDTUN not CTLFLAG_RWTUN. This will ensure that you can only change the variable on boot.

Ok, rebased and RWTUN changed to RDTUN

mashtizadeh (approved):
Looks good to me, much easier to reason about!

mhorne:
LGTM. Can you please confirm the following details:

Is it 'Sponsored by: AMD' ?
Is "Andre Silva <andasilv@amd.com>" the correct attribution? Your Signed-off-by does not match the commit authorship.

afscoelho:
> LGTM. Can you please confirm the following details:
> 
> Is it 'Sponsored by: AMD' ? Is "Andre Silva [andasilv@amd.com](mailto:andasilv@amd.com)" the correct attribution? Your Signed-off-by does not match the commit authorship.

Confirmed on both  and good catch on the Signed-off-by mismatch, thanks! That was my old email lingering in git config. Just force-pushed: authors fixed and Sponsored by: AMD added.
   Thanks!

PR #2141, opened 2026-04-15
hkbd/ukbd: sysctls to swap macbook kbd modifiers

tslight:
Many applications, desktop environments, window managers & text editors favor the usage of Alt or Ctrl over Super (Cmd). On a Macbook it is quite annoying that the Super (Cmd) key gets pride of place by the spacebar. 

The standard MacBook Cmd key location only really makes sense for macOS or maybe in some tiling wm if Mod4/Super is your main modifier. For most mainstream desktops and window managers, having Alt or Ctrl in that location makes much much more sense.

This patch adds two sysctls for swapping either Opt(Alt) or Ctrl with Cmd(Super).

Linux has similar sysctls to this; allowing a user to make an Apple keyboard more "orthodox"/useful at a level that takes effect independent of typing context - ie) tty, Xorg and/or wayland.

Having a sysctl to do these swaps means that a user doesn't have to faff about with both creating a custom vt keymap AND figure out which magic `setxkbmap` incantation one needs to make one's keyboard behave as desired across environments.

@wulf7

### DIFF
--- a/share/man/man4/hkbd.4
+++ b/share/man/man4/hkbd.4
@@ -151,6 +151,18 @@ tunables:
 Debug output level, where 0 is debugging disabled and larger values increase
 debug message verbosity.
 Default is 0.
+.It Va hw.hid.hkbd.apple_swap_cmd_opt
+Swap the Command & Option keys on Apple keyboards if set to 1.
+Default is 0.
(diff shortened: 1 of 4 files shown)


wulf7:
Committed. Thank you.

Could you add a brief description of previously submitted fn+key mappings to manpages?

PR #2143, opened 2026-04-16
hwpmc_ibs: Zero invalid IBS Op RIP samples

afscoelho:
Clear the exported RIP for IBS Op samples when hardware reports RIPINVALID.

Sponsored by: AMD

Signed-off-by: Andre Silva <andasilv@amd.com>


afscoelho:
@mashtizadeh @mhorne

sys/dev/hwpmc/hwpmc_ibs.c
mashtizadeh, line 365: Why don't we just avoid reading the MSR at this line

mashtizadeh (changes requested):
What's the goal of the change?  There are a half a dozen flags and flag combinations that invalidate various IBS fields and whole rdmsrs.

Either your goal is to eliminate rdmsr or log entries, which you didn't do with this change.

Otherwise this should be done by the consumer of pmclog.

FYI linux does not have this logic either, they have different logic to do with errata around invalid RIPs.

afscoelho:
> [mashtizadeh](/mashtizadeh)

/

> What's the goal of the change? There are a half a dozen flags and flag combinations that invalidate various IBS fields and whole rdmsrs.
> 
> Either your goal is to eliminate rdmsr or log entries, which you didn't do with this change.
> 
> Otherwise this should be done by the consumer of pmclog.
> 
> FYI linux does not have this logic either, they have different logic to do with errata around invalid RIPs.

Yeah, you're right, I'll drop this. Looking at it again, pmcstat_print_ibs_op already checks IBS_OP_DATA_RIPINVALID before printing the RIP, and the actual sample PC used for profiling comes from the trapframe in pmc_add_sample, not from the IBS Op RIP, the invalid bit is already in OP_DATA for anyone who wants to check it. And you're right about Linux too, I went and looked at arch/x86/events/amd/ibs.c. It doesn't touch the logged RIP either; 

Thanks for the pushback :-)

PR #2144, opened 2026-04-17
pfctl: clarify usage of load option flags

overhacked:
`pfctl -A`, `-N`, `-O`, and `-R` restrict which rule types and options are loaded. The man page language ("Load only...") does not make it clear that these options can be combined to (re)load multiple rule types and/or options without reloading the entire packet filter. Add language to make it explicitly clear that these flags combine.

Also, fix man page and error message for `-S` option. The `pfctl -S` flag was added to disable DNS resolution in https://reviews.freebsd.org/D50724 but documentation and error messages refer to a `pfctl -N` flag for the same purpose. The `pfctl -N` flag performs an unrelated function, so the docs and error messages need to be changed. Caught this when revising documentation for the `-N` flag.

### DIFF
--- a/sbin/pfctl/pfctl.8
+++ b/sbin/pfctl/pfctl.8
@@ -458,7 +458,7 @@ and
 flags to load multiple rule types without (re)loading all rules.
 .It Fl r
 Perform reverse DNS lookups on states and tables when displaying them.
-.Fl N
+.Fl S
 and
 .Fl r
 are mutually exclusive.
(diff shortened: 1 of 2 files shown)


kprovost:
Merged as https://cgit.freebsd.org/src/commit/?id=19eecf94fa16aac6f90a4610d2f342f70060674a and https://cgit.freebsd.org/src/commit/?id=18de44e7b44209d94999523a93d83031e26a5719

PR #2148, opened 2026-04-19
freebsd-version: fix pathname expansion bug

namiarj:
Hi,
I discovered several bugs in freebsd-version:

### Bugs:
1. `-j` option is susceptible to pathname expansion in the shell script due to the use of unquoted variable names in several places. I was able to run jails with special characters in name such as * and ?. Not only freebsd-version might not correctly work with these jail names but one can use characters like *, /, ? to reveal information in filesystem or trigger denial-of-service type of attacks.
2. freebsd-version -j "" exits silently.

### Examples:
```
# freebsd-version -j "/root/*"
jexec: jail "/root/Book" not found

# freebsd-version -j "/root/*/*/*/*"
jexec: jail "/root/Book/Jung/C.G." not found

# freebsd-version -j "/*/*/*/*/*/*/*/*/*"
# (excessive expansion leading to high resource usage)

- When provided with an empty string, the command exits silently:
# freebsd-version -j ""

# freebsd-version -j "/home/*/*/*/*/*/*/*/*"
jexec: jail "/home/nami/book/Prep/Education/Homeschool/Grade" not found

# freebsd-version -j "/etc/*conf"
jexec: jail "/etc/blacklistd.conf" not found

# freebsd-version -j "/jail/?????"
jexec: jail "/jail/nginx" not found

# freebsd-version -j "/jail/???"
jexec: jail "/jail/dev" not found

# freebsd-version -j "/jail/av/root/???"
jexec: jail "/jail/av/root/log" not found


```
This is the result on the patched version:
```
# freebsd-version -j "/*/*"
jexec: jail "/*/*" not found

# freebsd-version -j acme
15.0-RELEASE

# freebsd-version -j "?"
jexec: jail "?" not found

```
Commands like this on a slow HDD server take a lot of resources:
```
# time freebsd-version -j "/*/*/*/*/*"
jexec: jail "/jail/acme/boot/defaults/loader.conf" not found
0.133u 121.398s 2:01.70 99.8% 152+172k 0+0io 0pf+0
```
The patched version on the same server:
```
# time freebsd-version -j "/*/*/*/*/*"
jexec: jail "/*/*/*/*/*" not found
0.000u 0.077s 0:00.05 140.0% 379+358k 0+0io 0pf+0w
```
### Changes include:
- Adding quotation marks to the variable affected by the `-j` option.
- Setting `-f` to disable pathname expansion.
- Minor styling improvements.
- A check for an empty `-j` option.

### DIFF
--- a/bin/freebsd-version/freebsd-version.sh.in
+++ b/bin/freebsd-version/freebsd-version.sh.in
@@ -26,7 +26,7 @@
 #
 #
 
-set -e
+set -ef
 
 USERLAND_VERSION="@@REVISION@@-@@BRANCH@@"
 
@@ -88,7 +88,7 @@ userland_version() {
 #
 jail_version() {
 	for i in $jail; do
-		jexec -- $i freebsd-version
+		jexec -- "$i" freebsd-version
 	done
 }
 
--- a/bin/freebsd-version/freebsd-version.sh.in
+++ b/bin/freebsd-version/freebsd-version.sh.in
@@ -118,6 +118,10 @@ main() {
 			opt_u=1
 			;;
 		j)
+			if [ -z "$OPTARG" ] ; then
+				echo "No arg for -j option" >&2
+				usage
+			fi
 			if [ $opt_j ] ; then
 				jail="$jail $OPTARG"
 			else
--- a/bin/freebsd-version/freebsd-version.sh.in
+++ b/bin/freebsd-version/freebsd-version.sh.in
@@ -84,11 +84,11 @@ userland_version() {
 }
 
 #
-# Print the hardcoded userland version of a jail.
+# Print the hardcoded userland version(s) of jail(s).
 #
 jail_version() {
-	for i in $jail; do
-		jexec -- "$i" freebsd-version
+	for j in $jail; do
+		jexec -- "$j" freebsd-version
 	done
 }
 


spmzt (changes requested):
Hi,

Thank you for taking the time to submit this PR and security concerns.
I appreciate the effort, but I'd like to respectfully push back on a few points and suggest some revisions:

On the security/bug claims:
1. Pathname expansion and information disclosure: This isn't actually a vulnerability here. The behavior you're describing on attempting to use wildcards in jail names doesn't reveal filesystem information that the user doesn't already have read access to. Shell expansion happens locally, So you can only see what you have permission to see anyway.

2. Empty argument handling: While a warning could be a nice improvement, this isn't a bug. The script's behavior is predictable and expected. But I'm against this change since `-j` can be specified multiple times and this feature is already used in many shell scripts out there. I personally expect `freebsd-version -j "$NULL"` to do nothing in my for-loop. Considering it helps with `xargs` too.

3. Resource usage concerns: The sequential jexec execution you've shown isn't a denial-of-service vulnerability.
A single-threaded loop trying invalid jail names takes negligible resources (as your own `time` output shows: `0+0io 0pf+0`). This is normal script behavior, not a security issue.

The quoting of variables and use of `set -f` are legitimate improvements for correctness.
I'd support those changes independently.
- Focus on shell scripting best practices (proper quoting, disabling pathname expansion)
- Drop the empty argument check and styling.
- Update the PR description to reflect what this actually accomplishes

namiarj:
>
> 1- The behavior you're describing on attempting to use wildcards in jail
> names doesn't reveal filesystem information that the user doesn't already
> have read access to.

Sure, the FreeBSD user that is running the utility has access to the file
names. But if limited access is given to remote users to pass strings into
the -j option they could potentially could abuse it. Imagine the scenario
that an API used "freebsd-version -j" to query jail versions and file names
could be leaked through the API unintentionally.

Shell expansion happens locally, So you can only see what you have
permission to see anyway.

No it's happening in two places where unquoted variables are used. Setting
-f is enough to disable expansion in the whole script but. One of the
variable cannot be quoted because it's iterating over an space separated
list. The other variable can be quoted for correctness . You can disable
expansion in the script and still use file name expansion in your local
shell as shown below:

# freebsd-version -j "tes*"
jexec: jail "tes*" not found
# touch test
# freebsd-version -j tes*
jexec: jail "test" not found
# freebsd-version -j "tes*"
jexec: jail "tes*" not found

2- My issue was freebsd-version -j or freebsd-version -j"" would printout
meaningful error messages while freebsd-version -j "" would just exit
silently. I'm not sure what's happening here.

3- I'm not sure how the multiple depth level file expansions are using the
resources but on a machine with slow disk it can keep the script busy at
100% CPU for up to several minutes. Only on this machine with very slow HDD
on others it's much faster.

# time freebsd-version -j "/*"
jexec: jail "/COPYRIGHT" not found
0.008u 0.032s 0:00.04 75.0% 125+224k 0+0io 0pf+0w
# time freebsd-version -j "/*/*"
jexec: jail "/bin/[" not found
0.000u 0.761s 0:00.76 100.0% 149+177k 0+0io 0pf+0w
# time freebsd-version -j "/*/*/*"
jexec: jail "/boot/defaults/loader.conf" not found
0.007u 5.488s 0:05.49 99.8% 153+173k 0+0io 0pf+0w
# time freebsd-version -j "/*/*/*/*"
jexec: jail "/etc/periodic/daily/100.clean-disks" not found
0.007u 81.728s 1:21.73 99.9% 152+172k 0+0io 0pf+0w

About the styling issue I think the comment below is better to be plurar
since the jail_version function can print multiple jail names not one.
# Print the hardcoded userland version of a jail.

namiarj:
Alright reading your comment again I feel I might have misunderestood some
parts of it. My access to GitHub is very limited right now. Can your go on
and make the right changes yourself?

namiarj:
>
> Resource usage concerns: The sequential jexec execution you've shown isn't
> a denial-of-service vulnerability.
>
BTW, It's not the jexec execution that is taking resources it seems to be
the shell. Even on my SSD laptop it takes fare amount of resources and even
made some weird noices from the laptop I suppose from the SSD :D
# time freebsd-version -j "/*/*/*/*/*/*/*"
jexec: jail "/home/nami/bin/WinBox_Windows/assets/img/winbox.png" not found
       15.90 real         0.47 user         6.24 sys

PR #2150, opened 2026-04-19
iflib: add missing CURVNET_RESTORE() in iflib_rxeof()

pgzh:
CURVNET_RESTORE() needs to be explicitly called if ctx->isc_rxd_pkt_get() returns an error before returning from the function.
Signed-off-by: peter.ganzhorn@gmail.com
PR: <286179>


gmshake:
Committed via f49f61f19463c21125bb1215cf8e0530f52953e3 .

The commit message is amended a little, notably the `PR: <286179>` is removed, given the patch is not related to the reported issue.

Thanks for your contribution! This PR can be closed.

PR #2153, opened 2026-04-20
pmc: Remove old AMD events support

mhorne:
A quick draft. Mostly for heads-up / discussion if this is the desired direction.

The original "K8" AMD events have been effectively disabled since the switch to the newer "pmu-events" sources, from upstream linux (~2018, off the top of my head).

I want to bring this in line we the reality that we de-facto only support the use of AMD Zen1 and newer CPUs with hwpmc.

Our docs and code claim otherwise, and appear quite old. 

There is a straightforward clean-up of the old code here. Nonetheless I will likely put out a notice to the mailing lists about this.

TODO: update pmc.amd(3) to reflect the newer support reality.


mhorne:
@mashtizadeh @afscoelho

mashtizadeh (approved):
Looks great.

jkoshy (approved):

mashtizadeh:
We should follow this up with a rename the class and keep an alias for build compat.  If we rename the class string we should check that Aalok from AMD is okay with that.

If anyone wants to use the old events we could export these symbols as a json file, the code I wrote for the register detection should work with older and future processors up to the ranges AMD reserved.

PR #2154, opened 2026-04-20
powerpc: remove dead call into x86 perfmon

mashtizadeh:
FreeBSD and NetBSD copied these lines from the x86 architecture when porting to other architectures and forgot to delete them.

Sponsored by: Netflix


mashtizadeh:
@mhorne @bsdimp

mhorne (approved):

PR #2157, opened 2026-04-21
hwpmc: Harden ioctl interface for AMD processors

mashtizadeh:
In practice we cannot be sure about what is a valid value for a given processor as AMD has redefined these values and we do not want to disable NDA'ed usage of the counters.  Various reserved bits can and do lead to general protection faults.  To avoid this we should use the wrmsr_safe when writing values from userspace.

Sponsored by: Netflix

### DIFF
--- a/sys/dev/hwpmc/hwpmc_amd.c
+++ b/sys/dev/hwpmc/hwpmc_amd.c
@@ -505,6 +505,7 @@ amd_release_pmc(int cpu, int ri, struct pmc *pmc __unused)
 static int
 amd_start_pmc(int cpu __diagused, int ri, struct pmc *pm)
 {
+	int status;
 	const struct amd_descr *pd;
 	uint64_t config;
 
@@ -531,8 +532,9 @@ amd_start_pmc(int cpu __diagused, int ri, struct pmc *pm)
 
 	PMCDBG1(MDP, STA, 2, "amd-start config=0x%x", config);
 
-	wrmsr(pd->pm_evsel, config);
-	return (0);
+	status = wrmsr_safe(pd->pm_evsel, config);
+
+	return (status);
 }
 
 /*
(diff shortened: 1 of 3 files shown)


mashtizadeh:
@mhorne @bsdimp

My thought is to ensure the state machine works the way we assume and we use wrmsr_safe when setting up the counters on Intel and AMD.  I also have a patch to add rdpmc_safe since again AMD's layout makes me unsure that we should trust it won't change.

mhorne (commented):
I really like this. What about other cases, e.g. `amd_pmc_write()`?

mashtizadeh:
> I really like this. What about other cases, e.g. `amd_pmc_write()`?

So the problem is worse than you think in general rdmsr, wrmsr, and rdpmc all can GP if the register is invalid or you write to a reserved bit.  I also don't want to slow down any code inside the interrupt handler because we'll have problems stemming from too many slow NMIs.

I'm reasoning through the logic but I don't think we need to place it everywhere.  read/write should only touch the counter which is okay to change.  I need a few days but I'll check all the paths and finalize the patch.  I'll add Intel to this one too.

mhorne:
> 
> So the problem is worse than you think in general rdmsr, wrmsr, and rdpmc all can GP if the register is invalid or you write to a reserved bit. I also don't want to slow down any code inside the interrupt handler because we'll have problems stemming from too many slow NMIs.
> 

Agreed, we definitely don't want to slow down the fast path.

> I'm reasoning through the logic but I don't think we need to place it everywhere. read/write should only touch the counter which is okay to change. I need a few days but I'll check all the paths and finalize the patch. I'll add Intel to this one too.

Sounds great, take your time with it. If it helps we can land the AMD stuff first before Intel.

You are right, safety is not needed when writing to the counter itself.

mashtizadeh:
So far in my testing this seems to capture everything I can trigger from userspace.  I think I gotta rebase after Andre’s patch and we can land it.

mashtizadeh:
Waiting on Andre's patch to land.  It'll probably be easier for me to rebase than him.  https://reviews.freebsd.org/D56914

PR #2159, opened 2026-04-22
compat/linprocfs: Add some entries to /proc/sys/fs

ricardobranco777:
Add some entries to /proc/sys/fs that can be trivially mapped to the equivalent FreeBSD sysctls.

Also add /proc/sys/kernel/threads-max

Fixes https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=294713

### DIFF
--- a/sys/compat/linprocfs/linprocfs.c
+++ b/sys/compat/linprocfs/linprocfs.c
@@ -1737,6 +1737,23 @@ linprocfs_dotainted(PFS_FILL_ARGS)
 }
 
 /*
+ * Filler function for proc/sys/kernel/threads-max
+ */
+static int
+linprocfs_dothreads_max(PFS_FILL_ARGS)
+{
+	int res, error;
+	size_t size = sizeof(res);
+
+	error = kernel_sysctlbyname(curthread, "kern.maxproc",
+	    &res, &size, NULL, 0, 0, 0);
+	if (error != 0)
+		return (error);
+	sbuf_printf(sb, "%d\n", res);
+	return (0);
+}
+
+/*
  * Filler function for proc/sys/vm/min_free_kbytes
  *
  * This mirrors the approach in illumos to return zero for reads. Effectively,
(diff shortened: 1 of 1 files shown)


sys/compat/linprocfs/linprocfs.c
```diff
@@ -2215,6 +2215,129 @@ linprocfs_dosysvipc_shm(PFS_FILL_ARGS)
...
+	return (0);
+}
+
+/*
+ * Filler function for proc/sys/fs/overflowgid
+ */
+static int
+linprocfs_dooverflowgid(PFS_FILL_ARGS)
+{
+	sbuf_printf(sb, "%u\n", GID_NOBODY);
```
markjdb: This should be GID_NOGROUP probably.

sys/compat/linprocfs/linprocfs.c
```diff
@@ -2215,6 +2215,129 @@ linprocfs_dosysvipc_shm(PFS_FILL_ARGS)
...
+	if (error != 0)
+		check_uid = 0;
+	size = sizeof(check_gid);
+	error = kernel_sysctlbyname(curthread,
+	    "security.bsd.hardlink_check_gid",
+	    &check_gid, &size, NULL, 0, 0, 0);
+	if (error != 0)
+		check_gid = 0;
+	sbuf_printf(sb, "%d\n", (check_uid || check_gid) ? 1 : 0);
+	return (0);
```
markjdb, line 2348: From the description here, it looks like this should just be looking at `check_uid`? https://docs.kernel.org/admin-guide/sysctl/fs.html#protected-hardlinks

sys/compat/linprocfs/linprocfs.c
ricardobranco777: > This should be GID_NOGROUP probably.

Just grep'ed both on the source tree. Makes sense.

markjdb (approved):
Looks good

spmzt:

PR #2160, opened 2026-04-23
ctld: Fix uninitialized variables

PHSpline:
```
/etc/ctl.conf allows shorthand "port ioctl/n" syntax, which currently
results in an uninitialized variable being used in the /dev node name:

target ... {
    port ioctl/1
}

root@vm-fbsd:/ # ls -1 /dev/cam
ctl
ctl1.-871264432     <--- Incorrect

With this fix, the variable is now initialized properly:

root@vm-fbsd:/ # ls -1 /dev/cam
ctl
ctl1.0        <--- Correct
```

### DIFF
--- a/usr.sbin/ctld/ctld.cc
+++ b/usr.sbin/ctld/ctld.cc
@@ -2642,7 +2642,7 @@ conf::add_pports(struct kports &kports)
 {
 	std::unordered_map<struct pport *, struct target *> linked_ports;
 	struct pport *pp;
-	int ret, i_pp, i_vp;
+	int ret;
 
 	for (auto &kv : conf_targets) {
 		struct target *targ = kv.second.get();
@@ -2677,6 +2677,7 @@ conf::add_pports(struct kports &kports)
 			 * If this port is an ioctl port, create a new
 			 * port.
 			 */
+			int i_pp{}, i_vp{};
 			ret = sscanf(pport.c_str(), "ioctl/%d/%d", &i_pp,
 			    &i_vp);
 			if (ret == 2) {


PHSpline:
@bsdjhb

PHSpline:
@bsdjhb - gentle ping

asomers:
How can those variables be uninitialized after `sscanf` returns 2?

PHSpline:
> How can those variables be uninitialized after `sscanf` returns 2?

Because when "port ioctl/1" is provided as the input string, sscanf bails part way through and does not overwrite i_vp, leaving it as its original (uninitialized) value.

asomers:
> > How can those variables be uninitialized after `sscanf` returns 2?
> 
> Because when "port ioctl/1" is provided as the input string, sscanf bails part way through and does not overwrite i_vp, leaving it as its original (uninitialized) value.

If sscanf bails part way through, shouldn't it return less than 2?

PHSpline:
> > > How can those variables be uninitialized after `sscanf` returns 2?
> > 
> > 
> > Because when "port ioctl/1" is provided as the input string, sscanf bails part way through and does not overwrite i_vp, leaving it as its original (uninitialized) value.
> 
> If sscanf bails part way through, shouldn't it return less than 2?

Aha, sneaky... I see the code was changed recently from checking if > 0 to checking if == 2... when I rebased, I did not notice/retest.  Good call-out.  So now, I guess the "port name/pp" syntax described in the man page is no longer supported, instead requiring the full "port name/pp/vp", which resolves the issue.

PHSpline:
No longer needed

PR #2161, opened 2026-04-24
dev/ichsmb: disable block buffer if supported

ckoehne:
In order to improve the efficiency of block read/write calls, Intel has introduced a block buffer. Instead of generating an interrupt after receiving/sending a single byte, the data is buffered in the block buffer. It allows the SMBus controller to generate a single interrupt for the whole transfer. At the moment, we don't support that and don't expect the SMBus controller to behave in that way. Unfortunately, BIOS code can also access the SMBus controller and may enable the block buffer. Poorly written BIOS code may also keep the block buffer enabled breaking our driver. Therefore, we should check if the device supports a block buffer and disable it for every request because we don't know if some BIOS code has reconfigured the SMBus controller in between.


emaste:
I think ichsmb.c should have a comment explaining why this is done (basically what is in the commit message)

ckoehne:
> I think ichsmb.c should have a comment explaining why this is done (basically what is in the commit message)

Added a comment.

emaste (approved):
I don't really like abusing `.driver_data` like this, but I guess it's ok.

PR #2162, opened 2026-04-25
nullfs + inotify: fix kernel panic when renaming over a watched file (bug 292495)

jlaundry:
Earlier I opened [Bugzilla #292495](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=292495), as after upgrading from 14.3 to 15.0, I found that trying to use git inside Linux podman containers resulted in a kernel panic.

It only seems to happen when mapping an encrypted ZFS dataset into the container, where not all files are owned by `root:wheel`, and was able to reproduce this reliably in a fresh 15.0 install.

**Disclaimer:** while I've been a user for ages, I've never done any FreeBSD kernel development before, and my C skills are very out of date.

Mostly out of curiosity, I put Claude to work on the problem, and after an hour of gentle coaxing, it came up with the attached patch and a brief description of the issue:

> The root cause was that `vop_inotify` is missing from the nullfs vops table. Because it wasn't registered, VOP_INOTIFY on a nullfs vnode dispatched to `vop_stdinotify` (the generic default) instead of `null_bypass`. `null_copy_inotify` correctly set `VIRF_INOTIFY` on the upper nullfs vnode to proxy events down to the lower ZFS vnode's watches — but without `null_bypass` handling VOP_INOTIFY, `vn_inotify` ran directly on the nullfs vnode, which has `VIRF_INOTIFY` set but `v_pollinfo = NULL`, causing the page fault at offset `0x18` (the `vpi_lock` field).

After confirming the fix worked across arm64 and x86_64 virtual machines (full repro steps are in Bugzilla), I then tried to get some tests created to verify the fix... and after ~6 hours of back-and-forth Claude got there, and the `inotify_nullfs_rename_clobber` test will reliably panic a 15.0.0-p6 install, while passing on a patched kernel.

_However_, the `inotify_nullfs_rename` test doesn't crash on either 15.0.0 or 15.0.0-p6, so I'm not sure if it's needed at all:

> **Tests** (inotify_test.c)
> 
> - `inotify_nullfs_rename` — regression test for the `tvp == NULL` rename path.  Watches a file via its lower path, opens it via nullfs to stamp `VIRF_INOTIFY` on the upper vnode, then renames it to a new name.  Verifies that `IN_MOVE_SELF` is delivered on patched kernels and that unpatched kernels panic.
> 
> - `inotify_nullfs_rename_clobber` — regression test for the `tvp != NULL` crash path (the primary crash scenario).  Watches both a source and a destination file via their lower paths, opens both via nullfs, then renames source over destination.  On an unpatched kernel this reliably panics via the `_IN_MOVE_DELETE → VOP_INOTIFY(upper_tvp)` path.  On a patched kernel it consumes the expected four events: `IN_MOVE_SELF` (src, lower-level rename), `IN_DELETE_SELF` (dst), `IN_IGNORED` (dst watch auto-removed), and a second `IN_MOVE_SELF` (src, nullfs-level rename post — a consequence of `vop_rename_post` firing at both the lower-fs and nullfs stack levels).

### DIFF
--- a/sys/fs/nullfs/null_vnops.c
+++ b/sys/fs/nullfs/null_vnops.c
@@ -627,6 +627,12 @@ null_access(struct vop_access_args *ap)
 }
 
 static int
+null_inotify(struct vop_inotify_args *ap)
+{
+	return (null_bypass(&ap->a_gen));
+}
+
+static int
 null_accessx(struct vop_accessx_args *ap)
 {
 	struct vnode *vp = ap->a_vp;
@@ -1235,6 +1241,7 @@ struct vop_vector null_vnodeops = {
 	.vop_getattr =		null_getattr,
 	.vop_getlowvnode =	null_getlowvnode,
 	.vop_getwritemount =	null_getwritemount,
+	.vop_inotify =		null_inotify,
 	.vop_inactive =		null_inactive,
 	.vop_need_inactive =	null_need_inactive,
 	.vop_islocked =		vop_stdislocked,
--- a/sys/kern/vfs_inotify.c
+++ b/sys/kern/vfs_inotify.c
@@ -672,6 +672,9 @@ inotify_log(struct vnode *vp, const char *name, size_t namelen, int event,
 	KASSERT((event & ~(IN_ALL_EVENTS | IN_ISDIR | IN_UNMOUNT)) == 0,
 	    ("inotify_log: invalid event %#x", event));
 
+	if (__predict_false(vp->v_pollinfo == NULL))
+		return;
+
 	mtx_lock(&vp->v_pollinfo->vpi_lock);
 	TAILQ_FOREACH_SAFE(watch, &vp->v_pollinfo->vpi_inotify, vlink, tmp) {
 		KASSERT(watch->vp == vp,
--- a/tests/sys/kern/inotify_test.c
+++ b/tests/sys/kern/inotify_test.c
@@ -17,6 +17,7 @@
 #include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <glob.h>
 #include <limits.h>
 #include <mntopts.h>
 #include <stdio.h>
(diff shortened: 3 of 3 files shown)


jlaundry:
@markjdb if you could please take a look 🙂

markjdb:
I'm sorry the original bugzilla report didn't get looked at. :(

The patch isn't quite wrong (though I'm pretty skeptical that implementing vop_inotify in nullfs accomplishes anything), but really the root cause is that we're calling VOP_INOTIFY on a vnode that doesn't expect to handle it, and rather than making inotify_log() handle this special case it's better for nullfs to make sure that that case doesn't arise at all. I'd rather that vn_inotify() and inotify_log() not have to reason about what state the vnode is in.

The test case can be a fair bit simpler and the comments are pedantic and slightly wrong. I just wrote my own: https://reviews.freebsd.org/D56639

(Not to knock on claude, I think the VFS is particularly hard to understand, there's a lot of magic and generated code which obscures what's actually going on, especially with things like nullfs bypass, so source code analysis is going to miss some things without special prompting.)

jlaundry:
> I'm sorry the original bugzilla report didn't get looked at. :(

No worries at all! I had a workaround (having files owned by root), so I figured this might have been a 'just me' issue.

And thank you for the quick reply! I've just tested D56639, and confirmed it fixes my issue. I'll close this now, unless there's anything else you needed from me?

> (Not to knock on claude

I'm sure the machine won't be upset (for now...)

This was a fun learning experience, and I'm going to continue to iterate on the prompts to see if I can get it to the right answer, quicker; and in particular getting it to write a test that fails without crashing the system.

markjdb:
> > I'm sorry the original bugzilla report didn't get looked at. :(
> 
> No worries at all! I had a workaround (having files owned by root), so I figured this might have been a 'just me' issue.
> 
> And thank you for the quick reply! I've just tested D56639, and confirmed it fixes my issue. I'll close this now, unless there's anything else you needed from me?

Nope, let's close it.

> > (Not to knock on claude
> 
> I'm sure the machine won't be upset (for now...)
> 
> This was a fun learning experience, and I'm going to continue to iterate on the prompts to see if I can get it to the right answer, quicker; and in particular getting it to write a test that fails without crashing the system.

Given my experience with claude I'm a bit skeptical that it'll get to what I consider the right solution very easily. But having it generate a self-contained test case from a crash dump is extremely valuable.

PR #2164, opened 2026-04-26
Dynamic jail.conf

Crest:
Jails often have repetitive boilerplate that can't be factored out of `jail.conf` using `.include "<glob>";` or requires external lookups that require wrappers around [jail(8)](https://man.freebsd.org/cgi/man.cgi?jail) e.g. network interface creation or IP allocation, database lookups.

This patch checks if a [jail.conf(5)](https://man.freebsd.org/cgi/man.cgi?jail.conf) is executable by trying to reopen its file descriptor for execution in which case it runs it as child process of the jail command and parses the child's standard output.

This is a continuation of [D46284](https://reviews.freebsd.org/D46284), because Phabricator fails to update the diff to multiple files. Sorry for the duplication.

### DIFF
--- a/usr.sbin/jail/config.c
+++ b/usr.sbin/jail/config.c
@@ -331,15 +331,15 @@ open_config(const char *cfname, pid_t *const pid)
 	int read_fd, write_fd, exec_fd, pipes[2];
 	FILE *file;
 
-	/* Invalidate the child PID. */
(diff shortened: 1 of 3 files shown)


Crest:
These tests failed because the project ran out of compute budget and my pull request adds nothing that increases the build time by more than a few milliseconds.

PR #2168, opened 2026-04-27
kern_dump.c: fix progress output

mchoo7:
In ideal case, this should print multiple of tens, i.e. ..0%..10%..20%..30% and so on. However, those numbers fall into buckets that are previously visited, which results to ..0%..11%..21%..31% in real case scenario. Adjust bucket range to print multiple of tens unless they are skipped otherwise.


gmshake:
This is a cosmetic change rather than a fix. I think it is more valuable to rework `progress_track` ? Ideally a single integer is sufficient to indicate the progress.

PR #2169, opened 2026-04-28
bhyve: allow read/write to full CRB buffer

ckoehne:
For some reason, we've incorrectly calculated the size of the CRB data buffer register. There's no need to divide the CRB data buffer size by 4. We should allow access to the whole buffer instead.

### DIFF
--- a/usr.sbin/bhyve/tpm_intf_crb.c
+++ b/usr.sbin/bhyve/tpm_intf_crb.c
@@ -414,7 +414,7 @@ tpm_crb_mem_handler(struct vcpu *vcpu __unused, const int dir,
 		    4:
 		case offsetof(struct tpm_crb_regs,
 		    data_buffer) ... offsetof(struct tpm_crb_regs, data_buffer) +
-		    TPM_CRB_DATA_BUFFER_SIZE / 4:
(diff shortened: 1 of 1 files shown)


usr.sbin/bhyve/tpm_intf_crb.c
```diff
@@ -414,7 +414,7 @@ tpm_crb_mem_handler(struct vcpu *vcpu __unused, const int dir,
 		    4:
 		case offsetof(struct tpm_crb_regs,
 		    data_buffer) ... offsetof(struct tpm_crb_regs, data_buffer) +
-		    TPM_CRB_DATA_BUFFER_SIZE / 4:
+		    TPM_CRB_DATA_BUFFER_SIZE:
```
markjdb: Or write `sizeof(((struct tpm_crb_regs *)NULL)->data_buffer)`?

And shouldn't we subtract 1? We are allowing writes past the end of the buffer, it looks like, since the `...` ranges are inclusive. (I see there is some bounds checking on the offset at the beginning of the function, but still.)

ckoehne: I don't have a strong preference.

Thanks for catching the out of bounds write.

markjdb (approved):

PR #2170, opened 2026-04-28
ifconfig: Add Wi-Fi MAC address randomization

sanesecurityguy:
Allow the "random" keyword to be used with "wlanaddr" to generate a random MAC address when creating a network pseudo-device.


sanesecurityguy:
Oh, thanks, @spmzt. I thought I'd still have to send an email or something to get a review.

I've done functional tests on my end based on both tag release/15.0.0-p6 and branch releng/15.0 (branches main and stable wouldn't compile for unrelated reasons). In both cases, the new `random` keyword works as expected and as shown here:
[functional-test.txt](https://github.com/user-attachments/files/27204216/functional-test.txt)

sanesecurityguy:
Most of that code is ported from af_link.c here:
https://github.com/freebsd/freebsd-src/blob/release/15.0.0-p6/sbin/ifconfig/af_link.c#L191-L222

So it's not actually DRY, but this is the first repetition I'm aware of.

erikarn:
hi! which NIC did you test this against?

Not all of the wifi NICs allow random/arbitrary MAC address assignments. That's why I'm curious.

sanesecurityguy:
@erikarn This one:
```
$ pciconf -lv pci0:3:0:0
rtw880@pci0:3:0:0:      class=0x028000 rev=0x00 hdr=0x00 vendor=0x10ec device=0xc822 subvendor=0x1a3b subdevice=0x3750
    vendor     = 'Realtek Semiconductor Co., Ltd.'
    device     = 'RTL8822CE 802.11ac PCIe Wireless Network Adapter'
    class      = network
```

sbin/ifconfig/ifconfig.8
lwhsu: Needs to be updated when a non-trivial change to the manual page.

sbin/ifconfig/ifieee80211.c
```diff
@@ -6115,13 +6115,27 @@ set80211clone_wlanbssid(if_ctx *ctx __unused, const char *arg, int dummy __unuse
...
-	params.icp_flags |= IEEE80211_CLONE_MACADDR;
+	if (!strcmp(arg, "random")) {
+		sdl.sdl_len = sizeof(sdl);
+		sdl.sdl_alen = IEEE80211_ADDR_LEN;
+		sdl.sdl_nlen = 0;
+		sdl.sdl_family = AF_LINK;
+		arc4random_buf(&sdl.sdl_data, IEEE80211_ADDR_LEN);
+		/* Non-multicast and claim it is locally administered. */
+		sdl.sdl_data[0] &= 0xfc;
+		sdl.sdl_data[0] |= 0x02;
```
lwhsu, line 6129: I didn't know `wlanaddr` before, what I usually do is `ifconfig wlan0 ether random` after wlan0 has been created.

I am thinking if this generating random MAC part (sbin/ifconfig/af_link.c:198) can be extracted as a helper like:

```
void
random_macaddr(uint8_t mac[ETHER_ADDR_LEN])
{
        arc4random_buf(mac, ETHER_ADDR_LEN);
        mac[0] &= 0xfc; /* clear multicast and global bits */
        mac[0] |= 0x02; /* locally administered */
}
```

while indeed that `IEEE80211_ADDR_LEN` and `ETHER_ADDR_LEN` are both 6 but different in concept...

erikarn:
I think we need to have a bit of a think about it.

Ideally we'd not select a random address in ifconfig but we'd set a
flag (like the CLONE mac addr flag) that says "clone me a random
address".
And then net80211 / driver can choose what's sufficiently random and
supported, and fail if setting the MAC address directly isn't
supported.




-adrian

bzfbd:
I wonder which of the native drivers does not support setting the mac address while the interface is down given the above is basically what we've been doing by hand for two decades?  That said things have moved on and modern drivers do support this one or another way or still not at all or even more.

So I'd go even further in that the drivers should advertise this as a capability as they may set the address themselves (if net80211 doesn't do it);  at least a second bit would be needed to say if scanning with a random address is supported / should happen.  I'd have to go and see if there are even more "random" bits in the hand full of drivers I try to support.

But the changes are hard to make currently without breaking things before we start on massive userland-kernel changes (and for that we need to fix a dozen or so of major ports to use lib80211 or everyone will be cranky when their browser constantly doesn't build anymore).

I am not thinking about declining the above, I am more worried that the "random" keyword in the future will change in implementation and am cautious about that.

sbin/ifconfig/ifieee80211.c
sanesecurityguy: > what I usually do is ifconfig wlan0 ether random after wlan0 has been created.

Hmmm, either that doesn't work anymore or it doesn't work for just me. For me, running `ifconfig wlan0 ether random` on a working connection makes the connection stop working, even though `ifconfig wlan0` shows both the randomized MAC address in `ether` and the real MAC address in `hwaddr`. When creating the device with `ifconfig wlan0 create wlandev rtw880 ether random`, `ether` and `hwaddr` still show up correctly but the connection never works from the beginning.

`ifconfig` also complains if I try to change `wlanaddr` after creation:
```
# ifconfig wlan0 | grep ether
        ether e0:b5:2d:9e:c8:30
# ifconfig wlan0 wlanaddr "e0:b5:2d:9e:c8:31"
ifconfig: wlanaddr: bad value
```

That's why I ended up with `wlanaddr` during `create`.

A helper function like `random_macaddr` is probably a good idea. Where do you think would be the best place to put it? lib/libifconfig/libifconfig.c, perhaps?

sbin/ifconfig/ifconfig.8
sanesecurityguy: Done.

sanesecurityguy:
> (like the CLONE mac addr flag)

Where is the code for this so I can see how it works? Is it the parts mentioning `c_iscloneop` in ifconfig.c and ifconfig.h?

erikarn:
ok, there's a bunch of things we likely need to just go create PRs for and
chase down.

* yeah we should find out how / why we can change a vap mac address during
runtime and deny / fail that properly.
* the wifi driver attach puts the firmware/EEPROM MAC into the ieee80211com
struct (ic_macaddr), and that gets used for cloning/drivers.
* .. but creating a VAP with a random / different MAC doesn't guarantee the
firmware is going to be happy with it.
* the reason why creating VAPs with the that CLONE flag is to push the VAP
unique mac address configuration down into the NIC. The driver is the only
thing that knows what's an OK variant on MAC addresses - what's legal, what
minimises the bitmask used to match on receive frames, etc.
* .. and the reason for THAT is that you can't necessarily just go choose
random mac addresses if the hardware uses a single MAC + mask, since some
(atheros for sure) will actually generate ACKs in hardware to frames that
match!
* Creating the VAP via "create wlandev <x> ether random" changes the normal
ethernet address but likely isn't updating the VAP ethernet address and
isn't communicated through the VAP create method and stuff.
* And yeah, I'm aligned with Bjorn on more capability flags - especially
for "use a random mac address during scanning if you can."

Eg I'll take a recent native driver - mtw(4). /it/ does the right thing
currently because it's calling mtw_set_macaddr()  based on either the VAP
MAC address or the MAC address. But the driver only supports one VAP right
now and it doesn't look like the firmware for that cares.

For iwx(4), I know you pass in a MAC address when you create a mac context,
and that comes from the bss node MAC address. Which at that point should
match the VAP address.

For ath(4), this doesn't work - the MAC address comes from EEPROM, and will
always configure that as the base MAC address. It's not choosing one based
on the VAP MAC address. So we'd need driver changes here to support it when
operating in STA only mode.

Give me a couple days to pull this apart. I think it's worth finally filing
PRs and fixing "try to set the mac address when the vap is up" and "ether
<x> should work or fail properly" to begin with, and then figure out some
capability bits to teach drivers about it.

(And don't get me started on the limitations for random macs for scanning -
those also often can't be completely random because again there's only one
or two MACs programmable in hardware/firmware, everything else typically
involves some mask based matching, so iirc A Couple Of Well Known Vendors
didn't REALLY randomize the masks in the past as it would have caused
dropped traffic..)

erikarn:
So, look at IEEE80211_CLONE_* in sys/net80211/ieee80211_ioctl.h,
notably IEEE80211_CLONE_BSSID.

IEEE80211_CLONE_MACADDR is the hint to net80211/driver to use the
provided mac address rather than the device provided ic->ic_macaddr.
Look at ieee80211_freebsd.c:wlan_clone_create().
Drivers should be checking this to see if they need to explicitly be
setting the MAC address, but most of them just have a conditional to
use either vap->iv_myaddr or ic->ic_macaddr, which only works for STA.

For hostap operation, IEEE80211_CLONE_BSSID says "please create me a
unique MAC address that the hardware can use and return that". It
creates a new MAC address and passes that to ieee80211_vap_attach() to
complete attach with the updated MAC address.

It pushes the "assign me a unique address" thing into the driver
because that's the only one that knows what can be done.

Anyway, this is a lot of beard shaving to get this "correct-er". I
think for net80211 drivers we can start by adding some more capability
bits and enforce them. Then I think we should pushing the random
assignment into net80211 via an ioctl API flag, so we don't have to
worry about trying to do something the hardware/firmware isn't happy
with. Then userland shouldn't need changing, and we can worry about
cleaning up the drivers.


-adrian

sanesecurityguy:
I finally found time to look into this again and, while I'm still trying to wrap my head around it all, I think I finally figured out the answer to this question:
> * yeah we should find out how / why we can change a vap mac address during runtime and deny / fail that properly.

The assumption is:
> * the wifi driver attach puts the firmware/EEPROM MAC into the ieee80211com struct (ic_macaddr), and that gets used for cloning/drivers.

But it turns out that the rtw88 driver, which is ported from Linux, just naively writes the VAP's MAC address into the firmware's MAC address space, completely bypassing net80211.

Here are the relevant snippets of code:
- [rtw88/main.c](https://github.com/freebsd/freebsd-src/blob/e3fa020202d8e040242016bda275dde83c058549/sys/contrib/dev/rtw88/main.c#L2412)
- [SET_IEEE80211_PERM_ADDR](https://github.com/freebsd/freebsd-src/blob/e3fa020202d8e040242016bda275dde83c058549/sys/compat/linuxkpi/common/include/net/mac80211.h#L2299-L2303)
- [ether_addr_copy](https://github.com/freebsd/freebsd-src/blob/e3fa020202d8e040242016bda275dde83c058549/sys/compat/linuxkpi/common/include/linux/etherdevice.h#L88-L91)

I also managed to track down the data structures of the args to `SET_IEEE80211_PERM_ADDR` if you really wanna' look at that:
- [ieee80211_hw](https://github.com/freebsd/freebsd-src/blob/e3fa020202d8e040242016bda275dde83c058549/sys/compat/linuxkpi/common/include/net/mac80211.h#L479-L481)
- [wiphy](https://github.com/freebsd/freebsd-src/blob/e3fa020202d8e040242016bda275dde83c058549/sys/compat/linuxkpi/common/include/net/cfg80211.h#L1175)
- [rtw_dev](https://github.com/freebsd/freebsd-src/blob/e3fa020202d8e040242016bda275dde83c058549/sys/contrib/dev/rtw88/main.h#L2073)
- [efuse](https://github.com/freebsd/freebsd-src/blob/e3fa020202d8e040242016bda275dde83c058549/sys/contrib/dev/rtw88/rtw8822c.c#L26-L42)

So it looks like rtw88, and other Linux drivers, aren't using the `ieee80211com` struct at all.
```
user@localhost:/usr/src/sys/contrib/dev/rtw88 $ grep -rl ieee80211com
user@localhost:/usr/src/sys/contrib/dev/rtw88 $ cd..
user@localhost:/usr/src/sys/contrib/dev $ grep -rl ieee80211com
user@localhost:/usr/src/sys/contrib/dev $
```

What I haven't been able to figure out yet is when exactly `SET_IEEE80211_PERM_ADDR` is being called at runtime. As far as I can tell, it's being called [only during the driver probe phase](https://github.com/freebsd/freebsd-src/blob/e3fa020202d8e040242016bda275dde83c058549/sys/contrib/dev/rtw88/pci.c#L1891) but, unless I'm misunderstanding something, that doesn't explain why my [functional test](https://github.com/user-attachments/files/27204216/functional-test.txt) shows me being able to change the firmware MAC address at runtime. And I know my test is actually showing what `ifconfig` thinks is the unified VAP/hardware MAC address, but even my Wi-Fi router shows the new randomized MAC address connected to the network after I change it at runtime.

erikarn:
On Fri, 5 Jun 2026 at 07:34, Lambert ***@***.***> wrote:

>
Well yeah, they're linuxkpi drivers, the linuxkpi code has the net80211
bits in it. they hide it from the linuxkpi'ed wifi drivers.


> What I haven't been able to figure out yet is when exactly
> SET_IEEE80211_PERM_ADDR is being called at runtime. As far as I can tell,
> it's being called only during the driver probe phase
> <https://github.com/freebsd/freebsd-src/blob/e3fa020202d8e040242016bda275dde83c058549/sys/contrib/dev/rtw88/pci.c#L1891>
> but, unless I'm misunderstanding something, that doesn't explain why my functional
> test
> <https://github.com/user-attachments/files/27204216/functional-test.txt>
> shows me being able to change the firmware MAC address at runtime. And I
> know my test is actually showing what ifconfig thinks is the unified
> VAP/hardware MAC address, but even my Wi-Fi router shows the new randomized
> MAC address connected to the network after I change it at runtime.
>
>
It is likely changing via code in sys/compat/linuxkpi/ somewhere.

This is why I'd like to just make sure we've documented this and added
actual capabilities. I've been bitten by this in the past
and I don't want to enable / ship this and then debug it across a whole lot
of random NICs that don't quite support it in the
way you're hoping. :-)



-adrian

sanesecurityguy:
> Well yeah, they're linuxkpi drivers, the linuxkpi code has the net80211 bits in it. they hide it from the linuxkpi'ed wifi drivers.
> It is likely changing via code in sys/compat/linuxkpi/ somewhere.

Ah, got it. Found 'em in sys/compat/linuxkpi/common/src. Will look at them more closely some other time.

> I've been bitten by this in the past and I don't want to enable / ship this and then debug it across a whole lot of random NICs that don't quite support it in the way you're hoping. :-)

Yes, yes, I understand and trust your judgement!

sanesecurityguy:
> It is likely changing via code in sys/compat/linuxkpi/ somewhere.

I think I found it here in sys/compat/linuxkpi/common/src/linux_80211.c:
[linuxkpi_ieee80211_ifattach](https://github.com/freebsd/freebsd-src/blob/e3fa020202d8e040242016bda275dde83c058549/sys/compat/linuxkpi/common/src/linux_80211.c#L7100-L7102)

bzfbd:
> > It is likely changing via code in sys/compat/linuxkpi/ somewhere.
> 
> I think I found it here in sys/compat/linuxkpi/common/src/linux_80211.c: [linuxkpi_ieee80211_ifattach](https://github.com/freebsd/freebsd-src/blob/e3fa020202d8e040242016bda275dde83c058549/sys/compat/linuxkpi/common/src/linux_80211.c#L7100-L7102)

No.  What you found is setting the default HW address to the net80211 com.

For what you want there's various bits to this:

(a) random scanning https://github.com/freebsd/freebsd-src/blob/e3fa020202d8e040242016bda275dde83c058549/sys/compat/linuxkpi/common/src/linux_80211.c#L4838  (unimplemented as we have no way to pass this from user space to net80211 to LinuxKPI 802.11 currently; other work is needed on the user space API side first).

(b) Manually setting of a (random) mac address is documented in https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=277356#c4 and https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=277356#c5  ( I should really go and fix that in LinuxKPI 802.11 ).

But then again is you do what @lwhsu said in the code review comment that's really all that's needed.  Making it an ifconfig command line shortcut is likely independent of 802.11.  If wired or wireless HW does not support setting the "ether address" then it's a shoot yourself in the foot operation.  Otherwise we'd need a flag in each driver indication that this is supported but in the last 30 years no one bothered.

sanesecurityguy:
> No. What you found is setting the default HW address to the net80211 com.

Ah, crap, you're right. I got distracted and went off on a tangent there.

This is where it's really set and where the MAC address set by my code is actually used:
https://github.com/freebsd/freebsd-src/blob/e3fa020202d8e040242016bda275dde83c058549/sys/net80211/ieee80211_freebsd.c#L154-L157

That also means that rtw88 - and likely the other Linux drivers - is indeed using the `ieee80211com` struct after all.

And now it's occurred to me that what you guys mean by "runtime" might be different from what I thought. I was thinking "runtime" was after the driver attaches to the device, but what you guys mean by "runtime" might be after the VAP gets created or attached.

erikarn:
Would you mind doing me a favour?

Let's take a step back and think about the full design goals and
implementation details of this?
That way I/others can figure out what the right set of components are to go
add bits to.

What timezone are you in? Maybe a call or something to flesh it out?


-a

sbin/ifconfig/ifieee80211.c
sanesecurityguy: Turns out we already have a couple of functions for this in the linuxkpi:
https://github.com/freebsd/freebsd-src/blob/e3fa020202d8e040242016bda275dde83c058549/sys/compat/linuxkpi/common/include/linux/etherdevice.h#L118-L132

bzfbd: > Turns out we already have a couple of functions for this in the linuxkpi:
> 
> [freebsd-src/sys/compat/linuxkpi/common/include/linux/etherdevice.h](https://github.com/freebsd/freebsd-src/blob/e3fa020202d8e040242016bda275dde83c058549/sys/compat/linuxkpi/common/include/linux/etherdevice.h#L118-L132)
> 
> Lines 118 to 132 in [e3fa020](/freebsd/freebsd-src/commit/e3fa020202d8e040242016bda275dde83c058549)

And how do these functions help you in user space?

I am really sorry but I have to follow what Adrian said in this case.
Let's take a step back and think about design and what we want to accomplish rather than bouncing around between user space ifconfig and kernel space net80211 and drivers and a compat layer.

sanesecurityguy: > And how do these functions help you in user space?

I was just documenting its existence here since the question of a helper function came up.

> I am really sorry but I have to follow what Adrian said in this case.
> Let's take a step back and think about design...

Yes, I already replied to him that that sounds good to me, and then I deleted my reply to him a little over 24 hours later since I'm sure you both got an emailed copy of it and I just didn't want my email address to be written here unnecessarily.

Haven't heard from Adrian yet, and I've found email addresses for you both, so I guess I'll send you guys an email.

sanesecurityguy:
[This is a new comment replacing the old deleted comment]

> What timezone are you in? Maybe a call or something to flesh it out?

Yes, sounds good to me. Emailing you, Bjoern, and Li-Wen to arrange it.

PR #2175, opened 2026-04-29
linux: implement setfsuid(2) and setfsgid(2) as no-ops

ricardobranco777:
These system calls exist to decouple the Linux filesystem credentials from the effective credentials, avoiding signal exposure during privilege transitions.  The signal permission model that motivated this was revised in Linux 2.0, making these syscalls obsolete for new applications.

Implement both syscalls as no-ops that return the current effective UID/GID as the previous filesystem UID/GID.  Linux returns the previous filesystem UID/GID for these syscalls, with no error indication.

This silences spurious "syscall setfsxid not implemented" from programs such as bash that call them during session setup.

Fixes https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=294879

### DIFF
--- a/sys/compat/linux/linux_misc.c
+++ b/sys/compat/linux/linux_misc.c
@@ -3179,4 +3179,30 @@ linux_membarrier(struct thread *td, struct linux_membarrier_args *args)
 	return (0);
 }
 
+/*
+ * setfsuid() & setfsgid() exist to decouple the Linux filesystem credentials
+ * from the effective credentials, avoiding signal exposure during privilege
+ * transitions. The signal permission model that motivated this was revised in
+ * Linux 2.0, making these syscalls obsolete for new applications.
+ *
+ * As there's no FreeBSD equivalent, implement both syscalls as no-ops that
+ * return the current effective UID/GID as the previous filesystem UID/GID.
+ * Linux returns the previous filesystem UID/GID for these syscalls, with no
+ * error indication.
+ */
+
+int
+linux_setfsuid(struct thread *td, struct linux_setfsuid_args *args)
+{
+	td->td_retval[0] = td->td_ucred->cr_uid;
+	return (0);
+}
+
+int
+linux_setfsgid(struct thread *td, struct linux_setfsgid_args *args)
+{
+	td->td_retval[0] = td->td_ucred->cr_gid;
+	return (0);
+}
+
 MODULE_DEPEND(linux, mqueuefs, 1, 1, 1);
--- a/sys/compat/linux/linux_dummy.c
+++ b/sys/compat/linux/linux_dummy.c
@@ -56,8 +56,6 @@ UNIMPLEMENTED(query_module);	/* Added in Linux 2.2 removed in 2.6. */
 UNIMPLEMENTED(security);
 UNIMPLEMENTED(vserver);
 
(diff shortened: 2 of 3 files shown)


kostikbel (commented):
I think that at least a comment is due that there is not direct equivalent in the FreeBSD land and we implement the syscalls as nop.

ricardobranco777:
> I think that at least a comment is due that there is not direct equivalent in the FreeBSD land and we implement the syscalls as nop.

Agree. Added a comment.

sys/compat/linux/linux_misc.c
```diff
@@ -3179,4 +3179,29 @@ linux_membarrier(struct thread *td, struct linux_membarrier_args *args)
 	return (0);
 }
 
+/* setfsuid & setfsgid exist to decouple the Linux filesystem credentials
```
kostikbel: Fix the style please.  Also we traditionally use function() for reference to a function.

sys/compat/linux/linux_misc.c
```diff
@@ -3179,4 +3179,29 @@ linux_membarrier(struct thread *td, struct linux_membarrier_args *args)
...
+ * from the effective credentials, avoiding signal exposure during privilege
```
kostikbel, line 3184: BTW, do you mean signal(3)?

sys/compat/linux/linux_misc.c
ricardobranco777: Fixed.

sys/compat/linux/linux_misc.c
ricardobranco777: > BTW, do you mean signal(3)?

No.

kostikbel: Then what are 'signal' there?

BTW, the style for the comment is still not fixed.

ricardobranco777: > Then what are 'signal' there?

Signal exposure.  Signal is generic here to mean the standard UNIX signals.  What do you suggest?

> BTW, the style for the comment is still not fixed.

It should be fixed now.

kostikbel: So it is signal(3) after all.  It was not obvious from the text and your response contradicted.

BTW, aren't userspace file servers like samba and ganesha use these syscalls?

ricardobranco777: > So it is signal(3) after all. It was not obvious from the text and your response contradicted.

`signal(3)` is the libc function call that installs a **signal** handler.  We care about the **signal** concept, not the call.  Using `signal(3) exposure` here would be wrong.

> BTW, aren't userspace file servers like samba and ganesha use these syscalls?

I cloned samba and I don't see it called in the master branch.  ganesha-nfs calls `glfs_setfsuid` from glusterfs, and it does nothing special other than storing the value.

The system call is weird because the manpage says:

> On both success and failure, this call returns the previous filesystem user ID of the caller

kostikbel: signal(3) is what gives the (somewhat naive) overview of the signals on FreeBSD.

But ok.

kostikbel (approved):

spmzt (approved):

spmzt:

PR #2178, opened 2026-05-02
savecore: Create core.txt.last symlink

ricardobranco777:
When saving a coredump, savecore(8) maintains `.last` symlinks for info and vmcore artifacts, but not for the crashinfo text report.  Create it to point at the current `core.txt.<bounds>` file.

This makes /var/crash/core.txt.last track the same dump as info.last and vmcore.last.

Now I have `core.txt.last`:

```
$ ls -l /var/crash/
total 2689423
-rw-r--r--  1 root wheel          2 May  2 19:11 bounds
-rw-r--r--  1 root wheel        443 Apr 18 11:27 core.txt.0
-rw-r--r--  1 root wheel        443 Apr 18 11:37 core.txt.1
-rw-r--r--  1 root wheel     146891 Apr 18 11:43 core.txt.2
lrwxr-xr-x  1 root wheel         10 May  2 19:11 core.txt.last -> core.txt.2
-rw-------  1 root wheel        474 Apr 18 11:27 info.0
-rw-------  1 root wheel        474 Apr 18 11:37 info.1
-rw-------  1 root wheel        474 Apr 18 11:43 info.2
lrwxr-xr-x  1 root wheel          6 May  2 19:11 info.last -> info.2
-rw-r--r--  1 root wheel          5 Aug  8  2025 minfree
-rw-------  1 root wheel  854790144 Apr 18 11:27 vmcore.0
-rw-------  1 root wheel  854290432 Apr 18 11:37 vmcore.1
-rw-------  1 root wheel  853876736 Apr 18 11:43 vmcore.2
lrwxr-xr-x  1 root wheel          8 May  2 19:11 vmcore.last -> vmcore.2
```

### DIFF
--- a/sbin/savecore/savecore.c
+++ b/sbin/savecore/savecore.c
@@ -411,6 +411,7 @@ symlinks_remove(int savedirfd)
 
 	(void)unlinkat(savedirfd, "info.last", 0);
 	(void)unlinkat(savedirfd, "key.last", 0);
+	(void)unlinkat(savedirfd, "core.txt.last", 0);
 	(void)unlinkat(savedirfd, "vmcore.last", 0);
 	(void)unlinkat(savedirfd, "vmcore.last.gz", 0);
 	(void)unlinkat(savedirfd, "vmcore.last.zst", 0);
@@ -901,6 +902,12 @@ DoLiveFile(const char *savedir, int savedirfd, const char *device)
 		    savedir, "info.last");
 	}
 
+	snprintf(linkname, sizeof(linkname), "core.txt.%d", bounds);
+	if (symlinkat(linkname, savedirfd, "core.txt.last") == -1) {
+		logmsg(LOG_WARNING, "unable to create symlink %s/%s: %m",
+		    savedir, "core.txt.last");
+	}
+
 	snprintf(linkname, sizeof(linkname), "livecore.last");
 	if (compress)
 		strcat(linkname, kdhl.compression == KERNELDUMP_COMP_ZSTD ?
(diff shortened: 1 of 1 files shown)


mhorne (changes requested):
I think the idea is great. 

However this is not the right place to implement it. The `core.txt` files are generated by the `/usr/sbin/crashinfo` shell script, not by `savecore` itself. 

So the symlink should be generated there.

ricardobranco777:
> I think the idea is great.
> 
> However this is not the right place to implement it. The `core.txt` files are generated by the `/usr/sbin/crashinfo` shell script, not by `savecore` itself.
> 
> So the symlink should be generated there.

I took a look at crashinfo and it doesn't generate any symlinks at all.  All the symlinks are being generated by savecore.

mhorne:
> > I think the idea is great.
> > However this is not the right place to implement it. The `core.txt` files are generated by the `/usr/sbin/crashinfo` shell script, not by `savecore` itself.
> > So the symlink should be generated there.
> 
> I took a look at crashinfo and it doesn't generate any symlinks at all. All the symlinks are being generated by savecore.

That's correct. But savecore creates symlinks to files that it owns/generates. savecore does not know about `core.txt.N`.

Here is a quick demonstration of the inconsistency (your change is applied):

```
# mkdir /tmp/newdir
# savecore -L /tmp/newdir
savecore 20213 - - livedump
savecore 20213 - - writing core to /tmp/newdir/livecore.0
# ls -l /tmp/newdir
total 1332821
-rw-r--r--  1 root wheel          2 May 11 11:19 bounds
lrwxr-xr-x  1 root wheel         10 May 11 11:19 core.txt.last -> core.txt.0
-rw-------  1 root wheel        493 May 11 11:19 info.0
lrwxr-xr-x  1 root wheel          6 May 11 11:19 info.last -> info.0
-rw-------  1 root wheel 4357517312 May 11 11:19 livecore.0
lrwxr-xr-x  1 root wheel         10 May 11 11:19 livecore.last -> livecore.0
```

It has created a symlink to `core.txt.0`, but this file does not exist. Not unless I run `crashdump`. By default, this is done automatically by `rc(8)`, but is a distinct operation.

So, it belongs in crashdump.sh, and should be as simple as:
```
LINK=$CRASHDIR/core.txt.last
ln -s $FILE $LINK
```

ricardobranco777:
> So, it belongs in crashdump.sh, and should be as simple as:
> 
> ```
> LINK=$CRASHDIR/core.txt.last
> ln -s $FILE $LINK
> ```

You're right.  PTAL https://github.com/freebsd/freebsd-src/pull/2199

PR #2180, opened 2026-05-06
hwpmc: Promote hwpmc EXTERROR diagnostics

afscoelho:
Hi - this promotes the hwpmc extended-error work into upstream. Highlights:

- give the AMD core and IBS PMCs better extended-error reporting so users actually see the misconfig cases that come up
- add a root-only Kyua check (`hwpmc_exterr_test`) that walks through the typical negative scenarios

## Testing
- `make -j${NCPU} buildworld`
- `make -j${NCPU} buildkernel KERNCONF=${KERNCONF}`
- `kyua test -k /usr/tests/Kyuafile hwpmc_exterr_test`

### DIFF
--- a/tests/sys/kern/hwpmc_exterr_test.c
+++ b/tests/sys/kern/hwpmc_exterr_test.c
@@ -1,0 +1,460 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2026 Advanced Micro Devices, Inc.
+ */
+
+#include <sys/param.h>
+#include <sys/exterrvar.h>
+#include <sys/module.h>
+#include <sys/pmc.h>
+
+#include <dev/hwpmc/hwpmc_amd.h>
+#include <dev/hwpmc/hwpmc_ibs.h>
+
+#include <atf-c.h>
+#include <errno.h>
+#include <exterr.h>
+#include <pmc.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+
+static int
+hwpmc_syscall_num(void)
+{
+	struct module_stat ms;
+	int modid;
+
+	modid = modfind(PMC_MODULE_NAME);
+	if (modid < 0)
+		return (-1);
+
+	ms.version = sizeof(ms);
+	if (modstat(modid, &ms) < 0)
+		return (-1);
+
+	return (ms.data.intval);
+}
+
+static int
+hwpmc_call(int op, void *arg)
+{
+	int sc;
+
+	sc = hwpmc_syscall_num();
+	if (sc < 0)
+		return (-1);
+
+	return (syscall(sc, op, arg));
+}
+
+static void
+clear_exterr(void)
+{
+	char exterr[UEXTERROR_MAXLEN];
+
+	ATF_REQUIRE_ERRNO(EINVAL, exterrctl(EXTERRCTL_UD, 0, NULL) == -1);
+	ATF_REQUIRE_EQ(0, uexterr_gettext(exterr, sizeof(exterr)));
+	ATF_REQUIRE_STREQ("", exterr);
+}
+
+static void
+require_exterr(const char *needle)
+{
+	char exterr[UEXTERROR_MAXLEN];
+
+	ATF_REQUIRE_EQ(0, uexterr_gettext(exterr, sizeof(exterr)));
+	ATF_REQUIRE_MSG(strstr(exterr, needle) != NULL,
+	    "extended error \"%s\" does not contain \"%s\"", exterr, needle);
+}
+
+static void
+require_hwpmc(void)
+{
+
+	if (pmc_init() == 0)
+		return;
+
+	atf_tc_skip("hwpmc is unavailable: %s", strerror(errno));
+}
+
+static void
+require_class(enum pmc_class class, const char *name)
+{
+	const struct pmc_cpuinfo *pci;
+	uint32_t i;
+
+	ATF_REQUIRE_EQ(0, pmc_cpuinfo(&pci));
+	for (i = 0; i < pci->pm_nclass; i++) {
+		if (pci->pm_classes[i].pm_class == class)
+			return;
+	}
+	atf_tc_skip("%s PMCs are unavailable on this system", name);
+}
+
+static void
+allocate_soft_pmc(enum pmc_mode mode, int cpu, pmc_id_t *pmcid)
+{
+
+	ATF_REQUIRE_EQ(0, pmc_allocate("SOFT-CLOCK.HARD", mode, 0, cpu, pmcid,
+	    1));
+}
+
+ATF_TC_WITHOUT_HEAD(pmcallocate_invalid_mode);
+ATF_TC_BODY(pmcallocate_invalid_mode, tc)
+{
+	struct pmc_op_pmcallocate pa;
+
+	require_hwpmc();
+	clear_exterr();
+
+	memset(&pa, 0, sizeof(pa));
+	pa.pm_class = PMC_CLASS_SOFT;
+	pa.pm_mode = (enum pmc_mode)0xff;
+	pa.pm_cpu = PMC_CPU_ANY;
+
+	ATF_REQUIRE_ERRNO(EINVAL, hwpmc_call(PMC_OP_PMCALLOCATE, &pa) == -1);
+	require_exterr("Invalid PMC mode");
+}
+
+ATF_TC_WITHOUT_HEAD(pmcallocate_invalid_cpu);
+ATF_TC_BODY(pmcallocate_invalid_cpu, tc)
+{
+	struct pmc_op_pmcallocate pa;
+
+	require_hwpmc();
+	clear_exterr();
+
+	memset(&pa, 0, sizeof(pa));
+	pa.pm_class = PMC_CLASS_SOFT;
+	pa.pm_mode = PMC_MODE_TC;
+	pa.pm_cpu = (uint32_t)pmc_ncpu();
+
+	ATF_REQUIRE_ERRNO(EINVAL, hwpmc_call(PMC_OP_PMCALLOCATE, &pa) == -1);
+	require_exterr("Invalid PMC CPU");
+}
+
+ATF_TC_WITHOUT_HEAD(pmcallocate_invalid_flags);
+ATF_TC_BODY(pmcallocate_invalid_flags, tc)
+{
+	struct pmc_op_pmcallocate pa;
+
+	require_hwpmc();
+	clear_exterr();
+
+	memset(&pa, 0, sizeof(pa));
+	pa.pm_class = PMC_CLASS_SOFT;
+	pa.pm_mode = PMC_MODE_TC;
+	pa.pm_cpu = PMC_CPU_ANY;
+	pa.pm_flags = 1U << 31;
+
+	ATF_REQUIRE_ERRNO(EINVAL, hwpmc_call(PMC_OP_PMCALLOCATE, &pa) == -1);
+	require_exterr("Invalid PMC flags");
+}
+
+ATF_TC_WITHOUT_HEAD(pmcattach_system_mode);
+ATF_TC_BODY(pmcattach_system_mode, tc)
+{
+	struct pmc_op_pmcattach pa;
+	pmc_id_t pmcid;
+
+	require_hwpmc();
+	clear_exterr();
+	allocate_soft_pmc(PMC_MODE_SC, 0, &pmcid);
+
+	memset(&pa, 0, sizeof(pa));
+	pa.pm_pmc = pmcid;
+	pa.pm_pid = getpid();
+
+	ATF_REQUIRE_ERRNO(EINVAL, hwpmc_call(PMC_OP_PMCATTACH, &pa) == -1);
+	require_exterr("Cannot attach a system-mode PMC");
+
+	ATF_REQUIRE_EQ(0, pmc_release(pmcid));
+}
+
+ATF_TC_WITHOUT_HEAD(pmcattach_running_pmc);
+ATF_TC_BODY(pmcattach_running_pmc, tc)
+{
+	struct pmc_op_pmcattach pa;
+	pmc_id_t pmcid;
+
+	require_hwpmc();
+	clear_exterr();
+	allocate_soft_pmc(PMC_MODE_TC, PMC_CPU_ANY, &pmcid);
+	ATF_REQUIRE_EQ(0, pmc_start(pmcid));
+
+	memset(&pa, 0, sizeof(pa));
+	pa.pm_pmc = pmcid;
+	pa.pm_pid = getpid();
+
+	ATF_REQUIRE_ERRNO(EBUSY, hwpmc_call(PMC_OP_PMCATTACH, &pa) == -1);
+	require_exterr("PMC must be stopped before attach");
+
+	ATF_REQUIRE_EQ(0, pmc_stop(pmcid));
+	ATF_REQUIRE_EQ(0, pmc_release(pmcid));
+}
+
+ATF_TC_WITHOUT_HEAD(pmcrw_no_flags);
+ATF_TC_BODY(pmcrw_no_flags, tc)
+{
+	struct pmc_op_pmcrw prw;
+	pmc_id_t pmcid;
+
+	require_hwpmc();
+	clear_exterr();
+	allocate_soft_pmc(PMC_MODE_TC, PMC_CPU_ANY, &pmcid);
+
+	memset(&prw, 0, sizeof(prw));
+	prw.pm_pmcid = pmcid;
+
+	ATF_REQUIRE_ERRNO(EINVAL, hwpmc_call(PMC_OP_PMCRW, &prw) == -1);
+	require_exterr("PMCRW requires PMC_F_OLDVALUE");
+
+	ATF_REQUIRE_EQ(0, pmc_release(pmcid));
+}
+
+ATF_TC_WITHOUT_HEAD(pmcrw_write_running);
+ATF_TC_BODY(pmcrw_write_running, tc)
+{
+	struct pmc_op_pmcrw prw;
+	pmc_id_t pmcid;
+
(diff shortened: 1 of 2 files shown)


afscoelho:
Hi @mashtizadeh / @mhorne,  just a continuation of the first PR that introduced exterror. Thanks for your time

afscoelho:
@mashtizadeh / @mhorne

sys/dev/hwpmc/hwpmc_ibs.c
```diff
@@ -113,7 +113,10 @@ ibs_validate_fetch_config(uint64_t config)
 {
 
 	if ((config & ~(ibs_fetch_allowed_mask | ibs_fetch_extra_mask)) != 0)
-		return (EINVAL);
+		return (EXTERROR(EINVAL,
```
mashtizadeh, line 154: This one looks good

sys/dev/hwpmc/hwpmc_ibs.c
```diff
@@ -127,17 +130,21 @@ ibs_validate_op_config(uint64_t config)
 
 	if ((config & IBS_OP_CTL_LATFLTEN) != 0) {
 		if ((ibs_features & CPUID_IBSID_IBSLOADLATENCYFILT) == 0)
-			return (EINVAL);
+			return (EXTERROR(EINVAL,
+			    "IBS op load-latency filter is not supported"));
 		if ((config & IBS_OP_CTL_L3MISSONLY) == 0)
-			return (EINVAL);
+			return (EXTERROR(EINVAL,
+			    "IBS op load-latency filter requires L3MISS"));
```
mashtizadeh: Lower case l3miss to match the command line interface

sys/dev/hwpmc/hwpmc_amd.c
```diff
@@ -398,28 +401,38 @@ amd_allocate_pmc(int cpu __unused, int ri, struct pmc *pm,
...
 	if ((a->pm_flags & PMC_F_EV_PMU) == 0)
-		return (EINVAL);
+		return (EXTERROR(EINVAL, "AMD PMCs require PMC_F_EV_PMU"));
 
 	caps = pm->pm_caps;
 
 	if (((caps & PMC_CAP_PRECISE) != 0) &&
 	    ((pd->pd_caps & PMC_CAP_PRECISE) == 0))
-		return (EINVAL);
+		return (EXTERROR(EINVAL,
```
mashtizadeh: I’m worried about these checks before the subclass check below.  If the user gets an error the output is unreadable because we’re getting the errors from all the attempts by hwpmc_mod to find the correct index for a PMC.  This is really misleading and hard to read.

mashtizadeh: I’m thinking we just have to drop all of the ones here up to and including the subclass.  If you can test it try to trigger a bad bit in a DF counter and see what you see.

mashtizadeh (changes requested):
Can you check the output for a DF counter with a reserved bit set?  If it shows all those errors multiple times we have a problem then.

sys/dev/hwpmc/hwpmc_amd.c
afscoelho: > I’m worried about these checks before the subclass check below. If the user gets an error the output is unreadable because we’re getting the errors from all the attempts by hwpmc_mod to find the correct index for a PMC. This is really misleading and hard to read.

Got it! You are right. The hwpmc_mod capture helper only filters by pcd_class == class, since it can't see the AMD-specific rows. real validations like for example the config bits on a DF counter overwrites nothing, because the capture was already locked in.
My next push will separate between discriminator checks and true validation checks. I think the fix is to make only validation failures use the exterr. Discriminator checks will just return EINVAL

afscoelho: > I’m thinking we just have to drop all of the ones here up to and including the subclass. If you can test it try to trigger a bad bit in a DF counter and see what you see.

right, I test it, the user now sees exactly which bits violated the DF subclass mask. Before the fix, we get the message "AMD subclass 2 does not match row subclass 0" now we get "AMD PMU config has unsupported bits 0xe7fffff0ffbf0000" (the test amd_invalid_df_config_bits does that)

afscoelho:
@mashtizadeh

sys/dev/hwpmc/hwpmc_amd.c
```diff
@@ -405,21 +408,26 @@ amd_allocate_pmc(int cpu __unused, int ri, struct pmc *pm,
...
-	    ((pd->pd_caps & PMC_CAP_PRECISE) == 0))
-		return (EINVAL);
-
 	PMCDBG2(MDP, ALL, 1,"amd-allocate ri=%d caps=0x%x", ri, caps);
 
 	/* Validate sub-class. */
 	if (amd_pmcdesc[ri].pm_subclass != a->pm_md.pm_amd.pm_amd_sub_class)
 		return (EINVAL);
 
+	if (((caps & PMC_CAP_PRECISE) != 0) &&
```
mashtizadeh, line 414: This one will spew unwanted errors because the precise flag is only for the 2nd core counter.

sys/dev/hwpmc/hwpmc_mod.c
```diff
@@ -3673,7 +3765,8 @@ pmc_do_op_pmcrw(const struct pmc_op_pmcrw *prw, pmc_value_t *valp)
 
 	/* Must have at least one flag set. */
 	if ((prw->pm_flags & (PMC_F_OLDVALUE | PMC_F_NEWVALUE)) == 0)
-		return (EINVAL);
+		return (EXTERROR(EINVAL,
+		    "PMCRW requires OLDVALUE and/or NEWVALUE flags"));
```
mashtizadeh: We should be consistent, either spell out the flags everywhere or don’t.  See PMC_F_USERCALLCHAIN above.

sys/dev/hwpmc/hwpmc_mod.c
```diff
@@ -3445,11 +3510,15 @@ pmc_do_op_pmcallocate(struct thread *td, struct pmc_op_pmcallocate *pa)
...
 				continue;
 
+			td->td_pflags2 &= ~TDP2_EXTERR;
 			if (pcd->pcd_allocate_pmc(td->td_oncpu, adjri, pmc,
 			    pa) == 0) {
 				/* Success. */
 				break;
 			}
+			if (pcd->pcd_class == class &&
+			    pmc_capture_hwpmc_exterr(td, &alloc_ke))
```
mashtizadeh: Is there a reason to put all this code in the individual ops couldn’t you move it up to the main system call path?  Am I missing something?

sys/dev/hwpmc/hwpmc_amd.c
afscoelho: Right, reverted to bare `EINVAL`. Also added a class check in the loop to skip non-matching classes, so only relevant allocators run now.

sys/dev/hwpmc/hwpmc_mod.c
afscoelho: Fixed, changed to `PMC_F_OLDVALUE` and `PMC_F_NEWVALUE`.

sys/dev/hwpmc/hwpmc_mod.c
afscoelho: You're right, moved it up. Added `if (pcd->pcd_class != class) continue;` in both loops so only matching-class allocators run, and removed the per-iteration capture. Now we just capture once after the loop if it fails.

mashtizadeh: I don't think that it tests agains the wrong class, the problem is the AMD driver uses subclasses that the module doesn't know about.

afscoelho: right, the class skip is more of a optimization. But the subclass and other discriminator checks all return bare EINVAL without EXTERROR, so they don't pollute the thread state. The EXTERROR only fires after the subclass already matches, so the wrong subclass rows fail silently.

mashtizadeh: https://github.com/AMDESE/freebsd-src/blob/5590e2061a8f8e2bd93d89841f551a7429b3a7d5/sys/dev/hwpmc/hwpmc_mod.c#L3434

We already match the class earlier in the allocation, see above.  So I don’t think your optimization will ever run, unless we break the semantics and set different classes per index.

We probably should support this then subclasses could go away, but that would break our API, and I think a number of assumptions of hwpmc mod.

afscoelho: yeah, hmm. I think the asymmetry here is because this PR makes ibs_allocate_pmc's class-mismatch return EXTERROR(...) while amd_allocate_pmc returns bare EINVAL for the same condition. If I make IBS to return EINVAL for the "wrong class" path , there is not going to have the EXTERROR-pollution risk and the "pcd->pcd_class != class continue" really became redundant. Would you prefer I go that route, or do you have a different fix in mind?

mashtizadeh (approved):
@mhorne

afscoelho:
Hi @imp, would you be willing to take a look at this PR? It promotes AMD hwpmc extended-error diagnostics upstream, with a Kyua test covering the negative scenarios. @mashtizadeh has already approved. Happy to address any feedback. Thanks!

sys/dev/hwpmc/hwpmc_mod.c
```diff
@@ -3424,13 +3482,16 @@ pmc_do_op_pmcallocate(struct thread *td, struct pmc_op_pmcallocate *pa)
 		for (n = pcd->pcd_ri; n < md->pmd_npmc; n++) {
 			pcd = pmc_ri_to_classdep(md, n, &adjri);
 
+			if (pcd->pcd_class != class)
+				continue;
```
mhorne: This is not necessary, and is impossible based on pcd = pmc_class_to_classdep(class) above.
Besides, it is not in-scope for the proposed change. Is there something I am missing?
```suggestion

```

sys/dev/hwpmc/hwpmc_mod.c
```diff
@@ -3424,13 +3482,16 @@ pmc_do_op_pmcallocate(struct thread *td, struct pmc_op_pmcallocate *pa)
...
 			if (!pmc_can_allocate_row(n, mode) ||
 			    !pmc_can_allocate_rowindex(p, n, cpu))
 				continue;
 			if (!PMC_IS_UNALLOCATED(cpu, n) &&
 			    !PMC_IS_SHAREABLE_PMC(cpu, n))
 				continue;
 
+			td->td_pflags2 &= ~TDP2_EXTERR;
```
mhorne: Why? It does not seem possible to reach here with this flag set. In all EXTERROR cases above we return.

If you are messing with the flag manually, it is a good sign that something is wrong. 

What is the intent?

sys/dev/hwpmc/hwpmc_mod.c
```diff
@@ -3441,10 +3502,13 @@ pmc_do_op_pmcallocate(struct thread *td, struct pmc_op_pmcallocate *pa)
 		for (n = pcd->pcd_ri; n < md->pmd_npmc; n++) {
 			pcd = pmc_ri_to_classdep(md, n, &adjri);
 
+			if (pcd->pcd_class != class)
+				continue;
```
mhorne:
```suggestion

```

tests/sys/kern/hwpmc_exterr_test.c
```diff
@@ -0,0 +1,467 @@
...
+	}
+	return (false);
+}
+
+static void
+require_class(enum pmc_class class, const char *name)
+{
+
+	if (have_class(class))
+		return;
```
mhorne: `have_class()` is used only here; I think it should just be expanded inline.

tests/sys/kern/hwpmc_exterr_test.c
mhorne, line 1: Thank you very much for writing the tests. This is clearly an area where hwpmc is lacking, so, it is much appreciated.

I will have to try running them myself. I did not review this file thoroughly.

sys/dev/hwpmc/hwpmc_mod.c
```diff
@@ -3256,6 +3267,35 @@ pmc_class_to_classdep(enum pmc_class class)
...
+static int
+pmc_return_saved_exterr(const struct kexterr *ke, int error)
+{
+	struct kexterr saved;
+
+	saved = *ke;
+	saved.error = error;
+	exterr_set_from(&saved);
+	return (error);
+}
```
mhorne: I'm sorry, but I really do not understand what you are attempting to do with these functions. Can you clarify?

Is there an error case that is hard to handle with EXTERROR() due to complex control flow? My suggestion is not to worry too much about this; it is better if we get the simple cases merged first.

But, by my quick analysis these functions do not seem necessary.

mhorne (changes requested):
Hi. I acknowledge I have been slow to respond to this one. I beg your pardon.

One piece of feedback I need to highlight: Please keep the scope of your changes smaller. It will help the speed of review to approve or reject specific chunks. For example, this submission should logically be three commits:
 1. hwpmc_amd exterror messages (small, easy to review)
 2. hwpmc_mod exterror messages (medium, takes more time to review)
 3. New tests (large, some review + testing required)

Otherwise, thanks a lot for your substantial effort on this.

afscoelho:
> Hi. I acknowledge I have been slow to respond to this one. I beg your pardon.
> 
> One piece of feedback I need to highlight: Please keep the scope of your changes smaller. It will help the speed of review to approve or reject specific chunks. For example, this submission should logically be three commits:
> 
> 1. hwpmc_amd exterror messages (small, easy to review)
> 2. hwpmc_mod exterror messages (medium, takes more time to review)
> 3. New tests (large, some review + testing required)
> 
> Otherwise, thanks a lot for your substantial effort on this.

No problem, thanks for the review! I agree , I'll restructure the branch into three commits

Per your inline comments I'll also simplify  hwpmc_mod changes, I hope to make commit 2 much easier to review.

sys/dev/hwpmc/hwpmc_mod.c
afscoelho: As is, I think it is reachable, but I agree it shouldn't be needed. I think I will restore the driver convention in ibs_allocate_pmc(): row mismatches return a bare EINVAL, and only genuine validation failures use EXTERROR(). With that, I think both of these checks can go. Will do in the next push

sys/dev/hwpmc/hwpmc_mod.c
afscoelho: It can be set here from the second iteration on, but I agree that manually clearing the flag is odd. Since exterr_set() overwrites unconditionally, "last recorded error wins", so I'm dropping these clears

tests/sys/kern/hwpmc_exterr_test.c
afscoelho: Thanks! For running them: they need root and the hwpmc module loaded (they skip with a message otherwise), and the AMD/IBS cases skip on non-AMD hardware. kyua test -k /usr/tests/Kyuafile sys/kern/hwpmc_exterr_test as root should do it.

sys/dev/hwpmc/hwpmc_mod.c
afscoelho: The intent was to keep the driver's specific error (e.g. "AMD PMU config has unsupported bits ...") from being overwritten by the generic "No PMC row accepted the allocation request" EXTERROR after the loop, since exterr_set() overwrites unconditionally. but yeah, this is too much :- )

afscoelho:
Thanks again for the review. I restructured the commits and did the changes you suggested. Also rebased onto current main, which picked up the new Zen6 IBS work.

imp:
> Hi @imp, would you be willing to take a look at this PR? It promotes AMD hwpmc extended-error diagnostics upstream, with a Kyua test covering the negative scenarios. @mashtizadeh has already approved. Happy to address any feedback. Thanks!

Hi @afscoelho, thanks, but I don't think I am qualified to give any meaningful feedback on that work.

bsdimp (approved):
This looks good to my eye, modulo the merge conflicts that can be taken care of on landing by rerunning the exterrr gen script.

afscoelho:
Hi @mhorne, thanks for the review. I've pushed the changes you asked for.  When you get a chance, could you re-review them? Thanks again for the time.

afscoelho:
> > Hi @imp, would you be willing to take a look at this PR? It promotes AMD hwpmc extended-error diagnostics upstream, with a Kyua test covering the negative scenarios. @mashtizadeh has already approved. Happy to address any feedback. Thanks!
> 
> Hi @afscoelho, thanks, but I don't think I am qualified to give any meaningful feedback on that work.

No worries at all, @imp. Thanks for the reply, and appreciate your time either way.

afscoelho:
Hi @mhorne , just a heads up that I rebased this branch onto current main to resolve the merge conflicts bsdimp flagged (also double checked the EXTERROR category IDs added here, no collisions with anything landed since May, so no regen needed). CI is running on the new commit now.

Since the commits are already split per your earlier request and the inline feedback from June has been addressed, would you be able to take another look when you get a chance? Thanks again for your time on this.

mhorne:
> Hi @mhorne , just a heads up that I rebased this branch onto current main to resolve the merge conflicts bsdimp flagged (also double checked the EXTERROR category IDs added here, no collisions with anything landed since May, so no regen needed). CI is running on the new commit now.
> 
> Since the commits are already split per your earlier request and the inline feedback from June has been addressed, would you be able to take another look when you get a chance? Thanks again for your time on this.

Hi. I merged the functional patches several weeks ago, but there were some changes needed to tests that were tedious to communicate. I've made the changes myself and I will merge both to the tree today; I hope you do not mind, and thanks again for your patience.

My changes are according to this patch:
[fixups.patch](https://github.com/user-attachments/files/31801016/fixups.patch)

Most important was:
 - move tests to shared `tests/sys/pmc/` directory
 - add ATF_REQUIRE_FEATURE("exterr_strings")` to each test-case
 - removed failing test case `amd_missing_pmu_flag` in light of the recent change: 6c4d9b9af1a3

PR #2186, opened 2026-05-11
rc: add ${name}_pre_start, ${name}_post_stop script support

rozhuk-im:
Run a service-based pre_start running after setup but before start_precmd and post_stop script running after stop_postcmd.

Useful for automatic environment/chroot generation and cleanup.

### DIFF
--- a/libexec/rc/rc.subr
+++ b/libexec/rc/rc.subr
@@ -975,6 +975,13 @@ startmsg()
 #
 #	${name}_prepend	n	Command added before ${command}.
 #
+#	${name}_pre_start n	Command executed during start
+#				after ${name}_setup and before
+#				${rc_arg}_precmd is run.
+#
+#	${name}_post_stop n	Command executed during stop
+#				after ${rc_arg}_postcmd is run.
+#
 #	${name}_setup	n	Command executed during start, restart and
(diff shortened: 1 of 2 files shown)


concussious:
Hi! So, what's wrong with precmd and postcmd?

rozhuk-im:
It is ok, but it is hard to use it to call scripts from rc.conf:
```
prosody_enable="YES"
prosody_user="prosody"
prosody_group="prosody"
prosody_pidfile="/var/run/prosody.pid"
if [ 'prosody' == "${name}" ]; then
	start_precmd="/usr/local/etc/rc.d/nginx stop"
	start_postcmd="/usr/local/etc/rc.d/nginx start"
fi
```
It will work only if port does not use it inside rc.d script.

Same time after patch it is able to write:
```
turnserver_chroot="/var/run/chroot-turnserver/"
turnserver_pre_start="/root/bin/chroots/turnserver.sh start"
turnserver_post_stop="/root/bin/chroots/turnserver.sh stop"
```
where ```turnserver.sh``` create/destroy chroot env.

```turnserver_setup``` is designed to generate configs and no good way use it for same kind purposes.

As administrator of services I want to able to run some code before start and after stop service, this can have many use cases to customize logic outside port, on system level.

PR #2187, opened 2026-05-11
vfs_syscalls.c: add NULL check for basep in kern_getdirentries()

haowuge:
kern_getdirentries() unconditionally dereferences the basep parameter without checking for NULL.  When called with basep == NULL, this causes a kernel page fault.  The residp parameter is already properly guarded with a NULL check in the same function, but basep was missed.

Add a NULL check for basep to match the existing pattern for residp.

Observed as a fatal page fault during boot on loongarch64:
  panic: Fatal page fault at 0xffff8000004a6e04: 0000000000000000
  ra: kern_getdirentries + 0x240

### DIFF
--- a/sys/kern/vfs_syscalls.c
+++ b/sys/kern/vfs_syscalls.c
@@ -4396,7 +4396,8 @@ unionread:
 		goto unionread;
 	}
 	VOP_UNLOCK(vp);
-	*basep = loff;
+	if (basep != NULL)
+		*basep = loff;
 	if (residp != NULL)
 		*residp = auio.uio_resid;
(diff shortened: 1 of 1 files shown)


markjdb:
But where is kern_getdirentries() called with basep==NULL? I do not see any code which does that. That suggests that your kernel panic is caused by something else.

haowuge:
> But where is kern_getdirentries() called with basep==NULL? I do not see any code which does that. That suggests that your kernel panic is caused by something else.

Apologies, you are correct. The NULL pointer was passed by my own debug code that I added to enumerate devices under /dev during boot. Thank you for pointing this out. I will close this pull request.

PR #2188, opened 2026-05-11
libutil++: Include <cerrno> in stringf.cc

kfv:
With modern Clang/libc++ and C++20 and later language modes, incidental transitive exposure is being reduced (see [Header Removal Policy](https://libcxx.llvm.org/DesignDocs/HeaderRemovalPolicy.html),) making such dependencies increasingly fragile over time.

This change makes the dependency explicit and aligns the code with the standard library interface model, improving portability across current and future language modes.

No functional or behavioural change intended.

### DIFF
--- a/lib/libutil++/libutil++.hh
+++ b/lib/libutil++/libutil++.hh
@@ -13,6 +13,7 @@
 #include <netdb.h>
 #include <unistd.h>
 
+#include <cerrno>
 #include <cstdarg>
 #include <cstdio>
 #include <cstdlib>


bms:
Thank you for not resorting to the use of "complexity generators" (Dijkstra).

Hmm, didn't know about this LLVM effort. Seems like a subset of: https://include-what-you-use.org/ ; Nothing in-tree is using @bsdjhb -ware <libutil++.hh> but ctld(8) at the moment, so build impact minimal. Most of the stuff in the namespace I'd argue should maybe be done by explicit scope guards, it's just semantic sugar to my eyes which is fine.

(Daniel Stenberg of cURL called out format string sanitization yesterday in his GenAI-assisted security audit announcement, but that's another thing.)

kfv:
We're in the same boat, yeah? ;-D I think contributions tend to turn out better when the tradeoffs are understood and improvements (whether or not valid) stem from an actual need being twigged.

You're absolutely right on the diff herein. The goal is just to remove potential obstacles toward adopting newer C modes over time.

And regarding Daniel's post, I haven't seen it yet, will take a look now, it must be interesting. Cheers!

clausecker:
Landed in 296c05f7212c37bd6e6bf7e42af11d42a275d493.

PR #2190, opened 2026-05-12
tests/sys/arch/amd64: add a basic ptrace syscall tampering test

shkhln:
Per request elsewhere.

### DIFF
--- a/tests/sys/arch/amd64/Makefile
+++ b/tests/sys/arch/amd64/Makefile
@@ -1,6 +1,7 @@
 TESTSDIR=	${TESTSBASE}/sys/arch/amd64
 
 PLAIN_TESTS_C+=	int0x80
+PLAIN_TESTS_C+=	ptrace-sce-tamper
 
 BINDIR=		${TESTSDIR}
 
--- a/tests/sys/arch/amd64/ptrace-sce-tamper.c
+++ b/tests/sys/arch/amd64/ptrace-sce-tamper.c
@@ -1,0 +1,89 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2026 Alex S <iwtcex@gmail.com>
+ */
+
+#include <machine/reg.h>
+#include <sys/ptrace.h>
+#include <sys/syscall.h>
+#include <sys/wait.h>
+
+#include <assert.h>
+#include <err.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+#ifndef __amd64__
+#error "amd64 only"
+#endif
+
+/*
+ * This test substitutes exit(42) instead of getpid() using ptrace.
+ */
+
+static const int EXPECTED_EXIT_CODE = 42;
+
+static void
+tamper(pid_t pid)
+{
+	struct ptrace_lwpinfo info;
+	struct reg regs;
+
+	if (ptrace(PT_LWPINFO, pid, (caddr_t)&info, sizeof(info)) == -1)
+		err(1, "ptrace(PT_LWPINFO)");
+
+	if ((info.pl_flags & PL_FLAG_SCE) != 0 &&
+	    info.pl_syscall_code == SYS_getpid) {
+		if (ptrace(PT_GETREGS, pid, (caddr_t)&regs, sizeof(regs)) == -1)
+			err(1, "ptrace(PT_GETREGS)");
+
+		regs.r_rax = SYS_exit;
+		regs.r_rdi = EXPECTED_EXIT_CODE;
+
+		if (ptrace(PT_SETREGS, pid, (caddr_t)&regs, sizeof(regs)) == -1)
+			err(1, "ptrace(PT_SETREGS)");
+	}
+}
+
+int
+main(void)
+{
+	pid_t pid;
+	int status;
+
+	pid = fork();
+	if (pid == -1)
+		err(1, "fork");
+
+	if (pid == 0) {
+		(void)ptrace(PT_TRACE_ME, 0, 0, 0);
+		(void)getpid();
+		exit(0);
+	} else {
+		if (ptrace(PT_ATTACH, pid, 0, 0) == -1)
+			err(1, "ptrace(PT_ATTACH)");
+
+		for (;;) {
+			if (wait(&status) == -1)
+				err(1, "wait");
+
+			if (WIFEXITED(status)) {
+				if (WEXITSTATUS(status) == EXPECTED_EXIT_CODE) {
+					printf("exit code changed\n");
+					exit(0);
+				} else {
+					printf("unable to change exit code\n");
+					exit(1);
+				}
+			}
+
+			assert(WIFSTOPPED(status));
+			tamper(pid);
+
+			if (ptrace(PT_TO_SCE, pid, (caddr_t)1, 0) == -1)
+				err(1, "ptrace(PT_TO_SCE)");
+		}
+	}
+}


kostikbel (approved):
Please fix the style issues.  The code looks fine.

shkhln:
I made a few minor style adjustments. Does this look acceptable now?

tests/sys/arch/amd64/Makefile
```diff
@@ -1,6 +1,6 @@
 TESTSDIR=	${TESTSBASE}/sys/arch/amd64
 
-PLAIN_TESTS_C+=	int0x80
+PLAIN_TESTS_C+=	int0x80 ptrace-sce-tamper
```
kostikbel: Please add a line
```
PLAIN_TESTS_C+=	ptrace-sce-tamper
```
instead.

tests/sys/arch/amd64/ptrace-sce-tamper.c
```diff
@@ -0,0 +1,89 @@
...
+tamper(pid_t pid)
+{
+	struct ptrace_lwpinfo info;
+	struct reg regs;
+
+	if (ptrace(PT_LWPINFO, pid, (caddr_t)&info, sizeof(info)) == -1)
+		err(1, "ptrace(PT_LWPINFO)");
+
+	if ((info.pl_flags & PL_FLAG_SCE) != 0 &&
+	    info.pl_syscall_code == SYS_getpid) {
```
kostikbel, line 38: This line indent should be +4 spaces and not tab.

tests/sys/arch/amd64/ptrace-sce-tamper.c
```diff
@@ -0,0 +1,89 @@
...
+
+	if (pid == 0) {
+		(void)ptrace(PT_TRACE_ME, 0, 0, 0);
+		(void)getpid();
+		exit(0);
+	} else {
+		if (ptrace(PT_ATTACH, pid, 0, 0) == -1)
+			err(1, "ptrace(PT_ATTACH)");
+
+		while (1) {
```
kostikbel: ```for (;;) {```

kostikbel (commented):
It was/is acceptable without these changes, but look more stylish and more consistent with the base system.

tests/sys/arch/amd64/ptrace-sce-tamper.c
shkhln: There is a tab from the start of the line and 4 spaces afterwards. GitHub displays tabs with the width of 4 spaces, so it's not as clearly visible as I'd like it to be. (It will be visible if you try to select those spaces/tabs with the mouse, though.)

tests/sys/arch/amd64/ptrace-sce-tamper.c
shkhln: Done.

tests/sys/arch/amd64/Makefile
shkhln: Done.

kostikbel (approved):

emaste:
Is there a link to a discussion / PR / commit that introduces the desired behaviour?

kostikbel:
> Is there a link to a discussion / PR / commit that introduces the desired behaviour?

The feature was introduced in 5b1162b964d915cb7a35644fcce961d0fe1e042e, broken in 99976934274de6fa19f049a0b6eac10856710f96.  The fix is https://reviews.freebsd.org/D56928.

emaste:
Ah, thanks. Do you want to land this change when your D56928 lands? Or I'm happy to push it after that; I have it staged locally (and it currently fails, of course).

tests/sys/arch/amd64/Makefile
```diff
@@ -1,6 +1,7 @@
 TESTSDIR=	${TESTSBASE}/sys/arch/amd64
 
 PLAIN_TESTS_C+=	int0x80
+PLAIN_TESTS_C+=	ptrace-sce-tamper
```
ngie-eign, line 4: Why `PLAIN_TESTS_C` instead of different format, e.g., ATF, GoogleTest, or TAP? "Plain" tests tend to be more of a maintenance burden than they're worth in the longterm.

tests/sys/arch/amd64/ptrace-sce-tamper.c
```diff
@@ -0,0 +1,89 @@
...
+
+#include <assert.h>
+#include <err.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+#ifndef __amd64__
+#error "amd64 only"
+#endif
```
ngie-eign, line 20: `require.arch` handles this better: [atf-test-case (4)](https://man.freebsd.org/cgi/man.cgi?query=atf-test-case&sektion=4).

tests/sys/arch/amd64/ptrace-sce-tamper.c
ngie-eign: Please run through `clang-format -i` from the top-level directory.

tests/sys/arch/amd64/Makefile
emaste: No objection to using a test framework but this sort of test case seems fine enough as a single plain test IMO

tests/sys/arch/amd64/ptrace-sce-tamper.c
emaste: This snippet isn't really needed, this subdir is only built on amd64

tests/sys/arch/amd64/ptrace-sce-tamper.c
emaste: This is just a GitHub rendering bug.

tests/sys/arch/amd64/ptrace-sce-tamper.c
ngie-eign: I just checked and `require.arch` is only available with ATF; it's not available with the other test format types.

tests/sys/arch/amd64/Makefile
kostikbel: > Why `PLAIN_TESTS_C` instead of different format, e.g., ATF, GoogleTest, or TAP? "Plain" tests tend to be more of a maintenance burden than they're worth in the longterm.

Because ATF+kyua  is impossible to debug when it fails.

kostikbel:
> Ah, thanks. Do you want to land this change when your D56928 lands? Or I'm happy to push it after that; I have it staged locally (and it currently fails, of course).

Please push it yourself, I just landed the review in cb31a266167c8bde3488fbdd6bf24c38b6f59977.

tests/sys/arch/amd64/ptrace-sce-tamper.c
ngie-eign: > This snippet isn't really needed, this subdir is only built on amd64

Unless someone tries building the code out of band.

tests/sys/arch/amd64/Makefile
ngie-eign: @kostikbel : I disagree, but I'd rather we continue this discussion elsewhere else instead of cluttering up this PR.

shkhln:
Apparently this has a small chance of failing with `ptrace(PT_ATTACH): Device busy` because I plainly misunderstood intended PT_ATTACH/PT_TRACE_ME usage (only PT_TRACE_ME is really needed here). Would anybody mind removing PT_ATTACH call from this? I can submit another patch, but it seems a bit bureaucratic.

kostikbel:
> Apparently this has a small chance of failing with `ptrace(PT_ATTACH): Device busy` because I plainly misunderstood intended PT_ATTACH/PT_TRACE_ME usage (only PT_TRACE_ME is really needed here). Would anybody mind removing PT_ATTACH call from this? I can submit another patch, but it seems a bit bureaucratic.

Mail me the git-format patch.

shkhln:
On the other hand, I got rid of exec when I was converting my repro to this test, so PT_ATTACH is arguably more appropriate.

> Mail me the git-format patch.

Let's just stick to this thread, GitHub sends email anyway. Here is the patch:  https://github.com/shkhln/freebsd-src/commit/9d054e05f48fe6d565a0d5abf9b45b3ec4dc9560.patch. I'm not sure if raise(SIGSTOP) actually eliminates the race, but I can no longer make this test fail with high CPU load, which is an improvement.

PR #2191, opened 2026-05-13
mac: allow policies to restrict vnode-backed mmap max_prot

mcoelho80:
For vnode-backed mappings, MAC policies currently get a chance to check the protection requested by mmap(2). However, a mapping created without PROT_EXEC may still retain VM_PROT_EXECUTE in max_prot and later be upgraded with mprotect(PROT_EXEC):

	fd = open(path, O_RDONLY);
	addr = mmap(NULL, len, PROT_READ, MAP_PRIVATE, fd, 0);
	mprotect(addr, len, PROT_READ | PROT_EXEC);

At mprotect(2) time, the VM layer no longer has the same vnode context that was available during mmap(2). As a result, a MAC policy that intends to prevent executable mappings from a given vnode may be bypassed by first creating a non-executable mapping and then upgrading it later.

Without this enforcement point, the MAC framework does not provide policy authors with a complete mechanism to enforce vnode-based executable mapping restrictions. This can make a policy look correct while leaving a bypass path open.

Let MAC policies reduce max_prot during the vnode-backed mmap path, while the vnode is still available. This allows a policy to remove VM_PROT_EXECUTE from max_prot, causing a later mprotect(PROT_EXEC) to fail naturally in the VM layer.

If the reduced max_prot no longer contains the protection requested for the initial mmap, fail the mmap with EACCES.

This does not change behavior for policies that do not implement the downgrade hook.

### DIFF
--- a/sys/security/mac/mac_framework.h
+++ b/sys/security/mac/mac_framework.h
@@ -592,6 +592,9 @@ mac_vnode_check_mmap(struct ucred *cred, struct vnode *vp, int prot,
 	return (0);
 }
 
+void	mac_vnode_check_mmap_downgrade(struct ucred *cred, struct vnode *vp,
+	    int *prot);
+
 int	mac_vnode_check_open_impl(struct ucred *cred, struct vnode *vp,
 	    accmode_t accmode);
 #ifdef MAC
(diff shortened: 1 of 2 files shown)


mcoelho80:
Hi,

Could someone please check the queued CI jobs for this PR?

The amd64 checks have passed, but the remaining smoke tests appear to be stuck waiting to run. It may need a CI re-trigger or some attention from the test infrastructure side.

Thanks!

mcoelho80:
Hi,

The PR is ready for review, and the source/build checks that have run so far are green.

However, the remaining internal smoke tests still appear to be queued and not starting. Could someone with access to the FreeBSD CI infrastructure please re-trigger or inspect these jobs?

If this change needs to go through a different review path due to touching the MAC framework, please let me know and I will adjust accordingly.

Thanks.

PR #2196, opened 2026-05-14
bluetooth: signal devd on netgraph device events

charlesrocket:
Currently, devd emits events for external adapters only. Send Netgraph init/disconnect events to devd so the internal adapter's state could be asserted from userland.

### DIFF
--- a/sys/netgraph/bluetooth/hci/ng_hci_main.c
+++ b/sys/netgraph/bluetooth/hci/ng_hci_main.c
@@ -33,7 +33,10 @@
  */
 
 #include <sys/param.h>
+#include <sys/devctl.h>
+#include <sys/sbuf.h>
 #include <sys/systm.h>
+#include <sys/syslog.h>
 #include <sys/kernel.h>
 #include <sys/endian.h>
 #include <sys/malloc.h>
@@ -110,6 +113,19 @@ static int ng_hci_linktype_to_addrtype(int linktype)
 	}
 	return BDADDR_BREDR;
 }
+
+/*
+ * Append bdaddr_t as "xx:xx:xx:xx:xx:xx" to sbuf.
+ */
+
+static void
+ng_hci_append_bdaddr_to_sbuf(struct sbuf *sb, const bdaddr_t *ba)
+{
+	sbuf_printf(sb, "%02x:%02x:%02x:%02x:%02x:%02x",
+		ba->b[5], ba->b[4], ba->b[3],
+		ba->b[2], ba->b[1], ba->b[0]);
+} /* ng_hci_append_bdaddr_to_sbuf */
+
 /*****************************************************************************
  *****************************************************************************
  **                   Netgraph methods implementation
@@ -268,6 +284,28 @@ ng_hci_disconnect(hook_p hook)
 		/* Connection terminated by local host */
 		ng_hci_unit_clean(unit, 0x16);
 		unit->state &= ~(NG_HCI_UNIT_CONNECTED|NG_HCI_UNIT_INITED);
+
+		/* Signal power off to devd */
+		{
+			struct sbuf *sb;
+			sb = sbuf_new_auto();
+			sbuf_printf(sb, "node=%s bdaddr=",NG_NODE_NAME(NG_HOOK_NODE(hook)));
+			ng_hci_append_bdaddr_to_sbuf(sb, &unit->bdaddr);
+			sbuf_printf(sb, "\n");
+
+			if (sbuf_finish(sb) > 0) {
+				log(LOG_WARNING,
+					"hci: failed to signal bt device power off to devd: "
+					"%02x:%02x:%02x:%02x:%02x:%02x\n",
+					unit->bdaddr.b[5], unit->bdaddr.b[4],
+					unit->bdaddr.b[3], unit->bdaddr.b[2],
+					unit->bdaddr.b[1], unit->bdaddr.b[0]);
+			} else {
+				devctl_notify("BLUETOOTH", "HCI", "POWERED_OFF", sbuf_data(sb));
+			}
+
+			sbuf_delete(sb);
+		}
 	} else
 		return (EINVAL);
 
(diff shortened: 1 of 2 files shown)


charlesrocket:
## Motivation
I needed a piece of code that could listen for bluetooth adapter's state changes in userland (event-driven exclusively, _absolutely_ no polling). Since this is a built-in/static adapter (T480), `devd` does not produce any events on "power cycle" (one needs to physically pull the device out/disconnect to trigger these events, which is not possible with soldered devices). The only other way to catch these changes is via `netgraph`, but it requires _root_.

With this patch, `netgraph` sends "power cycle" events (`NGM_HCI_NODE_INIT` case and `ng_hci_disconnect()` call) to `devd`, allowing userland to access these events (`kqueue` on the socket/directly/etc).

sys/netgraph/bluetooth/hci/ng_hci_main.c
```diff
@@ -268,6 +282,17 @@ ng_hci_disconnect(hook_p hook)
...
 		ng_hci_unit_clean(unit, 0x16);
 		unit->state &= ~(NG_HCI_UNIT_CONNECTED|NG_HCI_UNIT_INITED);
+
+		{
+			char bdstr[18];
+			char data[96];
+
+			ng_hci_bdaddr_to_str(&unit->bdaddr, bdstr);
+			snprintf(data, sizeof(data), "node=%s bdaddr=%s",
+				NG_NODE_NAME(NG_HOOK_NODE(hook)), bdstr);
```
bsdimp: Since fixed buffers are a problem, we are encouraging people to use the sbuf routines to create the string.

bsdimp:
Left a review comment. I generally like this, however. Please use the safer routines or ask me questions if you need more info that I've provided.

charlesrocket:
@bsdimp Thank you, will push in a few.

charlesrocket:
Switched to sbuf(9)

sys/netgraph/bluetooth/hci/ng_hci_main.c
```diff
@@ -268,6 +284,25 @@ ng_hci_disconnect(hook_p hook)
...
 		unit->state &= ~(NG_HCI_UNIT_CONNECTED|NG_HCI_UNIT_INITED);
+
+		/* Signal power off to devd */
+		{
+			char	bdstr[18];
+			struct sbuf *sb;
+
+			sb = sbuf_new_auto();
+			ng_hci_bdaddr_to_str(&unit->bdaddr, bdstr);
+			sbuf_sprintf(sb, "node=%s bdaddr=%s\n",
```
TeddyEngel: I think you meant sbuf_printf?

sys/netgraph/bluetooth/hci/ng_hci_main.c
```diff
@@ -268,6 +284,25 @@ ng_hci_disconnect(hook_p hook)
...
+				NG_NODE_NAME(NG_HOOK_NODE(hook)), bdstr);
+
+			if (sbuf_finish(sb) != 0) err(1, "Could not generate sbuf message");
+
+			transmit_msg(sbuf_data(sb), sbuf_len(sb));
+			devctl_notify("BLUETOOTH", "HCI", "POWERED_OFF", sb->buf);
```
TeddyEngel: Use sbuf_data(sb) to access the buffer instead of direct access - also the buffer is at ->s_buf

sys/netgraph/bluetooth/hci/ng_hci_main.c
```diff
@@ -268,6 +284,25 @@ ng_hci_disconnect(hook_p hook)
...
+			struct sbuf *sb;
+
+			sb = sbuf_new_auto();
+			ng_hci_bdaddr_to_str(&unit->bdaddr, bdstr);
+			sbuf_sprintf(sb, "node=%s bdaddr=%s\n",
+				NG_NODE_NAME(NG_HOOK_NODE(hook)), bdstr);
+
+			if (sbuf_finish(sb) != 0) err(1, "Could not generate sbuf message");
+
+			transmit_msg(sbuf_data(sb), sbuf_len(sb));
```
TeddyEngel: where is transmit_msg defined?

sys/netgraph/bluetooth/hci/ng_hci_main.c
charlesrocket: Oh yea gotta adjust these

charlesrocket:
@TeddyEngel wow, how I got CI green with the current state :thinking: https://github.com/charlesrocket/freebsd-src/actions/runs/26329026100

sys/netgraph/bluetooth/hci/ng_hci_main.c
charlesrocket: Actually, this is a good one. I cannot find this object anywhere besides the man page reference in the example that I used. And somehow CI is green.

sys/netgraph/bluetooth/hci/ng_hci_main.c
```diff
@@ -268,6 +284,24 @@ ng_hci_disconnect(hook_p hook)
...
+		{
+			char	bdstr[18];
+			struct sbuf *sb;
+
+			sb = sbuf_new_auto();
+			ng_hci_bdaddr_to_str(&unit->bdaddr, bdstr);
+			sbuf_printf(sb, "node=%s bdaddr=%s\n",
+				NG_NODE_NAME(NG_HOOK_NODE(hook)), bdstr);
+
+			if (sbuf_finish(sb) != 0) err(1, "Could not generate sbuf message");
```
TeddyEngel: also, err() isn't available in kernel land - check other kernel code pieces for how errors are handled

charlesrocket: On `9` man page... This is hands down the most broken example I've ever seen in almost 30 years :grinning:

bsdimp: > On `9` man page... This is hands down the most broken example I've ever seen in almost 30 years :grinning: 

This remark isn't called for. It's not actionable and kinda mean spirited even though it's trying to be funny. Can you maybe share which example is wrong ?

charlesrocket:
Switched to logger to follow netgraph. Is it okay if I issue warnings on these errors?

sys/netgraph/bluetooth/hci/ng_hci_main.c
charlesrocket: @bsdimp Sorry, having issues dealing with this PR. The sbuf(9) man page example is confusing, mixing placeholders with valid objects. I will do more research later and propose a patch to ensure this case will be a single instance.

erikarn:
This looks fine to me! thanks!

sys/netgraph/bluetooth/hci/ng_hci_main.c
```diff
@@ -34,11 +34,15 @@
...
 #include <sys/systm.h>
+#include <sys/syslog.h>
+#include <sys/types.h>
 #include <sys/kernel.h>
 #include <sys/endian.h>
 #include <sys/malloc.h>
 #include <sys/mbuf.h>
+#include <sys/sbuf.h>
 #include <sys/queue.h>
+#include <sys/devctl.h>
```
bsdimp: So sys/param.h already includes sys/types.h, so that's not needed. And the sys/lines after sys/param.h should be sorted alphabeitcally.

bsdimp:
Just one silly style thing.. Once that's updated I'm happy. And anybody else can land this if I don't circle back quickly enought.

charlesrocket:
@bsdimp moved my imports up (minus `sys/types.h`).

sys/netgraph/bluetooth/hci/ng_hci_main.c
erikarn: So having this here but then using a sbuf below kinda breaks the usefulness of sbufs.

One of the points of sbufs is to avoid needing to do error-prone memory allocation, NUL termination and string length hijinx.

I recommend changing this function to something like

ng_hci_append_bdaddr_to_sbuf(sbuf_t sb, const bdaddr_t *ba)
{
  // do sbuf_printf() like the original function, which will append the string contents
}

.. 

then where you're using the sbuf_printf() stuff below, just do:

sbuf_printf(sb, "node=%s bdaddr=", NG_NODE_NAME(...));
ng_hci_append_badaddr_to_sbuf(sb, &unit->bdaddr);
sbuf_printf(sb, "\n");

that eliminates the need for that temporary stack bdstr[18] and hoping everything is properly NUL terminated.

charlesrocket: Done

PR #2197, opened 2026-05-14
rk_gpio: implement PIC masking methods and mask unhandled IRQs

b1nc0d3x:
## Summary

The Rockchip GPIO controller implements PIC operations for the INTRNG framework but is missing four masking methods that INTRNG calls during the filter / ithread handoff:

  * `pic_disable_intr`
  * `pic_enable_intr`
  * `pic_pre_ithread`
  * `pic_post_ithread`

Without them, level-sensitive interrupt sources connected to a Rockchip GPIO pin re-fire continuously while their ithread runs. On a RockPro64 with a FUSB302B Type-C controller (i2c) attached to gpio1 INT_N, the system enters a ~210 kHz interrupt storm the moment the fusb302 driver attaches and INT_N goes low. Observed via:

```
dtrace -n 'fbt::intr_event_schedule_thread:entry { @[stack()]=count(); }'
```

Edge-triggered GPIO IRQs were already self-acking and unaffected.

Two complementary changes:

1.  Add the four `pic_disable_intr` / `pic_enable_intr` / `pic_pre_ithread` / `pic_post_ithread` method bodies. Each toggles the pin's `RK_GPIO_INTMASK` bit so the source is masked during the in-flight ithread window and unmasked on return. This honours the generic INTRNG mask/unmask sequence and mirrors what other arm64 GPIO drivers do (qoriq_gpio, bcm2835_gpio).

2.  When the GPIO IRQ filter dispatches a pin and finds no consumer registered, mask the pin at the controller (INTMASK=1, INTEN=0) before continuing. Level-triggered sources keep asserting until acked, so a single stuck pin (e.g. fusb302 INT_N during attach races or after detach) used to flood the console with thousands of "Interrupt pin=N unhandled" lines per second. The mask survives until something re-attaches and re-enables the IRQ via the standard `pic_enable_intr` path. The single warning line is preserved (now reading "unhandled -- masked") so the underlying issue is still visible.

## Test plan

- [x] RockPro64 (rk3399), level-triggered IRQ on gpio1 RK_PA2 (fusb302 INT_N). Before: ~210 kHz storm, `top -SH` shows `swi1:gpio1` pinned. After: 0 spurious IRQs at steady state; pin masks correctly after a detach.
- [x] Same board, edge-triggered IRQ on gpio0 (button input) — still fires once per press, no regression.
- [x] Synthesize an unhandled IRQ by raising a GPIO pin with no FDT consumer wired: console shows one "unhandled -- masked" line, storm does not recur.

## Suggested reviewers

`manu`, `mhorne`, `andrew`, `mmel` — arm64 / Rockchip / INTRNG maintainers per `git log`. Not @-mentioned to avoid spam if the GitHub handles don't match; happy to ping on request.

## Notes

First in a series of patches landing the RockPro64 USB-C DisplayPort + audio path. Successor patches (`drm_fb_helper` hotplug fix, `usbc` PD framework, `fusb302`, `rk_typec_phy` DP altmode, `rk_cdn_dp`) will follow in their own PRs after this one gets a sign-of-life from a reviewer.

This patch is independent of the rest of that series and useful on its own for anyone running a Rockchip board with a level-triggered GPIO interrupt source.

Author: Kyle Crenshaw &lt;B1nc0d3x@gmail.com&gt;

### DIFF
--- a/sys/arm64/rockchip/rk_gpio.c
+++ b/sys/arm64/rockchip/rk_gpio.c
@@ -227,8 +227,22 @@ rk_gpio_intr(void *arg)
 
 		status &= ~(1 << pin);
 		if (intr_isrc_dispatch(RK_GPIO_ISRC(sc, pin), tf)) {
-			device_printf(sc->sc_dev, "Interrupt pin=%d unhandled\n",
-			    pin);
+			/*
+			 * Pin asserted but no consumer is registered for it
+			 * yet (or anymore).  Level-triggered sources keep
+			 * firing on every interrupt cycle, so a single stuck
+			 * pin floods the console with thousands of these
+			 * messages per second.  Mask the pin's IRQ at the
+			 * controller and disable further dispatches; if a
+			 * consumer attaches later it will re-enable through
+			 * pic_enable_intr / rk_gpio_pic_enable_intr.
+			 */
+			RK_GPIO_LOCK(sc);
+			rk_gpio_write_bit(sc, RK_GPIO_INTMASK, pin, 1);
+			rk_gpio_write_bit(sc, RK_GPIO_INTEN, pin, 0);
+			RK_GPIO_UNLOCK(sc);
+			device_printf(sc->sc_dev,
+			    "Interrupt pin=%d unhandled — masked\n", pin);
 			continue;
 		}
 
@@ -825,6 +839,51 @@ rk_pic_setup_intr(device_t dev, struct intr_irqsrc *isrc,
 	return (0);
 }
 
+/*
+ * INTRNG calls pic_disable_intr() to mask the source when it must hold off
+ * delivery -- e.g. between FILTER_SCHEDULE_THREAD and ithread completion.
+ * Without these, level-low IRQs (FUSB302 INT_N) re-fire continuously and
+ * starve the ithread (~210 kHz storm observed via dtrace).
+ */
+static void
+rk_pic_disable_intr(device_t dev, struct intr_irqsrc *isrc)
+{
+	struct rk_gpio_softc *sc = device_get_softc(dev);
+	struct rk_pin_irqsrc *rkisrc = (struct rk_pin_irqsrc *)isrc;
+
+	RK_GPIO_LOCK(sc);
+	rk_gpio_write_bit(sc, RK_GPIO_INTMASK, rkisrc->irq, 1);
+	RK_GPIO_UNLOCK(sc);
+}
+
+static void
+rk_pic_enable_intr(device_t dev, struct intr_irqsrc *isrc)
+{
+	struct rk_gpio_softc *sc = device_get_softc(dev);
+	struct rk_pin_irqsrc *rkisrc = (struct rk_pin_irqsrc *)isrc;
+
+	RK_GPIO_LOCK(sc);
+	rk_gpio_write_bit(sc, RK_GPIO_INTMASK, rkisrc->irq, 0);
+	RK_GPIO_UNLOCK(sc);
+}
+
+/*
+ * Called by INTRNG before delivering to the ithread; mask the source.
+ * Re-enabled in pic_post_ithread once the ithread has cleared the
+ * device-side interrupt cause.
+ */
+static void
+rk_pic_pre_ithread(device_t dev, struct intr_irqsrc *isrc)
+{
+	rk_pic_disable_intr(dev, isrc);
+}
+
+static void
+rk_pic_post_ithread(device_t dev, struct intr_irqsrc *isrc)
+{
+	rk_pic_enable_intr(dev, isrc);
+}
+
 static int
 rk_pic_teardown_intr(device_t dev, struct intr_irqsrc *isrc,
     struct resource *res, struct intr_map_data *data)
@@ -873,6 +932,10 @@ static device_method_t rk_gpio_methods[] = {
 	DEVMETHOD(pic_map_intr,		rk_pic_map_intr),
 	DEVMETHOD(pic_setup_intr,	rk_pic_setup_intr),
 	DEVMETHOD(pic_teardown_intr,	rk_pic_teardown_intr),
+	DEVMETHOD(pic_disable_intr,	rk_pic_disable_intr),
+	DEVMETHOD(pic_enable_intr,	rk_pic_enable_intr),
+	DEVMETHOD(pic_pre_ithread,	rk_pic_pre_ithread),
+	DEVMETHOD(pic_post_ithread,	rk_pic_post_ithread),
 
 	/* ofw_bus interface */
 	DEVMETHOD(ofw_bus_get_node,	rk_gpio_get_node),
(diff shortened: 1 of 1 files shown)


sys/arm64/rockchip/rk_gpio.c
```diff
@@ -825,6 +839,51 @@ rk_pic_setup_intr(device_t dev, struct intr_irqsrc *isrc,
 	return (0);
 }
 
+/*
+ * INTRNG calls pic_disable_intr() to mask the source when it must hold off
+ * delivery -- e.g. between FILTER_SCHEDULE_THREAD and ithread completion.
+ * Without these, level-low IRQs (FUSB302 INT_N) re-fire continuously and
+ * starve the ithread (~210 kHz storm observed via dtrace).
```
mhorne: Small correction: INTRNG calls the `PIC_PRE_ITHREAD()` method at this time. It only calls `PIC_DISABLE_INTR()` during teardown/removal of the final handler.

sys/arm64/rockchip/rk_gpio.c
```diff
@@ -227,8 +227,22 @@ rk_gpio_intr(void *arg)
...
+			 * firing on every interrupt cycle, so a single stuck
+			 * pin floods the console with thousands of these
+			 * messages per second.  Mask the pin's IRQ at the
+			 * controller and disable further dispatches; if a
+			 * consumer attaches later it will re-enable through
+			 * pic_enable_intr / rk_gpio_pic_enable_intr.
+			 */
+			RK_GPIO_LOCK(sc);
+			rk_gpio_write_bit(sc, RK_GPIO_INTMASK, pin, 1);
+			rk_gpio_write_bit(sc, RK_GPIO_INTEN, pin, 0);
```
mhorne, line 242: Are you sure you want to disable the interrupt, not just mask it? I don't see how it can be enabled again subsequently.

sys/arm64/rockchip/rk_gpio.c
mhorne:
```suggestion
```

Optionally, you could omit this line, leaving the `rk_pic_enable_intr()` method to handle the initial unmasking. The same would apply to `rk_pic_teardown_intr()` and `rk_pic_disable_intr()`.

I think this is a fair matching of the hw/sw scheme.

sys/arm64/rockchip/rk_gpio.c
```diff
@@ -825,6 +839,51 @@ rk_pic_setup_intr(device_t dev, struct intr_irqsrc *isrc,
...
+}
+
+static void
+rk_pic_post_ithread(device_t dev, struct intr_irqsrc *isrc)
+{
+	rk_pic_enable_intr(dev, isrc);
+}
+
 static int
 rk_pic_teardown_intr(device_t dev, struct intr_irqsrc *isrc,
```
mhorne: Please keep `rk_pic_setup_intr()` and `rk_pic_teardown_intr()` placed together within the file. Best practice is to keep the functions in the same order they are listed in the DEVMETHOD table.

sys/arm64/rockchip/rk_gpio.c
mhorne, line 222: Another observation: this seems kind of early to clear the pending interrupts, and may be contributing to the storm. I would expect it this step to be done _after_ running all handlers...

Edit: on second thought, while this is an oddity, I don't think it has much effect on the storm you are seeing.

mhorne (approved):
LGTM. I will leave it a few days for other reviewers, then merge.

sys/arm64/rockchip/rk_gpio.c
b1nc0d3x: Fixed in 74e32a36a74 — the comment now reflects that `pic_pre_ithread()` handles the in-flight masking window, and `pic_disable_intr()` is only invoked at teardown/removal of the final handler.

sys/arm64/rockchip/rk_gpio.c
b1nc0d3x: Good catch — `pic_enable_intr()` only unmasked, so the filter-dispatch path that clears INTEN had no way back. Fixed in 74e32a36a74 by making `pic_disable_intr` / `pic_enable_intr` operate on INTEN (as INTRNG expects), and moving the in-flight masking into `pic_pre_ithread` / `pic_post_ithread` with their own INTMASK writes inlined. The dispatch path's mask+disable is now properly recoverable by a later `pic_enable_intr()`.

sys/arm64/rockchip/rk_gpio.c
b1nc0d3x: Done in 74e32a36a74 — dropped the `INTMASK=0` / `INTEN=1` lines from `pic_setup_intr`'s success path, and the matching `INTEN=0` / `INTMASK=0` from `pic_teardown_intr`. `setup_intr` now configures DEBOUNCE + trigger only, leaving activation to `pic_enable_intr()`; `teardown_intr` undoes the trigger-side config (DEBOUNCE) and relies on INTRNG having called `pic_disable_intr` first. Agreed, this is a cleaner hw/sw match.

sys/arm64/rockchip/rk_gpio.c
b1nc0d3x: Reordered in 74e32a36a74 to match the DEVMETHOD table: `map / setup / teardown / disable / enable / pre_ithread / post_ithread`.

sys/arm64/rockchip/rk_gpio.c
b1nc0d3x: Agreed it's an oddity worth flagging. For the level-triggered case that drove this PR the EOI write is effectively a no-op until the source deasserts, so order vs. dispatch doesn't change observed behavior; for edges it does matter and the current ordering matches what the Linux `gpio-rockchip` driver does (ack-at-top of the filter). Happy to revisit in a follow-up if you'd prefer the post-dispatch placement — felt safer to keep the storm fix as a minimal change.

b1nc0d3x:
Thanks for the thorough review and the approval, @mhorne. The substantive bug at line 242 was a real semantic mistake on my part — your "I don't see how it can be enabled again subsequently" framing made the missing INTEN write obvious. Fixup `74e32a36a74` addresses all five threads. Verified on RockPro64 v2.1: kernel boots, all five GPIO banks attach, the unhandled `gpio1` pin masks once and stays at `vmstat -i` `total=1 rate=0` (no storm) across the new disable/enable + pre/post-ithread semantics.

sys/arm64/rockchip/rk_gpio.c
```diff
@@ -873,6 +963,10 @@ static device_method_t rk_gpio_methods[] = {
 	DEVMETHOD(pic_map_intr,		rk_pic_map_intr),
 	DEVMETHOD(pic_setup_intr,	rk_pic_setup_intr),
 	DEVMETHOD(pic_teardown_intr,	rk_pic_teardown_intr),
+	DEVMETHOD(pic_disable_intr,	rk_pic_disable_intr),
+	DEVMETHOD(pic_enable_intr,	rk_pic_enable_intr),
+	DEVMETHOD(pic_pre_ithread,	rk_pic_pre_ithread),
+	DEVMETHOD(pic_post_ithread,	rk_pic_post_ithread),
```
ehem, line 969: There hasn't been an enforcement mechanism before, but PICs are supposed to implement all 3 hooks (pic_post_filter is non-optional).  @strejda will confirm this.  There is a plan to add something which will blatantly complain (panic()) if any of these is missing.

PR #2199, opened 2026-05-15
crashinfo: Create core.txt.last symlink

ricardobranco777:
When saving a coredump, savecore(8) maintains .last symlinks for the info and vmcore artifacts, but not for the crashinfo text report.

Create it here to point at the current core.txt.<bounds> file.

This makes /var/crash/core.txt.last track the same core dump as info.last and vmcore.last.

Supersedes: https://github.com/freebsd/freebsd-src/pull/2178

### DIFF
--- a/usr.sbin/crashinfo/crashinfo.sh
+++ b/usr.sbin/crashinfo/crashinfo.sh
@@ -149,6 +149,7 @@ fi
 VMCORE=$CRASHDIR/vmcore.$DUMPNR
 INFO=$CRASHDIR/info.$DUMPNR
 FILE=$CRASHDIR/core.txt.$DUMPNR
+LINK=$CRASHDIR/core.txt.last
 HOSTNAME=`hostname`
 
 if $BATCH; then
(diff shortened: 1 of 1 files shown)


mhorne (approved):
LGTM! Thank you.

mhorne:
I tested the script and made two small tweaks:
 - Move link creation earlier, so that it is not dependent on the success of the kgdb commands (which can fail)
 - remove `basename`, it did not seem to be needed.

Please see the commit and let me know if there was a problem. I wanted to avoid an extra feedback cycle, so I just proceeded.

Thanks again.

ricardobranco777:
> I tested the script and made two small tweaks:
> 
> * Move link creation earlier, so that it is not dependent on the success of the kgdb commands (which can fail)
> * remove `basename`, it did not seem to be needed.
> 
> Please see the commit and let me know if there was a problem. I wanted to avoid an extra feedback cycle, so I just proceeded.
> 
> Thanks again.

I've just tested it and it works.  The resulting symlink is relative like the others so the basename was indeed redundant.

Thanks for merging.

PR #2200, opened 2026-05-15
hwpmc: add AMD Zen3 IBS errata workarounds (1293, 1238, 1347)

afscoelho:
## Summary

- hwpmc: add AMD Zen3 IBS errata workarounds (# 1293, # 1238, # 1347)

### DIFF
--- a/sys/dev/hwpmc/hwpmc_ibs.h
+++ b/sys/dev/hwpmc/hwpmc_ibs.h
@@ -149,8 +149,13 @@
 
 #define IBS_OP_DATA2			0xC0011036 /* IBS Op Data 2 */
 #define IBS_OP_DATA3			0xC0011037 /* IBS Op Data 3 */
+#define IBS_OP_DATA3_OPDCMISSOPENMEMREQSMASK					\
+	(0x3FULL << 26)		     /* DC Miss Open Mem Reqs */
+#define IBS_OP_DATA3_SWPF		(1ULL << 21) /* Software Prefetch */
+#define IBS_OP_DATA3_L2MISS		(1ULL << 20) /* L2 Cache Miss */
 #define IBS_OP_DATA3_DCPHYADDRVALID	(1ULL << 18) /* DC Physical Address */
 #define IBS_OP_DATA3_DCLINADDRVALID	(1ULL << 17) /* DC Linear Address */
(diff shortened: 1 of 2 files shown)


afscoelho:
Hi @mashtizadeh / @mhorne, Thanks for your time

emaste:
What are #1293/#1238/#1347 in the commit title? GitHub renders them as three miscellaneous unrelated pull requests.

afscoelho:
> What are #1293/#1238/#1347 in the commit title? GitHub renders them as three miscellaneous unrelated pull requests.

 Sorry, those are AMD Family 19h errata numbers, not GitHub PR refs. Will reword to avoid the # autolink.

mashtizadeh:
Again this is like that other patch, if the errata only causes bits to be questionable we should resolve this in the tools I that I showed you, not in the kernel.  Otherwise, wIt’s going to become a spaghetti of attempts to clear bits.

I didn’t handle any of them yet but with the new pmc tools you’ll be able to put them in one central place for post-processing.  It also has the cpuids saved so you can almost verbatim copy this code into it at some point.

afscoelho:
> mashtizadeh

Hmm, ok same logic as #2143, I'll move the DATA2 zeroing and the two DATA3 bit clears into pmcstat_print_ibs_op next to the existing 1238 handling, drop the kernel hunk, and resubmmit.

mashtizadeh:
> > mashtizadeh
> 
> Hmm, ok same logic as #2143, I'll move the DATA2 zeroing and the two DATA3 bit clears into pmcstat_print_ibs_op next to the existing 1238 handling, drop the kernel hunk, and resubmmit.

I wouldn’t even waste time in there, leave it for the new tool I’m about to upstream the first couple commands.  I was hoping to do it earlier but I wanted to finish one thing first.

afscoelho:
Closing this PR based on feedback from @mashtizadeh. The errata workarounds (Zen3 errata 1293, 1238, 1347) for IBS bits should be handled in userspace tooling (pmcstat/the upcoming pmc tools), not in the kernel — same rationale as #2143. The new pmc tooling being upstreamed by @mashtizadeh will provide a better centralized place for this post-processing logic. Will revisit once that tooling lands.

PR #2201, opened 2026-05-16
libc: Add free_sized() and free_aligned_sized() as per C23

kfv:
Add `free_sized()` (§7.24.3.4) and `free_aligned_sized()` (§7.24.3.5) as per C23.

For now they're simply thin wrappers around `free(3)` as the standard allows implementations to ignore the size and alignment hints (§7.24.3.4.3, §7.24.3.5.3,) so the behaviour is correct and conformant.

Once jemalloc is bumped to 5.3.1, it'd be well worth rewiring these to use the allocator's sized deallocation for the performance and security benefits.

### DIFF
--- a/lib/libc/stdlib/malloc/free_aligned_sized.c
+++ b/lib/libc/stdlib/malloc/free_aligned_sized.c
@@ -1,0 +1,21 @@
+/*
+ * Copyright (c) 2026 Faraz Vahedi <kfv@kfv.io>
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <sys/cdefs.h>
+#include <stdlib.h>
+
+void	__free(void *);
+
+void
+__free_aligned_sized(void *ptr, size_t alignment, size_t size)
+{
+
+	(void)alignment;
+	(void)size;
+	__free(ptr);
+}
+
+__weak_reference(__free_aligned_sized, free_aligned_sized);
--- a/lib/libc/stdlib/malloc/free_sized.3
+++ b/lib/libc/stdlib/malloc/free_sized.3
@@ -1,0 +1,100 @@
+.\"
+.\" Copyright (c) 2026 Faraz Vahedi <kfv@kfv.io>
+.\"
+.\" SPDX-License-Identifier: BSD-2-Clause
+.\"
+.Dd May 17, 2026
+.Dt FREE_SIZED 3
+.Os
+.Sh NAME
+.Nm free_sized ,
+.Nm free_aligned_sized
+.Nd C23 sized deallocation functions
+.Sh SYNOPSIS
+.Lb libc
+.In stdlib.h
+.Ft void
+.Fn free_sized "void *ptr" "size_t size"
+.Ft void
+.Fn free_aligned_sized "void *ptr" "size_t alignment" "size_t size"
+.Sh DESCRIPTION
+The
+.Fn free_sized
+function deallocates the memory referenced by
+.Fa ptr
+that was previously allocated by
+.Xr malloc 3 ,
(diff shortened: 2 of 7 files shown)


clausecker:
Please bump the man page date when you update a man page.

CC @bsdimp for additional review, though LGTM from my side.

kfv:
> Please bump the man page date when you update a man page.

Sure, my apologies. Done.

lib/libc/stdlib/malloc/free_aligned_sized.c
```diff
@@ -0,0 +1,21 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2026
```
bsdimp: Needs a copyright holder.

lib/libc/stdlib/memory.3
```diff
@@ -101,3 +109,10 @@ and
 .Fn posix_memalign
 functions conform to
 .St -p1003.1-2001 .
+.Pp
+The
+.Fn free_sized
+and
+.Fn free_aligned_sized
+functions conform to
+.St -isoC-2023 .
```
bsdimp, line 118: You should also provide a description earlier in the man page saying what they do.

kfv: I thought it'd be unnecessary as none of the other functions have a description therein and jemalloc would have it covered once we upgrade. I just added a separate manpage for the interim.

lib/libc/stdlib/malloc/free_sized.3
```diff
@@ -0,0 +1,101 @@
...
+.Pp
+Once the system allocator is upgraded to jemalloc 5.3.1 or later, these
+functions will be wired to the allocator's sized deallocation support, as with
+the rest of the memory allocation API, so that
+.Fa size
+and
+.Fa alignment
+hints are used for performance and security benefits.
+That behaviour conforms to the practice recommended for each function in the
+C standard.
```
brooksdavis: I don't think the manpage should contain forward-looking information in the much detail.

lib/libc/stdlib/malloc/free_aligned_sized.c
```diff
@@ -0,0 +1,21 @@
...
+#include <stdlib.h>
+
+void	__free(void *);
+
+void
+__free_aligned_sized(void *ptr, size_t alignment, size_t size)
+{
+
+	(void)alignment;
+	(void)size;
```
brooksdavis, line 17: I'd personally use `__unused` in the function definition, but this is fine.

lib/libc/stdlib/malloc/free_sized.3
kfv: You're right. I slightly modified it so that it only conveys that it currently conforms to the minimum requirements of the standard, without providing any hints to the compiler for optimisation, etc. Let me know if you think anything else should be changed as well.

lib/libc/stdlib/malloc/free_sized.c
```diff
@@ -0,0 +1,20 @@
+/*
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2026 Faraz Vahedi <kfv@kfv.io>
+ */
+
```
clausecker, line 6: Copyright comes before the SPDX-License-Identifier.  Please check style(9).

kfv: Thanks.

clausecker:
@bsdimp You ok with this change?  Want to land it?

clausecker:
Landed in 5f732742ad5b3133a860a8969cf2bd13dc9ac358.

PR #2203, opened 2026-05-17
libc: C23 conformance for <assert.h>

kfv:
- Add `__STDC_VERSION_ASSERT_H__` feature-test macro
- Restrict `static_assert` macro definition to pre-C23 modes
- Make `assert` variadic as per C23
- Update `assert.3` accordingly

### DIFF
--- a/include/assert.h
+++ b/include/assert.h
@@ -46,19 +46,19 @@
 #undef __assert_unreachable
 
 #ifdef NDEBUG
-#define	assert(e)	((void)0)
-#define	_assert(e)	((void)0)
+#define assert(e)	((void)0)
+#define _assert(e)	((void)0)
 #if __BSD_VISIBLE
-#define	__assert_unreachable()	__unreachable()
-#endif	/* __BSD_VISIBLE */
+#define __assert_unreachable()	__unreachable()
+#endif /* __BSD_VISIBLE */
 #else
-#define	_assert(e)	assert(e)
+#define _assert(e)	assert(e)
 
-#define	assert(e)	((e) ? (void)0 : __assert(__func__, __FILE__, \
+#define assert(e)	((e) ? (void)0 : __assert(__func__, __FILE__, \
 			    __LINE__, #e))
 #if __BSD_VISIBLE
-#define	__assert_unreachable()	assert(0 && "unreachable segment reached")
-#endif	/* __BSD_VISIBLE */
+#define __assert_unreachable()	assert(0 && "unreachable segment reached")
+#endif /* __BSD_VISIBLE */
 #endif /* NDEBUG */
 
 #ifndef _ASSERT_H_
@@ -77,7 +77,7 @@
  * _Static_assert, as keywords.
  */
 #if __ISO_C_VISIBLE >= 2011 && __ISO_C_VISIBLE < 2023 && !defined(__cplusplus)
-#define	static_assert	_Static_assert
+#define static_assert	_Static_assert
 #endif
 
 __BEGIN_DECLS
--- a/include/assert.h
+++ b/include/assert.h
@@ -46,16 +46,43 @@
 #undef __assert_unreachable
 
 #ifdef NDEBUG
-#define assert(e)	((void)0)
-#define _assert(e)	((void)0)
+#define assert(...)	((void)0)
+#define _assert(...)	((void)0)
 #if __BSD_VISIBLE
 #define __assert_unreachable()	__unreachable()
 #endif /* __BSD_VISIBLE */
 #else
-#define _assert(e)	assert(e)
-
-#define assert(e)	((e) ? (void)0 : __assert(__func__, __FILE__, \
-			    __LINE__, #e))
+#ifdef __cplusplus
+#if __cplusplus < 202002L
+/*
+ * C++ modes prior to C++20 cannot simultaneously satisfy all three
+ * desirable properties of the sanitiser:
+ *
+ *   Approach                       No double-eval  Lambda support  Arity check
+ *   -----------------------------  --------------  --------------  -----------
+ *   sizeof(cast(expression))       yes             no              yes
+ *   static_cast<bool>(expression)  no              yes             no
+ *   (void)bool(expression)         no              yes             no
+ *
+ *   NOTE: C++20 introduced lambdas in unevaluated contexts; see P0315R4.
+ *
+ * Since no approach satisfies all three below C++20, the least harmful
+ * choice is to forgo the check entirely rather than silently break one
+ * of the remaining guarantees.
+ *
+ */
+#define __assert_sanitize(...)	((void)0)
+#else
+#define __assert_sanitize(...)	(void)sizeof(((bool(*)(bool))0)(__VA_ARGS__))
+#endif /* __cplusplus < 202002L */
+#else
+#define __assert_sanitize(...)	(void)sizeof(((_Bool(*)(_Bool))0)(__VA_ARGS__))
+#endif /* __cplusplus */
+#define assert(...)	(__assert_sanitize(__VA_ARGS__),       \
+			    (__VA_ARGS__) ? (void)0 :          \
+			    __assert(__func__, __FILE__,       \
+			    __LINE__, #__VA_ARGS__))
+#define _assert(...)	assert(__VA_ARGS__)
 #if __BSD_VISIBLE
 #define __assert_unreachable()	assert(0 && "unreachable segment reached")
 #endif /* __BSD_VISIBLE */
@@ -76,7 +103,8 @@
  * C23 defines static_assert and its obsolescent alternative spelling,
  * _Static_assert, as keywords.
  */
-#if __ISO_C_VISIBLE >= 2011 && __ISO_C_VISIBLE < 2023 && !defined(__cplusplus)
+#if __ISO_C_VISIBLE >= 2011 && !defined(__cplusplus) && \
+    __STDC_VERSION__ < 202311L
 #define static_assert	_Static_assert
 #endif
 
--- a/include/assert.h
+++ b/include/assert.h
@@ -88,8 +88,8 @@
 #endif /* __BSD_VISIBLE */
 #endif /* NDEBUG */
 
-#ifndef _ASSERT_H_
-#define _ASSERT_H_
+#ifndef __STDC_VERSION_ASSERT_H__
+#define __STDC_VERSION_ASSERT_H__ 202311L
 
 /*
  * Static assertions.  In principle we could define static_assert for
@@ -112,4 +112,4 @@ __BEGIN_DECLS
 void __assert(const char *, const char *, int, const char *) __dead2;
 __END_DECLS
 
-#endif /* !_ASSERT_H_ */
+#endif /* !__STDC_VERSION_ASSERT_H__ */
--- a/include/assert.h
+++ b/include/assert.h
@@ -72,8 +72,11 @@
  * C++ template parameters may contain commas, even if not enclosed in
  * parentheses, causing the _Static_assert macro to be invoked with more
  * than two parameters.
+ *
+ * C23 defines static_assert and its obsolescent alternative spelling,
+ * _Static_assert, as keywords.
  */
-#if __ISO_C_VISIBLE >= 2011 && !defined(__cplusplus)
+#if __ISO_C_VISIBLE >= 2011 && __ISO_C_VISIBLE < 2023 && !defined(__cplusplus)
 #define	static_assert	_Static_assert
 #endif
 
--- a/share/man/man3/assert.3
+++ b/share/man/man3/assert.3
@@ -25,7 +25,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd April 20, 2021
+.Dd May 17, 2026
 .Dt ASSERT 3
 .Os
 .Sh NAME
@@ -40,15 +40,14 @@
 .Sh DESCRIPTION
 The
 .Fn assert
-macro tests the given
-.Ar expression
+macro tests the given scalar
+.Ar expression ,
 and if it is false,
-the calling process is terminated.
-A diagnostic message is written to
+a diagnostic message is written to
 .Dv stderr
 and the function
 .Xr abort 3
-is called, effectively terminating the program.
+is called, effectively terminating the calling process.
 .Pp
 If
 .Ar expression
@@ -57,6 +56,24 @@ the
 .Fn assert
 macro does nothing.
 .Pp
+The
+.Fn assert
+macro accepts a variable number of arguments.
+In all compilation modes,
+.Fn assert
+is defined as a macro with an ellipsis parameter, consistent with the
+C23 standard.
+This allows expressions containing commas to be passed directly without
+requiring an extra pair of enclosing parentheses.
+Only a single scalar expression is evaluated.
+Supplying multiple arguments is prohibited, and hence top-level comma
+operators.
+In particular, this guards against accidentally writing
+.Fn assert
+in the style of
+.Fn static_assert ,
+which would otherwise silently evaluate as always true via the comma
+operator.
 The
 .Fn assert
 macro
@@ -82,9 +99,9 @@ macro should only be used for ensuring the developer's expectations
 hold true.
 It is not appropriate for regular run-time error detection.
 .Pp
-The
+In pre-C23 compilation modes
 .Fn static_assert
-macro expands to
+is implemented as a macro and expands to
 .Fn _Static_assert ,
 and, contrarily to
 .Fn assert ,
@@ -94,8 +111,14 @@ message including the string literal message, if provided.
 The initial form of the
 .Fn _Static_assert
 containing a string literal message was introduced in C11 standard, and
-the other form with no string literal is to be implemented by C2x and
-some compilers may lack its adoption at present.
+the other form with no string literal conforms to C23 standard.
+.Pp
+In C23 and later,
+.Fn static_assert
+is a language keyword, and
+.Fn _Static_assert
+is provided as an obsolescent alternative spelling that should not be
+used for new code and development.
 .Sh EXAMPLES
 The assertion:
 .Dl "assert(1 == 0);"
@@ -111,6 +134,17 @@ Second, the code will disappear if
 .Dv NDEBUG
 is defined, changing the semantics of the program.
 .Pp
+The following example asserts that the
+.Va iov_len
+member of the compound literal, reflecting the value of
+.Va len ,
+is non-zero.
+The compound literal contains a comma that is not protected by
+parentheses, which the variadic
+.Fn assert
+macro handles transparently:
+.Dl assert((struct iovec){ buf, len }.iov_len);
+.Pp
 The following asserts that the size of the S structure is 16.
 Otherwise, it produces a diagnostic message which points at the
 constraint and includes the provided string literal:
@@ -123,14 +157,25 @@ If none is provided, it only points at the constraint.
 The
 .Fn assert
 macro conforms to
-.St -isoC-99 .
+.St -isoC-2023 .
 .Pp
 The
 .Fn static_assert
-macro conforms to
-.St -isoC-2011 .
+conforms to
+.St -isoC-2011 ,
+wherein defined as a macro, and
+.St -isoC-2023 ,
+as a language keyword, depending on the compilation mode.
 .Sh HISTORY
-An
+The
 .Nm
-macro appeared in
+macro first appeared in
 .At v7 .
+Starting with
+.Fx 15.2 ,
+it accepts a variadic argument list, allowing expressions
+containing commas, such as compound literals, to be passed
+without requiring extra enclosing parentheses.
+This conforms to
+.St -isoC-2023 ,
+but made available in all compilation modes.
--- a/share/man/man3/assert.3
+++ b/share/man/man3/assert.3
@@ -145,7 +145,9 @@ parentheses, which the variadic
 macro handles transparently:
 .Dl assert((struct iovec){ buf, len }.iov_len);
 .Pp
-The following asserts that the size of the S structure is 16.
+The following asserts that the size of the
+.Vt S
+structure is 16.
 Otherwise, it produces a diagnostic message which points at the
 constraint and includes the provided string literal:
 .Dl "static_assert(sizeof(struct S) == 16, ""size mismatch"");"
--- a/contrib/ldns/edns.c
+++ b/contrib/ldns/edns.c
@@ -10,6 +10,7 @@
  * See the file LICENSE for the license
  */
 
+#include <ldns/config.h>
 #include <ldns/ldns.h>
 
 #define LDNS_OPTIONLIST_INIT 8
--- a/include/rpcsvc/yp_prot.h
+++ b/include/rpcsvc/yp_prot.h
@@ -67,11 +67,6 @@
  * YPPROC_MAPLIST	takes (char *), returns (struct ypmaplist *).
  */
 
-#ifndef BOOL_DEFINED
-typedef u_int bool;
-#define BOOL_DEFINED
-#endif
-
 /* Program and version symbols, magic numbers */
 
 #define YPPROG		((u_long)100004)
--- a/sbin/mdmfs/mdmfs.c
+++ b/sbin/mdmfs/mdmfs.c
@@ -49,13 +49,12 @@
 #include <paths.h>
 #include <pwd.h>
 #include <stdarg.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <ctype.h>
 #include <unistd.h>
-
-typedef enum { false, true } bool;
 
 struct mtpt_info {
 	uid_t		 mi_uid;


clausecker:
CC @bsdimp 

Some thoughts on this one:

 - I think we can safely define assert to take the variable argument list in all modes as an extension, preventing friction when code bases are gradually extended to C23.
 - not sure what `_assert` is.  Should we touch it at all?
 - document `__STDC_VERSION_ASSERT_H__` please
 - your documentation change is somewhat obtuse.  It “takes a variable argument list,” not “takes an ellipsis.”  But even with that change, the average programmer will have a hard time understanding what that means.  Perhaps it would be better to explain that starting with C23, the macro is set up such that expressions including the comma operator can be used, whereas in previous versions of the C standard, such expressions would cause syntax error.  Give examples and explain how to work around this shortcoming in older language revisions (by using double parentheses)
 - if we chose to have `assert(...)` in all modes (which I recommend), document that this is supported as an extension
 - please check the output of your manpage with the `mandoc` command.  In particular, `.Fn` just produces the argument typeset as a function name.  It does not write “function” before the function name, so if you write

       The
       .Fn foo
       is bar

   you get something like “the *foo* is bar,” which is nonsensical.  Try writing

       The
       .Fn foo
       macro is bar

   or something similar.

kfv:
> * I think we can safely define assert to take the variable argument list in all modes as an extension, preventing friction when code bases are gradually extended to C23.

Agreed. Done.

> * not sure what `_assert` is.  Should we touch it at all?

Not sure if we should still keep it, but I think its sole purpose is backward compatibility as same-name functions with a leading underscore were seen as implementation internals back in early days of UNIX. I was not born back then, though, so I could be completely wrong.

> * document `__STDC_VERSION_ASSERT_H__` please

Why should we document feature test macros? That's not something user needs to be concerned with.

> * your documentation change is somewhat obtuse.  It “takes a variable argument list,” not “takes an ellipsis.”  But even with that change, the average programmer will have a hard time understanding what that means. 

It is not wrong, I believe. I've frequently seen "macros with an ellipsis parameter", "variadic macros", and similar terms—especially in the standard, proposal documents, and compiler documentation. But I understand your point, I'll take care of it shortly.

> Perhaps it would be better to explain that starting with C23, the macro is set up such that expressions including the comma operator can be used, whereas in previous versions of the C standard, such expressions would cause syntax error.  Give examples and explain how to work around this shortcoming in older language revisions (by using double parentheses)

Right, I'll take care of it tomorrow.

> * if we chose to have `assert(...)` in all modes (which I recommend), document that this is supported as an extension

Sure, I'll add a note for that.

> * please check the output of your manpage with the `mandoc` command.  In particular, `.Fn` just produces the argument typeset as a function name.  It does not write “function” before the function name, so if you write
>   ```
>   The
>   .Fn foo
>   is bar
>   ```
>     
>   you get something like “the _foo_ is bar,” which is nonsensical.  Try writing
>   ```
>   The
>   .Fn foo
>   macro is bar
>   ```
>  
>   or something similar.

I deliberately didn't use it for lines 105-107, but you're right about line 80. Sure. I'll do an update on the manpage tomorrow evening (better phrasing, new examples, etc.)

kfv:
By the way, please accept my apologies if the branch is named `libc/__STDC_VERSION_XXXX_H__`. I started out adding missing feature test macros but ended up taking care of `<assert.h>`.

clausecker:
> Why should we document feature test macros? That's not something user needs to be concerned with.

The macro is part of the C standard, but in retrospect I didn't add it to `<stdbit.h>` either, so maybe it's ok to not have it?

> It is not wrong, I believe. I've frequently seen "macros with an ellipsis parameter", "variadic macros", and similar terms—especially in the standard, proposal documents, and compiler documentation. But I understand your point, I'll take care of it shortly.

The important thing is that whatever term we use is consistent within FreeBSD's documentation.  Perhaps @concussious can say more here.

Looking forwards to your updates and don't worry about the branch name.

clausecker:
Another thing: please add to HISTORY the version in which `assert()` was extended to handle commas in its argument list (this should indicate 15.2 as the first version with the feature).

kfv:
> The macro is part of the C standard, but in retrospect I didn't add it to `<stdbit.h>` either, so maybe it's ok to not have it?

Conforming implementations must define `__STDC_VERSION_XXXX_H__` macros for all headers listed in the standard, so omitting them is not acceptable, I believe. [You also have it set correctly in <stdbit.h>](https://github.com/freebsd/freebsd-src/blob/30f500e97b57d36ebb8483cb8b8921507350cd0f/include/stdbit.h#L7-L8). For more details, see the final edition of [C23 Implications for C Libraries](https://gustedt.gitlabpages.inria.fr/c23-library/) by Jens Gustedt.

In the meantime, I am working on a separate PR to address the remaining missing `__STDC_VERSION_XXXX_H__` macros across the tree and will submit it within the next few hours.

As mentioned in the shared document, the macro can be defined before full conformance (e.g. as an include guard only) as long as its value is empty or a smaller number like `0`. Once that header is fully conforming to C23, the value must be `202311L`.

> The important thing is that whatever term we use is consistent within FreeBSD's documentation. Perhaps @concussious can say more here.

You're right, it now makes complete sense to me. Fair point. Thanks.

> Looking forwards to your updates and don't worry about the branch name.

Thanks, Robert.

> Another thing: please add to HISTORY the version in which assert() was extended to handle commas in its argument list (this should indicate 15.2 as the first version with the feature).

Sure, will do.

kfv:
> In the meantime, I am working on a separate PR to address the remaining missing `__STDC_VERSION_XXXX_H__` macros across the tree and will submit it within the next few hours.

Just opened the [PR](#2206) as a draft. I'll add description and mark it as ready for review once my tests complete, but happy to hear any early feedback or requests in the meantime.

bms:
OT: An adjacent data point to consider for assert(): if you're profiling code bases for source code complexity metric, allegedly its use bloats the SonarSource metric. as measured by clang-tidy, and this requires a specific workaround. Found because I've been doing background research to support falsifying certain other... hypotheses.

kfv:
> OT: An adjacent data point to consider for assert(): if you're profiling code bases for source code complexity metric, allegedly its use bloats the SonarSource metric. as measured by clang-tidy, and this requires a specific workaround. Found because I've been doing background research to support falsifying certain other... hypotheses.

Interesting! Thanks for the point. I'd love to hear more about your findings.

kfv:
Addressed Robert's notes as best I could. I kept the ellipsis term but tried to clarify the macro signature description around it (I'm open to removing it altogether, but couldn't think of a better phrasing, to be honest). Let me know if you any suggestions.

Cc: @clausecker, @concussious

include/assert.h
```diff
@@ -36,6 +36,8 @@
 
 #include <sys/cdefs.h>
 
+#include <stdbool.h>
```
clausecker: This is name space pollution.  Can we make do without including `stdbool.h`?

include/assert.h
```diff
@@ -46,23 +48,24 @@
...
 #endif	/* __BSD_VISIBLE */
 #else
-#define	_assert(e)	assert(e)
-
-#define	assert(e)	((e) ? (void)0 : __assert(__func__, __FILE__, \
-			    __LINE__, #e))
+#define	assert(...)	((void)sizeof(((bool(*)(bool))0)(__VA_ARGS__)),  \
+			    (__VA_ARGS__) ? (void)0 :                    \
+			    __assert(__func__, __FILE__,                 \
+			    __LINE__, #__VA_ARGS__))
```
clausecker: Could you explain the purpose if this construction?  It looks awfully complicated.  What was wrong with the previous approach, that is

    ((__VA_ARGS__) ? (void)0 : __assert(__func__, __FILE__, __LINE__, #__VA_ARGS__))

include/assert.h
```diff
@@ -72,13 +75,16 @@
  * C++ template parameters may contain commas, even if not enclosed in
  * parentheses, causing the _Static_assert macro to be invoked with more
  * than two parameters.
+ *
+ * C23 defines static_assert and its obsolescent alternative spelling,
+ * _Static_assert, as keywords.
  */
-#if __ISO_C_VISIBLE >= 2011 && !defined(__cplusplus)
+#if __ISO_C_VISIBLE >= 2011 && __ISO_C_VISIBLE <= 2023 && !defined(__cplusplus)
```
clausecker: You probably want

    __ISO_C_VISIBLE < 2023

so that the macro is invisible in C23 mode.

share/man/man3/assert.3
```diff
@@ -59,6 +58,24 @@ macro does nothing.
...
+requiring an extra pair of enclosing parentheses.
+Only a single scalar expression is evaluated.
+Supplying multiple arguments is prohibited, and hence top-level comma
+operators.
+In particular, this guards against accidentally writing
+.Fn assert
+in the style of
+.Fn static_assert ,
+which would otherwise silently evaluate as always true via the comma
+operator.
```
clausecker, line 76: Is this in accordance with C23?  I see no such rule in that standard.

share/man/man3/assert.3
```diff
@@ -82,9 +99,9 @@ macro should only be used for ensuring the developer's expectations
 hold true.
 It is not appropriate for regular run-time error detection.
 .Pp
-The
+In pre-C23 compilation modes, the
```
clausecker: Please remove `, the` as it is incorrect English grammar.

share/man/man3/assert.3
```diff
@@ -111,6 +134,17 @@ Second, the code will disappear if
 .Dv NDEBUG
 is defined, changing the semantics of the program.
 .Pp
+The following pedagogical example asserts that the
```
clausecker: Please remove `pedagogical `; it's an example, 'nuff said.

share/man/man3/assert.3
```diff
@@ -111,6 +134,17 @@ Second, the code will disappear if
...
+member of the compound literal, reflecting the value of
+.Va len ,
+is non-zero.
+The compound literal contains a comma that is not protected by
+parentheses, which the variadic
+.Fn assert
+macro handles transparently:
+.Dl assert((struct iovec){ buf, len }.iov_len);
+.Pp
 The following asserts that the size of the S structure is 16.
```
clausecker: Please use `.Vn` for `S`.

share/man/man3/assert.3
```diff
@@ -123,14 +157,28 @@ If none is provided, it only points at the constraint.
...
 .Nm
-macro appeared in
+macro first appeared in
 .At v7 .
+Starting with
+.Fx 15.2 ,
+it accepts a variadic argument list, allowing expressions
+containing commas, such as compound literals, to be passed
+without requiring extra enclosing parentheses.
+This is conforming to
```
clausecker: “This conforms to”

share/man/man3/assert.3
```diff
@@ -123,14 +157,28 @@ If none is provided, it only points at the constraint.
 The
 .Fn assert
 macro conforms to
-.St -isoC-99 .
+.St -isoC-99
+and
+.St -isoC-2023 ,
+depending on the compilation mode.
```
clausecker: As you removed the functionality switch, you can just say

```suggestion
The
.Fn assert
maco conforms to
.St -isoC-2023 .
```

clausecker (commented):
More corrections.

Is the complicated definition of `assert()` something Jens Gustedt cooked up?

kfv:
> Is the complicated definition of `assert()` something Jens Gustedt cooked up?

No, Peter Sommerlad.

include/assert.h
kfv: Peter Sommerlad explicitly explains [(§5.3)](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2829.htm#using-the-comma-operator-can-be-misapplied-to-an-always-true-assert-if-its-arguments-are-formed-as-for-static_assertconditionreason.-this-will-make-wrong-code-compile-that-today-doesnt.) that top-level commas could produce always-true assertions, and he addresses this with a proposed solution in [§9 Example implementation](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2829.htm#example-implementation-bsd-license-on-macos). The following excerpt from the [§3 Remedy](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2829.htm#remedy) section also covers the issue:

> The proposed solution prevents the use of the comma operator on top-level, to avoid accidentally creating always true assertions like the following:
>
>    assert(x > 0 , "x was not greater than zero");
>
>Those assertions can result from converting a static_assert to a regular assert.

So although this indeed does complicate what would otherwise be a simple macro, it is absolutely necessary to remain conformant.

clausecker: Ok, but the proposed wording doesn't say that always true assertions are forbidden.  The rational may do so, but the wording doesn't.  So I'm not sure if the proposed implementation is correct for the wording in the standard and would be inclined to say that it isn't.

clausecker: In particular, the draft says:

> Note: the following feature was removed, due to discussions of the R0/N2621 version of this paper. And a mechanism is enforced to detect misuse of the comma operator.

So no, this is not actually part of the standard.  Please simplify.

kfv: I disagree. The proposal discusses removing the behaviour that allows the variadic macro form of assert to accept additional diagnostic text via a top-level comma operator, as in:

> assert((void)"this cannot be true", -0.0);

See [P2264R0](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2264r0.html) of the proposal in the WG21 2020 papers.

So, my apologies, Robert, but I still believe we definitely need this.

Note that the standard only defines `assert()` with a single scalar expression. Furthermore, §7.2.1.1 explicitly states:

> NOTE Nevertheless, when NDEBUG is not defined, the macro acts as a function taking one parameter as indicated by the prototype given later in this subclause (7.2.2.1). For both assert() and assert(1, 1), the number of arguments does not agree with the number of parameters. A diagnostic is required by 6.5.3.3.

clausecker: Ok right, convinced.  Now can we do that without including `stdbool.h`?  Perhaps it suffices to just spell the type `_Bool` for C11 compliance.

share/man/man3/assert.3
clausecker: Resolved as per other discussion.

include/assert.h
kfv: Whoops! Good catch. Thanks.

clausecker:
@kfv Please try to avoid force-pushing unless you merely do a rebase.  Force pushes make it really hard to see what changes you did relatively to the previous state.  It's better if you push fix-up commits; we'll squash them on commit.

include/assert.h
kfv: Makes sense, we can change it to `bool` again once we default to C23 compilation mode later.

include/assert.h
kfv: Resolved as per another [trhead](https://github.com/freebsd/freebsd-src/pull/2203#discussion_r3288639161).

include/assert.h
clausecker: I think we'll have to stick with `_Bool` so `-std=c11` still works.  Not sure if this change works with C99 code though; might require more work.  I'll go do some tests.

share/man/man3/assert.3
kfv: Although I didn't quite understand why the definite article is considered wrong there, I did as requested.

clausecker: The definite article is usually not used with proper names (which `static_assert` is in this context), though there are exceptions.  It would be used in a  phrase like “the static_assert keyword” but not if you leave “keyword” out, as then the referent becomes a proper name.

kfv:
> @kfv Please try to avoid force-pushing unless you merely do a rebase. Force pushes make it really hard to see what changes you did relatively to the previous state. It's better if you push fix-up commits; we'll squash them on commit.

Understood, sure, will do.

share/man/man3/assert.3
kfv: Did you mean `.Vt`? `.Vn` doesn't appear to exist in [mdoc(7)](https://man.freebsd.org/cgi/man.cgi?query=mdoc). I used `.Vt` here, though `.Sy` would also be reasonable — happy to switch if you have a preference. By the way, I didn't fold it into a `fixup` since the file was untouched by my previous commits.

clausecker: Oh sorry, I meant `.Va` for variable name.  `.Vt` is for variable types, though I think you are right that it is better.  Not sure in any case.

clausecker (approved):
I think this looks pretty good.
Will do some final checks if this works in C99 mode, too, and if it does, land the change set.

share/man/man3/assert.3
kfv: Oh, that's meticulous. You're right. Thanks, Robert.

bsdimp:
I'm almost positive `_Bool` will work in c99 and c89 compilation environments. Both gcc and clang have these always on. They do for other __foo things.

clausecker:
Okay, this patch set fails during buildworld, as C++ does not have the `_Bool` spelling for `bool`.  Please amend the patch set so that in C++ mode, the type is spelled `bool` (or whatever is needed for C++).  Please also make sure to run a `make buildkernel buildworld` on each patch revision to catch such issues.

clausecker:
Note that just changing the spelling to `bool` does not seem to cut it.  LLVM then fails to build as follows:

```
--- Support/CommandLine.o ---
/usr/home/fuz/src/repo/freebsd-src/contrib/llvm-project/llvm/lib/Support/CommandLine.cpp:331:21: error: lambda expression in an unevaluated operand
  331 |                     [cat](const OptionCategory *Category) {
      |                     ^
/usr/home/fuz/src/repo/freebsd-src/contrib/llvm-project/llvm/lib/Support/CommandLine.cpp:341:21: error: lambda expression in an unevaluated operand
  341 |                     [sub](const SubCommand *Sub) {
      |                     ^
```

Please check.

clausecker:
I also get, further down:

```
/usr/home/fuz/src/repo/freebsd-src/contrib/llvm-project/compiler-rt/lib/builtins/cpu_model/x86.c:1160:3: error: 
      call to undeclared function 'static_assert'; ISO C99 and later do not support implicit function declarations
      [-Werror,-Wimplicit-function-declaration]
 1160 |   static_assert(sizeof(Features) / sizeof(Features[0]) == 4, "");
      |   ^
1 error generated.
```

This file is compiled with `-std=gnu17`.

kfv:
> Okay, this patch set fails during buildworld, as C++ does not have the `_Bool` spelling for `bool`. Please amend the patch set so that in C++ mode, the type is spelled `bool` (or whatever is needed for C++). Please also make sure to run a `make buildkernel buildworld` on each patch revision to catch such issues.

You're right, apologies, Robert. I ran `buildworld` on earlier revisions but forgot to update the final diff. That was my oversight.

> Note that just changing the spelling to `bool` does not seem to cut it. LLVM then fails to build as follows:
> 
> ```
> --- Support/CommandLine.o ---
> /usr/home/fuz/src/repo/freebsd-src/contrib/llvm-project/llvm/lib/Support/CommandLine.cpp:331:21: error: lambda expression in an unevaluated operand
>   331 |                     [cat](const OptionCategory *Category) {
>       |                     ^
> /usr/home/fuz/src/repo/freebsd-src/contrib/llvm-project/llvm/lib/Support/CommandLine.cpp:341:21: error: lambda expression in an unevaluated operand
>   341 |                     [sub](const SubCommand *Sub) {
>       |                     ^
> ```
> 
> Please check.

I'll push a fix shortly. I had completely overlooked C++ lambda expressions; the sanitisation approach needs a small adjustment to handle them correctly.

> I also get, further down:
> 
> ```
> /usr/home/fuz/src/repo/freebsd-src/contrib/llvm-project/compiler-rt/lib/builtins/cpu_model/x86.c:1160:3: error: 
>       call to undeclared function 'static_assert'; ISO C99 and later do not support implicit function declarations
>       [-Werror,-Wimplicit-function-declaration]
>  1160 |   static_assert(sizeof(Features) / sizeof(Features[0]) == 4, "");
>       |   ^
> 1 error generated.
> ```
> 
> This file is compiled with `-std=gnu17`.

Whoops! C23 check shall use `__STDC_VERSION__` rather than `__ISO_C_VISIBLE`.

kfv:
I'll push my fixup commits shortly. Currently I'm dealing with a few more cases. I'll update you once everything succeeds.

kfv:
Addressed all issues. I'll do one more test and push in a few hours.

kfv:
First and foremost, apologies for the force push — it was necessary to reorder the commits so the PR and commit history look sensible at a glance. Quite a bit has changed since I started targeting stricter conformance.

To avoid namespace pollution and unnecessary includes as discussed in comments, I added two distinct cases for C and C++, so both `bool` and `_Bool` remain correct under our current default mode where the former is not yet treated as a keyword in C.

I also had to make sure no breaking change is introduced in C++ modes prior to C++20, where we prefer working code over strict conformance. In those modes, `sizeof` cannot accept a lambda expression, and hence causing failures across the tree.

Candidates like `(void)bool(expression)` or `static_cast<bool>(expression)` were tried, but both break cases where double evaluation is not acceptable — such as with `std::set<int>` conditions. I’m currently convinced that falling back to `((void)0)` to keep things working is the best option available, but I should note that I'm not an expert C++ developer and my understanding of it is largely limited to studying the standard for the sake of interoperability and compatibility with C code, so I’d love to learn if there's a hack or workaround available in pre-C++20 modes that could covers all three desirable properties.

Finally, alongside the `assert.h` changes, I cleaned up a few spots in the tree where legacy `bool` definitions existed. Removing them is not only necessary for this patch set to land, but also helps pave the way for smoother upgrades in future by eliminating potential collisions and ensuring compatibility.

I must confess that after my first successful `buildworld`/`buildkernel` before our recent improvements, I didn't expect such small additions to break so much, and that was an oversight on my part underestimating the side-effects of foundational interfaces, my apologies for that. I'll do my best to ensure my upcoming contributions are better shaped and thought through, and of course fully tested before any iteration.

Please let me know if anything else needs to be changed or improved.

include/assert.h
```diff
@@ -46,19 +46,46 @@
...
+ * Since no approach satisfies all three below C++20, the least harmful
+ * choice is to forgo the check entirely rather than silently break one
+ * of the remaining guarantees.
+ *
+ */
+#define __assert_sanitize(...)	((void)0)
+#else
+#define __assert_sanitize(...)	(void)sizeof(((bool(*)(bool))0)(__VA_ARGS__))
+#endif /* __cplusplus < 202002L */
+#else
```
brooksdavis, line 78: Would benefit from a comment

brooksdavis (commented):
The feature test macros should probably be added last rather than before the features are updated.

The main commit making assert() variadic is made harder to review by including unrelated whitespace changes.  Please make those first. I found the use of comments on `#else` vs `#endif` inconsistent.

bsdimp:
@kfv Forced pushes for pull requests are routine and expected, so no worries there.

kfv:
> The feature test macros should probably be added last rather than before the features are updated.

Thanks for the catch! Done.

> The main commit making assert() variadic is made harder to review by including unrelated whitespace changes. Please make those first.

Fair point. I had originally intended to do exactly that but wasn't sure whether a standalone stylistic commit purely for whitespace consistency and `style(9)`-conformance would be considered noise in the history. Done. I'd just possibly rather squashing it into the following commit before it's merged, unless you'd prefer to keep them separate.

> I found the use of comments on `#else` vs `#endif` inconsistent.

I may be missing something, but to my eye the treatment is consistent: in all non-trivial blocks, every `#endif` carries a comment identifying its opening condition, and no `#else` branch does. The one `#endif` without an inline comment is the one closing the `static_assert` guard, though that block already has a thorough comment directly above it. Should I add one there too for uniformity?

One thing I did fix separately though, was a stray trailing blank line inside the C++ modes comment block I wrote.

include/assert.h
kfv: Doesn't the earlier comment block cover it already?

brooksdavis:
> > The main commit making assert() variadic is made harder to review by including unrelated whitespace changes. Please make those first.
> 
> Fair point. I had originally intended to do exactly that but wasn't sure whether a standalone stylistic commit purely for whitespace consistency and `style(9)`-conformance would be considered noise in the history. Done. I'd just possibly rather squashing it into the following commit before it's merged, unless you'd prefer to keep them separate.

There's a historic preference for keeping them separate. You can also add no-op commits to .git-blame-ignore-revs in a future pull request to cause them to vanish from git blame (once suitably configured).

> > I found the use of comments on `#else` vs `#endif` inconsistent.
> 
> I may be missing something, but to my eye the treatment is consistent: in all non-trivial blocks, every `#endif` carries a comment identifying its opening condition, and no `#else` branch does. The one `#endif` without an inline comment is the one closing the `static_assert` guard, though that block already has a thorough comment directly above it. Should I add one there too for uniformity?

"inconsistent" was perhaps poor word choice. I think the `#else` statements would benefit from comments as I found the unindented blocks hard to follow across the large comment.

clausecker:
Landed with some minor touchups to the man page.  
I eventually decided to remove the sentence “The `assert` macro accepts a variable number of arguments.” as it is confusing in context and seemingly contradicts the following text.

clausecker:
As per @kostikbel's comments on the mailing list, I'll revert the whitespace changes.  style(9) is clear that we put a tab between `#define` and the macro name, so that should not have been removed.

kfv:
[Excerpt from style(9)](https://github.com/freebsd/freebsd-src/blob/439710cf003b7849d2119e9a7694d7b27ab6d70f/share/man/man9/style.9#L206-L208):

> Put a single space or tab character between the `#define` and the macro name, but be consistent within a file.

I'm curious how everyone became convinced of that. I'd appreciate it if you could educate me.

kfv:
I wish we could have discussed it first. I still believe that commit was neither wrong nor unnecessary.

Don't get me wrong, I'm completely fine with reverting a commit when it's necessary, regardless of whether it's mine or someone else's. But in this case, I don't think that's justified. Either we should clarify the rule and improve the wording in style(9), or the change simply wasn't wrong to revert. Besides, I'm fairly sure there are several places in the source tree where the gap between the directive and the macro name is a single space rather than a tab.

`#define` occupies seven columns, so tabs and spaces are not visually distinguishable in that context. It can be frustrating when you're working on a file and your diff ends up full of misaligned or distorted lines caused by inconsistent formatting.

Although I believe tabs in this specific case can be visually confusing, I don't have any issue with them in general. I only care about consistency.

You might also want to try the following ad hoc check:

```
find /usr/src/lib/libc                                           \
  -type f \(                                                     \
    -name '*.c' -o -name '*.h' -o -name '*.cpp' -o -name '*.hpp' \
  \)                                                             \
  -exec awk '
    /^#define[ \t]/ {
      if ($0 ~ /^#define\t/) tab++
      else if ($0 ~ /^#define /) space++
    }
    END {
      printf "single tab:   %d\n", tab
      printf "single space: %d\n", space
    }
  ' {} +
```

On my side, this produces the following results at the time of writing:

```
single tab:   1420
single space: 2628
```

In any case, it's fine. Thanks for keeping me informed.

clausecker:
style(9) says:

> Put a single space or tab character between the #define and the macro name, but be consistent within a file.

It used to be that you were supposed to only use a tab, but the rules were recently relaxed.  In particular, if a file already uses tabs after `#define`, stick to that instead of changing it.  If the output looks weird, adjust the tab settings of your editor.

I'm somewhat relaxed about this stuff, but some people care a lot more.

kfv:
That’s true, `#define _ASSERT_H_` was the only line using a single space, IIRC.

As always, I do really appreciate genuine and kind support. Thanks, Robert!

Cheers.

clausecker:
There has been more discussion on your style change: https://reviews.freebsd.org/D57391

kfv:
Interesting! Thanks for letting me know, Robert.

DimitryAndric:
This PR still needs some adjustment to make C++20 work correctly. After the changes in assert.h, the following program:

```c++
#include <cassert>
#include <memory>

void f(std::unique_ptr<int> p)
{
 assert(p);
}
```

fails to compile, with:

```
unique-ptr-assert.cpp:6:10: error: no viable conversion from 'std::unique_ptr<int>' to 'bool'
   6 |   assert(p);
     |          ^
/usr/include/c++/v1/__memory/unique_ptr.h:276:64: note: explicit conversion function is not a candidate
 276 |   _LIBCPP_HIDE_FROM_ABI _LIBCPP_CONSTEXPR_SINCE_CXX23 explicit operator bool() const _NOEXCEPT {
     |                                                                ^
```

When compiled with -save-temps, it shows the eventual expression it dies on:

```
unique-ptr-assert.cpp:6:36: error: no viable conversion from 'std::unique_ptr<int>' to 'bool'
   6 |   ((void)sizeof(((bool(*)(bool))0)(p)), (p) ? (void)0 : __assert(__func__, "unique-ptr-assert.cpp", 6, "p"));
     |                                    ^
```

kfv:
Hey, I'll take care of it tomorrow. Thanks for the report.

clausecker:
See also [D57449](https://reviews.freebsd.org/D57449) where a somewhat drastic fix is proposed.

kfv:
Thanks for letting me know. I just left a comment there as well. Please allow me a day or two and we'll have it sorted.

PR #2205, opened 2026-05-18
 gpio: add generic Intel GPIO pin controller framework

ckoehne:
Add a platform-independent driver framework for Intel GPIO pin controllers
found on modern Intel SoCs. The driver accesses GPIO pad registers through
ACPI-provided memory-mapped resources and implements the gpio interface [1]
including pin enumeration, capability reporting, configuration, and
read/write/toggle operations. A common data model of communities and pad groups
allows individual SoC-specific drivers to supply their own pad tables and ACPI
hardware IDs while sharing all register-level logic.

[1] https://wiki.freebsd.org/GPIO

### DIFF
--- a/sys/dev/gpio/intel/adlngpio.c
+++ b/sys/dev/gpio/intel/adlngpio.c
@@ -1,0 +1,91 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2026 Beckhoff Automation GmbH & Co. KG
+ */
+
+#include <sys/param.h>
+#include <sys/systm.h>
+#include <sys/bus.h>
+#include <sys/gpio.h>
+#include <sys/kernel.h>
+#include <sys/module.h>
+
+#include <dev/gpio/gpiobusvar.h>
+
+#include "intelgpio.h"
+
+#include "gpio_if.h"
+#include "opt_acpi.h"
+
+static const struct intelgpio_padgroup adln_com0_groups[] = {
+	{ .first_pad = 0, .npads = 26, .gpio_base = 0, .name = "GPP_B" },
+	{ .first_pad = 26, .npads = 16, .gpio_base = 32, .name = "GPP_T" },
+	{ .first_pad = 42, .npads = 25, .gpio_base = 64, .name = "GPP_A" },
+};
+
+static const struct intelgpio_padgroup adln_com1_groups[] = {
+	{ .first_pad = 67, .npads = 8, .gpio_base = 96, .name = "GPP_S" },
+	{ .first_pad = 75, .npads = 20, .gpio_base = 128, .name = "GPP_I" },
+	{ .first_pad = 95, .npads = 24, .gpio_base = 160, .name = "GPP_H" },
+	{ .first_pad = 119, .npads = 21, .gpio_base = 192, .name = "GPP_D" },
+	{ .first_pad = 140, .npads = 29, .gpio_base = 224, .name = "vGPIO" },
+};
+
+static const struct intelgpio_padgroup adln_com4_groups[] = {
+	{ .first_pad = 169, .npads = 24, .gpio_base = 256, .name = "GPP_C" },
+	{ .first_pad = 193, .npads = 25, .gpio_base = 288, .name = "GPP_F" },
+	{ .first_pad = 218,
+	    .npads = 6,
+	    .gpio_base = INTELGPIO_GPIO_NOMAP,
+	    .name = "HVCMOS" },
+	{ .first_pad = 224, .npads = 25, .gpio_base = 320, .name = "GPP_E" },
+};
+
+static const struct intelgpio_padgroup adln_com5_groups[] = {
+	{ .first_pad = 249, .npads = 8, .gpio_base = 352, .name = "GPP_R" },
+};
+
+static const struct intelgpio_community adln_communities[] = {
+	{ .ngroups = nitems(adln_com0_groups), .groups = adln_com0_groups },
+	{ .ngroups = nitems(adln_com1_groups), .groups = adln_com1_groups },
+	{ .ngroups = nitems(adln_com4_groups), .groups = adln_com4_groups },
+	{ .ngroups = nitems(adln_com5_groups), .groups = adln_com5_groups },
+};
+
+static char *adln_hids[] = { "INTC1056", "INTC1057", "INTC1085", NULL };
+
+static const struct intelgpio_platform adln_platform = {
+	.communities = adln_communities,
+	.ncommunities = nitems(adln_communities),
+	.hids = adln_hids,
+	.desc = "Intel Alder Lake-N GPIO",
+};
+
+static int
+adln_probe(device_t dev)
+{
+	return (intelgpio_probe(dev, &adln_platform));
+}
+
+static int
+adln_attach(device_t dev)
+{
+	return (intelgpio_attach(dev, &adln_platform));
+}
+
+static device_method_t adln_methods[] = {
+	DEVMETHOD(device_probe, adln_probe),
+	DEVMETHOD(device_attach, adln_attach),
+
+	DEVMETHOD_END
+};
+
+DEFINE_CLASS_1(intelgpio, adln_driver, adln_methods,
+    sizeof(struct intelgpio_softc), intelgpio_driver);
+
+DRIVER_MODULE(adlngpio, acpi, adln_driver, NULL, NULL);
+MODULE_DEPEND(adlngpio, acpi, 1, 1, 1);
+MODULE_DEPEND(adlngpio, gpiobus, 1, 1, 1);
+MODULE_VERSION(adlngpio, 1);
+ACPI_PNP_INFO(adln_hids);
--- a/sys/dev/gpio/intel/intelgpio.c
+++ b/sys/dev/gpio/intel/intelgpio.c
@@ -1,0 +1,415 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2026 Beckhoff Automation GmbH & Co. KG
+ */
+
+#include <sys/param.h>
+#include <sys/systm.h>
+#include <sys/bus.h>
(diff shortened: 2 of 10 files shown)


ckoehne:
@bsdjhb @VexedUXR I'm not sure if you're able to review this. However, I've found you a few times in the git history of dev/gpio.

sys/modules/Makefile
```diff
@@ -400,7 +400,11 @@ SUBDIR=	\
 	sysvipc \
 	tarfs \
 	tcp \
+<<<<<<< HEAD
 	${_thunderbolt} \
+=======
+	${_tglhgpio} \
+>>>>>>> 96a13e988d65 (gpio: add Intel Tiger Lake-H GPIO driver)
```
VexedUXR: Merge conflict

sys/dev/gpio/intel/intelgpio.h
```diff
@@ -0,0 +1,116 @@
...
+	DEVMETHOD(device_detach,	intelgpio_detach),		\
+	DEVMETHOD(gpio_get_bus,		intelgpio_get_bus),		\
+	DEVMETHOD(gpio_pin_max,		intelgpio_pin_max),		\
+	DEVMETHOD(gpio_pin_getname,	intelgpio_pin_getname),		\
+	DEVMETHOD(gpio_pin_getcaps,	intelgpio_pin_getcaps),		\
+	DEVMETHOD(gpio_pin_getflags,	intelgpio_pin_getflags),		\
+	DEVMETHOD(gpio_pin_setflags,	intelgpio_pin_setflags),		\
+	DEVMETHOD(gpio_pin_get,		intelgpio_pin_get),		\
+	DEVMETHOD(gpio_pin_set,		intelgpio_pin_set),		\
+	DEVMETHOD(gpio_pin_toggle,	intelgpio_pin_toggle)
```
VexedUXR: IMO these should be implemented as a base class which the other drivers can inherit from.

sys/dev/gpio/intel/intelgpio.c
```diff
@@ -0,0 +1,393 @@
...
+int
+intelgpio_probe(device_t dev, const struct intelgpio_platform *plat)
+{
+	int rv;
+
+	if (acpi_disabled("intelgpio"))
+		return (ENXIO);
+
+	rv = ACPI_ID_PROBE(device_get_parent(dev), dev,
+	    __DECONST(char **, plat->hids), NULL);
```
VexedUXR: Can `plat->hids` just not be marked as `const` to begin with? If it's later determined that `ACPI_ID_PROBE` doesn't need a writable array (I'm not sure why it would), then there will likely be a sweeping commit for all drivers in the tree that use it. There are _a lot_ of drivers in the tree that don't mark their HID arrays as `const` because of this.

sys/dev/gpio/intel/adlngpio.c
```diff
@@ -0,0 +1,104 @@
...
+
+#include "opt_acpi.h"
+#include "gpio_if.h"
+
+static const struct intelgpio_padgroup adln_com0_groups[] = {
+	{ .first_pad =  0, .npads = 26, .gpio_base =  0, .name = "GPP_B" },
+	{ .first_pad = 26, .npads = 16, .gpio_base = 32, .name = "GPP_T" },
+	{ .first_pad = 42, .npads = 25, .gpio_base = 64, .name = "GPP_A" },
+};
+static const struct intelgpio_padgroup adln_com1_groups[] = {
```
VexedUXR, line 30: style nit
```suggestion
};

static const struct intelgpio_padgroup adln_com1_groups[] = {
```

sys/dev/gpio/intel/tglhgpio.c
```diff
@@ -0,0 +1,121 @@
...
+	{ .first_pad = 282, .npads = 9, .gpio_base = INTELGPIO_GPIO_NOMAP,
+	    .name = "JTAG" },
+};
+
+static const struct intelgpio_community tglh_communities[] = {
+	{ .ngroups = 4, .groups = tglh_com0_groups },
+	{ .ngroups = 5, .groups = tglh_com1_groups },
+	{ .ngroups = 2, .groups = tglh_com3_groups },
+	{ .ngroups = 3, .groups = tglh_com4_groups },
+	{ .ngroups = 2, .groups = tglh_com5_groups },
```
VexedUXR: I'll put this here once instead of cluttering the review with all of them.
Same goes for `adlngpio`. :)

```suggestion
	{ .ngroups = nitems(tglh_com5_groups), .groups = tglh_com5_groups },
```

VexedUXR (changes requested):
I did a quick pass and commented on some things that stood out to me. I'll take a more in-depth look once I have the time.

Thanks.

VexedUXR:
Also CC @strejda can't seem to add you to the review.

sys/dev/gpio/intel/intelgpio.h
```diff
@@ -0,0 +1,85 @@
...
+
+	struct resource *sc_mem_res[INTELGPIO_MAX_COMMUNITIES];
+	uint32_t sc_padbar[INTELGPIO_MAX_COMMUNITIES];
+};
+
+/*
+ * The intelgpio framework provides a common implementation for Intel GPIO
+ * controllers. Most of the gpio(4) interface methods (pin_max, pin_get,
+ * pin_set, etc.) as well as device_detach are already implemented in
+ * intelgpio.c and included via the INTELGPIO_DEVMETHODS macro.
```
VexedUXR: This comment needs to be updated.

sys/dev/gpio/intel/intelgpio.c
```diff
@@ -0,0 +1,417 @@
...
+		return (EINVAL);
+	if (res == NULL)
+		return (EINVAL);
+
+	INTELGPIO_LOCK(sc);
+	val = bus_read_4(res, pad_off);
+	INTELGPIO_UNLOCK(sc);
+
+	*value = (val & INTELGPIO_PADCFG0_GPIORXSTATE) ? GPIO_PIN_HIGH :
+							 GPIO_PIN_LOW;
```
VexedUXR:
```suggestion
	*value = (val & INTELGPIO_PADCFG0_GPIORXSTATE) ?
	    GPIO_PIN_HIGH : GPIO_PIN_LOW;
```

sys/dev/gpio/intel/adlngpio.c
```diff
@@ -0,0 +1,90 @@
...
+	DEVMETHOD_END
+};
+
+DEFINE_CLASS_1(intelgpio, adln_driver, adln_methods,
+    sizeof(struct intelgpio_softc), intelgpio_driver);
+
+DRIVER_MODULE(adlngpio, acpi, adln_driver, NULL, NULL);
+MODULE_DEPEND(adlngpio, acpi, 1, 1, 1);
+MODULE_DEPEND(adlngpio, gpiobus, 1, 1, 1);
+MODULE_VERSION(adlngpio, 1);
```
VexedUXR, line 93: You can add the HID info here using `ACPI_PNP_INFO` so the module is auto loaded.
Same for tglhgpio.

sys/dev/gpio/intel/intelgpio.c
```diff
@@ -0,0 +1,417 @@
...
+int
+intelgpio_attach(device_t dev, const struct intelgpio_platform *plat)
+{
+	struct intelgpio_softc *sc;
+	bool have_res;
+	int i, rid;
+
+	sc = device_get_softc(dev);
+	sc->sc_dev = dev;
+	sc->sc_handle = acpi_get_handle(dev);
```
VexedUXR: Why is this needed?

sys/dev/gpio/intel/intelgpio.c
```diff
@@ -0,0 +1,417 @@
...
+	    ("too many communities: %d > %d", plat->ncommunities,
+		INTELGPIO_MAX_COMMUNITIES));
+
+	INTELGPIO_LOCK_INIT(sc);
+
+	have_res = false;
+	for (i = 0; i < plat->ncommunities; i++) {
+		rid = i;
+		sc->sc_mem_res[i] = bus_alloc_resource_any(dev, SYS_RES_MEMORY,
+		    &rid, RF_ACTIVE);
```
VexedUXR: You can just pass `i` by value here.

sys/dev/gpio/intel/intelgpio.c
```diff
@@ -0,0 +1,417 @@
...
+		return (EINVAL);
+	if ((flags & GPIO_PIN_INPUT) && (flags & GPIO_PIN_OUTPUT))
+		return (EINVAL);
+
+	INTELGPIO_LOCK(sc);
+	val = bus_read_4(res, pad_off);
+
+	/*
+	 * Switch the pad to GPIO mode. Intel pads can operate in different
+	 * modes (native function 1-15 or GPIO). Since setflags is only called
```
VexedUXR, line 214: AFAICT this field is three bits (bits 12:10 of PMODE) for tiger lake and alder lake at least.
Is it different for other chipsets?

VexedUXR (commented):
Just a few small things, otherwise looks good to me.

sys/dev/gpio/intel/intelgpio.h
ckoehne: Looks unneeded now. Removed it. Thanks.

sys/dev/gpio/intel/intelgpio.c
ckoehne: Probably an artifact of an old version. Thanks for catching.

sys/dev/gpio/intel/intelgpio.c
ckoehne: I've checked the datasheet and you're right. For some reason, Linux uses bits 13:10 [1]. Nevertheless, I'm going to use 12:10.

[1] https://elixir.bootlin.com/linux/v7.1.2/source/drivers/pinctrl/intel/pinctrl-intel.c#L70

VexedUXR (approved):
Thanks.

PR #2213, opened 2026-05-19
 lib/libc/gen/fts.3: use 'options' consistently in fts_set() return description

Jeetubhati:
The RETURN VALUES section used "instr" to describe the fts_set()
argument, while the SYNOPSIS and all other references use "options".
Fix the inconsistency.

Reported during GSoC 2026 code review by Alan Somers.
Reviewed by: asomers

### DIFF
--- a/lib/libc/gen/fts.3
+++ b/lib/libc/gen/fts.3
@@ -25,7 +25,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd October 6, 2025
+.Dd May 21, 2026
 .Dt FTS 3
 .Os
 .Sh NAME
@@ -893,7 +893,7 @@ to an appropriate non-zero value.
 The
 .Fn fts_set
 function returns 0 on success and \-1 if its
-.Fa instr
+.Fa options
 argument is invalid.
 .Pp
 The


lib/libc/tests/gen/Makefile
```diff
@@ -7,6 +7,9 @@ ATF_TESTS_C+=		fmtcheck2_test
 ATF_TESTS_C+=		fmtmsg_test
 ATF_TESTS_C+=		fnmatch2_test
 ATF_TESTS_C+=		fpclassify2_test
+ATF_TESTS_C+=		fts_children_test
+ATF_TESTS_C+=		fts_open_test
+ATF_TESTS_C+=		fts_set_test
```
jillest: Source files are added for fts_children_test and fts_open_test but not fts_set_test. Is fts_set_test.c missing?

Also, the new test cases are not placed in alphabetical order like the existing ones.

lib/libc/tests/gen/fts_children_test.c
```diff
@@ -0,0 +1,323 @@
+/*-
```
concussious: Hi! You don't need to use this hyphen anymore, it was for a long-abandoned parser.

```suggestion
/*
```

asomers:
Don't forget to bump the .Dd date too.

asomers:
Also, you should add a `Sponsored by:   Google LLC (GSoC 2026)` tag to your commit messages.

lib/libc/tests/gen/Makefile
asomers: Those shouldn't even be part of this PR.  They got accidentally merged into this branch.

Jeetubhati: > Source files are added for fts_children_test and fts_open_test but not fts_set_test. Is fts_set_test.c missing?
> 
> Also, the new test cases are not placed in alphabetical order like the existing ones.

You are right on both points. The Makefile entries got accidentally 
included from another branch. fts_set_test.c belongs in a separate 
PR. The alphabetical ordering will be fixed.

Jeetubhati: > Those shouldn't even be part of this PR. They got accidentally merged into this branch.

Understood. Those entries will be removed from this PR — they belong 
in their respective PRs.

Jeetubhati:
> Also, you should add a `Sponsored by: Google LLC (GSoC 2026)` tag to your commit messages.

Will add "Sponsored by: Google LLC (GSoC 2026)" to all commit 
messages going forward and amend the existing ones.

Jeetubhati:
> Don't forget to bump the .Dd date too.

Ok, I'll do that

lib/libc/tests/gen/fts_children_test.c
Jeetubhati: Thanks for the heads up. I will remove the hyphen from the opening 
comment in all new files.

Jeetubhati: Fixed. Removed the hyphen from all new test files.

asomers (approved):

asomers:
Oh, one more thing.  Commit a802334d86c7173ddd7776a20065bd5682f25f67, which introduced "instr" into this man page, has already been merged to stable/14 and stable/15 . So you we should merge this one, too.  The way to do that is to add a "MFC after" tag to the commit message with some suggested time interval, like one week.  Then there's an automated script that will send an email reminder after that amount of time.  There's also http://mfc.freebsd.org which helps to track merges.  Generally, you should MFC any change that doesn't require pulling new stuff into contrib, doesn't cause any kind of backwards-compatibility concerns, and doesn't introduce major new features.

Jeetubhati:
> Oh, one more thing. Commit [a802334](https://github.com/freebsd/freebsd-src/commit/a802334d86c7173ddd7776a20065bd5682f25f67), which introduced "instr" into this man page, has already been merged to stable/14 and stable/15 . So you we should merge this one, too. The way to do that is to add a "MFC after" tag to the commit message with some suggested time interval, like one week. Then there's an automated script that will send an email reminder after that amount of time. There's also http://mfc.freebsd.org which helps to track merges. Generally, you should MFC any change that doesn't require pulling new stuff into contrib, doesn't cause any kind of backwards-compatibility concerns, and doesn't introduce major new features.

Done. Added "MFC after: 1 week" to the commit message and force
pushed. Good to know about the MFC process — I will add the tag
to any future fixes that apply to stable branches.

PR #2214, opened 2026-05-19
ipfilter: Fix NULL pointer dereferences when processing ICMPv6 error packets

TeddyEngel:
Fix kernel panic due to unchecked `M_LEN(fin->fin_m)` on synthetic ICMPv6 inner-packet parse with `fin_m == NULL`, plus dead length check replaced with `fin_dlen` validation.

Malformed/truncated ICMPv6 error packets are now rejected earlier. Valid packets meeting RFC-minimum encapsulation are unaffected.

@cschuber - this has been assigned to you for a while, so I took a shot at fixing it. Happy to defer if you'd prefer to handle it

### Root Cause

When `ipf_checkicmp6matchingstate()` processes an ICMPv6 error packet, it creates a synthesized `fr_info_t` structure (`ofin`) with `fin_m` explicitly set to NULL (with the comment "if dereferenced, panic XXX"). It then calls `ipf_makefrip()` which eventually reaches `ipf_pr_icmp6()`. This function calls `M_LEN(fin->fin_m)` without checking if `fin_m` is NULL, causing a kernel panic.

Additionally, the original length check (`fin->fin_plen < sizeof(ip6_t)`) was unreachable after the earlier `ICMP6ERR_MINPKTLEN` gate and therefore non-functional.

### Fixes

1. **fil.c: ipf_pr_icmp6() / ipf_pr_icmp()** - Add NULL check for `fin->fin_dp` before dereferencing
2. **fil.c: ipf_pr_icmp6()** - Add NULL check for `fin->fin_m` before calling `M_LEN()`
3. **ip_state.c: ipf_checkicmp6matchingstate()** - Add NULL checks for `ic6` and `oic` pointers
4. **ip_state.c: ipf_checkicmp6matchingstate()** - Fix length validation (original check was dead code)

### Testing

- [Reproducer](https://bugs.freebsd.org/bugzilla/attachment.cgi?id=262288) from PR 288333 no longer causes kernel panic
- Module builds and loads cleanly
- No regressions observed

### Problem Report

[PR 288333](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=288333)

### DIFF
--- a/sys/netpfil/ipfilter/netinet/fil.c
+++ b/sys/netpfil/ipfilter/netinet/fil.c
@@ -890,6 +890,8 @@ ipf_pr_icmp6(fr_info_t *fin)
 		ip6_t *ip6;
 
 		icmp6 = fin->fin_dp;
+		if (icmp6 == NULL)
+			return;
 
 		fin->fin_data[0] = *(u_short *)icmp6;
 
@@ -1198,6 +1200,8 @@ ipf_pr_icmp(fr_info_t *fin)
 	}
 
 	icmp = fin->fin_dp;
+	if (icmp == NULL)
+		return;
 
 	fin->fin_data[0] = *(u_short *)icmp;
 	fin->fin_data[1] = icmp->icmp_id;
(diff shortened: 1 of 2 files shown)


cschuber:
Thanks for the nudge. I'll review this.

cschuber:
Thanks for the nudge. It's pushed to freebsd/src.

cschuber:
Thank you for the high quality commits.

TeddyEngel:
Thanks @cschuber!

PR #2217, opened 2026-05-20
lib/libc/tests/gen: add fts_open() error and edge case tests

Jeetubhati:
Add ATF test cases covering fts_open() error conditions and edge cases:

- invalid option bits (outside FTS_OPTIONMASK) yield EINVAL
- empty argv yields EINVAL
- empty path string yields FTS_NS with ENOENT
- nonexistent path yields FTS_NS, not open failure
- trailing slash does not crash (SVN r49851 regression)
- unreadable directory yields FTS_D then FTS_DNR, never FTS_DP
- multiple root paths are all visited left to right

Reviewed by: asomers
Sponsored by: Google Summer of Code 2026

### DIFF
--- a/lib/libc/tests/gen/Makefile
+++ b/lib/libc/tests/gen/Makefile
@@ -11,6 +11,7 @@ ATF_TESTS_C+=		fpclassify2_test
 ATF_TESTS_C+=		fts_blocks_test
 .endif
 ATF_TESTS_C+=		fts_misc_test
+ATF_TESTS_C+=		fts_open_test
 ATF_TESTS_C+=		fts_options_test
 ATF_TESTS_C+=		ftw_test
 ATF_TESTS_C+=		getentropy_test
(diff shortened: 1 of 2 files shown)


lib/libc/tests/gen/Makefile
```diff
@@ -7,6 +7,9 @@ ATF_TESTS_C+=		fmtcheck2_test
 ATF_TESTS_C+=		fmtmsg_test
 ATF_TESTS_C+=		fnmatch2_test
 ATF_TESTS_C+=		fpclassify2_test
+ATF_TESTS_C+=		fts_children_test
```
asomers: This Makefile includes entries for fts_children_test and fts_set_test, even though those files aren't part of the review.  You should remove them.  Also, you should sort these tests alphabetically, as somebody pointed out on the earlier PR.

asomers (changes requested):

lib/libc/tests/gen/Makefile
Jeetubhati: Fixed in the latest commit. Removed the extra entries and 
confirmed fts_open_test is in the correct alphabetical position.

asomers (approved):

PR #2218, opened 2026-05-20
lib/libc/tests/gen: add fts_children() tests

Jeetubhati:
Add ATF test cases covering fts_children() behaviour:

- before fts_read returns root entry list
- empty directory returns NULL with errno 0
- non-empty directory returns all children in order
- called twice returns equivalent results
- FTS_NAMEONLY fills only fts_name, fts_info is FTS_NSOK
- non-directory node returns NULL with errno 0
- invalid options returns NULL with EINVAL

Reviewed by: asomers
Sponsored by: Google Summer of Code 2026

### DIFF
--- a/lib/libc/tests/gen/Makefile
+++ b/lib/libc/tests/gen/Makefile
@@ -10,6 +10,7 @@ ATF_TESTS_C+=		fpclassify2_test
 .if ${COMPILER_FEATURES:Mblocks}
 ATF_TESTS_C+=		fts_blocks_test
 .endif
+ATF_TESTS_C+=		fts_children_test
 ATF_TESTS_C+=		fts_misc_test
 ATF_TESTS_C+=		fts_open_test
 ATF_TESTS_C+=		fts_options_test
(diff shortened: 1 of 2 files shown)


lib/libc/tests/gen/Makefile
```diff
@@ -7,6 +7,9 @@ ATF_TESTS_C+=		fmtcheck2_test
 ATF_TESTS_C+=		fmtmsg_test
 ATF_TESTS_C+=		fnmatch2_test
 ATF_TESTS_C+=		fpclassify2_test
+ATF_TESTS_C+=		fts_children_test
+ATF_TESTS_C+=		fts_open_test
+ATF_TESTS_C+=		fts_set_test
```
asomers: You should remove the lines for files that aren't part of this review.

asomers (changes requested):

lib/libc/tests/gen/Makefile
Jeetubhati: Fixed in amended commit. Removed fts_open_test and fts_set_test entries.
This PR now only adds fts_children_test

asomers (approved):

PR #2221, opened 2026-05-22
Introduce initial Rust support for kernel builds

xaoseric:
This commit integrates the Rust compiler into the kernel and kernel module build systems. It adds a new `RUST_KERNEL` option, configured via `mk.conf`, which enables compilation of `.rs` files into kernel objects.

A simple `rust_echo` character device driver, composed of both C and Rust code, is included to demonstrate the new build infrastructure and C/Rust FFI patterns. The `config` utility is updated to generate appropriate Makefile rules for Rust source files.


bms:
Hard no from me, sorry.  I speak for myself and not the rest of the Project. You are of course free to experiment with kernel space Rust in your own branches. But please don't advocate for it in mainline FreeBSD.

Historically, there have been other efforts to implement e.g. C++ runtimes for kernel space in FreeBSD, and this was pushed back hard on for similar reasons, although I was amenable to it at the time.

These days, C23 (and tools such as CProver) are starting to offer most of the benefits desired from the kernel space use of e.g. Rust or C++ in that domain anyway, and supporting it seems to be a more viable strategy, as well as a more effective use of limited and valuable developer resource.

There are concrete concerns both about the stability of the Rust language and its ABI, as well as ongoing support for it use in the kernel build environment. This includes Rust's discouragement of and weak support for shared objects, and its concomitant preference for static linking. The claim that its syntax is similar to C/C++ is also demonstrably specious.

We understand CFFI issues are being mitigated however FreeBSD developers have other concerns in this area; these carry a significant maintenance cost. One developer made an excellent point which I paraphrase here: The best reason to not mix Rust into a C codebase is that the linkage between translation units mixed between C and Rust spoils checks both ways, and so is less safe than the original C.

It is a stronger argument against mixing where a full rewrite is impossible, although there is collateral damage: it can lend support to the "Rewrite in Rust in toto" position. That is clearly not going to happen. FreeBSD has always been implemented in C and that is extremely unlikely to change in the foreseeable future.

That history must be pointed out here is ironic, given the Rust community's apparent propensity to fail to learn from or otherwise deny computer science and political history, and its Foundation's questionable relationship with its own user base (e.g. the rather bizarre treatment of JeanHeyd Meneide at RustConf 2023).

The yt-dlp maintainer's experience with bun and subsequent Rust contributions amount to introducing Software of Unknown Provenance (SOUP), to use the IEC medical product definition, in direct contravention of established cybersecurity best practices.

So I do not see how adding Rust support for kernel modules adds any particular value for FreeBSD as a whole.

Thank you for your interest in FreeBSD.

asomers:
There are other people working on this problem, like Aaron Espinoza and Ayrton Muñoz .  Did you consult with them?  This is an important new feature, and it needs an architecture discussion before you can make a PR.  I suggest you email an RFC to freebsd-hackers@FreeBSD.org .  Make sure to CC a.munoz3327@gmail.com and acesp25@freebsd.org , as they might not be on that list.

bms:
Individuals advocating for Rust support in the Linux kernel seem to have been relying on the logical fallacy known as "appeal to novelty" and arguments from false authority (often expressed as "you shouldn't want that"). I have to stand firm in my objections to Rust being included in FreeBSD. The intention here is not to engage in anti-Rust polemic of any kind.

The technical objections remain. The main gain of Rust is often stated as "it is memory-safe", but it's difficult to the trained C-wielding ear not to hear this in the same tones of voice as "Brawndo's got Electrolytes" from Mike Judge's seminal satirical work "Idiocracy"; similar gains can be had by using e.g. CProver to begin moving towards provably-correct systems C use.

When implementing operating systems, there is a clear need to cleanly and consistently implement trees, sets, lists, skiplists etc. If Rust is not suited to this task, and relies solely on its collections library (with its concomitant maintenance, compile and runtime costs) then this undermines the case for its uptake in e.g. FreeBSD; an operating system which is already implemented in tight and well-reviewed C99/C11/C23.

Moreover, when posed with quite legitimate systems programming questions about the difficulties inherent in implementing such data structures in Rust itself, Rust advocates may falsely assert that these difficulties do not exist. This is in addition to the largely baseless claim that its syntax is similar to C/C++; empirical observation reveals only superficial similarities in appearance, nothing more. Its lexical structure and operator notation are quite different; the fact that structure declarations in Rust may differ depending on where the data is stored (stack vs heap) is another example.

At the moment, I lack the **tacit** knowledge to more fully refute a contention that Rust should be supported in FreeBSD, on a wholly empirical basis, so I rely on anecdotes and empirical study by other developers; e.g. the position that supporting Rust's runtime is an all-or-nothing proposition, e.g. the example of std::collections for LinkedList given by Felipec, and the performance constraints that its use might impose.

Christoph Hellwig already advanced grounded objections to Rust in the Linux kernel based on its maintenance overhead. Hector Martin's comments, by contrast, come across as a form of black propaganda; toxic, baseless, and has no place being anywhere near FreeBSD, let alone Linux.

If people really want to use Rust, they can do it in their own branches, but please don't ask for it to go into mainline FreeBSD; it doesn't belong there.

I'm not going to indulge conspiracy theory about license washing, "big tech" influence, or other such, but the fact that I am hearing any such things at all from people makes me cringe to even begin with. I have absolutely nothing against Rust's userland use in Firefox and I even encourage and applaud that.

I do not wish to learn Rust either, as it is of no personal interest to me whatsoever. I reason that I would not benefit from this other than pursuing language diversity; but this is irrelevant to anything I am currently doing or plan to do. I do not seek to control what others do, however.

As with so-called "AI", I am irritated that I must expend cycles refuting it.

bsdimp:
First off, Bruce doesn't speak for the project. Neither do I. Individuals don't have veto power.

Rust in the kernel is fine. We should allow people to experiment with it so we can see if it's a good idea or not. Right now, a decent case for it 'might be' useful with lots of unknowns. We should allow people wishing to experiment in this area to collaborate in our tree by having that in the base once it gets to a good enough point that wider testing can give new data. The people actually working in this area get to make the call.

xaoseric:
> First off, Bruce doesn't speak for the project. Neither do I. Individuals don't have veto power.
> 
> 
> 
> Rust in the kernel is fine. We should allow people to experiment with it so we can see if it's a good idea or not. Right now, a decent case for it 'might be' useful with lots of unknowns. We should allow people wishing to experiment in this area to collaborate in our tree by having that in the base once it gets to a good enough point that wider testing can give new data. The people actually working in this area get to make the call.

Yes, if it is done in a way that allows it to be built optionally, I think that is a very good compromise, and this PR has a flag that you need to pass if you want to build with rust support.

PR #2223, opened 2026-05-22
libc: Suppress <stdalign.h> content for C23 and later

kfv:
C23 deprecates `<stdalign.h>` and specifies that the header shall provide no content (§7.15.1).

### DIFF
--- a/include/stdalign.h
+++ b/include/stdalign.h
@@ -26,6 +26,8 @@
  * SUCH DAMAGE.
  */
 
+#if __STDC_VERSION__ < 202311L
+
 #ifndef __alignas_is_defined
 #define	__alignas_is_defined	1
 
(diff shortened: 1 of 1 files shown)


kfv:
Cc: @clausecker

bsdimp (approved):

bsdimp:
I've not verified the standard.
And I know know what this might break,

kfv:
Thanks for approving the PR and doing a review. I appreciate it, Warner.

Standard simply says in §7.15.1 that

> The header `<stdalign.h>` provides no content.

This is because both `alignas` and `alignof` are now keywords since C23, with obsolescent alternative spellings `_Alignas` and `_Alignof`, respectively (§6.4.1.3).

clausecker:

PR #2226, opened 2026-05-23
bhyveload: add a small ring buffer to the cb_putc function.

yuichiro-naito:
While bhyveload opens an nmdm device as a console, I often see that the boot loader messages are partly missing on my console. I investigated the bhyloveload with truss(1), I found that write(2) system call returns EGAIN while missing the message. The console output descriptor is opened with O_NONBLOCK. So, the write(2) system call may return the EAGAIN. It is described in the manual. An application should accept and handle the error case.

I added a small ring buffer for the output characters and retry writing in the error case. I see about 150 characters are missing with my debug message, so the 256-byte buffer is enough for me.

### DIFF
--- a/usr.sbin/bhyveload/bhyveload.c
+++ b/usr.sbin/bhyveload/bhyveload.c
@@ -57,6 +57,7 @@
 #include <sys/stat.h>
 #include <sys/disk.h>
 #include <sys/queue.h>
+#include <sys/uio.h>
 
 #include <machine/specialreg.h>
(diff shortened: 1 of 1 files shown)


yuichiro-naito:
I found that the `uint niov` variable doesn't need to be 'static'. I removed 'static' from the variable declaration.

yuichiro-naito:
I have a better idea for this issue. I'll make a new pull request instead of this pull request.
I'll close this.

PR #2229, opened 2026-05-23
ipfilter: Add NULL check for fin_dp in TCP, UDP, GRE, and AH packet h…

devnexen:
…andlers

Mirror commit 68ed81631afa, which added the same check in ipf_pr_icmp() and ipf_pr_icmp6().  ipf_pr_pullup() can return success while leaving fin->fin_dp NULL (e.g. when fin_m is NULL on a synthesized fr_info_t, as constructed by ipf_checkicmp6matchingstate()), so callers that immediately dereference fin_dp must check for NULL too.

The same pattern was missed in ipf_pr_tcpcommon(), ipf_pr_udpcommon(), ipf_pr_gre(), ipf_pr_gre6(), and ipf_pr_ah(), each of which would panic on a NULL dereference under the same conditions.

### DIFF
--- a/sys/netpfil/ipfilter/netinet/fil.c
+++ b/sys/netpfil/ipfilter/netinet/fil.c
@@ -1082,6 +1082,8 @@ ipf_pr_gre6(fr_info_t *fin)
 	}
 
 	gre = fin->fin_dp;
+	if (gre == NULL)
+		return;
 	if (GRE_REV(gre->gr_flags) == 1)
 		fin->fin_data[0] = gre->gr_call;
 }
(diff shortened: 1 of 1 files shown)


cschuber:
Thanks for the patch. (On phone app.) I'll review this when I get back on Monday.

cschuber:
I don't see how fin_dp could ever be NULL. It is always *ip + hlen. fin_dp always points to the data header following the IP header. Do you have a panic dump I can look at?

devnexen:
You're right ; ipf_makefrip() sets fin_dp = ip + hlen unconditionally,
  and the only later mutators add a non-negative shift, so it can't reach
  these handlers as NULL. Patterned this on the ICMP commit without a
  real trigger. Closing.

PR #2230, opened 2026-05-23
cut(1), rev(1): add Capsicum sandboxing

DominoTree:
Open file arguments through Casper's cap_fileargs service and enter capability mode after limiting stdio rights, so cut(1) and rev(1) run sandboxed.

cut(1) is a bootstrap tool, so the Makefile drops the Casper libraries and -DWITH_CASPER while BOOTSTRAPPING; <casper/cap_fileargs.h> then resolves fileargs_*() to plain fopen(3), matching wc(1)/head(1).

### DIFF
--- a/usr.bin/cut/Makefile
+++ b/usr.bin/cut/Makefile
@@ -2,6 +2,12 @@
 
 PROG=	cut
 
+.if ${MK_CASPER} != "no" && !defined(BOOTSTRAPPING)
+LIBADD+=	casper
+LIBADD+=	cap_fileargs
+CFLAGS+=	-DWITH_CASPER
+.endif
+
 HAS_TESTS=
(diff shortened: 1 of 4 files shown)


shkhln:
What is your use case?

DominoTree:
> What is your use case?

I think these are two of the last things in base that deal with a lot of untrusted user input by definition, and I don't see a lot of reason _not_ to plumb Capsicum - it's certainly not a lot of computational overhead or anything - and I think it's better for correctness sake

shkhln:
It's not exactly free — there are more CVEs against libcasper/libnv out there than against `rev`.

DominoTree:
Whoops I commented right as you did, that's a very valid point

PR #2211, opened 2026-05-18
usbc: USB-PD R3.2 policy SM + TCPC abstraction

b1nc0d3x:
## Summary

Add a new in-tree framework for USB Power Delivery state-machine logic that is independent of any specific Type-C port controller chip. The framework lives at `sys/dev/iicbus/usb/usbc/`, builds as a kernel module (`usbc.ko`), and exposes four layers:

1. **`usbc_pd.h`** — base types, enums, and timing constants from USB PD R3.2: SOP* packet types, control/data message numbers, header bit layout, PDO/RDO field encoders, protocol timers (`tSenderResponse`, `tPSTransition`, ...), VDM SVID / object position constants for DisplayPort altmode.

2. **`usbc_pd_msg.{h,c}`** — hardware-independent build/parse for PD message headers and data objects. No I/O, no locking; callable from interrupt context:
   - `usbc_pd_build_header()` — pack header fields to a u16
   - `usbc_pd_parse_header()` — inverse, with bounds checking
   - `usbc_pd_build_source_cap_pdo_fixed()` — 5V fixed-supply PDO
   - `usbc_pd_build_request_rdo()` — fixed-supply request
   - `usbc_pd_validate_pdo()` — sanity-check before TX

3. **`usbc_pd_policy.{h,c}`** — the protocol/policy engine. Implements the source-port policy state machine from PD chapter 8 (Power Negotiation), the soft-reset and hard-reset paths, the VDM exchange for Discover Identity / Discover SVIDs / Enter Mode / DP_Status, and the timer-driven retry/abort logic. Caller-facing API:
   - `usbc_pd_policy_attach(sc, tcpc_ops)`
   - `usbc_pd_policy_detach(sc)`
   - `usbc_pd_policy_rx(sc, msg)` — chip-driver IRQ handler entry
   - `usbc_pd_policy_role_change(sc, ...)`
   - `usbc_pd_policy_enter_dp_altmode(sc, port_pin_assign)`

4. **`usbc_tcpc.h`** — the abstraction the chip driver implements. A struct of function pointers (transmit a PD msg, set CC state, set vbus, set role, etc.) that the policy SM invokes. This is what makes the framework hardware-independent — the FUSB302B driver in the companion (forthcoming) review supplies one implementation, and a future TUSB320 / TPS6598x / MT6360 driver supplies another without changing a line of policy code.

## Why land this separately

- Dependency-free against the rest of the tree — no other in-tree driver consumes it yet.
[...]

PR #2231, opened 2026-05-24
fusb302: Fairchild FUSB302B USB-C / USB-PD controller driver

b1nc0d3x:
## Summary

Add an in-tree driver for the Fairchild (now ON Semiconductor)
FUSB302B USB Type-C port controller.  The chip is an i2c slave
that implements CC1/CC2 sensing, BMC PHY for USB Power Delivery
messages, DRP toggle for role detection, and a VBUS measure ADC.
It is the Type-C controller used on the RockPro64, Pinebook Pro,
ROCK Pi 4, Khadas Edge, and several Chromebook designs around the
RK3399 era.

The driver supplies the `struct usbc_tcpc_ops` vtable consumed by
the `usbc(4)` policy framework (D3, #2211).  All PD message
framing, policy state, VDM negotiation, and DisplayPort altmode
logic lives in usbc; this driver is the chip-specific I2C /
interrupt / FIFO half.

Functional coverage:

  * SRC + SNK port roles, with DRP toggle for autonomous role
    discovery on cold plug.
  * USB PD R3.2 messaging via the BMC TX/RX FIFOs (SOP only;
    SOP'/SOP'' cable messages scaffolded but not exercised).
  * Spec rev autoselect with PD 2.0 / PD 3.0 fallback (sysctl-
    selectable for compatibility testing against picky sinks).
  * Soft-reset and hard-reset paths.
  * Polled DRP retoggle on lost partner -- recovers from sinks
    that momentarily drop CC during their own re-orientation.
  * IRQ-driven; uses `sx(9)` for the soft state and never holds
    a spinlock across an I2C transaction (FUSB302's transfers can
    sleep on bus arbitration).
  * Debug output gated behind `dev.fusb302.<N>.debug`, default 0.

Modeled functionally on the Linux
`drivers/usb/typec/tcpm/fusb302.c` driver but is a clean
reimplementation -- no code copied from Linux.  Linux is the
reference for register sequencing only; the state machine,
locking discipline, and tunables are FreeBSD-native.

Includes a Rockchip-specific helper at
`sys/arm64/rockchip/rk3399_fusb302_helper.c` (built as a
separate module) that bridges FUSB302's role/orientation
detection to the RK3399 Type-C PHY mode-switch.  This file is the
only piece in the review that is Rockchip-specific; the
fusb302(4) driver itself runs on any platform that exposes an
i2c bus and a level-sensitive GPIO for INT_N.

## Dependencies

  Depends on D3 (`usbc` framework, #2211) being committed first.
  The vtable consumed (`struct usbc_tcpc_ops`, `usbc_pd_policy_*`
  entry points) is defined there.  Build will fail without D3
  applied.

## What's deliberately NOT in this review

  * The rk_typec_phy driver itself -- D5 (#2225).
  * The Cadence MHDP DisplayPort framer -- D6 (forthcoming).
  * Userspace control hooks (tcpmctl, sysctl-driven role
    override beyond debug knobs) -- reserved for a follow-up.
  * Programmable-power-supply or fast-role-swap support.

## Test plan

  * Pine64 RockPro64 v2.1 + XYM W156F1 USB-C portable monitor:
    cold boot completes a full PD source contract (SRC_CAP ->
    REQUEST -> ACCEPT -> PS_RDY) and a DP altmode entry within
    ~1.5 s of probe attach.  Verified across 20+ plug/unplug
    cycles; no leaked state, no IRQ storm (cf. D1, #2197, the
    rk_gpio PIC fix that makes this even reachable on cold boot).
  * Same board, USB-A type-C cable (no DP altmode advertised):
    PD source contract still completes; altmode entry is
    correctly declined; downstream rk_typec_phy stays in USB
    mode.
  * Spec-rev forced via `dev.fusb302.0.pd_spec_rev=20` (PD 2.0
    only) on three different PD sinks (XYM W156F1, Anker 535
    dock, a counterfeit Apple-pattern cable): all complete
    contract without falling back further.  `pd_spec_rev=30`
    (PD 3.0) also completes on the W156F1 and the Anker but is
    rejected by the counterfeit cable, which is expected.
  * Hard-reset recovery: forced via sysctl
    `dev.fusb302.0.force_hard_reset=1` mid-contract.  Driver
    returns to USBC_PD_SRC_STARTUP and renegotiates within
    300 ms.
  * `kldload fusb302; kldload rk3399_fusb302_helper; kldunload
    rk3399_fusb302_helper; kldunload fusb302` -- no leaks, no
    interrupts left enabled at the controller, no spurious GPIO
    edges observed under dtrace.

## Size note

At ~4.5k lines this is on the larger side for a single
Differential.  If reviewers prefer it split, the natural cut is:

  * D4a: `fusb302.c` + `fusb302_var.h` +
    `sys/modules/fusb302/Makefile` (the platform-independent
    driver, ~4.4k lines).
  * D4b: `rk3399_fusb302_helper.c` + its Makefile (Rockchip-
    specific extcon/typec-phy glue, ~140 lines).

I left it as one review because D4b is functionally meaningless
without D4a, and reviewing the helper in isolation is harder
than reviewing it as a small section of the same patch.  Happy
to split on request.

## Suggested reviewers

  bsdimp, emaste, wulf7 (recent USB / iicbus committers per
  `git log`); evadot, mhorne (sys/arm64/rockchip).

### DIFF
--- a/sys/arm64/rockchip/rk3399_typec_altmode_var.h
+++ b/sys/arm64/rockchip/rk3399_typec_altmode_var.h
@@ -1,0 +1,22 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#ifndef _ARM64_ROCKCHIP_RK3399_TYPEC_ALTMODE_VAR_H_
+#define _ARM64_ROCKCHIP_RK3399_TYPEC_ALTMODE_VAR_H_
+
+#include <sys/bus.h>
+#include <sys/types.h>
+
+struct rk3399_typec_dp_altmode_status {
+	bool		valid;
+	bool		dp_ready;
+	int		usb_ss;		/* -1 unknown, 0 DP-only, 1 USB3+DP */
+	uint32_t	pin_assignment;
+	uint32_t	dp_status;
+};
+
+int	rk3399_typec_dp_altmode_get_state(device_t dev,
+	    struct rk3399_typec_dp_altmode_status *status);
+
+#endif
--- a/sys/dev/iicbus/usb/fusb302.c
+++ b/sys/dev/iicbus/usb/fusb302.c
@@ -1,0 +1,4332 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2026 Kyle Crenshaw
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+/*
+ * FUSB302 USB Type-C controller with USB-PD source policy and DP Alt Mode
+ * VDM state machine.
+ *
+ * Implements the DFP (source) path: CC detection → source caps negotiation →
+ * Discover_Identity → Discover_SVIDs → Discover_Modes → Enter_Mode →
+ * DP_Status → DP_Config.  On success the result is exported via
+ * fusb302_get_dp_altmode_state() for rk_cdn_dp.
+ */
+
+#include <sys/param.h>
+#include <sys/systm.h>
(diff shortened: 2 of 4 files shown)


sys/arm64/rockchip/rk3399_fusb302_helper.c
```diff
@@ -0,0 +1,130 @@
...
+#include <sys/kernel.h>
+#include <sys/module.h>
+#include <sys/systm.h>
+#include <sys/sysctl.h>
+
+#include "rk3399_typec_altmode_var.h"
+#include <dev/iicbus/usb/fusb302_var.h>
+
+static struct sysctl_ctx_list rk3399_fusb302_helper_ctx;
+static struct sysctl_oid *rk3399_fusb302_helper_tree;
```
mhorne: This sysctl tree does not need to be declared dynamically (with `SYSCTL_ADD`). There exists machinery (in `sys/kern/kern_linker.c`) that will register and enable all statically-declared sysctls present in a module at load time. Same goes for unload.

`SYSCTL_ADD()` is used primarily when per-device sysctl context is needed. This does not seem to be the case here.

sys/arm64/rockchip/rk3399_fusb302_helper.c
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	return (0);
+}
+
+static moduledata_t rk3399_fusb302_helper_mod = {
+	"rk3399_fusb302_helper",
+	rk3399_fusb302_helper_modevent,
+	NULL
+};
```
mhorne: I am having trouble understanding the intended purpose of this module. It seems like the goal is to make `fusb302_get_dp_altmode_state()` overridable compared to the version provided by fusb302. But why? There is no actual RK3399 code here.

Which driver(s) will depend on this module?

mhorne (commented):
A couple quick comments as I skim, and get familiar with the structure of these reviews...

sys/arm64/rockchip/rk3399_fusb302_helper.c
b1nc0d3x: Thanks, Mitchell — 

On the helper module: you're right, there's no actual RK3399 code in there and it's not justifying its existence. It's bring-up scaffolding from when fusb302's PD/altmode negotiation wasn't reliable — we needed a sysctl-driven manual-override path to feed canned pin_assignment / dp_status values into rk_cdn_dp while we debugged the chip side. The PD path is healthy now (has been since early May), so the override hatch is no longer load-bearing. I'll drop rk3399_fusb302_helper.{c,ko} entirely in the next push.

On the dynamic sysctl point: also fair — the helper was the only place in this series using SYSCTL_ADD_* for non-per-device state, and it's going away anyway, so that critique resolves with the same delete. For future modules I'll reach for statically-declared SYSCTL_NODE / SYSCTL_INT at file scope and let the linker's link_set machinery register them at load.

 Heads-up that the same architectural smell is what you're flagging on #2236 (the linker_file_lookup_symbol call in rk_cdn_dp). Dropping the helper lets me answer that one cleanly too — rk_cdn_dp calls fusb302_get_dp_altmode_state as a plain extern, with a real MODULE_DEPEND(rk_cdn_dp, fusb302, …)  declaring the dependency you asked about. I'll reply on #2236 in the same sitting; one architectural answer, two PRs.  

I Will push the changes once the rest of the stack is in a comparable state. Thanks again for the eyes on these — the review surface is large and I truly appreciate your time and patience.

PR #2232, opened 2026-05-25
aq(4): Fix OOB write in RSS table packing (no link / panic)

DominoTree:
Fix an out-of-bounds stack write in aq_hw_rss_set().  RSS table entries are 3 bits (8 queues max), but with more than 8 RX rings rss_table[] holds larger values; the 32-bit write then spills one uint16_t past bitary[] and corrupts the stack, so the NIC never links or the kernel panics. Mask each value to 3 bits and pack 16 bits at a time to keep the write in bounds.

Confirmed on AQC107 (firmware 3.1.88) under FreeBSD 15.0-RELEASE: with the OOB write the interface attached but stayed "no carrier"; removed, link negotiates 1000baseT full-duplex and passes traffic.

cc @emaste

### DIFF
--- a/sys/dev/aq/aq_hw.c
+++ b/sys/dev/aq/aq_hw.c
@@ -878,8 +878,14 @@ aq_hw_rss_set(struct aq_hw_s *self,
 	memset(bitary, 0, sizeof(bitary));
 
 	for (i = HW_ATL_RSS_INDIRECTION_TABLE_MAX; i--;) {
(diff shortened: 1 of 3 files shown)


DominoTree:
Cross-posted at https://github.com/freebsd/freebsd-ports/pull/522 to fix things for existing users before aq(4) lands

DominoTree:
(Added a second small commit with another RSS fix)

DominoTree:
Closing PR, doing work here

https://reviews.freebsd.org/D57240

PR #2236, opened 2026-05-26
rk_cdn_dp: Cadence MHDP DPTX driver for RK3399 USB-C DisplayPort

b1nc0d3x:
## Summary

New in-tree driver for the Cadence Multi-Protocol High-speed DisplayPort (MHDP) Transmitter IP block at `0xfec00000` on the RK3399. The MHDP block runs a small embedded Cortex-M0 firmware blob that handles AUX channel transactions, DPCD reads, link training, and the DisplayPort main-link state machine; the host driver loads the firmware, walks a 19-stage bring-up sequence to get the framer ready, and then issues mailbox commands to drive link establishment and modesetting.

Attaches as a `drm_bridge`-style consumer of the `rk_typec_phy(4)` DP altmode added in #2225, and exposes DPCD readback, EDID readback, link training, MSA programming, and DPMS to higher layers.

## Functional scope

- Firmware loader for `rockchip-mhdp.bin` (signed Cadence blob; redistributable with the chip, handled the same way as `iwlwifi` firmware in base).
- 19-stage staged bring-up: reset assert/deassert, clk_enable, firmware load, uCPU release, mailbox handshake, set host caps, set lane count, set link rate, PHY enable handshake with typec_phy, framer enable, HPD wait, DPCD readback, link training (CR + EQ phases per lane), MSA + video timing, video enable, framer-ready confirmation. Each stage is sysctl-walkable via `dev.rk_cdn_dp.<unit>.stage=<n>` for diagnostics.
- DPCD over Cadence mailbox with the 5-byte-short-reply parser the firmware actually returns (upstream Linux mainline `cdn_dp` has a subtle bug here that the BSP variant avoids).
- Link training at RBR (1.62 Gbps) and HBR (2.7 Gbps), 1, 2, and 4 lanes. HBR2 reserved for follow-up.
- Framer SP_HSP / SP_VSP polarity programming per the panel's EDID (e.g. the XYM W156F1 USB-C portable requires NHSYNC + wide hsync, contrary to the CEA-VIC 1080p defaults).
- HPD-driven re-train + modeset on hot plug / unplug.
- Power-control sysctls for runtime DPMS-like sleep / wake.

## Two intentional deviations from mainline Linux

Both called out with source comments:

- **Reset ordering:** resets MUST be asserted BEFORE deasserting in stage 2. The Linux 4.4 Rockchip BSP does this; mainline Linux's `cdns-mhdp` does not, and on RK3399 mainline `cdns-mhdp` wedges the framer on first probe.
- **Framer polarity:** the framer's `SP_HSP` / `SP_VSP` bits are inverted relative to the DRM mode flags on the XYM panel (and any panel that ships a CEA-VIC 1080p EDID). The driver respects the panel's actual EDID timing, not the EDID's advertised polarity flags.

## Authorship + provenance

From-scratch reimplementation. The Cadence MHDP IP has Linux drivers in two flavours (the Rockchip 4.4 BSP `drivers/gpu/drm/rockchip/cdn-dp-*.c` and mainline `drivers/gpu/drm/bridge/cdns-mhdp8546-*`); both were used as register/sequence reference but no code is copied. License is BSD-2-Clause; SPDX identifier headers present on every file.

## Dependencies

Depends on #2225 (`rk_typec_phy` DP altmode) being committed first. Stages 9..12 invoke phynode methods that #2225 introduces.

A stacked follow-up review will add the DP audio path (Cadence audio packetizer + SDP infoframe + `rk_hdmi_codec_stub` wiring into `audio_soc`).

## Test Plan

- RockPro64 v2.1 + XYM W156F1 USB-C portable monitor: cold boot completes all 19 stages within ~2.5 s of fusb302 attach, 1920x1080 @ 60 Hz HBR (2.7 Gbps) on 2 lanes, EDID parsed.
- Hot replug: 10 cycles, no leaks, retrain succeeds every time.
- RBR (1.62 Gbps) forced via `dev.rk_cdn_dp.0.force_link_rate=rbr`: panel still trains and shows video.
- DPCD readback round-trip via `dev.rk_cdn_dp.0.dpcd_read=<addr>` -- short-reply parser decodes 5-byte and 7-byte replies correctly for 100+ random addresses.

## Diff scope

```
sys/conf/files.arm64                          +1
sys/arm64/rockchip/rk_cdn_dp.c             +6461
sys/modules/rockchip/rk_cdn_dp/Makefile      +16
                                           ------
Total                                      +6478
```

All new code; no existing FreeBSD files modified except the one-line `files.arm64` hookup.

## Suggested reviewers

@manu @bsdimp @mhorne61 @mmel-FreeBSD


sys/arm64/rockchip/rk_cdn_dp.c
```diff
@@ -0,0 +1,6461 @@
...
+	 * rk3399_typec_altmode_helper module's hardcoded defaults. Per-file
+	 * "try fusb302 first then helper" doesn't work because whichever
+	 * module is iterated first wins, and the helper is loaded earlier.
+	 */
+	sym = 0;
+	(void)linker_file_foreach(rk_cdn_dp_lookup_altmode_fusb302_cb, &sym);
+	if (sym == 0)
+		(void)linker_file_foreach(
+		    rk_cdn_dp_lookup_altmode_helper_cb, &sym);
+	return ((int (*)(device_t, struct rk3399_typec_dp_altmode_status *))sym);
```
mhorne: Something is off track here. We don't tend to mess with the linker interface in driver code, there is probably a different way of achieving your desired effect.

Does `rk_cdn_dp` have a functional dependency on an `fusb302` device being attached? Please help me to understand the device/driver relationship here.

sys/modules/rockchip/rk_cdn_dp/Makefile
```diff
@@ -0,0 +1,16 @@
+.PATH: ${SRCTOP}/sys/arm64/rockchip
+
+KMOD=	rk_cdn_dp
```
mhorne, line 3: Good, but you will need to update `SUBDIR` in `sys/modules/rockchip/Makefile` as well.

Otherwise, the module will not be built.

sys/arm64/rockchip/rk_cdn_dp.c
b1nc0d3x: Answering your structural question here along with the helper-module question you raised on #2231, since they're really the same architectural call.

The linker_file_lookup_symbol call at line 5067 is the runtime-resolution side of the same hack that the rk3399_fusb302_helper module sits on. Today rk_cdn_dp looks up fusb302_get_dp_altmode_state at runtime so that either the real implementation in fusb302(4) or the manual-override implementation in the helper could win, depending on load order. Functionally yes — rk_cdn_dp has a hard dependency on fusb302 being attached and having walked PD altmode entry; without that, the Cadence framer can't be brought up. The runtime lookup obscured that dependency instead of declaring it.  

In the next push I'll drop the linker_file_lookup_symbol call entirely, replace it with a direct extern int fusb302_get_dp_altmode_state(...) declared  in fusb302_var.h (already where the other cross-module fusb302 entry points live after the D4 style pass), and add MODULE_DEPEND(rk_cdn_dp, fusb302, 1,1, 1) so the relationship is declared in the normal place. The rk3399_fusb302_helper module goes away in the same change, which removes the reason the lookup ever existed. 

And on the Makefile note — good catch, I'll update SUBDIR in sys/modules/rockchip/Makefile so the module actually builds when someone runs make -C sys/modules/rockchip. Folding that into the same push. 

Thanks again for the substantive look — these are the architectural pushes the series needs before it lands.

mhorne: Great, I like the sounds of this plan!

PR #2238, opened 2026-05-26
socket: Add missing so_rerror handling cases

stephane-rochoy-stormshield:
Also wrap the logic into `soerror()` and `soerrorfetch()` to help handle
`so_error` and `so_rerror` consistently.

My tests regarding this patch are quite limited: the problem was discovered in
a very specific context were a lot of interfaces were created which triggered a
lot of requests sent to a `PF_ROUTE` socket.

And I didn't find any Kuya test regarding `SO_ERROR` and the fact, that since
`so_rerror` was introduced, it may be necessary to call it twice to pop all the
errors currently bound to the socket.

Also, I was unable to find a way to simulate something that looks like an mbuf
starvation to trigger the use of `so_rerror`, so feel free to give me some hints
if possible to add tests ;)

### DIFF
--- a/lib/libsys/getsockopt.2
+++ b/lib/libsys/getsockopt.2
@@ -25,7 +25,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd April 21, 2026
+.Dd May 26, 2026
 .Dt GETSOCKOPT 2
 .Os
 .Sh NAME
@@ -517,6 +517,9 @@ returns any pending error on the socket and clears
 the error status.
 It may be used to check for asynchronous errors on connected
 datagram sockets or for other asynchronous errors.
+Note that more than one error may be pending on the socket when
+.Dv SO_RERROR
+is set.
 .Dv SO_RERROR
 indicates that receive buffer overflows should be handled as errors.
 Historically receive buffer overflows have been ignored and programs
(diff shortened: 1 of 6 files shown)


lib/libsys/getsockopt.2
```diff
@@ -513,6 +513,9 @@ returns any pending error on the socket and clears
 the error status.
 It may be used to check for asynchronous errors on connected
 datagram sockets or for other asynchronous errors.
+Note that more than one error may be set one the socket when
```
dodo040: Small typo / wording suggestion:
```suggestion
Note that more than one error may be pending on the socket when
```

stephane-rochoy-stormshield: Good catch!

sys/kern/uipc_usrreq.c
```diff
@@ -2268,10 +2268,9 @@ uipc_soreceive_dgram(struct socket *so, struct sockaddr **psa, struct uio *uio,
 	while ((m = so->so_rcv.uxdg_peeked) == NULL &&
 	    (sb = TAILQ_FIRST(&so->so_rcv.uxdg_conns)) == NULL &&
 	    (m = STAILQ_FIRST(&so->so_rcv.uxdg_mb)) == NULL) {
-		if (so->so_error) {
-			error = so->so_error;
-			if (!(flags & MSG_PEEK))
-				so->so_error = 0;
+		if (soerror(so)) {
+			error = flags & MSG_PEEK ? soerror(so) : \
+			    soerrorfetch(so);
```
TeddyEngel, line 2273: minor issue, but there aren't many continuation uses outside of macro definitions - and it looks like the ternary would fit in one line < 80 characters? other ternaries seem to either break to nextline after : without continuation or be in one line

stephane-rochoy-stormshield: Well, I initially rearranged this ternary due to `checkstyle9.pl`:
```
WARNING: line over 80 characters
#2272: FILE: sys/kern/uipc_usrreq.c:2272:
+			error = flags & MSG_PEEK ? soerror(so) : soerrorfetch(so);
```
So let's remove the useless backslash.

PR #2239, opened 2026-05-26
loader.efi: Search boot device before foreign ZFS pools

kfv:
This fixes [Bug 295289](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=295289), where booting a 16.0-CURRENT memstick on a system with an installed ZFS root incorrectly boots from the internal disk instead of the installer media.

### DIFF
--- a/stand/efi/loader/main.c
+++ b/stand/efi/loader/main.c
@@ -365,6 +365,49 @@ try_as_currdev(pdinfo_t *hd, pdinfo_t *pp)
 }
 
 /*
+ * Search the boot device first (i.e. the ESP and any sibling partitions).
+ * Per the UEFI specification, filesystems on other devices must not be
+ * preferred until the boot device has been fully exhausted.
+ */
+static int
+try_boot_device_partitions(void)
+{
+	pdinfo_t *dp, *pp, *espdp;
+	CHAR16 *text;
+
+	dp = efiblk_get_pdinfo_by_handle(boot_img->DeviceHandle);
+	if (dp == NULL)
+		return (ENOENT);
+
+	text = efi_devpath_name(dp->pd_devpath);
+	if (text != NULL) {
+		printf("Trying ESP: %S\n", text);
+		efi_free_devpath_name(text);
+	}
+	set_currdev_pdinfo(dp);
+	if (sanity_check_currdev())
+		return (0);
+
+	if (dp->pd_parent == NULL)
+		return (ENOENT);
+
+	espdp = dp;
+	dp = dp->pd_parent;
+	STAILQ_FOREACH(pp, &dp->pd_part, pd_link) {
+		if (espdp == pp)
+			continue;
+		text = efi_devpath_name(pp->pd_devpath);
+		if (text != NULL) {
+			printf("Trying: %S\n", text);
+			efi_free_devpath_name(text);
+		}
+		if (try_as_currdev(dp, pp))
+			return (0);
+	}
+	return (ENOENT);
+}
+
+/*
  * Sometimes we get filenames that are all upper case
  * and/or have backslashes in them. Filter all this out
  * if it looks like we need to do so.
@@ -535,10 +578,9 @@ match_boot_info(char *boot_info, size_t bisz)
 static int
 find_currdev(bool do_bootmgr, char *boot_info, size_t boot_info_sz)
 {
-	pdinfo_t *dp, *pp;
+	pdinfo_t *dp;
 	EFI_DEVICE_PATH *devpath, *copy;
 	EFI_HANDLE h;
-	CHAR16 *text;
 	struct devsw *dev;
 	int unit;
 	uint64_t extra;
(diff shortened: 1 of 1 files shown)


stand/efi/loader/main.c
```diff
@@ -364,6 +364,49 @@ try_as_currdev(pdinfo_t *hd, pdinfo_t *pp)
...
+
+	dp = efiblk_get_pdinfo_by_handle(boot_img->DeviceHandle);
+	if (dp == NULL)
+		return (ENOENT);
+
+	text = efi_devpath_name(dp->pd_devpath);
+	if (text != NULL) {
+		printf("Trying ESP: %S\n", text);
+		efi_free_devpath_name(text);
+	}
```
bsdimp, line 386: We usually don't print things like that, but it's likely OK.

bsdimp:
This looks really good.... maybe too many printfs... I'm unsure about the extra trying messages... In the past in the boot loader we've tried very hard not to do that.

I'm not 100% sure it's right, even though it looks good, for the case where there's a mix of UFS and ZFS bootable images around. It's one of the reasons that I've been a little slow to fix this. What are your thoughts about what this patch will do in that case?

kfv:
> I'm not 100% sure it's right, even though it looks good, for the case where there's a mix of UFS and ZFS bootable images around. It's one of the reasons that I've been a little slow to fix this. What are your thoughts about what this patch will do in that case?

I had the same concern, so I ran the following set of tests in VMware Fusion with all possible combinations I could think of.

## Environment

- UEFI boot, Secure Boot off.

- Patch loader applied to all EFI partitions.

- Disks:
  - nda0: initial bootable ZFS install (zroot)
  - nda1: 16.0-CURRENT UFS installer memstick image (EFI + UFS)
  - nda2: bootable UFS install
  - nda3: non-bootable data-only ZFS (no EFI)
  - nda4: EFI + intentionally non-bootable UFS root
  - nda5: second bootable ZFS install (nroot)

For reference, the `gpart show` and `efibootmgr -v` outputs are attached below:

```gpart
=>       40  157286320  nda0  GPT  (75G)
         40     532480     1  efi  (260M)
     532520       2008        - free -  (1.0M)
     534528    4194304     2  freebsd-swap  (2.0G)
    4728832  152555520     3  freebsd-zfs  (73G)
  157284352       2008        - free -  (1.0M)

=>     34  2732312  nda1  GPT  (1.3G)
       34    66584     1  efi  (33M)
    66618  2665728     2  freebsd-ufs  (1.3G)

=>      40  41942960  nda2  GPT  (20G)
        40    532480     1  efi  (260M)
    532520  39313408     2  freebsd-ufs  (19G)
  39845928   2097072     3  freebsd-swap  (1.0G)

=>      40  41942960  nda3  GPT  (20G)
        40  41942960     1  freebsd-zfs  (20G)

=>      40  41942960  nda4  GPT  (20G)
        40    532480     1  efi  (260M)
    532520  41410480     2  freebsd-ufs  (20G)

=>      40  41942960  nda5  GPT  (20G)
        40    532480     1  efi  (260M)
    532520      2008        - free -  (1.0M)
    534528   4194304     2  freebsd-swap  (2.0G)
   4728832  37212160     3  freebsd-zfs  (18G)
  41940992      2008        - free -  (1.0M)

=>     63  2012243  cd0  MBR  (3.8G)
       63  2012243       - free -  (3.8G)

=>     63  2012243  iso9660/15_0_RELEASE_AARCH64_DVD  MBR  (3.8G)
       63  2012243                                    - free -  (3.8G)
```

```efibootmgr
Boot to FW : false
BootCurrent: 0000
BootOrder  : 0000, 0001, 0004, 0002, 0003, 0005, 0006, 0008, 0009, 000A
+Boot0000* EFI VMware Virtual NVME Namespace (NSID 1) PciRoot(0x0)/Pci(0x15,0x1)/Pci(0x0,0x0)/NVMe(0x1,00-00-00-00-00-00-00-00)
 Boot0001* EFI VMware Virtual SATA CDROM Drive (1.0) PciRoot(0x0)/Pci(0x11,0x0)/Pci(0x3,0x0)/Sata(0x1,0x0,0x0)
 Boot0004* EFI Internal Shell (Unsupported option) MemoryMapped(0xb,0x27f617000,0x27faf03ff)/FvFile(7c04a583-9e3e-4f1c-ad65-e05268d0b4d1)
 Boot0002* EFI VMware Virtual NVME Namespace (NSID 2) PciRoot(0x0)/Pci(0x15,0x1)/Pci(0x0,0x0)/NVMe(0x2,00-00-00-00-00-00-00-00)
 Boot0003* EFI VMware Virtual NVME Namespace (NSID 3) PciRoot(0x0)/Pci(0x15,0x1)/Pci(0x0,0x0)/NVMe(0x3,00-00-00-00-00-00-00-00)
 Boot0005* EFI Network PciRoot(0x0)/Pci(0x15,0x0)/Pci(0x0,0x0)/MAC(000c293decd8,0x0)
 Boot0006* EFI Network 1 PciRoot(0x0)/Pci(0x16,0x0)/Pci(0x0,0x0)/MAC(000c293dece2,0x0)
 Boot0008* EFI VMware Virtual NVME Namespace (NSID 4) PciRoot(0x0)/Pci(0x15,0x1)/Pci(0x0,0x0)/NVMe(0x4,00-00-00-00-00-00-00-00)
 Boot0009* EFI VMware Virtual NVME Namespace (NSID 5) PciRoot(0x0)/Pci(0x15,0x1)/Pci(0x0,0x0)/NVMe(0x5,00-00-00-00-00-00-00-00)
 Boot000A* EFI VMware Virtual NVME Namespace (NSID 6) PciRoot(0x0)/Pci(0x15,0x1)/Pci(0x0,0x0)/NVMe(0x6,00-00-00-00-00-00-00-00)


Unreferenced Variables:
```

## Test intent

- Bootable selected device should win first (UFS or ZFS).
- Relaxed foreign-ZFS fallback should still work when selected device has no bootable root.

## Results

1. Booting from nda0 (bootable ZFS install, pool zroot) -> boots nda0 (zroot)
2. Booting from nda1 (UFS installer memstick) -> boots installer media (the original bug case)
3. Booting from nda2 (bootable UFS install) -> boots nda2
4. Booting from nda3 (data-only ZFS, non-bootable) -> falls through and boots nda0 ZFS (expected as relaxed behaviour)
5. Booting from nda4 (EFI + non-bootable UFS root) -> falls through and boots nda0 ZFS (expected as relaxed behaviour)
6. Booting from nda5 (bootable ZFS install, pool nroot) -> boots nda5 (nroot)

## Conclusion

This matches intended semantics of the patch, I believe.

- We now honour boot-device-first behaviour (UFS installer on selected media no longer gets preempted by unrelated internal ZFS pools).
- We preserve relaxed fallback behaviour when the selected device is not bootable (e.g. nda3/nda4 cases still find a foreign bootable ZFS root).
- Both bootable UFS and bootable ZFS selected-device boots continue to work.

So in mixed environments, the patch changes priority from “global ZFS probe first” to “selected boot device first, then global ZFS probe if relaxed policy”.

stand/efi/loader/main.c
```diff
@@ -606,65 +648,42 @@ find_currdev(bool do_bootmgr, char *boot_info, size_t boot_info_sz)
 		return (0);
 #endif /* MD_IMAGE_SIZE */
 
-#ifdef EFI_ZFS_BOOT
-	zfsinfo_list_t *zfsinfo = efizfs_get_zfsinfo_list();
-	zfsinfo_t *zi;
+	if (try_boot_device_partitions() == 0)
```
bsdimp, line 651: This needs a big block comment above it. I know a couple of other things were added w/o it, but the comments are supposed to document the order we do things in.

stand/efi/loader/main.c
```diff
@@ -606,65 +648,42 @@ find_currdev(bool do_bootmgr, char *boot_info, size_t boot_info_sz)
...
+		/*
+		 * Try ZFS pool(s) on the boot device not reachable via
+		 * the partition walk above.
+		 */
+		STAILQ_FOREACH(zi, zfsinfo, zi_link) {
+			if (zi->zi_handle != boot_img->DeviceHandle)
+				continue;
+			printf("Trying ZFS pool 0x%jx\n", zi->zi_pool_guid);
+			if (probe_zfs_currdev(zi->zi_pool_guid))
+				return (0);
```
bsdimp, line 668: I think this for loop belongs inside a `#ifdef EFI_ZFS_BOOT` in try_boot_device_partitions. I'd do it first so that we always have a ZFS over UFS policy. Otherwise, for a drive with both UFS and ZFS, we'll prefer UFS, but in a system with two disks we'll prefer ZFS. This is the only case that wasn't in the test procedure, and although rare, it should be done to keep things consistent. We have to prefer something, since there's at least an implicit order in the filesystems we probe.

Ideally, I'd like to see ZFS made more regular so we don't need special cases. However, that's a lot of work, and it's not clear that we can ultimately do it.

bsdimp:
So the first time I read this I had a reaction that suggested caution. I've re-read this patch a lot and it's close and I've discovered my unease. We should always prefer ZFS to UFS, and there's one case this isn't the case: if there's both a bootable ZFS dataset and a bootable UFS partition, we'll choose UFS in this one case only. That irregularity is what I was worried about. I couldn't come up with any other scenarios outside of your test cases. It's also a rare case, but I've encountered it in upgrade situations.

Thank you so much for getting this going. Please accept my apologies for delaying on getting this in due to my misgivings. I've been burned by order changes fairly often, so I wanted to be extra sure. It didn't help that the extra printfs lead me to a conclusion that turned out to be false (which I won't repeat here and again apologize for).

Do you have a script to create that test setup? I'd love to make things more testable here. That won't gate committing this once the small move the ZFS probe like I suggested.

kfv:
Thanks for your review and accept my apologies for the late response.

> So the first time I read this I had a reaction that suggested caution. I've re-read this patch a lot and it's close and I've discovered my unease. We should always prefer ZFS to UFS, and there's one case this isn't the case: if there's both a bootable ZFS dataset and a bootable UFS partition, we'll choose UFS in this one case only. That irregularity is what I was worried about. I couldn't come up with any other scenarios outside of your test cases. It's also a rare case, but I've encountered it in upgrade situations.

I see your point now. I didn’t consider a single device with both filesystems, so, neither in my testing nor in my head did I ever hit a single drive of UFS+ZFS bootable layout, you're right. For that I will soon give it a try and will let you know.

> Thank you so much for getting this going. Please accept my apologies for delaying on getting this in due to my misgivings. I've been burned by order changes fairly often, so I wanted to be extra sure. It didn't help that the extra printfs lead me to a conclusion that turned out to be false (which I won't repeat here and again apologize for).

I understand. I just saw there were already  four "Trying..." prints, so, yeah, that was it. But I'm more than happy to remove any unnecessary prints or perform cleanups in general. And don't worry about what we discussed on the other side, it's all water under the bridge.

bsdimp:
Great! when you have changes, I'll review them qiuckly.

bsdimp:

bsdimp:
So I was testing `ghpr`, and had this staged and it was pushed in with the flaw we talked about. So, to recover, if you're willing, can you do the change against main and open a new PR? Does that work for you? Less churn in the repo for what already is a much better than before change if we don't go the backout route.

kfv:
Hey, sure! And thanks for landing the change, by the way.

Also, sorry for being a bit tied up last week. I unfortunately made a careless submission that introduced an incompatibility in the C++ stack, which ended up breaking a number of ports and receiving a few reports. I had to get that cleaned up first. Now that that's resolved, I should be able to spend some time on this next week. I actually love to get a bit more involved with the boot loader, so count me in. I'll keep you posted as I make progress.

Thanks again for all the support, and please let me know if there's anything in particular you'd like me to look at or keep in mind as we move forward.

Cheers!

PR #2240, opened 2026-05-26
bsdinstall: Create separate dataset for /root

ricardobranco777:
Otherwise there could be data loss in ZFS boot environment rollbacks.

Fixes: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=295619


bsdimp:
So this was done in 2014 and then reverted saying more work was needed. It's unclear what that addition work is.

PR #2243, opened 2026-05-28
libc: Guard mergesort() allocation size arithmetic

kfv:
`mergesort()` used unchecked `size_t` arithmetic for allocation and subsequent range calculations. For large inputs, `nmemb * size` could wrap, leading to `stack-buffer-underflow` and out-of-bounds access.

### Diff

- Add an overflow guard in `mergesort()` to reject combinations that would overflow `nmemb * size + PSIZE`.
- Return `-1` with `errno = EINVAL` for invalid size combinations instead of proceeding with unsafe pointer/length arithmetic.

### Note

- No ABI/API changes.
- Normal sorting semantics are unchanged for valid inputs.

### DIFF
--- a/lib/libc/stdlib/merge.c
+++ b/lib/libc/stdlib/merge.c
@@ -49,6 +49,7 @@
 #include <sys/param.h>
 
 #include <errno.h>
+#include <stdckdint.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -109,7 +110,7 @@ mergesort_b(void *base, size_t nmemb, size_t size, cmp_t cmp)
 mergesort(void *base, size_t nmemb, size_t size, cmp_t cmp)
 #endif
 {
-	size_t i;
+	size_t i, nbytes, asize;
 	int sense;
 	int big, iflag;
 	u_char *f1, *f2, *t, *b, *tp2, *q, *l1, *l2;
@@ -123,16 +124,21 @@ mergesort(void *base, size_t nmemb, size_t size, cmp_t cmp)
 	if (nmemb == 0)
 		return (0);
 
+	if (ckd_mul(&nbytes, nmemb, size) || ckd_add(&asize, nbytes, PSIZE)) {
+		errno = EINVAL;
+		return (-1);
+	}
+
 	iflag = 0;
 	if (__is_aligned(size, ISIZE) && __is_aligned(base, ISIZE))
 		iflag = 1;
 
-	if ((list2 = malloc(nmemb * size + PSIZE)) == NULL)
+	if ((list2 = malloc(asize)) == NULL)
 		return (-1);
 
 	list1 = base;
 	setup(list1, list2, nmemb, size, cmp);
-	last = list2 + nmemb * size;
+	last = list2 + nbytes;
 	i = big = 0;
 	while (*EVAL(list2) != last) {
 	    l2 = list1;
@@ -227,10 +233,10 @@ COPY:	    			b = t;
 	    tp2 = list1;	/* swap list1, list2 */
 	    list1 = list2;
 	    list2 = tp2;
-	    last = list2 + nmemb*size;
+	    last = list2 + nbytes;
 	}
 	if (base == list2) {
-		memmove(list2, list1, nmemb*size);
+		memmove(list2, list1, nbytes);
 		list2 = list1;
 	}
 	free(list2);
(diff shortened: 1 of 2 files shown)


kfv:
Cc: @clausecker

lib/libc/stdlib/merge.c
```diff
@@ -123,6 +123,11 @@ mergesort(void *base, size_t nmemb, size_t size, cmp_t cmp)
 	if (nmemb == 0)
 		return (0);
 
+	if (nmemb > (SIZE_MAX - PSIZE) / size) {
```
clausecker: Perhaps something like

```suggestion
	size_t *dummy;
        
	if (ckd_mul(&dummy, nmemb, size)) {
```

instead?  But please declare dummy near the top of the file.

clausecker: Near the top of the function of course.  Possibly also needs to account for the extra PSIZE.

kfv: Brilliant! You're right, I hadn't considered `<stdckdint.h>`. Glad we have it now. I've updated the diff accordingly, with two minor adjustments:

1. To avoid redundant recalculations, I replaced `dummy` with two named variables (`nbytes` and `asize`), which are then reused where appropriate.
2. Correctly checking `(nmemb * size + PSIZE)` requires chaining both `ckd_mul` and `ckd_add`, which the updated code now does.

Let me know if anything. Thanks!

clausecker:
I think the patch looks ok.  
One thing I am thinking about is whether we should actually return `ENOMEM` on overflow to be consistent with other allocating routines like `calloc`.  Let me discuss this with some people and then land the change with what we agreed on.

bms:
+1 for EINVAL given it's what the malloc family from jemalloc do.

clausecker:
> +1 for EINVAL given it's what the malloc family from jemalloc do.

This is incorrect, jemalloc treats overflow as an OOM condition and returns ENOMEM: https://github.com/freebsd/freebsd-src/blob/main/contrib/jemalloc/src/jemalloc.c#L2502

kfv:
> I think the patch looks ok. One thing I am thinking about is whether we should actually return `ENOMEM` on overflow to be consistent with other allocating routines like `calloc`. Let me discuss this with some people and then land the change with what we agreed on.

Your concern is valid, Robert. However, `mergesort()` fails before calling `malloc()` when `nmemb` and `size` don't form a representable workspace, therefore making it an invalid argument pair, in the same class as the `size < PSIZE / 2` check earlier in the function. That's why `EINVAL` seemed like the better fit to me.

> > +1 for EINVAL given it's what the malloc family from jemalloc do.
> 
> This is incorrect, jemalloc treats overflow as an OOM condition and returns ENOMEM: https://github.com/freebsd/freebsd-src/blob/main/contrib/jemalloc/src/jemalloc.c#L2502

You're right, `jemalloc` uses `ENOMEM` on overflow, but `jemalloc` and `mergesort()` operate at different semantic layers with different contracts. `errno` should reflect the API contract of the function reporting the error, not necessarily the implementation detail underneath.

Besides, I also need to update the man page to cover this case. I'll do shortly. And please let me know if you think `ENOMEM` could still be the better choice here.

clausecker:
Both `EINVAL` and `ENOMEM` are reasonable.

My way of thinking is that if `nmemb * size + PSIZE` overflows a `size_t`, we will be unable to allocate that much memory, as that's more than what the computer can handle.  So `ENOMEM` is a truthful error return.  The same logic applies to routines like `calloc`.

But likewise, as the caller has passed in an array that is supposedly of size `nmemb * size`, he has lied to us in one way or another if `nmemb * size` overflows a `size_t`, which justified the `EINVAL` error code.

I think it probably doesn't matter too much which error we return; I just want to seek some consensus before baking this decision into our API.

kfv:
Thanks for the explanation, I understand your reasoning now. I'd just add that an overflow can still occur even when the caller is conceptually honest about the intended array dimensions; the allocation simply becomes unrepresentable once `nmemb * size + PSIZE` exceeds `SIZE_MAX`. The rationale makes sense regardless.

kfv:
Just updated the man page. We can update it later if we decide to go with `ENOMEM`.

clausecker:
Please remember to bump the man page date when you change a man page.

kfv:
That I always forget, my apologies. Done.

clausecker:

PR #2246, opened 2026-05-30
kern/kexec: Input validation checks for kexec_load(2) arguments

geekrusman:
Missing sanity checks in kexec_load(2) allowed malformed arguments to trigger a kernel panic during syscall execution. Added NULL pointer checks, integer overflow checks for size parameters (used in segment_fits), and segment overlap checks for the mem parameter.

PR:             295521
Reported by:    Changseon Lee

Fixes: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=295521

### DIFF
--- a/sys/kern/kern_kexec.c
+++ b/sys/kern/kern_kexec.c
@@ -154,6 +154,42 @@ pa_for_pindex(struct kexec_segment_stage *segs, int count, vm_pindex_t pind)
 	panic("No segment for pindex %ju\n", (uintmax_t)pind);
 }
 
+/* Perform sanity checks on the kexec segments */
+static int
+validate_kexec_segs(struct kexec_segment *segs, u_long nseg)
+{
+	vm_paddr_t v, prev_end;
+	int i;
+
+	for (i = 0; i < nseg; i++) {
+		v = (vm_paddr_t)(uintptr_t)segs[i].mem;
+
+		if (segs[i].bufsz > 0 && segs[i].buf == NULL)
+			return (EINVAL);
+
+		if (segs[i].memsz == 0)
+			return (EINVAL);
+
+		if (segs[i].bufsz > segs[i].memsz)
+			return (EINVAL);
+
+		// Check for overflows in the mem range
+		if (v + segs[i].memsz - 1 < v)
+			return (EINVAL);
+	}
+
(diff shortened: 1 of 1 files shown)


sys/kern/kern_kexec.c
```diff
@@ -154,6 +154,42 @@ pa_for_pindex(struct kexec_segment_stage *segs, int count, vm_pindex_t pind)
...
+			return (EINVAL);
+
+		if (segs[i].memsz == 0)
+			return (EINVAL);
+
+		if (segs[i].bufsz > segs[i].memsz)
+			return (EINVAL);
+
+		// Check for overflows in the mem range
+		if (v + segs[i].memsz < v)
```
bsdimp: doesn't this have an off by one error?

So if v the last page of the address space, and memsz is 1 page, then v + memsz is 0. Shouldn't this have a -1 at the end?

sys/kern/kern_kexec.c
```diff
@@ -154,6 +154,42 @@ pa_for_pindex(struct kexec_segment_stage *segs, int count, vm_pindex_t pind)
...
+	}
+
+	/* Check for segment overlaps in the sorted segment list */
+	for (i = 1; i < nseg; i++) {
+		prev_end = (vm_paddr_t)(uintptr_t)segs[i-1].mem + segs[i-1].memsz;
+		v = (vm_paddr_t)(uintptr_t)segs[i].mem;
+
+		if (v < prev_end)
+			return (EINVAL);
+	}
```
bsdimp, line 188: Is there a requirement that this list be sorted?

sys/kern/kern_kexec.c
geekrusman: You're right, @bsdimp. I've addressed this in the latest patch.

sys/kern/kern_kexec.c
geekrusman: Correct, @bsdimp. The overlap check relies on the segments being sorted first, which is why it called immediately after the qsort call at line 231.

"qsort(segtmp, nseg, sizeof(*segtmp), seg_cmp);"

geekrusman:
@bsdimp Could you help review the code changes?

PR #2247, opened 2026-05-31
mrsas: Free event log info DMA resources on timeout

lhx2000:
mrsas_get_seq_num() allocates event log info DMA resources before issuing MR_DCMD_CTRL_EVENT_GET_INFO. If mrsas_issue_blocked_cmd() returns ETIMEDOUT, the function jumps to dcmd_timeout before calling mrsas_free_evt_log_info_cmd(), leaving those resources allocated.

Move mrsas_free_evt_log_info_cmd() to the common exit path so the resources are released on both the successful path and the DCMD timeout path.


bsdimp (approved):
I'm not super familiar with mrsas, but it's similar to mpr and mps and I belive for both those this would be right, so I'm pretty sure this is right.

PR #2248, opened 2026-06-01
lib/libc/tests/gen: add misc fts traversal tests

Jeetubhati:
Extend fts_misc_test.c with additional test cases covering
traversal invariants and regression scenarios:

- FTS_NOCHDIR with absolute paths: application chdir between
  fts_read() calls does not corrupt traversal; CWD remains
  unchanged throughout
- fts_name is always NUL-terminated with correct fts_namelen
- FTS_D/FTS_DP are paired and fts_level increments correctly
- FTSENT fts_errno/fts_dev/fts_ino/fts_nlink are correct
- circular symlink loop under FTS_PHYSICAL terminates
- cycle via symlink under FTS_LOGICAL yields FTS_DC
- fts_close after root deletion must not crash
- fts_close after root rename restores CWD (SVN r77497)
- FTS_NOCHDIR + empty directory does not corrupt path (SVN r49772)
- FTS_NS entry has non-zero fts_errno
- FTS_XDEV does not cross mount points (tested with tmpfs)

Reviewed by: asomers

### DIFF
--- a/lib/libc/tests/gen/fts_misc_test.c
+++ b/lib/libc/tests/gen/fts_misc_test.c
@@ -1,16 +1,23 @@
-/*-
+/*
  * Copyright (c) 2025 Klara, Inc.
+ * Copyright (c) 2026 Jitendra Bhati
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <sys/mount.h>
+#include <sys/param.h>
 #include <sys/stat.h>
+#include <sys/syslimits.h>
+#include <sys/uio.h>
 
+#include <errno.h>
 #include <fcntl.h>
 #include <fts.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
 #include <unistd.h>
 
 #include <atf-c.h>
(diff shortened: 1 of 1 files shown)


lib/libc/tests/gen/Makefile
```diff
@@ -14,6 +14,7 @@ ATF_TESTS_C+=		fts_children_test
 ATF_TESTS_C+=		fts_misc_test
 ATF_TESTS_C+=		fts_open_test
 ATF_TESTS_C+=		fts_options_test
+ATF_TESTS_C+=		fts_regress_test
```
asomers: This change snuck in by mistake; this PR doesn't add that file.

asomers (changes requested):

lib/libc/tests/gen/fts_misc_test.c
```diff
@@ -69,10 +75,488 @@ ATF_TC_BODY(fts_unrdir_nochdir, tc)
...
+	ATF_REQUIRE(fts_read(fts) != NULL);
+	ATF_REQUIRE_EQ(0, unlink("dir/file"));
+	ATF_REQUIRE_EQ(0, rmdir("dir"));
+
+	/* Drain traversal -- errors expected, no crash. */
+	while (fts_read(fts) != NULL)
+		;
+
+	/* fts_close() must not crash regardless of return value. */
+	(void)fts_close(fts);
```
jillest: This seems a little light on assertions. An error could be checked for, and the cwd like the below `close_after_root_moved`.

jillest (commented):
These seem useful tests.

lib/libc/tests/gen/Makefile
Jeetubhati: Sorry about that! I was testing fts_regress_test locally and forgot to clean it up. It’s all fixed now.

lib/libc/tests/gen/fts_misc_test.c
Jeetubhati: Added errno check after drain loop and verify CWD after fts_close()
Follows the pattern established in close_after_root_moved.

asomers (approved):

jillest (approved):

lib/libc/tests/gen/fts_misc_test.c
```diff
@@ -69,10 +75,511 @@ ATF_TC_BODY(fts_unrdir_nochdir, tc)
...
+	char *paths[] = { "dir", NULL };
+	FTS *fts;
+	FTSENT *ent;
+	int crossed;
+
+	ATF_REQUIRE_EQ(0, mkdir("dir", 0755));
+	ATF_REQUIRE_EQ(0, mkdir("dir/mnt", 0755));
+	ATF_REQUIRE_EQ(0, close(creat("dir/file", 0644)));
+
+	if (mount("tmpfs", "dir/mnt", 0, NULL) != 0)
```
asomers: One more problem: it's almost impossible to use mount(2) correctly on FreeBSD.  In fact, it doesn't work as you used it here.  Instead, you should use nmount(2).  Can you please change it?

asomers (changes requested):

lib/libc/tests/gen/fts_misc_test.c
Jeetubhati: I fixed this by replacing mount(2) with nmount(2).

lib/libc/tests/gen/fts_misc_test.c
```diff
@@ -533,32 +534,44 @@ ATF_TC_HEAD(xdev, tc)
...
+        FTSENT *ent;
+        int crossed;
+
+        ATF_REQUIRE_EQ(0, mkdir("dir", 0755));
+        ATF_REQUIRE_EQ(0, mkdir("dir/mnt", 0755));
+        ATF_REQUIRE_EQ(0, close(creat("dir/file", 0644)));
+
+        iov[0].iov_base = (void *)"fstype";
+        iov[0].iov_len  = sizeof("fstype");
+        iov[1].iov_base = fstype;
```
asomers: Why write "fstype" and "fspath" inline, but create a variable for "tmpfs"?  It's inconsistent.  I think you should put "tmpfs" inline, too.

lib/libc/tests/gen/fts_misc_test.c
```diff
@@ -533,32 +534,44 @@ ATF_TC_HEAD(xdev, tc)
...
+        iov[1].iov_len  = sizeof(fstype);
+        iov[2].iov_base = (void *)"fspath";
+        iov[2].iov_len  = sizeof("fspath");
+        iov[3].iov_base = fspath;
+        iov[3].iov_len  = strlen(fspath) + 1;
```
asomers: Why use size of fstype, but strlen for fspath?

lib/libc/tests/gen/fts_misc_test.c
```diff
@@ -533,32 +534,44 @@ ATF_TC_HEAD(xdev, tc)
...
-	    "FTS_XDEV must not descend into tmpfs mount point");
-
-	ATF_REQUIRE_EQ_MSG(0, fts_close(fts), "fts_close(): %m");
+        struct iovec iov[4];
+        char fstype[] = "tmpfs";
+        char fspath[] = "dir/mnt";
+        char *paths[] = { "dir", NULL };
+        FTS *fts;
+        FTSENT *ent;
+        int crossed;
```
asomers: Again, this should be a bool.

lib/libc/tests/gen/fts_misc_test.c
```diff
@@ -533,32 +534,44 @@ ATF_TC_HEAD(xdev, tc)
...
-	crossed = 0;
-	while ((ent = fts_read(fts)) != NULL) {
-		if (strcmp(ent->fts_name, "inside") == 0)
-			crossed = 1;
-	}
-	ATF_CHECK_MSG(crossed == 0,
-	    "FTS_XDEV must not descend into tmpfs mount point");
-
-	ATF_REQUIRE_EQ_MSG(0, fts_close(fts), "fts_close(): %m");
+        struct iovec iov[4];
```
asomers: It seems like you replaced hard tabs with eight spaces everywhere in this function.  I don't think that was intended, and it's certainly not the standard style.  Please switch back.

asomers (changes requested):

lib/libc/tests/gen/fts_misc_test.c
Jeetubhati: Fixed. Also put "tmpfs" inline for consistency with the other
string arguments.

lib/libc/tests/gen/fts_misc_test.c
Jeetubhati: Now that both are inline string literals, using
sizeof consistently is correct. Fixed.

asomers (approved):

PR #2249, opened 2026-06-01
nat64lsn: Fix type confusion panic when using wrong NAT64 instance type

TeddyEngel:
When an ipfw rule references a NAT64 instance by name using 'nat64lsn', the kernel looks up the instance in the shared srvstate[] array without verifying the instance type.  If the named instance is instead a nat64clat or nat64stl instance, the code incorrectly casts the instance to nat64lsn_instance and dereferences the ->cfg pointer, which causes a kernel panic.

The root cause is that all NAT64 instance types share the same srvstate[] array but have different struct layouts.  For nat64lsn_instance, the field after 'no' is a pointer to nat64lsn_cfg.  For nat64clat_cfg, the same offset contains an embedded nat64_config struct.

Fixed by adding a type check after NAT64_LOOKUP() to verify that the instance's etlv matches IPFW_TLV_NAT64LSN_NAME before proceeding. If the type doesn't match, return IP_FW_DENY to reject the packet safely rather than crashing.

Tested with the official reproducer from the bug report on FreeBSD 16.0-CURRENT, no kernel panic with the change.

PR: [292023](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=292023)


bu7cher (commented):
Looks good to me.

spmzt:

PR #2251, opened 2026-06-02
fsck_msdosfs: don't report FSERROR for a lost chain that was cleared

Sheng5Qing:
While looking into why an SD card had to be re-inserted twice before it
would mount, I traced it to a stale FSERROR exit code from fsck_msdosfs.

In checklost(), when a lost cluster chain is found but reconnect() fails
(LOST.DIR full in my case), reconnect() returns FSERROR and the old code
ORs it into `mod` right away via `mod |= ret = reconnect(...)`. If the
following Clear succeeds, clearchain() frees the chain and sets FSFATMOD,
but the FSERROR bit is still set, so checkfilesys() returns 8 on a volume
that is actually clean. A mounter that keys off the exit code rejects it
on the first pass; the second pass returns 0 only because the first run
already repaired it.

The fix defers folding a reconnect() failure into `mod` until we know
whether the chain gets cleared. Cleared -> resolved, don't report FSERROR.
Not cleared -> keep FSERROR as before. FSFATAL still takes `ret != FSERROR`
so the `mod & FSFATAL` break is unchanged.

The patch now tracks whether FSERROR came from reconnect(), so the change is
limited to the reconnect-failure fallback path. FSERROR returned by checkchain()
keeps the old behavior.

Tested on arm64. Same corrupted FAT16 image (orphan chains + full LOST.DIR,
`fsck_msdos -p -f -y`): old binary returns 8 then 0 on a re-run, patched
returns 0 on the first run, and the repair output is identical between the
two. Also ran it over 135 randomly generated images (varying chain counts,
lengths, and a partly-full LOST.DIR so some chains reconnect and some get
cleared in the same run) and the patched binary never returned 8.

### DIFF
--- a/sbin/fsck_msdosfs/fat.c
+++ b/sbin/fsck_msdosfs/fat.c
@@ -1258,6 +1258,7 @@ checklost(struct fat_descriptor *fat)
 	cl_t head;
 	int mod = FSOK;
 	int dosfs, ret;
+	bool reconnect_failed;
 	size_t chains, chainlength;
 	struct bootblock *boot;
 
@@ -1283,19 +1284,32 @@ checklost(struct fat_descriptor *fat)
 			continue;
 		}
 		if (fat_is_cl_head(fat, head)) {
+			reconnect_failed = false;
 			ret = checkchain(fat, head, &chainlength);
 			if (ret != FSERROR && chainlength > 0) {
 				pwarn("Lost cluster chain at cluster %u\n"
 				    "%zd Cluster(s) lost\n",
 				    head, chainlength);
-				mod |= ret = reconnect(fat, head,
+				ret = reconnect(fat, head,
 				    chainlength);
+				/*
+				 * Defer folding a reconnect() failure into
+				 * mod: if the lost chain is cleared below, it
+				 * has been resolved and must not be reported as
+				 * an unrecovered error (FSERROR).
+				 */
+				if (ret == FSERROR)
+					reconnect_failed = true;
+				else
+					mod |= ret;
 			}
 			if (mod & FSFATAL)
 				break;
 			if (ret == FSERROR && ask(0, "Clear")) {
 				clearchain(fat, head);
 				mod |= FSFATMOD;
+			} else if (reconnect_failed) {
+				mod |= FSERROR;
 			}
 			chains--;
 		}


Sheng5Qing:
Ping -- could someone take a look at this when you get a chance?

@delphij you maintain sbin/fsck_msdosfs and committed a similar Android-originated
fix there last month (the FAT header correction cache one), so this might be in
your area.

Short version: in checklost(), a reconnect() failure ORs FSERROR into `mod`, and
when the chain is then cleared the bit is left set, so fsck exits 8 on a volume
it just fixed. The patch holds off on folding the reconnect() failure into `mod`
until it knows whether the chain gets cleared.

Tested on arm64 against a corrupted FAT16 image (orphan chains + full LOST.DIR):
old binary exits 8 then 0 on a re-run, patched exits 0 the first time, identical
repair output. Also fuzzed ~130 random images with a partly-full LOST.DIR (so
some chains reconnect and some get cleared in one run) and never saw a stray 8.

There's one behavior change I called out in the description (the else-if also
covers a declined Clear in interactive mode) -- happy to split that off if you'd
rather keep this minimal.

Sheng5Qing:
Update: I revised the patch to keep the scope narrower and rebased it on current `main`.

The new version only defers FSERROR when it came from `reconnect()`; FSERROR returned by `checkchain()` keeps the old behavior. This removes the unrelated interactive-mode behavior change I called out above.

Test summary is unchanged: on a minimal corrupted FAT16 image with orphan chains and a full `LOST.DIR`, the unpatched binary exits 8 on the first run and 0 on a re-run, while the patched binary exits 0 on the first run. The first-run diagnostics and repair output are identical; only the exit code changes. I also ran a randomized differential harness over 135 generated FAT16 images, including mixed reconnect-success / clear-fallback cases, with no patched run returning 8.

Sheng5Qing:
Reopening. This also came up on AOSP — enh@ said external/ changes need to go
upstream first and reassigned it to you:
https://android-review.googlesource.com/c/platform/external/fsck_msdos/+/4267862

The bug: when LOST.DIR is full, reconnect() returns FSERROR and the old code
folds it into mod right away. If the chain then gets cleared, the volume is
repaired but that stale FSERROR is still set, so fsck exits 8 on a clean fs.
On Android this makes vold refuse the card on first insertion.

@delphij mind taking a look?

delphij:
Hi, I noticed some additional issues so I have revised your change and created some test cases for it.

The core fix was amended as https://github.com/delphij/freebsd-src/commit/0492d152ef341ad8c21dc5b9b1f569b97be1f028 

I've also added some positive and negative test cases at https://github.com/delphij/freebsd-src/commit/4af0a2eebd6a32442b0f3b6417476168e8c0b2c9 . My changes were mainly to make `corrupted_lost_chain_left_alone` and `corrupted_lost_chain_reconnected` pass.

Could you please take a look at my version?

Sheng5Qing:
Nice catch. I focused on the stale FSERROR from reconnect() and missed that the same assignment also overwrites checkchain()'s status. Your version makes sense to me, and the tests cover the additional cases.

I'm happy to go with your version. Thanks for taking this on.

delphij:
Committed, thanks!

PR #2252, opened 2026-06-02
mvs: free DMA slots on channel attach failure

lhx2000:
mvs_ch_attach() allocates per-slot DMA maps with mvs_slotsalloc() before several later attach steps can fail.  The detach path frees those maps with mvs_slotsfree(), but attach failure does not reach mvs_ch_detach(), so the maps are leaked on the error path.

Call mvs_slotsfree() before destroying the channel mutex in the common attach failure path.


kev009:
I think it needs a couple related changes:  add mvs_dmafini(dev) after mvs_slotsfree(dev), and move both above the bus_release_resource() call so the error path mirrors mvs_ch_detach()

amotin (approved):
Makes sense, but it seems it also misses `mvs_dmafini()`.

PR #2254, opened 2026-06-02
netpfil: add PF netlink command decoding support

ishan0803:
Convert PFNL_CMD values in pf_nl.h from an enum to #define constants, add a pfnl_cmd table definition for mktable, and implement the corresponding command decoding helpers in libsysdecode.

This allows mktable to generate PF netlink command lookup tables and enables symbolic decoding of PF netlink commands.

Sponsored-by: Google LLC (GSoC 2026)


lib/libsysdecode/sysdecode.h
```diff
@@ -66,6 +66,7 @@ const char *sysdecode_ipproto(int _protocol);
 void	sysdecode_kevent_fflags(FILE *_fp, short _filter, int _fflags,
 	    int _base);
 const char *sysdecode_itimer(int _which);
+const char *sysdecode_pfnl_cmd(int cmd);
```
TeddyEngel, line 69: looks like the parameter should be int _cmd to match the existing convention in the header file

PR #2255, opened 2026-06-03
arm/bcm2835: Release mailbox resources on attach failure

lhx2000:
bcm_mbox_attach() allocates the mailbox memory resource before allocating the IRQ resource and installing the interrupt handler.

If IRQ allocation fails, the memory resource is leaked. If bus_setup_intr() fails, only the IRQ resource is released.

Track the memory and IRQ resource IDs separately and unwind all resources acquired so far before returning ENXIO.

### DIFF
--- a/sys/arm/broadcom/bcm2835/bcm2835_mbox.c
+++ b/sys/arm/broadcom/bcm2835/bcm2835_mbox.c
@@ -161,9 +161,8 @@ bcm_mbox_attach(device_t dev)
 {
 	struct bcm_mbox_softc *sc = device_get_softc(dev);
 	int i;
-	int rid = 0;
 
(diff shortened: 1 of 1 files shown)


VexedUXR (commented):
Looks okay to me, but it could be made simpler by passing the rid by value. That way you don't need two variables.

lhx2000:
Thanks, updated. I changed bus_alloc_resource_any() to pass the rid by value and removed the extra rid variable. I kept bus_release_resource() in the existing four-argument form with rid 0.

VexedUXR (approved):
The removal of the variable should be mentioned in the commit message, but the diff LGTM.

Thanks.

VexedUXR:

PR #2256, opened 2026-06-03
bnxt: clean up PCI mappings on attach mapping failure

lhx2000:
If bnxt_pci_mapping() succeeds in mapping BAR0 but fails while mapping BAR2, bnxt_attach_pre() previously skipped bnxt_pci_mapping_free() and only disabled bus mastering. Route the failure path through free_pci_map so any partially mapped PCI BAR resources are released.

Remove the now-unused pci_map_fail label.


kev009:
bnxt_attach_pre() also leaves the softc on the global pf_list on every failure path

PR #2257, opened 2026-06-03
fts: add fts regression tests

Jeetubhati:
Add ATF regression tests for previously-fixed fts(3) bugs:

- PR 45723: directory with read but no execute is traversed via FTS_DONTCHDIR fallback, not silently skipped (commit 1e03bff7f2b7)
- PR 196724: FTS_SLNONE must not be returned for a non-symlink; time-bounded race test runs for 1 second with concurrent file creation/deletion (commit bf4374c54589)
- PR 262038: readdir(2) errors produce FTS_DNR with fts_errno set, not silently treated as end-of-directory (commit 0cff70ca6654)
- SVN r246641: normal traversal works correctly with O_DIRECTORY fix in fts_safe_changedir() (commit f9928f1705ee)
- SVN r261589: no crash when tree modified during traversal; time-bounded race test runs for 1 second with concurrent file creation/deletion (commit c6d38f088e5c)

### DIFF
--- a/lib/libc/tests/gen/Makefile
+++ b/lib/libc/tests/gen/Makefile
@@ -14,6 +14,7 @@ ATF_TESTS_C+=		fts_children_test
 ATF_TESTS_C+=		fts_misc_test
 ATF_TESTS_C+=		fts_open_test
 ATF_TESTS_C+=		fts_options_test
+ATF_TESTS_C+=		fts_regress_test
 ATF_TESTS_C+=		fts_set_test
 ATF_TESTS_C+=		ftw_test
 ATF_TESTS_C+=		getentropy_test
@@ -99,6 +100,7 @@ LIBADD.fpsetround_test+=m
 LIBADD.siginfo_test+=	m
 
 LIBADD.nice_test+=	pthread
+LIBADD.fts_regress_test+=	pthread
 LIBADD.syslog_test+=	pthread
 
 CFLAGS+=		-I${.CURDIR}
--- a/lib/libc/tests/gen/fts_regress_test.c
+++ b/lib/libc/tests/gen/fts_regress_test.c
@@ -1,0 +1,315 @@
+/*
+ * Copyright (c) 2026 Jitendra Bhati
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+/*
+ * Regression tests for specific FreeBSD bug reports fixed in fts(3).
+ */
+
+#include <sys/stat.h>
+#include <sys/time.h>
+
+#include <errno.h>
+#include <fcntl.h>
+#include <fts.h>
+#include <pthread.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+
+#include <atf-c.h>
+
+/*
+ * Thrash function for file-based race tests: repeatedly creates and
+ * deletes a regular file at the given path.
+ */
+static volatile bool race_stop;
+
+static void *
+race_thrash(void *arg)
+{
+	const char *path = arg;
+
+	while (!race_stop) {
+		(void)close(creat(path, 0644));
+		(void)unlink(path);
+	}
+	return (NULL);
+}
+
+/*
+ * Thrash function for directory-based race tests: repeatedly removes
+ * and recreates a directory at the given path.
+ */
+static void *
+dir_thrash(void *arg)
+{
+	const char *path = arg;
+
+	while (!race_stop) {
+		(void)rmdir(path);
+		(void)mkdir(path, 0755);
+	}
+	return (NULL);
+}
+
+/*
+ * PR 45723: A directory with read but no execute permission must be
+ * traversed.  Before the fix, fts_build() gave up silently when
+ * chdir() failed, producing no output at all.  The fix falls back to
+ * FTS_DONTCHDIR mode so the directory is still traversed using full
+ * relative paths.
+ *
+ * Requires an unprivileged user because root ignores permissions.
+ */
+ATF_TC(read_no_exec_dir);
+ATF_TC_HEAD(read_no_exec_dir, tc)
+{
+	atf_tc_set_md_var(tc, "descr",
+	    "directory with read but no execute is traversed via "
+	    "FTS_DONTCHDIR fallback");
+	atf_tc_set_md_var(tc, "require.user", "unprivileged");
+}
+ATF_TC_BODY(read_no_exec_dir, tc)
+{
+	char *paths[] = { "dir", NULL };
+	FTS *fts;
+	FTSENT *ent;
+	bool saw_d, saw_file;
+
+	ATF_REQUIRE_EQ(0, mkdir("dir", 0755));
+	ATF_REQUIRE_EQ(0, close(creat("dir/file", 0644)));
+	ATF_REQUIRE_EQ(0, chmod("dir", 0400));
+
+	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL, NULL)) != NULL);
+
+	/*
+	 * Before the fix, zero entries were produced.  After the fix,
+	 * fts falls back to FTS_DONTCHDIR and traverses using full paths.
+	 * Verify the directory is not silently skipped.
+	 */
+	saw_d = false;
+	saw_file = false;
+	while ((ent = fts_read(fts)) != NULL) {
+		if (ent->fts_info == FTS_D &&
+		    strcmp(ent->fts_name, "dir") == 0)
+			saw_d = true;
+		if (strcmp(ent->fts_name, "file") == 0)
+			saw_file = true;
+	}
+
+	ATF_CHECK_MSG(saw_d,
+	    "FTS_D not returned for directory with mode 0400");
+	ATF_CHECK_MSG(saw_file,
+	    "file inside mode 0400 directory was not visited");
+
+	ATF_REQUIRE_EQ_MSG(0, fts_close(fts), "fts_close(): %m");
+}
+
+/*
+ * PR 196724: FTS_SLNONE must not be returned for a non-symlink.
+ *
+ * The fix ensures that FTS_SLNONE is only returned when lstat confirms
+ * the entry is actually a symlink.  Exercised by a time-bounded race
+ * where a background thread creates and deletes a regular file while
+ * fts traverses with FTS_LOGICAL.
+ */
+ATF_TC(no_slnone_for_nonsymlink);
+ATF_TC_HEAD(no_slnone_for_nonsymlink, tc)
+{
+	atf_tc_set_md_var(tc, "descr",
+	    "FTS_SLNONE must not be returned for a non-symlink");
+}
+ATF_TC_BODY(no_slnone_for_nonsymlink, tc)
+{
+	pthread_t thr;
+	char *paths[] = { "dir", NULL };
+	FTS *fts;
+	FTSENT *ent;
+	struct timespec start, now, elapsed;
+
+	ATF_REQUIRE_EQ(0, mkdir("dir", 0755));
+	ATF_REQUIRE_EQ(0, symlink("nonexistent", "dir/dead"));
+
+	race_stop = false;
+	ATF_REQUIRE_EQ(0, pthread_create(&thr, NULL, race_thrash,
+	    __DECONST(void *, "dir/victim")));
+
+	clock_gettime(CLOCK_MONOTONIC, &start);
+	for (;;) {
+		clock_gettime(CLOCK_MONOTONIC, &now);
+		timespecsub(&now, &start, &elapsed);
+		if (elapsed.tv_sec >= 1)
+			break;
+		fts = fts_open(paths, FTS_LOGICAL, NULL);
+		ATF_REQUIRE(fts != NULL);
+		while ((ent = fts_read(fts)) != NULL) {
+			if (ent->fts_info == FTS_SLNONE &&
+			    ent->fts_statp->st_mode != 0 &&
+			    !S_ISLNK(ent->fts_statp->st_mode))
+				ATF_CHECK_MSG(0,
+				    "FTS_SLNONE returned for non-symlink '%s'",
+				    ent->fts_name);
+		}
+		fts_close(fts);
+	}
+
+	race_stop = true;
+	pthread_join(thr, NULL);
+}
+
+/*
+ * PR 262038: fts_build() must detect readdir(2) errors and not treat
+ * them as end-of-directory.  The man page specifies that FTS_DNR must
+ * immediately follow FTS_D, in place of FTS_DP.
+ *
+ * Requires an unprivileged user because root ignores permissions.
+ */
+ATF_TC(readdir_error_detected);
+ATF_TC_HEAD(readdir_error_detected, tc)
+{
+	atf_tc_set_md_var(tc, "descr",
+	    "readdir errors produce FTS_DNR with fts_errno set");
+	atf_tc_set_md_var(tc, "require.user", "unprivileged");
+}
+ATF_TC_BODY(readdir_error_detected, tc)
+{
+	char *paths[] = { "dir", NULL };
+	FTS *fts;
+	FTSENT *ent;
+
+	ATF_REQUIRE_EQ(0, mkdir("dir", 0755));
+	ATF_REQUIRE_EQ(0, close(creat("dir/file", 0644)));
+
+	/*
+	 * Mode 0100: execute only, no read.  chdir() succeeds but
+	 * opendir/readdir fails.  fts must return FTS_D then FTS_DNR
+	 * (not FTS_DP) per the man page.
+	 */
+	ATF_REQUIRE_EQ(0, chmod("dir", 0100));
+
+	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL, NULL)) != NULL);
+
+	ATF_REQUIRE((ent = fts_read(fts)) != NULL);
+	ATF_CHECK_EQ_MSG(FTS_D, ent->fts_info,
+	    "expected FTS_D, got %d", ent->fts_info);
+
+	ATF_REQUIRE((ent = fts_read(fts)) != NULL);
+	ATF_CHECK_EQ_MSG(FTS_DNR, ent->fts_info,
+	    "expected FTS_DNR, got %d", ent->fts_info);
+	ATF_CHECK_MSG(ent->fts_errno != 0,
+	    "FTS_DNR must have non-zero fts_errno");
+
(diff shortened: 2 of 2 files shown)


lib/libc/tests/gen/Makefile
```diff
@@ -98,8 +99,9 @@ LIBADD.fpclassify_test+=m
 LIBADD.fpsetround_test+=m
 LIBADD.siginfo_test+=	m
 
-LIBADD.nice_test+=	pthread
-LIBADD.syslog_test+=	pthread
+LIBADD.nice_test+=	 pthread
+LIBADD.fts_regress_test+=pthread
+LIBADD.syslog_test+=	 pthread
```
asomers: What's going on with the nice_test and syslog_test lines?  Did you change some whitespace?

lib/libc/tests/gen/fts_regress_test.c
```diff
@@ -0,0 +1,339 @@
+/*
+ * Copyright (c) 2026 Jitendra Bhati
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+/*
+ * Regression tests for specific FreeBSD bug reports fixed in fts(3).
+ * Each test documents the original bug and verifies the fix is present.
```
asomers: Apart from line 8, the rest of this comment block should go away.  It's redundant with the comment block at the beginning of each test case, and it's harder to maintain when it's all the way up here.

lib/libc/tests/gen/fts_regress_test.c
```diff
@@ -0,0 +1,339 @@
...
+ *
+ * With mode 0400 (read only, no execute):
+ *  - fts can open the directory for readdir (read permission)
+ *  - chdir() fails (no execute permission)
+ *  - the fix: fall back to FTS_DONTCHDIR mode
+ *  - fts_accpath uses the full relative path "dir/file"
+ *
+ * Requires an unprivileged user because root ignores permissions.
+ */
+ATF_TC(pr45723_read_no_exec_dir);
```
asomers: I don't think you should put the old pr number in the test case name.  It will if anything just confuse an operator who is running the test.  You should simply call it `read_no_exec_dir`.

lib/libc/tests/gen/fts_regress_test.c
```diff
@@ -0,0 +1,339 @@
...
+	}
+	return (NULL);
+}
+
+ATF_TC(pr196724_slnone_race);
+ATF_TC_HEAD(pr196724_slnone_race, tc)
+{
+	atf_tc_set_md_var(tc, "descr",
+	    "PR 196724: FTS_SLNONE must not be returned for a non-symlink "
+	    "(race test, 1 second)");
```
asomers: IMHO this description is too verbose.  If I were writing it, I would leave out the second line.

lib/libc/tests/gen/fts_regress_test.c
```diff
@@ -0,0 +1,339 @@
...
+	FTSENT *ent;
+	time_t end;
+
+	ATF_REQUIRE_EQ(0, mkdir("dir", 0755));
+	ATF_REQUIRE_EQ(0, symlink("nonexistent", "dir/dead"));
+
+	pr196724_stop = 0;
+	ATF_REQUIRE_EQ(0, pthread_create(&thr, NULL, pr196724_thrash, NULL));
+
+	end = time(NULL) + 1;
```
asomers: time(3) doesn't have enough resolution here.  As written, the test could run for anywhere in the range (0, 2) seconds.  Worse, it could run for a very long time if the system clock gets adjusted while the test is ongoing.
Instead, you should use `clock_gettime(CLOCK_MONOTONIC, ...)`

lib/libc/tests/gen/fts_regress_test.c
```diff
@@ -0,0 +1,339 @@
...
+	 * Mode 0100: execute only, no read.  chdir() succeeds but
+	 * opendir/readdir fails.
+	 */
+	ATF_REQUIRE_EQ(0, chmod("dir", 0100));
+
+	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL, NULL)) != NULL);
+
+	saw_dnr = 0;
+	while ((ent = fts_read(fts)) != NULL) {
+		if (ent->fts_info == FTS_DNR) {
```
asomers: The man page is more specific than this.  It says that FTS_DNR should immediately follow FTS_D, in place of FTS_DP.  Could you tighten up the assertions in this test case to match the man page?

lib/libc/tests/gen/fts_regress_test.c
```diff
@@ -0,0 +1,339 @@
...
+ * non-directory cannot be substituted between stat and open.  We verify
+ * normal traversal still completes correctly after the fix.
+ */
+ATF_TC(svnr246641_odirectory_changedir);
+ATF_TC_HEAD(svnr246641_odirectory_changedir, tc)
+{
+	atf_tc_set_md_var(tc, "descr",
+	    "SVN r246641: normal traversal works with O_DIRECTORY fix");
+}
+ATF_TC_BODY(svnr246641_odirectory_changedir, tc)
```
asomers: What does this test accomplish?  How does it ensure that O_DIRECTORY is used?

lib/libc/tests/gen/fts_regress_test.c
```diff
@@ -0,0 +1,339 @@
...
+ * concurrently modified during traversal.
+ *
+ * The test exercises the race for one second with a background thread
+ * rapidly creating and deleting a file.  The test passes if fts does
+ * not crash or hang during that time.
+ */
+static volatile int svnr261589_stop;
+
+static void *
+svnr261589_thrash(void *arg)
```
asomers: You can actually reuse the same thrash function for both test cases.  ATF will ensure that two test cases are not running at the same time in the same process.

asomers (changes requested):

lib/libc/tests/gen/fts_regress_test.c
Jeetubhati: It doesn't prove O_DIRECTORY is used, it only checks normal traversal works
Should I remove it entirely, or is there a way to test the TOCTOU
defence that you have in mind?

asomers: Can you test it with a 1-second thrash loop like the others?
The other possibility I can think of would be to use dtrace to validate that the correct flag is passed to open(2).  But I think I would prefer a thrash loop.

lib/libc/tests/gen/fts_regress_test.c
Jeetubhati: Removed. Each test case has its own comment now.

lib/libc/tests/gen/fts_regress_test.c
Jeetubhati: Done.

lib/libc/tests/gen/fts_regress_test.c
Jeetubhati: Shortened all descriptions in the file.

lib/libc/tests/gen/fts_regress_test.c
Jeetubhati: Fixed. Using clock_gettime(CLOCK_MONOTONIC) now.

lib/libc/tests/gen/fts_regress_test.c
Jeetubhati: I updated it to verify FTS_DNR immediately follows FTS_D with no
FTS_DP in between, matching the man page.

lib/libc/tests/gen/fts_regress_test.c
Jeetubhati: I replaced it with a 1-second thrash loop that races to substitute
dir/sub with a regular file. With O_DIRECTORY the open fails
safely; without it fts would misbehave.

lib/libc/tests/gen/fts_regress_test.c
Jeetubhati: Ok. merged into a single race_thrash function with the path
passed as an argument. Both race tests now use it.

lib/libc/tests/gen/Makefile
Jeetubhati: Sorry, I accidentally changed whitespace on those two
lines while editing. Now i restored them to their original state.

lib/libc/tests/gen/fts_regress_test.c
asomers: BTW, attempting to measure a duration of time using the wall clock is a classic beginner mistake.  It's basically impossible, because the wall clock can be changed at any time by the user or by ntpd.  That's the main reason that clock_gettime exists.

lib/libc/tests/gen/fts_regress_test.c
```diff
@@ -0,0 +1,309 @@
...
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+
+#include <atf-c.h>
+
+#include "fts_test.h"
```
asomers: I don't think this header file is needed anymore.

lib/libc/tests/gen/fts_regress_test.c
```diff
@@ -48,123 +25,114 @@
...
+	race_stop = 0;
+	ATF_REQUIRE_EQ(0, pthread_create(&thr, NULL, race_thrash,
+	    __DECONST(void *, "dir/victim")));
 
-	end = time(NULL) + 1;
-	while (time(NULL) < end) {
-		fts = fts_open(paths, FTS_LOGICAL, fts_lexical_compar);
+	clock_gettime(CLOCK_MONOTONIC, &start);
+	while (clock_gettime(CLOCK_MONOTONIC, &now) == 0 &&
+	    now.tv_sec - start.tv_sec < 1) {
```
asomers: For comparing the two timestamps, you should use `timespecsub`

lib/libc/tests/gen/fts_regress_test.c
Jeetubhati: Understood, thanks for explaining. Is there anything else that
needs updating in the current implementation, or is CLOCK_MONOTONIC
sufficient as-is?

lib/libc/tests/gen/fts_regress_test.c
asomers: For tests like this, I wonder if the while loop isn't more trouble than it's worth.  In this case, the return sequence from fts_read should be fully specified.  There is no ambiguity about which file to return first.  So I would think that you should be able to remove the while loop and simply do something like
```
fts_open()
assert fts_read() returns FTS_D
assert fts_read() returns FTS_DNR
assert fts_read() returns NULL
```

lib/libc/tests/gen/fts_regress_test.c
```diff
@@ -48,123 +25,114 @@
...
-	 * returned (i.e. the directory is not silently skipped).
+	 * Before the fix, zero entries were produced.  After the fix,
+	 * fts falls back to FTS_DONTCHDIR and traverses using full paths.
+	 * Verify the directory is not silently skipped.
 	 */
 	saw_d = 0;
-	saw_ns = 0;
+	saw_file = 0;
 	entries = 0;
 	while ((ent = fts_read(fts)) != NULL) {
```
asomers, line 98: Just like I mention below, you can probably eliminate the while loop here.  It might be more readable.

lib/libc/tests/gen/fts_regress_test.c
```diff
@@ -48,123 +25,114 @@
...
 	atf_tc_set_md_var(tc, "require.user", "unprivileged");
 }
-ATF_TC_BODY(pr45723_read_no_exec_dir, tc)
+ATF_TC_BODY(read_no_exec_dir, tc)
 {
 	char *paths[] = { "dir", NULL };
 	FTS *fts;
 	FTSENT *ent;
-	int saw_d, saw_ns, entries;
+	int saw_d, saw_file, entries;
```
asomers: If you don't unroll the while loop, flag variables like this ought to be `bool`, as a matter of style.

lib/libc/tests/gen/fts_regress_test.c
```diff
@@ -216,124 +182,128 @@ ATF_TC_BODY(pr262038_readdir_error_detected, tc)
...
+	 * fails safely.  Without it fts would open the file as a
+	 * directory and misbehave.
+	 */
+	race_stop = 0;
+	ATF_REQUIRE_EQ(0, pthread_create(&thr, NULL, race_thrash,
+	    __DECONST(void *, "dir/sub")));
+
+	clock_gettime(CLOCK_MONOTONIC, &start);
+	while (clock_gettime(CLOCK_MONOTONIC, &now) == 0 &&
+	    now.tv_sec - start.tv_sec < 1) {
```
asomers: Also use `timespecsub` here.

lib/libc/tests/gen/fts_regress_test.c
```diff
@@ -216,124 +182,128 @@ ATF_TC_BODY(pr262038_readdir_error_detected, tc)
...
-	ATF_CHECK_EQ_MSG(5, entries,
-	    "expected 5 entries, got %d", entries);
+	/*
+	 * Background thread races to replace dir/sub with a regular
+	 * file between fts stat and open.  With O_DIRECTORY the open
+	 * fails safely.  Without it fts would open the file as a
+	 * directory and misbehave.
+	 */
+	race_stop = 0;
+	ATF_REQUIRE_EQ(0, pthread_create(&thr, NULL, race_thrash,
```
asomers: There are two problems with the thrash loop here:

1. The pthread uses `unlink` on a directory.  That won't work.  It needs to use either `rmdir` or `unlinkat` with `AT_REMOVEDIR`.
2. After the thrash loop removes the directory the first time, nothing recreates it.  Instead, the thrash loop will recreate it as a regular file.  You need to create is as a directory

lib/libc/tests/gen/fts_regress_test.c
```diff
@@ -216,124 +182,128 @@ ATF_TC_BODY(pr262038_readdir_error_detected, tc)
...
+	    __DECONST(void *, "dir/victim")));
 
-	end = time(NULL) + 1;
-	while (time(NULL) < end) {
-		fts = fts_open(paths, FTS_PHYSICAL, fts_lexical_compar);
+	clock_gettime(CLOCK_MONOTONIC, &start);
+	while (clock_gettime(CLOCK_MONOTONIC, &now) == 0 &&
+	    now.tv_sec - start.tv_sec < 1) {
```
asomers: And here as well, use `timespecsub`.

lib/libc/tests/gen/fts_regress_test.c
```diff
@@ -216,124 +182,128 @@ ATF_TC_BODY(pr262038_readdir_error_detected, tc)
...
+		fts = fts_open(paths, FTS_PHYSICAL, NULL);
 		if (fts == NULL)
 			continue;
```
asomers: Why would `fts_open` legitimately return NULL here?

asomers (changes requested):

lib/libc/tests/gen/fts_regress_test.c
asomers: Just the way that you compare timestamps.  You should use timespecsub instead of manually comparing the seconds fields.

lib/libc/tests/gen/fts_regress_test.c
Jeetubhati: Removed it.

lib/libc/tests/gen/fts_regress_test.c
Jeetubhati: Fixed. Using timespecsub in all three race tests now.

lib/libc/tests/gen/fts_regress_test.c
Jeetubhati: Changed flags to bool. I Kept the loop since the exact sequence
depends on fts_build behavior with mode 0400 and we only need
to verify the directory is not silently skipped.

lib/libc/tests/gen/fts_regress_test.c
Jeetubhati: Fixed both using rmdir/mkdir in the thrash function so dir/sub
is always recreated as a directory, not a regular file.

lib/libc/tests/gen/fts_regress_test.c
Jeetubhati: I added it as a defensive check thinking fts_open could fail if
the directory was temporarily absent during the thrash. But you
are right dir itself is never removed, only dir/sub is, so
fts_open has no reason to fail. I removed it in latest changes.

lib/libc/tests/gen/fts_regress_test.c
Jeetubhati: Replaced with sequential fts_read() calls as you suggested.
Now it looks much clean.

lib/libc/tests/gen/fts_regress_test.c
```diff
@@ -46,127 +24,130 @@
 
 #include <atf-c.h>
 
-#include "fts_test.h"
+/*
+ * Thrash function for file-based race tests: repeatedly creates and
+ * deletes a regular file at the given path.
+ */
+static volatile int race_stop;
```
asomers: Why int instead of bool?

lib/libc/tests/gen/fts_regress_test.c
asomers: Ok, that makes sense.

lib/libc/tests/gen/fts_regress_test.c
```diff
@@ -169,85 +182,78 @@ ATF_TC_BODY(readdir_error_detected, tc)
...
-		if (prev_info == FTS_D && ent->fts_info == FTS_DP)
-			ATF_CHECK_MSG(0,
-			    "FTS_DP returned for unreadable dir, "
-			    "expected FTS_DNR");
-		prev_info = ent->fts_info;
-	}
-	ATF_CHECK_EQ_MSG(FTS_DNR, prev_info,
-	    "expected last entry to be FTS_DNR, got %d", prev_info);
+	ent = fts_read(fts);
+	ATF_REQUIRE(ent != NULL);
```
asomers: You can condense these statements into one line, like this:
```
ATF_REQUIRE((ent = fts_read(fts)) != NULL);
```

asomers (changes requested):
The thrash loops look much better.  I just have a couple of more comments about style.

lib/libc/tests/gen/fts_regress_test.c
Jeetubhati: Replaced it.

lib/libc/tests/gen/fts_regress_test.c
Jeetubhati: Changed to bool.

PR #2258, opened 2026-06-03
lib/libc/tests/gen: address post-merge feedback on fts_set_test.c

Jeetubhati:
Per review by dag-erling:

- Remove fts_check_debug() which has no effect in fts_set_test.c
- Add errno = 0 before fts_read() loop so the errno check is meaningful
- Remove fts_lexical_compar where traversal order does not matter
- Change int boolean flags to bool throughout
- Move fts_set_clientptr/fts_get_clientptr test inside fts_read() loop so the pointer is exercised during active traversal

Fixes:		https://github.com/freebsd/freebsd-src/commit/940142d6103746bb1158c408da443d0240b836d9 ("lib/libc/tests/gen: add fts_set() tests")

### DIFF
--- a/lib/libc/tests/gen/fts_set_test.c
+++ b/lib/libc/tests/gen/fts_set_test.c
@@ -68,10 +68,10 @@ ATF_TC_BODY(again, tc)
 	ATF_REQUIRE_EQ(0, mkdir("dir", 0755));
 	ATF_REQUIRE_EQ(0, close(creat("dir/file", 0644)));
 
-	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL,
-	    fts_lexical_compar)) != NULL);
+	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL, NULL)) != NULL);
 
 	revisit_count = 0;
+	errno = 0;
 	while ((ent = fts_read(fts)) != NULL) {
 		if (ent->fts_info == FTS_F && revisit_count == 0) {
 			ATF_REQUIRE_EQ_MSG(0,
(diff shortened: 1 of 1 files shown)


lib/libc/tests/gen/fts_set_test.c
```diff
@@ -346,7 +357,6 @@ ATF_TC_BODY(get_stream_backpointer, tc)
 
 ATF_TP_ADD_TCS(tp)
 {
-	fts_check_debug();
```
asomers, line 349: If you're going to remove it here, then you should probably remove it from fts_children_test.c, too.

asomers (changes requested):
I don't see any problems, but we should let des review it too.

Also, you should add a `Fixes: 940142d6103 ("lib/libc/tests/gen: add fts_set() tests")` tag to the commit message.  That way the MFC reminder will remind you (or, more likely, me) to MFC these two at the same time.

lib/libc/tests/gen/fts_set_test.c
Jeetubhati: Okay

Jeetubhati: Removed in a separate PR since fts_children_test is already merged:
https://github.com/freebsd/freebsd-src/pull/2260

lib/libc/tests/gen/fts_set_test.c
```diff
@@ -68,10 +68,10 @@ ATF_TC_BODY(again, tc)
 	ATF_REQUIRE_EQ(0, mkdir("dir", 0755));
 	ATF_REQUIRE_EQ(0, close(creat("dir/file", 0644)));
 
-	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL,
-	    fts_lexical_compar)) != NULL);
+	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL, NULL)) != NULL);
 
 	revisit_count = 0;
+	errno = 0;
```
dag-erling: You need to clear `errno` before _each_ `fts_read()` call.

lib/libc/tests/gen/fts_set_test.c
dag-erling: I believe this is now unused.

dag-erling (changes requested):

lib/libc/tests/gen/fts_set_test.c
Jeetubhati: I fixed it using the for loop idiom: for (errno = 0; (ent = fts_read(fts)) != NULL; errno = 0)

lib/libc/tests/gen/fts_set_test.c
Jeetubhati: Removed.

dag-erling (approved):

asomers (approved):

asomers:
Closed by ee213339f4b21782cd1b44086ff9b7fe1fd682c5 .

PR #2259, opened 2026-06-04
nat64: Fix type confusion panic when using wrong NAT64 instance types

TeddyEngel:
When an ipfw rule references a NAT64 instance by name, the kernel looks up the instance in the shared srvstate[] array without verifying the instance type. If the named instance is not the expected one, the code casts the instance to the wrong type and will eventually cause a kernel panic.

The root cause is that all NAT64 instance types share the same srvstate[] array but have different struct layouts. 

Fixed by adding a type check after NAT64_LOOKUP() to verify that the instance's etlv matches the expected type before proceeding. If the type doesn't match, return IP_FW_DENY to reject the packet safely rather than crashing.

Tested with the reproducers from the bug report on FreeBSD 16.0-CURRENT, no kernel panic with the changes.

PR: [295794](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=295794)


spmzt (approved):

spmzt:

PR #2260, opened 2026-06-04
fts: address post-merge feedback on fts_children_test.c

Jeetubhati:
- Remove fts_check_debug() which is only needed by test cases
  that use the fts_test() helper; fts_children_test.c does not
  call fts_test()
- Remove fts_lexical_compar where traversal order does not matter

Fixes:		e624417db8a1 ("lib/libc/tests/gen: add fts_children() tests")

### DIFF
--- a/lib/libc/tests/gen/fts_children_test.c
+++ b/lib/libc/tests/gen/fts_children_test.c
@@ -21,7 +21,11 @@
 
 #include <atf-c.h>
 
-#include "fts_test.h"
+static int
+fts_lexical_compar(const FTSENT * const *a, const FTSENT * const *b)
+{
+	return (strcmp((*a)->fts_name, (*b)->fts_name));
+}
 
 /*
  * fts_children() before fts_read() returns the list of root entries.
@@ -164,8 +168,7 @@ ATF_TC_BODY(called_twice, tc)
 	ATF_REQUIRE_EQ(0, close(creat("dir/x", 0644)));
 	ATF_REQUIRE_EQ(0, close(creat("dir/y", 0644)));
 
-	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL,
-	    fts_lexical_compar)) != NULL);
+	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL, NULL)) != NULL);
 
 	ent = fts_read(fts);
 	ATF_REQUIRE(ent != NULL);
@@ -218,8 +221,7 @@ ATF_TC_BODY(nameonly, tc)
 	ATF_REQUIRE_EQ(0, close(creat("dir/f1", 0644)));
 	ATF_REQUIRE_EQ(0, close(creat("dir/f2", 0644)));
 
-	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL,
-	    fts_lexical_compar)) != NULL);
+	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL, NULL)) != NULL);
 
 	ent = fts_read(fts);
 	ATF_REQUIRE(ent != NULL);
@@ -266,8 +268,7 @@ ATF_TC_BODY(nondirectory, tc)
 	ATF_REQUIRE_EQ(0, mkdir("dir", 0755));
 	ATF_REQUIRE_EQ(0, close(creat("dir/file", 0644)));
 
-	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL,
-	    fts_lexical_compar)) != NULL);
+	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL, NULL)) != NULL);
 
 	ent = fts_read(fts);	/* FTS_D dir */
 	ATF_REQUIRE(ent != NULL);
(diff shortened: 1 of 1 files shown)


dag-erling:
Do any of your tests actually rely on the order the results are returned in? If not, you should pass `NULL` instead of `fts_lexical_compar` and drop `fts_test.h`. If yes, you should probably just copy `fts_lexical_compar()` from `fts_test.h` instead of including it. It's not just “useful stuff for FTS tests”, it is code shared by `fts_misc_test`, `fts_open_test`, and `fts_options_test`.

Can you also please just mark your commits with `fts:` and not `lib/libc/tests/gen:`, the latter is not useful at all.

dag-erling:
> fts_check_debug() has no effect in test programs.

This is not true, and frankly it worries me that you wrote it. Do you actually understand the code you are submitting?

asomers (changes requested):
The change is good, but you should edit the commit message:

- The "Fixes" tag must reference a commit hash, not a PR number.  There are automated tools that expect that.
- "has no effect in test programs" isn't quite accurate.  "Is only needed by test cases that use the `fts_test` helper" is better.

lib/libc/tests/gen/fts_children_test.c
```diff
@@ -165,7 +165,7 @@ ATF_TC_BODY(called_twice, tc)
 	ATF_REQUIRE_EQ(0, close(creat("dir/y", 0644)));
 
 	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL,
-	    fts_lexical_compar)) != NULL);
+	    NULL)) != NULL);
```
dag-erling: pull up

lib/libc/tests/gen/fts_children_test.c
```diff
@@ -219,7 +219,7 @@ ATF_TC_BODY(nameonly, tc)
 	ATF_REQUIRE_EQ(0, close(creat("dir/f2", 0644)));
 
 	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL,
-	    fts_lexical_compar)) != NULL);
+	    NULL)) != NULL);
```
dag-erling: pull up

lib/libc/tests/gen/fts_children_test.c
```diff
@@ -267,7 +267,7 @@ ATF_TC_BODY(nondirectory, tc)
 	ATF_REQUIRE_EQ(0, close(creat("dir/file", 0644)));
 
 	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL,
-	    fts_lexical_compar)) != NULL);
+	    NULL)) != NULL);
```
dag-erling: pull up

lib/libc/tests/gen/fts_children_test.c
dag-erling: Drop

dag-erling (changes requested):

Jeetubhati:
> Do any of your tests actually rely on the order the results are returned in? If not, you should pass `NULL` instead of `fts_lexical_compar` and drop `fts_test.h`. If yes, you should probably just copy `fts_lexical_compar()` from `fts_test.h` instead of including it. It's not just “useful stuff for FTS tests”, it is code shared by `fts_misc_test`, `fts_open_test`, and `fts_options_test`.
> 
> Can you also please just mark your commits with `fts:` and not `lib/libc/tests/gen:`, the latter is not useful at all.

Checked each test case. nonempty_dir checks entries in alphabetical
order so it needs fts_lexical_compar. The other three only count
entries or check names without relying on order, so I removed
fts_lexical_compar from those. fts_test.h stays since fts_lexical_compar
is still used in nonempty_dir.

Jeetubhati:
> This is not true, and frankly it worries me that you wrote it. Do you actually understand the code you are submitting?

You are right to question that, fts_check_debug() only sets up debug logs for fts_test(). Since fts_children_test.c never uses fts_test(), we do not need fts_check_debug() here at all. I fixed the commit message to explain this clearly.

Jeetubhati:
> Can you also please just mark your commits with `fts:` and not `lib/libc/tests/gen:`, the latter is not useful at all.

Noted. Using fts: going forward.

Jeetubhati:
> * The "Fixes" tag must reference a commit hash, not a PR number.  There are automated tools that expect that.

Fixed to reference the commit hash e624417db8a1.

Jeetubhati:
> * "has no effect in test programs" isn't quite accurate.  "Is only needed by test cases that use the `fts_test` helper" is better.

Updated the commit

lib/libc/tests/gen/fts_children_test.c
Jeetubhati: Done

lib/libc/tests/gen/fts_children_test.c
Jeetubhati: Done

lib/libc/tests/gen/fts_children_test.c
Jeetubhati: Done

lib/libc/tests/gen/fts_children_test.c
Jeetubhati: Dropped fts_test.h and copied fts_lexical_compar locally.

asomers (approved):

dag-erling (approved):

dag-erling:
Thank you. You don't need to reply to every single review comment, you know. Updating the patch is enough in most cases.

PR #2261, opened 2026-06-04
Add support for Linux-compatible Unix abstract sockets

ricardobranco777:
From the Linux's [unix(7) manpage](https://man7.org/linux/man-pages/man7/unix.7.html):

> an abstract socket address is distinguished (from a  pathname socket) by the fact that sun_path[0] is a null byte ('\0').  The socket's address in this namespace is given by the additional bytes in sun_path that are covered by the specified length of the address structure.  (Null bytes in the name have no special significance.)  The name has no connection with filesystem pathnames.  (...)

Also:

> Socket permissions have no meaning for abstract sockets: the process umask(2) has no effect when binding an abstract socket, and changing the ownership and permissions of the object (via fchown(2) and fchmod(2)) has no effect on the accessibility of the socket. Abstract sockets automatically disappear when all open references to the socket are closed.

This PR adds support for Linux-compatible Unix abstract sockets.

It passes some tests but before I continue feel free to dismiss this as a Linuxism now.  ;)

TODO
- [x] Add tests. 
- [ ] Add introspection via sysctls?
- [ ] Update commands such as fstat(1).  Please suggest others.
- [ ] Update manpages.
- [ ] Update the Linuxulator (in another PR).

### DIFF
--- a/sys/kern/uipc_usrreq.c
+++ b/sys/kern/uipc_usrreq.c
@@ -65,8 +65,10 @@
 #include <sys/fcntl.h>
 #include <sys/file.h>
 #include <sys/filedesc.h>
+#include <sys/hash.h>
 #include <sys/jail.h>
 #include <sys/kernel.h>
+#include <sys/libkern.h>
 #include <sys/lock.h>
 #include <sys/malloc.h>
 #include <sys/mbuf.h>
@@ -77,6 +79,7 @@
 #include <sys/proc.h>
 #include <sys/protosw.h>
 #include <sys/queue.h>
+#include <sys/refcount.h>
 #include <sys/resourcevar.h>
 #include <sys/rwlock.h>
 #include <sys/socket.h>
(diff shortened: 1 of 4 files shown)


sys/kern/uipc_usrreq.c
```diff
@@ -585,6 +820,11 @@ uipc_bindat(int fd, struct socket *so, struct sockaddr *nam, struct thread *td)
 	if (namelen <= 0)
 		return (EINVAL);
 
+	if (soun->sun_path[0] == '\0')
+		return (namelen == 1 ?
+		    unp_bind_abstract_auto(so, td) :
+		    unp_bind_abstract(so, nam, namelen, td));
+
```
bms, line 828: Perhaps another explicit conditional is preferable to the ternary, given this has an if() scope already?

bms:
Yeah I've been looking around in this area for related things. There is some value in the abstract namespace being native rather in the Linuxolator, though it may not directly prevent namespace race conditions; that's a job for bindat() etc.

The lack of POSIX permissions probably mandates more client use of LOCAL_PEERCRED etc. (we still carry around an ONC/SunRPC wrapper for this in libc).

sys/kern/uipc_usrreq.c
```diff
@@ -559,6 +625,175 @@ uipc_attach(struct socket *so, int proto, struct thread *td)
 	return (0);
 }
 
+/*
+ * Look up an abstract binding in pr's namespace.  On hit, returns true with
```
kevans91, line 629: I think the semantics w.r.t. jails are fine here for now, but I wonder if we should consider adding a new jail param that gives us the option for a jail to inherit the prison's abstract namespace (think a dual state like `vnet`; `unixns=(inherit|new)`).  I think that implies that we'd want a different organization entirely to better keep track of which prison has access to which namespace (and that we'd maybe just attach the `unp_absent` to the prison instead, so maybe it's worth answering the question up front: would that be useful behavior?  If Linux allows copying or sharing abstract namespaces, that might give us some kind of answer here.

kevans91: Actually, https://man7.org/linux/man-pages/man7/network_namespaces.7.html would seem to imply that the abstract namespace is tied to the netns. VNET()'ifying this would probably make a lot of sense, and then we don't need another jail param.

kevans91:
re: the usefulness of this, I would like to understand if you have a use-case in mind or if this is just closing the feature gap.  I've considered this in the past for things, but always settled on other ways to accomplish what I was doing.

ricardobranco777:
> Yeah I've been looking around in this area for related things. There is some value in the abstract namespace being native rather in the Linuxolator, though it may not directly prevent namespace race conditions; that's a job for bindat() etc.

It's not currently supported in the Linuxulator:
https://github.com/freebsd/freebsd-src/blob/main/sys/compat/linux/linux.c#L403

sys/kern/uipc_usrreq.c
ricardobranco777: The abstract namespace is tied to the network namespace, yes, and VNET is the best fit.  I can drop the prison logic then.

kevans91: Sounds good to me- we'll just want to be sure to document that well in the unix(4) addition that covers the feature.  I would almost predict that someone will complain and want some sysctls to enable greater isolation of the abstract namespace within the same vnet- it might be worth stashing off the originator's ucred, but that could be punted until later.  I could perhaps also see wanting some MAC integration, which might also want a ucred attached to the socket.

ricardobranco777:
> re: the usefulness of this, I would like to understand if you have a use-case in mind or if this is just closing the feature gap. I've considered this in the past for things, but always settled on other ways to accomplish what I was doing.

Just closing the feature gap as a nice to have.  It's not a killer feature.  Other than that:
- Support more Linux applications.
- Some patches in ports adding `#ifdef __linux__` for abstract sockets in [mutter](https://github.com/freebsd/freebsd-ports/blob/main/x11-wm/mutter/files/patch-src_wayland_meta-wayland-buffer.c)

kevans91:
> > re: the usefulness of this, I would like to understand if you have a use-case in mind or if this is just closing the feature gap. I've considered this in the past for things, but always settled on other ways to accomplish what I was doing.
> 
> Just closing the feature gap as a nice to have. It's not a killer feature. Other than that:
> 
> * Support more Linux applications.
> * Some patches in ports adding `#ifdef __linux__` for abstract sockets in [mutter](https://github.com/freebsd/freebsd-ports/blob/main/x11-wm/mutter/files/patch-src_wayland_meta-wayland-buffer.c)

I don't see a reason we would reject the feature, personally.  @glebius might have a stronger opinion here, but I suspect that it's fine if we nail the implementation details down just right.

sys/kern/uipc_usrreq.c
kevans91: After discussing it with other folks, there's some concern we'd end up in the same spot as sysvipc if it's shared even within a vnet, so it'd probably be better to do both per-prison and per-VNET namespaces and add that jail parameter anyways.  We'd effectively proceed with VNET()'ing this and stashing a `unp_abshead *pr_unixns` off into `struct prison`.  With these params:

 - vnet=new: the prison just takes `V_unp_absent`, unixns=inherit is an error
 - vnet=inherit, unixns=new: we allocate a new unp_abshead for the prison
 - vnet=inherit, unixns=inherit: we use the parent pr_unixns

Thus, inheriting the namespace across vnets is a non-starter.

glebius:
I noticed the pull request. Sorry for not reviewing promptly. I'll try to find time, but it could be that won't find any until BSDCan (next week).

sys/kern/uipc_usrreq.c
bapt: I think we should keep it per prison and never allow inheritance. I plan to use this in rcd as it solves a huge problem I have; if this socket is inherited in a jail it would be a problem.

PR #2181, opened 2026-05-07
compat/linux: return EBADF for O_PATH xattr operations

ilovegrape:
This fixes LTP open13, which expects O_PATH mmap() and fgetxattr()
to fail with EBADF, but FreeBSD returned EACCES for mmap() and
EOPNOTSUPP for fgetxattr().

Apply the same EBADF handling to fsetxattr(), fremovexattr(), and
flistxattr() so the xattr paths stay consistent.

PR #2263, opened 2026-06-05
linuxulator: improve getxattr() error handling

ilovegrape:
- Return ERANGE for short getxattr buffers: when the caller provides a buffer smaller than the xattr value, linux_getxattr() and linux_lgetxattr() should set the return value to the full size of the attribute and return ERANGE, matching Linux behavior. This fixes the LTP getxattr01 test.
- Return ENODATA for getxattr on unsupported file types: for file types that do not support extended attributes (fifo, char device, block device, UNIX domain socket), the FreeBSD VFS layer returns EOPNOTSUPP via vop_eopnotsupp(). Map this to ENOATTR so that the Linux errno translation produces ENODATA (61), matching Linux behavior. This fixes the LTP getxattr02 test.

### DIFF
--- a/sys/compat/linux/linux_xattr.c
+++ b/sys/compat/linux/linux_xattr.c
@@ -26,8 +26,10 @@
  */
 
 #include <sys/param.h>
+#include <sys/capsicum.h>
 #include <sys/extattr.h>
 #include <sys/fcntl.h>
+#include <sys/file.h>
 #include <sys/namei.h>
 #include <sys/proc.h>
 #include <sys/syscallsubr.h>
@@ -132,11 +134,20 @@ listxattr(struct thread *td, struct listxattr_args *args)
 {
 	char attrname[LINUX_XATTR_NAME_MAX + 1];
 	char *data, *prefix, *key;
+	cap_rights_t rights;
+	struct file *fp = NULL;
 	struct uio auio;
 	struct iovec aiov;
 	unsigned char keylen;
 	size_t sz, cnt, rs, prefixlen, pairlen;
 	int attrnamespace, error;
+
+	if (args->path == NULL) {
+		error = getvnode(td, args->fd,
+		    cap_rights_init_one(&rights, CAP_EXTATTR_LIST), &fp);
+		if (error != 0)
+			return (error);
+	}
 
 	if (args->size != 0)
 		sz = min(LINUX_XATTR_LIST_MAX, args->size);
@@ -162,7 +173,7 @@ listxattr(struct thread *td, struct listxattr_args *args)
 			error = kern_extattr_list_path(td, args->path,
 			    attrnamespace, &auio, args->follow, UIO_USERSPACE);
 		else
-			error = kern_extattr_list_fd(td, args->fd,
+			error = kern_extattr_list_fp(td, fp,
 			    attrnamespace, &auio);
 		rs = sz - auio.uio_resid;
 		if (error == EPERM)
@@ -204,6 +215,8 @@ listxattr(struct thread *td, struct listxattr_args *args)
 	if (error == 0)
 		td->td_retval[0] = cnt;
 	free(data, M_LINUX);
+	if (fp != NULL)
+		fdrop(fp, td);
 	return (error_to_xattrerror(attrnamespace, error));
 }
 
@@ -253,18 +266,33 @@ static int
 removexattr(struct thread *td, struct removexattr_args *args)
 {
 	char attrname[LINUX_XATTR_NAME_MAX + 1];
+	struct file *fp = NULL;
+	cap_rights_t rights;
 	int attrnamespace, error;
+
+	if (args->path == NULL) {
+		error = getvnode(td, args->fd,
+		    cap_rights_init_one(&rights, CAP_EXTATTR_DELETE), &fp);
+		if (error != 0)
+			return (error);
+	}
 
 	error = xattr_to_extattr(args->name, &attrnamespace, attrname);
 	if (error != 0)
-		return (error);
+		goto out_err;
 	if (args->path != NULL)
 		error = kern_extattr_delete_path(td, args->path, attrnamespace,
 		    attrname, args->follow, UIO_USERSPACE);
 	else
-		error = kern_extattr_delete_fd(td, args->fd, attrnamespace,
+		error = kern_extattr_delete_fp(td, fp, attrnamespace,
 		    attrname);
+	if (fp != NULL)
+		fdrop(fp, td);
 	return (error_to_xattrerror(attrnamespace, error));
+out_err:
+	if (fp != NULL)
+		fdrop(fp, td);
+	return (error);
 }
 
 int
@@ -310,17 +338,30 @@ static int
 getxattr(struct thread *td, struct getxattr_args *args)
 {
 	char attrname[LINUX_XATTR_NAME_MAX + 1];
+	struct file *fp = NULL;
+	cap_rights_t rights;
 	int attrnamespace, error;
+
+	if (args->path == NULL) {
+		error = getvnode(td, args->fd,
+		    cap_rights_init_one(&rights, CAP_EXTATTR_GET), &fp);
+		if (error != 0)
+			return (error);
+	}
 
 	error = xattr_to_extattr(args->name, &attrnamespace, attrname);
 	if (error != 0)
-		return (error);
+		goto out_err;
 	if (args->path != NULL)
 		error = kern_extattr_get_path(td, args->path, attrnamespace,
 		    attrname, args->value, args->size, args->follow, UIO_USERSPACE);
 	else
-		error = kern_extattr_get_fd(td, args->fd, attrnamespace,
+		error = kern_extattr_get_fp(td, fp, attrnamespace,
 		    attrname, args->value, args->size);
+
+out_err:
+	if (fp != NULL)
+		fdrop(fp, td);
 	return (error == EPERM ? ENOATTR : error);
 }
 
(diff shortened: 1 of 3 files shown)


kostikbel:
Using td_retval is rude.  Change the signature of kern_extattr_get_path()/get_fd() to take a pointer to size.  Then set td_retval[0] in sys_extattr_get(), and use this in linuxolator.

ilovegrape:
> Using td_retval is rude. Change the signature of kern_extattr_get_path()/get_fd() to take a pointer to size. Then set td_retval[0] in sys_extattr_get(), and use this in linuxolator.

Thanks for the review, I didn't want to change too much at the beginning.
I just have the fix amended according to the comments and re-tested, the result is positive.

sys/kern/vfs_extattr.c
```diff
@@ -489,19 +492,22 @@ user_extattr_get_path(struct thread *td, const char *path, int attrnamespace,
...
 	int error;
 
 	error = copyinstr(uattrname, attrname, sizeof(attrname), NULL);
 	if (error)
 		return (error);
-	return (kern_extattr_get_path(td, path, attrnamespace,
-	    attrname, data, nbytes, follow, UIO_USERSPACE));
+	error = kern_extattr_get_path(td, path, attrnamespace,
+	    attrname, data, nbytes, follow, UIO_USERSPACE, &retval);
+	td->td_retval[0] = retval;
```
kostikbel:
```
if (error == 0)
   td->td_retval[0] = retval;
```

sys/kern/vfs_extattr.c
```diff
@@ -419,18 +419,21 @@ int
...
 	int error;
 
 	error = copyinstr(uap->attrname, attrname, sizeof(attrname), NULL);
 	if (error)
 		return (error);
-	return (kern_extattr_get_fd(td, uap->fd, uap->attrnamespace,
-	    attrname, uap->data, uap->nbytes));
+	error = kern_extattr_get_fd(td, uap->fd, uap->attrnamespace,
+	    attrname, uap->data, uap->nbytes, &retval);
+	td->td_retval[0] = retval;
```
kostikbel:
```
if (error == 0)
   td->td_retval[0] = retval;
```

sys/compat/linux/linux_xattr.c
```diff
@@ -310,18 +310,35 @@ static int
...
+		return (ERANGE);
+	if (args->path != NULL)
+		error = kern_extattr_get_path(td, args->path, attrnamespace,
+		    attrname, args->value, args->size, args->follow,
+		    UIO_USERSPACE, &size);
+	else
+		error = kern_extattr_get_fd(td, args->fd, attrnamespace,
+		    attrname, args->value, args->size, &size);
+
+	td->td_retval[0] = size;
```
kostikbel:
```
if (error == 0)
   td->td_retval[0] = retval;
```

ilovegrape:
All Done~

sys/compat/linux/linux_xattr.c
```diff
@@ -310,18 +310,35 @@ static int
...
 	else
 		error = kern_extattr_get_fd(td, args->fd, attrnamespace,
-		    attrname, args->value, args->size);
-	return (error == EPERM ? ENOATTR : error);
+		    attrname, NULL, 0, &size);
+	if (error != 0)
+		goto out;
+	if (args->value == NULL || args->size == 0)
+		return (0);
+	if (size > args->size)
```
kostikbel: What if the size of the attribute changes between the first and second extattr_get()?

ilovegrape: There is another related fix https://github.com/freebsd/freebsd-src/pull/2181, I need to consider both at the same time.

ilovegrape:
I pushed 2 changes below as I was thinking I need to make the change based #2181 .
1. linuxulator: O_PATH file descriptors return wrong errno for f*xattr(2), - this is passed the review in #2181
2. linuxulator: improve getxattr() error handling with atomic extattr_get_vp() - it seems that there is no need to base on #2181 after tried with different ways on the fix.

I just rebased the code to main, and will push it one more time if there is no compile issue in my local, to see if the compile issue can be resolved or not.

Can you please let me know whether I can keep both commit or keep only the 2nd one.

Xin

sys/compat/linux/linux_xattr.c
ilovegrape: See the updated code for this comments.

ilovegrape:
I rebased to main, but there is still falling in cross-check...

sys/kern/vfs_extattr.c
```diff
@@ -362,55 +362,65 @@ extattr_get_vp(struct vnode *vp, int attrnamespace, const char *attrname,
...
-		td->td_retval[0] = cnt;
-	} else
+	/*
+	 * If the buffer is too small, return ERANGE with the full size so
+	 * the caller can retry (Linux getxattr semantics).
+	 */
+	if (size > nbytes) {
 		td->td_retval[0] = size;
-#ifdef MAC
+		error = ERANGE;
```
kostikbel: I do not see how this could work.  If an error is returned, td_retval[0] is overwritten.

ilovegrape: My brain was short-circuited here when thinking of how to return the size for previous step...
To be more precise, it's better to remove the `td->td_retval[0] = size;`

sys/compat/linux/linux_xattr.c
```diff
@@ -310,18 +338,33 @@ static int
...
 		    attrname, args->value, args->size, args->follow, UIO_USERSPACE);
 	else
-		error = kern_extattr_get_fd(td, args->fd, attrnamespace,
+		error = kern_extattr_get_fp(td, fp, attrnamespace,
 		    attrname, args->value, args->size);
-	return (error == EPERM ? ENOATTR : error);
+	if (fp != NULL)
+		fdrop(fp, td);
+	return (error == EPERM || error == EOPNOTSUPP ? ENOATTR : error);
+out_err:
```
kostikbel: Why the out_err code is needed?  You might just as well to put out_err at the line 361.  I do not see much wrong with always translating EPERM/EOPNOTSUPP.

sys/kern/vfs_extattr.c
```diff
@@ -353,55 +362,64 @@ extattr_get_vp(struct vnode *vp, int attrnamespace, const char *attrname,
...
 #endif
 
-	error = VOP_GETEXTATTR(vp, attrnamespace, attrname, auiop, sizep,
+	/*
+	 * Probe the attribute size.  When data is NULL or nbytes is 0 the
+	 * caller only wants the size, so we are done after this.
+	 */
+	error = VOP_GETEXTATTR(vp, attrnamespace, attrname, NULL, &size,
 	    td->td_ucred, td);
+	if (error)
```
kostikbel: if (error != 0)

sys/kern/vfs_extattr.c
```diff
@@ -353,55 +362,64 @@ extattr_get_vp(struct vnode *vp, int attrnamespace, const char *attrname,
...
+		goto done;
+	if (data == NULL || nbytes == 0) {
```
kostikbel: Does this check change the semantic for native FreeBSD calls?
Also, for FreeBSD, for the case when data != NULL, do we need to do the newly added VOP_GETEXTATTR() call with uio == NULL, at all?

I still suggest what I proposed at the very beginning, AFAIR: add a bool argument to extattr_get_vp() which requests either FreeBSD or Linux semantic (treating nbytes == 0 as the size query).

ilovegrape: Thanks for the input.

I get your point, but I am trying to avoid adding such additional parameters for any functions in sys/kern as well as more wrapper functions...

But you are right, this changed a little with the original FreeBSD semantic that when data != NULL and the native FreeBSD application might not handle ERANGE as there is no such error.

I think the first version, that doing the size query and real query only in linuxulator layer without other changes, might be a good choice, although there is small chance of race condition.

Based on your suggestion, I am going to implement a new linux_extattr_get_vp() and keep the original extattr_get_vp() for FreeBSD to avoid too many function changes and making extattr_get_vp() more complex. ( I implemented in this way at the beginning, but I gave it up due to duplicated codes...)

ilovegrape: It seems that it's not as complex as what I thought, code updated based on the suggestion by adding a new parameter `is_linux` in `extattr_get_vp()`

sys/kern/vfs_extattr.c
```diff
@@ -391,6 +400,33 @@ extattr_get_vp(struct vnode *vp, int attrnamespace, const char *attrname,
 		goto done;
 #endif
 
+	if (is_linux && (data != NULL)) {
```
kostikbel: if (is_linux && data != NULL) {

sys/kern/vfs_extattr.c
```diff
@@ -348,7 +357,7 @@ kern_extattr_set_path(struct thread *td, const char *path, int attrnamespace,
  */
 static int
 extattr_get_vp(struct vnode *vp, int attrnamespace, const char *attrname,
-    void *data, size_t nbytes, struct thread *td)
+    void *data, size_t nbytes, bool is_linux, struct thread *td)
```
kostikbel: I suggest to call the parameter 'return_erange' or like.

Or after looking at it some more, it might be not a bad idea to just implement linux_extattr_get_vp() in the linuxolator, essentially copying extattr_get_vp() with your modifications.

ilovegrape: I finally take the 2nd option that having a new `linux_extattr_get_vp()` in the linuxulator~

sys/compat/linux/linux_xattr.c
```diff
@@ -306,22 +337,115 @@ linux_fremovexattr(struct thread *td, struct linux_fremovexattr_args *args)
...
+
+	if (nbytes > IOSIZE_MAX)
+		return (EINVAL);
+
+	vn_lock(vp, LK_SHARED | LK_RETRY);
+
+#ifdef MAC
+	error = mac_vnode_check_getextattr(td->td_ucred, vp, attrnamespace,
+	    attrname);
+	if (error)
```
kostikbel: error != 0

kostikbel (approved):

sys/compat/linux/linux_xattr.c
ilovegrape: done

ilovegrape:
Thanks @kostikbel for the continues reviews and the final approval.
hi @spmzt, no more changes for this pull request, you can take your time to move it forward~

spmzt:

PR #2264, opened 2026-06-05
iflib: Fix panic when netmap VALE drops packets to iflib driver

TeddyEngel:
When a netmap VALE switch drops packets destined for an iflib-based network driver (e.g., vmxnet3, igb, em), the kernel panics with a page fault in the driver's `isc_txd_encap()` function.

The crash occurs because:

1. `nm_vale_flush()` drops oversized packets by setting `slot->len = 0`
2. `iflib_netmap_txsync()` processes the zero-length slot but `seg_idx` stays 0 since `len == 0`
3. `isc_txd_encap()` is called with `pi->ipi_nsegs = 0`
4. The driver loops zero times but unconditionally sets the EOP flag on an uninitialized descriptor pointer

To fix, we ensure at least one segment (seg_idx = 1) for zero-length packets. The segment already has a valid address from the PNMB call and ds_len = 0, written unconditionally at loop top before the if (len) guard. This keeps the netmap and NIC ring indices in lockstep (both advance together at loop bottom), and the driver writes a valid zero-length descriptor with EOP instead of accessing invalid memory.

We also skip the transmit counter update (IFCOUNTER_OPACKETS/OBYTES) when pkt_len == 0, so packets VALE intended to drop are not reported as transmitted.

The fix is in iflib rather than individual drivers because all iflib drivers assume `nsegs > 0`, and we want to protect all drivers with a single fix. 

Tested on FreeBSD 16.0-CURRENT using the reproducer script from the bug report.

PR: [294726](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=294726)

### DIFF
--- a/sys/net/iflib.c
+++ b/sys/net/iflib.c
@@ -1051,6 +1051,13 @@ iflib_netmap_txsync(struct netmap_kring *kring, int flags)
 			}
 
 			if (!(slot->flags & NS_MOREFRAG)) {
+				/*
+				 * Ensure a packet made only of zero-length
+				 * slots still emits one descriptor.
+				 */
+				if (seg_idx == 0)
+					seg_idx = 1;
+
 				pi.ipi_len = pkt_len;
 				pi.ipi_nsegs = seg_idx;
 				pi.ipi_pidx = nic_i_start;
@@ -1062,8 +1069,10 @@ iflib_netmap_txsync(struct netmap_kring *kring, int flags)
 				DBG_COUNTER_INC(tx_encap);
 
 				/* Update transmit counters */
-				tx_bytes += pi.ipi_len;
-				tx_pkts++;
+				if (pkt_len > 0) {
+					tx_bytes += pi.ipi_len;
+					tx_pkts++;
+				}
 
 				/* Reinit per-packet info for the next one. */
 				flags = seg_idx = pkt_len = 0;


TeddyEngel:
CI failures seem unrelated to this PR - they're in lib/libc/gen/getcap.c (conflicting types for cgetent/cgetfirst/cgetnext). Same failures also appeared on PR #2263. Style Checker and commit checks pass.

TeddyEngel:
Hey @spmzt  - I was wondering if you know who could review this one?

markjdb:
I'm not sure this is right--aren't the netmap ring and NIC ring indices supposed to stay in sync? I don't think we can advance the netmap consumer index without also using a slot in the hardware ring.

See how iflib_netmap_txsync() derives `nic_i` from `nm_i` before the loop: with your patch the indices will get out of sync. Did you verify that the NIC is still able to pass traffic after it hits the 0-length packet case?

TeddyEngel:
@markjdb You are right about the lockstep problem, great catch

I reworked the patch so the zero-length dropped-packet case no longer skips the common nm_i/nic_i advancement path.

The current version keeps the existing zero-length segment compaction behavior, but handles the all-zero packet case by emitting one descriptor instead of calling isc_txd_encap() with ipi_nsegs == 0. It also avoids counting the dropped packet. 
Tested on em and vmxnet3 with the original reproducer + tested for normal traffic after the drop.

While testing I also found a case compaction unfortunately can't handle: multi-slot NS_MOREFRAG chains zeroed by VALE (e.g. pkt-gen -F 2 -l 10000) leave fewer initialized descriptors than slots consumed, so isc_txd_flush() publishes descriptors encap never wrote — on vmxnet3 the stale generation bit wedges the TX queue until reattach. 

A no-compaction patch variant (one descriptor per consumed slot, zero-length included) fixes this on em and vmxnet3, but exposes zero-length descriptors to drivers that the compaction currently shields, notably ixl/iavf (just to clarify - i couldn't test with the actual drivers, this is just from code inspection).

So I think this version might be the safer narrow fix for the reported panic, but there's a deeper question I'd appreciate guidance on. Tolerating zero-length descriptors doesn't look viable as a general driver contract since on ixl/iavf a zero-length buffer size is a hardware MDD violation (the ZERO_BSIZE checks). 
That seems to leave either iflib normalizing dropped slots itself, or VALE/netmap not consuming destination NIC slots for dropped packets in the first place. Happy to rework the patch in whichever direction you prefer.

PR #2265, opened 2026-06-06
libc: Fix assert() sanitiser for C++ contextual bool conversion

kfv:
Reported by: [dim@](https://github.com/freebsd/freebsd-src/pull/2203#issuecomment-4624496419), [aokblast@](https://reviews.freebsd.org/D57449)

Cc: @clausecker, @DimitryAndric, @aokblast

### DIFF
--- a/include/assert.h
+++ b/include/assert.h
@@ -73,7 +73,10 @@
  */
 #define __assert_sanitize(...)	((void)0)
 #else
-#define __assert_sanitize(...)	(void)sizeof(((bool(*)(bool))0)(__VA_ARGS__))
+char __assert_sanitize_arity(auto &&);
+#define __assert_sanitize(...)                                 \
+	((void)sizeof(__assert_sanitize_arity(__VA_ARGS__)),   \
+	 (void)sizeof((__VA_ARGS__) ? 1 : 1))
 #endif /* __cplusplus < 202002L */
 #else
 #define __assert_sanitize(...)	(void)sizeof(((_Bool(*)(_Bool))0)(__VA_ARGS__))
--- a/include/assert.h
+++ b/include/assert.h
@@ -46,8 +46,8 @@
 #undef __assert_unreachable
 
 #ifdef NDEBUG
-#define assert(e)	((void)0)
-#define _assert(e)	((void)0)
+#define assert(...)	((void)0)
+#define _assert(...)	((void)0)
 #if __BSD_VISIBLE
 #define __assert_unreachable()	__unreachable()
 #endif /* __BSD_VISIBLE */
--- a/include/assert.h
+++ b/include/assert.h
@@ -53,38 +53,15 @@
 #endif /* __BSD_VISIBLE */
 #else
 #ifdef __cplusplus
-#if __cplusplus < 202002L
-/*
- * C++ modes prior to C++20 cannot simultaneously satisfy all three
- * desirable properties of the sanitiser:
- *
- *   Approach                       No double-eval  Lambda support  Arity check
- *   -----------------------------  --------------  --------------  -----------
- *   sizeof(cast(expression))       yes             no              yes
- *   static_cast<bool>(expression)  no              yes             no
- *   (void)bool(expression)         no              yes             no
- *
- *   NOTE: C++20 introduced lambdas in unevaluated contexts; see P0315R4.
- *
- * Since no approach satisfies all three below C++20, the least harmful
- * choice is to forgo the check entirely rather than silently break one
- * of the remaining guarantees.
- *
- */
-#define __assert_sanitize(...)	((void)0)
+#define assert(...)	((void)(bool(__VA_ARGS__) ? ((void)0) :           \
+			    __assert(__func__, __FILE__, __LINE__,        \
+			    #__VA_ARGS__)))
 #else
-char __assert_sanitize_arity(auto &&);
-#define __assert_sanitize(...)                                 \
-	((void)sizeof(__assert_sanitize_arity(__VA_ARGS__)),   \
-	 (void)sizeof((__VA_ARGS__) ? 1 : 1))
-#endif /* __cplusplus < 202002L */
-#else
-#define __assert_sanitize(...)	(void)sizeof(((_Bool(*)(_Bool))0)(__VA_ARGS__))
-#endif /* __cplusplus */
-#define assert(...)	(__assert_sanitize(__VA_ARGS__),       \
-			    (__VA_ARGS__) ? (void)0 :          \
-			    __assert(__func__, __FILE__,       \
+#define assert(...)	((void)sizeof(((_Bool(*)(_Bool))0)(__VA_ARGS__)), \
+			    (__VA_ARGS__) ? (void)0 :                     \
+			    __assert(__func__, __FILE__,                  \
 			    __LINE__, #__VA_ARGS__))
+#endif /* __cplusplus */
 #define _assert(...)	assert(__VA_ARGS__)
 #if __BSD_VISIBLE
 #define __assert_unreachable()	assert(0 && "unreachable segment reached")


clausecker:
This looks like it could work.

aokblast:
Hi, I just run the full lit testsuites:

The following tests breaks with your patch:

```
  llvm-libc++-shared.cfg.in :: /home/blast/jenkins/workspace/llvm-build/build/runtimes/runtimes-bins/libcxx/test/libcxx/system_reserved_names.gen.py/cassert.compile.pass.cpp
  llvm-libc++-shared.cfg.in :: extensions/gnu/hash/specializations.verify.cpp
  llvm-libc++-shared.cfg.in :: extensions/libcxx/depr/depr.c.headers/extern_c.pass.cpp
  llvm-libc++-shared.cfg.in :: std/numerics/cfenv/cfenv.syn/cfenv.pass.cpp
```

Please make sure you have built libcxx before testing.

This is the command I use for testing:

```
 libcxx/utils/libcxx-lit ./build-crt ./libcxx/test/extensions/gnu/hash/specializations.verify.cpp
```

aokblast:
Also, please rebase your code as something has changed between your patch the latest commit.

kfv:
Sure. I made a small change and it appears to have fixed the issue. Tests are still running, I'll update once finished (might take a few hours.) Thanks!

DimitryAndric:
I'm testing this now with a limited ports build. I know that previously some ports were falling over, so I'll specifically be checking those. Will post results later.

kfv:
Although my tests are still in progress, I’ve pushed my changes in case you’d like to review them in the meantime.

kfv:
Alright, figured it out. The failure in `extensions/gnu/hash/specializations.verify.cpp` is essentially a false positive, I might say. The test expects exactly one instance of "does not provide a call operator", but our assert macro at the time expanding `__VA_ARGS__` thrice, so the count comes out wrong. We can probably just ignore it, right? Unless you'd like me to think of another tweak?

kfv:
BTW, I did consider doing the arity check as a lambda directly inside the macro per se (that I would like better, to be honest) but wasn't sure about the readability hit, so went with the out-of-line declaration instead. Curious what you think.

clausecker:
I do wonder if we should just do the inline function as given in the original C23 proposal.

Something like

```
static inline void
__assert__(const char *__func, const char *__FILE, int __LINE, const char *__cond, bool __outcome)
{
    if (!__outcome)
        __assert(__func, __FILE, __LINE, __cond);
}

#define assert(...) __assert__(__func__, __FILE__, __LINE__, #__VA_ARGS__, __VA_ARGS__ ? true : false)
```

This seems like it should do the right thing in both C and C++ modes.  Note that only `,` binds less than `?:`, so if `__VA_ARGS__` resolves to a comma expression, we end up passing the wrong number of arguments to `__assert__`.

kfv:
I don't think this is comparable to what we have currently. It'd only handle contextual conversion I believe. The difficulty is that we need all four properties simultaneously: arity enforcement, lambda support, no double evaluation, and contextual bool conversion (as raised in the recent bug report.) Actually, it turned out to be a surprisingly painful piece of code to get right across both C and C++ and their respective modes. Deceptively simple on the surface; anything but underneath. The current implementation is the only approach I've found so far that handles all of them.

clausecker:
The proposed code handles arity enforcement, does not evaluate the argument twice, and handles contextual bool conversion.  What would the problem be if a lambda was passed?

Arity is enforced in that a wrong number of arguments to `assert()` leads to a wrong number of arguments passed to `__assert__`, e.g. `assert(foo, bar)` translates to

    __assert__("some_func", "some_file.c", 1234, "foo, bar", foo, bar ? 1 : 0)

which then fails to compile as we pass one more arguments then expected (`foo` becomes an argument on its own).

kfv:
Lovely! Now I understand. You’re right. I’ll test it early tomorrow morning, but your explanation is clear and makes perfect sense.

aokblast:
> Alright, figured it out. The failure in `extensions/gnu/hash/specializations.verify.cpp` is essentially a false positive, I might say. The test expects exactly one instance of "does not provide a call operator", but our assert macro at the time expanding `__VA_ARGS__` thrice, so the count comes out wrong. We can probably just ignore it, right? Unless you'd like me to think of another tweak?

It is not the false positive, it is an intended behavior.
That is the reason I disable sanitizer in C++ case. I don't mind to mark the test case XFAILED so that it won't continuously complain on this. @DimitryAndric what is your opinion?

DimitryAndric:
> > Alright, figured it out. The failure in `extensions/gnu/hash/specializations.verify.cpp` is essentially a false positive, I might say. The test expects exactly one instance of "does not provide a call operator", but our assert macro at the time expanding `__VA_ARGS__` thrice, so the count comes out wrong. We can probably just ignore it, right? Unless you'd like me to think of another tweak?
> 
> It is not the false positive, it is an intended behavior. That is the reason I disable sanitizer in C++ case. I don't mind to mark the test case XFAILED so that it won't continuously complain on this. @DimitryAndric what is your opinion?

I may be missing something, but why are we now evaluating the expression multiple times?

kfv:
@clausecker's proposed solution yields only once, so it should be fine. The only issue I've noticed so far is misleading diagnostics in cases of arity misuse, which should be easy to fix. At his request, I'm testing another version on my end mimicking the [original example](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2264r4.html#example-implementation-bsd-license-on-macos) provided by Peter Sommerlad in his official proposal. Give me about half an hour, and I'll push changes shortly.

kfv:
@clausecker: Also, for what it's worth, I think I owe you an apology for not looking more closely at your comment last night. Your solution does indeed handle all the cases. It was late, and for some reason what I was seeing didn't look correct to me, but I was completely mistaken and rushed to judgment. My apologies, mate.

kfv:
Initial tests passed, and C++ modes before C++20 now handled correctly. Additional tests, including LLVM’s, are still running. I'll update you once they finish.

kfv:
@aokblast: No need for XFAIL; the recent change means this now passes as expected.

> $ libcxx/utils/libcxx-lit .build libcxx/test/extensions/gnu/hash/specializations.verify.cpp
> [  0%] Built target unwind_shared_objects
> [  1%] Built target unwind_shared
> [  1%] Built target generate-cxxabi-headers
> [ 77%] Built target generate-cxx-headers
> [ 78%] Built target cxxabi_shared_objects
> [ 78%] Built target cxxabi_shared
> [ 82%] Built target cxx_shared
> [ 83%] Built target unwind_static_objects
> [ 83%] Built target unwind_static
> [ 85%] Built target cxxabi_static_objects
> [ 85%] Built target cxxabi_static
> [ 90%] Built target cxx_static
> [ 90%] Built target cxx
> [ 90%] Built target cxx_experimental
> [100%] Built target generate-cxx-modules
> [100%] Built target cxx-test-depends
> llvm-lit: /usr/ports/devel/llvm19/work-default/llvm-project-19.1.7.src/libcxx/utils/libcxx/test/config.py:24: note: (llvm-libc++-shared.cfg.in) Using %{cxx} substitution: '/usr/bin/c++'
> llvm-lit: /usr/ports/devel/llvm19/work-default/llvm-project-19.1.7.src/libcxx/utils/libcxx/test/config.py:24: note: (llvm-libc++-shared.cfg.in) Using %{flags} substitution: '-pthread --target=aarch64-unknown-freebsd15.0'
> llvm-lit: /usr/ports/devel/llvm19/work-default/llvm-project-19.1.7.src/libcxx/utils/libcxx/test/config.py:24: note: (llvm-libc++-shared.cfg.in) Using %{compile_flags} substitution: '-nostdinc++ -I %{target-include-dir} -I %{include-dir} -I %{libcxx-dir}/test/support -std=c++26 -Werror -Wall -Wctad-maybe-unsupported -Wextra -Wshadow -Wundef -Wunused-template -Wno-unused-command-line-argument -Wno-attributes -Wno-pessimizing-move -Wno-noexcept-type -Wno-atomic-alignment -Wno-reserved-module-identifier -Wdeprecated-copy -Wdeprecated-copy-dtor -Wno-user-defined-literals -Wno-tautological-compare -Wsign-compare -Wunused-variable -Wunused-parameter -Wunreachable-code -Wno-unused-local-typedef -Wno-local-type-template-args -Wno-c++11-extensions -Wno-unknown-pragmas -Wno-pass-failed -Wno-mismatched-new-delete -Wno-redundant-move -Wno-self-move -D_LIBCPP_HAS_NO_PRAGMA_SYSTEM_HEADER -D_LIBCPP_ENABLE_EXPERIMENTAL -D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_NONE -Werror=thread-safety -Wuser-defined-warnings'
> llvm-lit: /usr/ports/devel/llvm19/work-default/llvm-project-19.1.7.src/libcxx/utils/libcxx/test/config.py:24: note: (llvm-libc++-shared.cfg.in) Using %{link_flags} substitution: '-lc++experimental -nostdlib++ -L %{lib-dir} -Wl,-rpath,%{lib-dir} -lc++'
> llvm-lit: /usr/ports/devel/llvm19/work-default/llvm-project-19.1.7.src/libcxx/utils/libcxx/test/config.py:24: note: (llvm-libc++-shared.cfg.in) Using %{exec} substitution: '%{executor} --execdir %T -- '
> llvm-lit: /usr/ports/devel/llvm19/work-default/llvm-project-19.1.7.src/libcxx/utils/libcxx/test/config.py:24: note: (llvm-libc++-shared.cfg.in) All available features: -faligned-allocation, -fsized-deallocation, LIBCXX-FREEBSD-FIXME, buildhost=freebsd, buildhost=freebsd15, c++26, c++experimental, can-create-symlinks, clang, clang-19, clang-19.1, clang-19.1.7, diagnose-if-support, executor-has-no-bash, freebsd, gcc-style-warnings, has-1024-bit-atomics, has-64-bit-atomics, has-fblocks, has-fconstexpr-steps, has-no-cxx-module-support, has-unix-headers, large_tests, libcpp-abi-version=1, libcpp-hardening-mode=none, libcpp-has-no-availability-markup, libcpp-has-thread-api-pthread, locale.cs_CZ.ISO8859-2, locale.en_US.UTF-8, locale.fr_CA.ISO8859-1, locale.fr_FR.UTF-8, locale.ja_JP.UTF-8, locale.ru_RU.UTF-8, locale.zh_CN.UTF-8, long_tests, no-tzdb, objective-c++, optimization=none, stdlib=libc++, stdlib=llvm-libc++, target=aarch64-unknown-freebsd15.0, thread-safety, verify-support
> -- Testing: 1 tests, 1 workers --
> PASS: llvm-libc++-shared.cfg.in :: extensions/gnu/hash/specializations.verify.cpp (1 of 1)
> 
> Testing Time: 5.04s
> 
> Total Discovered Tests: 1
>   Passed: 1 (100.00%)

kfv:
One more test remaining (with a very small patch Ima push afterwards.)

kfv:
Everything is good now. I would appreciate it if you could take another look.

aokblast:
> Everything is good now. I would appreciate it if you could take another look.

Looks like std/numerics/cfenv/cfenv.syn/cfenv.pass.cpp fails in this case.

kfv:
Could you please share what it complains about? It passes successfully on my end:

```
$ libcxx/utils/libcxx-lit .build libcxx/test/std/numerics/cfenv/cfenv.syn/cfenv.pass.cpp
[  0%] Built target unwind_shared_objects
[  1%] Built target unwind_shared
[  1%] Built target generate-cxxabi-headers
[ 77%] Built target generate-cxx-headers
[ 78%] Built target cxxabi_shared_objects
[ 78%] Built target cxxabi_shared
[ 82%] Built target cxx_shared
[ 83%] Built target unwind_static_objects
[ 83%] Built target unwind_static
[ 85%] Built target cxxabi_static_objects
[ 85%] Built target cxxabi_static
[ 90%] Built target cxx_static
[ 90%] Built target cxx
[ 90%] Built target cxx_experimental
[100%] Built target generate-cxx-modules
[100%] Built target cxx-test-depends
llvm-lit: /usr/ports/devel/llvm19/work-default/llvm-project-19.1.7.src/libcxx/utils/libcxx/test/config.py:24: note: (llvm-libc++-shared.cfg.in) Using %{cxx} substitution: '/usr/bin/c++'
llvm-lit: /usr/ports/devel/llvm19/work-default/llvm-project-19.1.7.src/libcxx/utils/libcxx/test/config.py:24: note: (llvm-libc++-shared.cfg.in) Using %{flags} substitution: '-pthread --target=aarch64-unknown-freebsd15.0'
llvm-lit: /usr/ports/devel/llvm19/work-default/llvm-project-19.1.7.src/libcxx/utils/libcxx/test/config.py:24: note: (llvm-libc++-shared.cfg.in) Using %{compile_flags} substitution: '-nostdinc++ -I %{target-include-dir} -I %{include-dir} -I %{libcxx-dir}/test/support -std=c++26 -Werror -Wall -Wctad-maybe-unsupported -Wextra -Wshadow -Wundef -Wunused-template -Wno-unused-command-line-argument -Wno-attributes -Wno-pessimizing-move -Wno-noexcept-type -Wno-atomic-alignment -Wno-reserved-module-identifier -Wdeprecated-copy -Wdeprecated-copy-dtor -Wno-user-defined-literals -Wno-tautological-compare -Wsign-compare -Wunused-variable -Wunused-parameter -Wunreachable-code -Wno-unused-local-typedef -Wno-local-type-template-args -Wno-c++11-extensions -Wno-unknown-pragmas -Wno-pass-failed -Wno-mismatched-new-delete -Wno-redundant-move -Wno-self-move -D_LIBCPP_HAS_NO_PRAGMA_SYSTEM_HEADER -D_LIBCPP_ENABLE_EXPERIMENTAL -D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_NONE -Werror=thread-safety -Wuser-defined-warnings'
llvm-lit: /usr/ports/devel/llvm19/work-default/llvm-project-19.1.7.src/libcxx/utils/libcxx/test/config.py:24: note: (llvm-libc++-shared.cfg.in) Using %{link_flags} substitution: '-lc++experimental -nostdlib++ -L %{lib-dir} -Wl,-rpath,%{lib-dir} -lc++'
llvm-lit: /usr/ports/devel/llvm19/work-default/llvm-project-19.1.7.src/libcxx/utils/libcxx/test/config.py:24: note: (llvm-libc++-shared.cfg.in) Using %{exec} substitution: '%{executor} --execdir %T -- '
llvm-lit: /usr/ports/devel/llvm19/work-default/llvm-project-19.1.7.src/libcxx/utils/libcxx/test/config.py:24: note: (llvm-libc++-shared.cfg.in) All available features: -faligned-allocation, -fsized-deallocation, LIBCXX-FREEBSD-FIXME, buildhost=freebsd, buildhost=freebsd15, c++26, c++experimental, can-create-symlinks, clang, clang-19, clang-19.1, clang-19.1.7, diagnose-if-support, executor-has-no-bash, freebsd, gcc-style-warnings, has-1024-bit-atomics, has-64-bit-atomics, has-fblocks, has-fconstexpr-steps, has-no-cxx-module-support, has-unix-headers, large_tests, libcpp-abi-version=1, libcpp-hardening-mode=none, libcpp-has-no-availability-markup, libcpp-has-thread-api-pthread, locale.cs_CZ.ISO8859-2, locale.en_US.UTF-8, locale.fr_CA.ISO8859-1, locale.fr_FR.UTF-8, locale.ja_JP.UTF-8, locale.ru_RU.UTF-8, locale.zh_CN.UTF-8, long_tests, no-tzdb, objective-c++, optimization=none, stdlib=libc++, stdlib=llvm-libc++, target=aarch64-unknown-freebsd15.0, thread-safety, verify-support
-- Testing: 1 tests, 1 workers --
PASS: llvm-libc++-shared.cfg.in :: std/numerics/cfenv/cfenv.syn/cfenv.pass.cpp (1 of 1)

Testing Time: 0.46s

Total Discovered Tests: 1
  Passed: 1 (100.00%)
```

aokblast (approved):
Sorry. It breaks not because of your patch. This patch works. Thanks!

kfv:
Cheers!

clausecker:
Will do some final testing and then land this one in the next days.  Thank you for the review, @aokblast.

clausecker:
I don't like that this still expands `__VA_ARGS__` twice in C mode and that the implementation differs between C and C++ mode. But I guess we can always revise the code at a later point.

DimitryAndric:
I still think it would be safest to do a ports exp-run with the changes applied. There are so many places where it could go wrong, it's difficult to predict in advance where something unexpected will occur.

But still, the version in main is now definitely breaking things, so getting a fix in quickly is probably more important.

kfv:
> I don't like that this still expands `__VA_ARGS__` twice in C mode and that the implementation differs between C and C++ mode. But I guess we can always revise the code at a later point.

You were right. I've just taken care of that. The only thing left is making it a single logic for both, which, as you mentioned, we can probably tackle later.

include/assert.h
```diff
@@ -46,42 +46,21 @@
...
-#define __assert_sanitize(...)	((void)0)
-#else
-#define __assert_sanitize(...)	(void)sizeof(((bool(*)(bool))0)(__VA_ARGS__))
-#endif /* __cplusplus < 202002L */
+#define assert(...)	((void)(bool(__VA_ARGS__) ? ((void)0) :           \
+			    __assert(__func__, __FILE__, __LINE__,        \
+			    #__VA_ARGS__)))
 #else
-#define __assert_sanitize(...)	(void)sizeof(((_Bool(*)(_Bool))0)(__VA_ARGS__))
+#define assert(...)	((void)sizeof(((_Bool(*)(_Bool))0)(__VA_ARGS__)), \
```
clausecker, line 60: How is that supposed to work now?  This just calls `__assert` unconditionally, i.e. all assertions fail in C mode unless `NDEBUG` is defined.

I'm not sure why you keep saying “you are right” when I don't even say that there is something wrong.  If you are using an AI to write these responses, please stop.  You are smart enough to write your own comments and I don't like talking to bots.

I would preferably go ahead with committing your previous patch set, i.e. the one approved by @aokblast (https://github.com/freebsd/freebsd-src/commit/9e035303dd48382787d53e4d7e7b59841a58c070) for now, as that has been verified to some extent.  You ok with that?

kfv: Did you even test it or is this just another assumption chucked at me without checking?

kfv:
```
root@dev:~/Developer/Tests # cat empty.c
#include <assert.h>

void f(void) { assert(); }
root@dev:~/Developer/Tests # cat arity-pass.c 
#include <assert.h>

void
f(void)
{
	struct A {int x,y;};
	assert((struct A){1,2}.x);
}

root@dev:~/Developer/Tests # cat arity-fail.c 
#include <assert.h>

void
f(void)
{
	struct A {int x,y;};
	assert((struct A){1,2}.x, "FAIL");
}

root@dev:~/Developer/Tests # cat simple.c 
#include <assert.h>

void f(void) { assert(2 > 1); }
root@dev:~/Developer/Tests # cc -c empty.c
empty.c:3:16: error: too few arguments to function call, expected 1, have 0
    3 | void f(void) { assert(); }
      |                ^~~~~~~~
/usr/include/assert.h:60:67: note: expanded from macro 'assert'
   60 | #define assert(...)     ((void)sizeof(((_Bool(*)(_Bool))0)(__VA_ARGS__)), \
      |                                       ~~~~~~~~~~~~~~~~~~~~            ^
1 error generated.
root@dev:~/Developer/Tests # cc -c arity-fail.c 
arity-fail.c:7:28: error: too many arguments to function call, expected 1, have 2
    7 |         assert((struct A){1,2}.x, "FAIL");
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
/usr/include/assert.h:60:56: note: expanded from macro 'assert'
   60 | #define assert(...)     ((void)sizeof(((_Bool(*)(_Bool))0)(__VA_ARGS__)), \
      |                                       ~~~~~~~~~~~~~~~~~~~~ ^~~~~~~~~~~
1 error generated.
root@dev:~/Developer/Tests # cc -c arity-pass.c 
root@dev:~/Developer/Tests # cc -c simple.c 
root@dev:~/Developer/Tests # 
```

clausecker: This is not about the arity check, but about the core assertion mechanism.  The `assert` macro must only call `__assert` if the assertion has actually failed, but you have removed the ternary expression that makes that happen.

Try this C source file:

```
#include <assert.h>

int main()
{
    assert(true);
}
```

With your updated `assert.h`, this will now yield a failed assertion at runtime, whereas the assertion is clearly true.

kfv: > I'm not sure why you keep saying “you are right” when I don't even say that there is something wrong.

You don’t need to necessarily say I’m wrong, mate. "you’re right” is just a polite way of acknowledging what’s been pointed out, innit? Just trying to be polite.

>  If you are using an AI to write these responses, please stop.

Why do people keep assuming I’m using AI every single time over and over and over again? What’s wrong with the way I’m writing or coding or whatever that makes they jump to that conclusion every time?

> You are smart enough to write your own comments and I don't like talking to bots.

Thanks for the compliment, but you're talking to me, mate, I assure you, no artificial agent sitting at my computer writing or coding on my behalf.

kfv: I’ve got no issue with that, but please take a proper look at my recent patch as well. I might have missed something, that’s entirely possible, I ain't hubristically claiming that I'm always right (quite the contrary, I've managed to end up with so many oversights that I feel terribly ashamed for,) but it wasn’t just thrown in without any checking or thought.

clausecker: > You don’t need to necessarily say I’m wrong, mate. "you’re right” is just a polite way of acknowledging what’s been pointed out, innit? Just trying to be polite.

This may be a language barrier thing, but starting every response with “You're right” strikes me as a very sycophantic way of communication.  We are here to work out your patch together as equals, and if you tell me I'm right even when I am not or when I haven't made any claim to be tested, I can no longer be sure that you actually mean that or that I have actually convinced you of anything.  The signal of truth is lost, making the conversation much harder to continue.

Starting every response with an affirmation like that is also something AIs really like to do, so my bullshit alarm goes off like crazy when a human does it.  This is also why I suspected you of AI use, though I am relieved to hear that I am talking to you, not a bot.

The reason I care is that when people outsource their communication to AIs, they often stop understanding the technical details of the conversation and it can become impossible to get through to the user behind the bot.  I can run my own AI tools if necessary; a reporter having me talk to his AI is a waste of my time and quite disrespectful.  I will disengage from bug reports if that happens.

aokblast: @kfv I run your latest patch with llvm libcxx tests. It passes correctly. But I also have same problem on NDEBUG as @clausecker. Please give me more time to check if there is any hiding issue.

clausecker: > I’ve got no issue with that, but please take a proper look at my recent patch as well. I might have missed something, that’s entirely possible, I ain't hubristically claiming that I'm always right (quite the contrary, I've managed to end up with so many oversights that I feel terribly ashamed for,) but it wasn’t just thrown in without any checking or thought.

I did take a look and have observed that the new definition of `assert()` for C source files seems defective in that while there is no compile-time error, all assertions in C code will now fail even if the condition holds true.

Please check the example program I provided.

aokblast: > I’ve got no issue with that, but please take a proper look at my recent patch as well. I might have missed something, that’s entirely possible, I ain't hubristically claiming that I'm always right (quite the contrary, I've managed to end up with so many oversights that I feel terribly ashamed for,) but it wasn’t just thrown in without any checking or thought.

I write a sample example to reproduce @clausecker's point.
libcxx runs in C++ mode so there is no any problem as your C++ change is correct.

```
blast@blast-devserv ~/P/C/assertion_test> cat main.c
#include <assert.h>

int main() {
    assert(1);
}
blast@blast-devserv ~/P/C/assertion_test> cc main.c -o main
blast@blast-devserv ~/P/C/assertion_test> ./main
Assertion failed: (1), function main, file main.c, line 4.
fish: Job 1, './main' terminated by signal SIGABRT (Abort)
```

Intended behavior: No assertion happens.

kfv: > This is not about the arity check, but about the core assertion mechanism. The assert macro must only call __assert if the assertion has actually failed, but you have removed the ternary expression that makes that happen.
>
> Try this C source file:
>
> #include <assert.h>
>
> int main()
> {
>     assert(true);
> }
> With your updated assert.h, this will now yield a failed assertion at runtime, whereas the assertion is clearly true.

My sincere apologies, I had removed `(__VA_ARGS__) ? (void)0 :` line by mistake, clumsy me 😖

Cc: @clausecker, @aokblast

kfv: By getting rid of the `__assert_arity`, and hence its expansion, on the following code:

```
#include <assert.h>

void
f(void)
{
	struct A {int x,y;};
	assert((struct A){1,2}.x, "FAIL");
}
```

We get:

```
arity-fail.c:7:28: error: too many arguments to function call, expected 1, have 2
    7 |         assert((struct A){1,2}.x, "FAIL");
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
/usr/include/assert.h:60:56: note: expanded from macro 'assert'
   60 | #define assert(...)     ((void)sizeof(((_Bool(*)(_Bool))0)(__VA_ARGS__)), \
      |                                       ~~~~~~~~~~~~~~~~~~~~ ^~~~~~~~~~~
arity-fail.c:7:25: warning: left operand of comma operator has no effect [-Wunused-value]
    7 |         assert((struct A){1,2}.x, "FAIL");
      |                ~~~~~~~~~~~~~~~ ^
/usr/include/assert.h:61:9: note: expanded from macro 'assert'
   61 |                             (__VA_ARGS__) ? (void)0 :                     \
      |                              ^~~~~~~~~~~
1 warning and 1 error generated.
```

Instead of the following:

```
root@dev:~/Developer/Tests # cc arity-fail.c
arity-fail.c:7:28: error: too many arguments to function call, expected 1, have 2
    7 |         assert((struct A){1,2}.x, "FAIL");
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
/usr/include/assert.h:61:37: note: expanded from macro 'assert'
   61 | #define assert(...)     (__assert_arity(__VA_ARGS__),       \
      |                          ~~~~~~~~~~~~~~~^~~~~~~~~~~~
/usr/include/assert.h:60:63: note: expanded from macro '__assert_arity'
   60 | #define __assert_arity(...)     (void)sizeof(((_Bool(*)(_Bool))0)(__VA_ARGS__))
      |                                              ~~~~~~~~~~~~~~~~~~~~ ^~~~~~~~~~~
arity-fail.c:7:25: warning: left operand of comma operator has no effect [-Wunused-value]
    7 |         assert((struct A){1,2}.x, "FAIL");
      |                ~~~~~~~~~~~~~~~ ^
/usr/include/assert.h:62:9: note: expanded from macro 'assert'
   62 |                             (__VA_ARGS__) ? (void)0 :       \
      |                              ^~~~~~~~~~~
1 warning and 1 error generated.
```

Which is how I just wanted to address your double-expansion concern.

kfv: > This may be a language barrier thing, but starting every response with “You're right” strikes me as a very sycophantic way of communication. We are here to work out your patch together as equals, and if you tell me I'm right even when I am not or when I haven't made any claim to be tested, I can no longer be sure that you actually mean that or that I have actually convinced you of anything. The signal of truth is lost, making the conversation much harder to continue.
>
> Starting every response with an affirmation like that is also something AIs really like to do, so my bullshit alarm goes off like crazy when a human does it. This is also why I suspected you of AI use, though I am relieved to hear that I am talking to you, not a bot.

I understand where you're coming from and that does make sense, but it's not our first time working together, right? And you've seen me disagree with you, so I'm not like AI agents saying "Ah, right" etc. to keep things smooth and please others. But I understand, it can look like AI sometimes, especially as I sometimes oversell stuff (and that's nothing good, I know.)

And I don't think it's solely a language barrier since I've heard this in my home country as well (and not recently but so many years ago before agents come to life; at the time, my colleagues would sometimes joke about that.) So it's possibly something I've got to fix on my end. I don't know. I'll find a way through it. But thanks for explaining. Now it's more clear why people might think like that sometimes.

> The reason I care is that when people outsource their communication to AIs, they often stop understanding the technical details of the conversation and it can become impossible to get through to the user behind the bot. I can run my own AI tools if necessary; a reporter having me talk to his AI is a waste of my time and quite disrespectful. I will disengage from bug reports if that happens.

I completely agree with you. I’ve encountered the same situation many times at work where I just wanted to pull my hair out. So I can assure you that, even if I sometimes come across as clumsy or a bit awkward, we’re on the same page and I'm just me. And I apologise if I occasionally show up careless, I must make sure to test everything more thoroughly before submitting anything.

kfv: @clausecker And sorry if I lost my temper and overreacted, just got too upset for a second.

clausecker:
Looks fine now.  Will do another build test and then try to land this change and #2223.

kfv:
Thanks.

include/assert.h
clausecker: > By getting rid of the `__assert_arity`, and hence its expansion, on the following code:
> 
> ```
> #include <assert.h>
> 
> void
> f(void)
> {
> 	struct A {int x,y;};
> 	assert((struct A){1,2}.x, "FAIL");
> }
> ```
> 
> We get:
> 
> ```
> arity-fail.c:7:28: error: too many arguments to function call, expected 1, have 2
>     7 |         assert((struct A){1,2}.x, "FAIL");
>       |         ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
> /usr/include/assert.h:60:56: note: expanded from macro 'assert'
>    60 | #define assert(...)     ((void)sizeof(((_Bool(*)(_Bool))0)(__VA_ARGS__)), \
>       |                                       ~~~~~~~~~~~~~~~~~~~~ ^~~~~~~~~~~
> arity-fail.c:7:25: warning: left operand of comma operator has no effect [-Wunused-value]
>     7 |         assert((struct A){1,2}.x, "FAIL");
>       |                ~~~~~~~~~~~~~~~ ^
> /usr/include/assert.h:61:9: note: expanded from macro 'assert'
>    61 |                             (__VA_ARGS__) ? (void)0 :                     \
>       |                              ^~~~~~~~~~~
> 1 warning and 1 error generated.
> ```
> 
> Instead of the following:
> 
> ```
> root@dev:~/Developer/Tests # cc arity-fail.c
> arity-fail.c:7:28: error: too many arguments to function call, expected 1, have 2
>     7 |         assert((struct A){1,2}.x, "FAIL");
>       |         ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
> /usr/include/assert.h:61:37: note: expanded from macro 'assert'
>    61 | #define assert(...)     (__assert_arity(__VA_ARGS__),       \
>       |                          ~~~~~~~~~~~~~~~^~~~~~~~~~~~
> /usr/include/assert.h:60:63: note: expanded from macro '__assert_arity'
>    60 | #define __assert_arity(...)     (void)sizeof(((_Bool(*)(_Bool))0)(__VA_ARGS__))
>       |                                              ~~~~~~~~~~~~~~~~~~~~ ^~~~~~~~~~~
> arity-fail.c:7:25: warning: left operand of comma operator has no effect [-Wunused-value]
>     7 |         assert((struct A){1,2}.x, "FAIL");
>       |                ~~~~~~~~~~~~~~~ ^
> /usr/include/assert.h:62:9: note: expanded from macro 'assert'
>    62 |                             (__VA_ARGS__) ? (void)0 :       \
>       |                              ^~~~~~~~~~~
> 1 warning and 1 error generated.
> ```
> 
> Which is how I just wanted to address your double-expansion concern.

Ah ok, that makes sense.  I meant something else with “double-expansion.”  It's that `__VA_ARGS__` is expanded twice; once for the arity check and once for the assertion check.  So besides blowing up the number of tokens in the preprocessed source, if there is anything in `__VA_ARGS__` that triggers a warning, that will now be printed twice, once for each expansion.

Once again, this is just a minor issue and I'll just go ahead and land your patch for now (unless I find some late-breaking issue).

Please don't worry too much about the communication style stuff.  There are many different opinions about it and mine are just mine.  I do think you submit good changes and I appreciate your attention to detail.  Testing is hard; don't worry about stuff going wrong until it's shipped in a release somewhere.  We can always fix things or revert if an oversight is discovered.

aokblast (approved):
Again as LLVM lit is happy on the new change.

clausecker:

PR #2271, opened 2026-06-08
stand: zero struct stat in fstat() to avoid uninitialized fields

Defenso-EBO:
## Summary

The standalone `fstat()` in `stand/libsa/fstat.c` returns a `struct stat`
whose fields were never zero-initialized. Filesystems that don't populate
every field leave the rest as stack garbage. The loader's ZFS implementation
(`zfs_dnode_stat` in `stand/libsa/zfs/zfsimpl.c`) only sets `st_mode`,
`st_uid`, `st_gid` and `st_size` — never `st_dev`.

This breaks veriexec on ZFS root: `fingerprint_info_lookup()` in
`lib/libsecureboot/veopen.c` compares the manifest's `st_dev` against the
target file's `st_dev`. With both uninitialized, they almost always differ,
the matching manifest entry is skipped, and the kernel fails verification with
a spurious `no entry` under UEFI Secure Boot. UFS is unaffected because
`stand/libsa/ufs.c` sets `st_dev` from the filesystem id.

## Fix

Zero the `struct stat` at the start of `fstat()` so unset fields read as 0
rather than stack garbage. Since `stat()` calls `fstat()` internally, both
paths are covered. This is a general robustness improvement, not specific to
veriexec.

## Testing

Tested on FreeBSD 16.0-CURRENT (amd64), ZFS-on-GELI root, EFI loader built with
`WITH_BEARSSL WITH_LOADER_VERIEXEC WITH_LOADER_VERIEXEC_VECTX
WITH_LOADER_EFI_SECUREBOOT`. Before: `Unverified /boot/kernel/kernel: no entry`.
After (with `veopen.c` unmodified): `Verified /boot/kernel/kernel`, system boots
under Secure Boot with verified boot active.

Bugzilla: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=295935

### DIFF
--- a/stand/efi/boot1/zfs_module.c
+++ b/stand/efi/boot1/zfs_module.c
@@ -143,6 +143,7 @@ load(const char *filepath, dev_info_t *devinfo, void **bufp, size_t *bufsize)
 	dnode_phys_t dn;
 	struct stat st;
 	uint64_t rootobj;
+	uint64_t objnum;
 	int err;
 	void *buf;
 
@@ -205,7 +206,7 @@ load(const char *filepath, dev_info_t *devinfo, void **bufp, size_t *bufsize)
 		return (EFI_NOT_FOUND);
 	}
 
-	if ((err = zfs_lookup(&zmount, filepath, &dn)) != 0) {
+	if ((err = zfs_lookup(&zmount, filepath, &dn, &objnum)) != 0) {
 		if (err == ENOENT) {
 			DPRINTF("Failed to find '%s' on pool '%s' (%d)\n",
 			    filepath, spa->spa_name, err);
@@ -216,7 +217,9 @@ load(const char *filepath, dev_info_t *devinfo, void **bufp, size_t *bufsize)
 		return (EFI_INVALID_PARAMETER);
 	}
 
-	if ((err = zfs_dnode_stat(spa, &dn, &st)) != 0) {
+	/* zmount.fsid_guid was set by zfs_mount_impl() above. */
+	if ((err = zfs_dnode_stat(spa, &dn, &st, zmount.fsid_guid,
+	    objnum)) != 0) {
 		printf("Failed to stat '%s' on pool '%s' (%d)\n", filepath,
 		    spa->spa_name, err);
 		return (EFI_INVALID_PARAMETER);
--- a/stand/libsa/zfs/zfs.c
+++ b/stand/libsa/zfs/zfs.c
@@ -85,6 +85,7 @@ struct fs_ops zfs_fsops = {
 struct file {
 	off_t		f_seekp;	/* seek pointer */
 	dnode_phys_t	f_dnode;
+	uint64_t	f_objnum;	/* object number (st_ino) */
 	uint64_t	f_zap_type;	/* zap type for readdir */
 	uint64_t	f_num_leafs;	/* number of fzap leaf blocks */
 	zap_leaf_phys_t	*f_zap_leaf;	/* zap leaf buffer */
@@ -120,7 +121,7 @@ zfs_open(const char *upath, struct open_file *f)
 		return (ENOMEM);
 	f->f_fsdata = fp;
 
-	rc = zfs_lookup(mount, upath, &fp->f_dnode);
+	rc = zfs_lookup(mount, upath, &fp->f_dnode, &fp->f_objnum);
 	fp->f_seekp = 0;
 	if (rc) {
 		f->f_fsdata = NULL;
@@ -214,10 +215,11 @@ static int
 zfs_stat(struct open_file *f, struct stat *sb)
 {
 	struct devdesc *dev = f->f_devdata;
-	const spa_t *spa = ((struct zfsmount *)dev->d_opendata)->spa;
+	struct zfsmount *zm = dev->d_opendata;
 	struct file *fp = (struct file *)f->f_fsdata;
 
-	return (zfs_dnode_stat(spa, &fp->f_dnode, sb));
+	return (zfs_dnode_stat(zm->spa, &fp->f_dnode, sb, zm->fsid_guid,
+	    fp->f_objnum));
 }
 
 static int
--- a/stand/libsa/zfs/zfsimpl.c
+++ b/stand/libsa/zfs/zfsimpl.c
@@ -48,6 +48,7 @@ struct zfsmount {
 	const spa_t		*spa;
 	objset_phys_t		objset;
 	uint64_t		rootobj;
+	uint64_t		fsid_guid;	/* ds_fsid_guid of mounted dataset */
 	STAILQ_ENTRY(zfsmount)	next;
 };
 
@@ -3433,7 +3434,8 @@ done:
  * and return its details in *objset
  */
 static int
-zfs_mount_dataset(const spa_t *spa, uint64_t objnum, objset_phys_t *objset)
+zfs_mount_dataset(const spa_t *spa, uint64_t objnum, objset_phys_t *objset,
+    uint64_t *fsid_guid)
 {
 	dnode_phys_t dataset;
 	dsl_dataset_phys_t *ds;
@@ -3449,6 +3451,17 @@ zfs_mount_dataset(const spa_t *spa, uint64_t objnum, objset_phys_t *objset)
 		    (uintmax_t)objnum);
 		return (EIO);
 	}
+
+	/*
+	 * Capture the dataset's intrinsic on-disk identifier so callers can
+	 * expose it as st_dev (matches the kernel, which derives the fsid from
+	 * dmu_objset_fsid_guid()/ds_fsid_guid).  ds_fsid_guid is a 56-bit ID
+	 * that may change to avoid collisions; ds_guid is a never-changing
+	 * 64-bit ID and would be the alternative if absolute stability over
+	 * time were required -- to be decided in review.
+	 */
+	if (fsid_guid != NULL)
+		*fsid_guid = ds->ds_fsid_guid;
 
 	return (0);
 }
@@ -3518,7 +3531,7 @@ zfs_mount_impl(const spa_t *spa, uint64_t rootobj, struct zfsmount *mount)
 		return (EIO);
 	}
 
-	if (zfs_mount_dataset(spa, rootobj, &mount->objset)) {
+	if (zfs_mount_dataset(spa, rootobj, &mount->objset, &mount->fsid_guid)) {
 		printf("ZFS: can't open root filesystem\n");
 		return (EIO);
 	}
@@ -3703,8 +3716,16 @@ zfs_spa_init(spa_t *spa)
 }
 
 static int
-zfs_dnode_stat(const spa_t *spa, dnode_phys_t *dn, struct stat *sb)
+zfs_dnode_stat(const spa_t *spa, dnode_phys_t *dn, struct stat *sb,
+    uint64_t fsid_guid, uint64_t objnum)
 {
+
+	/*
+	 * Zero the whole struct first so fields this function does not fill
+	 * (st_nlink, timestamps, st_blocks, ...) hold 0 rather than stack
+	 * garbage.  st_dev/st_ino are then overwritten with real values below.
+	 */
+	memset(sb, 0, sizeof(*sb));
 
 	if (dn->dn_bonustype != DMU_OT_SA) {
 		znode_phys_t *zp = (znode_phys_t *)dn->dn_bonus;
@@ -3753,6 +3774,15 @@ zfs_dnode_stat(const spa_t *spa, dnode_phys_t *dn, struct stat *sb)
 		    SA_SIZE_OFFSET);
 		free(buf);
 	}
+
+	/*
+	 * Populate the identity fields used by veriexec to tell apart the same
+	 * path on different datasets/filesystems.  dev_t and ino_t are 64-bit
+	 * on FreeBSD, so the dataset GUID and object number fit directly with
+	 * no hashing or truncation.
+	 */
+	sb->st_dev = (dev_t)fsid_guid;
+	sb->st_ino = (ino_t)objnum;
 
 	return (0);
 }
@@ -3818,7 +3848,8 @@ struct obj_list {
  * Lookup a file and return its dnode.
  */
 static int
-zfs_lookup(const struct zfsmount *mount, const char *upath, dnode_phys_t *dnode)
+zfs_lookup(const struct zfsmount *mount, const char *upath,
+    dnode_phys_t *dnode, uint64_t *objnum_out)
 {
 	int rc;
 	uint64_t objnum;
@@ -3904,7 +3935,8 @@ zfs_lookup(const struct zfsmount *mount, const char *upath, dnode_phys_t *dnode)
 		element[q - p] = 0;
 		p = q;
 
-		if ((rc = zfs_dnode_stat(spa, &dn, &sb)) != 0)
+		/* Only st_mode is inspected here, so identity is irrelevant. */
+		if ((rc = zfs_dnode_stat(spa, &dn, &sb, 0, 0)) != 0)
 			goto done;
 		if (!S_ISDIR(sb.st_mode)) {
 			rc = ENOTDIR;
@@ -3929,7 +3961,8 @@ zfs_lookup(const struct zfsmount *mount, const char *upath, dnode_phys_t *dnode)
 		/*
 		 * Check for symlink.
 		 */
-		rc = zfs_dnode_stat(spa, &dn, &sb);
+		/* Only st_mode is inspected here, so identity is irrelevant. */
+		rc = zfs_dnode_stat(spa, &dn, &sb, 0, 0);
 		if (rc)
 			goto done;
 		if (S_ISLNK(sb.st_mode)) {
@@ -3976,6 +4009,14 @@ zfs_lookup(const struct zfsmount *mount, const char *upath, dnode_phys_t *dnode)
 	}
 
 	*dnode = dn;
+	/*
+	 * objnum tracks the object number of the dnode we just resolved (the
+	 * loader's equivalent of the kernel's z_id/db_object); hand it back so
+	 * the file can expose it as st_ino.  Callers that do not need it pass
+	 * NULL.
+	 */
+	if (objnum_out != NULL)
+		*objnum_out = objnum;
 done:
 	STAILQ_FOREACH_SAFE(entry, &on_cache, entry, tentry)
 		free(entry);


stand/libsa/fstat.c
kevans91: The blank line between declarations and code must be preserved

Defenso-EBO: I will clean the commit tree and fix it when I go back to my laptop. Thank you for the review

VexedUXR:
4 out of the 14 filesystems in libsa already zero out the struct (note that splitfs, gzipfs and bzipfs just call `fstat` on the underlying filesystem, so it's more like 4 out of 11). Maybe it should be initialized in each filesystem's callback instead? The kernel also does it that way, FWIW.

I'm also not sure that initializing both `st_dev` and `st_ino` to zero is the right approach, don't they need to be unique here?
CC @sgerraty 

libsa(3) indicates that some fields are left undefined. This list should probably be updated to include `st_dev` and `st_ino` (and maybe also say that other fields not listed here should be initialized to zero?). Likely out of scope for this PR though.
>      int fstat(int fd, struct stat *sb)
>
>                 The stat() and fstat() functions only fill out the following
>                 fields in the sb structure:
>                 st_mode,st_nlink,st_uid,st_gid,st_size.  The tftp file system
>                 cannot provide meaningful values for this call, and the
>                 cd9660 file system always reports files having uid/gid of
>                 zero.

sgerraty:
On Thu, 11 Jun 2026 13:04:02 +0000 (UTC), Ahmad Khalifa writes:
>4 out of the 14 filesystems in libsa already zero out the struct (note that sp
>litfs, gzipfs and bzipfs just call `fstat` on the underlying filesystem, so it
>'s more like 4 out of 11). Maybe it should be initialized in each filesystem's
> callback instead? The kernel also does it that way, FWIW.
>
>I'm also not sure that initializing both `st_dev` and `st_ino` to zero is the 
>right approach, don't they need to be unique here?
>CC @sgerraty

They do if doing LOADER_VERIEXEC, to ensure that the same pathname on
two different filesystems are tracked separately.

>libsa.3 indicates that some fields are left undefined. This list should probab
>ly be updated to include `st_dev` and `st_ino` (and maybe also say that other 
>fields not listed here should be initialized to zero?). Likely out of scope fo
>r this PR though.
>>      int fstat(int fd, struct stat *sb)
>>
>>                 The stat() and fstat() functions only fill out the following
>>                 fields in the sb structure:
>>                 st_mode,st_nlink,st_uid,st_gid,st_size.  The tftp file syste
>m
>>                 cannot provide meaningful values for this call, and the
>>                 cd9660 file system always reports files having uid/gid of
>>                 zero.

kevans91:
> They do if doing LOADER_VERIEXEC, to ensure that the same pathname on two different filesystems are tracked separately.

I wonder if we should do somethiing like hash the `spa_guid` + `spa_bootenv` combo to derive a stable `st_dev`.

sgerraty:
>> They do if doing LOADER_VERIEXEC, to ensure that the same pathname on two di
>fferent filesystems are tracked separately.
>
>I wonder if we should do somethiing like hash the `spa_guid` + `spa_bootenv` c
>ombo to derive a stable `st_dev`.

If they are available and easily accessible.
The right answer tends to vary by filesystem.

Defenso-EBO:
I've reworked this based on the feedback from @VexedUXR, @sgerraty and
@kevans91 — thanks. The fix has moved from fstat.c into the loader's ZFS code.

Instead of zeroing st_dev/st_ino, the loader now populates them with
the same intrinsic identifiers the kernel uses:

- **st_dev** = the dataset's `ds_fsid_guid` — exactly what the kernel derives
  its fsid from (`dmu_objset_fsid_guid()` → `dsl_dataset_fsid_guid()` →
  `ds->ds_fsid_guid`). It was already read in `zfs_mount_dataset()` but
  discarded; it's now kept in `struct zfsmount`.
- **st_ino** = the object number resolved in `zfs_lookup()` (the loader's
  equivalent of the kernel's `zp->z_id` = `db->db_object`), kept in
  `struct file`.

`dev_t`/`ino_t` are 64-bit on FreeBSD, so both are assigned directly.
 A `memset()` at the top of `zfs_dnode_stat()` zeroes the
remaining fields so nothing holds stack garbage.

**Tested** on 16.0-CURRENT (amd64), ZFS-on-GELI root: rebuilt and re-signed the
EFI loader; the system boots under UEFI Secure Boot with `mac_veriexec` active.
 (The OS-level st_dev differs from the loader's value because the
kernel folds ds_fsid_guid through vn_fsid while the loader assigns it directly —
only intra-loader stability matters for veriexec.)

(Note: the branch is still named after the original fstat.c approach; the fix
itself has moved to the ZFS code.)

kevans91:
> I've reworked this based on the feedback from @VexedUXR, @sgerraty and @kevans91 — thanks. The fix has moved from fstat.c into the loader's ZFS code.
> 
> Instead of zeroing st_dev/st_ino, the loader now populates them with the same intrinsic identifiers the kernel uses:
> 
> * **st_dev** = the dataset's `ds_fsid_guid` — exactly what the kernel derives
>   its fsid from (`dmu_objset_fsid_guid()` → `dsl_dataset_fsid_guid()` →
>   `ds->ds_fsid_guid`). It was already read in `zfs_mount_dataset()` but
>   discarded; it's now kept in `struct zfsmount`.

I love it- I realized after writing that it was a bad idea, but didn't apparently get around to actually writing back here.

Defenso-EBO:
Hello @kevans91 do you know when this PR is going to be accepted ?

johannes-bruegmann:
Independent test of this patch on release hardware, with a setup that differs from
the one in the commit message — in particular veriexec in **strict** mode and
`/boot` on a separate ZFS pool.

**Tree:** `releng/15.1`, commit `aadd58dddcbc` (15.1-RELEASE-p2, as reported by
`uname`). Applied by cherry-picking `refs/pull/2271/head`; it needed no adjustment.
The two places where `main` and `releng/15.1` differ in these files are outside the
hunks (`DECODE_ERROR` vs `EFI_ERROR_CODE` in `zfs_module.c:106`, and
`if (path != NULL)` vs `if (mnt->path != NULL)` in `zfs.c:410`).

**Machine:** amd64, UEFI with Secure Boot enforcing and my own PK/KEK/db. Loader
built `WITH_LOADER_VERIEXEC` and `LOADER_VERIEXEC_VECTX`, signed with the db key.
`/boot` is a small ZFS pool on a removable microSD card; the root pool and a data
pool are GELI-encrypted (6486219 and 4390575 PKCS5v2 iterations). veriexec runs in
strict mode, i.e. `/boot/loader.ve.strict` sets `accept_no_fp = VE_WANT`.

### 1. It boots, and manifest entries are matched correctly

Without the fix this is precisely what failed here: files that *are* listed in the
manifest were reported as `Unverified ... no entry`, because `st_dev`/`st_ino` held
stack garbage. With the patch:

```
loader.ve.hashed = loader.lua,cli.lua,config.lua,hook.lua,core.lua,color.lua,password.lua,screen.lua,kernel,zroot.key,zroot.key,zfs.ko,geom_eli.ko
loader.ve.pcr    = 5c6e901f0605f57a69eac880702a2f3cd85dc9c2d9e4c815dc45a34184258832
```

13 files verified, no spurious failure, and the pseudo-PCR is computed.

### 2. A tampered file is still rejected

Uninitialised identity fields could in principle also make an entry match the
*wrong* file, and so accept a modified one. To exclude that, I appended a single
comment line to a file that is listed in the manifest:

```
Unverified /boot/lua/color.lua: 7f9dadab44ea37dcdc5735c12c632dc3951cceb2e3e9783540f0709e259c0346 != eb533bc685107bb631fd53905719708e1af5e131cb62c56e4e49c8ccf6e9e016
ERROR: /boot/lua/loader.lua:35: module 'color' not found
```

`eb533b…` is the manifest entry, `7f9dad…` the file after the edit; the second is
the sha256 of the first plus `"\n-- T2 tamper (PR 2271)\n"`, so both are
reproducible. The file is named, the two hashes differ, and the boot does not reach
a usable system — entries are matched to the file they belong to.

(Aside, not caused by this patch: for a Lua module the rejection is not fatal.
`severity_guess()` returns `VE_MUST` for `.lua`, but the `panic()` in
`verify_file.c:601` requires `severity > accept_no_fp`, and strict mode is switched
on by a file that `config.lua` reads — so it is not yet in force while the loader's
own Lua files are being verified. I intend to report that separately.)

### 3. It builds on releng/15.1

Worth stating because size limits in `stand/` are a classic blocker. I briefly
suspected this patch of overflowing `stand/i386/loader_lua`; **it does not**, and I
withdraw that. The overflow I saw came from an unrelated local change of my own:
with that change reverted and this patch applied, the loader has 3340 bytes left of
its `LOADERSIZE` budget of 500000.

### Also

Bug 297553, which I filed, is a duplicate of 295935 as referenced in the commit
message — feel free to close mine. The inline patch in my report is superseded by
this one and should not be applied.

Happy to test a rebased version, or to provide any other measurement from this
machine.

kevans91:
> (Aside, not caused by this patch: for a Lua module the rejection is not fatal.
severity_guess() returns VE_MUST for .lua, but the panic() in
verify_file.c:601 requires severity > accept_no_fp, and strict mode is switched
on by a file that config.lua reads — so it is not yet in force while the loader's
own Lua files are being verified. I intend to report that separately.)

I think what you want is a way to bake an elevated mode into the boot chain, right?
 
> Bug 297553, which I filed, is a duplicate of 295935 as referenced in the commit
message — feel free to close mine. The inline patch in my report is superseded by
this one and should not be applied.

Thanks for confirming, we'll just tag your PR in the commits we land from this one so that one can track the lifecycle of the change through there.

johannes-bruegmann:
Yes — that is exactly it, and you put it better than I did.

As I understand the model, strict has to be **in force before the first
must-verify file is opened**, or its guarantee is void for precisely the files it
exists for. Is that right, or have I misunderstood the intent? On my machine it is
not in force by then, and I think that part is provable rather than merely
observed.

**The circularity.** `verify_tweak()` fires when a file whose name begins with
`loader.ve.` is verified, and the only route I could find to get such a file read is
to list it in `loader_conf_files` — i.e. `config.lua` reads it. But `config.lua` is
itself a `VE_MUST` file, and it has to be verified and loaded before it can read
anything. So anything it reads necessarily takes effect *after* the loader's own Lua
files have been verified. Not a timing accident: a cycle.

**The measurement.** With `/boot/loader.ve.strict` in place and veriexec otherwise
working, I appended one comment line to `/boot/lua/color.lua`, a file listed in the
manifest:

```
Unverified /boot/lua/color.lua: 7f9dadab… != eb533bc6…
ERROR: /boot/lua/loader.lua:35: module 'color' not found
```

To be clear about what is and is not surprising here: the second column is the
**default** mode behaving as designed, and I do not think that is wrong. What I am
reporting is that I had configured strict and it had not taken effect yet.

|                           | strict, which I had configured | what actually took effect                        |
|---------------------------|--------------------------------|--------------------------------------------------|
| listed `.lua`, hash wrong | `panic("cannot continue")`     | rejected, boot continues into a Lua module error |

The absence of the panic is what proves the mode: `verify_file.c:601` requires
`rc == VE_FINGERPRINT_WRONG && severity > accept_no_fp`. Here `rc` was
`VE_FINGERPRINT_WRONG` (two hashes shown, not "no entry") and `severity_guess()`
gives `.lua` `VE_MUST` = 2. No panic means `2 > accept_no_fp` was false, so
`accept_no_fp` was still ≥ `VE_MUST`. Strict would have made it `VE_WANT`.

**Why it matters beyond the missing panic.** `loader.lua` aborted at its `require`
for that module, before `config.load()`. So `loader.conf` was never processed: no
`password`, no `bootlock_password`, no menu — I was dropped at an `OK` prompt. On a
machine where `/boot` is on removable media, one modified Lua file is therefore
enough to reach a prompt with no password, without touching anything that is signed.
(What did survive was a local addition of my own that runs from C before the
interpreter starts, which is what made me look at the ordering in the first place.)

Shapes I can imagine — I have no standing to choose between them:

- a build-time knob, the counterpart of `LOADER_VERIEXEC_TESTING`, setting
  `ACCEPT_NO_FP_DEFAULT` to `VE_WANT` instead of weakening it;
- opening the tweak file from C before the interpreter starts, so the mode is in
  force for everything the interpreter then touches;
- deriving it from the platform: if the firmware launched us with Secure Boot
  enforcing, elevate automatically — which also removes the need for a file whose
  own verification is what switches the mode on.

I am not much use for the design side of this, but if a patch turns up I can run it:
I build this loader from source anyway, and the tamper case takes a few minutes to
reproduce, so I could at least report back on whether the failure becomes fatal at
the right moment.

And thanks for the duplicate handling — nothing further needed from me there.

kevans91:

kevans91:
> deriving it from the platform: if the firmware launched us with Secure Boot
enforcing, elevate automatically — which also removes the need for a file whose
own verification is what switches the mode on.

I think it'll end up being this at a minimum.  A build knob also makes sense, though, given that you already have to use one to build a veriexec-enabled boot chain.

johannes-bruegmann:
Thanks for merging, and for the point about the knob — you are right that it costs
nothing in usability, since a veriexec boot chain already requires one.

If it would help rather than duplicate work you already have in flight, I could
attempt the elevation part and send it as a PR. I read `SecureBoot` and `SetupMode`
from the loader here already, amongst other things, so the awkward part — getting at
them before the interpreter starts — is something I have working code for.

I should be honest about the limits of that offer, though: I can share what I have,
but I cannot really judge whether it would be useful to you. I only have my own
situation in view and no feel for what fits the tree, and it was written with AI
assistance. Smaller refactorings are no problem at all; being told the shape is
wrong is equally fine.

Either way I can test: this machine reproduces the case in a few minutes — modify a
manifest-listed `.lua` and expect the panic at the right moment instead of a Lua
module error.

Just say which you prefer; I would rather ask than send something unwanted.

kevans91:
> If it would help rather than duplicate work you already have in flight, I could
attempt the elevation part and send it as a PR. I read SecureBoot and SetupMode
from the loader here already, amongst other things, so the awkward part — getting at
them before the interpreter starts — is something I have working code for.

Can you toss up a PR and tag @bsdimp, @sgerraty, and myself, please?  I think that's a fine start.  We probably want to do it based on these even before we'd load loader.env, if that's not already what you're doing.

sgerraty:
>Yes =E2=80=94 that is exactly it, and you put it better than I did.
>
>As I understand the model, strict has to be **in force before the first
>must-verify file is opened**, or its guarantee is void for precisely the =
>files it
>exists for. Is that right, or have I misunderstood the intent? On my mach=
>ine it is
>not in force by then, and I think that part is provable rather than merel=
>y
>observed.

Yes, unless you have strict enabled very early, then any tweaks to mode
apply only after they are set.

>**The circularity.** `verify_tweak()` fires when a file whose name begins=
> with
>`loader.ve.` is verified, and the only route I could find to get such a f=
>ile read is
>to list it in `loader_conf_files` =E2=80=94 i.e. `config.lua` reads it. B=
>ut `config.lua` is
>itself a `VE_MUST` file, and it has to be verified and loaded before it c=
>an read
>anything. So anything it reads necessarily takes effect *after* the loade=
>r's own Lua
>files have been verified. Not a timing accident: a cycle.

The loader.ve. tweak files may be of limited use on a normal FreeBSD
install.  Indeed trying to support loader veriexec in a single /boot
directory would be very difficult.


In (Junos) we support multiple sets of packages being present, and the
loader is configured to look at for example
/packages/sets/active/boot which contains symlinks to the boot/ subdir
of packages which have things to contribute to the loader.

Thus each individual package can contribute its own modules,
loader.conf, init.4th etc (we cannot use LUA yet as x86 BIOS loader
cannot fit both LUA and VERIEXEC, and we still have to support those)
which can be separately verified by per package manifests etc.

eg. loader-ve-strict.tgz contains
drwxr-xr-x  0 root   wheel       0 Aug  4 12:20 boot/
-rw-r--r--  0 root   wheel     668 Aug  7  2024 boot/init.4th
-rw-r--r--  0 root   wheel      25 Aug  7  2024 boot/loader.ve.strict
-rw-r--r--  0 root   wheel     327 Aug  4 12:20 manifest
-rw-r--r--  0 root   wheel    6480 Jun 30 15:25 manifest.ecerts
-rw-r--r--  0 root   wheel     284 Aug  4 12:20 manifest.esig
-rw-r--r--  0 root   wheel     899 Aug  4 12:20 package.xml

The files above do not live in /boot but in
/packages/db/loader-ve-strict-$VERSION and
/packages/sets/active/loader-ve -> /packages/db/loader-ve-strict-$VERSION 
/packages/sets/active/boot/loader-ve -> /packages/db/loader-ve-strict-$VERSION/boot

The init.4th exists simply to read boot/loader.ve.strict
and it is verifying that path is what communicates control to the loader.
loader.ve.strict itself just contains a comment.

We use init.4th because that's always required to be verified.
Using a loader.conf would be simpler, but we allow a loader.conf to be
unverified so not good enough.

In case you are wondering; the link name loader-ve above allows packages
like loader-ve-{strict,try,...} to all be mutually exclusive.

johannes-bruegmann:
Follow-up: I opened #2377 with the read moved into C -- in interact(), before
interp_preinit() and interp_init().

@sgerraty this is your init.4th mechanism transposed. loader_lua has no equally
early verified file (loader.lua is itself the first one), so the read had to
happen before the interpreter starts. That was the piece I was missing, thank
you.

Measured on 15.1-RELEASE-p2, with both older routes removed so the result cannot
be ambiguous: a tampered file under /boot/lua now panics, where before it dropped
to an unrestricted prompt -- loader.lua aborts at its require, config.load()
never runs, so loader.conf and the password in it never happen.

@kevans91 the ordering question you raised is answered in the PR: it cannot sit
before loader.env, because ve_efi_init() and find_currdev() both run after it,
and it does not need to, because "loader.ve." is already a restricted prefix by
then.

kevans91:
> @kevans91 the ordering question you raised is answered in the PR: it cannot sit
before loader.env, because ve_efi_init() and find_currdev() both run after it,
and it does not need to, because "loader.ve." is already a restricted prefix by
then.

Yes, sorry- I meant specifically that something keying off of whether we're operating under SecureBoot should happen that early.  I just misunderstood your references to `SecureBoot` and `SetupMode`, it seems.

johannes-bruegmann:
That was my wording, not your reading -- I put two different things in the same
comment. SecureBoot and SetupMode belong to a separate piece of code, not to this
PR.

That piece does sit that early: it runs immediately before parse_args() in
efi_main(), ahead of read_loader_env(). It reads SecureBoot and SetupMode from the
EFI global GUID via efi_global_getenv() and records what it found as loader
variables. It measures only.

What led me there: veriexec verifies everything it loads, but nothing checks the
root of its own chain -- SecureBoot and SetupMode do not appear anywhere under
stand/ today. On a machine where Secure Boot was switched off in setup, a loader
built WITH_LOADER_VERIEXEC still reports a fully verified boot.

I would like to offer it separately rather than here, and in whatever form suits
you -- PR, bug report, or a list. On the harder half, what should happen when a
measurement comes back wrong, I have no settled view and would rather hear yours.

sgerraty:
>What led me there: veriexec verifies everything it loads, but nothing checks t
>he
>root of its own chain -- SecureBoot and SetupMode do not appear anywhere under
>stand/ today. On a machine where Secure Boot was switched off in setup, a load
>er
>built WITH_LOADER_VERIEXEC still reports a fully verified boot.

That's because verification of the loader is the responsibility of the
firmware (BIOS, UEFI, UBOOT etc).  The loader does not know if it has
been verified, though if the system is doing secure boot, it can infer
from the fact that it is running that it has been verified successfully
;-)

PR #2272, opened 2026-06-09
hwpmc: Disable AMD PMCs if in an unsupported VM

mashtizadeh:
AMD does not have a CPUID bit to indicate the lack of K8 PMCs.  If all other PMC features are not present we should test an event selector to see if it stores and returns a value.  If the VM is implemented correctly, this should result in a #GP on the initial wrmsr_safe.  Bhyve and a few other VMs ignore writes, so I got one step further and test that it retains the OS and USR bits.

Tested on Zen 5 native and a Zen 5 Bhyve virtual machine.  This code should not run on any recent hardware, except in a VM, as it checks that the core counter extension is missing.

Reported by: Sandipan Das
Sponsored by: Netflix

### DIFF
--- a/sys/dev/hwpmc/hwpmc_amd.c
+++ b/sys/dev/hwpmc/hwpmc_amd.c
@@ -869,6 +869,8 @@ amd_pcpu_fini(struct pmc_mdep *md, int cpu)
 struct pmc_mdep *
 pmc_amd_initialize(void)
 {
+	int status;
+	uint64_t reg;
 	struct pmc_classdep *pcd;
 	struct pmc_mdep *pmc_mdep;
 	enum pmc_cputype cputype;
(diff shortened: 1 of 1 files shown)


mashtizadeh:
@bsdimp @mhorne

bsdimp (approved):

sys/dev/hwpmc/hwpmc_amd.c
```diff
@@ -869,6 +869,8 @@ amd_pcpu_fini(struct pmc_mdep *md, int cpu)
 struct pmc_mdep *
 pmc_amd_initialize(void)
 {
+	int status;
+	uint64_t reg;
 	struct pmc_classdep *pcd;
 	struct pmc_mdep *pmc_mdep;
 	enum pmc_cputype cputype;
```
mhorne, line 876: There are some difficult style(9) rules about variable declaration order here, roughly "larger comes first". I will make an adjustment before merging.

mhorne (approved):
LGTM. Thank you!

This is also [PR 268943](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=268943).

sys/dev/hwpmc/hwpmc_amd.c
mashtizadeh: Ah I forgot about that I will keep that in mind for the next patch.

PR #2273, opened 2026-06-09
fts: add fts_openat() API

Jeetubhati:
fts: add fts_openat() API

Add `fts_openat(int dirfd, char * const *argv, int options, int (*compar)(...))` as a new entry point for fts(3).

`fts_open()` is now a thin wrapper calling `fts_openat(AT_FDCWD, ...)`. When `dirfd` is `AT_FDCWD` the behaviour is identical to `fts_open()`. Passing a pre-opened directory fd allows fts traversal inside Capsicum capability mode where path-based operations are not permitted after `cap_enter(2)`.

**Implementation:**

- `fts_open()` becomes a one-liner wrapper calling `fts_openat(AT_FDCWD, ...)`
- `fts_openat()` `dup()`s the caller's `dirfd` and passes it as `rootfd` to `__fts_open()`; `AT_FDCWD` is passed through as-is since it is not a real file descriptor
- When a real `dirfd` is provided, root-level entries have `fts_dirfd` set to a dup of the caller's `dirfd`, making `openat(ent->fts_dirfd, ent->fts_name)` work consistently at all directory levels — not just non-root entries
- When `rootfd != AT_FDCWD`, `__fts_open()` uses it directly as `fts_rfd` instead of opening `.` — eliminating a redundant dup
- `fts_openat` exported in `FBSD_1.9` version map

**Tests:**

`fts_openat_test.c` with 1 test case:

- `atfdcwd_matches_fts_open` — verifies `fts_openat(AT_FDCWD, ...)` produces identical results to `fts_open()` by comparing entry count, `fts_info`, and `fts_name` for every entry

Depends on: PR #2303

### DIFF
--- a/include/fts.h
+++ b/include/fts.h
@@ -1,4 +1,4 @@
-/*-
+/*
  * SPDX-License-Identifier: BSD-3-Clause
  *
  * Copyright (c) 1989, 1993
@@ -147,6 +147,8 @@ void	*fts_get_clientptr(FTS *);
 FTS	*fts_get_stream(FTSENT *);
 #define	 fts_get_stream(ftsent)	((ftsent)->fts_fts)
 FTS	*fts_open(char * const *, int,
+	    int (*)(const FTSENT * const *, const FTSENT * const *));
+FTS	*fts_openat(int, char * const *, int,
 	    int (*)(const FTSENT * const *, const FTSENT * const *));
 #ifdef __BLOCKS__
 FTS	*fts_open_b(char * const *, int,
--- a/lib/libc/gen/fts.3
+++ b/lib/libc/gen/fts.3
@@ -25,7 +25,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd August 9, 2026
+.Dd August 13, 2026
 .Dt FTS 3
 .Os
 .Sh NAME
@@ -37,6 +37,8 @@
 .In fts.h
 .Ft FTS *
 .Fn fts_open "char * const *path_argv" "int options" "int (*compar)(const FTSENT * const *, const FTSENT * const *)"
+.Ft FTS *
+.Fn fts_openat "int dirfd" "char * const *path_argv" "int options" "int (*compar)(const FTSENT * const *, const FTSENT * const *)"
 .Ft FTS *
 .Fn fts_open_b "char * const *path_argv" "int options" "int (^compar)(const FTSENT * const *, const FTSENT * const *)"
 .Ft FTSENT *
@@ -621,6 +623,36 @@ except that it takes a block pointer instead of a function pointer.
 The block is copied before
 .Fn fts_open_b
 returns, so the original can safely go out of scope or be released.
+.Ss Fn fts_openat
+The
+.Fn fts_openat
+function is identical to
+.Fn fts_open
+except that it accepts a file descriptor
+.Fa dirfd
+as its first argument.
+If
+.Fa dirfd
+is
+.Dv AT_FDCWD ,
+the behaviour is identical to
+.Fn fts_open .
+Otherwise,
+.Fa dirfd
+must be an open file descriptor referring to a directory,
+and the traversal is rooted there.
+This allows
+.Fn fts_openat
+to be used inside Capsicum capability mode
+.Pq Xr capsicum 4 ,
+where path-based operations are not permitted.
+The
+.Fn fts_openat
+function duplicates
+.Fa dirfd
+internally, so the caller may close it after
+.Fn fts_openat
+returns.
 .Ss Fn fts_read
 The
 .Fn fts_read
--- a/lib/libc/gen/fts.c
+++ b/lib/libc/gen/fts.c
@@ -155,6 +155,10 @@ __fts_open(FTS *sp, char * const *argv, int rootfd)
 	if ((parent = fts_alloc(sp, "", 0)) == NULL)
 		goto mem2;
 	parent->fts_level = FTS_ROOTPARENTLEVEL;
+	parent->fts_dirfd = AT_FDCWD;
+
+	if (rootfd != AT_FDCWD)
+		parent->fts_dirfd = rootfd;
 
 	/* Shush, GCC. */
 	tmp = NULL;
@@ -169,11 +173,20 @@ __fts_open(FTS *sp, char * const *argv, int rootfd)
 		p->fts_accpath = p->fts_name;
 		p->fts_info = fts_stat(sp, p,
 		    ISSET(FTS_COMFOLLOWDIR) ? -1 : ISSET(FTS_COMFOLLOW),
-		    -1);
+		    rootfd == AT_FDCWD ? -1 : rootfd);
 
 		/* Command-line "." and ".." are real directories. */
 		if (p->fts_info == FTS_DOT)
 			p->fts_info = FTS_D;
+		if (p->fts_info == FTS_D) {
+			if (strcmp(p->fts_name, ".") == 0)
+				p->fts_dirfd = rootfd != AT_FDCWD ?
+				    _dup(rootfd) :
+				    _open(".", O_RDONLY | O_DIRECTORY | O_CLOEXEC);
+			else
+				p->fts_dirfd = _openat(rootfd, p->fts_name,
+				    O_RDONLY | O_DIRECTORY | O_CLOEXEC);
+		}
 
 		/*
 		 * If comparison routine supplied, traverse in sorted
@@ -213,10 +226,16 @@ __fts_open(FTS *sp, char * const *argv, int rootfd)
 	 * descriptor we run anyway, just more slowly.  We use _openat rather
 	 * than _dup because rootfd may be AT_FDCWD, not a real descriptor.
 	 */
-	if (!ISSET(FTS_NOCHDIR) &&
-	    (sp->fts_rfd = _openat(rootfd, ".", O_RDONLY |
-	    O_CLOEXEC, 0)) < 0)
-		SET(FTS_NOCHDIR);
+
+	if (!ISSET(FTS_NOCHDIR)) {
+		if (rootfd != AT_FDCWD)
+			sp->fts_rfd = _dup(rootfd);
+		if (sp->fts_rfd < 0)
+			SET(FTS_NOCHDIR);
+		else if ((sp->fts_rfd =
+			_open(".", O_RDONLY | O_CLOEXEC, 0)) < 0)
+			SET(FTS_NOCHDIR);
+	}
 	return (sp);
 
 mem3:	fts_lfree(root);
@@ -230,8 +249,16 @@ FTS *
 fts_open(char * const *argv, int options,
     int (*compar)(const FTSENT * const *, const FTSENT * const *))
 {
+	return (fts_openat(AT_FDCWD, argv, options, compar));
+}
+
+FTS *
+fts_openat(int dirfd, char * const *argv, int options,
+    int (*compar)(const FTSENT * const *, const FTSENT * const *))
+{
 	struct _fts_private *priv;
 	FTS *sp;
+	int rootfd;
 
 	/* Options check. */
 	if (options & ~FTS_OPTIONMASK) {
@@ -239,7 +266,7 @@ fts_open(char * const *argv, int options,
 		return (NULL);
 	}
 
-	/* fts_open() requires at least one path */
+	/* fts_openat() requires at least one path */
 	if (*argv == NULL) {
 		errno = EINVAL;
 		return (NULL);
@@ -251,10 +278,15 @@ fts_open(char * const *argv, int options,
 	sp = &priv->ftsp_fts;
 	sp->fts_compar = compar;
 	sp->fts_options = options;
+	if (dirfd == AT_FDCWD)
+		rootfd = AT_FDCWD;
+	else if ((rootfd = _dup(dirfd)) < 0) {
+		free(priv);
+		return (NULL);
+	}
+	return (__fts_open(sp, argv, rootfd));
 
-	return (__fts_open(sp, argv, AT_FDCWD));
 }
-
 #ifdef __BLOCKS__
 FTS *
 fts_open_b(char * const *argv, int options,
@@ -781,7 +813,16 @@ fts_build(FTS *sp, int type)
 		oflag = DTF_NODUP;
 	else
 		oflag = DTF_HIDEW | DTF_NODUP;
-	if ((dirp = __opendir2(cur->fts_accpath, oflag)) == NULL) {
+	if (cur->fts_dirfd >= 0) {
+		int fd;
+		fd = _openat(cur->fts_dirfd, ".", O_RDONLY | O_DIRECTORY | O_CLOEXEC);
+		if (fd >= 0)
+			dirp = fdopendir(fd);
+		else
+			dirp = __opendir2(cur->fts_accpath, oflag);
+	} else
+		dirp = __opendir2(cur->fts_accpath, oflag);
+	if (dirp == NULL) {
 		if (type == BREAD) {
 			cur->fts_info = FTS_DNR;
 			cur->fts_errno = errno;
@@ -789,7 +830,8 @@ fts_build(FTS *sp, int type)
 		return (NULL);
 	}
 
-	cur->fts_dirfd = _dup(_dirfd(dirp));
+	if (cur->fts_dirfd < 0)
+		cur->fts_dirfd = _dup(_dirfd(dirp));
 
 	/*
 	 * In the FTS_PHYSICAL | FTS_NOSTAT case, we want to avoid calling
@@ -927,8 +969,12 @@ mem1:				saved_errno = errno;
 
 		p->fts_level = level;
 		p->fts_parent = sp->fts_cur;
+		if (dp->d_type == DT_DIR) {
+			p->fts_dirfd = _openat(_dirfd(dirp),
+			p->fts_name,
+			O_RDONLY | O_DIRECTORY | O_CLOEXEC);
+		}
 		p->fts_pathlen = len + dnamlen;
-
 		if (dp->d_type == DT_WHT)
 			p->fts_flags |= FTS_ISW;
 
--- a/lib/libc/tests/gen/Makefile
+++ b/lib/libc/tests/gen/Makefile
@@ -16,6 +16,7 @@ ATF_TESTS_C+=		fts_blocks_test
 ATF_TESTS_C+=		fts_children_test
 ATF_TESTS_C+=		fts_misc_test
 ATF_TESTS_C+=		fts_open_test
+ATF_TESTS_C+=		fts_openat_test
 ATF_TESTS_C+=		fts_options_test
 ATF_TESTS_C+=		fts_regress_test
 ATF_TESTS_C+=		fts_set_test
--- a/lib/libc/tests/gen/fts_openat_test.c
+++ b/lib/libc/tests/gen/fts_openat_test.c
@@ -1,0 +1,250 @@
+/*
+ * Copyright (c) 2026 Jitendra Bhati
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+/*
+ * Basic tests for fts_openat().  When called with AT_FDCWD the
+ * behaviour must be identical to fts_open().
+ */
+
+#include <sys/stat.h>
+
+#include <limits.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <fcntl.h>
+#include <errno.h>
+#include <fts.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/capsicum.h>
+
+#include <atf-c.h>
+
+#define	FTS_TEST_MAXENTRIES 64
+
+static int
+fts_lexical_compar(const FTSENT * const *a, const FTSENT * const *b)
+{
+	return (strcmp((*a)->fts_name, (*b)->fts_name));
+}
+
+/*
+ * fts_openat(AT_FDCWD, ...) must behave identically to fts_open().
+ */
+ATF_TC(atfdcwd_matches_fts_open);
+ATF_TC_HEAD(atfdcwd_matches_fts_open, tc)
+{
+	atf_tc_set_md_var(tc, "descr",
+	    "fts_openat(AT_FDCWD) behaves identically to fts_open");
+}
+
+ATF_TC_BODY(atfdcwd_matches_fts_open, tc)
+{
+	char *cwd, *abspath;
+	char *paths[2];
+	FTS *fts;
+	FTSENT *ent;
+
+	int *info1, *info2;
+	char (*names1)[NAME_MAX + 1], (*names2)[NAME_MAX + 1];
+	int n1, n2, i;
+	
+	ATF_REQUIRE((info1 = malloc(FTS_TEST_MAXENTRIES *
+            sizeof(*info1))) != NULL);
+        ATF_REQUIRE((info2 = malloc(FTS_TEST_MAXENTRIES *
+            sizeof(*info2))) != NULL);
+        ATF_REQUIRE((names1 = malloc(FTS_TEST_MAXENTRIES *
+            sizeof(*names1))) != NULL);
+        ATF_REQUIRE((names2 = malloc(FTS_TEST_MAXENTRIES *
+            sizeof(*names2))) != NULL);
+
+	cwd = malloc(PATH_MAX);
+	ATF_REQUIRE(cwd != NULL);
+	abspath = malloc(PATH_MAX * 2);
+	ATF_REQUIRE(abspath != NULL);
+
+	ATF_REQUIRE(getcwd(cwd, PATH_MAX) != NULL);
+	ATF_REQUIRE_EQ(0, mkdir("dir", 0755));
+	ATF_REQUIRE_EQ(0, mkdir("dir/sub", 0755));
+	ATF_REQUIRE_EQ(0, close(creat("dir/sub/file", 0644)));
+	ATF_REQUIRE_EQ(0, close(creat("dir/other", 0644)));
+
+	snprintf(abspath, PATH_MAX * 2, "%s/dir", cwd);
+	paths[0] = abspath;
+	paths[1] = NULL;
+
+	/* Collect fts_open results. */
+	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL,
+	    fts_lexical_compar)) != NULL);
+	for (n1 = 0;
+	    (ent = fts_read(fts)) != NULL && n1 < FTS_TEST_MAXENTRIES;
+	    n1++) {
+		info1[n1] = ent->fts_info;
+		strlcpy(names1[n1], ent->fts_name, NAME_MAX + 1);
+	}
+	ATF_REQUIRE_EQ_MSG(0, fts_close(fts), "fts_close: %m");
+
+	/* Collect fts_openat results. */
+	ATF_REQUIRE((fts = fts_openat(AT_FDCWD, paths, FTS_PHYSICAL,
+	    fts_lexical_compar)) != NULL);
+	for (n2 = 0;
+	    (ent = fts_read(fts)) != NULL && n2 < FTS_TEST_MAXENTRIES;
+	    n2++) {
+		info2[n2] = ent->fts_info;
+		strlcpy(names2[n2], ent->fts_name, NAME_MAX + 1);
+	}
+	ATF_REQUIRE_EQ_MSG(0, fts_close(fts), "fts_close: %m");
+
+	/* Compare. */
+	ATF_CHECK_EQ_MSG(n1, n2,
+	    "entry count mismatch: fts_open=%d fts_openat=%d", n1, n2);
+	for (i = 0; i < n1 && i < n2; i++) {
+		ATF_CHECK_EQ_MSG(info1[i], info2[i],
+		    "fts_info mismatch at entry %d: "
+		    "fts_open=%d fts_openat=%d name=%s",
+		    i, info1[i], info2[i], names1[i]);
+		ATF_CHECK_STREQ_MSG(names1[i], names2[i],
+		    "fts_name mismatch at entry %d: "
+		    "fts_open='%s' fts_openat='%s'",
+		    i, names1[i], names2[i]);
+	}
+
+	free(cwd);
+	free(abspath);
+	free(info1);
+	free(info2);
+	free(names1);
+	free(names2);
+}
+
+/*
+ * fts_openat() with a real dirfd must work in Capsicum capability mode.
+ */
+ATF_TC(openat_capsicum);
+ATF_TC_HEAD(openat_capsicum, tc)
+{
+	atf_tc_set_md_var(tc, "descr",
+	    "fts_openat() with dirfd works in Capsicum capability mode");
+}
+ATF_TC_BODY(openat_capsicum, tc)
+{
+	char *paths[] = { ".", NULL };
+	FTS *fts;
+	FTSENT *ent;
+	int dirfd;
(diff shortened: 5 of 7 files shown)


lib/libc/tests/gen/fts_openat_test.c
```diff
@@ -0,0 +1,132 @@
...
+ATF_TC(atfdcwd_matches_fts_open);
+ATF_TC_HEAD(atfdcwd_matches_fts_open, tc)
+{
+	atf_tc_set_md_var(tc, "descr",
+	    "fts_openat(AT_FDCWD) behaves identically to fts_open");
+}
+ATF_TC_BODY(atfdcwd_matches_fts_open, tc)
+{
+	char cwd[PATH_MAX];
+	char abspath[PATH_MAX * 2];
```
asomers: Better to put such large arrays on the heap than the stack.

lib/libc/tests/gen/fts_openat_test.c
```diff
@@ -0,0 +1,132 @@
...
+	/* Collect fts_openat results. */
+	ATF_REQUIRE((fts = fts_openat(AT_FDCWD, paths, FTS_PHYSICAL,
+	    NULL)) != NULL);
+	n2 = 0;
+	while ((ent = fts_read(fts)) != NULL && n2 < 32) {
+		info2[n2] = ent->fts_info;
+		strlcpy(names2[n2], ent->fts_name, sizeof(names2[0]));
+		n2++;
+	}
+	ATF_REQUIRE_EQ_MSG(0, fts_close(fts), "fts_close: %m");
```
asomers, line 100: Instead of running fts_open, reading the results into an array, then running fts_openat, reading the results into an array, and then comparing the two arrays, why not run the two in lockstep?  Something like this?
```
fts0 = fts_open(...);
fts1 = fts_openat(AT_FDCWD, ...);
while {
    ent0 = fts_read(fts0);
    ent1 = fts_read(fts1);
    ATF_REQUIRE((ent0 == NULL) ^ !(ent1 == NULL));
    if (!ent0)
        break;
    ATF_REQUIRE(each field of ent0 matches ent1);
}
```

lib/libc/gen/fts.c
```diff
@@ -249,10 +255,41 @@ fts_open(char * const *argv, int options,
...
+        }
+
+        /* Allocate/initialize the stream. */
+        if ((priv = calloc(1, sizeof(*priv))) == NULL)
+                return (NULL);
+        sp = &priv->ftsp_fts;
+        sp->fts_compar = compar;
+        sp->fts_options = options;
+        priv->ftsp_rootfd = dirfd;
+
```
asomers: All of this stuff is duplicated between fts_open and fts_openat.  Instead of duplicating code, why not just make fts_open call fts_openat(AT_FDCWD)?  Isn't that what we'd discussed?

lib/libc/gen/fts.c
```diff
@@ -211,9 +212,14 @@ __fts_open(FTS *sp, char * const *argv)
 	 * and ".." are all fairly nasty problems.  Note, if we can't get the
 	 * descriptor we run anyway, just more slowly.
 	 */
-	if (!ISSET(FTS_NOCHDIR) &&
-	    (sp->fts_rfd = _open(".", O_RDONLY | O_CLOEXEC, 0)) < 0)
-		SET(FTS_NOCHDIR);
+	if (!ISSET(FTS_NOCHDIR)) {
+                struct _fts_private *priv = (struct _fts_private *)sp;
+                if (priv->ftsp_rootfd >= 0)
+                        sp->fts_rfd = priv->ftsp_rootfd;
```
asomers: You can't simply copy ftsp_rootfd here, because fts_close will close() it.  And we don't want to close the file descriptor provided by our caller; that's not part of the contract.  Instead, you should call `dup()`.

asomers (changes requested):
This is a good start, but we can't merge it until the function is actually capable of doing something.  Didn't the GSoC schedule call for doing the internal refactor of fts_open to eliminate global namespace access first, so that the addition of fts_openat later would be quick and easy?

Jeetubhati:
> This is a good start, but we can't merge it until the function is actually capable of doing something. Didn't the GSoC schedule call for doing the internal refactor of fts_open to eliminate global namespace access first, so that the addition of fts_openat later would be quick and easy?

I got the ordering wrong. Will do the internal refactor first so that fts_openat actually works in capability mode, then fts_openat becomes a thin wrapper on top.

lib/libc/tests/gen/fts_openat_test.c
Jeetubhati: Will move cwd and abspath to heap with malloc().

Jeetubhati:
> internal refactor of fts_open to eliminate global namespace a

I replaced all _open(".", ...) calls with _openat(ROOTFD(sp), ".", ...)
in __fts_open(), fts_read(), and fts_children(). Added FTSP() and ROOTFD()
macros for clean access to ftsp_rootfd.

lib/libc/gen/fts.c
Jeetubhati: fts_open() is now a one-liner calling fts_openat(AT_FDCWD, ...). as we'd decided.

lib/libc/gen/fts.c
Jeetubhati: fts_openat() calls _dup(dirfd) so fts_close() does not close a fd the caller still owns.

lib/libc/tests/gen/fts_openat_test.c
Jeetubhati: Attempted lockstep but it fails in ATF because fts_open uses fchdir internally which changes the process CWD, causing the second traversal to see a different AT_FDCWD. Using sequential traversal instead with fts_lexical_compar for deterministic comparison.

Jeetubhati: fts_safe_changedir still uses _open(path, ...) rather than _openat, it operates relative to the current traversal position, not the root fd. Fixing this properly requires threading the current directory fd through the call chain, which is the next step along with fts_capsicum_test.c.

lib/libc/tests/gen/fts_openat_test.c
Jeetubhati: Done.

lib/libc/gen/fts.c
```diff
@@ -472,19 +452,20 @@ fts_read(FTS *sp)
...
-				p->fts_errno = errno;
-				p->fts_info = FTS_ERR;
-			} else
-				p->fts_flags |= FTS_SYMFOLLOW;
-		}
-		return (p);
-	}
+            (p->fts_info == FTS_SL || p->fts_info == FTS_SLNONE)) {
+                p->fts_info = fts_stat(sp, p, 1, -1);
+                if (p->fts_info == FTS_D && !ISSET(FTS_NOCHDIR)) {
```
asomers: Something's wrong with the indentation here.  It  looks like you expanded tabs into spaces.

lib/libc/gen/fts.c
```diff
@@ -569,7 +550,8 @@ next:	tmp = p;
 			p->fts_info = fts_stat(sp, p, 1, -1);
 			if (p->fts_info == FTS_D && !ISSET(FTS_NOCHDIR)) {
 				if ((p->fts_symfd =
-				    _open(".", O_RDONLY | O_CLOEXEC, 0)) < 0) {
+                                    _openat(ROOTFD(sp), ".", O_RDONLY |
```
asomers: Indentation here, too.

asomers (changes requested):
This PR now does two different things: it internally refactors fts_open, and it adds fts_openat.  Please split it up into two separate PRs.

lib/libc/gen/fts.c
```diff
@@ -212,7 +213,7 @@ __fts_open(FTS *sp, char * const *argv)
 	 * descriptor we run anyway, just more slowly.
 	 */
 	if (!ISSET(FTS_NOCHDIR) &&
-	    (sp->fts_rfd = _open(".", O_RDONLY | O_CLOEXEC, 0)) < 0)
+	    (sp->fts_rfd = _openat(rootfd, ".", O_RDONLY | O_CLOEXEC, 0)) < 0)
```
asomers: This line is almost equivalent to `dup`, if `rootfd` isn't `AT_FDCWD`.  But down on line 263, we already called `dup`.  That's redundant.  I think you should eliminate one of the two dups.

lib/libc/gen/fts.c
```diff
@@ -249,8 +258,14 @@ fts_open(char * const *argv, int options,
...
 	sp->fts_options = options;
+	if (dirfd == AT_FDCWD)
+		rootfd = AT_FDCWD;
+	else if ((rootfd = _dup(dirfd)) < 0) {
+		free(priv);
+		return (NULL);
+	}
 
-	return (__fts_open(sp, argv));
+	return (__fts_open(sp, argv, rootfd));
```
asomers, line 287: `__fts_open` does not take ownership of `rootfd`.  So this line is a file descriptor leak.  To fix it, you should either eliminate the `openat` in line 216, or else eliminate the `_dup` on line 263.

lib/libc/tests/gen/fts_openat_test.c
```diff
@@ -0,0 +1,147 @@
...
+}
+
+ATF_TC_BODY(atfdcwd_matches_fts_open, tc)
+{
+	char *cwd, *abspath;
+	char *paths[2];
+	FTS *fts;
+	FTSENT *ent;
+	int info1[64], info2[64];
+	char names1[64][NAME_MAX + 1], names2[64][NAME_MAX + 1];
```
asomers: These two arrays sum to 32 kB.  You should allocate them on the heap, not the stack.  Or, you could even write this test in ATF-C++.  That would let you use an STL container here instead of a C array.  It's your choice.

lib/libc/tests/gen/fts_openat_test.c
```diff
@@ -0,0 +1,147 @@
...
+
+	snprintf(abspath, PATH_MAX * 2, "%s/dir", cwd);
+	paths[0] = abspath;
+	paths[1] = NULL;
+
+	/* Collect fts_open results. */
+	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL,
+	    fts_lexical_compar)) != NULL);
+	n1 = 0;
+	while ((ent = fts_read(fts)) != NULL && n1 < 64) {
```
asomers: Better to define a constant than to keep copying 64 everywhere.  Also, this while loop would be better written as a for loop.

lib/libc/tests/gen/fts_openat_test.c
```diff
@@ -0,0 +1,147 @@
...
+		info1[n1] = ent->fts_info;
+		strlcpy(names1[n1], ent->fts_name, sizeof(names1[0]));
```
asomers: While technically correct, this is a bit confusing.  And it's likely to confuse static analysis tools.  The size would be better written as `sizeof(names1[n1])`.  Or, just use `NAME_MAX+1`, which you'll have to do anyway if you switch to heap allocation.

lib/libc/tests/gen/fts_openat_test.c
```diff
@@ -0,0 +1,147 @@
...
+	atf_tc_set_md_var(tc, "descr",
+	    "fts_openat with empty argv returns NULL with EINVAL");
+}
+ATF_TC_BODY(empty_argv, tc)
+{
+	char *paths[] = { NULL };
+
+	ATF_REQUIRE_ERRNO(EINVAL,
+	    fts_openat(AT_FDCWD, paths, FTS_PHYSICAL, NULL) == NULL);
+}
```
asomers, line 242: I think it would be ok to eliminate these two tests.  It's enough to test this behavior with either `fts_open` or `fts_openat`.  You don't need to use both.

asomers (changes requested):

lib/libc/tests/gen/fts_openat_test.c
```diff
@@ -44,10 +46,20 @@ ATF_TC_BODY(atfdcwd_matches_fts_open, tc)
...
 	FTSENT *ent;
-	int info1[64], info2[64];
-	char names1[64][NAME_MAX + 1], names2[64][NAME_MAX + 1];
+
+	int *info1, *info2;
+	char (*names1)[NAME_MAX + 1], (*names2)[NAME_MAX + 1];
 	int n1, n2, i;
 
+	info1 = malloc(FTS_TEST_MAXENTRIES * sizeof(*info1));
+	ATF_REQUIRE(info1 != NULL);
```
asomers: As a matter of style, these lines could be written more succinctly like this:
```c
ATF_REQUIRE((info1 = malloc(...)) != NULL);
```

asomers (approved):
This LGTM (though I haven't tested it yet).  But I think that we shouldn't merge it until it's actually capable of running in capsicum mode.  That requires re-merging all of the stuff from PR #2278 , along with #2298 .

asomers:
cc @dag-erling

lib/libc/tests/gen/fts_openat_test.c
Jeetubhati: Fixed. Also noted that #2273 depends on the combined refactor PR landing first. I'll wait for that to merge before this one.

dag-erling:
I'm still waiting for the new dirfd patch, which should logically come before this one.

asomers:
> I'm still waiting for the new dirfd patch, which should logically come before this one.

See #2303 .

dag-erling (changes requested):
Please sort out the division between this and #2303 and clarify which order you intend them to go in.

Jeetubhati:
> Please sort out the division between this and #2303 and clarify which order you intend them to go in.

The intended merge order is:

1. PR #2303 first — internal refactor, `fts_dirfd` in `_fts_private`, `fts_get_dirfd()` accessor, man page update
2. PR #2273 second — `fts_openat()` API, built on top of #2303

PR #2303 is ready for review. Once it lands, #2273 will be rebased on top of it.

dag-erling:
I won't be able to review these until early next week, please hold off until then.

include/fts.h
```diff
@@ -148,6 +148,8 @@ FTS	*fts_get_stream(FTSENT *);
 #define	 fts_get_stream(ftsent)	((ftsent)->fts_fts)
 FTS	*fts_open(char * const *, int,
 	    int (*)(const FTSENT * const *, const FTSENT * const *));
+FTS     *fts_openat(int, char * const *, int,
```
asomers: The whitespace here looks like it doesn't match its neighbors.

lib/libc/gen/fts.c
```diff
@@ -167,6 +167,8 @@ __fts_open(FTS *sp, char * const *argv, int rootfd)
 		p->fts_level = FTS_ROOTLEVEL;
 		p->fts_parent = parent;
 		p->fts_accpath = p->fts_name;
+		if (rootfd != AT_FDCWD)
+                        p->fts_dirfd = _dup(rootfd);
```
asomers: More whitespace mismatch here.

lib/libc/tests/gen/Makefile
```diff
@@ -72,6 +70,9 @@ NETBSD_ATF_TESTS_C+=	cpuset_test
 NETBSD_ATF_TESTS_C+=	dir_test
 NETBSD_ATF_TESTS_C+=	floatunditf_test
 NETBSD_ATF_TESTS_C+=	fnmatch_test
+NETBSD_ATF_TESTS_C+=	fpclassify_test
+NETBSD_ATF_TESTS_C+=	fpsetmask_test
+NETBSD_ATF_TESTS_C+=	fpsetround_test
```
asomers: Why did these change from `ATF_TESTS_C` to `NETBSD_ATF_TESTS_C`?

asomers (changes requested):
This looks mostly good.  But it's hard to be sure I've reviewed it accurately while the changes from #2303 are also here.
Also, could you add a test that ensures fts_openat works in capability mode?

lib/libc/tests/gen/Makefile
Jeetubhati: The `fpclassify_test`, `fpsetmask_test` and `fpsetround_test` change to `NETBSD_ATF_TESTS_C` is from upstream main — not introduced by this PR. Our Makefile correctly inherits that change.

Jeetubhati:
> could you add a test that ensures fts_openat works in capability mode?

Added openat_capsicum test verifying` fts_openat()` works in Capsicum capability mode. The test passes when PR #2303 (`fd-relative fts_build changes`) is also installed — the two PRs are designed to work together.

include/fts.h
Jeetubhati: Fixed whitespace in include/fts.h and fts.c.

lib/libc/tests/gen/Makefile
asomers: Yes, it is introduced by this PR.  It's introduced by change 9465988a5a160f08c29c37abceb575e31e3309ba .  It appears that you made an error resolving a conflict during a merge or rebase operation.

Jeetubhati: Sorry about that i think it was introduced during a rebase conflict resolution. I fixed it now, `fpclassify_test`, `fpsetmask_test `and `fpsetround_test `are restored to their correct positions in `ATF_TESTS_C`.

lib/libc/tests/gen/fts_capsicum_test.c
```diff
@@ -0,0 +1,231 @@
...
+	
+	/*
+         * Open fts before entering capability mode — fts_openat
+         * will dup dirfd internally so we can close our copy.
+         */
+        ATF_REQUIRE((fts = fts_openat(dirfd, paths,
+            FTS_PHYSICAL | FTS_NOCHDIR, NULL)) != NULL);
+
+        /* Enter capability mode — no more path-based syscalls. */
+        ATF_REQUIRE_EQ(0, cap_enter());
```
asomers: Screwed up whitespace here.

lib/libc/tests/gen/fts_capsicum_test.c
```diff
@@ -0,0 +1,231 @@
...
+	ATF_REQUIRE_EQ(0, close(creat("dir/other", 0644)));
+
+	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL, NULL)) != NULL);
+
+	while ((ent = fts_read(fts)) != NULL) {
+		/* Root entry has no parent directory fd. */
+		if (ent->fts_level == FTS_ROOTLEVEL)
+			continue;
+		/* Post-order directory visit — same entry as pre-order. */
+		if (ent->fts_info == FTS_DP)
```
asomers: Why skip the assertions in directory post-order?  I still think it would be valuable to ensure that fts_dirfd is set correctly during post-order.

lib/libc/tests/gen/fts_capsicum_test.c
```diff
@@ -0,0 +1,231 @@
...
+
+	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL, NULL)) != NULL);
+
+	while ((ent = fts_read(fts)) != NULL) {
+		if (ent->fts_info == FTS_DP ||
+		    ent->fts_level == FTS_ROOTLEVEL)
+			continue;
+
+		if (ent->fts_dirfd < 0)
+			continue;
```
asomers: I think that you should assert that fts_dird >=0, rather than skip this portion of the test.

lib/libc/tests/gen/fts_capsicum_test.c
```diff
@@ -0,0 +1,231 @@
...
+		    "lstat failed for '%s': %m", ent->fts_accpath);
+
+		ATF_CHECK_EQ_MSG(sb_accpath.st_ino, sb_dirfd.st_ino,
+		    "inode mismatch at depth %ld for '%s'",
+		    ent->fts_level, ent->fts_name);
+
+		depth_checked++;
+	}
+
+	ATF_CHECK_MSG(depth_checked > 0, "must have checked at least one entry");
```
asomers: Surely you can be more specific than "> 0" ?

Jeetubhati: Addressed all review comments in new PR #2332  : 

- Fixed whitespace around `fts_openat `and `cap_enter `calls
- `fts_dirfd_valid `now checks `fts_dirfd `in post-order (`FTS_DP`)  entries as well
- Changed silent skip to `ATF_REQUIRE_MSG `for `fts_dirfd >= 0`
- `fts_dirfd_deep_tree `now asserts exactly 7 entries instead of > 0

Please review the new PR #2332

asomers (changes requested):
Something is screwed up here.  The review includes lots of files that shouldn't be changed.  Can you please rebase?

Jeetubhati:
Rebased cleanly on origin/main. Added `cur->fts_dirfd` initialization in `fts_build` for root directories.

asomers:
> Rebased cleanly on origin/main. Added `cur->fts_dirfd` initialization in `fts_build` for root directories.

Something's wrong.  There are conflicts between this PR and the main branch.  When you rebased, I don't think you rebased onto a recent enough version of the main branch.  Can you please rebase again?

lib/libc/gen/fts.3
```diff
@@ -25,11 +25,12 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd August 9, 2026
+.Dd Aug 12, 2026
```
asomers: Spell out the entire month; don't abbreviate.

lib/libc/gen/fts.3
```diff
@@ -25,11 +25,12 @@
...
 .Dt FTS 3
 .Os
 .Sh NAME
-.Nm fts
+.Nm fts ,
+.Nm fts_openat
```
asomers: No need to mention `fts_openat` here.  This isn't the place for function names.  It's the name of the man page.

lib/libc/gen/fts.3
```diff
@@ -621,6 +616,37 @@ except that it takes a block pointer instead of a function pointer.
...
+Otherwise,
+.Fa dirfd
+must be an open file descriptor referring to a directory,
+and the traversal is rooted there.
+This allows
+.Fn fts_openat
+to be used inside Capsicum capability mode
+.Pq Xr capsicum 4 ,
+where path-based operations are not permitted after
+.Xr cap_enter 2 .
```
asomers: "after cap_enter" is redundant with "inside Capsicum capability mode".  I would delete the "after cap_enter" part.

lib/libc/gen/fts.c
```diff
@@ -251,8 +265,14 @@ fts_open(char * const *argv, int options,
 	sp = &priv->ftsp_fts;
 	sp->fts_compar = compar;
 	sp->fts_options = options;
+	if (dirfd == AT_FDCWD)
+		rootfd = AT_FDCWD;
+	else if ((rootfd = _dup(dirfd)) < 0) {
```
asomers, line 283: I think there's a file descriptor leak here.  I don't see anything that will ever close `rootfd`.  It gets duped again in the loop at line 171, and those duped entries get freed at line 382, but `rootfd` itself does not get closed.

lib/libc/gen/fts.c
```diff
@@ -787,7 +803,8 @@ fts_build(FTS *sp, int type)
 		return (NULL);
 	}
 
-	cur->fts_dirfd = _dup(_dirfd(dirp));
+	if (cur->fts_dirfd < 0)
+		cur->fts_dirfd = _dup(_dirfd(dirp));
```
asomers, line 834: Are you sure this is still needed, now that we only keep fts_dirfd for DIR_P entries?

lib/libc/gen/fts.c
```diff
@@ -167,6 +167,8 @@ __fts_open(FTS *sp, char * const *argv, int rootfd)
 		p->fts_level = FTS_ROOTLEVEL;
 		p->fts_parent = parent;
 		p->fts_accpath = p->fts_name;
+		if (rootfd != AT_FDCWD)
+			p->fts_dirfd = _dup(rootfd);
```
asomers: Ever since dd532ad13371dcabc07d05052a7a256fc83c6ead , we've only stored fts_dirfd for DIR_P entries, and it refers to the entry itself.  But here, rootfd refers to the original dirfd passed to `fts_openat`, not the entry itself.  Instead, I think you should store `rootfd` in `parent->dirfd` at line 157.  Then here, you should open a new dirfd only if `p` is a directory.

lib/libc/tests/gen/fts_openat_test.c
```diff
@@ -0,0 +1,240 @@
...
+	ATF_REQUIRE_EQ(0, mkdir("dir/sub", 0755));
+	ATF_REQUIRE_EQ(0, close(creat("dir/sub/file", 0644)));
+	ATF_REQUIRE_EQ(0, close(creat("dir/other", 0644)));
+
+	ATF_REQUIRE((dirfd = open("dir", O_RDONLY | O_DIRECTORY)) >= 0);
+
+	ATF_REQUIRE((fts = fts_openat(dirfd, paths,
+	    FTS_PHYSICAL | FTS_NOCHDIR, NULL)) != NULL);
+
+	ATF_REQUIRE_EQ(0, cap_enter());
```
asomers: You should call `cap_enter` _before_ `fts_openat`.

lib/libc/tests/gen/fts_openat_test.c
```diff
@@ -0,0 +1,240 @@
...
+	ATF_REQUIRE_EQ(0, mkdir("dir/sub", 0755));
+	ATF_REQUIRE_EQ(0, close(creat("dir/sub/file", 0644)));
+
+	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL, NULL)) != NULL);
+
+	while ((ent = fts_read(fts)) != NULL) {
+		if (ent->fts_info != FTS_F)
+			continue;
+		if (strcmp(ent->fts_name, "file") != 0)
+			continue;
```
asomers: These restrictions are too restrictive.  We should assert that this access pattern works for directories like `sub` as well as files.  In fact, it should even work for `dir`.  Back in #2303 we initialized the root entry's `fd_parent->fts_dirfd` to -1.  But I realize now that it really ought to be initialized to the dirfd passed to `fts_openat`, which may be `AT_FDCWD`.

Furthermore, I think you should expand `paths` to include two directories, so we can be sure that each root directory gets initialized correctly.  That would detect bugs like double-close.

lib/libc/gen/fts.3
asomers, line 43: You should add a prototype for `fts_openat` here.

asomers (changes requested):

lib/libc/gen/fts.c
```diff
@@ -251,10 +278,18 @@ fts_open(char * const *argv, int options,
...
 	sp->fts_options = options;
+	if (dirfd == AT_FDCWD)
+		rootfd = AT_FDCWD;
+	else if ((rootfd = _dup(dirfd)) < 0) {
+		free(priv);
+		return (NULL);
+	}
+	sp = __fts_open(sp, argv, rootfd);
+	if (rootfd != AT_FDCWD)
+		(void)_close(rootfd);
```
asomers: This is a use-after-free.  In your most recent commit, `__fts_open` stores `rootfd` in `parent->fts_dirfd`.  So you mustn't close it right here.  Instead, close it during `fts_close`.

Jeetubhati:
Fixed all review comments. All 47 fts tests pass including `openat_capsicum` and `fts_capsicum_test`.` parent->fts_dirfd` is now initialized to `AT_FDCWD`, and root directory entries get `fts_dirfd `set via `openat `for both `fts_open `and `fts_openat`.

lib/libc/gen/fts.c
```diff
@@ -779,15 +814,25 @@ fts_build(FTS *sp, int type)
 		oflag = DTF_NODUP;
 	else
 		oflag = DTF_HIDEW | DTF_NODUP;
-	if ((dirp = __opendir2(cur->fts_accpath, oflag)) == NULL) {
+	if (cur->fts_dirfd >= 0) {
+		int fd;
+		fd = _openat(cur->fts_dirfd, ".", O_RDONLY | O_DIRECTORY | O_CLOEXEC);
```
asomers, line 818: It's unfortunate that we have to open an additional file descriptor here just to use `readdir`.  I think there is an optimization possibility here, based on using `fdopendir`/`fdclosedir` to eliminate the extra open.  But that's a task for another day.

lib/libc/gen/fts.c
asomers: Again, are you sure it's needed here?  Is there any guarantee that cur is a directory?

asomers (changes requested):

Jeetubhati:
> Again, are you sure it's needed here? Is there any guarantee that cur is a directory?

`cur` is always a directory when `fts_build `is called (`fts_build `is only called when `cur->fts_info == FTS_D`). The
if` (cur->fts_dirfd < 0)` check is needed to avoid overwriting `fts_dirfd `that was already set for root entries in `__fts_open`.

lib/libc/gen/fts.c
Jeetubhati: Fixed the `rootfd `use-after-free removed the premature close after `__fts_open`. `parent->fts_dirfd` is now closed properly by `fts_close `when it frees the dummy parent.

lib/libc/gen/fts.c
Jeetubhati: Should i leave it for future or optimise it now?

asomers: I suggest leaving it for the future.

asomers (approved):

lib/libc/gen/fts.c
```diff
@@ -213,10 +226,16 @@ __fts_open(FTS *sp, char * const *argv, int rootfd)
...
-	    O_CLOEXEC, 0)) < 0)
-		SET(FTS_NOCHDIR);
+
+	if (!ISSET(FTS_NOCHDIR)) {
+		if (rootfd != AT_FDCWD)
+			sp->fts_rfd = _dup(rootfd);
+		if (sp->fts_rfd < 0)
+			SET(FTS_NOCHDIR);
+		else if ((sp->fts_rfd =
+			_open(".", O_RDONLY | O_CLOEXEC, 0)) < 0)
```
asomers: This leaks a file descriptor.  First, _dup succeeds on line 232.  Then, _open overwrites the just-opened file descriptor on line 236.  I think what you mean to to is this:

```suggestion
		if (rootfd != AT_FDCWD) {
			sp->fts_rfd = _dup(rootfd);
				if (sp->fts_rfd < 0)
				SET(FTS_NOCHDIR);
		}
		else if ((sp->fts_rfd =
			_open(".", O_RDONLY | O_CLOEXEC, 0)) < 0)
```

asomers (changes requested):

lib/libc/gen/fts.c
```diff
@@ -169,11 +173,20 @@ __fts_open(FTS *sp, char * const *argv, int rootfd)
...
 		if (p->fts_info == FTS_DOT)
 			p->fts_info = FTS_D;
+		if (p->fts_info == FTS_D) {
+			if (strcmp(p->fts_name, ".") == 0)
+				p->fts_dirfd = rootfd != AT_FDCWD ?
+				    _dup(rootfd) :
+				    _open(".", O_RDONLY | O_DIRECTORY | O_CLOEXEC);
+			else
+				p->fts_dirfd = _openat(rootfd, p->fts_name,
+				    O_RDONLY | O_DIRECTORY | O_CLOEXEC);
```
asomers: Do you really need special handling for "."?  Can you not just pass that file name directly to openat, like this?
```suggestion
			p->fts_dirfd = _openat(rootfd, p->fts_name,
			    O_RDONLY | O_DIRECTORY | O_CLOEXEC);
```

asomers (approved):
This LGTM now.  But I want to merge #2373 before this PR.  Bugs before features.

asomers:
I missed one thing in my previous review: you need to add a link to lib/libc/gen/Makefile.inc to make the fts_openat man page.  I'll add that myself before I merge.

PR #2277, opened 2026-06-10
hwpmc: Avoid using AMD PMCs if in use by firmware

mashtizadeh:
Some firmwares use the PMCs to monitor OS performance.  We can't be certain that the BIOS would detect any change to the counters if we reprogram them.  In cases where the firmware is using the PMCs to control power management this could have dangerous side effects or unexpected performance effects.  We detect if any of the counters are enabled and disable the AMD PMCs.

Reported by: Sandipan Das
Sponsored by: Netflix

### DIFF
--- a/sys/dev/hwpmc/hwpmc_amd.c
+++ b/sys/dev/hwpmc/hwpmc_amd.c
@@ -864,15 +864,83 @@ amd_pcpu_fini(struct pmc_mdep *md, int cpu)
 }
 
 /*
+ * Check that the PMC hardware is safe to use.  First, we check that the PMCs
+ * are not in use by firmware or another module.  Second, if none of the PMC
+ * feature flags are set, we check that the event selector is working, because
+ * virtual machines have no way to communicate the absence of PMCs.
+ */
+static int
+amd_hwcheck(void)
+{
+	uint64_t reg;
+	int error, i;
+
+	/*
+	 * Some PC vendors enable the core counters in firmware to track
+	 * performance.  The best guess is that this is being used to control
+	 * power management from within the SMM mode.  We shouldn't just take
+	 * over the PMCs in this case.  The user should try disabling any
+	 * performance monitoring or power management functions in the BIOS to
+	 * safely make use of the counters.
+	 */
+	for (i = 0; i < amd_core_npmcs; i++) {
+		error = rdmsr_safe(amd_pmcdesc[i].pm_evsel, &reg);
+		if (error != 0) {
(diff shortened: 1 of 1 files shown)


mashtizadeh:
I need to write a similar patch for Intel.

I couldn't find an AMD machine with this problem, but I have an old Intel HP box that exhibits this behavior.

Sandipan is one of AMD's Linux committers that mentioned the problems in our last meeting.   He cited this patch:
https://github.com/torvalds/linux/commit/4407204c5c9037763aadce39b025529dfbfcac9e

mashtizadeh:
@bsdimp @mhorne

sys/dev/hwpmc/hwpmc_amd.c
```diff
@@ -907,6 +907,36 @@ pmc_amd_initialize(void)
...
 
+	/*
+	 * Some PC vendors enable the core counters in firmware to track
+	 * performance.  The best guess is that this is being used to control
+	 * power management from within the SMM mode.  We shouldn't just take
+	 * over the PMCs in this case.  The user should try disabling any
+	 * performance monitoring or power management functions in the BIOS to
+	 * safely make use of the counters.
+	 */
+	for (i = 0; i < AMD_PMC_CORE_DEFAULT; i++) {
```
mhorne: Should these tests come after we have queried and set `amd_core_npmcs`? Or is it only important to check the first 6 counters?

mhorne (commented):
Bad BIOS!

Generally, LGTM. You might split this logic into a helper function.

sys/dev/hwpmc/hwpmc_amd.c
mashtizadeh: It's fine were it is, we should only see this on older machines.  HWP and other mechanisms are providing alternative ways for them to continue to implement similar functionality without using the counters anymore.  I can't repro it on any recent Intel or AMD machines I bought from those brands that I know have done this in the past.

Don't forget it needs to come above the VM check as that will touch the first PMC.

mashtizadeh: As for the helper function, I'll make that change and update this diff tomorrow or Saturday.

mashtizadeh: I also owe you a patch to clean up the SYSTEM/USER flag it's going to work as you suggested, but I want to implement a best effort filtering.  IBS isn't like other counters we get accurate samples but still suffer from skid during delivery.

mashtizadeh:
I refactored both checks as you suggested into a separate function.

mhorne (approved):

PR #2278, opened 2026-06-12
fts: refactor to use fd-relative operations internally

Jeetubhati:
Replace all `_open(".", O_RDONLY|O_CLOEXEC)` calls with fd-relative equivalents in `__fts_open()`, `fts_read()`, and `fts_children()`.

Add `fts_dirfd` to `struct _ftsent`, set to the file descriptor of the parent directory in `fts_build()`. Callers can use `openat(ent->fts_dirfd, ent->fts_name, ...)` to access files safely without relying on `fts_accpath`, which is important for:

 - Capsicum capability mode, where path-based operations are not permitted after `cap_enter(2)`
 - Security-sensitive programs that need to avoid TOCTOU races by using an fd+name pair instead of a path

`fts_dirfd` is initialized to -1 in `fts_alloc()` and set to `_dup(_dirfd(dirp))` for each entry built in `fts_build()`. The duplicated fd is owned by the entry and closed when the entry is freed.

The root fd is now passed as an explicit argument to `__fts_open()` rather than stored in `struct _fts_private`. This avoids the risk of descendant functions incorrectly using a stale root fd after fts has internally `fchdir`'d. `fts_open()` and `fts_open_b()` pass `AT_FDCWD`; a future `fts_openat()` will pass the caller's dirfd.


lib/libc/gen/fts.c
```diff
@@ -246,6 +249,7 @@ fts_open(char * const *argv, int options,
 	/* Allocate/initialize the stream. */
 	if ((priv = calloc(1, sizeof(*priv))) == NULL)
 		return (NULL);
+	priv->ftsp_rootfd = AT_FDCWD;
```
asomers: This is only going to work for small directory trees.  For larger ones, fts might internally chdir.  Then AT_FDCWD would be pointing to the wrong place.  So it would be better to use an actual file descriptor.  Like fts_rfd.  Why can't you just use that?

lib/libc/gen/fts.c
```diff
@@ -97,6 +99,7 @@ struct _fts_private {
 	struct statfs	ftsp_statfs;
 	dev_t		ftsp_dev;
 	int		ftsp_linksreliable;
+	int		ftsp_rootfd;
```
asomers: Why add a new file descriptor, instead of reusing fts_rfd?

lib/libc/gen/fts.c
```diff
@@ -439,8 +443,8 @@ fts_read(FTS *sp)
 	    (p->fts_info == FTS_SL || p->fts_info == FTS_SLNONE)) {
 		p->fts_info = fts_stat(sp, p, 1, -1);
 		if (p->fts_info == FTS_D && !ISSET(FTS_NOCHDIR)) {
-			if ((p->fts_symfd = _open(".", O_RDONLY | O_CLOEXEC,
-			    0)) < 0) {
+			if ((p->fts_symfd = _openat(ROOTFD(sp), ".",
```
asomers: If we've already changed directories, then ROOTFD(sp) will be the wrong thing to use here.

lib/libc/gen/fts.c
```diff
@@ -532,7 +536,8 @@ next:	tmp = p;
 			p->fts_info = fts_stat(sp, p, 1, -1);
 			if (p->fts_info == FTS_D && !ISSET(FTS_NOCHDIR)) {
 				if ((p->fts_symfd =
-				    _open(".", O_RDONLY | O_CLOEXEC, 0)) < 0) {
+                                    _openat(ROOTFD(sp), ".", O_RDONLY |
```
asomers: Just as above, ROOTFD(sp) will be wrong if we've already changed directories.

lib/libc/gen/fts.c
```diff
@@ -671,7 +676,7 @@ fts_children(FTS *sp, int instr)
 	    ISSET(FTS_NOCHDIR))
 		return (sp->fts_child = fts_build(sp, instr));
 
-	if ((fd = _open(".", O_RDONLY | O_CLOEXEC, 0)) < 0)
+	if ((fd = _openat(ROOTFD(sp), ".", O_RDONLY | O_CLOEXEC, 0)) < 0)
```
asomers: Here, too, ROOTFD(sp) might be wrong.

asomers (changes requested):
What about the _open in fts_safe_changedir?  I think that instead of changing directories, it would be better to open the child directory, and use that for future openat calls.  But since the child directory is _already_ opened, there might not be anything to do.  In fact, we might be able to remove fts_safe_changedir altogether.

Jeetubhati:
> What about the _open in fts_safe_changedir? I think that instead of changing directories, it would be better to open the child directory, and use that for future openat calls. But since the child directory is _already_ opened, there might not be anything to do. In fact, we might be able to remove fts_safe_changedir altogether.

if the child directory is already opened via fts_safe_changedir, we might not need path-based opens at all.
I need to think through whether fts_safe_changedir can be removed or simplified. Will look at this more carefully.

lib/libc/gen/fts.c
Jeetubhati: I've changed ROOTFD(sp) to AT_FDCWD for the mid-traversal CWD saves in fts_read() and fts_children().

lib/libc/gen/fts.c
asomers: I don't see anything that uses this field outside of __fts_open.  So I don't think it makes sense to store it in this structure.  Or is the plan that this field will be used by `fts_openat`, which will itself call `__fts_open`, once you add that function?  If the intention is that `ftsp_rootfd` will only ever be used by `__fts_open` and not by any descendant functions, then I think it's risky to include it in the struct.  It might be tempting to use it later on, but that would be wrong, given the possibility of chdir.  Instead, I think it would be more appropriate to pass this file descriptor as an argument to `__fts_open`.

lib/libc/gen/fts.c
```diff
@@ -439,8 +443,8 @@ fts_read(FTS *sp)
 	    (p->fts_info == FTS_SL || p->fts_info == FTS_SLNONE)) {
 		p->fts_info = fts_stat(sp, p, 1, -1);
 		if (p->fts_info == FTS_D && !ISSET(FTS_NOCHDIR)) {
-			if ((p->fts_symfd = _open(".", O_RDONLY | O_CLOEXEC,
-			    0)) < 0) {
+			if ((p->fts_symfd = _openat(AT_FDCWD, ".",
```
asomers: This isn't fd-relative enough.  It will fail for any capability-mode program.  Can you use fts_dirfd here?  And in the next two locations too?

lib/libc/gen/fts.c
```diff
@@ -671,7 +676,7 @@ fts_children(FTS *sp, int instr)
 	    ISSET(FTS_NOCHDIR))
 		return (sp->fts_child = fts_build(sp, instr));
 
-	if ((fd = _open(".", O_RDONLY | O_CLOEXEC, 0)) < 0)
+	if ((fd = _openat(AT_FDCWD, ".", O_RDONLY | O_CLOEXEC, 0)) < 0)
```
asomers: Also here, can you use fts_dirfd?

asomers (changes requested):
Don't forget to mirror any changes that you make to `fts_open` to `fts_open_b`, too.

lib/libc/gen/fts.c
Jeetubhati: Done. I removed ftsp_rootfd from `struct _fts_private` and the
`FTSP()/ROOTFD()` macros. The `fd `is now passed directly as a
`rootfd `argument to `__fts_open()`.` fts_open()` and `fts_open_b()`
pass AT_FDCWD; `fts_openat()` will pass the caller's `dirfd`.

Jeetubhati:
> Don't forget to mirror any changes that you make to `fts_open` to `fts_open_b`, too.

Done. `fts_open_b() `now passes `AT_FDCWD` to `__fts_open()` to match `fts_open()`.

lib/libc/gen/fts.c
Jeetubhati: All three locations now use `fts_dirfd` instead of `AT_FDCWD`. Since `fts_dirfd `is -1 before `fts_build()` has run (e.g. when `fts_children()` is called before `fts_read()`), we fall back to `AT_FDCWD` in that case. `fts_dirfd` is now dup'd in `fts_build()` so it remains valid after `closedir()`, and closed in `fts_lfree()` and other free sites.

lib/libc/gen/fts.c
Jeetubhati: Done. `fts_children()` now uses `fts_dirfd` with `AT_FDCWD` fallback.

lib/libc/gen/fts.c
```diff
@@ -360,6 +357,8 @@ fts_close(FTS *sp)
 			p = p->fts_link != NULL ? p->fts_link : p->fts_parent;
 			free(freep);
 		}
+		if (p->fts_dirfd >= 0)
+                        (void)_close(p->fts_dirfd);
```
asomers: Whitespace problem here.  Needs spaces -> tabs.

lib/libc/gen/fts.c
```diff
@@ -563,6 +564,8 @@ name:		t = sp->fts_path + NAPPEND(p->fts_parent);
 		 * can distinguish between error and EOF.
 		 */
 		free(tmp);
+		if (p->fts_dirfd >= 0)
+                        (void)_close(p->fts_dirfd);
```
asomers: Whitespace problem here, too.

lib/libc/gen/fts.c
```diff
@@ -443,7 +442,8 @@ fts_read(FTS *sp)
 	    (p->fts_info == FTS_SL || p->fts_info == FTS_SLNONE)) {
 		p->fts_info = fts_stat(sp, p, 1, -1);
 		if (p->fts_info == FTS_D && !ISSET(FTS_NOCHDIR)) {
-			if ((p->fts_symfd = _openat(AT_FDCWD, ".",
+			if ((p->fts_symfd = _openat(p->fts_dirfd >= 0 ?
```
asomers: If I understand correctly, fts_dirfd might not be set only for the root directory.  Is that correct?  Would it make more sense to require that `fts_dirfd` always be valid, and initialize it to `AT_FDCWD` during `fts_alloc`?

lib/libc/gen/fts.c
```diff
@@ -1211,9 +1216,11 @@ fts_lfree(FTSENT *head)
 
 	/* Free a linked list of structures. */
 	while ((p = head)) {
-		head = head->fts_link;
-		free(p);
-	}
+                head = head->fts_link;
```
asomers: Whitespace problem here, too.

lib/libc/gen/fts.c
```diff
@@ -439,8 +442,9 @@ fts_read(FTS *sp)
 	    (p->fts_info == FTS_SL || p->fts_info == FTS_SLNONE)) {
 		p->fts_info = fts_stat(sp, p, 1, -1);
 		if (p->fts_info == FTS_D && !ISSET(FTS_NOCHDIR)) {
-			if ((p->fts_symfd = _open(".", O_RDONLY | O_CLOEXEC,
-			    0)) < 0) {
+			if ((p->fts_symfd = _openat(p->fts_dirfd >= 0 ?
+                            p->fts_dirfd : AT_FDCWD, ".",
+                            O_RDONLY | O_CLOEXEC, 0)) < 0) {
```
jillest: In the `p->fts_dirfd >= 0` case, this is just a slower dup with a less suitable permission check. Also, perhaps this can reuse `p->fts_dirfd` without any additional system calls.

lib/libc/gen/fts.c
```diff
@@ -907,6 +917,7 @@ mem1:				saved_errno = errno;
 		}
 
 		p->fts_level = level;
+		p->fts_dirfd = _dup(_dirfd(dirp));
```
jillest, line 920: This limits the directory tree depth to the number of file descriptors that can be opened, and I think imposing such a limit was deliberately avoided before. Also note that it allows better performance because it avoids the need for the `open("..")/fstat/fchdir` dance.

The ideal approach would be to keep a particular number of file descriptors open and close the excess (starting from the numerically lowest levels). Then, if we need to return to a directory and the descriptor was closed, use `openat("..")` and `fstat` (checking dev and ino) to open it again. If Capsicum does not allow opening `..`, it would have to be opened again starting from the root level (also checking dev and ino), which increases the risk of aborting for concurrent changes and increases code complexity even more because the pathname may be longer than `{PATH_MAX}` (requiring the pathname to be split up into multiple `openat` calls). An alternative could be to accept the limitation on the directory tree depth if Capsicum is in use (note that this assumes that `find` run from scripts like `/etc/periodic/daily/110.clean-tmps` will not use Capsicum, lest an attacker dodge the cleaner by creating a really deep directory tree).

This is also what the `maxfds` parameter in `ftw` and `nftw` is for.

Also, `..` will not work if a symlink had been followed (which is why `fts_symfd` already existed). I don't care too much about limiting the number of nested symlink follows, since symlinks cannot be followed when traversing an untrusted directory tree anyway.

jillest, line 920: Some refactoring should allow using our non-standard `fdclosedir`, saving a `dup` and a `close` system call.

lib/libc/gen/fts.c
jillest: I don't think so, because it will cause the application to access wrong files for the root level if it uses the combination of `fts_dirfd` and `fts_name`.

Hmm, I realize that "the combination of `fts_dirfd` and `fts_name`" may be too simple. I think things should work out such that the application accesses files at the root level using the pathnames it originally passed to `fts_open` or similar (and with the directory file descriptor in the case of `fts_openat`), even if those pathnames contain slashes. For example, a command like `chmod -R o-rwx dirwithlotsoffiles/*` should not open and close `dirwithlotsoffiles` for every file.

lib/libc/gen/fts.c
asomers: On my most memory-challenged FreeBSD device, `ulimit -n` returns 116802.  So trying to economize on file descriptors strikes me as premature optimization.  So I for one am ok with optimizing for the capsicum-use-case.

lib/libc/gen/fts.c
Jeetubhati: Fixed. All three locations now use tabs consistently.

lib/libc/gen/fts.c
Jeetubhati: For root-level entries, `fts_dirfd` is -1 because fts_build() has not yet run when the root entries are stat'd in `__fts_open(`). Initializing to `AT_FDCWD` would be incorrect since openat(AT_FDCWD,fts_name) would fail for paths like "dirwithlotsoffiles/file" that contain slashes.

The correct `fd` for root-level entries is the one passed by thecaller to `fts_openat()` that will be handled when `fts_openat()` is implemented in the next PR. For now, -1 is the right sentinel
to indicate "use fts_accpath instead".

lib/libc/gen/fts.c
Jeetubhati: Thank you for the detailed analysis. The fd-per-entry approach does impose a depth limit, but as Alan notes the default fd limit on FreeBSD is well over 100k which covers any realistic directory tree depth in practice.

The maxfds approach you describe would be the right solution for a production-quality implementation, but it adds significant complexity especially the Capsicum case where `openat("..")` is not available and re-opening from the root with dev/ino checks would be needed. That complexity is out of scope for this GSoC project.

For now the fd-per-entry approach is correct and safe for the Capsicum use case. The depth limitation is an acceptable trade-off given the fd limit on modern FreeBSD systems. If this becomes a concern in the future, the `maxfds` optimization can be added as a follow-up.

Jeetubhati: Good point. I looked at fdclosedir and it would save one dup and one close per directory. I am deferring this optimization to a follow-up PR to keep this one focused. I think the current approach is correct even if not maximally efficient.

lib/libc/gen/fts.c
Jeetubhati: Fixed.  Changed to `_dup(p->fts_dirfd)` directly when `fts_dirfd` is valid, avoiding the unnecessary `open(".")` call entirely.

asomers: That's a little bit better.  But are you sure that you need to dup the file descriptor at all? Would it be possible to simply copy `fts_dirfd`?

asomers: Actually, I take that back.  Trying to eliminate the `_dup` at this point would be a premature optimization.  The original code always opened a new file descriptor at this point, so it's reasonable for your refactor to do the same.  Eliminating the extra syscall (and its corresponding `close`) would be a good future optimization, but doesn't need to be part of this PR.

lib/libc/gen/fts.c
```diff
@@ -212,9 +213,9 @@ __fts_open(FTS *sp, char * const *argv)
 	 * descriptor we run anyway, just more slowly.
 	 */
 	if (!ISSET(FTS_NOCHDIR) &&
-	    (sp->fts_rfd = _open(".", O_RDONLY | O_CLOEXEC, 0)) < 0)
+	    (sp->fts_rfd = _openat(rootfd, ".", O_RDONLY |
```
asomers, line 216: This code _looks_ like it could be replaced by `_dup`.  But it actually can't, because `rootfd` may be `AT_FDCWD`.  Might be worth a comment.

lib/libc/gen/fts.c
Jeetubhati: Understood. Keeping the` _dup()` as-is for now. The optimization can be done in a follow-up.

asomers (approved):
LGTM.  I'll leave it open for a little while in case jilles has any final comments.

lib/libc/gen/fts.c
Jeetubhati: Added a note to the existing comment explaining why` _dup` cannot be used here —> `rootfd` may be `AT_FDCWD`, not a real descriptor.

lib/libc/gen/fts.c
```diff
@@ -205,16 +206,18 @@ __fts_open(FTS *sp, char * const *argv)
 	sp->fts_cur->fts_info = FTS_INIT;
 
 	/*
-	 * If using chdir(2), grab a file descriptor pointing to dot to ensure
-	 * that we can get back here; this could be avoided for some paths,
-	 * but almost certainly not worth the effort.  Slashes, symbolic links,
-	 * and ".." are all fairly nasty problems.  Note, if we can't get the
-	 * descriptor we run anyway, just more slowly.
-	 */
+         * If using chdir(2), grab a file descriptor pointing to dot to ensure
```
asomers: Arghh, your editor turned tabs into spaces again.  Please change them back.

asomers (changes requested):

lib/libc/gen/fts.c
Jeetubhati: Fixed.

asomers (approved):

lib/libc/gen/fts.c
jillest: One issue here is that having `fts_dirfd` be a public field makes it harder to make a change later that closes it temporarily, since applications may grow to depend on it being valid. Or do you think it is sufficient to have a caveat in the man page like the one about `fts_path` and `fts_accpath`?

And we do want FreeBSD to be production-quality, right?

It need not block merging this pull request.

jillest (approved):

lib/libc/gen/fts.c
Jeetubhati: Good point about fts_dirfd being public making future changes harder. A man page caveat similar to `fts_path/fts_accpath` would be appropriate something like "`fts_dirfd` is valid only for the duration of the `fts_read()` call that returned the entry".

The `maxfds` optimization and any `fd` management improvements can be done in a follow-up PR after the Capsicum work lands. The immediate next steps are:

- PR #2273 (`fts_openat`): `fts_openat()` API on top of this refactor
- A new PR for `fts_capsicum_test.c` using `openat(fts_dirfd, fts_name)`
- Man page update documenting `fts_dirfd` lifetime

asomers: If we don't want it in the public API, should we move `fts_dirfd` into `struct _fts_private`?

Jeetubhati: > If we don't want it in the public API, should we move `fts_dirfd` into `struct _fts_private`?

`fts_dirfd` needs to be in the public struct for callers to use `openat(ent->fts_dirfd, ent->fts_name)` that is the whole point of adding it for Capsicum compatibility. Moving it to `_fts_private` would make it inaccessible to callers.

The man page caveat should be sufficient: "`fts_dirfd `is valid only for the duration of the` fts_read()` call that returned the entry, and must not be closed by the caller." This is the same approach used for `fts_path` and `fts_accpath`.

jillest: > > If we don't want it in the public API, should we move `fts_dirfd` into `struct _fts_private`?

> `fts_dirfd` needs to be in the public struct for callers to use `openat(ent->fts_dirfd, ent->fts_name)` that is the whole point of adding it for Capsicum compatibility. Moving it to `_fts_private` would make it inaccessible to callers.

> The man page caveat should be sufficient: "`fts_dirfd `is valid only for the duration of the` fts_read()` call that returned the entry, and must not be closed by the caller." This is the same approach used for `fts_path` and `fts_accpath`.

And another caveat that `openat(ent->fts_dirfd, ent->fts_name)` won't work if `entry->fts_level == FTS_ROOTLEVEL`, since `openat(initial_fd_or_AT_FDCWD, ent->fts_accpath)` is necessary in that case. Such a contorted API is hard to use. This could be improved by moving `fts_dirfd` to `struct _fts_private` and providing access via a function (which would return an fd/path pair somehow), by adding said pair to the `FTSENT` struct or otherwise making `openat(ent->fts_some_fd, ent->fts_some_path)` work for all levels.

Jeetubhati: > And another caveat that `openat(ent->fts_dirfd, ent->fts_name)` won't work if `entry->fts_level == FTS_ROOTLEVEL`, since `openat(initial_fd_or_AT_FDCWD, ent->fts_accpath)` is necessary in that case. Such a contorted API is hard to use. This could be improved by moving `fts_dirfd` to `struct _fts_private` and providing access via a function (which would return an fd/path pair somehow), by adding said pair to the `FTSENT` struct or otherwise making `openat(ent->fts_some_fd, ent->fts_some_path)` work for all levels.

I hadn't considered the root level inconsistency. You are right that requiring callers to check `fts_level` and use different approaches for root vs non-root entries makes the API awkward and error-prone.

I think the cleanest fix would be to hold `fts_dirfd` until `fts_openat()` is ready, so that root-level entries can receive the caller's `dirfd` and `openat(ent->fts_dirfd, ent->fts_name)` works consistently at all levels. That way the API lands complete rather than half-done.

dch:
This seems to break running poudriere-devel. I can only test on arm64 presently:
```
[00:00:10] Warning: (accessibility/at-spi2-core): jexec: jail_attach(2): Operation not permitted
```

Bisecting over the last few days leads to this specific commit:

https://cgit.freebsd.org/src/commit/commit?id=e03ed9daeb49fffa1d16b8d00240c65e92650d01

NB that normal 
```
dch@a05 ~> gently poudriere bulk -j 15_1_a64 -C dns/doggo
[00:00:00] Creating the reference jail... done
[00:00:06] Mounting system devices for 15_1_a64-default
[00:00:06] Stashing existing package repository
[00:00:06] Mounting ports from: /usr/ports
[00:00:06] Mounting packages from: /usr/local/poudriere/data/packages/15_1_a64-default
[00:00:06] Mounting distfiles from: /var/cache/distfiles
[00:00:06] Copying /var/db/ports from: /usr/local/etc/poudriere.d/options
[00:00:06] Appending to make.conf: /usr/local/etc/poudriere.d/make.conf
[00:00:06] Warning: DEVELOPER=yes ignored from make.conf. Use 'bulk -t' or 'testport' for testing instead.
/etc/resolv.conf -> /usr/local/poudriere/data/.m/15_1_a64-default/ref/etc/resolv.conf
[00:00:06] Starting jail 15_1_a64-default
Updating /var/run/os-release done.
[00:00:07] Creating user dch done
[00:00:07] Will build as dch:wheel (1002:0)
[00:00:08] Setting schg on jail base paths
[00:00:09] Ports supports: FLAVORS SUBPACKAGES SELECTED_OPTIONS
[00:00:09] Acquiring build logs lock for 15_1_a64-default... done
[00:00:09] Logs: /usr/local/poudriere/data/logs/bulk/15_1_a64-default/2026-06-25_09h53m16s
[00:00:09] WWW: https://a05.cabal5.net/poudriere/build.html?mastername=15_1_a64-default&build=2026-06-25_09h53m16s
[00:00:09] Loading MOVED for /usr/local/poudriere/data/.m/15_1_a64-default/ref/usr/ports
[00:00:09] Gathering ports metadata
[00:00:09] Warning: (lang/go126): jexec: jail_attach(4): Operation not permitted
```

Can we hold off MFC until we figure out where this error comes from?

asomers:
> Can we hold off MFC until we figure out where this error comes from?

Yes, definitely.  In fact, we've temporarily reverted this change while @Jeetubhati works on a new version.  See bd1c0ff49c48950dbd999c024a17199022f81d36 .

PR #2279, opened 2026-06-12
Allow FreeBSD mount NetBSD's FFSv2 WAPBL read-only

ricardobranco777:
ufs: Allow read-only mounting of NetBSD FFSv2 WAPBL filesystems
    
    Skip UFS2 fs_metaspace upper-bound validation that rejects NetBSD FFSv2
    WAPBL filesystems due to differing superblock layouts.
    
    Detect the condition during mount instead and permit read-only mounts
    while rejecting read-write mounts with EROFS.  This follows NetBSD's
    recommendation for systems without WAPBL support and avoids modifying
    unsupported journal metadata.

Fixes https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=296022

### DIFF
--- a/sys/ufs/ffs/ffs_vfsops.c
+++ b/sys/ufs/ffs/ffs_vfsops.c
@@ -604,6 +604,16 @@ ffs_mount(struct mount *mp)
 			if (error) {
 				return (error);
 			}
+			/*
+			 * Refuse upgrade of NetBSD WAPBL filesystem to
+			 * read-write; see validate_sblock() for details.
+			 */
+			if (fs->fs_magic == FS_UFS2_MAGIC &&
+			    fs->fs_metaspace > fs->fs_fpg / 2) {
+				vfs_mount_error(mp,
+				    "WAPBL filesystem must be mounted read-only");
+				return (EROFS);
+			}
 			fs->fs_flags &= ~FS_UNCLEAN;
 			if (fs->fs_clean == 0) {
 				fs->fs_flags |= FS_UNCLEAN;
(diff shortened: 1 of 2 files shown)


asomers:
This patch may be sufficient to mount, but what guarantees are there that it will work correctly?  That all types of metadata will be readable, and that performing writes on FreeBSD won't render the volume corrupt from NetBSD's perspective?  It will take a very detailed understanding of both operating systems' on-disk formats to answer those questions.
cc @realchonk @mckusick .

ricardobranco777:
> This patch may be sufficient to mount, but what guarantees are there that it will work correctly? That all types of metadata will be readable, and that performing writes on FreeBSD won't render the volume corrupt from NetBSD's perspective? It will take a very detailed understanding of both operating systems' on-disk formats to answer those questions. cc @realchonk @mckusick .

`fs_unrefs` is used only once in sys/ufs/ufs/ufs_gjournal.c.  Ditto for `fs_providersize` in sys/geom/label/g_label_ufs.c.  These code paths are touched only if the user runs `gjournal label` on the device.

`fs_metaspace` seems safe for read-only mounts but I'm not so sure for read-write.

I'd be honoured to have a review by Marshall Kirk McKusick. :)

sys/ufs/ffs/ffs_vfsops.c
```diff
@@ -926,6 +936,23 @@ ffs_mountfs(struct vnode *odevvp, struct mount *mp, struct thread *td)
...
+	if (fs->fs_magic == FS_UFS2_MAGIC &&
+	    fs->fs_metaspace > fs->fs_fpg / 2) {
+		if ((mp->mnt_flag & MNT_RDONLY) == 0) {
+			vfs_mount_error(mp, "NetBSD WAPBL filesystem must be "
+			    "mounted read-only");
+			error = EROFS;
+			goto out;
+		}
+		printf("WARNING: %s: NetBSD WAPBL filesystem mounted "
+		    "read-only (fs_metaspace %jd, fs_fpg/2 %jd)\n",
```
bsdimp: we've been trying to keep the string all on one line for better grepability. Here and elsewhere.

bsdimp:
So gjournal is going into the shredder
this change looks good to me.

bsdimp (commented):
So apart from the strings thing, I like this commit.

sys/ufs/ffs/ffs_vfsops.c
```diff
@@ -926,6 +936,23 @@ ffs_mountfs(struct vnode *odevvp, struct mount *mp, struct thread *td)
 		    ffs_use_bread);
 	if (error != 0)
 		goto out;
+	/*
+	 * Per NetBSD's wapbl(4), systems without WAPBL support should mount it
+	 * read-only; see validate_sblock() for details.
+	 */
+	if (fs->fs_magic == FS_UFS2_MAGIC &&
+	    fs->fs_metaspace > fs->fs_fpg / 2) {
```
bsdjhb, line 944: Is this the canonical way to detect WAPBL?

sys/ufs/ffs/ffs_vfsops.c
ricardobranco777: Hopefully fixed.  I removed the "NetBSD" string and adjusted them.  Only one message is no longer respecting the 80 chars style limit.

sys/ufs/ffs/ffs_vfsops.c
ricardobranco777: > Is this the canonical way to detect WAPBL?

We can't rely on fs->fs_flags alone because NetBSD's FS_DOWAPBL (0x100) collides with FreeBSD's FS_NFS4ACLS (NetBSD uses 0x800 for this, but that's another history).

We can't also rely on fs->fs_flags because this value is merged from fs->fs_old_flags in ffs_oldfscompat_read().

bsdimp:
Kirk was going to land this, but hit a snag I don't understand, so I'm going to land it instead.

bsdimp:

PR #2283, opened 2026-06-15
fts: remove chdir, use openat for directory traversal

Jeetubhati:
fts(3) currently uses chdir(2) to descend into directories during traversal, relying on fts_safe_changedir() for safety and fts_rfd/fts_symfd to restore the original directory afterwards. This makes fts incompatible with Capsicum capability mode where chdir() is not permitted.

This patch series removes all chdir() usage from fts and replaces it with openat() using directory file descriptors:

Commit 1: fts_build() stores the directory fd in sp->fts_rfd and always uses the full path for fts_accpath (previously only done under FTS_NOCHDIR). The fts_safe_changedir() call is removed.

Commit 2: fts_read() removes all fchdir()/fts_safe_changedir() calls for descending and returning to parent. fts_close() no longer restores the original CWD since it is never changed.

The FTS_NOCHDIR flag and fts_safe_changedir() will be removed in follow-up commits.

### DIFF
--- a/lib/libc/gen/fts.c
+++ b/lib/libc/gen/fts.c
@@ -343,7 +343,6 @@ int
 fts_close(FTS *sp)
 {
 	FTSENT *freep, *p;
-	int saved_errno;
 
 	/*
 	 * This still works if we haven't read anything -- the dummy structure
@@ -377,20 +376,6 @@ fts_close(FTS *sp)
 #endif /* __BLOCKS__ */
 	}
 
-	/* Return to original directory, save errno if necessary. */
-	if (!ISSET(FTS_NOCHDIR)) {
-		saved_errno = fchdir(sp->fts_rfd) ? errno : 0;
-		(void)_close(sp->fts_rfd);
-
-		/* Set errno and return. */
-		if (saved_errno != 0) {
-			/* Free up the stream pointer. */
-			free(sp);
-			errno = saved_errno;
-			return (-1);
-		}
-	}
-
 	/* Free up the stream pointer. */
 	free(sp);
 	return (0);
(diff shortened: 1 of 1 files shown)


jillest (commented):
Paths containing slashes in `fts_accpath` make programs vulnerable to symlink attacks and other race conditions. For example, an attacker might trick a `find` cleaning up `/tmp` as root into deleting a file somewhere else by replacing a directory with a symlink after the directory has been read but before the contained file is deleted. The `find` code is paranoid about this with `-delete`.

For compatibility, the default mode will have to continue designating files by `fts_accpath` not containing a slash (if at all possible) relative to the current directory. A new thread-safe mode can designate files by a new fd and `fts_accpath` pair.

asomers:
Good catch, @jillest .  If I understand correctly, the danger is that somebody might do something like `find /tmp ... -delete`, and, if a file like `/tmp/a/b/c` exists, then the attacker might replace b with a symlink, like `/tmp/a/b -> /root`, tricking the user into deleting `/root/c`?  In that case, I guess we'll have to do as you suggest, and continue to change directories (but using `fchdir` instead of `chdir`).  Then in a future commit we can add a `int fts_dirfd` field to `struct ftsent`, guaranteeing that `openat(fts_dirfd, fts_name)` will return the correct file, even in the face of symlink attacks, and even if `FTS_NOCHDIR` is in use.  Would that work?

Jeetubhati:
Closing. Removing chdir breaks the security guarantee that fts_accpath is a single filename relative to CWD. Per jillest's review and Alan's suggestion, the correct approach is to keep fchdir and add fts_dirfd to FTSENT instead. PR https://github.com/freebsd/freebsd-src/pull/2278 will be updated with the correct internal changes.

PR #2284, opened 2026-06-16
CHERI: syscall implementation wrappers

brooksdavis:
With CHERI the set of system calls that need to be translated in compatibility layers changes. Instead of systems calls that point to objects whose underlying type changes (e.g., pointers in freebsd32 are 32-bit integers), all systems calls that take a pointer need to be changed.  This is because all access to userspace in a CHERI kernel is via a capability and we need to make one from the integer pointer arguments. This series of commits adds kern_ or, where a kern_ that isn't suitable already exists, user_ wrapper functions. These new functions will take capability arguments either directly from sys_ system call implementations or from freebsd64_ (and perhaps eventually freebsd32_) implementations that create capabilities to pass in.

Most of these functions are trivial (often to the point that they could theoretically be generated), but in a few cases, such as sendfile(2), I've done a more significant refactoring of existing code.

In the creation of user_ functions I've strongly biased toward reduction of duplicate code based on my experiences writing first compat/cheriabi (the initial CHERI syscall interface) and compat/freebsd64. The goal is to make freebsd64_ implementations as thin as possible.

Changes are generally split by file to allow bisection and easy reverts should they become necessary.

### DIFF
--- a/sys/compat/freebsd32/freebsd32_misc.c
+++ b/sys/compat/freebsd32/freebsd32_misc.c
@@ -2163,81 +2163,21 @@ struct sf_hdtr32 {
 };
 
 static int
-freebsd32_do_sendfile(struct thread *td,
-    struct freebsd32_sendfile_args *uap, int compat)
+freebsd32_copyin_hdtr(const struct sf_hdtr32 *uhdtr,
+    struct sf_hdtr *hdtr)
 {
 	struct sf_hdtr32 hdtr32;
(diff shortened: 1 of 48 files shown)


brooksdavis:
I've run the FreeBSD test suite using bricoler before and after these changes on amd64 and get the same set of results. Should anyone want any of these reviews desperately (e.g., sendfile or jail changes), I'm happy to pull them over to Phabricator.

PR #2286, opened 2026-06-18
Update rc.subr.8 : document ${name}_umask

eborisch:
Documented in the rc.conf man page, but missed in the rc.subr man page.

Signed-off-by: eborisch@gmail.com

### DIFF
--- a/share/man/man8/rc.subr.8
+++ b/share/man/man8/rc.subr.8
@@ -811,6 +811,8 @@ uses
 Only supported after
 .Pa /usr
 is mounted.
+.It Va ${name}_umask
+Run the service using this umask(1) value.
 .It Va ${name}_group
 Group to run the chrooted
 .Va command


share/man/man8/rc.subr.8
```diff
@@ -811,6 +811,8 @@ uses
 Only supported after
 .Pa /usr
 is mounted.
+.It Va ${name}_umask
+Run the service using this umask(1) value.
```
concussious, line 815: Thanks for the patch!

I see this is what it says in rc.conf(5), but umask(1) is a link to builtin(1), which tells you to go read sh(1), which tells you to go read umask(2), which doesn't even tell you anything... so this text is totally useless. I think the following would be a liiitle better?
```suggestion
Run the service using this file creation mask.
```
Please change it in `share/man/man5/rc.conf.5` in this patch too.

concussious (changes requested):

PR #2288, opened 2026-06-19
libc: Implement qualifier-preserving standard library functions

kfv:
This PR implements the qualifier-preserving standard library functions as proposed in [N3020](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3020.pdf). I've all the cases I could think of. Both world and kernel builds complete successfully as well.

The one thing on my radar is the upcoming C23 migration of the base. There are still a bunch of inconsistencies in the tree that need sorting out before we get there, so I've already started chipping away at the cleanup. Should have it wrapped up within a week or two.

Cc: @clausecker

### DIFF
--- a/sys/sys/cdefs.h
+++ b/sys/sys/cdefs.h
@@ -228,6 +228,27 @@
 #endif
 
 /*
+ * __qualsel() is the building block for C23 qualifier-preserving macros
+ * as proposed in N3020: it selects cexpr when the pointer expression p
+ * references a const-qualified object, and expr otherwise.
+ *
+ * The conditional operator's composite-type rule collapses every
+ * pointer-to-const-object type onto the single "const void *"" association,
+ * so callers passing any such pointer are matched even though _Generic()
+ * otherwise compares types exactly.
+ *
+ * The (__uintptr_t) round-trip only suppresses -Wcast-qual on the throwaway
+ * second operand.
+ */
+#if (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L) || \
+    __has_extension(c_generic_selections)
+#define	__qualsel(p, cexpr, expr)					\
+	_Generic(1 ? (p) : (void *)(__uintptr_t)(p),			\
+	    const void *: (cexpr),					\
+	    default: (expr))
+#endif
+
+/*
  * C99 Static array indices in function parameter declarations.  Syntax such as:
  * void bar(int myArray[static 10]);
  * is allowed in C99 but not in C++.  Define __min_size appropriately so
--- a/include/stdlib.h
+++ b/include/stdlib.h
@@ -414,4 +414,16 @@ errno_t	 qsort_s(void *, rsize_t, rsize_t,
 __END_DECLS
 __NULLABILITY_PRAGMA_POP
 
+#if defined(__qualsel) && !defined(__cplusplus) && \
+    defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L
+#define	bsearch(key, base, nmemb, size, compar)		__qualsel((base),    \
+	(const void *)(bsearch)((key), (base), (nmemb), (size), (compar)),   \
+	(bsearch)((key), (base), (nmemb), (size), (compar)))
+#ifdef __BLOCKS__
+#define	bsearch_b(key, base, nmemb, size, compar)	__qualsel((base),    \
+	(const void *)(bsearch_b)((key), (base), (nmemb), (size), (compar)), \
+	(bsearch_b)((key), (base), (nmemb), (size), (compar)))
+#endif
+#endif
+
 #endif /* !_STDLIB_H_ */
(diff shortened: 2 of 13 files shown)


clausecker:
How does this interact with ssp macros for the same functions?  I suppose we'd need the qualifier selection for the SSP macros, too?

kfv:
Negative. None of the affected functions have corresponding SSP variants. They are write-destination wrappers that don't return a pointer derived from the input buffer, so there's no qualifier to discard.

kfv:
By the way, do you think we should update the man pages? And if so, should we align the synopsis with the standard, e.g. `QChar *strchr(QChar *s, int c);`, or simply note that these are qualifier-preserving in C23 and later modes? I’ve considered this a few times but couldn't reach a clear conclusion.

clausecker:
> By the way, do you think we should update the man pages? And if so, should we align the synopsis with the standard, e.g. `QChar *strchr(QChar *s, int c);`, or simply note that these are qualifier-preserving in C23 and later modes? I’ve considered this a few times but couldn't reach a clear conclusion.

I think it might actually be better to leave the prototypes as they are in the man pages and then to just add a note.  glibc's current man pages have the full prototypes, making the man pages really hard to read.

> Negative. None of the affected functions have corresponding SSP variants. They are write-destination wrappers that don't return a pointer derived from the input buffer, so there's no qualifier to discard.

That's a relief to hear!

Please add the man page changes you envision and I'll try to do a full review.

kfv:
> I think it might actually be better to leave the prototypes as they are in the man pages and then to just add a note. glibc's current man pages have the full prototypes, making the man pages really hard to read.

Agreed, as for new users unfamiliar with the standard, names like `QVoid`, `QChar`, and `QWchar_t` could be so much awkward, and they might reasonably try to use them and then find out they don’t exist. Some mysterious source of unnecessary confusion.

But, you know, sadly, once we land this change, our current synopsis won't be quite right either. For example, `strstr()` is classically declared as returning `char*` and taking its first `s1` parameter as `const char*`. That would no longer be valid under our current wording, because the only constraint the standard (and subsequently we, shortly) enforce is that the return type and the first parameter, `big` in our case, shall have the same type. So I don’t think it’s enough to just add a note and leave the synopsis as is.

I’ll add notes for now, but please let me know what you think about this side of the issue.

kfv:
I believe the cleanest is to follow the same synopsis as in the standard. Let me know if you're positive and I'll make the necessary updates.

kfv:
I’ve pushed my recommended change, which I believe is much cleaner than introducing separate qualified and non-qualified pointer types for each function synopsis. Please take a look, and if it looks good, we can squash it into the previous commit before landing.

clausecker:
I'm sorry for my slow response.  It's very hot outside and I find it hard to concentrate.

kfv:
No worries mate, take your time! I can relate, it’s dead hot here as well and AC ain’t helping much either. Hopefully summer passes soon.

sys/sys/cdefs.h
```diff
@@ -227,6 +227,27 @@
...
+ * otherwise compares types exactly.
+ *
+ * The (__uintptr_t) round-trip only suppresses -Wcast-qual on the throwaway
+ * second operand.
+ */
+#if (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L) || \
+    __has_extension(c_generic_selections)
+#define	__qualsel(p, cexpr, expr)					\
+	_Generic(1 ? (p) : (void *)(__uintptr_t)(p),			\
+	    const void *: (cexpr),					\
```
clausecker, line 247: Does this work correctly given that we some times have `char *` and some times `void *`?  I am not super familiar with the semantics of `_Generic`.

kfv: Yes. The ternary collapses both to either `void *` or `const void *`, based on which _Generic selects. The specific pointer type doesn't matter.

clausecker: Great!  Then I think the patch set is fine and will land it with #2188 and #2301 as soon as the latter is ready and my tests pass.

kfv: With the other two PRs already landed, thanks to you, I think we're good to go now.

clausecker:
So it seems like I forgot to land this one, which means you get to do it yourself!  Please squash rebase and prepare commit messages for approval to land.

CC @devinteske

kfv:
Sure, roger that! Thanks.

clausecker:
There's an unrelated commit in your branch which seems to have already landed (?) but apart from that LGTM and approved for commit.

kfv:
Whoops! Sounds like I pushed at exactly the same time Mark’s commit had landed in our git repo but had not yet synced with here. Perhaps I should have done a `pull --rebase` from here to avoid this.

kfv:
Yeah, that was indeed the case. Fixed now.

clausecker:
LGTM.  Approved for commit.

PR #2289, opened 2026-06-20
netinet6: Remove unused rip6_usrreqs declaration

private-yusuke:
The rip6_usrreqs extern declaration has been stale since https://github.com/freebsd/freebsd-src/commit/e7d02be19d40063783d6b8f1ff2bc4c7170fd434. pr_usrreqs is not declared anywhere, and rip6_usrreqs is no longer referenced either.

This PR will remove the stale extern declaration.


glebius (approved):
Pushed

private-yusuke:
Thanks!

This might be a bit off-topic but could you also please look into or assign someone (or a team) for https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=295434 ?

PR #2290, opened 2026-06-21
if_bridge: Remove unused disable parameter in bridge_stop

Acesp25:
The ```int disable``` parameter is included in the bridge_stop function signature but is not used in the function body.

This parameter originally appeared when importing the driver from NetBSD. However, the FreeBSD ifnet library no longer requires an if_stop function. Meaning that the function signature can be changed to only contain needed parameters for our bridge driver.

Discussed with: freebsd-net@ mailing list


gmshake (approved):
Looks good to me. ( reviewed by zlei@ )

Acesp25:
Code has been merged, thanks everybody!

PR #2291, opened 2026-06-21
implemented flag decoding in sysdecode_netlink() using print_mask_int()

androvonx95:
Decode common nlmsg_flags in sysdecode_netlink() instead of always printing hex. Unknown/overloaded bits stay as hex since they depend on message type.

CHANGES:
1.Decode known Netlink base flags in lib/libsysdecode/netlink.c

2.Add a regression test for fully-decoded and mixed flag cases in sysdecode_test.c

### DIFF
--- a/lib/libsysdecode/flags.c
+++ b/lib/libsysdecode/flags.c
@@ -70,7 +70,6 @@
 #include <netgraph/bluetooth/include/ng_l2cap.h>
 #include <netgraph/bluetooth/include/ng_btsocket.h>
 #include <netpfil/pf/pf_nl.h>
-#include <netlink/netlink.h>
 
 #include "support.h"
 
@@ -1218,8 +1217,3 @@ sysdecode_pfnl_cmd(int cmd)
 }
 
 const char *
-sysdecode_nlm_flag(int flag)
-{
-
-	return (lookup_value(nlm_flag, flag));
-}
--- a/lib/libsysdecode/netlink.c
+++ b/lib/libsysdecode/netlink.c
@@ -26,6 +26,32 @@
  * to fallback to a standard hex/string dump.
  */
 
+/*
+ * These nlmsg_flags bits have a unique meaning regardless of the
+ * message direction or operation type.  Bits in the 0x100-0x800 range
+ * are operation-specific and are left as raw hex to avoid ambiguity.
+ */
+static struct name_table netlink_base_flags[] = {
+	{ NLM_F_REQUEST,	"NLM_F_REQUEST" },
+	{ NLM_F_MULTI,		"NLM_F_MULTI" },
+	{ NLM_F_ACK,		"NLM_F_ACK" },
+	{ NLM_F_ECHO,		"NLM_F_ECHO" },
+	{ NLM_F_DUMP_INTR,	"NLM_F_DUMP_INTR" },
+	{ NLM_F_DUMP_FILTERED,	"NLM_F_DUMP_FILTERED" },
+	{ 0,			NULL },
+};
+
+static void
+print_netlink_flags(FILE *fp, uint16_t flags)
+{
+	int rem;
+
+	if (!print_mask_int(fp, netlink_base_flags, flags, &rem))
+		fprintf(fp, "0x%x", rem);
+	else if (rem != 0)
+		fprintf(fp, "|0x%x", rem);
+}
+
 static struct name_table *family_table = NULL;
 static size_t num_family = 0;
 
@@ -79,11 +105,7 @@ sysdecode_netlink(FILE *fp, const void *buf, size_t len, int protocol)
 		}
 
 		fprintf(fp, "flags=");
-		const char *nlm_f = sysdecode_nlm_flag(nl->nlmsg_flags);
-		if (nlm_f != NULL)
-			fprintf(fp, "%s", nlm_f);
-		else
-			fprintf(fp, "0x%x", nl->nlmsg_flags);
+		print_netlink_flags(fp, nl->nlmsg_flags);
 
 		fprintf(fp, ",seq=%u,pid=%u", nl->nlmsg_seq, nl->nlmsg_pid);
 
--- a/lib/libsysdecode/mktables
+++ b/lib/libsysdecode/mktables
@@ -190,7 +190,6 @@ fi
 gen_table "shmflags"  "SHM_[A-Z_]+[[:space:]]+0x[0-9]+"                    "sys/mman.h"         "SHM_ANON"
 gen_table "itimerwhich"     "ITIMER_[A-Z]+[[:space:]]+[0-9]+"              "sys/time.h"
 gen_table_enum "pfnl_cmd"   "PFNL_CMD_[A-Z_]+"                             "netpfil/pf/pf_nl.h"
-gen_table "nlm_flag"        "NLM_F_[A-Z]+[[:space:]]+0x[0-9]+"             "netlink/netlink.h"
 
 # Generate a .depend file for our output file
 if [ -n "$output_file" ]; then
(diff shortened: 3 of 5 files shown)


androvonx95:
@mhorne @bsdimp

androvonx95:
@mhorne @bsdimp I understand you guys are busy and its not easy to review everything fast but if you could take a look at this it would mean a lot to me. Thanks

lib/libsysdecode/netlink.c
```diff
@@ -15,9 +15,36 @@
...
+	{ NLM_F_MULTI,		"NLM_F_MULTI" },
+	{ NLM_F_ACK,		"NLM_F_ACK" },
+	{ NLM_F_ECHO,		"NLM_F_ECHO" },
+	{ NLM_F_DUMP_INTR,	"NLM_F_DUMP_INTR" },
+	{ NLM_F_DUMP_FILTERED,	"NLM_F_DUMP_FILTERED" },
+	{ 0,			NULL },
+};
+
+static void
+print_netlink_flags(FILE *fp, uint16_t flags)
```
mhorne, line 45: Is `sysdecode_netlink()` supposed to call this? Right now it seems to be missing from the diff.

lib/libsysdecode/netlink.c
mhorne: It looks like this existing session attempts to decode the flags field. Does it not work? Or, do I misunderstand the intent?

mhorne (commented):
> @mhorne @bsdimp I understand you guys are busy and its not easy to review everything fast but if you could take a look at this it would mean a lot to me. Thanks

Hi, thanks for the submission. Netlink and libsysdecode are not really my domain... I'll do my best to help.

Can you share an example command where `truss(1)` fails to decode the field?

androvonx95:
@mhorne 

Any command that sets multiple flags in the Netlink header will fail to decode with the current implementation. For example, a standard routing request (like running route monitor or netstat -rn over Netlink) usually sets NLM_F_REQUEST | NLM_F_DUMP (0x01 | 0x300 = 0x301). Because the current implementation uses lookup_value(), it looks for an exact match for 0x301, fails to find one, and falls back to printing raw hex.

Thanks for looking into this

lib/libsysdecode/netlink.c
androvonx95: Yes, it is supposed to call it. I think you are looking at an older version of the PR prior to my latest force-push. The updated diff correctly replaces the old call with print_netlink_flags(fp, nl->nlmsg_flags); inside sysdecode_netlink().

lib/libsysdecode/netlink.c
androvonx95: he existing sysdecode_nlm_flag() function does attempt to decode the flag  but like it doesn't work correctly for bitmasks. It uses lookup_value(), which only checks for an exact integer match but because Netlink flags are frequently combined (like the NLM_F_REQUEST | NLM_F_DUMP example above) the exact match almost always fails in practice, causing it to fall back to raw hex.

My change replaces it with print_mask_int(), which correctly checks and prints the individual bits. I've also removed the old sysdecode_nlm_flag() implementation entirely since it is no longer needed

lib/libsysdecode/flags.c
```diff
@@ -1218,8 +1217,3 @@ sysdecode_pfnl_cmd(int cmd)
 }
 
 const char *
```
mhorne, line 1219: You left something behind :)

mhorne (approved):
LGTM. (I removed the one compile error)

It does indeed seem to work now:
```
$ truss -o out.log netstat -rn
$ grep netlink out.log
sendto(3,netlink{flags=NLM_F_REQUEST|0x300,seq=1,pid=0,len=32,type=18},32,0,NULL,0) = 32 (0x20)
recvmsg(3,{0x3827746f1384,12,netlink{flags=NLM_F_REQUEST|NLM_F_MULTI|0x300,seq=1,pid=0,len=344,type=,flags=NLM_F_REQUEST|NLM_F_MULTI|0x300,seq=1,pid=0,len=320,type=,flags=NLM_F_REQUEST|NLM_F_MULTI|0x300,seq=1,pid=0,len=352,type=,flags=NLM_F_REQUEST|NLM_F_MULTI|0x300,seq=1,pid=0,len=360,type=,flags=0x0,seq=1,pid=0,len=20,type=NLMSG_DONE,<truncated>},1,{},0,0},0) = 1396 (0x574)
sendto(3,netlink{flags=NLM_F_REQUEST|0x300,seq=2,pid=0,len=36,type=26},36,0,NULL,0) = 36 (0x24)
recvmsg(3,{0x3827746f1384,12,netlink{flags=NLM_F_REQUEST|NLM_F_MULTI|0x300,seq=2,pid=0,len=84,type=24,flags=NLM_F_REQUEST|NLM_F_MULTI|0x300,seq=2,pid=0,len=76,type=24,flags=NLM_F_REQUEST|NLM_F_MULTI|0x300,seq=2,pid=0,len=76,type=24,flags=NLM_F_REQUEST|NLM_F_MULTI|0x300,seq=2,pid=0,len=76,type=24,flags=NLM_F_REQUEST|NLM_F_MULTI|0x300,seq=2,pid=0,len=76,type=24,flags=NLM_F_REQUEST|NLM_F_MULTI|0x300,seq=2,pid=0,len=88,type=24,flags=NLM_F_REQUEST|NLM_F_MULTI|0x300,seq=2,pid=0,len=88,type=24,flags=NLM_F_REQUEST|NLM_F_MULTI|0x300,seq=2,pid=0,len=88,type=24,flags=NLM_F_REQUEST|NLM_F_MULTI|0x300,seq=2,pid=0,len=88,type=24,flags=NLM_F_REQUEST|NLM_F_MULTI|0x300,seq=2,pid=0,len=88,type=24,flags=NLM_F_REQUEST|NLM_F_MULTI|0x300,seq=2,pid=0,len=88,type=24,flags=NLM_F_REQUEST|NLM_F_MULTI|0x300,seq=2,pid=0,len=88,type=24,flags=0x0,seq=2,pid=0,len=20,type=NLMSG_DONE,<truncated>},1,{},0,0},0) = 1024 (0x400)
```

Tagging the original author @ishan0803 in case he has any opinion. Otherwise I will merge soon.

As a follow-up change, it seems `sysdecode_netlink()` is missing from `sysdecode_mask(3)` man page. You are welcome to submit a change for that too (or not, if you prefer).

lib/libsysdecode/netlink.c
```diff
@@ -26,6 +26,32 @@
...
+ */
+static struct name_table netlink_base_flags[] = {
+	{ NLM_F_REQUEST,	"NLM_F_REQUEST" },
+	{ NLM_F_MULTI,		"NLM_F_MULTI" },
+	{ NLM_F_ACK,		"NLM_F_ACK" },
+	{ NLM_F_ECHO,		"NLM_F_ECHO" },
+	{ NLM_F_DUMP_INTR,	"NLM_F_DUMP_INTR" },
+	{ NLM_F_DUMP_FILTERED,	"NLM_F_DUMP_FILTERED" },
+	{ 0,			NULL },
+};
```
ishan0803, line 42: Not sure why are we removing NLM_F table generation using mktable and having hardcoded table here?

ishan0803 (changes requested):

kprovost:
As Ishan pointed out, we should keep using gen_table "nlm_flag", but I think the patch does correctly identify and resolve a bug.
Also, I like the addition of the tests.

kprovost:
Also, I missed "Co-authored-by: Cursor <cursoragent@cursor.com>"

The project is still in the process of forumlating a policy on LLM contributions. Mine is "No", and until the project decides something we should not take this.

mhorne:
> Also, I missed "Co-authored-by: Cursor [cursoragent@cursor.com](mailto:cursoragent@cursor.com)"
> 
> The project is still in the process of forumlating a policy on LLM contributions. Mine is "No", and until the project decides something we should not take this.

Generally agreed with Kristof here; whatever you get from an LLM should be reviewed/refined such that it can reasonably be called "your work".

Thanks @androvonx95 for the report and submission. It looks like @ishan0803 addressed the immediate bug in 407b7bcc9386. I found another minor issue with the generated table in #2340.

There remains a valid concern that the current implementation doesn't handle the overlapping/overloaded flag values very well.

PR #2292, opened 2026-06-22
hwpmc: Add the IBS definitions for pmc

mashtizadeh:
Small change to add additional IBS register definitions for the new pmc tools.  Most of the definitions are for Zen 4 and above where we get detailed information regarding the source of a completed memory operation.

Sponsored by: Netflix


mashtizadeh:
@mhorne @bsdimp

mhorne (approved):

PR #2297, opened 2026-06-25
makefs: Fix build on systems without st_birthtime such as Linux

bsdjhb:
Fixes:		0a301f33306c ("makefs cd9660: Populate creation time stamps in RockRidge extensions")


emaste (approved):

bsdjhb:
makefs is kind of busted in that we don't provide nbtool_config.h for builds on FreeBSD, so right now birthtimes don't work for any image which I'll work on fixing separately (*sigh*), but this should unbreak cross builds

PR #2298, opened 2026-06-26
fts: add fts_get_dirfd() accessor for parent directory fd

Jeetubhati:
Add ftsp_dirfd to struct `_fts_private `to store the `fd `of the current directory being traversed. This `fd `is set in `fts_build()` to `_dirfd(dirp)` and provides safe access to the parent directory without exposing an `fd `in the public struct `_ftsent`.

Add `fts_get_dirfd(const FTS *)` as the public accessor. Callers can use `openat(fts_get_dirfd(fts), ent->fts_name, ...)` to access files safely, which works in Capsicum capability mode where path-based operations are not permitted.

Replace statfs(ent->fts_path) with `fstatfs(ftsp_dirfd)` in `fts_ufslinks()` when the `fd ` is available, avoiding a path-based syscall.

Export `fts_get_dirfd ` in `FBSD_1.9 `version map.

Sponsored by:Google LLC (GSoC 2026)

### DIFF
--- a/lib/libc/gen/Symbol.map
+++ b/lib/libc/gen/Symbol.map
@@ -476,6 +476,7 @@ FBSD_1.8 {
 
 FBSD_1.9 {
 	freadlink;
+	fts_get_dirfd;
 	posix_spawn_file_actions_addchdir;
 	posix_spawn_file_actions_addfchdir;
 	posix_spawnattr_getexecfd_np;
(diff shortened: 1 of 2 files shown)


asomers (changes requested):
This PR can't be merged as is since we had to revert #2278.  Could you please combine it with #2278 ?  Also, could you add a mention of fts_get_dirfd to the man page?

Jeetubhati:
> This PR can't be merged as is since we had to revert #2278. Could you please combine it with #2278 ? Also, could you add a mention of fts_get_dirfd to the man page?

Combined the refactor from #2278 with fts_get_dirfd() into a new PR #2303. The new PR includes:

- Internal refactor using `fd`-relative operations
- `fts_dirfd `stored in struct `_fts_private (not public struct)`
- `fts_get_dirfd(const FTS *)` as the public accessor
- `fts_openat()` and `fts_get_dirfd()` documented in `fts(3)`

asomers:
Can we close this, now that #2303 is merged?

PR #2301, opened 2026-06-27
libc: Add strfromd, strfromf, and strfroml per C23

kfv:
`strfromd()`, `strfromf()`, `strfroml()` are thin wrappers over `snprintf()`, which  precisely what the C23 spec says they should be. Jens Gustedt's [reference implementation](https://www.openwall.com/lists/musl/2023/05/31/28), LLVM libc, and glibc all grow their own format-string parsers and invoke internal float converters directly, hence skipping `vfprintf` for a somewhat small but positive overhead gain. To my understanding, we don't expose `gdtoa` or `__ldtoa` as internal APIs and they are tightly woven into vfprintf alongside flag handling, padding, locale, output-buffer protocol etc. Reimplementing that surrounding logic would be an overkill, and any divergence from `vfprintf`'s behaviour such as NaN payloads, rounding edge cases, locale decimal points et al would possibly become a silent conformance bug. Therefore, delegating to `snprintf` means every future correctness or improvements therein is inherited automatically.

The only non-trivial piece is probably the `strfroml`: C23 says the length modifier is determined by the function suffix and not the format string, so manual insertion of `L` is required (as also covered in Jens Gustedt's reference implementation.) I know VLAs are not much encouraged in the base system, but that's honestly what I found most suitable since we would have needed unnecessary conditionals otherwise. This way, the allocation remains on the stack, proportional to the actual format length, no need for further validations (where none is basically required,) and costs nothing noticeable beyond what the digit scan already traversed.

Besides, as C23 §7.24.1.3 declares non-conforming format strings undefined behaviour and imposes no requirements, I did neither the `-1` done by Jens Gustedt nor `abort()` by glibc, since they would create a false portability contract that callers might rely on and other implementations would never honour. Although the `abort()` approach in case of a UB is permitted by the standard, that I find equally a policy choice rather than a requirement, which might cause same set of possible burdens.

Cc: @clausecker


clausecker:
We do actually use `gdtoa` internally in libc (see `contrib/gdtoa` and the implementation of our printf functions; the gdtoa code is built into libc but not exposed as part of the libc interface).  So having these go directly to the `gdtoa` library calls is on the table and likely the better solution in my opinion.

lib/libc/stdlib/strfroml.c
```diff
@@ -0,0 +1,40 @@
...
+	if (fmt[i] == '.')
+		for (++i; (unsigned)(fmt[i] - '0') <= 9u; ++i)
+			continue;
+
+	/*
+	 * Insert 'L' before the conversion specifier so snprintf treats the
+	 * argument as long double as the caller's format intentionally omits
+	 * the length modifier per C23 §7.24.1.3.
+	 */
+	char ff[i + 3];
```
clausecker: Instead of using a VLA, please estimate an upper bound for the length of the `ff` array and use that as a constant length.  Then add bounds checks to ensure we do not exceed it.  VLAs are iffy as they make it impossible to statically analyse the stack usage of a function and I would prefer to avoid them.

lib/libc/stdlib/strfroml.c
```diff
@@ -0,0 +1,40 @@
...
+
+	/*
+	 * Advance past '%' to locate the conversion specifier, then past the
+	 * optional '.' and any decimal digits.  This scan is not validation,
+	 * so use of an invalid format string is UB per C23 §7.24.1.3 and no
+	 * error is signalled.
+	 */
+	i = 1;
+	if (fmt[i] == '.')
+		for (++i; (unsigned)(fmt[i] - '0') <= 9u; ++i)
```
clausecker: Please avoid this trick, it's not immediately obvious, the compiler knows to do this optimisation itself and conversion of a negative signed value to unsigned is technically speaking undefined behaviour anyway.

Maybe

    while (isdigit((unsigned char)fmt[++i])
            ;

instead?  Or avoiding locale questions,

    char c;
    while (c = fmt[++i], '0' <= c && c <= '9')
            ;

lib/libc/stdlib/strfromd.3
```diff
@@ -0,0 +1,95 @@
...
+as does any undefined behaviour inherited from
+.Fn snprintf
+per se.
+.Sh RETURN VALUES
+The
+.Fn strfromd ,
+.Fn strfromf ,
+and
+.Fn strfroml
+functions, as per
```
clausecker, line 77: Maybe

    functions, like
    .Xr snprintf 3 ,

For references to functions from other man pages, please use the `.Xr` request instead of `.Fn`.

lib/libc/stdlib/strfromd.3
```diff
@@ -0,0 +1,95 @@
...
+followed by one of the conversion specifiers
+.Cm a , A , e , E , f , F , g ,
+or
+.Cm G .
+The conversion specifier applies to the type indicated by the
+function suffix, rather than by a length modifier.
+.Pp
+Use of any other format string results in undefined behaviour,
+as does any undefined behaviour inherited from
+.Fn snprintf
```
clausecker: Use `.Xr` instead of `.Fn`.

kfv:
Brilliant, so I convert this PR to draft and I'll take a deeper look tomorrow (or the day after it.)  Thanks for your swift review.

lib/libc/stdlib/strfromf.c
```diff
@@ -0,0 +1,15 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+
+int
+strfromf(char * __restrict s, size_t n, const char * __restrict fmt, float fp)
+{
+
```
clausecker, line 45: We no longer do blank lines before the first statement if no variables are declared, see style(9).

clausecker (changes requested):

lib/libc/stdlib/strfromf.c
kfv: Oh, that I've missed. Thanks!

lib/libc/stdlib/strfromd.3
kfv: Sure.

lib/libc/stdlib/strfromd.3
kfv: Roger that.

lib/libc/stdlib/strfroml.c
kfv: Agreed. Initially thought of utilising the conventional `INT_STRLEN_BOUND`, but I didn't like the fact that in case of an invalid precision we'd have to set it to `INT_MAX` or something. But your point is right and I'd forgotten how importance that is. Thanks.

lib/libc/stdlib/strfroml.c
kfv: > conversion of a negative signed value to unsigned is technically speaking undefined behaviour anyway.

I believe it is not UB according to §6.3.1.3 (though please correct me if I'm wrong.) But I agree that it's a bit tricky, so yeah, perhaps it's better to avoid it.

clausecker:
Oh and it would be useful to have unit tests for these.  Not a must, but helpful in any case.

kfv:
Just pushed the reimplementation commit. Sorry it took longer than expected, it took me quite a while to understand `gdtoa` and our conventions for using it, and besides, there was more work involved than I anticipated since I had to make sure all the edge cases from the standard are correctly covered.

Please take a look when you get a chance and let me know if anything. In the meantime, I'll keep working on unit tests. They're somewhat ready now, to be honest, but I'm not fully happy with them yet, so I'd like to spend a bit more time on them.

Cc: @clausecker

clausecker:
Sorry for the delay, will try to deal with this one soon.

kfv:
It's OK, no worries at all.

lib/libc/stdlib/strfrom.h
```diff
@@ -0,0 +1,462 @@
+/*
```
clausecker, line 1: Why are these in a header file?
This means that the routines are compiled three times into libc, at a waste of space.  It would be better to have them present only once.
You can strop them with two leading underscores to hide them for static linking.

lib/libc/stdlib/strfrom.h
```diff
@@ -0,0 +1,462 @@
...
+	if (n > 0 && (size_t)*pos < n - 1)
+		s[*pos] = c;
+	(*pos)++;
+}
+
+/*
+ * Write len bytes from src.
+ */
+static inline void
+sf_write(char *s, size_t n, int *pos, const char *src, int len)
```
clausecker: You can simplify your logic by having an `sf_puts` routine that appends a NUL-terminated string, as you need that semantic in a variety of places.

lib/libc/stdlib/strfrom.h
```diff
@@ -0,0 +1,462 @@
...
+	c = *p;
+
+	switch (c) {
+	case 'a': case 'A':
+	case 'e': case 'E':
+	case 'f': case 'F':
+	case 'g': case 'G':
+		break;
+	default:
+		abort();
```
clausecker: This seems a bit harsh.  What if a future version of the standard has more format options?  It is probably better to return something clearly bogus in case of parser error (such as `"???"`) so that future or badly written programs don't crash immediately.

If you do wish to crash immediately, I recommend preceding the crash with an error message on `stderr` so that the crash is less mysterious.

lib/libc/stdlib/strfrom.h
```diff
@@ -0,0 +1,462 @@
...
+
+	if (*fmt != '%')
+		abort();
+	p = fmt + 1;
+
+	*prec = -1;
+	if (*p == '.') {
+		*prec = 0;
+		while (*++p >= '0' && *p <= '9')
+			*prec = *prec * 10 + (*p - '0');
```
clausecker: You could use `strtoul` for this.

lib/libc/stdlib/strfrom.h
```diff
@@ -0,0 +1,462 @@
...
+ * is_nan: non-zero if digits[0] == 'N' (NaN), zero for Infinity.
+ */
+static int
+sf_special(char *s, size_t n, char conv, int signflag, int is_nan)
+{
+	int pos;
+	int upper;
+
+	pos = 0;
+	upper = (conv >= 'A' && conv <= 'Z');
```
clausecker: Please use `isupper()` from `ctype.h`.  Same for all the other instances where you e.g. check for a digit.

lib/libc/stdlib/strfrom.h
```diff
@@ -0,0 +1,462 @@
...
+
+#include "../stdio/floatio.h"
+#include "gdtoa.h"
+
+/*
+ * Write c to s[*pos] if within the n-1 writable bytes; always advance *pos.
+ * n is the total buffer capacity, including the null slot.
+ */
+static inline void
+sf_putc(char *s, size_t n, int *pos, char c)
```
clausecker: I do wonder if it would make things easier to put `s`, `n`, and `pos` into a structure that you then pass by reference.

lib/libc/stdlib/strfromd.c
```diff
@@ -0,0 +1,42 @@
...
+#include "strfrom.h"
+
+int
+strfromd(char * __restrict s, size_t n, const char * __restrict fmt, double fp)
+{
+	char conv, lc, *digits, *dend;
+	int prec, decpt, signflag, ret, mode, ndig_req;
+
+	conv = sf_parse_fmt(fmt, &prec);
+	lc = conv | 0x20;
```
clausecker: Please use `tolower()` from `ctype.h`.  Do not make assumptions about the character set except for those given in ANSI C if possible.

clausecker (changes requested):
Finally got some time to review this one.

Looks quite complex, but I guess it has to be.

There are some items that could be improved.
Most importantly, I really don't like half the implementation being in a header that is included thrice.  Perhaps make it a libc-internal source file and strop the external symbols.

kfv:
Hey Robert, thanks for the review! I'll go through your notes tomorrow. I won't be around much today.

lib/libc/stdlib/strfrom.h
kfv: That was on me, right. I’d put helpers in a header file while figuring out how we keep internal routines unexported, and left them there carelessly. Thanks for the catch!

lib/libc/stdlib/strfrom.h
kfv: Done. Added `sf_puts()` for null terminated writes (`Inf`/`NaN` and the locale decimal point). `sf_write()` remains for counted `gdtoa` digit spans, though.

lib/libc/stdlib/strfrom.h
kfv: Done.

lib/libc/stdlib/strfromd.c
kfv: Neat point, right. Done.

lib/libc/stdlib/strfrom.h
kfv: Done.

lib/libc/stdlib/strfrom.h
kfv: I'd prefer not to. `strtoul()`/`isdigit()` are locale-aware, while the format grammar is basic-source digits and must stay equivalent to `snprintf()`; an explicit `'0'`..`'9'` loop keeps that contract.

lib/libc/stdlib/strfrom.h
kfv: I'd prefer to keep bare `abort()` here. Invalid formats are undefined behaviour and terminating is a valid and conforming response to undefined behaviours per standard. It is also an established pattern for impossible paths aborting.

Besides, “print then die” path would pull `stdio` into this `stdlib` conversion for a case that should never run on correct input, and `stderr` is often closed or unreliable by the time we get there.

Delegating wholly to `snprintf()` would have made a softer failure mode easier (and perhaps more natural, I might add); but now with the `gdtoa`-based path the parser is ours, and fail-fast `abort` keeps the contract sharp without extra apparatus. Soft `"???"` is also standard-legal, you're right, but in addition to aforementioned reasons, it would also hide caller bugs, so it'd basically just do more harm than good.

Happy to discuss further if you think otherwise, but I prefer to keep it as-is, if you're OK.

clausecker: Ok, let's get some other opinions.  This is a formatting routine and I strongly expect its spec to be extended in the future.  If we crash now on invalid format, this'll become a minefield for future software trying to use newly defined formats.

bsdphk:
As a general rule in FreeBSD, our libraries should never call abort() unless:

A) There are _no_ other safe options.
or
B) It was explicitly asked for. (for instance in malloc debugging flags)

I personally would simply return the string "EDOOFUS" which is the error code for "Programming error".

lib/libc/stdlib/strfrom.h
kfv: Convinced. I've dropped `abort()` and now treat an invalid format as suggested by phk@.

kfv:
@bsdphk: Thanks for the review. Done as you and @clausecker described.

lib/libc/stdlib/strfrom.h
kfv: Your “clearly bogus” idea is what's done, in fact; just that `EDOOFUS` sat better with me.

clausecker:
Landed in f68d7bfc1479042184e09431bd55771c50c47f68.  Will MFC in a month.

kfv:
Thank you so much, Robert.

PR #2303, opened 2026-06-29
fts: refactor to use fd-relative operations and add fts_get_dirfd()

Jeetubhati:
Combines PR #2278 (reverted) and PR #2298 into a single PR.

- Refactor `fts(3)` internals to use `fd-relative` operations replacing `_open(".")` calls with `_openat()` equivalents
- Add `fts_dirfd` to `struct _ftsent` for per-entry parent directory `fd`
- Add `fts_get_dirfd(const FTS *) `as public accessor stored in `struct _fts_private `— not exposed in public struct
- Replace `statfs()` with `fstatfs()` using `ftsp_dirfd` when available
- Document `fts_openat() `and `fts_get_dirfd()` in `fts(3)` man page

This enables `fts(3`) traversal inside Capsicum capability mode where path-based operations are not permitted after `cap_enter(2)`.

### DIFF
--- a/include/fts.h
+++ b/include/fts.h
@@ -92,6 +92,8 @@ struct _ftsent {
 	char *fts_path;			/* root path */
 	int fts_errno;			/* errno for this node */
 	int fts_symfd;			/* fd for symlink */
+	int fts_dirfd;                  /* fd for parent directory */
+	int __fts_reserved[3];          /* reserved for future use */
 	__size_t fts_pathlen;		/* strlen(fts_path) */
 	__size_t fts_namelen;		/* strlen(fts_name) */
 
--- a/lib/libc/gen/fts-compat15.c
+++ b/lib/libc/gen/fts-compat15.c
@@ -1,4 +1,4 @@
-/*-
+/*
  * SPDX-License-Identifier: BSD-3-Clause
  *
  * Copyright (c) 1990, 1993, 1994
@@ -28,7 +28,11 @@
  * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
  * SUCH DAMAGE.
  *
- * $OpenBSD: fts.c,v 1.22 1999/10/03 19:22:22 millert Exp $
+ */
+/*
+ * Compatibility shim for FBSD_1.5 fts(3) ABI.
+ * This file provides the old fts functions for binaries compiled
+ * against FreeBSD 15.x and earlier without fts_dirfd in FTSENT15.
  */
 
 #include "namespace.h"
@@ -39,6 +43,7 @@
 #include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
+#include "fts-compat15.h"
 #include <fts.h>
 #include <stdalign.h>
 #include <stdint.h>
@@ -54,7 +59,7 @@
 #else
 #include "block_abi.h"
 typedef DECLARE_BLOCK(int, fts_block,
-    const FTSENT * const *, const FTSENT * const *);
+    const FTSENT15 * const *, const FTSENT15 * const *);
 void qsort_b(void *, size_t, size_t, fts_block);
 #endif /* __BLOCKS__ */
 /* only present if linked with blocks runtime */
@@ -62,18 +67,17 @@ void *_Block_copy(const void *) __weak_symbol;
 void _Block_release(const void *) __weak_symbol;
 extern void *_NSConcreteGlobalBlock[] __weak_symbol;
 
-static FTSENT	*fts_alloc(FTS *, char *, size_t);
-static FTSENT	*fts_build(FTS *, int);
-static void	 fts_lfree(FTSENT *);
-static void	 fts_load(FTS *, FTSENT *);
+static FTSENT15	*fts_alloc(FTS15 *, char *, size_t);
+static FTSENT15	*fts_build(FTS15 *, int);
+static void	 fts_lfree(FTSENT15 *);
+static void	 fts_load(FTS15 *, FTSENT15 *);
 static size_t	 fts_maxarglen(char * const *);
-static FTS	*__fts_open(FTS *, char * const *, int);
-static void	 fts_padjust(FTS *, FTSENT *);
-static int	 fts_palloc(FTS *, size_t);
-static FTSENT	*fts_sort(FTS *, FTSENT *, size_t);
-static int	 fts_stat(FTS *, FTSENT *, int, int);
-static int	 fts_safe_changedir(FTS *, FTSENT *, int, char *);
-static int	 fts_ufslinks(FTS *, const FTSENT *);
+static void	 fts_padjust(FTS15 *, FTSENT15 *);
+static int	 fts_palloc(FTS15 *, size_t);
+static FTSENT15	*fts_sort(FTS15 *, FTSENT15 *, size_t);
+static int	 fts_stat(FTS15 *, FTSENT15 *, int, int);
+static int	 fts_safe_changedir(FTS15 *, FTSENT15 *, int, char *);
+static int	 fts_ufslinks(FTS15 *, const FTSENT15 *);
 
 #define	ISDOT(a)	(a[0] == '.' && (!a[1] || (a[1] == '.' && !a[2])))
 
@@ -84,17 +88,17 @@ static int	 fts_ufslinks(FTS *, const FTSENT *);
 #define	FCHDIR(sp, fd)	(!ISSET(FTS_NOCHDIR) && fchdir(fd))
 
 /* fts_build flags */
-#define	BCHILD		1		/* fts_children */
-#define	BNAMES		2		/* fts_children, names only */
-#define	BREAD		3		/* fts_read */
+#define	BCHILD		1		/* freebsd15_fts_children */
+#define	BNAMES		2		/* freebsd15_fts_children, names only */
+#define	BREAD		3		/* freebsd15_fts_read */
 
 /*
- * Internal representation of an FTS, including extra implementation
- * details.  The FTS returned from fts_open points to this structure's
+ * Internal representation of an FTS15, including extra implementation
+ * details.  The FTS15 returned from freebsd15_fts_open points to this structure's
  * ftsp_fts member (and can be cast to an _fts_private as required)
  */
 struct _fts_private {
-	FTS		ftsp_fts;
+	FTS15		ftsp_fts;
 	struct statfs	ftsp_statfs;
 	dev_t		ftsp_dev;
 	int		ftsp_linksreliable;
@@ -129,11 +133,11 @@ static const char *ufslike_filesystems[] = {
 		uint64_t: UINT64_MAX,					\
 		default: 0)
 
-static FTS *
-__fts_open(FTS *sp, char * const *argv, int rootfd)
+static FTS15 *
+__fts_open(FTS15 *sp, char * const *argv)
 {
-	FTSENT *p, *root;
-	FTSENT *parent, *tmp;
+	FTSENT15 *p, *root;
+	FTSENT15 *parent, *tmp;
 	size_t len, nitems;
 
 	/* Logical walks turn on NOCHDIR; symbolic links are too hard. */
@@ -196,7 +200,7 @@ __fts_open(FTS *sp, char * const *argv, int rootfd)
 		root = fts_sort(sp, root, nitems);
 
 	/*
-	 * Allocate a dummy pointer and make fts_read think that we've just
+	 * Allocate a dummy pointer and make freebsd15_fts_read think that we've just
 	 * finished the node before the root(s); set p->fts_info to FTS_INIT
 	 * so that everything about the "current" node is ignored.
 	 */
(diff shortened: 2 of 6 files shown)


include/fts.h
```diff
@@ -133,6 +133,7 @@ struct _ftsent {
 	struct stat *fts_statp;		/* stat(2) information */
 	char *fts_name;			/* file name */
 	FTS *fts_fts;			/* back pointer to main FTS */
+	int fts_dirfd;                  /* fd for parent directory */
```
asomers: You should remove this line.  This field has been moved to the other structure.

lib/libc/gen/fts.3
asomers: Don't forget to change this line.

lib/libc/gen/fts.3
```diff
@@ -599,6 +605,55 @@ except that it takes a block pointer instead of a function pointer.
...
+the behaviour is identical to
+.Fn fts_open .
+Otherwise,
+.Fa dirfd
+must be an open file descriptor referring to a directory,
+and the traversal is rooted there.
+This allows
+.Fn fts_openat
+to be used inside Capsicum capability mode
+.Pq Xr capsicum 4 ,
```
asomers: Just my suggestion, but I would write it like this:
```suggestion
to be used inside
.Xr capsicum 4
capability mode ,
```

lib/libc/gen/fts.c
```diff
@@ -97,6 +99,7 @@ struct _fts_private {
 	struct statfs	ftsp_statfs;
 	dev_t		ftsp_dev;
 	int		ftsp_linksreliable;
+	int             ftsp_dirfd;
```
asomers: Spaces to tabs here

lib/libc/gen/fts.3
```diff
@@ -38,6 +40,8 @@
 .Ft FTS *
 .Fn fts_open "char * const *path_argv" "int options" "int (*compar)(const FTSENT * const *, const FTSENT * const *)"
 .Ft FTS *
+.Fn fts_openat "int dirfd" "char * const *path_argv" "int options" "int (*compar)(const FTSENT * const *, const FTSENT * const *)"
```
asomers: The `fts_openat` function isn't actually included in fts.c in this review, so you shouldn't add it to the man page.

asomers (changes requested):

lib/libc/gen/fts.3
```diff
@@ -29,7 +29,9 @@
 .Dt FTS 3
 .Os
 .Sh NAME
-.Nm fts
+.Nm fts ,
+.Nm fts_openat ,
+.Nm fts_get_dirfd
```
dag-erling: We don't list any of the other functions in the NAME section. You should either add them all or drop this.

include/fts.h
dag-erling, line 146: Missing a prototype for `fts_get_dirfd()` here.

lib/libc/gen/fts.3
```diff
@@ -599,6 +605,55 @@ except that it takes a block pointer instead of a function pointer.
...
+.Pq Xr capsicum 4 ,
+where path-based operations are not permitted after
+.Xr cap_enter 2 .
+The
+.Fn fts_openat
+function duplicates
+.Fa dirfd
+internally, so the caller may close it after
+.Fn fts_openat
+returns.
```
dag-erling: Not in this PR

lib/libc/gen/fts.c
```diff
@@ -67,6 +67,7 @@ static FTSENT	*fts_build(FTS *, int);
 static void	 fts_lfree(FTSENT *);
 static void	 fts_load(FTS *, FTSENT *);
 static size_t	 fts_maxarglen(char * const *);
+static FTS	*__fts_open(FTS *, char * const *, int);
```
dag-erling, line 70: not needed

dag-erling (changes requested):

include/fts.h
Jeetubhati: Kept `fts_dirfd` in `struct _ftsent` as an internal field (like `fts_symfd`) but removed it from the public header documentation. The field is not exposed to callers — `fts_get_dirfd()` is the public accessor.

lib/libc/gen/fts.3
Jeetubhati: Updated to July 7, 2026.

lib/libc/gen/fts.3
Jeetubhati: Fixed per suggestion.

lib/libc/gen/fts.c
Jeetubhati: Fixed.

lib/libc/gen/fts.3
Jeetubhati: Removed. `fts_openat `will be documented in PR #2273.

lib/libc/gen/fts.3
Jeetubhati: Reverted to just .Nm fts since we are not listing all functions.

include/fts.h
Jeetubhati: Added `int fts_get_dirfd(const FTS *)` to` include/fts.h`.

lib/libc/gen/fts.3
Jeetubhati: Removed. Not in this PR.

lib/libc/gen/fts.c
Jeetubhati: Keeping the 3-arg version as it is preparatory for `fts_openat()` in PR #2273. Without `rootfd`, `fts_open(`) cannot pass `AT_FDCWD` to `__fts_open` for the `_openat` call.

include/fts.h
asomers: It's not enough to remove it from the public documentation.  Changing the size of this struct is probably what caused the Poudriere failures that DCH reported.  We need to move fts_dirfd out of this struct entirely.

lib/libc/gen/fts.c
```diff
@@ -81,6 +82,7 @@ static int	 fts_ufslinks(FTS *, const FTSENT *);
 #define	SET(opt)	(sp->fts_options |= (opt))
 
 #define	FCHDIR(sp, fd)	(!ISSET(FTS_NOCHDIR) && fchdir(fd))
+#define FTSP(sp)        ((struct _fts_private *)(sp))
```
asomers: The whitespace here doesn't match the previous line.  Perhaps a tabs/spaces disagreement?

include/fts.h
asomers: I think it should work to move this field into struct `_fts_private`.

Jeetubhati: I moved `fts_dirfd `completely out of `struct _ftsent` into `struct _fts_private` as `ftsp_dirfd`. All per-entry `p->fts_dirfd` references replaced with `FTSP(sp)->ftsp_dirfd`.

lib/libc/gen/fts-compat15.h
```diff
@@ -0,0 +1,97 @@
...
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the University nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
```
asomers, line 17: Usually, for new files you should use a 2-clause BSD license, put yourself as the copyright holder, with the current year.  HOWEVER, since the entire contents of this file are copy-pasted, it's actually appropriate to copy/paste the copyright header, too.

lib/libc/gen/fts-compat15.c
```diff
@@ -0,0 +1,1360 @@
...
+
+#include <dirent.h>
+#include <errno.h>
+#include <fcntl.h>
+#include "fts-compat15.h"
+#include <fts.h>
+#undef FTSENT
+#undef FTS
+#define FTSENT FTSENT15
+#define FTS FTS15
```
asomers: Some places below you explicitly reference FTSENT15 and FTS15.  But here, you also defined the old names to be macros for the new names.  That's inconsistent.  You should do it one way or the other.  I think you should use the new names everywhere.  The macros, I think, might get confusing.

include/fts.h
```diff
@@ -92,6 +92,7 @@ struct _ftsent {
 	char *fts_path;			/* root path */
 	int fts_errno;			/* errno for this node */
 	int fts_symfd;			/* fd for symlink */
+	int fts_dirfd;                  /* fd for parent directory */
```
asomers, line 95: I don't want to go through this whole ELF-symbol-versioning exercise again any time soon.  So I suggest that you add a few more "int __reserved" fields to this struct for future use.  Maybe four ints worth.

lib/libc/gen/fts.c
```diff
@@ -97,6 +99,7 @@ struct _fts_private {
 	struct statfs	ftsp_statfs;
 	dev_t		ftsp_dev;
 	int		ftsp_linksreliable;
+	int		ftsp_dirfd;
```
asomers: Do we even need ftsp_dirfd, since we're adding fts_dirfd to struct FTSENT ?

lib/libc/gen/fts.c
```diff
@@ -711,6 +726,12 @@ fts_set_clientptr(FTS *sp, void *clientptr)
 	sp->fts_clientptr = clientptr;
 }
 
+int
+fts_get_dirfd(const FTS *sp)
+{
+        return (FTSP(sp)->ftsp_dirfd);
```
asomers: Would it be more useful to return FTSENT.fts_dirfd instead?  That way callers could use it to open a file with `openat`.  For that matter, do we even need the accessor function, or can we just document the `fts_dirfd` member in the man page and allow users to use it directly?

asomers (changes requested):

include/fts.h
asomers: On second thought, I regret my recommendation.  On an LP64 architecture, there will now be 4 bytes of padding in the structure.  That's not ideal.  It will still work, but it can be surprising to find padding.  So instead, it would be better to add an odd number of ints' worth of padding.  Could you please change it?

lib/libc/gen/fts-compat15.c
```diff
@@ -0,0 +1,1360 @@
...
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE REGENTS OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ *
+ * $OpenBSD: fts.c,v 1.22 1999/10/03 19:22:22 millert Exp $
```
asomers: Err, even though it's appropriate to copy the Copyright part, this RCS tag has no business being here.  In fact, RCS tags are deprecated everywhere now.  Please remove this one.

lib/libc/gen/fts-compat15.c
asomers: Please don't forget to address this.

asomers (changes requested):
BTW, how did you test this?  The easiest way to test the fts-compat15 stuff is probably to build and install your own ATF tests from _before_ this change.  Then build and install libc (but not its tests) with this change.  The ATF tests should still work.  And you can verify that they're using the correct symbol versions by doing `objdump -T <program>` to see the symbol versions they're linked against.

include/fts.h
Jeetubhati: `__fts_reserved` changed from [4] to [3] to avoid padding on LP64

lib/libc/gen/fts-compat15.c
Jeetubhati: Removed RCS tag.

asomers (approved):
LGTM.

@dch can you please test this revision with whatever you did that broke last time?

asomers:
> LGTM.
> 
> @dch can you please test this revision with whatever you did that broke last time?

Actually, hold off @dch .  We've found another problem.  This change breaks nftw.  Investigating now.

asomers:
> Actually, hold off @dch . We've found another problem. This change breaks nftw. Investigating now.

It was my fault, actually, @dch .  Can you please test this change?

dag-erling:
I won't be able to review these until early next week, please hold off until then.

asomers (approved):

jillest:
How could a simple fts(3) user like bin/chflags/chflags.c be changed to take advantage of this? (So it can work with a current directory that cannot be opened, and as a preparation for Capsicumizing it.)

asomers:
> How could a simple fts(3) user like bin/chflags/chflags.c be changed to take advantage of this? (So it can work with a current directory that cannot be opened, and as a preparation for Capsicumizing it.)

What do you mean "cannot be opened"?  And the answer is that chflags.c should use `fts_openat`, which will be added by PR #2273 .

asomers (approved):
LGTM.  Thanks for adding the HISTORY note, too.

PR #2307, opened 2026-07-03
arm64: fill NULL stream bus_space slots on memmap_bus

b1nc0d3x:
## Summary

`memmap_bus` in `sys/arm64/arm64/bus_machdep.c` leaves every "stream" access slot NULL except the multi variants. Any caller that reaches for e.g. `bus_read_region_stream_2()` on arm64 jumps through a NULL function pointer.

`bhnd(4)` exercises this path. On RockPro64 (RK3399) with a BCM4312 attached under `bwn(4)`, `bhndb_pci_attach()` succeeds and `bhnd_nvram0 <SPROM/OTP>` is announced; the first SPROM read inside `bhnd_nvram_sprom_ident()` then panics:

```
Fatal data abort:
 x9:  memmap_bus
 x16: bhnd_nvram_iores_read+0x0
 lr:  bhnd_nvram_iores_read+0x12c
 elr: 0x0000000000000000
 far: 0x0000000000000000
 esr: 0x86000004         (instruction abort, address 0)
```

Callchain: `bwn_pci_attach → bhndb_attach_bridge → bhndb_pci_attach → bhnd_sprom_attach → bhnd_nvram_store_parse_new → bhnd_nvram_data_new → bhnd_nvram_sprom_new → bhnd_nvram_sprom_ident → bhnd_nvram_iores_read` (dispatching `bus_read_region_stream_2`) → `memmap_bus.bs_rr_2_s == NULL`.

Fill in the NULL slots by aliasing them to their non-stream twins. On aarch64 (little-endian CPU) accessing little-endian PCIe/MMIO, "stream" (no host↔bus swap) and "non-stream" access are byte-identical, so no functional change for existing callers — the multi-stream slots are already handled the same way.

## Test plan

- [x] arm64 kernel build (RockPro64, RP64KERN_KMS)
- [x] `bwn(4)` BCM4312 attach through full bhnd/SIBA enumeration; no NULL-fn panic
- [x] `wpa_supplicant(8)` WPA2/CCMP 4-way handshake completes
- [x] ICMP round-trip over the air to the AP

### DIFF
--- a/sys/arm64/arm64/bus_machdep.c
+++ b/sys/arm64/arm64/bus_machdep.c
@@ -242,11 +242,11 @@ struct bus_space memmap_bus = {
 	.bs_c_4 = NULL,
 	.bs_c_8 = NULL,
 
-	/* read single stream */
-	.bs_r_1_s = NULL,
-	.bs_r_2_s = NULL,
-	.bs_r_4_s = NULL,
-	.bs_r_8_s = NULL,
+	/* read single stream (LE CPU + LE bus: no swap, same as non-stream) */
+	.bs_r_1_s = generic_bs_r_1,
+	.bs_r_2_s = generic_bs_r_2,
+	.bs_r_4_s = generic_bs_r_4,
+	.bs_r_8_s = generic_bs_r_8,
 
 	/* read multiple stream */
 	.bs_rm_1_s = generic_bs_rm_1,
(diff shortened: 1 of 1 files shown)


sys/arm64/arm64/bus_machdep.c
```diff
@@ -242,41 +242,41 @@ struct bus_space memmap_bus = {
...
 	.bs_rm_2_s = generic_bs_rm_2,
 	.bs_rm_4_s = generic_bs_rm_4,
 	.bs_rm_8_s = generic_bs_rm_8,
 
-	/* read region stream */
-	.bs_rr_1_s = NULL,
-	.bs_rr_2_s = NULL,
-	.bs_rr_4_s = NULL,
-	.bs_rr_8_s = NULL,
+	/* read region stream (LE CPU + LE bus: no swap, same as non-stream) */
```
zxombie, line 257: You don't need to update these comments, we don't have this extra comment for `write multiple stream` below

PR #2308, opened 2026-07-03
vt: reprogram display controller on KDSETMODE KD_TEXT for active window

Defenso-QTH:
When a graphical application (X.Org display server, Wayland compositor) calls KDSETMODE to set the active VT back to KD_TEXT, vt(4) does not reset the CRTC to point to its framebuffer, leaving the last image of the graphical application visible instead of the text console until the next VT switch.

Add a call to vd_postswitch to reset the CRTC, set VDF_INVALID to force a redraw and schedule the flush timer, like vt(4) does in vt_window_switch().

### DIFF
--- a/sys/dev/vt/vt_core.c
+++ b/sys/dev/vt/vt_core.c
@@ -2924,6 +2924,23 @@ skip_thunk:
 		case KD_TEXT:
 		case KD_TEXT1:
 		case KD_PIXEL:
+			if ((vw->vw_flags & VWF_GRAPHICS) &&
+			    vw == vd->vd_curwindow) {
+				/*
+				 * When leaving graphics mode on the currently
+				 * active window, reprogram the display
+				 * controller back to text mode immediately.
+				 * Without this, the CRTC retains the graphical
+				 * state and the console stays invisible until
+				 * the next VT switch.
+				 */
+				VT_LOCK(vd);
+				vd->vd_flags |= VDF_INVALID;
+				VT_UNLOCK(vd);
+				if (vd->vd_driver->vd_postswitch)
+					vd->vd_driver->vd_postswitch(vd);
+				vt_resume_flush_timer(vw, 0);
+			}
 			vw->vw_flags &= ~VWF_GRAPHICS;
 			break;
 		case KD_GRAPHICS:
--- a/sys/dev/vt/vt_core.c
+++ b/sys/dev/vt/vt_core.c
@@ -2923,9 +2923,31 @@ skip_thunk:
 		switch (*(int *)data) {
 		case KD_TEXT:
 		case KD_TEXT1:
-		case KD_PIXEL:
+		case KD_PIXEL: {
+			bool restore = false;
+			VT_LOCK(vd);
+			if ((vw->vw_flags & VWF_GRAPHICS) &&
+			    vw == vd->vd_curwindow) {
+				/*
+				 * When leaving graphics mode on the currently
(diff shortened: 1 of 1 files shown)


sys/dev/vt/vt_core.c
```diff
@@ -2924,6 +2924,23 @@ vtterm_ioctl(struct terminal *tm, u_long cmd, caddr_t data,
 		case KD_TEXT:
 		case KD_TEXT1:
 		case KD_PIXEL:
+			if ((vw->vw_flags & VWF_GRAPHICS) &&
+			    vw == vd->vd_curwindow) {
```
VexedUXR, line 2930: `VT_LOCK` should be held before this check to avoid a race with `vt_window_switch`.

VexedUXR (changes requested):
I think this patch is okay, but `VT_LOCK` should called before accessing `vd->vd_curwindow`.

Looking at the drm-kmod side of things, linux passes a "force" flag to `__drm_fb_helper_restore_fbdev_mode_unlocked` when switching modes in this case [1]. That flag is always false on FreeBSD. I'm not sure what the repercussions for this are...

[1] https://github.com/freebsd/drm-kmod/blob/8378b4d738751/drivers/gpu/drm/drm_fb_helper.c#L1384

sys/dev/vt/vt_core.c
Defenso-QTH: Done.

Defenso-QTH (commented):
Moved the lock before accessing `vd_curwindow`.

Defenso-QTH:
> Looking at the drm-kmod side of things, linux passes a "force" flag to `__drm_fb_helper_restore_fbdev_mode_unlocked` when switching modes in this case [1]. That flag is always false on FreeBSD. I'm not sure what the repercussions for this are...
> 
> [1] https://github.com/freebsd/drm-kmod/blob/8378b4d738751/drivers/gpu/drm/drm_fb_helper.c#L1384

From what I understand, Linux needs this because `systemd-logind` keeps DRM master even when applications close. On FreeBSD, we do not have `logind`. `xorg` does drop master before issuing KDSETMODE, and so does `seatd`.

### `xorg`
https://gitlab.freedesktop.org/xorg/xserver/-/blob/master/hw/xfree86/common/xf86Init.c#L805
In `ddxGiveUp`:
* `xf86Screens[i]->LeaveVT` drops DRM master
* `xf86CloseConsole` sets KD_TEXT

### `seatd`
https://git.sr.ht/~kennylevinsen/seatd/tree/9d1f83e2825442c4aeca467160c918f7ff128392/item/seatd/seat.c#L254
* `seat_close_device` drops DRM master
* `vt_close` sets KD_TEXT

Defenso-QTH:
Following additional testing, I confirm the bug only manifests when several applications (e.g. wayland compositors) are holding holding fds to the drm device and the currently displayed one is closed.

This can be explained by this check in `drm-kmod`: https://github.com/freebsd/drm-kmod/blob/512101d90dcd2fc69419986d898f0cee4e83042d/drivers/gpu/drm/drm_file.c#L438

I still believe this is a `vt` bug worth patching. `vt` is supposed to tell the graphics driver what to do, not let it figure it out in its own.

emaste:
I've added this to my staging tree, and expect to take a look soon.

VexedUXR (approved):
I think this looks good.

Thanks for putting up with my quest to figure out what exactly was going on!

PR #2309, opened 2026-07-05
libc: Fix stack buffer overflow in cgetnext()

kfv:
`cgetnext()` copies a capability record's name field into a fixed 1024 byte on-stack buffer (`nbuf[BSIZE]`) one character at a time, but never checks that the write pointer stays within the buffer. Therefore, a record name that is longer than `BSIZE` (either on a single line, or accumulated across several `\`-continued lines as the pointer is not reset amidst continuation) writes past the end of `nbuf` and smashes the stack.

The name field is read directly from the capability file, so any consumer of `cgetnext()`/`cgetfirst()` that processes an untrusted `termcap(5)`- or `printcap(5)`-style database (e.g. `cap_mkdb(1)`, `lpr(1)`, or any third-party caller) can be exposed and hence made to corrupt the stack.

### DIFF
--- a/lib/libc/gen/getcap.3
+++ b/lib/libc/gen/getcap.3
@@ -28,7 +28,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd March 22, 2002
+.Dd July 23, 2026
 .Dt GETCAP 3
 .Os
 .Sh NAME
@@ -549,6 +549,19 @@ as follows:
 .Bl -tag -width Er
 .It Bq Er ENOMEM
 No memory to allocate.
+.El
+.Pp
+The
+.Fn cgetfirst
+and
+.Fn cgetnext
+functions
+may fail and set
+.Va errno
+as follows:
+.Bl -tag -width Er
+.It Bq Er ENAMETOOLONG
+The name field of a capability record is too long.
 .El
 .Sh SEE ALSO
 .Xr cap_mkdb 1 ,
--- a/lib/libc/gen/getcap.c
+++ b/lib/libc/gen/getcap.c
@@ -709,6 +709,11 @@ cgetnext(char **bp, char **db_array)
 		np = nbuf;
 		for (;;) {
 			for (cp = line; *cp != '\0'; cp++) {
+				if (np >= nbuf + sizeof(nbuf) - 1) {
+					(void)cgetclose();
+					errno = ENAMETOOLONG;
+					return (-1);
+				}
 				if (*cp == ':') {
 					*np++ = ':';
 					done = 1;


clausecker:
That seems like a sensible fix, but `ENAMETOOLONG` is probably the wrong error message.  Maybe `ENOMEM`?  Or `EBADMSG`?  Or when in doubt, `EINVAL` always works.

kfv:
I see your concern, but this wouldn't be the first place where we use `ENAMETOOLONG` for something other than a file name; [gethostname.c](https://github.com/freebsd/freebsd-src/blob/759ce9a2b38e1de70c14c81dee7e245bf0bc6b94/lib/libc/gen/gethostname.c#L46-L50), [ns_name.c](https://github.com/freebsd/freebsd-src/blob/759ce9a2b38e1de70c14c81dee7e245bf0bc6b94/lib/libc/nameser/ns_name.c#L822-L826), and [setlocale.c](https://github.com/freebsd/freebsd-src/blob/759ce9a2b38e1de70c14c81dee7e245bf0bc6b94/lib/libc/locale/setlocale.c#L312-L314), to name a few. So I'd personally keep `ENAMETOOLONG`, as it's the closest match semantically and communicates the actual cause of the failure more clearly.

ngie-eign (commented):
This looks reasonable to me. Could you please update the manpage to account for the new error condition?

kfv:
Done.

lib/libc/gen/getcap.3
```diff
@@ -550,6 +550,19 @@ as follows:
...
+.Fn cgetfirst
+and
+.Fn cgetnext
+functions
+may fail and set
+.Va errno
+as follows:
+.Bl -tag -width Er
+.It Bq Er ENAMETOOLONG
+Capability name fits not in the name buffer.
```
ngie-eign: Some minor word smithing:
```suggestion
One or more capabilties will not fit in
.Fa buf .
```

ngie-eign (commented):
@concussious : does the manpage change seem good?

ngie-eign (commented):
I'm a bit on the fence about `ENOMEM` and `ENAMETOOLONG`, given that this is actually an issue with an internally allocated buffer on the stack, not an issue with [direct] user provided input.

lib/libc/gen/getcap.3
kfv: The buffer holds the record's name field; the `name|alias|…` list before the first `:`.  So "the name field of a capability record" reads better, I believe. What do you think?

kfv:
> I'm a bit on the fence about `ENOMEM` and `ENAMETOOLONG`, given that this is actually an issue with an internally allocated buffer on the stack, not an issue with [direct] user provided input.

The buffer is internal, yes, but the data filling it isn't. The name field is read verbatim from the capability database, so an untrusted external input. Besides, `errno` is consumer-facing, so it should describe the failure accurately.

`ENOMEM` would mislead, implying an allocation failed when none did.  `ENAMETOOLONG` is the established idiom for "a name exceeded a length limit" whether that limit is the caller's or the implementation's.

See [`realpath(3)`](https://man.freebsd.org/realpath(3)), [`readlink(2)`](https://man.freebsd.org/readlink(2)), and [`lstat(2)`](https://man.freebsd.org/lstat(2)) for similar cases.

PR #2316, opened 2026-07-08
sched_ule: add per-entity load tracking (PELT)

rdmsr:
## Summary
The ULE scheduler currently relies on a simple thread count (`tdq_count`) for load balancing decisions, this is simple but does not take into account the nature of the threads running on each CPU. As such, a busier CPU, with say, 10 CPU-bound threads, would still get selected over a CPU with 11 IO-bound threads. Looking at existing techniques, per-entity load tracking (PELT) from Linux seemed to fit the simplicity of ULE and it had already been proven in production. This patch adds support for it via a per-thread load average value updated whenever a thread's state is updated, this per-thread load average is summed into its queue's own load average and is used to break ties between CPUs when making decisions for load balancing. This is not really meant to replace `tdq_load` entirely for load-balancing decisions but rather as an additional heuristic which could provide better decisions.

Differences with Linux:
- The PELT state is updated on a per-tick basis instead of per-period (1024us) like Linux, this might be slightly less accurate than Linux's metrics, but this considerably simplifies the implementation and ULE already works on a per-tick basis, so this *should* be good enough.
- Only the 'runnable' (i.e. running or in run queue) state is kept track of, Linux also keeps track of CPU utilization and a load average, the former would be useful for CPU frequency scaling and energy-aware scheduling, and the latter is essentially the same metric we already keep track of (runnable) but weighted according to CFS's weights.

See the comments in the code that goes over the math.

## Test plan
I tested this functionality via sysctl hook that prints the `tdq_load_avg` of each CPU:
```c
static int
sysctl_kern_sched_load_avg(SYSCTL_HANDLER_ARGS)
{
  struct sbuf *sb;
  struct tdq *tdq;
  const char *sep;
  int error, i;

  error = sysctl_wire_old_buffer(req, 0);
  if (error != 0)
    return (error);
  sb = sbuf_new_for_sysctl(NULL, NULL, 128, req);
  if (sb == NULL)
    return (ENOMEM);
  sep = "";
  CPU_FOREACH(i) {
    tdq = TDQ_CPU(i);
    sbuf_printf(sb, "%scpu%d %d/%d", sep, i, TDQ_LOAD(tdq),
        TDQ_LOAD_AVG(tdq));
    sep = " ";
  }
  error = sbuf_finish(sb);
  sbuf_delete(sb);
  return (error);
}
```

I then tested a variety of CPU-bound and IO-bound programs to check if the load averages were getting properly updated.
 I verified there were no regressions by timing a full kernel build with and without PELT, and noting the differences were minimal (in fact, the build with PELT seemed to have been shorter by a second).
It would be helpful to test and tune parameters on a larger machine (I only have a small desktop CPU) to see if there is a meaningful improvement in a certain system metric.

### DIFF
--- a/sys/kern/sched_ule.c
+++ b/sys/kern/sched_ule.c
@@ -96,6 +96,8 @@ struct td_sched {
 	u_int		ts_slptime;	/* Number of ticks we vol. slept */
 	u_int		ts_runtime;	/* Number of ticks we were running */
 	u_int		ts_ticks;	/* pctcpu window's running tick count */
+	u_int		ts_load_avg;	/* decayed runnable average */
+	u_int		ts_lavg_tick;	/* tick of last PELT update */
 #ifdef KTR
 	char		ts_name[TS_NAME_LEN];
 #endif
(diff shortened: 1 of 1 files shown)


mashtizadeh:
Linux and BSD’s load balancing already behave pretty differently depending on which scheduler your looking at.  FreeBSD’s rebalances more slowly (a thread at a time) leading to better long term fairness, considers the cache hierarchy, and cpu load.  I think we need a set of benchmarks to show a measurable improvement while maintaining fairness.

https://www.usenix.org/conference/atc18/presentation/bouron

rdmsr:
> Linux and BSD’s load balancing already behave pretty differently depending on which scheduler your looking at. FreeBSD’s rebalances more slowly (a thread at a time) leading to better long term fairness, considers the cache hierarchy, and cpu load. I think we need a set of benchmarks to show a measurable improvement while maintaining fairness.
> 
> https://www.usenix.org/conference/atc18/presentation/bouron

I'm not sure I understand, the idea behind this patch is to try and make the "cpu load" metric a bit more accurate/fairer while keeping all the other FreeBSD policies.
I'm not entirely sure what kind of benchmarks would work here.

rdmsr:
Well, I have ran a suite of benchmarks comprising mostly of an hackbench-like thing, and have measured only marginal (~1-1.5%) improvements with this patch turned on. I will close this PR for now as I am not happy with the overall quality, but I believe a feature like this could still be useful perhaps under a finer-grained tracking (time unit instead of ticks) and on SMT systems.

PR #2317, opened 2026-07-09
netmap: fix VALE offset handling in nm_vale_flush()

TeddyEngel:
Backport of upstream netmap commits ef8fa11e3 and e0de5b529 by Giuseppe Lettieri, fixing multiple offset handling bugs in nm_vale_flush():

- Inverted cache-line alignment comparison (`dst_co < src_co` → `src_co < dst_co`) could place data before the requested offset in the destination buffer.
- nm_write_offset() called unconditionally, silently failing on rings with offset_mask == 0 while pkt_copy already placed data at a non-zero offset the consumer can't read.
- No fallback for destinations without offset support — now uses plain memcpy when offset_mask == 0 and source is unaligned.

Tested on FreeBSD 16.0-CURRENT (DEBUG kernel) with VALE vp-to-vp and VALE-to-iflib forwarding using various buffer offsets (0–63). Confirmed the bug reproduces on the unpatched kernel and is fixed by this patch.

PR: [294491](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=294491)

### DIFF
--- a/sys/dev/netmap/netmap_vale.c
+++ b/sys/dev/netmap/netmap_vale.c
@@ -974,7 +974,7 @@ retry:
 					char *dst, *src = ft_p->ft_buf;
 					size_t copy_len = ft_p->ft_len, dst_len = copy_len;
 					uintptr_t src_cb;
-					uint64_t dstoff, dstoff_cb;
+					uint64_t dstoff = 0, dstoff_cb = 0;
 					int src_co, dst_co;
 					const uintptr_t mask = NM_BUF_ALIGN - 1;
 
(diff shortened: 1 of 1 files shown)


TeddyEngel:
(The style warnings seem consistent with the code surrounding the changes - which are already at 80+ character lines due to deep nesting)

jlduran (approved):
These look good, thank you @TeddyEngel and @giuseppelettieri! The differences with upstream are starting to grow. I'll start submitting the trivial ones first.

PR #2318, opened 2026-07-09
igc: Disable ASPM L1.2 on I226 to prevent RX stalls

michael-adler:
I226 parts advertise support for the PCIe L1.2 link substate, but a hardware erratum makes the exit latency from that low-power state longer than the packet buffer can absorb under load. This stalls the inbound packet stream. Disabling ASPM system-wide (BIOS or OS ASPM policy) does not fix it. The L1.2 enable bit must be cleared directly in the device's own PCIe L1 PM extended capability.

Add igc_is_device_id_i226() to identify affected parts and igc_disable_broken_aspm_l1_2() to clear the ASPM L1.2 enable bit on attach and after resume, since PCIe config space can be reset across a suspend/resume cycle.

Add register/bit definitions for the L1 PM substates capability (PCIZ_L1PM) to pcireg.h.

Ported from the Linux igc driver:

  0325143b59c6 igc: disable L1.2 PCI-E link substate to avoid
               performance issue
  1468c1f97cf3 igc: fix disabling L1.2 PCI-E link substate on I226
               on init

Fixes: Bugzilla #279245 - igc(4) I226 (and I225) TX hangups


michael-adler:
Also applies cleanly to releng/15.1

markjdb (commented):
Thank you for the patch. I've forwarded this on to some users who are seeing the problem.

A couple of notes on the commit itself:
- Please don't use the Fixes tag for this, that field is used to link commits together. To reference a bugzilla issue just use `PR: 279245`.
- Please modify pcireg.h in a separate commit

AdSchellevis:
While looking at a similar issue on our end, it seems that a very simple and fast way to reproduce the issue is to toggle the `dev.igc.0.eee_control` switch, this calls `igc_if_init(ctx);` as well, which for as far as I can find, isn't allowed in this scope (and kills the rx path). 

I just want to mention I'm not an expert here, but what we're trying at the moment is 
https://github.com/opnsense/src/commit/effb6c32aed12ed8e7a84bfc55236f854e04dd24 (commit message contains steps to reproduce). 

This PR simply skips `igc_if_resume()`, which might be correct, but if that's the case, should highly likely do the exact same thing for the i225.

michael-adler:
> While looking at a similar issue on our end, it seems that a very simple and fast way to reproduce the issue is to toggle the `dev.igc.0.eee_control` switch, this calls `igc_if_init(ctx);` as well, which for as far as I can find, isn't allowed in this scope (and kills the rx path).
> 
> I just want to mention I'm not an expert here, but what we're trying at the moment is [opnsense/src@effb6c3](https://github.com/opnsense/src/commit/effb6c32aed12ed8e7a84bfc55236f854e04dd24) (commit message contains steps to reproduce).

Are you sure this is similar? I read your patch as a replacing init with reset on a few paths triggered by user commands that adjust the device. The patch here disables a broken HW ASPM feature that doesn't wake up quickly enough to avoid overrun on the RX stream. The two patches seem to be solving different problems, though share the same RX hang. Am I missing something?

AdSchellevis:
I'm certainly not sure it's fully similar, but calling `igc_if_init()` seems to lead to the same result (a broken rx path). From what I could find, using a bit of AI as well, the `igc_if_init()` shouldn't be called directly as its dispatched to `iflib`.

As I don't have a reproducible case, I can't debug this myself, but there is something fishy going on. 

Out of curiosity, did you add logging during your debugging? It would be interesting to know if this path is triggered commonly when there is an issue (and after the initial call when it changed the aspm settings), I assume your patch solved the problem you had when it occurred.

It looks like I misread part of your patch earlier, which led me to believe the init was skipped in the i226 case now,  which obviously isn't the case. 

So far I could only test using the manual toggle, which indeed triggers the same behavior, but isn't the exact same thing. By tracing the path using a lot of extra logging in the various callers I found the `igc_msix_link` wasn't called when flipping with the  `dev.igc.0.eee_control` setting and was being called when a regular down/up happened.

But again, I'm certainly no expert on the topic.

gonzopancho:
After reading the R2.2 spec update for I225/I226, I have doubts that this applies *at all* to i226, in fact, said spec update says it's "Fixed in I226".

michael-adler:
> After reading the R2.2 spec update for I225/I226, I have doubts that this applies _at all_ to i226, in fact, said spec update says it's "Fixed in I226".

@gonzopancho - Are you referring to IGC_DEV_ID_I226_LMVP and IGC_DEV_ID_I226_BLANK_NVM? If so, you're probably right. I missed that those aren't in the Linux igc_is_device_id_i226().

kev009:
@gonzopancho is directly citing that this doesn't apply to any I226 per latest intel documentation, which I can confirm is what it says.

But counter to the documentation, the one related thing Linux does is disable PCIE_LINK_STATE_L1_2 for I226, which is only the ASPM part (not PCIM_L1PM_CTL1_PCIPM_L1_2).

Are you actually seeing an improvement for this or just trying to match what Linux does?

michael-adler:
I see the bug on i226-V (8086:125c) with the latest firmware. Without this patch, Ethernet RX eventually hangs.

@kev009 you are right about Linux only disabling the ASPM part. I will retest without changing PCI-PM, remove the two i226 devices not in the Linux version, and update this PR with the result.

michael-adler:
@kev009 and @gonzopancho - updated patch to reduce the set of devices and change only the ASPM L1.2 bit. The machine remains stable.

Thank you for the feedback.

PR #2320, opened 2026-07-12
mbuf: Fix escape sequence for M_PROTO11 in M_FLAG_PROTOBITS, sync mbuf(9) content

private-yusuke:
- Fix the number in escape sequence for M_PROTO11 in M_FLAG_PROTOBITS
- Replace the space with a tab in the definition of M_TSTMP_LRO
- Sync mbuf flags, types and external buffer types of mbuf(9) with the current mbuf.h

I checked the changes do not bring new warnings with igor(1).

As the CONTRIBUTING.md states the following, I would like you to change the date written in mbuf.9 before committing the changes:

> Proposed changes to manual pages should not bump the document date until merged.

### DIFF
--- a/sys/sys/mbuf.h
+++ b/sys/sys/mbuf.h
@@ -478,7 +478,7 @@ m_epg_pagelen(const struct mbuf *m, int pidx, int pgoff)
 #define	M_TSTMP_HPREC	0x00000800 /* rcv_tstmp is high-prec, typically
 				      hw-stamped on port (useful for IEEE 1588
 				      and 802.1AS) */
-#define M_TSTMP_LRO	0x00001000 /* Time LRO pushed in pkt is valid in (PH_loc) */
+#define	M_TSTMP_LRO	0x00001000 /* Time LRO pushed in pkt is valid in (PH_loc) */
 
 #define	M_PROTO1	0x00002000 /* protocol-specific */
 #define	M_PROTO2	0x00004000 /* protocol-specific */
@@ -521,7 +521,7 @@ m_epg_pagelen(const struct mbuf *m, int pidx, int pgoff)
 #define	M_FLAG_PROTOBITS \
     "\16M_PROTO1\17M_PROTO2\20M_PROTO3\21M_PROTO4" \
     "\22M_PROTO5\23M_PROTO6\24M_PROTO7\25M_PROTO8\26M_PROTO9" \
-    "\27M_PROTO10\28M_PROTO11"
+    "\27M_PROTO10\30M_PROTO11"
 #define	M_FLAG_PRINTF (M_FLAG_BITS M_FLAG_PROTOBITS)
 
 /*
--- a/share/man/man9/mbuf.9
+++ b/share/man/man9/mbuf.9
@@ -215,7 +215,7 @@ flag bits are defined as follows:
 #define	M_EXT		0x00000001 /* has associated external storage */
 #define	M_PKTHDR	0x00000002 /* start of record */
 #define	M_EOR		0x00000004 /* end of record */
-#define	M_RDONLY	0x00000008 /* associated data marked read-only */
+#define	M_RDONLY	0x00000008 /* associated data is marked read-only */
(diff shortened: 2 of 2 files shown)


sys/sys/mbuf.h
```diff
@@ -478,7 +478,7 @@ m_epg_pagelen(const struct mbuf *m, int pidx, int pgoff)
 #define	M_TSTMP_HPREC	0x00000800 /* rcv_tstmp is high-prec, typically
 				      hw-stamped on port (useful for IEEE 1588
 				      and 802.1AS) */
-#define M_TSTMP_LRO	0x00001000 /* Time LRO pushed in pkt is valid in (PH_loc) */
+#define	M_TSTMP_LRO	0x00001000 /* Time LRO pushed in pkt is valid in (PH_loc) */
```
concussious, line 481: It looks like you put the tab in the wrong spot.

sys/sys/mbuf.h
```diff
@@ -521,7 +521,7 @@ m_epg_pagelen(const struct mbuf *m, int pidx, int pgoff)
 #define	M_FLAG_PROTOBITS \
     "\16M_PROTO1\17M_PROTO2\20M_PROTO3\21M_PROTO4" \
     "\22M_PROTO5\23M_PROTO6\24M_PROTO7\25M_PROTO8\26M_PROTO9" \
-    "\27M_PROTO10\28M_PROTO11"
+    "\27M_PROTO10\30M_PROTO11"
```
concussious, line 524: This change looks correct to me, but I don't know if anything relies on the incorrect behavior or if something complicated is going on. I will ask some senior colleagues to take a look.

concussious (approved):
Thank you for doing this. I see the M_PROTOs were changed in e57b2d0e51265f7f1603077e0ca5af6c8e450751 for FreeBSD 13.0, but we forgot to fix the manual. I accept the man page change and will merge it. Let me ask a senior colleague to review the kernel fix.

sys/sys/mbuf.h
private-yusuke: This change is intended as it would align the `#define` and the latter parts in the same way as other definitions in this file.

sys/sys/mbuf.h
private-yusuke: Thank you for the help! I look forward to hearing from them.

To my understanding, `M_FLAG_PROTOBITS` ultimately will only be used in bce_dump_mbuf() if `BCE_DEBUG` is defined, which should not happen unless you are debugging with mbuf.

private-yusuke (commented):
Thank you for your prompt assistance.

sys/sys/mbuf.h
concussious: I see now, they are all tabs. Sorry, tabs and spaces are hard sometimes.

sys/sys/mbuf.h
concussious: That explains why no one has noticed this yet

ngie-eign: @private-yusuke : could you please help me understand why `\28` is incorrect and `\30` is correct? Are you repeating the jump from `\17` -> `\20`?

concussious: These numbers are octal, so octal 30 is the next number after octal 27. I checked them with `jot -w %o 30 1 | cat -n`.

private-yusuke: The numbers `nnn` appearing in an escape sequence`\nnn` in C would be interpreted as octal.

[printf(9)](https://man.freebsd.org/cgi/man.cgi?query=printf&apropos=0&sektion=9&manpath=FreeBSD+16.0-CURRENT&format=html) manpage also mentions about this indirectly.

> Each bit	identifier be-
>        gins with a character specifying	the number of the bit  (starting  from
>        1)  this	 identifier  describes.	 The characters	from \01 to \40	can be
>        used to specify bit numbers in the range	from 1 to  32  and  characters
>        from  \200  to  \377 to specify bit numbers in the range	from 1 to 128.

PR #2319, opened 2026-07-09
virtio: Add feature bits up to VirtIO v1.3 and accept VIRTIO_F_RING_RESET in the modern PCI transport

kfv:
FreeBSD/arm64 guests under macOS `Virtualization.framework` boot to multiuser but have no network: Apple's `virtio-net` rejects `FEATURES_OK` when `VIRTIO_F_RING_RESET` (bit 40) is not acknowledged, and we do not define bits above 33.

This PR adds the VirtIO 1.1+ transport feature bit definitions, acknowledges `RING_RESET` in the modern PCI transport when offered (safe: it only permits per-`virtqueue` reset), and prints the host/written masks on negotiation failure so such mismatches are diagnosable.

I have tested arm64 GENERIC under `QEMU`/`HVF` (no regressions; all `virtio` devices attach and networking works) and under `Virtualization.framework` on macOS/arm64 (negotiation proceeds past the `RING_RESET` rejection).

For Apple, note that this is a prerequisite rather than a complete fix. I'm currently working on the remaining issues and will address them in a separate PR once they're resolved.

PR #2322, opened 2026-07-12
vtnet: Feature negotiation fallback, GUEST_ANNOUNCE, and CTRL_RX_EXTRA

kfv:
This is the follow-up I promised in #2319. That PR dealt with the transport side of the problem (the missing VirtIO 1.1+ feature bit definitions and `RING_RESET`), but as noted, it was just a prerequisite rather than a complete fix for Apple's `Virtualization.framework`. This all started while I was experimenting with `Virtualization.framework`, and then it grew into a larger piece of work.

I tested on an M1 Pro MacBook under both `Virtualization.framework` (vtnet now attaches correctly and gets a DHCP lease with the default configuration) and QEMU (no regressions; the fallback never triggers there).

But I should say up front that although I've done my best to follow the spec and existing driver conventions, VirtIO internals and the network stack aren't my usual territory, so please educate me if there's anything I've missed or done wrong.

### DIFF
--- a/sys/dev/virtio/network/if_vtnet.c
+++ b/sys/dev/virtio/network/if_vtnet.c
@@ -758,7 +758,7 @@ vtnet_negotiate_features(struct vtnet_softc *sc)
 {
 	device_t dev;
 	uint64_t features, negotiated_features;
-	int no_csum;
+	int error, no_csum;
 
 	dev = sc->vtnet_dev;
 	features = virtio_bus_is_modern(dev) ? VTNET_MODERN_FEATURES :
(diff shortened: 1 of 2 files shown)


sys/dev/virtio/network/if_vtnet.c
```diff
@@ -4044,6 +4064,86 @@ vtnet_ctrl_mq_cmd(struct vtnet_softc *sc, uint16_t npairs)
 	return (error);
 }
 
+static int
```
erikarn, line 4067: This looks like a copy/paste from vtnet_ctrl_mq_cmd() and maybe other functions? Maybe we should refactor those out first into a generic "send this command to the ring" routine so we do the work in one function?

erikarn (changes requested):
looks good other than the duplicate code. Do you want to tackle de-duplicating it now or in a subsequent pull request?

sys/dev/virtio/network/if_vtnet.c
kfv: You're right, it's indeed a copy of the existing pattern. I did so to keep this revision a pure feature change and idiomatically aligned with the rest of the code. Honestly, I would have liked to do a refactor, but wasn't sure it'd be appropriate or welcome since this is my first time touching our virtio implementation. But I'd be happy to take care of that, so I'll put up a preparatory refactor that all the control commands can sit on, and also address a few other improvements I have in mind.

kfv:
> looks good other than the duplicate code. Do you want to tackle de-duplicating it now or in a subsequent pull request?

I'd say let's keep this PR as is, and I'll shortly open a separate PR focused solely on refactoring. This only is a huge file, and there are many spots in our implementation that I think could be improved (including the control commands we just [discussed](https://github.com/freebsd/freebsd-src/pull/2322#discussion_r3580209330)). That would need me to thoroughly and carefully test everything afterwards, so it might take a bit more time. But I'm OK either way, so I'll follow your call.

kfv:
Whoops! Sorry for accidentally closing the PR, wrong click it was! :-))

spmzt:
Hi @erikarn

Is this ready to commit or you prefer doing the refactor here?

IMHO, it should be on another PR.

erikarn:
We can refactor it in a follow-up PR. Land it when you're ready!

-a

spmzt:
@erikarn Ok then, I'll commit it if you approve this PR.
Thank you!

dch:
@spmzt with @erikarn's above OK, I think you can commit this now.

spmzt:

PR #2323, opened 2026-07-15
sound: uaudio: don't let an idle stream reprogram a shared UAC2 clock

delleceste:
Bugzilla: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=295933 (PR 295933)

## Problem

Some UAC2 devices expose a single Clock Source entity shared between their
playback and capture interfaces (it appears in both the output and input clock
bitmaps). On such a device `uaudio(4)` programs the sample rate for both
directions when a stream starts. If playback runs at a 44.1 kHz-family rate
while the idle capture channel is left at its 48 kHz-family default, the capture
`SET_CUR(CUR_SAM_FREQ)` is issued *after* the playback one and overwrites it on
the shared clock. The device's converter then runs at ~48 kHz while the host
streams 44.1 kHz data, causing a continuous input-FIFO underrun inside the
device: the DAC repeatedly drops and re-acquires USB streaming lock (audible
dropouts, front-panel play/idle flicker). The 48 kHz family is unaffected
because both directions then agree on the rate.

The host side is verifiably healthy throughout (`underruns 0`, channel stays
`RUNNING`); the fault is the wrong clock rate programmed into the device. The
same device plays the 44.1 kHz family correctly under Linux's `snd-usb-audio`,
which does not let an idle capture stream reprogram the clock used by active
playback.

## Fix

- **Shared-clock guard.** Before issuing `SET_CUR` to a clock id, if that clock
  is shared between playback and capture and the other direction is already
  streaming at a different rate, skip it. The first active stream owns the
  clock; a later one follows it.
- **Rate-align the jitter record stream.** When the record channel is
  auto-started only as a source of jitter information for asynchronous
  playback, align its nominal rate to the playback rate before starting it, so
  it neither reprograms the shared clock to a conflicting rate nor produces
  mismatched frame sizes.
- **Always submit the explicit-feedback SYNC transfer** so
  `dev.pcm.%d.feedback_rate` stays live as a diagnostic even when a capture
  stream is present.

## Testing

Reproduced and fixed on an OKTO RESEARCH DAC8 STEREO (`0x152a:0x88c5`), whose
vestigial capture interface never streams. With the patch, 44.1 kHz-family
rates lock cleanly and stay stable, the flicker is gone, and the capture
interface is left in place. Builds `-Werror`-clean; in daily use for over a
week on FreeBSD 15.1, amd64.

### DIFF
--- a/sys/dev/sound/usb/uaudio.c
+++ b/sys/dev/sound/usb/uaudio.c
@@ -1358,6 +1358,46 @@ uaudio_max_buffer_size(struct uaudio_chan *ch, uint8_t alt)
 	return (buf_size);
 }
 
+static bool
+uaudio20_clock_is_shared(struct uaudio_softc *sc, unsigned int x)
+{
+	/*
+	 * A clock entity that feeds both an OUTPUT (playback) and an
+	 * INPUT (capture) terminal shows up in both bitmaps: it is a
+	 * sample clock shared between the two directions.
+	 */
+	return ((sc->sc_mixer_clocks.bit_output[x / 8] & (1 << (x % 8))) != 0 &&
+	    (sc->sc_mixer_clocks.bit_input[x / 8] & (1 << (x % 8))) != 0);
+}
+
+static uint32_t
+uaudio_dir_running_rate(struct uaudio_chan *chans)
+{
+	unsigned int i;
+
+	for (i = 0; i != UAUDIO_MAX_CHILD; i++) {
+		struct uaudio_chan *ch = &chans[i];
+
+		if (ch->running != 0 && ch->cur_alt < ch->num_alt)
+			return (ch->usb_alt[ch->cur_alt].sample_rate);
+	}
+	return (0);		/* nothing streaming in this direction */
+}
+
+static bool
+uaudio_chan_match_rate(struct uaudio_chan *ch, uint32_t rate, uint8_t *p_alt)
+{
+	uint8_t x;
+
+	for (x = 0; x != ch->num_alt; x++) {
+		if (ch->usb_alt[x].sample_rate == rate) {
+			*p_alt = x;
+			return (true);
+		}
+	}
+	return (false);
+}
+
 static void
 uaudio_configure_msg_sub(struct uaudio_softc *sc,
     struct uaudio_chan *chan, int dir)
@@ -1451,6 +1491,35 @@ uaudio_configure_msg_sub(struct uaudio_softc *sc,
 			} else {
 				if (!(sc->sc_mixer_clocks.bit_input[x / 8] &
 				    (1 << (x % 8)))) {
+					continue;
+				}
+			}
+
+			/*
+			 * Shared-clock guard.  If this clock entity is
+			 * shared between the playback and capture paths,
+			 * and the OTHER direction is already streaming at
+			 * a different rate, do not reprogram the clock --
+			 * the rate that is already locked wins.  This
+			 * stops an idle or secondary stream from yanking
+			 * the shared clock out from under an active
+			 * stream and dropping USB stream lock.  The first
+			 * active stream owns the clock; a later one
+			 * follows it (see uaudio_chan_start(), which
+			 * re-aligns the jitter-info record stream to the
+			 * playback rate before it is started).
+			 */
+			if (uaudio20_clock_is_shared(sc, x)) {
+				uint32_t other = (dir == PCMDIR_PLAY) ?
+				    uaudio_dir_running_rate(sc->sc_rec_chan) :
+				    uaudio_dir_running_rate(sc->sc_play_chan);
+
+				if (other != 0 &&
+				    other != chan_alt->sample_rate) {
+					DPRINTF("shared clock ID=%u busy at "
+					    "%u Hz; not reprogramming to "
+					    "%u Hz\n", x, other,
+					    chan_alt->sample_rate);
 					continue;
 				}
 			}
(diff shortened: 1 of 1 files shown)


sys/dev/sound/usb/uaudio.c
```diff
@@ -2334,13 +2403,14 @@ uaudio_chan_play_sync_callback(struct usb_xfer *xfer, usb_error_t error)
...
-		 * Check if the recording stream can be used as a
-		 * source of jitter information to save some
-		 * isochronous bandwidth:
+		 * Always run the SYNC transfer.  When a recording
+		 * stream exists it is used as the source of jitter
+		 * information instead of the feedback value (see
+		 * USB_ST_TRANSFERRED above), but the explicit feedback
+		 * value is still tracked in ch->feedback_rate as a
+		 * diagnostic (dev.pcm.%d.feedback_rate).  This costs
+		 * one isochronous IN frame per second.
```
christosmarg: This comment talks about SYNC tranfers and `feedback_rate` which is set in the case above, but this is the SETUP case. Am I missing something?

christosmarg (commented):
In the commit message:

> the capture SET_CUR(CUR_SAM_FREQ) is issued

`CUR_SAM_FREQ` does not exist, did you mean `UA20_CS_SAM_FREQ_CONTROL`?

> The device's converter then runs at ~48 kHz while the host streams 44.1 kHz data

Is the a reason you are saying "converted" and "host" instead of just "playback" and "recording" streams?

> causing a continuous input-FIFO underrun inside the device: the DAC repeatedly drops and re-acquires USB streaming lock

I think I understand what you are trying to say, but the wording is a bit confusing (especially "input-FIFO"). Please simplify or clarify this.

> When the record channel

Typo, it should be "recording".

sys/dev/sound/usb/uaudio.c
delleceste: The SETUP case is where the transfer gets (re)submitted — the old code broke out here without submitting whenever a record channel existed, so USB_ST_TRANSFERRED never ran and feedback_rate
 stayed 0. Removing the skip is what keeps the feedback value updated; the comment referenced TRANSFERRED because that's where the value is computed. Reworded it to describe just what this case
does. Note this hunk is diagnostic-only (keeps dev.pcm.%d.feedback_rate meaningful); happy to drop it if the extra isochronous IN frame isn't wanted.

delleceste:
> In the commit message:
> 
> > the capture SET_CUR(CUR_SAM_FREQ) is issued
> 
> `CUR_SAM_FREQ` does not exist, did you mean `UA20_CS_SAM_FREQ_CONTROL`?
> 
> > The device's converter then runs at ~48 kHz while the host streams 44.1 kHz data
> 
> Is the a reason you are saying "converted" and "host" instead of just "playback" and "recording" streams?
> 
> > causing a continuous input-FIFO underrun inside the device: the DAC repeatedly drops and re-acquires USB streaming lock
> 
> I think I understand what you are trying to say, but the wording is a bit confusing (especially "input-FIFO"). Please simplify or clarify this.
> 
> > When the record channel
> 
> Typo, it should be "recording".

Thank you for the careful review — going through both the code and the commit message in this much detail is genuinely appreciated, especially for a first contribution to this part of the tree.
  
   All points are addressed in the force-pushed revision:
  
 - commit message now says SET_CUR(UA20_CS_SAM_FREQ_CONTROL) — you're right, the name I used doesn't exist in the tree;
  - reworded the "converter"/"host" sentence. I had avoided "playback vs. recording streams" because the mismatch is between the device's clock and the playback data rate (the recording stream is
  idle at that point); the new wording hopefully says that plainly;
 - simplified the underrun sentence;
 - "record channel" → "recording channel";
 - shortened the USB_ST_SETUP comment as suggested;
 - added the Assisted-By: trailer (also on #2324).
  
Thanks again for your time!

sys/dev/sound/usb/uaudio.c
```diff
@@ -2334,13 +2403,10 @@ uaudio_chan_play_sync_callback(struct usb_xfer *xfer, usb_error_t error)
 
 	case USB_ST_SETUP:
 		/*
-		 * Check if the recording stream can be used as a
-		 * source of jitter information to save some
-		 * isochronous bandwidth:
+		 * Submit the transfer even when the recording stream
+		 * provides the jitter information, so that the feedback
+		 * rate keeps being sampled for diagnostic purposes.
```
christosmarg, line 2408: What is the benefit of making this unconditional, with regards to the bug we are trying to fix though?

christosmarg:
I've been testing with my USB sound card, and this actually fixes the sample rate bug mentioned in the BUGS section of sound(4)'s man page!

christosmarg:
Committed: https://cgit.freebsd.org/src/commit/?id=755685dd665ef209912c59da6a7d0e7f2c9f464b

I also added some additional text in the commit message. Thank you!

delleceste:
> I've been testing with my USB sound card, and this actually fixes the sample rate bug mentioned in the BUGS section of sound(4)'s man page!

Yes, that's true, I've just read the note in sound (4)!
Happy to hear that we've got the fix!

sys/dev/sound/usb/uaudio.c
delleceste: Sorry, I missed this before you committed — answering for the record.

You're right that it's not needed to fix the flicker: with a capture channel present the feedback value is still only stored in ch->feedback_rate and never reaches ch->jitter_curr, so it's 
diagnostic-only.

 The motivation is that feedback_rate is the only thing FreeBSD exposes that reports the device's clock rather than the rate the host requested — and before this change it was unavailable on 
 precisely the devices where those two can disagree, i.e. anything with a capture interface, which is the situation this PR is about. I use it in userland to close the loop after a rate change: 
 request a rate, then confirm the DAC actually locked to it before starting playback. That check is what made this bug visible in the first place; without it a shared-clock mismatch is silent 
apart from the audible symptom.

On cost: since USB_ST_TRANSFERRED breaks rather than falling through to tr_setup:, the transfer is one-shot, and uaudio_chan_play_callback() only kicks it once per second. So this is one 
isochronous IN packet of 3–4 bytes per second per playing channel, on an endpoint whose bandwidth is reserved by the alt setting whether or not we submit. The alternative was 
 hw.usb.uaudio.debug, which costs two log lines a second from the same callback.

 If you'd still rather not pay it unconditionally, gating it on a hw.usb.uaudio tunable instead of uaudio_debug would work fine for me — happy to send that as a follow-up.

PR #2325, opened 2026-07-16
ntpdate temporal consistency check

larkmjc:
discussions about time system attacks causing time corruption resulted in the idea of a temporal consistency check that treats
large time steps by ntpdate as a potential vulnerability. the check uses time powered down to calculate a maximum clock drift.

this series adds an `-x` flag to `last` that reports time in seconds.
this series adds an `-m` flag to `ntpdate` for a maximum absolute offset.
this series adds a few rc.conf parameter defaults to rc.d/ntpdate:

- ntpdate_reboot_error_check=NO
  - temporal consistency check disabled
- ntpdate_reboot_error_dev=0.0001
  - 0.01% deviation or ~8.64 seconds per day
- ntpdate_reboot_offset_max=600
  - 10 mins maximum absolute time offset
- ntpdate_reboot_offset_min=0.01
  - 10 millisecond minimum absolute time offset

the -m maxoffset value is caculated as follows:

    min(max(t_error * abs(t_boot - t_shutdown), t_min), t_max)

shut down time unavailable, e.g. first boot, sets date as usual.
shut down for 60 seconds, maximum allowed offset is ~ +/-0.01 sec.
shut down for one week, maximum allowed offset is ~ +/-1 min.

example normal operation:

```
$ sudo /etc/rc.d/ntpdate start
Setting date via ntp.
maximum absolute offset: 0.0175 secs
16 Jul 17:30:14 ntpdate[3357]: step time server 2001:4428:0:dc::20 offset -0.005959 sec
```

example error operation:

```
$ sudo /etc/rc.d/ntpdate start
Setting date via ntp.
maximum absolute offset: 0.0175 secs
16 Jul 17:33:27 ntpdate[3446]: absolute offset +0.017500 exceeded: offset +1.054452 sec
```

this patch is a proof of concept for review purposes and is not expected to be merged.

### DIFF
--- a/contrib/ntp/ntpdate/ntpdate.c
+++ b/contrib/ntp/ntpdate/ntpdate.c
@@ -161,6 +161,7 @@ int sys_numservers = 0; 	/* number of servers to poll */
 int sys_authenticate = 0;	/* true when authenticating */
 u_int32 sys_authkey = 0;	/* set to authentication key in use */
 u_long sys_authdelay = 0;	/* authentication delay */
+l_fp sys_maxoffset;      	/* maximum offset (zero is disabled) */
 int sys_version = NTP_VERSION;	/* version to poll with */
 
 /*
@@ -350,7 +351,7 @@ ntpdatemain (
 	/*
 	 * Decode argument list
 	 */
-	while ((c = ntp_getopt(argc, argv, "46a:bBde:k:o:p:qst:uv")) != EOF)
+	while ((c = ntp_getopt(argc, argv, "46a:bBde:k:o:p:qst:m:uv")) != EOF)
 		switch (c)
 		{
 		case '4':
@@ -419,6 +420,16 @@ ntpdatemain (
 				sys_timeout = ((LFPTOFP(&tmp) * TIMER_HZ)
 					   + 0x8000) >> 16;
 				sys_timeout = max(sys_timeout, MINTIMEOUT);
+			}
+			break;
+		case 'm':
+			if (!atolfp(ntp_optarg, &tmp)) {
+				(void) fprintf(stderr,
+					   "%s: maxoffset %s is undecodeable\n",
+					   progname, ntp_optarg);
+				errflg++;
+			} else {
+				sys_maxoffset = tmp;
 			}
 			break;
 		case 'v':
@@ -1262,6 +1273,7 @@ clock_adjust(void)
 {
 	register struct server *sp, *server;
 	int dostep;
+	l_fp offset;
 
 	for (sp = sys_servers; sp != NULL; sp = sp->next_server)
 		clock_filter(sp);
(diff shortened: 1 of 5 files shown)


larkmjc:
re: Persistent sampling of RTC drift in ntpd (this PR #2325 as example of that)

this PR is a good example of the underlying gap or "blind spot" during system downtime, though the fix I'm suggesting belongs in ntpd, not this PR.

ntpd's only persistent state is the drift file: one ppm value? fit while running and disciplined against trusted network peers. It has no memory of the "internally controlled" regime — RTC free-running while ntpd/the system is off. this PR works around this per-boot for ntpdate specifically, using last -x timestamps and a fixed maximum bound on error rate — but it only ever considers the last blind interval, and the result goes to syslog rather than anywhere designed for accumulation.

suggestion: ntpd itself could keep a small fixed-capacity ring alongside the drift file — N records of t_shutdown, t_boot, offset_at_loss, offset_at_regain, implied_ppm — so RTC-while-off drift gets characterized empirically per machine over many cycles, rather than assumed as a constant or discarded after one use. This is distinct from the existing drift file, which only ever samples the disciplined regime.

as an aside: authenticating peers against timing attacks, in potentially hostile networks, is its own issue — NTS/autokey and NTPSec territory — not addressed by this ring buffer idea, which only covers the unauthenticated RTC-free-running interval. if they are doing periodic re-keying, then I guess one could also gather an RRD on the pubkey reputation. this is more just thinking aloud as this sounds like a lot of work with many different implementation strategies depending on the chosen crypto ecosystem: X.509/TLS, gpg, and ssh.

jrtc27:
ntpdate is obsolete: https://github.com/freebsd/freebsd-src/blob/2b1df6149e8a2d50a09d13c64d1574dad91e10b1/usr.sbin/ntp/doc/ntpdate.8#L19-L35

larkmjc:
according to the freebsd manpage comment it is. it doesn't mean one doesn't fix security vulnerabilities. so that comment is from 2001 and may have been imported from ntp HTML docs, according to the commit message, after looking for it with git annotate, but it seems, on inspection, it is not present in other ntpdate manpages such as linux and netbsd. openbsd has rdate.

jrtc27:
Then go file it against either upstream or a specific distro that doesn't regard it as deprecated. I refer you also to https://github.com/freebsd/freebsd-src/commit/c153a35bfd716fd74e4b3955cc60541b78ab3d8e, if you want some more recent proof.

And yes, if there is a security vulnerability or serious bug we will surely fix it, but not add new features. Are you reporting a security vulnerability or serious bug?

larkmjc:
unbounded time steps are a security vulnerability, and similar checks have apparently been added to ntpd related to the same vulnerability. it is a security feature that mitigates a security vulnerability. you could make a commit to delete ntpdate, I guess as another way of fixing it. I haven't checked the mitigation in ntpd but it was mentioned to me that it has bounded offset, although I doubt it uses an error term on the blind time as the bound. there is a principle behind that.

jrtc27:
If this is of concern to you then don't use the deprecated ntpdate.

bsdimp:
Ntp doesn't do authentication so unbounded time is the least of your worries

bsdimp:
Ntp doesn't do authentication so bounding the result is kinda worthless

larkmjc:
some ntpd implementations implement authentication and limit absolute time steps I am told but I have not authenticated the statement by auditing the code. ntpd is much harder to audit btw. trying to find the start-up step before it switches to drift adjustment. there is something to be said about simplicity.

here is the ntpdate simulation mode:

https://github.com/freebsd/freebsd-src/blob/2b1df6149e8a2d50a09d13c64d1574dad91e10b1/contrib/ntp/ntpd/ntp_loopfilter.c#L502-L527

this seems to be leap-second related:

https://github.com/freebsd/freebsd-src/blob/2b1df6149e8a2d50a09d13c64d1574dad91e10b1/contrib/ntp/ntpd/ntp_timer.c#L651-L691

I haven't found the start up step code. it may not call `step_systime`.

jrtc27:
Please go elsewhere to brainstorm, and if you have bug reports or feature requests against either ntpd or ntpdate, take them to the upstream sources. We aren't going to carry patches for them unless necessary, i.e. there's a critical vulnerability that upstream isn't going to fix, or some compatibility issue that blocks building FreeBSD.

PR #2326, opened 2026-07-16
geom: zoned: Introduce Zoned Storage emulator

voidanix:
geom: zoned: Introduce Zoned Storage emulator

gzoned is a new GEOM class that exposes host-managed zoned devices
(similar to ZAC/ZBC drives) on top of regular, non-zoned providers.

The new class has several features to realistically emulate zoned
drives: per-zone write pointers, BIO_ZONE zone-management commands
support, fault emulation through zone conditions (RWP recommended,
offline, R/O), URSWRZ bit toggling and concurrent open zone limits.

Created zoned devices are split into zones of equal size, by default
sequential-write-required. Such zones can be turned into conventional
zones if desired. The zoned drive's state is persistent (through
metadata at the tail of the provider).

ATF tests are also included.

Discussed with: asomers, fuz, ken
Sponsored by:   Google Summer Of Code 2026


tests/sys/geom/class/zoned/zoned_test.sh
```diff
@@ -0,0 +1,469 @@
...
+
+	alloc_zoned_md
+	atf_check gzoned create -s 256m ${md}
+	atf_check -o not-match:"Conventional" \
+	    zonectl -d /dev/${md}.zoned -c rz
+}
+sequential_cleanup()
+{
+	gzoned_test_cleanup
+}
```
asomers: IMHO this test is unnecessary.  The defaults might change, and nobody should be relying on them anyway.

lib/geom/zoned/geom_zoned.c
```diff
@@ -0,0 +1,285 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2026 voidanix <voidanix@FreeBSD.org>
+ * All rights reserved.
```
clausecker: You can omit the license as well as the "All rights reserved" text, see style(9).

tests/sys/geom/class/zoned/zoned_test.sh
```diff
@@ -0,0 +1,469 @@
...
+	atf_set "require.user" "root"
+}
+conventional_body()
+{
+	gzoned_test_setup
+
+	alloc_zoned_md
+	atf_check gzoned create -s 256m -c 1,3 ${md}
+	atf_check_equal "2" \
+	    "$(zonectl -d /dev/${md}.zoned -c rz | grep -c Conventional)"
```
asomers: A little bit easier:
```suggestion
	    "$(zone_count nonwp)"
```

tests/sys/geom/class/zoned/zoned_test.sh
```diff
@@ -0,0 +1,469 @@
...
+	atf_set "require.user" "root"
+}
+conventional_range_body()
+{
+	gzoned_test_setup
+
+	alloc_zoned_md
+	atf_check gzoned create -s 256m -c 0-2 ${md}
+	atf_check_equal "3" \
+	    "$(zonectl -d /dev/${md}.zoned -c rz | grep -c Conventional)"
```
asomers:
```suggestion
	    "$(zone_count nonwp)"
```

tests/sys/geom/class/zoned/zoned_test.sh
```diff
@@ -0,0 +1,469 @@
...
+}
+all_zones_same_cleanup()
+{
+	gzoned_test_cleanup
+}
+
+atf_test_case report_filter cleanup
+report_filter_head()
+{
+	atf_set "descr" "zonectl -o only reports zones matching the filter"
```
asomers: This test doesn't exercise gzoned.  Rather, it seems designed to exercise zonectl.  So it really belongs in usr.sbin/zonectl/tests, not here.  Furthermore, this creates a testing loop: gzoned is validated by zonectl, but zonectl is validated by gzoned.

asomers (changes requested):
I've only reviewed the tests so far, but I have a few comments.  Can you please add tests for:
* the maximum number of open zones
* a write is allowed to cross the boundary between two conventional zones
* "gzoned stop -f"
* "gzoned create -s" can adjust zone size
* "gzoned create -r" with a comma-separated list of more than one range
* "gzoned create" on top of an already-zoned device: nothing bad should happen
* "gzoned -c rwp" Does the Right Thing for conventional or empty zones

lib/geom/zoned/geom_zoned.c
```diff
@@ -0,0 +1,285 @@
...
+	if (zoned_parse_conv(req, conv, nzones, &md) != 0)
+		return;
+	maxopen = gctl_get_intmax(req, "maxopen");
+	if (maxopen < 0 || maxopen > nzones) {
+		gctl_error(req, "Maximum open zones must be between 0 and "
+		    "the number of zones (%u).", nzones);
+		return;
+	}
+
+	strlcpy(md.md_magic, G_ZONED_MAGIC, sizeof(md.md_magic));
```
clausecker: Better use `memcpy` here as `md` is already pre-zeroed.

sys/geom/zoned/g_zoned.c
```diff
@@ -0,0 +1,1574 @@
...
+
+	mtx_assert(&sc->sc_lock, MA_OWNED);
+
+	if (g_zoned_cond_is_open(z->zone_condition) !=
+	    g_zoned_cond_is_open(cond)) {
+		if (g_zoned_cond_is_open(cond))
+			sc->sc_nopen++;
+		else
+			sc->sc_nopen--;
+	}
```
clausecker: Maybe clearer?

```suggestion
	bool is_open, want_open;

	is_open = g_zoned_cond_is_open(z->zone_condition);
	want_open = g_zoned_cond_is_open(cond);
	if (!is_open && want_open)
			sc->sc_nopen++;
	else if (is_open && !want_open)
			sc->sc_nopen--;
```

Or shorter:

```suggestion
        sc->sc_nopen += (uint32_t)g_zoned_cond_is_open(cond)
            - (uint32_t)g_zoned_cond_is_open(z->zone_condition);
```

sys/geom/zoned/g_zoned.c
```diff
@@ -0,0 +1,1574 @@
...
+	g_topology_unlock();
+
+	valid = false;
+	buf = g_read_data(cp, sc->sc_taboff, sc->sc_secsize, &error);
+	if (buf != NULL) {
+		valid = strncmp((char *)buf, G_ZONED_TABLE_MAGIC,
+		    sizeof(G_ZONED_TABLE_MAGIC) - 1) == 0 &&
+		    le32dec(buf + 20) == sc->sc_nzones;
+		g_free(buf);
```
clausecker, line 256: The zone table should have a structure, maybe with a flexible array member at the end.  Magic byte offsets are very annoying to read and make the code unmaintainable.

sys/geom/zoned/g_zoned.c
```diff
@@ -0,0 +1,1574 @@
...
+		chunk = MIN(total - done, (off_t)maxphys);
+		chunk -= chunk % sc->sc_secsize;
+		error = g_write_data(cp, off + done, buf + done, chunk);
+	}
+	if (error == 0)
+		error = g_io_flush(cp);
+	if (error != 0)
+		G_ZONED_DEBUG(0,
+		    "Final zone table write on %s failed (error=%d).",
+		    cp->provider->name, error);
```
clausecker: This seems cleaner with a break to underline that this is a bog standard counting loop with an exceptional exit path.

```suggestion
	for (done = 0; done < total; done += chunk) {
		chunk = MIN(total - done, (off_t)maxphys);
		chunk -= chunk % sc->sc_secsize;
		error = g_write_data(cp, off + done, buf + done, chunk);
		if (error != 0)
			break;	
	}

	if (error == 0)
		error = g_io_flush(cp);
	if (error != 0)
		G_ZONED_DEBUG(0,
		    "Final zone table write on %s failed (error=%d).",
		    cp->provider->name, error);
```

sys/geom/zoned/g_zoned.c
```diff
@@ -0,0 +1,1574 @@
...
+		    DISK_ZONE_CLOSE_SUP | DISK_ZONE_FINISH_SUP |
+		    DISK_ZONE_RWP_SUP | DISK_ZONE_MAX_SEQ_SET;
+		if (!sc->sc_rdrestrict)
+			p->flags |= DISK_ZONE_DISK_URSWRZ;
+		/* Optimal zone counts are host-aware concepts; leave unset. */
+		p->optimal_seq_zones = 0;
+		p->optimal_nonseq_zones = 0;
+		/* 0xffffffff is the ZBC "unlimited" sentinel. */
+		p->max_seq_zones = (sc->sc_maxopen != 0) ? sc->sc_maxopen :
+		    0xffffffff;
```
clausecker: Can we get a symbolic constant for the sentinel?  Elsewhere you use `UINT64_MAX` for it, which is not `0xffffffff`.

sys/geom/zoned/g_zoned.h
```diff
@@ -0,0 +1,237 @@
...
+ * Version history:
+ * 1 - Initial version number.
+ */
+#define	G_ZONED_VERSION		1
+
+/*
+ * On-disk layout, carved out of the end of the backing provider:
+ *
+ *   [ ... usable (zoned) region ... | table header | zone entries | metadata ]
+ *                                     (1 sector)     (N entries)   (last sec)
```
clausecker, line 34: Please add a struct for the table header and use it instead of manually computing offsets.

sys/geom/zoned/g_zoned.h
```diff
@@ -0,0 +1,237 @@
...
+ * Table header + zone entries hold the live per-zone state and are maintained
+ * entirely by the kernel.
+ */
+/* On-disk zone-state table entry */
+struct g_zoned_disk_entry {
+	uint8_t		de_type;		/* Zone type. */
+	uint8_t		de_condition;		/* Zone condition. */
+	uint8_t		de_flags;		/* Zone attribute flags. */
+	uint8_t		de_reserved[5];
+	uint64_t	de_write_pointer_lba;	/* Write pointer LBA. */
```
clausecker: If I understood the spec correctly, the write pointer is a 32 bit number within the zone (meaning each zone can have up to 2³² - 1 blocks).  Maybe you can do it the same way for consistency?  Then the structure would fit into 8 bytes in total, saving disk space.

sys/geom/zoned/g_zoned.h
```diff
@@ -0,0 +1,237 @@
...
+	uint32_t	md_sectorsize;	/* Provider sector size in bytes. */
+	uint64_t	md_provsize;	/* Provider size in bytes. */
+	uint32_t	md_nconv;	/* Conventional ranges in use. */
+	struct g_zoned_convrange md_conv[G_ZONED_MAXCONV];
+	uint32_t	md_flags;	/* G_ZONED_MD_* flags. */
+	uint32_t	md_maxopen;	/* Open zone limit, 0 = unlimited. */
+};
+
+static __inline void
+zoned_metadata_encode(const struct g_zoned_metadata *md, u_char *data)
```
clausecker, line 117: Can we have similar functions for the zone table header?

asomers (changes requested):
Testing has revealed a bug: gzoned reports conventional zones as having a write pointer equal to the start LBA, not -1:

What zonectl  show for a gzoned device:
```
> sudo zonectl -c rz  -d /dev/md0.zoned
3 zones, Maximum LBA 0x17ffff (1572863)
Zone lengths are the same, types vary
  Start LBA  Length       WP LBA      Zone Type      Condition      Sequential             Reset
          0, 524288,           0,  Conventional,           NWP,     Sequential,  No Reset Needed
    0x80000, 524288,     0x80000,  Conventional,           NWP,     Sequential,  No Reset Needed
   0x100000, 524288,    0x100000,  Seq Required,           NWP,     Sequential,  No Reset Needed
```
What it shows for a real ZBC disk:
```
 sudo zonectl -c rz  -d /dev/da3 | head
55880 zones, Maximum LBA 0xda47ffff (3662151679)
Zone lengths are the same, types vary
  Start LBA  Length       WP LBA      Zone Type      Condition      Sequential             Reset
          0,  65536,          -1,  Conventional,           NWP,     Sequential,  No Reset Needed
    0x10000,  65536,          -1,  Conventional,           NWP,     Sequential,  No Reset Needed
    0x20000,  65536,          -1,  Conventional,           NWP,     Sequential,  No Reset Needed
    0x30000,  65536,          -1,  Conventional,           NWP,     Sequential,  No Reset Needed
```

sys/geom/zoned/g_zoned.c
voidanix: The table header and entries now have their own structs, so this part is indeed cleaner.

> Magic byte offsets are very annoying to read and make the code unmaintainable.

If you are referring to the metadata: what would be ideal here? Would it be acceptable to derive the codec offsets from the structs via `__offsetof()`, then pin them with asserts? Something like:

```c
le32enc(data + __offsetof(struct g_zoned_metadata, md_version), md->md_version);
```

It would also mean realigning the metadata layout to place `md_conv` at the end as you suggested. Currently it starts at byte 52 and the 64-bit fields cannot line up without `__packed`...

> maybe with a flexible array member at the end

The table is not contiguous on disk. The secsize of the provider is a runtime property and the gap between the header (at sector 0) and entries (starting at sector 1) depends on this: using a flexible array member would lead to sector 0 storing the header *and* a number of entries that depends on the sector size.

The current split makes every entry sector hold the same number of entries and makes header updates dirty a different sector than entry updates. The dirty-sector flush relies on this.

clausecker: There are basically two options:

 - define a structure with the same layout as the on-disk format and use it for marshalling, perhaps using packed structures. Slightly fragile, but workable in practice
 - define enumeration constants with the byte offsets and marshall through these

In neither case would you need `offsetof` as you can just use a cast like this:

    struct g_zoned_metadata *md_ondisk = (struct g_zoned_metadata *)data;

    le32enc(md_ondisk->md_version, md->md_version)

or use the enumeration constants from the other approach.

clausecker:
Looking pretty good here.

sys/geom/zoned/g_zoned.c
```diff
@@ -0,0 +1,1556 @@
...
+		 * Sequential-write-required zones accept writes at the
+		 * WP only.
+		 */
+		if (z->zone_condition == DISK_ZONE_COND_FULL ||
+		    lba != z->write_pointer_lba) {
+			G_ZONED_LOGREQ(bp,
+			    "Out-of-order write to zone %u (lba %ju, wp %ju).",
+			    zno, (uintmax_t)lba,
+			    (uintmax_t)z->write_pointer_lba);
+			return (EIO);
```
asomers: It would be very awesome if this and other error cases used exterror(9), a new facility in FreeBSD 15.  Something like this:

```suggestion
			return (EXTERROR(EIO, "Write pointer violation.  lba=%ju wp=%ju", lba, z->write_pointer_lba));
```

sys/geom/zoned/g_zoned.h
```diff
@@ -0,0 +1,268 @@
...
+	return ((uint32_t)(((uint64_t)mediasize - reserve) /
+	    (uint64_t)zonesize));
+}
+
+#ifdef _KERNEL
+#define	G_ZONED_DEBUG(lvl, ...) \
+    _GEOM_DEBUG("GEOM_ZONED", g_zoned_debug, (lvl), NULL, __VA_ARGS__)
+#define	G_ZONED_LOGREQLVL(lvl, bp, ...) \
+    _GEOM_DEBUG("GEOM_ZONED", g_zoned_debug, (lvl), (bp), __VA_ARGS__)
+#define	G_ZONED_LOGREQ(bp, ...)	G_ZONED_LOGREQLVL(2, bp, __VA_ARGS__)
```
asomers, line 199: This macro is handy.  But it would be handier if you logged errors at debug level 1, and all accesses at debug level 2.

sys/geom/zoned/g_zoned.c
```diff
@@ -0,0 +1,1556 @@
...
+		mtx_lock(&sc->sc_lock);
+		if (sc->sc_rdrestrict) {
+			error = g_zoned_read_check(sc, bp);
+			if (error != 0) {
+				mtx_unlock(&sc->sc_lock);
+				g_io_deliver(bp, error);
+				return;
+			}
+		}
+		sc->sc_reads++;
```
asomers, line 1200: Why increment sc_reads here, but increment sc_writes within g_zoned_write_check?  I don't like the inconsistency.

sys/geom/zoned/g_zoned.c
```diff
@@ -0,0 +1,1556 @@
...
+
+	cbp = g_clone_bio(bp);
+	if (cbp == NULL) {
+		g_io_deliver(bp, ENOMEM);
+		return;
+	}
+	cbp->bio_done = (bp->bio_cmd == BIO_WRITE) ? g_zoned_write_done :
+						     g_std_done;
+	G_ZONED_LOGREQ(cbp, "Sending request.");
+	g_io_request(cbp, LIST_FIRST(&gp->consumer));
```
asomers: What happens if cbp fails a write request for a reason like EIO or ENXIO?  You've already advanced the write pointer.  If the underlying request fails, then the write pointer will be inconsistent with what's actually on disk.  I think you should only advance the write pointer after the underlying request succeeds.

sys/geom/zoned/g_zoned.c
```diff
@@ -0,0 +1,1556 @@
...
+
+	zno = g_zoned_zoneno(sc, lba);
+	z = &sc->sc_zones[zno];
+
+	if (z->zone_type == DISK_ZONE_TYPE_CONVENTIONAL) {
+		last = (end > lba) ? g_zoned_zoneno(sc, end - 1) : zno;
+		for (i = zno + 1; i <= last; i++) {
+			if (sc->sc_zones[i].zone_type !=
+			    DISK_ZONE_TYPE_CONVENTIONAL) {
+				G_ZONED_LOGREQ(bp, "Read crosses from a "
```
asomers: Does ZBC/ZAC really restrict reads from crossing zone boundaries _only_ if URSWRZ is set?  Even if the read starts in a conventional zone and extends into a sequential zone but remains beneath the sequential zone's write pointer?

sys/geom/zoned/g_zoned.c
```diff
@@ -0,0 +1,1556 @@
...
+		zno = g_zoned_zoneno(sc, lba);
+		z = &sc->sc_zones[zno];
+		if (z->zone_type != DISK_ZONE_TYPE_CONVENTIONAL &&
+		    z->write_pointer_lba == end) {
+			z->write_pointer_lba = lba;
+			g_zoned_set_cond(sc, z,
+			    (lba == z->zone_start_lba) ?
+				DISK_ZONE_COND_EMPTY :
+				DISK_ZONE_COND_IMPLICIT_OPEN);
+			g_zoned_mark_dirty(sc, zno);
```
asomers: This will roll back an optimistically full zone to IMPLICIT_OPEN if the write fails.  But what if the zone was previously explicitly open or closed?  I don't think we want it to transition from explicitly open to implicitly open.

sys/geom/zoned/g_zoned.c
```diff
@@ -0,0 +1,1556 @@
...
+ * drive would.
+ *
+ * Persistence:
+ *   - The provider's last sector holds a metadata block, written by the
+ *     userland "create" command. The kernel tastes it on every provider
+ *     arrival and re-creates the zoned device automatically.
+ *   - The sectors just before it hold the per-zone state (condition + write
+ *     pointer). That table is read at taste time and rewritten lazily.
+ *     Zone-state changes are committed to disk on BIO_FLUSH (and the table is
+ *     also flushed when zone-management commands run), mirroring a drive whose
```
asomers: It doesn't actually flush the zone table during zone commands.  It just marks the table as dirty.
```suggestion
 *     Zone-state changes are committed to disk on BIO_FLUSH mirroring a drive whose
```

sys/geom/zoned/g_zoned.c
```diff
@@ -0,0 +1,1556 @@
...
+
+static void
+g_zoned_flush_write_done(struct bio *bp)
+{
+	struct g_zoned_flush *fc = bp->bio_caller1;
+	int error = bp->bio_error;
+
+	g_destroy_bio(bp);
+	if (error != 0) {
+		G_ZONED_DEBUG(0, "Zone table write failed (error=%d).", error);
```
asomers, line 365: If flush failed, then we didn't really flush the data.  But we've already cleared the dirty flag.  Can we restore the dirty flag here?

asomers (commented):
Overall, this looks quite good.  I haven't read every line, but I've read most of them, at least cursorily.  There might be some lingering off-by-one arithmetic errors or something, but the good test coverage makes me feel confident enough even so.  I just have a few concerns before I'm ready to approve this.

Could you please add a test for resize?

Could you also add a check in userland during "gzoned create" that the conventional zone ranges don't overlap?

sys/geom/zoned/g_zoned.c
voidanix: Should I drop the G_ZONED_LOGREQs then?  No messages will show without `options EXTERR_STRINGS`.

asomers: That's not exactly the same thing.  _GEOM_DEBUG is for printing debug messages to the console whenever you want them.  But exterror(9) is about returning detailed error messages to userland instead of simple errnos.  Both are good to have.

voidanix: You are fully correct... sleepless brain thought they were somehow redundant.

sys/geom/zoned/g_zoned.c
voidanix: Good catch... ZBC-3 4.5.2.3 indeed says that it's a READ BOUNDARY VIOLATION. Addressed in cfad08a4467099892c45762a39e7e8d48aff9a0b.

voidanix:
Rebased and addressed all feedback, thank you!

91e57de6dc2680a66c9b and d8647fa5daacd114cd2e were also added while I was at it, as a follow-up to #2345 . I am happy to split those commits into a separate PR if they are deemed superfluous.

sys/geom/zoned/g_zoned.h
```diff
@@ -0,0 +1,286 @@
...
+#define	G_ZONED_MD_FLAGSMASK		G_ZONED_MD_RESTRICTED_READS
+
+struct g_zoned_metadata {
+	char		md_magic[16];	/* Magic value. */
+	uint32_t	md_version;	/* Version number. */
+	uint32_t	md_id;		/* Unique ID. */
+	uint64_t	md_zonesize;	/* Zone size in bytes. */
+	uint32_t	md_nzones;	/* Number of zones. */
+	uint32_t	md_sectorsize;	/* Provider sector size in bytes. */
+	uint64_t	md_provsize;	/* Provider size in bytes. */
```
asomers: These four parameters over-specify the device.  They're redundant, since `md_nzones` can be uniquely calculated from the other three.  Including redundant information in a config format is a good way to get very confused when things are corrupted.  I suggest removing the `md_nzones` field from the metadata, and recomputing it during taste.

sys/geom/zoned/g_zoned.c
```diff
@@ -0,0 +1,1647 @@
...
+			    (uint64_t)i));
+		}
+		if (i > zno && sc->sc_zones[i].zone_type != z->zone_type) {
+			G_ZONED_LOGREQLVL(1, bp, "Read crosses into a zone of "
+			    "a different type.");
+			return (G_ZONED_EXTERR(bp, EIO,
+			    "Read crosses into a zone of a different type."
+			    "  lba=%ju zone=%ju", (uint64_t)lba,
+			    (uint64_t)i));
+		}
```
asomers, line 1095: Does ZBC really prohibit all reads crossing into zones of a different type, regardless of `URSWRZ`?  The HGST HSH721415AL42M4 disk that Robert sent me allows it (though perhaps it's not fully ZBC compliant).

asomers (commented):
This PR is looking very good.  I like it.  I especially like the exterror feature.  It works even better than I knew.  Check out this zonectl output:

```
> sudo zonectl -d /dev/md1.zoned -c rwp -l 1
zonectl: DIOCZONECMD ioctl failed: Invalid argument (Zone has no write pointer to manage.  zone=0 cond=0)
```

sys/geom/zoned/g_zoned.c
voidanix: Yes, that is the case: see ZBC-3 4.5.3.1.6. I've also mentioned 4.5.2.3 before, which is related. I think your drive is ZBC-1 which may behave differently compared to ZBC-3.

asomers (approved):
LGTM!

clausecker:
@asomers Do you want to land this one?

asomers:
> @asomers Do you want to land this one?

Sure.  But first, @voidanix can you please squash your commits and finalize the commit message?

voidanix:
> @voidanix can you please squash your commits and finalize the commit message?

Squashed everything except the last commit to ease reviewing, as there was a serious issue while testing this on a real 8TiB drive with `-s 512`, yielding almost 17 billion zones (completely unrealistic for real drives).

The uint32_t nzones counter would wrap and continuously try to allocate 240G of memory for the zone table. I hard capped the nzones limit to UINT32_MAX and made the metadata writing possible only after the zone table malloc succeeds (see `create` gctl verb).

Let me know if everything can be squashed together after this, or whether I should leave it as-is.

EDIT: here's the new output while testing with a 8 TiB vnode (md1):

```
## Too many zones
$ gzoned create -s 512 -c 0-100 md1
gzoned: Zone size 512 is unusable for md1: it must yield at least one zone and at most 4294967295.

## The host has only 6G of RAM, fails cleanly
$ gzoned create -s 2k -c 0-100 md1
gzoned: Cannot create a zoned device on md1.
GEOM_ZONED: Cannot allocate a 4278190079 zone table for md1.
```

sys/geom/zoned/g_zoned.h
```diff
@@ -0,0 +1,291 @@
...
+	md->md_flags = le32toh(d.md_flags);
+	md->md_maxopen = le32toh(d.md_maxopen);
+}
+
+/*
+ * Number of zones a provider of a given geometry can hold after reserving room
+ * at the tail for the metadata sector and the zone-state table. Returns 0 if
+ * the zone size does not leave room for even a single zone, or for more than
+ * G_ZONED_MAXZONES allows.
+ */
```
asomers: This comment is no longer accurate, because there are now two reasons for returning zero.

sys/geom/zoned/g_zoned.c
```diff
@@ -1352,9 +1359,6 @@ g_zoned_create(struct g_class *mp, const struct g_zoned_metadata *md,
 		g_detach(cp);
 	g_destroy_consumer(cp);
 	g_destroy_provider(newpp);
-	mtx_destroy(&sc->sc_lock);
-	free(sc->sc_zones, M_ZONED);
-	g_free(sc);
```
asomers: Why did you delete these lines?  Aren't they still needed if `g_attach` or `g_access` fails?

asomers (commented):
Good catch.  And good that you tested those limits, however unrealistic they may be.  I only have one question, inline, about the new commit.

sys/geom/zoned/g_zoned.c
voidanix: No, and they would result in a double free. The path already calls `g_destroy_provider()` at the line above which, in turn, calls `g_zoned_providergone()` that already does the cleanup.

`mtx_destroy()` here gets you a `panic: assertion mtx_unowned(m) failed at /usr/src/sys/kern/kern_mutex.c:1203`, which can be reached with a read-only provider as `g_access(cp, 1, 1, 1)` fails, as you said.

asomers (approved):

PR #2328, opened 2026-07-17
man: Fix RFC 1918 network prefix lengths

private-yusuke:
According to [RFC 1918](https://datatracker.ietf.org/doc/html/rfc1918), the following IP prefixes are reserved for private internets:

- 10.0.0.0/8
- 172.16.0.0/12
- 192.168.0.0/16

This PR fixes the prefix lengths in references to private networks ("RFC 1918 networks", "the standard private IP address ranges").
The changes are limited to man pages.


cschuber:
Thank you for the patch. It's committed now.

PR #2330, opened 2026-07-18
gpart: Allow partitioning zoned devices (GPT only)

voidanix:
Partition metadata cannot be written inside sequential-write-required
zones on host-managed drives, so committing a partition table would fail
halfway through with I/O errors.

To overcome this, let gpart know up-front whether the metadata ranges
can be written, as long as they fit within conventional zones. Those
ranges are defined for GPT only, as usecases for other schemes are
limited on zoned storage.

As host-managed drives usually offer sequential zones at the end of the
drive, only the backup GPT at the end of the provider remains unwritable
and is thus omitted.

Reported by:    fuz, asomers
Discussed with: fuz, asomers, ken
Sponsored by:   Google Summer of Code 2026

Depends on: #2326


voidanix:
As I am fairly unoriginal with tests, nothing of relevance came up other than the run-of-the-mill ones e.g. "check if not corrupt when retasting" or "check if other partitioning schemes fail"... Getting `gzoned(8)` into shape first certainly make testing feasible, although I would definitely appreciate some feedback on what could be added on the ATF side of things.

lib/geom/part/gpart.8
```diff
@@ -613,6 +613,20 @@ in the EBR scheme.
 It also prevents any type of actions on such partitions.
 .El
 .Pp
+On host-managed zoned providers
```
asomers, line 616: Don't forget to bump the .Dd date in the man page, too.

sys/geom/part/g_part.c
```diff
@@ -874,6 +875,146 @@ g_part_ctl_bootcode(struct gctl_req *req, struct g_part_parms *gpp)
...
+		zar.zone_params.report.entries_allocated = 1;
+		zar.zone_params.report.entries = &entry;
+		error = g_io_zonecmd(&zar, cp);
+		if (error != 0)
+			return (error);
+		/* Guard against a report that is empty or fails to advance. */
+		if (zar.zone_params.report.entries_filled != 1 ||
+		    entry.zone_start_lba + entry.zone_length <= lba)
+			return (ENXIO);
+		if (entry.zone_type != DISK_ZONE_TYPE_CONVENTIONAL) {
```
asomers: What about SEQ_PREFERRED zones?  I think gpart should treat those like conventional zones.  I don't think that zone type should ever show up here, given that this function is only called for host-managed disks.  But still, program defensively.

sys/geom/part/g_part.c
```diff
@@ -874,6 +875,146 @@ g_part_ctl_bootcode(struct gctl_req *req, struct g_part_parms *gpp)
...
+	 * scrub maps matter for it.
+	 */
+	for (idx = 0; table->gpt_scheme != &g_part_null_scheme; idx++) {
+		error = G_PART_GETMDRANGE(table, pp, idx, &start, &length,
+		    &flags);
+		if (error == ENOENT)
+			break;
+		if (error != 0) {
+			gctl_error(req, "%d %s: partition scheme %s is not "
+			    "supported on host-managed zoned providers",
```
asomers: This error message is misleading; the current code won't support host-aware either.

asomers (changes requested):
Looks good so far.  How about some tests?

sys/geom/part/g_part.c
voidanix: Done in 9de2ff34c16f760a710cb1e5f5dea9e55acefa29

sys/geom/part/g_part.c
voidanix: Reworded it better in d9879e27ea75acaad8b4d485fc5ea284a0e5d961 , hopefully this is enough.

voidanix:
Rebased on top of gzoned branch (mainly for tests). Will switch this to a WIP PR considering the circumstances.

PR #2332, opened 2026-07-18
fts: add Capsicum capability mode tests

Jeetubhati:
Add `fts_capsicum_test.c` with three test cases verifying fts(3)
behavior with `fts_dirfd` and in Capsicum capability mode.

**Test cases:**

- `fts_dirfd_valid` — verifies `fts_dirfd` is set to a valid fd for all non-root entries and that `openat(ent->fts_dirfd,
  ent->fts_name)` identifies the same inode as `fts_accpath`

- `fts_dirfd_capsicum` — verifies complete fts traversal works in Capsicum capability mode using `fts_openat()` with a
  pre-opened `dirfd` and `FTS_NOCHDIR`. Accesses files via `openat(ent->fts_dirfd, ent->fts_name)` without any
  path-based syscalls after `cap_enter(2)`

- `fts_dirfd_deep_tree` — verifies `fts_dirfd` + `fts_name` correctly identifies files at all directory depths

**Dependencies:**

- PR #2303 — `fts_dirfd` in `struct _ftsent`, fd-relative `fts_build` using `openat` + `fdopendir`
- PR #2273 — `fts_openat()` API

`fts_dirfd_valid` and `fts_dirfd_deep_tree` pass independently.
`fts_dirfd_capsicum` requires both PRs to be installed.

### DIFF
--- a/lib/libc/tests/gen/fts_capsicum_test.c
+++ b/lib/libc/tests/gen/fts_capsicum_test.c
@@ -49,16 +49,9 @@ ATF_TC_BODY(fts_dirfd_valid, tc)
 		if (ent->fts_level == FTS_ROOTLEVEL)
 			continue;
 
-		ATF_CHECK_MSG(ent->fts_dirfd >= 0,
+		ATF_REQUIRE_MSG(ent->fts_dirfd >= 0,
 		    "fts_dirfd must be valid for '%s' at level %ld info=%d",
 		    ent->fts_name, ent->fts_level, ent->fts_info);
-
-		if (ent->fts_dirfd < 0)
-			continue;
-
-		/* Skip post-order — same entry as pre-order. */
-		if (ent->fts_info == FTS_DP)
-			continue;
 
 		ATF_REQUIRE_EQ_MSG(0,
 		    fstatat(ent->fts_dirfd, ent->fts_name, &sb_dirfd,
@@ -118,6 +111,8 @@ ATF_TC_BODY(fts_dirfd_capsicum, tc)
 	 */
 	ATF_REQUIRE((fts = fts_openat(dirfd, paths,
 	    FTS_PHYSICAL | FTS_NOCHDIR, NULL)) != NULL);
+	close(dirfd);
+
 
 	/* Enter capability mode — no more path-based syscalls. */
 	ATF_REQUIRE_EQ(0, cap_enter());
--- a/lib/libc/tests/gen/fts_capsicum_test.c
+++ b/lib/libc/tests/gen/fts_capsicum_test.c
@@ -1,0 +1,222 @@
+/*
+ * Copyright (c) 2026 Jitendra Bhati
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Tests for fts(3) in Capsicum capability mode using fts_dirfd.
+ */
+
+#include <sys/capsicum.h>
+#include <sys/stat.h>
+
+#include <errno.h>
+#include <fcntl.h>
+#include <fts.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#include <atf-c.h>
+
+/*
+ * Verify that fts_dirfd is set to a valid file descriptor for every
+ * entry returned by fts_read(), and that openat(fts_dirfd, fts_name)
+ * correctly identifies the same file as fts_accpath.
+ */
+ATF_TC(fts_dirfd_valid);
+ATF_TC_HEAD(fts_dirfd_valid, tc)
+{
+	atf_tc_set_md_var(tc, "descr",
+	    "fts_dirfd is valid for all non-root entries");
+}
+ATF_TC_BODY(fts_dirfd_valid, tc)
+{
+	char *paths[] = { "dir", NULL };
+	FTS *fts;
+	FTSENT *ent;
+	struct stat sb_accpath, sb_dirfd;
+
+	ATF_REQUIRE_EQ(0, mkdir("dir", 0755));
+	ATF_REQUIRE_EQ(0, mkdir("dir/sub", 0755));
+	ATF_REQUIRE_EQ(0, close(creat("dir/sub/file", 0644)));
+	ATF_REQUIRE_EQ(0, close(creat("dir/other", 0644)));
+
+	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL, NULL)) != NULL);
+
+	while ((ent = fts_read(fts)) != NULL) {
+		/* Root entry has no parent directory fd. */
+		if (ent->fts_level == FTS_ROOTLEVEL)
+			continue;
+
+		ATF_CHECK_MSG(ent->fts_dirfd >= 0,
+		    "fts_dirfd must be valid for '%s' at level %ld info=%d",
+		    ent->fts_name, ent->fts_level, ent->fts_info);
+
+		if (ent->fts_dirfd < 0)
+			continue;
+
+		/* Skip post-order — same entry as pre-order. */
+		if (ent->fts_info == FTS_DP)
+			continue;
+
+		ATF_REQUIRE_EQ_MSG(0,
+		    fstatat(ent->fts_dirfd, ent->fts_name, &sb_dirfd,
+		    AT_SYMLINK_NOFOLLOW),
+		    "fstatat(fts_dirfd, fts_name) failed for '%s': %m",
+		    ent->fts_name);
+		ATF_REQUIRE_EQ_MSG(0,
+		    lstat(ent->fts_accpath, &sb_accpath),
+		    "lstat(fts_accpath) failed for '%s': %m",
+		    ent->fts_accpath);
+
+		ATF_CHECK_EQ_MSG(sb_accpath.st_ino, sb_dirfd.st_ino,
+		    "inode mismatch for '%s': "
+		    "fts_accpath ino=%ju fts_dirfd ino=%ju",
+		    ent->fts_name,
+		    (uintmax_t)sb_accpath.st_ino,
+		    (uintmax_t)sb_dirfd.st_ino);
+		ATF_CHECK_EQ_MSG(sb_accpath.st_dev, sb_dirfd.st_dev,
+		    "device mismatch for '%s'", ent->fts_name);
+	}
+
+	ATF_REQUIRE_EQ_MSG(0, fts_close(fts), "fts_close(): %m");
+}
+
+/*
+ * Verify that fts traversal works correctly in Capsicum capability
+ * mode using openat(fts_dirfd, fts_name) to access files.
+ */
+ATF_TC(fts_dirfd_capsicum);
+ATF_TC_HEAD(fts_dirfd_capsicum, tc)
+{
+	atf_tc_set_md_var(tc, "descr",
+	    "fts traversal using fts_dirfd works in Capsicum capability mode");
+}
+ATF_TC_BODY(fts_dirfd_capsicum, tc)
+{
+	if (!feature_present("security_capabilities") ||
+	    !feature_present("security_capability_mode"))
+		atf_tc_skip("Capsicum not available");
+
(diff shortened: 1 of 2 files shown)


lib/libc/tests/gen/fts_capsicum_test.c
```diff
@@ -0,0 +1,222 @@
...
+		/* Root entry has no parent directory fd. */
+		if (ent->fts_level == FTS_ROOTLEVEL)
+			continue;
+
+		ATF_CHECK_MSG(ent->fts_dirfd >= 0,
+		    "fts_dirfd must be valid for '%s' at level %ld info=%d",
+		    ent->fts_name, ent->fts_level, ent->fts_info);
+
+		if (ent->fts_dirfd < 0)
+			continue;
```
asomers: This check is redundant with the above one.  Remove it.

lib/libc/tests/gen/fts_capsicum_test.c
```diff
@@ -0,0 +1,222 @@
...
+
+		/* Skip post-order — same entry as pre-order. */
+		if (ent->fts_info == FTS_DP)
+			continue;
```
asomers: Why skip?  I think it would be worthwhile to perform the checks in both post and pre order.

lib/libc/tests/gen/fts_capsicum_test.c
```diff
@@ -0,0 +1,222 @@
...
+
+	ATF_REQUIRE((dirfd = open("dir", O_RDONLY | O_DIRECTORY)) >= 0);
+
+	/*
+	 * Open fts before entering capability mode — fts_openat
+	 * will dup dirfd internally so we can close our copy.
+	 */
+	ATF_REQUIRE((fts = fts_openat(dirfd, paths,
+	    FTS_PHYSICAL | FTS_NOCHDIR, NULL)) != NULL);
+
```
asomers: How about closing dirfd here?  As is, the test leaks that file descriptor.

asomers (changes requested):

lib/libc/tests/gen/fts_capsicum_test.c
Jeetubhati: Removed.

lib/libc/tests/gen/fts_capsicum_test.c
Jeetubhati: Earlier I skipped post-order to avoid duplicate `stat `checks since
`FTS_DP `revisits the same directory. Fixed now — `fts_dirfd `is verified valid in post-order too, but the `fstatat `check is still skipped since `fts_name `in post-order refers to the directory itself which was already checked in pre-order.

lib/libc/tests/gen/fts_capsicum_test.c
Jeetubhati: I added `close(dirfd)` after `fts_openat()` — `fts_openat` dups it internally so closing the caller's copy is safe.

lib/libc/tests/gen/fts_capsicum_test.c
asomers: Why skip the fstatat checks?  They won't noticeably slow the test down.

Jeetubhati: You're right, no good reason to skip. I'll remove the post-order early continue and run `fstatat `in both orders.

lib/libc/tests/gen/fts_capsicum_test.c
```diff
@@ -0,0 +1,225 @@
...
+		if (ent->fts_level == FTS_ROOTLEVEL)
+			continue;
+
+		ATF_REQUIRE_MSG(ent->fts_dirfd >= 0,
+		    "fts_dirfd must be valid for '%s' at level %ld info=%d",
+		    ent->fts_name, ent->fts_level, ent->fts_info);
+
+		/* In post-order, fts_dirfd should still be valid. */
+		if (ent->fts_info == FTS_DP) {
+			ATF_CHECK_MSG(ent->fts_dirfd >= 0,
```
asomers: This check is redundant.  It will never fail, because the check 6 lines above will always fail first.  You should delete this check.

asomers (changes requested):

lib/libc/tests/gen/fts_capsicum_test.c
Jeetubhati: Already removed in the latest commit — the entire `FTS_DP `block in `fts_dirfd_valid `is gone.

asomers (approved):

lib/libc/tests/gen/fts_capsicum_test.c
```diff
@@ -0,0 +1,217 @@
...
+	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL, NULL)) != NULL);
+
+	while ((ent = fts_read(fts)) != NULL) {
+		/* Root entry has no parent directory fd. */
+		if (ent->fts_level == FTS_ROOTLEVEL)
+			continue;
+
+		ATF_REQUIRE_MSG(ent->fts_dirfd >= 0,
+		    "fts_dirfd must be valid for '%s' at level %ld info=%d",
+		    ent->fts_name, ent->fts_level, ent->fts_info);
```
asomers: This is no longer true, right?  Shouldn't fts_dirfd only be valid for `FTS_D` and `FTS_DP`?

lib/libc/tests/gen/fts_capsicum_test.c
```diff
@@ -0,0 +1,217 @@
...
+
+		ATF_REQUIRE_EQ_MSG(0,
+		    fstatat(ent->fts_dirfd, ent->fts_name, &sb_dirfd,
```
asomers:
```suggestion
		    fstatat(ent->fts_parent->fts_dirfd, ent->fts_name, &sb_dirfd,
```

lib/libc/tests/gen/fts_capsicum_test.c
```diff
@@ -0,0 +1,217 @@
...
+		    ent->fts_info == FTS_D)
+			saw_sub = true;
+
+		if (strcmp(ent->fts_name, "file") == 0 &&
+		    ent->fts_info == FTS_F) {
+			saw_file = true;
+
+			struct stat sb;
+			ATF_CHECK_EQ_MSG(0,
+			    fstatat(ent->fts_dirfd, ent->fts_name, &sb,
```
asomers:
```suggestion
			    fstatat(ent->fts_parent->fts_dirfd, ent->fts_name, &sb,
```

lib/libc/tests/gen/fts_capsicum_test.c
```diff
@@ -0,0 +1,217 @@
...
+	ATF_REQUIRE_EQ(0, close(creat("dir/root", 0644)));
+
+	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL, NULL)) != NULL);
+
+	while ((ent = fts_read(fts)) != NULL) {
+		if (ent->fts_info == FTS_DP ||
+		    ent->fts_level == FTS_ROOTLEVEL)
+			continue;
+
+		ATF_REQUIRE_MSG(ent->fts_dirfd >= 0,
```
asomers: Only true for directories

lib/libc/tests/gen/fts_capsicum_test.c
```diff
@@ -0,0 +1,217 @@
...
+		    "fts_dirfd must be valid for '%s' at level %ld",
+		    ent->fts_name, ent->fts_level);
+
+		ATF_REQUIRE_EQ_MSG(0,
+		    fstatat(ent->fts_dirfd, ent->fts_name, &sb_dirfd,
```
asomers:
```suggestion
		    fstatat(ent->fts_parent->fts_dirfd, ent->fts_name, &sb_dirfd,
```

asomers (changes requested):

Jeetubhati:
Rebased to contain only the capsicum test files. Updated to use `fts_parent->fts_dirfd + fts_name` per new` fts_dirfd` semantics.

lib/libc/tests/gen/fts_capsicum_test.c
```diff
@@ -0,0 +1,217 @@
...
+	ATF_REQUIRE_EQ(0, mkdir("dir/sub", 0755));
+	ATF_REQUIRE_EQ(0, close(creat("dir/sub/file", 0644)));
+	ATF_REQUIRE_EQ(0, close(creat("dir/other", 0644)));
+
+	ATF_REQUIRE((fts = fts_open(paths, FTS_PHYSICAL, NULL)) != NULL);
+
+	while ((ent = fts_read(fts)) != NULL) {
+		/* Root entry has no parent directory fd. */
+		if (ent->fts_level == FTS_ROOTLEVEL)
+			continue;
```
asomers: As I explained in #2273 , I think we should ensure that the root level has a parent with a valid dirfd too.  It might be AT_FDCWD.

lib/libc/tests/gen/fts_capsicum_test.c
```diff
@@ -0,0 +1,217 @@
...
+	 * Open fts before entering capability mode — fts_openat
+	 * will dup dirfd internally so we can close our copy.
+	 */
+	ATF_REQUIRE((fts = fts_openat(dirfd, paths,
+	    FTS_PHYSICAL | FTS_NOCHDIR, NULL)) != NULL);
+	close(dirfd);
+
+
+	/* Enter capability mode — no more path-based syscalls. */
+	ATF_REQUIRE_EQ(0, cap_enter());
```
asomers: If we're using `fts_openat` instead of `fts_open`, then we should enter capability mode first.

lib/libc/tests/gen/fts_capsicum_test.c
```diff
@@ -0,0 +1,217 @@
...
+		    ent->fts_info == FTS_F) {
+			saw_file = true;
+
+			struct stat sb;
+			ATF_CHECK_EQ_MSG(0,
+			    fstatat(ent->fts_parent->fts_dirfd, ent->fts_name, &sb,
+			    AT_SYMLINK_NOFOLLOW),
+			    "fstatat(fts_dirfd, fts_name) failed for "
+			    "'%s' in capability mode: %m",
+			    ent->fts_name);
```
asomers: No reason to restrict this check for regular files.  It should work for directories too.

asomers (changes requested):
It seems to me that these tests are overlapping with those in fts_openat_test.c from the other PR.  We might want to consider merging the two files.

Jeetubhati:
Fixed review comments:
- `fts_dirfd_valid`: check `fts_parent->fts_dirfd `for all entries
- `fts_dirfd_capsicum`: cap_enter before `fts_openat`, check all entry types not just `FTS_F`
All 3 capsicum tests pass.

lib/libc/tests/gen/fts_capsicum_test.c
```diff
@@ -0,0 +1,214 @@
...
+	ATF_REQUIRE_EQ(0, mkdir("dir", 0755));
+	ATF_REQUIRE_EQ(0, mkdir("dir/sub", 0755));
+	ATF_REQUIRE_EQ(0, close(creat("dir/sub/file", 0644)));
+	ATF_REQUIRE_EQ(0, close(creat("dir/other", 0644)));
+
+	ATF_REQUIRE((dirfd = open("dir", O_RDONLY | O_DIRECTORY)) >= 0);
+
+	/*
+	 * Open fts before entering capability mode — fts_openat
+	 * will dup dirfd internally so we can close our copy.
```
asomers: This comment is out-of-date now.

asomers (commented):
This LGTM to me now, with the exception of that one inaccurate comment.

lib/libc/tests/gen/fts_capsicum_test.c
Jeetubhati: Fixed the comment.

asomers (approved):

asomers (approved):

PR #2334, opened 2026-07-21
unix: allow `connectat(2)` to name the peer socket by descriptor

Ericson2314:
The second commit is the heart of the change (and from whence I am getting the PR title).

The first commit is a preparatory refactor without any behavior change, and the third and final commit is a test.

The idea is that, if we have the listening socket already open, there is no reason to have to go consult the file system again to find it, we can just connect to it directly.

If this change is accepted, the natural follow-up would be also allow listening on unbound sockets, and then the file system can be avoided entirely!

The existing capsicum system of fine-grained permissions on file descriptors works very well for this, I am happy to report. :)

*Draft because I am currently figuring out how to test this. I have run my tests on an stock FreeBSD machine and seen them fail as expected, but I am still getting cross-compiling the kernel working*

### DIFF
--- a/sys/kern/uipc_usrreq.c
+++ b/sys/kern/uipc_usrreq.c
@@ -294,6 +294,8 @@ static int	unp_connect(struct socket *, struct sockaddr *,
 		    struct thread *);
 static int	unp_connectat(int, struct socket *, struct sockaddr *,
 		    struct thread *, bool);
+static int	unp_connect_peer(struct socket *, struct unpcb *,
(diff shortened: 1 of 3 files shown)


Ericson2314:
I was able to built the kernel and run the tests, and everything passes! Can confirm the tests fail with the upstream kernel and pass with my changes --- just as intended.

Ericson2314:
I really want stacked PRs for this, so I am going to switch to phabricator. https://reviews.freebsd.org/D58406

PR #2335, opened 2026-07-23
linprocfs fix in /proc/pid/maps for mmap04 ltp case

ilovegrape:
1. The /proc/pid/maps file on Linux does not zero-pad addresses to a fixed width
1. Distinguish private vs shared anonymous mappings in maps for ltp mmap04 test case


ilovegrape:
Hi @dag-erling could you please help to take a look on this change?

PR #2336, opened 2026-07-23
e1000: Fix link state stuck DOWN at 10/100 after TSO automask

kfv:
This patch fixes a link-state regression from 2ddf24f8f525.

On [lem(4)](https://man.freebsd.org/cgi/man.cgi?lem(4))/[em(4)](https://man.freebsd.org/cgi/man.cgi?em(4)) sub-gigabit media, `em_automask_tso()` disables TSO and forces a reset, so `LINK_STATE_UP` is deferred; but `sc->link_active` is left set, and hence the post-reset pass never re-enters the up-transition branch and the UP is never delivered.

`if_link_state` then stays DOWN while the link is actually up, breaking consumers that key off link state.
Clearing `sc->link_active` when the UP is deferred is desired, so the post-reset pass redelivers it.

Cc: @spmzt

### DIFF
--- a/sys/dev/e1000/if_em.c
+++ b/sys/dev/e1000/if_em.c
@@ -2301,10 +2301,16 @@ em_if_update_admin_status(if_ctx_t ctx)
 		if (hw->mac.type < igb_mac_min)
 			automasked = em_automask_tso(ctx);
 
-		/* Automasking resets the interface so don't mark it up yet */
+		/*
+		 * Automasking resets the interface, so don't mark it up;
+		 * clear link_active instead so the post-reset pass redelivers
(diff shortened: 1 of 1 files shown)


gmshake:
Also CC @kev009

kev009:
That fixes the immediate issue.  There are a couple related inconsistencies from the complexity, can you test this patch which turns it into a more obvious state machine: https://reviews.freebsd.org/D58425

kfv:
Hey, sure! I'll do it in a few hours.

kfv:
Closing in favour of [D58425](https://reviews.freebsd.org/D58425).

PR #2338, opened 2026-07-24
libsysdecode: use local sysdecode.h

ishan0803:
Replace <sysdecode.h> with "sysdecode.h" so local builds use the in-tree header in lib/libsysdecode instead of a stale installed copy in /usr/include, avoiding build failures after updating sysdecode.h.

Sponsored-by : Google LLC (GSoC 2026)


bsdimp:
`make buildenv BUILDENV_SHELL=make` is a very useful command to pickup the in-tree version rather than the installed version of things.

PR #2345, opened 2026-07-31
zonectl: Consistently report ZAC conv. zones WP LBA

voidanix:
zonectl: Consistently report ZAC conv. zones WP LBA

On ZAC drives, conventional zones conventionally report a write pointer
LBA of 0xffffffffffff. This field is 48 bits wide, unlike ZBC's 64 bits.

Recognize both ZAC and ZBC all-ones behaviour in the WRITE POINTER LBA
field to indicate non-valid information.

Tested by:      fuz
Discussed with: fuz, asomers, ken
Fixes:          4735ef6196bc ("zonectl: display conventional zones better during RZ")
Sponsored by:   Google Summer Of Code 2026

(sorry for manually having to fill in PR bodies/titles... gh CLI is not collaborating...)

### DIFF
--- a/usr.sbin/zonectl/zonectl.c
+++ b/usr.sbin/zonectl/zonectl.c
@@ -292,7 +292,8 @@ zonectl_print_rz(struct disk_zone_report *report, zone_output_flags out_flags,
 		    (uintmax_t)entry->zone_start_lba,
 		    field_widths[ZONE_FW_LEN],
 		    (uintmax_t)entry->zone_length);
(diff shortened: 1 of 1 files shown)


asomers (approved):
I see a potential for confusion with ZBC drives only once they exceed 1 EiB in size.  But at that point, ZAC will have to expand its LBA size anyway.  And since this output is only intended for human consumption, I think the risk is acceptable.

PR #2346, opened 2026-07-31
libpmc: Fix AMD L3 counters

mashtizadeh:
Fix two small bugs affecting the event parsing of AMD L3 counters. AMD's manual and JSON disagree about the naming scheme on recent processors.  I use the naming scheme present in the recent PPRs to be consistent, so in the json parser we rename allslices to  allsources just as we already do with sliceid and sourceid.  Also ensure that we parse the 0x prefix present in the newer json files.

Sponsored by: Netflix

### DIFF
--- a/lib/libpmc/libpmc_pmu_util.c
+++ b/lib/libpmc/libpmc_pmu_util.c
@@ -371,13 +371,13 @@ pmu_parse_event(struct pmu_event_desc *ped, const char *eventin)
 		else if (strcmp(key, "l3_slice_mask") == 0)
 			ped->ped_l3_slice = strtol(value, NULL, 16);
 		else if (strcmp(key, "sourceid") == 0)
-			ped->ped_sourceid = strtol(value, NULL, 16);
+			ped->ped_sourceid = strtol(value, NULL, 0);
 		else if (strcmp(key, "coreid") == 0)
-			ped->ped_coreid = strtol(value, NULL, 16);
+			ped->ped_coreid = strtol(value, NULL, 0);
 		else if (strcmp(key, "allcores") == 0)
-			ped->ped_allcores = strtol(value, NULL, 10);
+			ped->ped_allcores = strtol(value, NULL, 0);
 		else if (strcmp(key, "allsources") == 0)
-			ped->ped_allsources = strtol(value, NULL, 10);
+			ped->ped_allsources = strtol(value, NULL, 0);
 		else if (strcmp(key, "pebs") == 0)
 			ped->ped_pebs = strtol(value, NULL, 10);
 		else {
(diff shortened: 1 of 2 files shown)


mashtizadeh:
@mhorne @bsdimp

mashtizadeh:
Removed end of line white spaces and improved the commit message.

lib/libpmc/libpmc_pmu_util.c
```diff
@@ -375,9 +375,9 @@ pmu_parse_event(struct pmu_event_desc *ped, const char *eventin)
 		else if (strcmp(key, "coreid") == 0)
 			ped->ped_coreid = strtol(value, NULL, 16);
 		else if (strcmp(key, "allcores") == 0)
-			ped->ped_allcores = strtol(value, NULL, 10);
+			ped->ped_allcores = strtol(value, NULL, 16);
```
mhorne: So, is it always hex in our sources today?

lib/libpmc/pmu-events/jevents.c
```diff
@@ -623,7 +623,14 @@ static int json_events(const char *fn,
 			} else if (json_streq(map, field, "EnAllCores")) {
 				addfield(map, &allcores, "", "allcores=", val);
 			} else if (json_streq(map, field, "EnAllSlices")) {
-				addfield(map, &allslices, "", "allslices=", val);
```
mhorne, line 626: It's not important to keep this alias?

mhorne (approved):

lib/libpmc/pmu-events/jevents.c
mashtizadeh: Yea I think that's the right thing to do.  I was double checking and I forgot AMD uses slices for some Zen 4 PPRs, then switches to sources for everything else.  I don't see any other json using the keyword slices or allslices or enallslices.

lib/libpmc/libpmc_pmu_util.c
mashtizadeh: Yes everything is hex with the 0x prefix, but maybe we should switch to base = 0 in general since we have the 0x prefixes.  That makes sense more general for the parser particularly with another change I need to put up.

mashtizadeh:
I'll refresh this with both changes in a bit.  Thanks Mitchell!

lib/libpmc/libpmc_pmu_util.c
mashtizadeh: I changed it zero.

lib/libpmc/pmu-events/jevents.c
mashtizadeh: Left it as source after checking with AMD.

mashtizadeh:
Ready to land

mhorne (approved):

PR #2348, opened 2026-08-03
nullfs: allow mounting a single unix-domain socket

Defenso-QTH:
Extend nullfs file-mounts to VSOCK so a single socket can be exposed (e.g. into a jail) without sharing the directory that contains it. The motivating case is handing a jail exactly one daemon socket such as seatd's or a jailed bhyve's rfb unix socket instead of a directory in which anything else may later appear.

Note: If the socket is unbound and re-created, the mount needs to be torn down and set up again.

Sponsored by: Defenso

### DIFF
--- a/sbin/mount_nullfs/mount_nullfs.8
+++ b/sbin/mount_nullfs/mount_nullfs.8
@@ -62,12 +62,26 @@ but in other respects it is indistinguishable from the original.
 .Pp
 The
 .Nm
-utility supports mounting both directories and single files.
+utility supports mounting directories, single files, and single
+.Xr unix 4
+domain sockets.
 Both
 .Ar target
 and
 .Ar mount_point
 must be the same type.
+.Pp
+Note that a socket mount refers to one specific socket, not to a name:
+if the serving daemon unlinks and re-binds its socket, as it typically does
+when restarted, the mount continues to refer to the old socket and
+.Xr connect 2
+fails.
+Mounting the containing directory does not have this limitation.
+.Pp
+When mounting a single file or socket, the
+.Ar target
+must have exactly one hard link, and the resulting mount point cannot
+itself be covered by a further mount.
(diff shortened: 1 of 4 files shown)


kostikbel (commented):
When would such thing useful?  You cannot open existing unix socket for listening.
So for this to be used, somebody must create a socket, start listening, and only then the socket can be mount over something. If the listener exits, the socket must be unmounted.

Also, since this patch requires equal types for the mount point and the node to mount over it, you must have a socket created just to mount something on it.

Overall, I want the explanation how is it going to be used.

Defenso-QTH:
Thanks for your attention!

Here is my use case:
- start `seatd` on my host
- create a jail
- nullfs mount `/var/run/seatd.sock` in the jail
- run a wayland compositor in the jail that gets access to display and input devices by enumerating through sysctls (see related https://github.com/wulf7/libudev-devd/pull/12 and https://github.com/wulf7/libudev-devd/pull/18 for instance) and obtaining fds through `seatd`.

`seatd` is expected to keep running continuously as long as the host is running so the socket remains up and available to other consumers.

(You can see a list of other PRs / commits related to this project here: https://wiki.freebsd.org/QuentinThebault)

This change avoids maintaining patches on ports that do not allow configurable socket paths - or exposing the whole `/var/run` directory through a nullfs mount.

kostikbel (approved):
I hesitated to handle this submission for some time.
Technically it is correct.  But the feature it introduced is very awkward to use, that is, the quirk of rebinding is IMO the very serious limitation.

Eventually I decided to not block the change, since the code modifications are small and not questionable.  But the feature is.

Defenso-QTH:
Made obsolete by https://github.com/freebsd/freebsd-src/commit/2c68ad49f13ddfa33735bd9bb6a3ca170a472ac0.

PR #2350, opened 2026-08-04
fsck_msdosfs: fix left shift signed integer overflow in readboot()

sutes-work:
Avoid signed integer promotion shift overflow when decoding 32-bit fields (e.g., FSFree, FSNext, bpbHiddenSecs, etc.) by using le32dec() from <sys/endian.h> instead of shifting raw u_char bytes into signed int.


delphij (approved):

delphij:
Committed, thanks!

PR #2351, opened 2026-08-04
fsck_msdosfs: fix memory leaks in checkfilesys()

sutes-work:
Invoke releasefat(fat) in checkfilesys() prior to free(fat) on exit paths so that fatbuf, headbitmap.map, and fat32_cache entries are properly freed.


delphij (approved):

delphij:
Committed, thanks!

PR #2206, opened 2026-05-18
libc: Add missing C23 feature test macros

kfv:
|   Header File   |        Feature Test Macro        |    Value    |  Conformance  |        Landed        |
|  :------------  |  :-----------------------------  |  :-------:  |  :----------  |  :-----------------  |
|  <assert.h>     |  `__STDC_VERSION_ASSERT_H__`     |  `202311L`  |  Full         |  2126cab             |
|  <complex.h>    |  `__STDC_VERSION_COMPLEX_H__`    |  `202311L`  |  Partial         |                      |
|  <fenv.h>       |  `__STDC_VERSION_FENV_H__`       |  `202311L`  |  Partial      |                      |
|  <float.h>      |  `__STDC_VERSION_FLOAT_H__`      |  `202311L`  |  Partial      |                      |
|  <inttypes.h>   |  `__STDC_VERSION_INTTYPES_H__`   |  `202311L`  |  Full      |  733ea4ee584db9b25b470ee48243135a321b4eed  |
|  <limits.h>     |  `__STDC_VERSION_LIMITS_H__`     |  `202311L`  |  Full      |  1f09e354297ce571353210c23e50ef6fc19ce382             |
|  <math.h>       |  `__STDC_VERSION_MATH_H__`       |  `202311L`  |  Partial      |                      |
|  <setjmp.h>     |  `__STDC_VERSION_SETJMP_H__`     |  `202311L`  |  Full         |  4aeed6e9d2132366fe5d7fa46d47d73be9de263b  |
|  <stdarg.h>     |  `__STDC_VERSION_STDARG_H__`     |  `202311L`  |  Partial      |                      |
|  <stdatomic.h>  |  `__STDC_VERSION_STDATOMIC_H__`  |  `202311L`  |  Full         |  85becda             |
|  <stdbit.h>     |  `__STDC_VERSION_STDBIT_H__`     |  `202311L`  |  Full         |  6296500             |
|  <stdckdint.h>  |  `__STDC_VERSION_STDCKDINT_H__`  |  `202311L`  |  Full         |  e6615b1             |
|  <stddef.h>     |  `__STDC_VERSION_STDDEF_H__`     |  `202311L`  |  Partial      |                      |
|  <stdint.h>     |  `__STDC_VERSION_STDINT_H__`     |  `202311L`  |  Full         |                      |
[...]

PR #2352, opened 2026-08-04
libc: Complete C23 <limits.h> conformance

kfv:
This is complementary to #2206.

### DIFF
--- a/sys/sys/limits.h
+++ b/sys/sys/limits.h
@@ -69,6 +69,7 @@
 #endif
 
 #if __ISO_C_VISIBLE >= 2023
+#define BOOL_MAX	1		/* max value for a bool */
 #define BOOL_WIDTH	1		/* number of bits in a bool */
 
 #define CHAR_WIDTH	__CHAR_BIT	/* number of bits in a char */
@@ -86,6 +87,12 @@
 
 #define LLONG_WIDTH	__LLONG_BIT	/* number of bits in a long long */
 #define ULLONG_WIDTH	__LLONG_BIT	/* number of bits in an unsigned long long */
+
+#ifdef __BITINT_MAXWIDTH__
+#define BITINT_MAXWIDTH	__BITINT_MAXWIDTH__
+#else
+#define BITINT_MAXWIDTH	ULLONG_WIDTH
+#endif
 #endif
 
 #if __POSIX_VISIBLE || __XSI_VISIBLE
--- a/include/limits.h
+++ b/include/limits.h
@@ -29,8 +29,8 @@
  * SUCH DAMAGE.
  */
 
-#ifndef _LIMITS_H_
-#define	_LIMITS_H_
+#ifndef __STDC_VERSION_LIMITS_H__
+#define __STDC_VERSION_LIMITS_H__	202311L
 
 #include <sys/cdefs.h>
 
@@ -148,4 +148,4 @@
 #include <sys/syslimits.h>
 #endif
 
-#endif /* !_LIMITS_H_ */
+#endif /* !__STDC_VERSION_LIMITS_H__ */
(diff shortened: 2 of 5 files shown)


kfv:
Cc: @clausecker

sys/arm/include/_limits.h
```diff
@@ -41,11 +41,14 @@
  * unsigned int for 32-bit two's complement ANSI compilers (section 3.1.3.2).
  */
 
+#define __BOOL_BIT	1		/* number of bits in a bool */
```
clausecker: I had intentionally not added such definitions earlier as `BOOL_BIT` and `BOOL_MAX` are architecture-invariant.  It would be nice to not pollute `sys/${ARCH}/include` with stuff that is not actually architecture dependent as to make it easier to find out what actually does need to be adjusted.

sys/sys/limits.h
```diff
@@ -69,7 +69,9 @@
 #endif
 
 #if __ISO_C_VISIBLE >= 2023
-#define BOOL_WIDTH	1		/* number of bits in a bool */
+#define BOOL_MAX	__BOOL_MAX	/* max value for a bool */
```
clausecker: Instead just define the width and maximum of `bool` right here as constants.  They'll always be the same anyway.

clausecker (changes requested):
This is a good candidate for one of your first commits.

Please familiarise yourself with [Committer's Guide](https://docs.freebsd.org/en/articles/committers-guide/) and go through the process of setting up your commit bit.  @devinteske will do the parts needed to be done on the mentor's side as your primary mentor.

To land this commit, you'll need:

 - successful testing of the commit on your part; I recommend ensuring kernel and world build with your change and to run the test suite once (there should be no new failures).  Ideally also do runtime testing as appropriate (such as, installing world with your changes and testing if some source file using the new symbols build)
 - positive code review (optional, but highly recommended)
 - approval from one of your mentors

For the first few commits, I kindly ask you to provide your proposed commit message as a comment before you get mentor approval to land the commit so we can make sure that it'll follow our conventions.  You can find documentation on that in the committer's guide and examples by reading through the git history of the project.  If you wish to do review through pull requests, this is easily achieved by you rebasing and squashing the commits in your PR as you envision them to land in the tree.

Later on we'll trust you to get your commit messages right on your own.

kfv:
Sure, it sounds great. Thanks!

sys/arm/include/_limits.h
kfv: Makes sense, right.

sys/sys/limits.h
kfv: Correct.

clausecker:
LGTM.  Please squash rebase and prepare commit messages as per committer's guide, so I can do a final check and approve for commit.
I think this could be one or two commits (you could put the unit test stuff into a separate commit, but it's so little that it doesn't seem to make much sense).
I think @devinteske hasn't set up your commit access yet, so we'll have to wait for that in any case.

kfv:
Do the commit messages look good now? I combined the three related commits into one, but kept the unit test and typo fix commits separate. Please let me know if there's anything else I should change.

kfv:
* Changed `Reviewed by` to Approved by` and dropped `Signed-off-by`.

clausecker:
Looking good.  It might be helpful if you went into the “why” in addition to the “what” in your commit message.  For example, you should say that these symbols are added for compliance with C23 and that the feature test macro is defined now that we believe the header to be fully in compliance.

Bug fixes can be MFC'ed after less time than new features, I usually try to do an MFC after about a week for these.  Waiting a full month is fine though.

You can add both `Reviewed by` and `Approved by`.  The former indicates that code review was conducted, while the latter only indicates that approval was granted.

Put in `Approved by: fuz (mentor)` as I approve of your comment in my function as your mentor.  For some parts of the tree, approval is needed to commit for other reasons (e.g. maintainer approval is needed for modifying a port), so this distinction is important.

kfv:
Thanks for the notes. Done.

clausecker:
Looks good to me.

Approved for commit.

PR #2354, opened 2026-08-05
fts: fix fts_accpath regression in fts_read

Jeetubhati:
`fts_safe_changedir()` was incorrectly changed to use `p->fts_dirfd` instead of -1 in` PR #2303`, which broke `fts_accpath` for entries in newly descended directories. `fts_safe_changedir `needs -1 to use` p->fts_accpath` for the `chdir `operation.

Reported by: asomers

### DIFF
--- a/lib/libc/gen/fts.c
+++ b/lib/libc/gen/fts.c
@@ -489,8 +489,7 @@ fts_read(FTS *sp)
 		 * FTS_STOP or the fts_info field of the node.
 		 */
 		if (sp->fts_child != NULL) {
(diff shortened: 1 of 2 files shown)


asomers (changes requested):
Can you also please add a regression test that fails with the unfixed fts_open, and passes with the fixed one?

asomers (changes requested):
The new test case does not work.  It passes regardless of whether libc is fixed.

asomers (approved):

PR #2355, opened 2026-08-05
libzfs_core add get zfs dataset properties support

elliejs:
This PR adds the ability to get dataset properties to libzfs_core since that's the stable API. You can supply NULL to get all properties on a dataset or a partial nvlist to get only those properties (mirrors the behavior with bookmarks). The returned properties nvlist returns the properties selected (or all), as well as potentially the source of the property (eg the dataset the property was inherited from).

setting properties is a separate concern which could be worked on next, but I didn't want to clog up this PR.

Tested by rebuilding the patch into my FreeBSD box `FreeBSD KIDON 15.1-RELEASE-p2 FreeBSD 15.1-RELEASE-p2 libzfs_core-get-zprops-n283597-09d163c8a7a2 GENERIC amd64` and running this [test_get_props.c](https://github.com/user-attachments/files/30753021/test_get_props.c) MRE.

MRE compiled with:
```
cc -o test_get_props test_get_props.c -lzfs_core -lnvpair \
  -I/usr/src/sys/contrib/openzfs/include \
  -I/usr/src/sys/contrib/openzfs/lib/libspl/include \
  -I/usr/src/sys/contrib/openzfs/lib/libspl/include/os/freebsd
  ```


elliejs:
sorry for the mess, it took me a while to move the pr to main (testing took place with this commit based on releng/15.1)

working on bringing this into spec with CONTRIBUTING.md, sorry, first time. will rebase later

elliejs:
all of this is sys/contrib/openzfs, which is an active upstream. closing here, will re-open there.

PR #2356, opened 2026-08-05
man9: document cold, nitems, and powerof2

joaomboni:
**Task from the FreeBSD JuniorJobs list.**

Add missing kernel man pages listed in JuniorJobs / bug 108980:
- cold(9): early-boot global from sys/systm.h
- nitems(9): array length macro from sys/param.h
- powerof2(9): power-of-two test macro from sys/param.h
Verified with mandoc -T lint.


bsdimp:
Can you check we don't document nitems and powerof2 in cdefs man page? I went through all the macros like that, i though, and documented them 12-18 months ago iirc. Maybe we just need MLINKS for them?

joaomboni:
Thanks for the review — I checked.

nitems and powerof2 are not in cdefs(9) (nor defined in sys/cdefs.h).
They come from sys/param.h via sys/_param.h.

cdefs(9) documents related convenience macros (__offsetof, __rangeof,
__containerof), but not these.

So I don't think MLINKS alone would work unless we first add the text
somewhere (cdefs(9) or a param-oriented page). Happy to:
1) keep the dedicated nitems(9)/powerof2(9) pages, or
2) fold them into an existing/new page and use MLINKS if you prefer.

cold(9) is separate (sys/systm.h) and unrelated to cdefs.

share/man/man9/powerof2.9
```diff
@@ -0,0 +1,36 @@
...
+.\" Copyright (c) 2026 João Bonifácio <joaoboni017@gmail.com>
+.\"
+.\" SPDX-License-Identifier: BSD-2-Clause
+.\"
+.Dd August 5, 2026
+.Dt POWEROF2 9
+.Os
+.Sh NAME
+.Nm powerof2
+.Nd test whether a value is a power of two
```
ngie-eign:
```suggestion
.Nd test whether or not a value is a power of two
```

share/man/man9/powerof2.9
```diff
@@ -0,0 +1,36 @@
...
+.In sys/param.h
+.Fn powerof2 "x"
+.Sh DESCRIPTION
+The
+.Fn powerof2
+macro evaluates to a non-zero value if
+.Fa x
+is a power of two, and to zero otherwise.
+It is equivalent to
+.Li (((x) - 1) & (x)) == 0 .
```
ngie-eign, line 24: I wanted to offer some different wording given that the user should be testing for boolean true or false instead of expecting the value returned to be an integral quantity.
```suggestion
The
.Fn powerof2
macro evaluates to boolean true if the quantity represented by
.Fa x
is a power of two. The macro evaluates to boolean false otherwise.
.Fn powerof2
is equivalent to
.Li (((x) - 1) & (x)) == 0 .
```

share/man/man9/nitems.9
```diff
@@ -0,0 +1,38 @@
...
+.Nd number of elements in an array
+.Sh SYNOPSIS
+.In sys/param.h
+.Fn nitems "array"
+.Sh DESCRIPTION
+The
+.Fn nitems
+macro expands to the number of elements in
+.Fa array .
+It is equivalent to
```
ngie-eign:
```suggestion
.Fa nitems
is equivalent to
```

share/man/man9/nitems.9
```diff
@@ -0,0 +1,38 @@
...
+.Pq sizeof(array) / sizeof((array)[0]) .
+.Pp
+.Fa array
+must be a true array object.
+If
+.Fa array
+is a pointer, the result is wrong because
+.Li sizeof
+yields the size of the pointer, not the size of the pointed-to object.
```
ngie-eign: Subtly different wording. Someone else in docs may disagree with this suggestion.
```suggestion
Providing a pointer quantity will result in an incorrect result because
.Li sizeof
for a pointer will compute size of the pointer type, not the size of the memory which the pointer refers to.
```

share/man/man9/cold.9
```diff
@@ -0,0 +1,35 @@
+.\" Copyright (c) 2026 João Bonifácio <joaoboni017@gmail.com>
+.\"
+.\" SPDX-License-Identifier: BSD-2-Clause
+.\"
+.Dd August 4, 2026
+.Dt COLD 9
+.Os
+.Sh NAME
+.Nm cold
+.Nd indicate that the system is still booting
```
ngie-eign: Slight wording tweak.
```suggestion
.Nd early system boot indicator
```

share/man/man9/cold.9
```diff
@@ -0,0 +1,35 @@
...
+.Sh SYNOPSIS
+.In sys/systm.h
+.Vt extern int cold;
+.Sh DESCRIPTION
+The global variable
+.Va cold
+is non-zero while the kernel is still performing early boot
```
ngie-eign:
```suggestion
is set to a non-zero value while the kernel is still performing early boot
```

share/man/man9/cold.9
```diff
@@ -0,0 +1,35 @@
...
+It is initialized to
+.Dv 1
+during machine-dependent startup and is set to
+.Dv 0
+when autoconfiguration has finished.
+.Pp
+Kernel code uses
+.Va cold
+to decide whether the system is still in the early boot phase,
+when sleeping or relying on fully configured devices may not be
```
ngie-eign: Maybe this is better since I originally read this and thought you were missing an Oxford comma between `sleeping` and `or` :/?
```suggestion
to decide whether the system is still in the early boot phase, i.e.,
when sleeping or relying on fully configured devices may not be
```

share/man/man9/powerof2.9
concussious: Nice, just don't forget to break line 20 where the new sentence starts.

joaomboni: thanks for the review, i fixed the line wrapping.

joaomboni:
Thanks for the review. I've incorporated the suggested wording and formatting updates.

share/man/man9/Makefile
```diff
@@ -245,7 +246,7 @@ MAN=	accept_filter.9 \
 	mutex.9 \
 	namei.9 \
 	netisr.9 \
-	nhop.9 \
```
ngie-eign, line 248: Why was this line deleted?

ngie-eign: @joaomboni : ping?

PR #2357, opened 2026-08-06
fts: fix fts_accpath regression after fts_children()

Jeetubhati:
When fts_children() is called, sp->fts_child is set. On the next fts_read() call, fts_safe_changedir() was incorrectly passed p->fts_dirfd (pointing to the parent directory) instead of -1. This caused fts to fchdir to the parent instead of the child directory, silently skipping the contents of 3rd-level subdirectories. This was observed as a failure in nmtree_test:mtree_create which calls fts_children() internally.

Add regression test accpath_correct_after_descent that calls fts_children() on each directory entry and verifies files at depth 3 are still visited correctly. The test fails with the unfixed libc and passes with the fix.

Reported: asomers
Sponsored by:Google LLC (GSoC 2026)

### DIFF
--- a/lib/libc/gen/fts.c
+++ b/lib/libc/gen/fts.c
@@ -489,8 +489,7 @@ fts_read(FTS *sp)
 		 * FTS_STOP or the fts_info field of the node.
 		 */
 		if (sp->fts_child != NULL) {
-			if (fts_safe_changedir(sp, p,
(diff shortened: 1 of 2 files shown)


Jeetubhati:
Superseded by PR https://github.com/freebsd/freebsd-src/pull/2354 on jbhati/fts-accpath-fix which has been updated with the correct fix and regression test.

PR #2358, opened 2026-08-07
bhyve: rtc_pl031: Fix PeriphID and CellID values

mrkajetanp:
PeriphID and CellID values are determined by macros which take an
index. They currently receive a bus offset which has a stride of 4 bytes.
This causes the ID1-3 registers to report incorrect values.
Scale the offset before passing it to the macro to fix this.

Tested with kvm-unit-tests/arm/pl031.

### DIFF
--- a/usr.sbin/bhyve/rtc_pl031.c
+++ b/usr.sbin/bhyve/rtc_pl031.c
@@ -236,13 +236,13 @@ rtc_pl031_read(struct rtc_pl031_softc *sc, int offset)
 	case RTCPeriphID1:
 	case RTCPeriphID2:
 	case RTCPeriphID3:
-		reg = RTCPeriphID_VAL(offset - RTCPeriphID0);
+		reg = RTCPeriphID_VAL((offset - RTCPeriphID0) >> 2);
 		break;
 	case RTCCellID0:
 	case RTCCellID1:
 	case RTCCellID2:
 	case RTCCellID3:
-		reg = RTCCellID_VAL(offset - RTCCellID0);
+		reg = RTCCellID_VAL((offset - RTCCellID0) >> 2);
 		break;
 	default:
 		/* Return 0 in reads from unasigned registers */


mrkajetanp:
@jrtc27 @zxombie @sarah-walker-arm

usr.sbin/bhyve/rtc_pl031.c
```diff
@@ -236,13 +236,13 @@ rtc_pl031_read(struct rtc_pl031_softc *sc, int offset)
 	case RTCPeriphID1:
 	case RTCPeriphID2:
 	case RTCPeriphID3:
-		reg = RTCPeriphID_VAL(offset - RTCPeriphID0);
+		reg = RTCPeriphID_VAL((offset - RTCPeriphID0) / sizeof(reg));
```
jrtc27: I think I'd just do >> 2, it's tied to the address stride (in particular, of the ID registers), which doesn't necessarily have to be the same as the register size.

mrkajetanp: fine by me, sure!

jrtc27:
> PeriphID and CellID values are determined by macros which expect an index that is then multiplied by 8 and used as a bit shift. Instead, they currently receive offsets and shift a 32-bit register by more than 32 bits. This makes the resulting values incorrect.
>
> Divide the offset by the register size and convert it to an index before passing it to the macro. Tested with kvm-unit-tests/arm/pl031.

This feels like a lot of words of detail without some of the key points. I think it would be better expressed as something like:

> The offset is a bus offset, which has a stride of 4 bytes, but the macro takes an index, causing the ID1-3 registers to report the wrong values. Scale the offset to fix this.

(Note that dividing the offset *is* converting it to an index, so "and ..." isn't appropriate)

jrtc27:
Thanks, pushed, which should close this shortly when the GitHub sync runs.

PR #2360, opened 2026-08-07
fts: close fts_dirfd eagerly when advancing to next sibling

Jeetubhati:
`fts_build()` calls `_dup(_dirfd(dirp))` for every child entry in a directory. Previously these `fds` were held open until each entry was freed, meaning a directory with N children would hold N simultaneous `fds`. On wide directories this could exhaust the per-process `fd `limit.

Fix by closing `fts_dirfd `in `fts_read()` as soon as we advance past a non-directory entry to its sibling. Directory entries must keep their `fts_dirfd `alive longer since fts may need it when descending. This reduces the number of simultaneously open `fts_dirfd ``fds `from N (one per sibling) to at most the current depth of the traversal.

Reported by: Mark Johnston <markj@FreeBSD.org>
Fixes:  4bd01d6ae016 (fts: refactor to use fd-relative operations internally)

### DIFF
--- a/lib/libc/gen/fts.3
+++ b/lib/libc/gen/fts.3
@@ -25,7 +25,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd July 7, 2026
+.Dd August 9, 2026
 .Dt FTS 3
 .Os
 .Sh NAME
@@ -316,9 +316,10 @@ file is a member.
 A parent structure for the initial entry point is provided as well,
 however, only the
 .Fa fts_level ,
-.Fa fts_number
-and
+.Fa fts_number ,
 .Fa fts_pointer
+and
+.Fa fts_dirfd
 fields are guaranteed to be initialized.
 .It Fa fts_link
 Upon return from the
@@ -350,19 +351,24 @@ A pointer to
 .Xr stat 2
 information for the file.
 .It Fa fts_dirfd
-A file descriptor open on the parent directory of this entry.
-It may be used with
+A file descriptor open on this directory entry.
+It is set only for directory entries
+.Pq Dv FTS_D
+and is \-1 for all other entry types.
+To access a file using fd-relative operations without relying
+on path-based syscalls, required in
+.Xr capsicum 4
+capability mode, use
+.Fa fts_parent->fts_dirfd
+with
 .Xr openat 2
 and
-.Fa fts_name
-to access the file without relying on path-based operations,
-which is required in
-.Xr capsicum 4
-capability mode.
-The descriptor is valid only until the next call to
-.Fn fts_read
+.Fa fts_name .
+The descriptor is valid until the directory's post-order visit
+.Pq Dv FTS_DP
 and must not be closed by the caller.
-For root-level entries,
+For root-level entries opened with
+.Fn fts_open ,
 .Fa fts_dirfd
 is \-1.
 .El
--- a/lib/libc/gen/fts.c
+++ b/lib/libc/gen/fts.c
@@ -443,8 +443,8 @@ fts_read(FTS *sp)
 	    (p->fts_info == FTS_SL || p->fts_info == FTS_SLNONE)) {
 		p->fts_info = fts_stat(sp, p, 1, -1);
 		if (p->fts_info == FTS_D && !ISSET(FTS_NOCHDIR)) {
-			if ((p->fts_symfd = p->fts_dirfd >= 0 ?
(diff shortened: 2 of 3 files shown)


asomers (approved):
This change looks incrementally good to me.  But it doesn't go as far as what we discussed on Discord.  Are you still planning to make a change that would avoid initializing fd_dirfd for non-directories in the first place?

Jeetubhati:
> This change looks incrementally good to me. But it doesn't go as far as what we discussed on Discord. Are you still planning to make a change that would avoid initializing fd_dirfd for non-directories in the first place?

Yes, I am planning all the changes we discussed — this commit is just the first step. I will update this PR with the complete fix: adding a private fd field on directory FTSENT entries, copying it to children's fts_dirfd instead of dup() ing per entry, and skipping fts_dirfd entirely for non-directory entries.

Working on it now.

asomers:
cc @markjdb

asomers (changes requested):
Don't forget to update fts.3 too.  And the fts test suite should included something demonstrating the intended use of fts_dirfd (but it's ok if that comes in with the fts_openat PR).

Jeetubhati:
> Don't forget to update fts.3 too. And the fts test suite should included something demonstrating the intended use of fts_dirfd (but it's ok if that comes in with the fts_openat PR).

Updated `fts.3` with the new semantics. Will update it again here once the fts_dirfd design is finalized in `jbhati/fts-dirfd-fix`.

Also added `fts_dirfd_openat `test in the `openat `PR demonstrating the intended use: `fts_parent->fts_dirfd + fts_name` with `openat(2)` to access files without path-based operations.

markjdb:
This patch seems to fix the test suite problems I was seeing, thanks.

lib/libc/gen/fts.3
```diff
@@ -350,17 +350,21 @@ A pointer to
...
 information for the file.
 .It Fa fts_dirfd
-A file descriptor open on the parent directory of this entry.
-It may be used with
+A file descriptor open on this directory entry.
+It is set only for directory entries
+.Pq Dv FTS_D
+and is \-1 for all other entry types.
+To access a file using fd-relative operations without relying
+on path-based syscalls, which is required in
```
asomers:
```suggestion
on path-based syscalls, required in
```

lib/libc/gen/fts.3
asomers: Is this part still true?

asomers:
```suggestion
.Dd August 8, 2026
```

lib/libc/gen/fts.c
```diff
@@ -443,8 +443,8 @@ fts_read(FTS *sp)
 	    (p->fts_info == FTS_SL || p->fts_info == FTS_SLNONE)) {
 		p->fts_info = fts_stat(sp, p, 1, -1);
 		if (p->fts_info == FTS_D && !ISSET(FTS_NOCHDIR)) {
-			if ((p->fts_symfd = p->fts_dirfd >= 0 ?
-			    _dup(p->fts_dirfd) :
+			if ((p->fts_symfd = p->fts_parent->fts_dirfd >= 0 ?
```
asomers, line 446: Are we sure that fts_parent is non-null here?  What if the root itself is a symlink?  The man page says that it is, but I would double check.

lib/libc/gen/fts.3
asomers, line 323: I recommend adding `fts_dirfd` to this list.  And make sure that it really is initialized.  To -1, that is.

asomers (changes requested):

lib/libc/gen/fts.c
Jeetubhati: `fts_parent `is always non-null — all entries including root get a dummy parent allocated in `__fts_open()`. For a root symlink, `fts_parent->fts_dirfd` is -1 so it correctly falls back to `_open(".", O_RDONLY | O_CLOEXEC, 0)`.

lib/libc/gen/fts.3
```diff
@@ -366,9 +369,17 @@ and
...
+.Fn fts_open ,
 .Fa fts_dirfd
 is \-1.
+Root-level entries opened with
+.Fn fts_openat
+using a real
+.Fa dirfd
+argument have
+.Fa fts_dirfd
+set to a duplicate of that descriptor.
```
asomers: References to `fts_openat` don't belong here, since that function hasn't been committed yet.  Please remove this part.

asomers (changes requested):

lib/libc/gen/fts.3
Jeetubhati: Removed it.

PR #2366, opened 2026-08-12
libpmc: Enable overriding PMU counter flags

mashtizadeh:
Typically, PMU counters defined in the JSON have PMC arguments set by
the JSON definition. This change allows the PMC flags to be overridden
by a user. This is useful for some of the advanced flags that allow you
filter on types of events, cores, or threads. While we should expose the
core/thread filters through the CPU mask the problem is that in a number
of systems those ids do match APIC IDs.

For example:
pmcstat -s l3_xi_sampled_latency.all,allcores=0,coreid=0x1

This overrides the default behavior of monitoring the entire processor,
and filters for requests originating from the 2nd core.  Note that in
this case the coreid does not necessarily match the APIC ID.

Sponsored by: Netflix

### DIFF
--- a/lib/libpmc/libpmc.c
+++ b/lib/libpmc/libpmc.c
@@ -1247,17 +1247,13 @@ pmc_allocate(const char *ctrspec, enum pmc_mode mode,
 	 * this is unsupported on the platform, or the event is not found,
 	 * continue with searching the regular event tables.
 	 */
-	r = spec_copy = strdup(ctrspec);
-	ctrname = strsep(&r, ",");
 	if (pmc_pmu_enabled()) {
-		errno = pmc_pmu_pmcallocate(ctrname, &pmc_config);
+		errno = pmc_pmu_pmcallocate(ctrspec, &pmc_config);
 		if (errno == 0)
 			goto found;
 		if (errno == EOPNOTSUPP)
 			goto out;
 	}
-	free(spec_copy);
-	spec_copy = NULL;
 
 	/* replace an event alias with the canonical event specifier */
 	if (pmc_mdep_event_aliases)
--- a/lib/libpmc/libpmc_pmu_util.c
+++ b/lib/libpmc/libpmc_pmu_util.c
@@ -322,6 +322,14 @@ pmc_pmu_event_get_by_idx(const char *cpuid, int idx)
 	return (pme->table[idx].name);
 }
 
+static void
+pmu_init_event(struct pmu_event_desc *ped)
+{
+	bzero(ped, sizeof(*ped));
+	ped->ped_period = DEFAULT_SAMPLE_COUNT;
+	ped->ped_umask = -1;
+}
+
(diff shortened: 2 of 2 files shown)


mashtizadeh:
@mhorne @bsdimp

I ran a tinderbox build to check that I didn't break anything.  Tested the change with AMD's L3 counters.

mhorne:
How would a user do so / what would the command invocation look like?

This should be mentioned in the commit message.

mashtizadeh:
Fixed the commit message, no changes to the code.

lib/libpmc/libpmc_pmu_util.c
```diff
@@ -322,6 +322,14 @@ pmc_pmu_event_get_by_idx(const char *cpuid, int idx)
 	return (pme->table[idx].name);
 }
 
+static void
+pmu_init_event(struct pmu_event_desc *ped)
+{
+	bzero(ped, sizeof(*ped));
+	ped->ped_period = DEFAULT_SAMPLE_COUNT;
+	ped->ped_umask = -1;
+}
```
mhorne, line 331: I am not sure why you split this out... it seems to precede every call to `pmu_parse_event()`, thus no change?

lib/libpmc/libpmc.c
mhorne, line 1267: This is now always true, and should somehow be simplified.

lib/libpmc/libpmc_pmu_util.c
```diff
@@ -712,17 +740,18 @@ pmc_pmu_pmcallocate_md(const char *event_name, struct pmc_op_pmcallocate *pm)
...
+pmc_pmu_pmcallocate_md(const char *ctrspec, struct pmc_op_pmcallocate *pm)
 {
 	const struct pmu_event *pe;
 	struct pmu_event_desc ped;
 	int idx = -1;
 
-	event_name = pmu_alias_get(event_name);
-	if ((pe = pmu_event_get(NULL, event_name, &idx)) == NULL)
+	ctrspec = pmu_alias_get(ctrspec);
+	if ((pe = pmu_event_get(NULL, ctrspec, &idx)) == NULL)
```
mhorne, line 750: I think this will create a regression, as `pmu_event_get()` expects the event name only. Same with the powerpc64 case below.

mhorne (changes requested):

lib/libpmc/libpmc_pmu_util.c
mashtizadeh: No I call parse event twice now from the x86 path to allow us to override parameters.  No other processor has additional parameters like that at the moment.

lib/libpmc/libpmc_pmu_util.c
mashtizadeh: Your right, I forgot to put the strsep code here.

PR #2368, opened 2026-08-13
pmc: Add support for AMD UMC counters

mashtizadeh:
This change adds support for AMD's UMC performance counters.  It is a bit more complicated than existing counters because the enable bit has moved.  This supports Zen 4 through most Zen 6 chips as UMC counters are per-node, where a node does not necessarily translate to a NUMA domain. A few follow up changes to PMC will address this limitation.

Sponsored by: Netflix

### DIFF
--- a/sys/dev/hwpmc/hwpmc_amd.h
+++ b/sys/dev/hwpmc/hwpmc_amd.h
@@ -35,6 +35,7 @@
 #define	CPUID_EXTPERFMON	0x80000022
 #define	EXTPERFMON_CORE_PMCS(x)	((x) & 0x0F)
 #define	EXTPERFMON_DF_PMCS(x)	(((x) >> 10) & 0x3F)
+#define	EXTPERFMON_UMC_PMCS(x)	(((x) >> 16) & 0xFF)
 
 /* AMD K8 PMCs */
 #define	AMD_PMC_EVSEL_0		0xC0010000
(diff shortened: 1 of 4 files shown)


mashtizadeh:
@mhorne @bsdimp

mashtizadeh:
Tested core counters and UMC counters, including maxing ou the counters at the maximum supported by the machine.

mashtizadeh:
Also Mitchell one of the AMD folks is going to add the alloc/free bits for allocated counters, because it’s going to be a problem otherwise for counter multiplexing.

mashtizadeh:
@mhorne @bsdimp Ping.  Can we get this one moving so that one of the AMD folks can help with the Zen 6 bits.

sys/dev/hwpmc/hwpmc_amd.h
```diff
@@ -170,9 +171,34 @@
...
+
+#define	AMD_PMC_UMC_IS_STOPPED(evsel) ((rdmsr((evsel)) & AMD_PMC_UMC_ENABLE) == 0)
+
+#define	AMD_PMC_UMC_TO_EVENTMASK(x)	((x) & AMD_PMC_UMC_EVENTMASK)
+#define	AMD_PMC_UMC_TO_RDWRMASK(x)	(((x) << 8) & AMD_PMC_UMC_RDWRMASK)
+
 #define	AMD_NPMCS_K8		4
 #define AMD_NPMCS_MAX		(AMD_PMC_CORE_MAX + AMD_PMC_L3_MAX + \
-				 AMD_PMC_DF_MAX)
+				 AMD_PMC_DF_MAX + AMD_PMC_UMC_MAX)
```
mhorne, line 201: This is quickly becoming a massive allocation, and probably needs to be handled in a smarter/more dynamic way. Consider that many processors will not have any use for 256 UMC counters, but this is bloating the per-CPU allocations as well as the static event descriptors `amd_pmcdesc[AMD_NPMCS_MAX]`.

mhorne (approved):
LGTM.

sys/dev/hwpmc/hwpmc_amd.h
emaste: Yeah, we have a number of allocations that are scaled based on some maximum (often `MAXCPU`) rather than dynamically based on the actual number of CPUs, descriptors, etc. I'd like to have some common, greppable comment that we can use to find these.

Maybe `XXXBLOAT`

mashtizadeh: I agree, we need to have this fixed anyways as part of the patches for counter multiplexing so it won't be this way very long.

mhorne: That's great; I will merge this change ASAP.

PR #2371, opened 2026-08-15
fix: use IANA reserved NFS-over-RDMA port

viniciusferrao:
This fixes a typo in the NFSoRDMA IANA
reserved port which is 20049.


viniciusferrao:
@rmacklem just a quick typo fix.

viniciusferrao:
Already landed on https://github.com/freebsd/freebsd-src/commit/3ce28106a887766383c99e5a607396f66ba2e5c1

PR #2372, opened 2026-08-15
fts: fix trailing-slash regression in fts_read after fts_children

Jeetubhati:
When `fts_read()` descends into a directory whose children were already prefetched by `fts_children()` (as ls -R does), it changed directory using `p->fts_name` instead of `p->fts_accpath.`

With a trailing slash on a relative root path (e.g. '`dir/`'), the bare name was resolved relative to the wrong directory, so every sibling directory after the first failed with `ENOENT `and was reported as `FTS_DNR`.  This manifested as '`ls -lR dir/`' skipping the contents of all but the first subdirectory.

Restore the use of `p->fts_accpath`, matching the behavior prior to `4bd01d6ae016`.

Add a regression test that reproduces the exact conditions: `fts_children()` on each directory, `FTS_PHYSICAL `without `FTS_NOCHDIR`, and a trailing slash on the root path.

Reported by:    Michael Butler `<imb@protected-networks.net>`
Fixes:          `4bd01d6ae016`

### DIFF
--- a/lib/libc/gen/fts.c
+++ b/lib/libc/gen/fts.c
@@ -489,7 +489,7 @@ fts_read(FTS *sp)
 		 * FTS_STOP or the fts_info field of the node.
 		 */
 		if (sp->fts_child != NULL) {
-			if (fts_safe_changedir(sp, p, -1, p->fts_name)) {
+			if (fts_safe_changedir(sp, p, -1, p->fts_accpath)) {
 				p->fts_errno = errno;
 				p->fts_flags |= FTS_DONTCHDIR;
 				for (p = sp->fts_child; p != NULL;
--- a/lib/libc/tests/gen/fts_regress_test.c
+++ b/lib/libc/tests/gen/fts_regress_test.c
@@ -362,6 +362,63 @@ ATF_TC_BODY(accpath_correct_after_descent, tc)
 	ATF_REQUIRE_EQ_MSG(0, fts_close(fts), "fts_close(): %m");
 }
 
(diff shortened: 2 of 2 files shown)


asomers (commented):
The test looks good.  But I don't understand the bug.  Where is the trailing-slash-sensitive code?  And which "wrong directory" was the path resolved relative to?

Jeetubhati:
> The test looks good. But I don't understand the bug. Where is the trailing-slash-sensitive code? And which "wrong directory" was the path resolved relative to?

Where the trailing-slash-sensitive code is: the `sp->fts_child != NULL` branch in `fts_read `(the path taken when `fts_children()` prefetched the children, i.e. ls -R). The trailing slash makes the root entry's fts_name empty, so its subdirectory children get `fts_name = 2026-07-03` but `fts_accpath = george/2026-07-03`.

Which wrong directory: when descending from the root into its first child, fts hasn't `chdir'd` into george yet `CWD `is still the starting directory. `fts_safe_changedir(sp, p, -1, p->fts_name)` does `_open(p->fts_name)` relative to `CWD`, so `_open("2026-07-03")` resolves against the starting `dir `instead of george → `ENOENT`. `p->fts_accpath (george/2026-07-03)` is correct relative to that `CWD`, which is why restoring it fixes it.

I confirmed the `CWD `by instrumenting it: `name='2026-07-03' accpath='george/2026-07-03' CWD=/home/jitendra`.

asomers (approved):

PR #2373, opened 2026-08-15
fts: fix file descriptor leak in fts_close

Jeetubhati:
`fts_build()` stores a `dup'd` file descriptor in each directory entry's `fts_dirfd`.  When a traversal is abandoned before completion and `fts_close()` is called, the cleanup loop freed each pending entry with `free()` without first closing its `fts_dirfd`, leaking one descriptor per pending directory.

Close `fts_dirfd `before freeing each entry in the cleanup loop, matching the handling already applied to the dummy parent entry after the loop.

Add a regression test that descends a couple of levels, abandons the traversal, closes, and asserts the open descriptor count is unchanged.

PR:             `297557`
Reported by:    asomers
Fixes:          `4bd01d6ae016`

### DIFF
--- a/lib/libc/tests/gen/fts_misc_test.c
+++ b/lib/libc/tests/gen/fts_misc_test.c
@@ -6,6 +6,7 @@
  */
 
 #include <sys/mount.h>
+#include <sys/sysctl.h>
 #include <sys/param.h>
 #include <sys/stat.h>
 #include <sys/syslimits.h>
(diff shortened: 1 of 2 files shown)


lib/libc/tests/gen/fts_misc_test.c
```diff
@@ -576,6 +577,61 @@ ATF_TC_CLEANUP(xdev, tc)
...
+	 * those fds; previously it freed the entries without closing
+	 * fts_dirfd, leaking one fd per pending directory.
+	 */
+	ATF_REQUIRE_EQ(0, mkdir("sub1", 0755));
+	ATF_REQUIRE_EQ(0, mkdir("sub1/deep", 0755));
+	ATF_REQUIRE_EQ(0, close(creat("sub1/deep/file", 0644)));
+
+	/* Count currently-open descriptors. */
+	fds_before = 0;
+	for (fd = 0; fd < 256; fd++)
```
asomers: This loop is hacky.  I think you can look up the file descriptor count more easily by using the `kern.proc.nfds` sysctl.  Could you please try it that way?

asomers (changes requested):

lib/libc/tests/gen/fts_misc_test.c
```diff
@@ -577,6 +577,23 @@ ATF_TC_CLEANUP(xdev, tc)
...
+ */
+static int
+count_open_fds(void)
+{
+	int mib[4] = { CTL_KERN, KERN_PROC, KERN_PROC_NFDS, 0 };
+	int nfds;
+	size_t len = sizeof(nfds);
+
+	ATF_REQUIRE_EQ_MSG(0,
+	    sysctl(mib, 4, &nfds, &len, NULL, 0),
```
asomers: Best to avoid magic numbers:
```c
    sysctl(mib, nitems(mib), &nfds, &len, NULL, 0),
```

asomers (changes requested):

asomers (approved):

PR #2374, opened 2026-08-16
man: Refine historical details about cut(1) and paste(1)

dhgutteridge:


dhgutteridge:
Two simple changes here. The commit that added the authorship attributions[1] seems to have flipped them between files. cut(1) has the dual authorship of Adam S. Moskowitz and Marciano Pitargue, going by the copyright notice in cut.c (which is also what the OpenBSD man page that this commit sourced from states). paste.c only lists Adam S. Moskowitz.

Separately, paste(1) was changed to state it first appeared in AT&T System III vs. 32V, which altered the earlier commit by Tim J. Robbins[2]. Going by the TUHS archive of 32V, that's not correct: paste(1) appeared in (some) UNIX sources before cut(1) did. Looking at the review[3], I see there was discussion about this, but imp@ stated that cut(1) wasn't in 32V, not that paste(1) was not. It seems this feedback was conflated to apply to both man pages.

```
% tar tvzf 32v_usr.tar.gz | grep paste
-rwxr-xr-x mhol/wheel     3800 1979-03-25 01:53 usr/bin/paste
-rw-r--r-- mhol/wheel     2994 1978-11-05 23:03 usr/src/cmd/paste.c
% tar tvzf 32v_usr.tar.gz | grep cut
%
```

1. [https://github.com/freebsd/freebsd-src/commit/9bdb5158d6016798187e4d7eb943791a64809194](https://github.com/freebsd/freebsd-src/commit/9bdb5158d6016798187e4d7eb943791a64809194)
2. [https://github.com/freebsd/freebsd-src/commit/30f8b6c18279162a2f4efdc9e85a49683a8aafbf](https://github.com/freebsd/freebsd-src/commit/30f8b6c18279162a2f4efdc9e85a49683a8aafbf)
3. [https://reviews.freebsd.org/D36048](https://reviews.freebsd.org/D36048)

bsdimp:
So, looking at TUHS, this does appear to be in 32V, but no manual is in 32V. The source that's in our 32V tape matches the System III sources that are extant and the System V sources. There's no paste in BSD until after 4.3BSD and that source is completely different. There's references in it to 127 paste command in the sources. So the system III path makes sense: USDL copied it from a similar command that was circulating at bell labs, but wasn't part officially of research unix (and I don't think we have a copy of it). The paste.c that's in 8th edition is from the system III / 32V sources. It's weird that it's in 32V, but not in V7 (32V is supposed to be just the V7 sources ported minimally to the VAX), nor in 3BSD, which is derived directly from 32V. On the mailing list, it was documented that there were several versions of 32V, but we only have one surviving. What's even weirder is that it's from before many of the surviving v7 tapes. The versions and revisions of Unix in this time period are a bit weird.

Anyway, the one in SysIII is clearly a later version of the 32V, so I suppose this is OK. But I'm not sure how to interpret that 3BSD, which we know is derived from 32V, doesn't have it.

dhgutteridge:
> So, looking at TUHS, this does appear to be in 32V, but no manual is in 32V. The source that's in our 32V tape matches the System III sources that are extant and the System V sources. There's no paste in BSD until after 4.3BSD and that source is completely different. There's references in it to 127 paste command in the sources. So the system III path makes sense: USDL copied it from a similar command that was circulating at bell labs, but wasn't part officially of research unix (and I don't think we have a copy of it). The paste.c that's in 8th edition is from the system III / 32V sources. It's weird that it's in 32V, but not in V7 (32V is supposed to be just the V7 sources ported minimally to the VAX), nor in 3BSD, which is derived directly from 32V. On the mailing list, it was documented that there were several versions of 32V, but we only have one surviving. What's even weirder is that it's from before many of the surviving v7 tapes. The versions and revisions of Unix in this time period are a bit weird.
> 
> Anyway, the one in SysIII is clearly a later version of the 32V, so I suppose this is OK. But I'm not sure how to interpret that 3BSD, which we know is derived from 32V, doesn't have it.

This history is more complicated than I imagined, though I shouldn't be surprised. I was partly just leaning on the commit from Tim J. Robbins, some of whose other work with cut(1) I'd heavily borrowed recently. Now I'm curious about 3BSD, too.

PR #2375, opened 2026-08-18
chflags: use Capsicum capability mode

Jeetubhati:
**`chflags`: use Capsicum capability mode**

Converts `chflags(1)` to run its file-hierarchy traversal inside Capsicum capability mode using the `fts_openat(3)` API. This confines `chflags `to the directory hierarchies it was actually asked to operate on, so a malicious or buggy tree cannot redirect it at files outside the intended paths.

**Approach**

`AT_FDCWD `is rejected in capability mode, even for relative paths, so the traversal needs a real directory descriptor to root each argument at. Before` cap_enter()`:

A descriptor for the current directory `(.)` is opened once and used as the base for every plain relative argument.
Absolute paths and paths containing .. cannot be resolved relative to another descriptor in capability mode, so each of those gets its own pre-opened parent descriptor, with the argument rewritten to the trailing component.

With every needed descriptor opened up front, `cap_enter()` is called and the traversal runs entirely through fd-relative operations: `chflagsat(fts_parent->fts_dirfd, fts_name, ...)`.

**Testing**

No functional change for any argument form. Verified: relative, absolute, and .. paths; -R recursion; -H/-L/-P symlink modes; and -h operating on a symlink itself. New functional tests cover relative, absolute, .., recursive, and mixed arguments, and skip on filesystems that don't support the `uchg `flag.

### DIFF
--- a/bin/chflags/chflags.1
+++ b/bin/chflags/chflags.1
@@ -29,7 +29,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd July 9, 2026
+.Dd August 21, 2026
 .Dt CHFLAGS 1
 .Os
 .Sh NAME
@@ -38,6 +38,7 @@
 .Sh SYNOPSIS
 .Nm
 .Op Fl fhvx
+.Op Fl -dereference-links-unsafely
 .Oo
 .Fl R
 .Op Fl H | Fl L | Fl P
@@ -70,11 +71,20 @@ and hence unaffected by the command.
 If the
 .Ar file
 is a symbolic link,
-change the file flags of the link itself rather than the file to which it points.
+change the file flags of the link itself
+rather than the file to which it points.
 .It Fl L
 If the
 .Fl R
-option is specified, all symbolic links are followed.
+option is specified, symbolic links are followed.
+A symbolic link encountered during the traversal that points outside
+the file hierarchy being traversed is not followed unless the
+.Fl -dereference-links-unsafely
+option is also given;
+see the description of that option below.
+A symbolic link named directly on the command line is followed unless
+.Fl h
+is specified.
 .It Fl P
 If the
 .Fl R
@@ -97,6 +107,21 @@ option is specified more than once, the old and new flags of the file
 will also be printed, in octal notation.
 .It Fl x
 Do not cross mount points.
+.It Fl -dereference-links-unsafely
+By default,
+.Nm
+runs its traversal inside a
+.Xr capsicum 4
+sandbox, which prevents it from following a symbolic link whose
+target lies outside the file hierarchy named on the command line.
+This option disables the sandbox so that such symbolic links are
+followed, restoring historical behavior.
+It is only meaningful together with
+.Fl L
+or the default logical behavior, and should be used with caution,
+as it allows a crafted symbolic link to redirect
+.Nm
+at files outside the intended hierarchy.
 .El
 .Pp
 The flags are specified as an octal number or a comma separated list
--- a/bin/chflags/chflags.c
+++ b/bin/chflags/chflags.c
@@ -133,14 +133,14 @@ main(int argc, char *argv[])
 		if (val < 0)
 			errno = ERANGE;
 		if (errno)
-                        err(1, "invalid flags: %s", flags);
-                if (*ep)
-                        errx(1, "invalid flags: %s", flags);
+			err(1, "invalid flags: %s", flags);
+		if (*ep)
+			errx(1, "invalid flags: %s", flags);
 		set = val;
-                oct = 1;
+		oct = 1;
 	} else {
 		if (strtofflags(&flags, &set, &clear))
-                        errx(1, "invalid flag: %s", flags);
+			errx(1, "invalid flag: %s", flags);
 		clear = ~clear;
 		oct = 0;
 	}
--- a/bin/chflags/chflags.c
+++ b/bin/chflags/chflags.c
@@ -30,19 +30,28 @@
  */
 
 #include <sys/types.h>
+#include <sys/capsicum.h>
 #include <sys/stat.h>
 
+#include <assert.h>
+#include <capsicum_helpers.h>
 #include <err.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <fts.h>
+#include <getopt.h>
+#include <libgen.h>
+#include <limits.h>
 #include <signal.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
 
 static volatile sig_atomic_t siginfo;
+
+#define	OPT_DEREF_UNSAFE	(CHAR_MAX + 1)
 
 static void usage(void) __dead2;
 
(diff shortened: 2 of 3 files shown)


asomers (changes requested):
This commit mixes style changes and functional changes.  In this case, the style changes fix existing style bugs (i.e. wrong indentation style on line 136).  But still, that makes review difficult.  Please either revert the style changes, or else separate them into their own commit, preceding the functional changes.

bin/chflags/tests/chflags_test.sh
```diff
@@ -55,8 +68,91 @@ no_arguments_body()
...
 }
 
+atf_test_case relative_path
+relative_path_head()
+{
+	atf_set "descr" "chflags sets flags on a relative path"
+}
+relative_path_body()
+{
+require_chflags
```
asomers: Why don't you indent `require_chflags`?

asomers (commented):
Another troublesome use case is using `chflags -RL` on a link whose destination may point outside of the command line argument's starting directory.  For example:
```
file
foo/
    bar/
        baz/
            file@ -> ../../file
```
In this case, running `chflags -Rl hidden foo/bar` would open a dirfd for `foo/bar`, so attempting to follow the `file@` link would fail.  However, prior to your change, that operation would succeed and it would change the flags on `file`.  Achieving perfect backwards-compatibility would probably require skipping `cap_enter` whenever `-L` is provided.  However, tricky symlinks like this are *exactly* the kind of threat that we want to guard against!  Arguably, chflags *shouldn't* follow symlinks like that.  Should we preserve existing behavior or add a new guardrail?  That decision probably warrants a discussion on freebsd-hackers@FreeBSD.org .

bin/chflags/tests/chflags_test.sh
```diff
@@ -55,8 +68,136 @@ no_arguments_body()
...
+{
+	atf_set "descr" "chflags sets flags on a relative path"
+}
+relative_path_body()
+{
+	require_chflags
+	touch file
+	atf_check chflags uchg file
+	atf_check -o match:uchg stat -f "%Sf" file
+	atf_check chflags nouchg file
```
asomers: Are you setting nouchg for purposes of cleanup?  If so, you should delete it.  ATF will delete it anyway.  Unless ATF can't delete it, because that flag is set?  In that case, it's not the best flag to use.  You could try a flag like `hidden` or `nodump` instead.

bin/chflags/tests/chflags_test.sh
```diff
@@ -55,8 +68,136 @@ no_arguments_body()
...
+	touch target
+	ln -s ../../../target foo/bar/baz/link
+	atf_check -s not-exit:0 -e ignore chflags -RL uchg foo/bar
+	atf_check -o match:"target[[:space:]]*-" stat -f "%N %Sf" target
+}
+
+outside_symlink_rejected_cleanup()
+{
+	chflags -R --dereference-links-unsafely 0 foo target 2>/dev/null || true
+}
```
asomers: This whole step can probably be removed if you use a different flag than uchg.

bin/chflags/chflags.1
asomers: Don't forget to change .Dd

asomers: "all" is no longer accurate.  You should update this sentence.

bin/chflags/tests/chflags_test.sh
```diff
@@ -55,8 +68,136 @@ no_arguments_body()
...
 	atf_add_test_case invalid_usage
 	atf_add_test_case no_arguments
+	atf_add_test_case relative_path
+	atf_add_test_case absolute_path
+	atf_add_test_case dotdot_path
+	atf_add_test_case recursive
+	atf_add_test_case mixed_paths
+	atf_add_test_case outside_symlink_rejected
+	atf_add_test_case outside_symlink_unsafe
 }
```
asomers, line 238: Could you also add a test for `chflags -h`?  We want to ensure that we don't break that.

bin/chflags/chflags.1
```diff
@@ -97,6 +99,21 @@ option is specified more than once, the old and new flags of the file
...
+.Nm
+runs its traversal inside a
+.Xr capsicum 4
+sandbox, which prevents it from following a symbolic link whose
+target lies outside the file hierarchy named on the command line.
+This option disables the sandbox so that such symbolic links are
+followed, restoring historical behavior.
+It is only meaningful together with
+.Fl L
+or the default logical behavior, and should be used with caution,
```
asomers, line 121: What's this about "default logical behavior"?

bin/chflags/chflags.1
```diff
@@ -97,6 +99,21 @@ option is specified more than once, the old and new flags of the file
...
 Do not cross mount points.
+.It Fl -dereference-links-unsafely
+By default,
+.Nm
+runs its traversal inside a
+.Xr capsicum 4
+sandbox, which prevents it from following a symbolic link whose
+target lies outside the file hierarchy named on the command line.
+This option disables the sandbox so that such symbolic links are
+followed, restoring historical behavior.
```
asomers, line 118: IMHO, capsicum is an implementation detail.  We needn't mention it here.  Also, I think we should mention the release when it's behavior changed.  So I suggest wording like this:
```
refuses to follow symbolic links whose target lies outside of its parent directory
named on the command line.
This option causes all symbolic links to be followed, restoring the historic behavior of
.Fx 15.1 .
```

bin/chflags/chflags.c
```diff
@@ -53,19 +60,168 @@ siginfo_handler(int sig __unused)
...
+	slash = strrchr(*path, '/');
+	if (slash == NULL)
+		return (open(".", O_RDONLY | O_DIRECTORY | O_CLOEXEC));
+	if (slash == *path) {
+		/* Path is directly under "/". */
+		fd = open("/", O_RDONLY | O_DIRECTORY | O_CLOEXEC);
+		*path = slash + 1;
+		return (fd);
+	}
+	dir = strndup(*path, slash - *path);
```
asomers: This is going to behave differently depending on whether path contains a trailing slash: that is `/var/tmp/` vs `/var/tmp`.  I think it would be better to use `dirname(3)`, which will handle the variable trailing slash for you.

bin/chflags/chflags.c
```diff
@@ -138,80 +297,96 @@ main(int argc, char *argv[])
 			errx(1, "invalid flags: %s", flags);
 		set = val;
 		oct = 1;
+		clear = 0;
```
asomers: Why this line?

bin/chflags/chflags.c
```diff
@@ -138,80 +297,96 @@ main(int argc, char *argv[])
...
-	for (rval = 0; errno = 0, (p = fts_read(ftsp)) != NULL;) {
-		int atflag;
+	/*
+	 * With --dereference-links-unsafely the traversal may follow a
+	 * symlink to a file outside the hierarchy named on the command
+	 * line, which capability mode cannot allow.  Keep the historical
+	 * path-based traversal rooted at AT_FDCWD in that case.
+	 */
+	if (unsafe) {
+		rval = chflags_fts(AT_FDCWD, argv, fts_options, false, set,
```
asomers: Why do you skip opening dirds in unsafe mode?  Are you doing that just as an optimization?  If so, I advise against it.  unsafe mode will be rare.  Better, IMHO, to keep the code as similar as possible between the two cases.  In unsafe mode, we can probably behave exactly like we do in safe mode, but skip `cap_enter`.

bin/chflags/chflags.c
```diff
@@ -138,80 +297,96 @@ main(int argc, char *argv[])
...
-		if (newflags == p->fts_statp->st_flags)
-			continue;
-		if (chflagsat(AT_FDCWD, p->fts_accpath, newflags,
-		    atflag) == -1) {
-			e = errno;
-			if (!fflag) {
-				warnc(e, "%s", p->fts_path);
+	relpaths = calloc(argc + 1, sizeof(*relpaths));
+	ownfd = calloc(argc, sizeof(*ownfd));
+	ownpath = calloc(argc, sizeof(*ownpath));
```
asomers: Don't forget to free these.  Even in a short-lived program where dynamically memory is used for the program's entire lifetime, it's important to free it so it won't be reported by tools like valgrind.  Otherwise valgrind users won't be able to tell which memory was accidentally leaked and which memory was intended to live until the program exitted.

bin/chflags/chflags.c
```diff
@@ -138,80 +297,96 @@ main(int argc, char *argv[])
...
-			}
-		} else if (vflag || siginfo) {
-			(void)printf("%s", p->fts_path);
-			if (vflag > 1 || siginfo)
-				(void)printf(": 0%lo -> 0%lo",
-				    (u_long)p->fts_statp->st_flags,
-				    newflags);
-			(void)printf("\n");
-			siginfo = 0;
+			twopath = calloc(2, sizeof(*twopath));
```
asomers: Don't forget to free this too.  In fact, you should probably put `twopath` on the stack, since it's only 16 bytes.  If you do that, then you can store `path` directly in `ownpath` on line 359, and assign `twopath[0] = ownpath[i]` just before calling `chflags_fts` on line 379.

asomers (changes requested):

bin/chflags/tests/chflags_test.sh
asomers: Err, actually I mean chflags directly on a symlink, _without_ `-h`.  That's the more interesting case.

bin/chflags/tests/chflags_test.sh
```diff
@@ -179,14 +168,23 @@ outside_symlink_unsafe_body()
...
+symlink_no_h_head()
 {
-	chflags -R --dereference-links-unsafely 0 foo target 2>/dev/null || true
+	atf_set "descr" "chflags on a symlink without -h changes the target"
+}
+symlink_no_h_body()
+{
+	require_chflags
+	touch target
+	ln -s target link
```
asomers: This test is not as good as it could be, because the link and its target lie in the same directory.  It would be more interesting if the target were in the link's parent's directory.  That would ensure that capsicum isn't accidentally breaking following links that are specified on the command line.

bin/chflags/chflags.c
```diff
@@ -309,75 +288,70 @@ main(int argc, char *argv[])
...
 		err(1, "caph_limit_stdio");
-	if (caph_enter() < 0)
+	/*
+	 * With --dereference-links-unsafely the traversal may follow a
+	 * symlink to a file outside the hierarchy named on the command
+	 * line, which capability mode would block, so skip cap_enter() in
+	 * that case.
+	 */
+	if (!unsafe && caph_enter() < 0)
 		err(1, "cap_enter");
```
asomers: Slight spelling error here: caph_enter instead of cap_enter .

bin/chflags/chflags.c
```diff
@@ -309,75 +288,70 @@ main(int argc, char *argv[])
...
 
 	/* Process each absolute / ".."-containing path under its own fd. */
+	for (i = 0; i < nown; i++) {
+		twopath[0] = ownbase[i];
+		twopath[1] = NULL;
+		rval |= chflags_fts(ownfd[i], twopath, fts_options, set,
+		    clear, oct, Rflag, fflag, vflag);
+	}
+
 	for (i = 0; i < nown; i++)
```
asomers: Why have a separate loop here?  I think you can combine it with the above loop.

bin/chflags/tests/chflags_test.sh
asomers, line 2: Oopsie , need to update the copyright header.

bin/chflags/tests/chflags_test.sh
```diff
@@ -180,11 +180,29 @@ symlink_no_h_head()
...
+}
+symlink_rootlevel_H_body()
+{
+	require_chflags
+	mkdir dir realdir
+	touch realdir/file
+	ln -s ../realdir dir/dlink
+	atf_check chflags -RH nodump dir/dlink
+	atf_check -o match:nodump stat -f "%Sf" realdir/file
+	atf_check -o match:"dlink[[:space:]]*-" stat -f "%N %Sf" dir/dlink
```
asomers: It might be more clear to do `atf_check -o not-match:nodump stat -f "%Sf" dir/dlink`

bin/chflags/chflags.1
```diff
@@ -77,11 +77,12 @@ rather than the file to which it points.
...
 .Fl R
 option is specified, symbolic links are followed.
-Symbolic links that point outside the file hierarchy being traversed
-are not followed unless the
+A symbolic link encountered during the traversal that points outside
+the file hierarchy being traversed is not followed unless the
 .Fl -dereference-links-unsafely
 option is also given;
 see the description of that option below.
+A symbolic link named directly on the command line is always followed.
```
asomers: Not "always", because it won't be followed when `-h` is used.

bin/chflags/chflags.c
```diff
@@ -318,7 +340,7 @@ main(int argc, char *argv[])
 			ownfd[nown] = fd;
 			nown++;
 		} else {
-			relpaths[nrel++] = argv[i];
+			relpaths[nrel++] = arg;
```
asomers: If `realpath` succeeded, then `arg` will point to `resolved_path`, which is on the stack, and will fall out of scope on line 345.  It will be a dangling pointer when passed to `chflags_fts`.

bin/chflags/chflags.c
```diff
@@ -308,8 +308,30 @@ main(int argc, char *argv[])
...
+		 * own parent.
+		 */
+		if ((fts_options & (FTS_LOGICAL | FTS_COMFOLLOW)) &&
+		    lstat(arg, &sb) == 0 && S_ISLNK(sb.st_mode)) {
+			if (realpath(arg, resolved) == NULL) {
+				warn("%s", arg);
+				rval = 1;
+				continue;
+			}
+			arg = resolved;
```
asomers, line 331: It would be good to add another test case that uses `-RH` on a dead link, just to exercise this path.

asomers (changes requested):

bin/chflags/tests/chflags_test.sh
asomers: Why did you leave Shivansh Rai's copyright here?  Does any part of this file date from his work, except for the copyright header itself?

bin/chflags/tests/chflags_test.sh
```diff
@@ -202,7 +203,21 @@ symlink_rootlevel_H_body()
...
+{
+	atf_set "descr" "chflags -RH on a broken symbolic link fails" \
+	    " gracefully"
+}
+symlink_rootlevel_H_dead_body()
+{
+	require_chflags
+	mkdir dir
+	ln -s ../nonexistent dir/dlink
+	atf_check -s not-exit:0 -e not-empty chflags -RH nodump dir/dlink
```
asomers: It would be good to assert that "No such file or directory" appears in the stderr.  What we don't want is something about "Capabilities insufficient"

bin/chflags/tests/chflags_test.sh
asomers: Jitendra explained on Discord that two of the tests in this file actually do date from Shivansh's time.  I missed that.

asomers (approved):
I think this looks good now. I'll test it and hopefully commit later today.

PR #2377, opened 2026-08-19
stand: compile in a stricter veriexec default, established early

johannes-bruegmann:
**Summary.** Move setting veriexec's strict mode into C, so a stricter default is
compiled into the signed loader and is in force before any file is read — instead
of relying on a deletable `/boot/loader.ve.strict`. That is the change ([1]);
[2]–[4] are related weak spots below, offered for discussion, not fixed here.

An apology for the length — you were right that the original was hard to approach.
The line above is the "what"; the detail below is the "why", including why a
deletable file was there on purpose, because it looks arbitrary without it.

**What changed.** This supersedes the earlier approach (an early verified read of
`/boot/loader.ve.strict`). Following @bsdimp's point that a deletable file is a
poor place for the decision, it moves the strict default into the compiled,
signed loader. Two opt-in knobs under `LOADER_VERIEXEC`, wired like
`LOADER_VERIEXEC_VECTX`:

- **`LOADER_VERIEXEC_STRICT`** — `ACCEPT_NO_FP_DEFAULT = VE_WANT`. There is no
  knob for a stricter default today (`LOADER_VERIEXEC_TESTING` only relaxes it).
- **`LOADER_VERIEXEC_ELEVATED`** (implies STRICT) — a new `ve_verifying_set()`
  turns verification on early and fail-closed in `efi_main()`, so the default
  takes effect before any file or unverified input is read. Requires a
  compiled-in trust anchor.

**What this is for.** I run FreeBSD with `/boot` on a removable microSD card — a
medium I must assume is exposed to tampering — and I'm building a small set of
trust gates into the boot path: local tamper detection, not prevention. For that,
strict verification has to be in force as early as possible, so that every file
the loader uses has genuinely been verified first.

**On the deletable file.** In hindsight the original version wasn't as careless as
it looked. In a detection mindset a soft "this should be here" is arguably harder
on an attacker than a hard wall: they cannot tell whether tripping it was noticed,
or what happens next. It also keeps maximum compatibility with today's boot — with
no gates compiled in, behaviour is exactly as before, and a user who simply does
not want strict can still delete the file, as they can today. The compile-in
removes that escape hatch; whether that matters is your call, not mine. Both
approaches work for me, and detection does not depend on which is chosen — so I
have reworked this in the compile-in direction you asked for.

The findings in [1]–[4] raise the question of whether the elevation to strict
mode takes hold early enough, and reliably enough, to underpin that. **This change
addresses [1].** [2]–[4] are laid out in full below as the basis for a discussion,
not fixed here; each would be a separate, small follow-up.

A caveat on scope: these are a first assessment of where the elevation looks thin,
not an exhaustive audit. I've already noted further candidates I haven't worked
through yet — for instance typed data-blob preloads (e.g. `/boot/entropy`) that
appear to load without going through `verify_file` — and there are very likely
more. I'd rather offer this as an evolving list than imply it is complete.

None of this is a complaint about the code. FreeBSD doesn't really support this
use today — establishing verification state early and acting on it — and doing it
in a way that fits the tree's idioms is where I'm unsure, because I don't know
where you intend veriexec to go; I'd rather align with that than guess.

---

## [1] Strict mode is not established early or reliably (addressed here)

**The threshold is inert while verification is off.** `ACCEPT_NO_FP_DEFAULT`
(`verify_file.c:522`) is consulted only at the no-fingerprint tolerance
(`:581`, `severity < accept_no_fp`) and the tampered-file check (`:602`,
`severity > accept_no_fp`). Both sit behind `verify_prep()`, which returns
"accept" as soon as verification is off:

```
verify_prep():  if (!Verifying || fd < 0) return (0);   /* :469 */
verify_file():  rc = verify_prep(...); if (!rc) return (0);
```

So while `Verifying < 1`, a compiled-in `VE_WANT` changes nothing — every file is
accepted. Threshold and activation are independent levers; a compile-in default
is one, not both.

**Activation is a lazy side effect that can fail open.** `Verifying`
(`:301`, init `-1`) is set once, on the first verified open, from
`ve_trust_init()` (`:459`). If that returns 0 — no or failed trust anchor — it
latches off for the whole boot, and the loader still reports a verified boot.
`ve_trust_init()` is itself a coarse signal: it sums anchor counts and a revoked
anchor still counts (revocation moves it to a separate `forbidden_anchors` vector,
`vets.c`), so `> 0` means "an anchor object exists", not "the chain can verify".

**The self-test is computed and discarded.** `verify_prep` runs `ve_self_tests()`
(`:465`) — which cryptographically checks the trust anchor against its own
signature — but ignores the result; only `verify_tweak("strict")` acts on it, and
that path needs the tweak file.

**The only early route today is a deletable file, via a cycle.** Raising the mode
before the interpreter means a verified `loader.ve.*` read, but the tweak arrives
through `loader_conf_files` → `config.lua`, and `config.lua` is itself `VE_MUST`,
loaded and verified first — under the *previous* mode. And nothing requires the
file to exist (see [3]); deleting `/boot/loader.ve.strict` silently reverts to the
`VE_MUST` default.

**The config chain then admits unlisted overrides.** With only `VE_MUST` in force,
the chaining variables — `loader_conf_files`, `loader_conf_dirs`,
`local_loader_conf_files` (defaults: `/boot/loader.conf.d`,
`/boot/loader.conf.local`), and `nextboot.conf` (`config.lua:825`) — are
unrestricted and read *after* `loader.conf`, so they override it. An unlisted
`.conf` is `VE_WANT`, and `severity < accept_no_fp` (`1 < 2`) tolerates it
(`:581`); since config-file assignment goes through `setenv()` (`lutils.c:221`),
not the `is_restricted_var`-gated `boot_setenv()` (`environment.c:294`), it can set
*any* variable. So a planted `/boot/loader.conf.local` overrides `loader.conf`
unrestricted — no signed file touched.

**What this change does.** `ve_verifying_set()` establishes `Verifying` once,
early, before the first file: it takes `ve_trust_init()`, **panics** if the trust
base does not come up (fail-closed, not the current silent fail-open), and
**enforces** `ve_self_tests()`. With `VE_WANT` then in force from the first open,
`config.lua` and the whole chaining/override surface above refuse unlisted files.
`ve_verifying_set()` lives under `LOADER_VERIEXEC_ELEVATED` (needs a compiled-in
anchor); the tweak-open in `interact()` remains as the file fallback, consulted
only while `Verifying < 1` — i.e. never, once the switch has run.

Whether the elevated mode is *still* in force at runtime is in turn something a
gate can measure and publish — that companion detection work is separate from this
change.

**Measured** (15.1-RELEASE-p2, amd64, Secure Boot enforcing, `/boot` on ZFS): a
clean boot passes with the elevated mode in force. The discriminating runs — a
panic on a tampered `VE_MUST` file, refusal of a planted `loader.conf.local`, a
panic on a failed trust base — follow from the paths above; I can supply the exact
boots on request.

One related input path this change does **not** close by itself: `loader.env`
(`read_loader_env()` in `main.c`) is parsed raw, outside `verify_file`, so it is
never verified. Its reach is bounded by `is_restricted_var`, but that still leaves
`loader_conf_files`, `module_path`, `password`. We treat it as a hole too: on this
machine we compile it out entirely — we moved the preceding `#endif` to sit *after*
the `read_loader_env()` calls. I'd be glad to fold that into this PR if you'd want
it, or to verify/gate the file instead.

## [2] Classes tolerated — and unreported — even under strict (severity floor)

`severity_guess()` maps `.cookie`, `.hints`, `.order`, `.tgz` to `VE_TRY`. Under
strict (`accept_no_fp = VE_WANT`), `VE_TRY < VE_WANT` so these are still tolerated
(`:581`), and the code comments that it "will not report them" — so a swap is
silent. `device.hints` is the concrete case: it sits in the default
`loader_conf_files`, is parsed as config, and an unlisted one is accepted.

Separately, read paths do not verify at all: `more`/`page_file` reads a file
without `verify_file` (observed: `more /boot/keys/zroot.key` succeeds at a strict
prompt). Verification covers *load* paths, not *read* paths.

A gated `severity_guess()` → `VE_MUST` floor would close the first. It overrides a
deliberate upstream tolerance ("files we expect to always be unverified"), so it
belongs behind the same opt-in, not in the default.

## [3] Files the mechanism does not require to exist (presence gate)

Verification only ever appraises a file that is *read*; absence is invisible.
Three layers of this:

- The whole appraisal block is guarded by `loaded_manifests` (`:542`): with no
  manifest loaded, it is skipped entirely.
- Nothing at boot walks the signed manifest to confirm its listed files are still
  present.
- `loader.conf` is not mandatory: a missing or failed one is skipped
  (`config.readConf`, silent parse) and the boot proceeds with defaults —
  bypassing any `loader.conf`-based control (`bootlock_password`, the menu).

Walkthrough (default `VE_MUST`): the manifest lists `loader.conf`; the attacker
deletes it (nothing prevents this) → defaults leave
`local_loader_conf_files="/boot/loader.conf.local"` in force → the attacker
creates that unlisted file → it is tolerated (`:581`) and read last, overriding.
Kernel and modules are explicit `VE_MUST` (`load_elf.c:328`, `module.c:715`), so
this does not load unsigned code, and GELI still needs its passphrase — the effect
is a boot-policy/bootlock bypass, not code execution.

A presence check — critical files must exist *and* verify, else halt or prompt —
would close it. A pinned `loader.conf` cannot: if it is the file that is deleted,
the permissive defaults return.

## [4] The root of the chain is unread (surface SecureBoot/SetupMode)

veriexec verifies everything it loads, but nothing under `stand/` checks the root
of its own chain: `SecureBoot` and `SetupMode` are never read. On a machine where
Secure Boot was switched off in setup, a loader built `WITH_LOADER_VERIEXEC` still
reports a fully verified boot. The predicate already exists —
`efi_secure_boot_enabled()` computes `SecureBoot == 1 && SetupMode == 0` — but its
result is used only internally and then discarded. Surfacing it would let a gate
notice the "Secure Boot disabled, loader intact" case (CMOS reset / evil maid).
@sgerraty flagged this half in #2271 and asked what should happen on a wrong
measurement; I'd rather build to your answer than guess.

---

Adjacent, not part of the above: a verification refusal surfaces as
`ENOENT`/"unknown error (80)" because `EAUTH` is unknown to `libsa`'s `strerror`
(reported separately); and `zfs_stat()` leaving `st_dev`/`st_ino` uninitialised —
which broke strict on ZFS `/boot` entirely — is fixed by #2271.

I'd gladly follow up on [2]–[4] in whatever form suits you — a PR, a bug, or a
list — and to hear which you'd even want. I'm a software developer and architect by
trade, just not deep in OS internals; parts of this were drafted with AI assistance
and then checked on the machine, so please tell me where it doesn't fit the tree's
conventions and I'll rework it. If there's testing I can help with, I'm glad to —
this machine reproduces the measurements in a few minutes.


bsdimp:
The long, detailed explanation is difficult to approach. The long discussion in the other PR does help. Both PRs suffer from not having a concise summary at the start (this much more so than the one it references).

"Move setting veriexec strict mode into C so we can compile in a stricter default." for example (though I know this adds an extra file early in the interpreter's initialization).

The analysis is somewhat flawed in its reasoning, or I'm not understanding something about veriexec: Don't lua and 4th have the same problem. 4th has an obscure thing that can be used, but lua doesn't have the extra obscure thing. Both are a hole in the system. loader.env is a clever way around it, but it itself isn't tamperproof.

I'm not sure why having an extra file that's examined is secure. If I'm an attacker, can't I just delete that file? Wouldn't it be better to just compile in the strict mode instead of having an extra layer of indirection?

Please help me understand...

sgerraty:
BTW glad you are finding this stuff useful.

On Wed, 19 Aug 2026 03:56:45 -0700, johannes-bruegmann writes:
>### Why the existing route cannot do it

FWIW among the tweaks I have not upstreamed is:

Index: stand/efi/loader/main.c
--- stand/efi/loader/main.c	2026-08-12 16:08:29.492681000 -0700
+++ stand/efi/loader/main.c	2026-08-12 17:09:26.560750000 -0700
@@ -1302,6 +988,8 @@
 #ifdef LOADER_VERIEXEC
 	/* tell boot_setenv to be careful */
 	set_check_restricted(true);
+	/* and start enforcing now */
+	ve_verifying_set();
 #endif
 

Fine for us as an embedded system, but not necessarily a good idea for
upstream.   Though perhaps another #ifdef around that like

#ifdef ENFORCE_LOADER_VERIEXEC

would make it suitable?

That might be better than hard coding /boot/loader.ve.strict
anywhere.

A knob to control whether VE_TWEAK_FILE is defined might be good.

Once you have enforcement triggered, the timing on other possible
loader.ve.* is generally not critical.  Eg. we have a package which adds
loader.ve.trust.revoke.jnxdev

that file's content is a trust anchor to be rejected by the loader.
Useful when you don't want a box to boot with dev signed packages.
It does not need to be processed first up as the result does not matter
until you start getting to kernel modules etc.

sgerraty:
On Wed, 19 Aug 2026 07:12:17 -0700, Warner Losh writes:
>"Move setting veriexec strict mode into C so we can compile in a stricter defa
>ult." for example (though I know this adds an extra file early in the interpre
>ter's initialization).

And per my earlier response, the neatest solution IMO to that is;

Index: stand/efi/loader/main.c
--- stand/efi/loader/main.c	2026-08-12 16:08:29.492681000 -0700
+++ stand/efi/loader/main.c	2026-08-12 17:09:26.560750000 -0700
@@ -1302,6 +988,8 @@
 #ifdef LOADER_VERIEXEC
 	/* tell boot_setenv to be careful */
 	set_check_restricted(true);
+	/* and start enforcing now */
+	ve_verifying_set();
 #endif

but for upstream it might be better as

 #ifdef LOADER_VERIEXEC
 	/* tell boot_setenv to be careful */
 	set_check_restricted(true);
 #ifdef LOADER_VERIEXEC_ENFORCE
	/* and start enforcing now */
	ve_verifying_set();
 #endif
 #endif

s/ENFORCE/ALWAYS/ or other better name.

Johannes's patch is doing something similar, but based on the presense
of a verified file "loader.ve.strict"
It is the act of verifying that path that communicates info to the
loader.

Unlike mac_veriexec which deals only with dev_t,inode_t,gen tuples and
does not know nor care what name is associated with them, the
loader implementation is pathname based - thus a filename like
"loader.ve.strict" can be "trusted" if verified.

But since I can simply remove /boot/loader.ve.strict
the call to ve_verifying_set() above is better.

>The analysis is somewhat flawed in its reasoning, or I'm not understanding som
>ething about veriexec: Don't lua and 4th have the same problem. 4th has an obs

For 4th we have loader.rc and *.4th on the list of things which must
always be verified.  We do not want *.conf on that list as there are
countless reasons for wanting to tweak a loader.conf file - we just
restrict what you can set via an unverified loader.conf.
Same goes for loader.env btw

If I understand Johannes correctly, LUA only has *.lua and loader.conf ?
and there's no convenient *must* be verified file that can be used to
trigger a read of "loader.ve.strict" (actually the open is all that
matters), before you get to loader.lua and if that is tampered with
the loader drops to OK prompt rather than panic.

Thus his solution was to do that open from the loader itself, before
loader.lua is read. 

>cure thing that can be used, but lua doesn't have the extra obscure thing. Bot
>h are a hole in the system. loader.env is a clever way around it, but it itsel
>f isn't tamperproof.

Nor do we/I want it to be tamperproof - for the same reasons as
loader.conf

>I'm not sure why having an extra file that's examined is secure. If I'm an att
>acker, can't I just delete that file? Wouldn't it be better to just compile in
> the strict mode instead of having an extra layer of indirection?

It isn't since it can be removed.  The only secure solution is to call
ve_verifying_set() from loader long before you read anything.
Good if you are an embedded vendor, not so much for a generic OS
distribution, the extra #ifdef though might help.

--sjg

johannes-bruegmann:
> The long, detailed explanation is difficult to approach. 

@bsdimp — thanks, this is fair. Leading with the summary you asked for:

**Move setting veriexec's strict mode into C, so a stricter default is compiled into
the signed loader and in force before any file is read — instead of relying on a
deletable `/boot/loader.ve.strict`.** That's the change; the long part is only the
"why". I've cut it down and put this summary at the top of the description.

On your main question — you're right, and it's exactly what this does now. The
strict default is compiled in; the examined file is **no longer the mechanism**. You
can delete it and strict is still on. Its only remaining job is to be a canary whose
absence we *notice*. So I fully agree it isn't secure as an activation mechanism —
that's why the decision moved into C. Keeping a soft "this should be here" was a
tamper-*detection* idea (notice the change, don't necessarily stop the boot), but
you're right it's the wrong place for the *switch*. Even the original,
error-handling-free version served that detection purpose for us — but compile-in is
better, and I'm glad to build it whatever way you'd want.

On lua vs 4th: you're right it isn't lua-specific. That's why the fix sits *below*
the interpreter — in `verify_file` / `efi_main` — so it holds for both lua and 4th.
On the "obscure thing" in 4th I'm not sure I follow you; could you point me at it, so
the writeup is accurate?

On `loader.env`: agreed, it isn't tamperproof — we treat it as a hole too. Locally
we compile it out entirely (we moved the preceding `#endif` to sit after the
`read_loader_env()` calls). I'd be glad to fold that into this PR if you'd want it.

Please tell me where this still doesn't fit, and I'll rework it.

johannes-bruegmann:
@sgerraty — you said the harder half, what should happen when a measurement comes
back wrong, is unsettled, and that's exactly the part I've been building. Rather
than describe it in the abstract, I've published it: 

  https://github.com/johannes-bruegmann/freebsd-src/tree/platform-trust-gates.

It's a small mechanism/policy layer — measure → claim → gate → policy → action, one
policy table per phase. It appraises the platform early (`SecureBoot`/`SetupMode`, a board
identity from SMBIOS, a boot marker in the boot entry's LoadOptions) and, before
the interpreter, the loader's own prerequisites. The split is deliberate — the
measuring never decides, the policy does — so "what happens on a wrong measurement"
is a *table*, not hardcoded. Report-only is a first-class policy, which keeps the
whole thing honest as tamper **detection**, not prevention.

Concretely, the answer we landed on for a wrong measurement:

- **Publish always.** Every gate exposes its verdict to `loader.trust.<gate>.*`,
  so a wrong measurement is at least always *visible*.
- **Soft guarantees are watched, not enforced.** A separate publish-only gate
  records whether strict is actually active and whether the `loader.ve.strict`
  marker is present — and never halts. That's the crux for us: we know the medium
  can be tampered, so we limit ourselves to *noticing*. It also keeps full
  compatibility with a stock boot (no gates compiled in → behaviour unchanged).
- **Enforcement is one config-driven layer up, in Lua.** The loader publishes;
  `password.lua` reads the verdicts and, per a `loader_trust_<gate>_require`
  policy, may demand a passphrase. A compiled-in C backstop covers the one case
  the Lua cannot — a missing or tampered `loader.conf`.

Honest caveat on the failure action: today it is enforcement-shaped (a wrong
passphrase halts). That reflects where our tooling is, not where we think it should
end up. The direction we are aiming at is softer — a prompt that merely *captures*
an answer and hands it to a later gate to adjudicate — but that presupposes gates
we cannot yet place in those later layers. So it is not fully thought through, and
your steer would be welcome.

I'm not attached to our shape. If gates like these are something you'd consider in
principle, I'd much rather build them the way you'd want — general enough that
anyone could use them — than carry a private fork that drifts from where you take
veriexec. Two things I'm genuinely unsure about:

1. Whether an early, *forced* activation belongs in the loader at all, or whether
   there's a pattern you'd prefer.
2. Your own open question, turned back: for the failure action, is a config-driven
   policy (report / prompt / passphrase-against-a-compiled-in-hash) the right
   shape, or were you picturing something else?

Would you be willing to look and point us in the right direction? Happy to reshape
it entirely.

johannes-bruegmann:
@sgerraty: Follow-up: policy layer + tooling on top of libsecureboot/veriexec

Hi Simon,

a follow-up from the loader ZFS veriexec corner (the st_dev/st_ino stat fix): I have since built a small policy layer on top of your libsecureboot/veriexec infrastructure — "platform trust gates" for the EFI loader. Providers measure the platform (Secure Boot state, PK/KEK/db digest, board identity, an NVRAM boot marker, file prerequisites via verify_file, load origin from LoadedImage), claims weigh them against compiled baselines, per-phase policies publish the appraisal to kenv. Strictly detection, not enforcement — the design assumes the medium can be tampered with and concentrates on noticing, with loader.ve.pcr as the soft aggregate alongside.

The tooling that compiles those decisions (and the veriexec manifests, signing, deployment) is now public:

https://github.com/enk-ode/elvboot

The loader-side patch series will follow once it is presentable; the generated foundation.c in the repo's docs shows the interface. If any of it is useful upstream — or if you see problems with how I lean on verify_file/ve.pcr — I would value your feedback.

Thanks for the infrastructure this all stands on,
Johannes

bsdimp (approved):
This looks good...

PR #2381, opened 2026-08-20
rm(1): guard against accidental removal of /* and /dev

ZhangQiyue2011:
This change adds an early argument check in rm to reject attempts to remove `/*`, `/dev`, `/*/`, or `/dev/`.
This mitigation only targets accidental mass deletion from patterns such as `rm -rf /*`; deliberate destructive operations are not blocked.

This guard only blocks removal of the top-level `/dev` and `/dev/` directory, rather than extending the blocklist to `/bin`, `/usr` or other system hierarchies, for two primary reasons:
1. `/dev` is a required fundamental directory on a conforming POSIX system; a running system cannot function without it.
2. There are essentially no legitimate, routine use cases to delete the entire top-level `/dev` directory tree on an active, booted FreeBSD instance.

By contrast, `/bin` and `/usr` do have valid scenarios for full removal, such as chroot construction, embedded system customization, and root filesystem refactoring. Blocking them would interfere with legitimate system-administration workflows.

Additionally, an argument list containing `/dev` is a very strong signal of an accidental mass deletion originating from patterns like `rm -rf /*`. In such a failure mode, we abort the entire operation immediately instead of filtering out only the dangerous path and continuing to delete remaining items. Broken scripts or catastrophic typos should not be partially tolerated.

Tested:
- `rm -ri /*` → blocked, prints error and aborts
- `rm -ri /*/` → blocked, prints error and aborts
- `rm -ri /dev` → blocked, prints error and aborts
- `rm -ri /dev/` → blocked, prints error and aborts
- Ordinary file/directory removal continues to work as before

### DIFF
--- a/bin/rm/rm.c
+++ b/bin/rm/rm.c
@@ -55,6 +55,7 @@ static int	check(const char *, const char *, struct stat *);
 static int	check2(char **);
 static void	checkdot(char **);
 static void	checkslash(char **);
+static void	checkhazard(char **);
 static void	rm_file(char **);
 static void	rm_tree(char **);
 static void siginfo(int __unused);
@@ -141,6 +142,7 @@ main(int argc, char *argv[])
 
 	checkdot(argv);
 	checkslash(argv);
+	checkhazard(argv);
 	uid = geteuid();
 
 	(void)signal(SIGINFO, siginfo);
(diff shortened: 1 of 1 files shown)


ZhangQiyue2011:
Ping reviewers.

This change adds early argument checks inside rm(1), defending against accidental removal of /*, /*/, /dev, and /dev/.
This mitigation only targets human‑error fat‑finger scenarios, and is not intended to block deliberate malicious operations.

The work originates from my own real‑world experience: I have actually executed a destructive "rm -rf /*" by mistake and damaged my system, which motivated this safety guard.
Limitations of this mitigation are noted in the PR description.

PR #2386, opened 2026-08-24
gpio: add Intel Elkhart Lake GPIO driver

ckoehne:
Add a GPIO driver for the Intel Elkhart Lake platform based on the generic intelgpio framework.

Elkhart Lake exposes each GPIO community as a separate ACPI device with _HID INTC1020. The community number is specified by _UID, which the driver uses to select the corresponding single-community configuration.

Add the pad group definitions for communities 0
through 5, together with the required module build infrastructure and files.x86 integration.

Most BIOS versions for this platform report the DSW community (UID 2) with _STA == 0, so it is hidden from normal probing.  Keep the UID 2 entry in the table anyway, so that the remaining communities keep their UID-based numbering and resource mapping.

### DIFF
--- a/sys/dev/gpio/intel/ehlgpio.c
+++ b/sys/dev/gpio/intel/ehlgpio.c
@@ -1,0 +1,175 @@
+/*-
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ * Copyright (c) 2026 Beckhoff Automation GmbH & Co. KG
+ */
+
+#include <sys/param.h>
+#include <sys/systm.h>
+#include <sys/bus.h>
+#include <sys/gpio.h>
+#include <sys/kernel.h>
+#include <sys/module.h>
+
+#include <contrib/dev/acpica/include/acpi.h>
+
+#include <dev/acpica/acpivar.h>
+#include <dev/gpio/gpiobusvar.h>
+
+#include "intelgpio.h"
+
+#include "gpio_if.h"
+#include "opt_acpi.h"
+
+static const struct intelgpio_padgroup ehl_com0_groups[] = {
+	{ .first_pad =  0, .npads = 26, .gpio_base =  0, .name = "GPP_B" },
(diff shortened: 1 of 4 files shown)


sys/dev/gpio/intel/ehlgpio.c
```diff
@@ -0,0 +1,175 @@
...
+
+#include "intelgpio.h"
+
+#include "gpio_if.h"
+#include "opt_acpi.h"
+
+static const struct intelgpio_padgroup ehl_com0_groups[] = {
+	{ .first_pad =  0, .npads = 26, .gpio_base =  0, .name = "GPP_B" },
+	{ .first_pad = 26, .npads = 16, .gpio_base = 26, .name = "GPP_T" },
+	{ .first_pad = 42, .npads = 25, .gpio_base = 42, .name = "GPP_G" },
```
VexedUXR: There are 24 pins here according to the datasheet.

```suggestion
	{ .first_pad = 42, .npads = 24, .gpio_base = 42, .name = "GPP_G" },
```

sys/dev/gpio/intel/ehlgpio.c
```diff
@@ -0,0 +1,175 @@
...
+		.desc = "Intel Elkhart Lake GPIO community 5",
+	}
+};
+
+static const struct intelgpio_platform *
+ehl_get_platform(device_t dev)
+{
+	ACPI_HANDLE handle;
+	ACPI_STATUS status;
+	int uid;
```
VexedUXR, line 125: `acpi_GetInteger` expects a pointer to a `UINT32`.

```suggestion
	UINT32 uid;
```

After that you can do away with the < 0 check too. :)

VexedUXR (changes requested):
I wonder why a lot of these pins are undocumented in the datasheet, but exist in the linux driver.

Anyways, I don't want to block this just because of that. Just one small thing, otherwise LGTM.

PR #2387, opened 2026-08-25
ports.7: Document test target

patmaddox:
While here, document WITHOUT_DEBUG_PORTS.

### DIFF
--- a/share/man/man7/ports.7
+++ b/share/man/man7/ports.7
@@ -25,7 +25,7 @@
 .\" (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
 .\" THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 .\"
-.Dd February 24, 2026
+.Dd August 25, 2026
 .Dt PORTS 7
 .Os
 .Sh NAME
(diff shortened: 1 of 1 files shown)


share/man/man7/ports.7
```diff
@@ -294,6 +294,8 @@ Remove an installed port from the system, similar to
 Remove all installed ports with the same
 .Va PKGORIGIN
 from the system.
+.It Cm test
+Run the port's tests, if it has any.
```
kevans91:
```suggestion
Run the port's tests, if it has defined a
.Dv TEST_TARGET
value or a
.Cm do-test
target.
```

kevans91:
I don't think we need to define how the testing part of the ports framework works in entirety, but I think providing cookies like `TEST_TARGET` and `do-test` would be a helpful pointer to provide here.

share/man/man7/ports.7
arrowd: Yes, I agree.

ngie-eign:
This seems like a very reasonable change to admit once @kevans91 's feedback has been addressed.
One thing I'm curious about -- does it make sense to distribute a separate package tied to the ports tree that handles this documentation? That way the docs live with the sources they document. Just thinking out loud..

patmaddox:
thanks for the feedback, updated

concussious (approved):

kevans91 (approved):
Carry forth and wreak havoc unto the tree, good sir.

ngie-eign (approved):

PR #2392, opened 2026-08-27
hwpmc/amd: add UMC PMC counter support

ojanerif:
## Summary

AMD's Unified Memory Controller (UMC) exposes per-node performance
counters accessible via MSRs starting at 0xC0010800 (evsel at BASE +
2*i, perfctr at BASE + 2*i + 1). The counter count and node topology
are reported by CPUID leaf 0x80000022: EBX[23:16] gives the total PMC
count, ECX is a bitmask of Data Fabric nodes present.

UMC counters differ from core/L3/DF counters in two important ways:

- The enable bit is at bit 31 (0x80000000) rather than the standard
  position used by other AMD counter types.
- UMC counters do not trigger overflow NMIs, so the stop path needs no
  overflow-wait loop.

Counters are registered as `K8-UMC<node>-<i>`, where the node index is
derived from `popcntq(ECX)` to identify which Data Fabric node (and
thus which memory controller) a counter belongs to. On single-node
systems the naming is `K8-UMC0-0` .. `K8-UMC0-N`.

UMC MSRs are not addressable via RDPMC, so `amd_get_msr()` returns
EINVAL for the UMC subclass rather than falling through to the DF
branch.

This is the kernel-side change. A companion libpmc change (userland
event mapping and allocator) will follow as a separate PR.

## Test plan

Tested on EPYC 9654 (Zen 4, `AuthenticAMD-25-11-1`):

- [x] CPUID 0x80000022 reports 4 UMC PMCs across 1 node
- [x] Counters register as `K8-UMC0-{0..3}` (visible in dmesg as `K8/32`: 6 core + 6 L3 + 16 DF + 4 UMC)
- [x] 4 simultaneous system-wide counting PMCs allocate successfully
- [x] A 5th allocation correctly returns EINVAL
- [x] `buildkernel GENERIC` completes cleanly with `-Werror`

Sponsored by:	AMD

### DIFF
--- a/sys/dev/hwpmc/hwpmc_amd.c
+++ b/sys/dev/hwpmc/hwpmc_amd.c
@@ -64,7 +64,7 @@ struct amd_descr {
 };
 
 static int amd_npmcs;
-static int amd_core_npmcs, amd_l3_npmcs, amd_df_npmcs;
+static int amd_core_npmcs, amd_l3_npmcs, amd_df_npmcs, amd_umc_npmcs;
 static struct amd_descr amd_pmcdesc[AMD_NPMCS_MAX];
 struct amd_event_code_map {
 	enum pmc_event	pe_ev;	 /* enum value */
@@ -190,6 +190,8 @@ static uint64_t amd_df_allowed_mask;
 static uint64_t amd_core_extra_mask;
 static uint64_t amd_l3_extra_mask;
 static uint64_t amd_df_extra_mask;
+static uint64_t amd_umc_allowed_mask;
+static uint64_t amd_umc_extra_mask;
 
 SYSCTL_DECL(_kern_hwpmc);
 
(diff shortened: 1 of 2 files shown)


mashtizadeh:
This is basically the same as my diff I sent you I thought you were going to add the logic to check the DF nodes.

ojanerif:
Closing in favor of #2368 which covers the same kernel-side UMC counter support. A follow-up PR will address the CPU topology mask construction (per Data Fabric node CPU ID bits) needed to make the counter allocation interface usable for multiplexing.

mashtizadeh:
I thought you were going to do the follow up diffs we talked about:

1. Add a get cpu mask call to pmc that is more general than the current flags since DF nodes don't match anything else.
2. Add DF node support on top of the patch

PR #2395, opened 2026-08-29
localtime compatibility with upstream

eggert:
The two patches in this pull request adjust contrib/tzcode/localtime.c so that it is identical to upstream (as of the current TZDB release, 2026c). This should improve coordination of FreeBSD's localtime implementation with upstream. TZDB 2026d (not yet released) is planned to have some patches will make localtime work better in the presence of adversarial TZif files, and when that happens the goal is for the 2026d localtime.c to drop in unchanged.

### DIFF
--- a/contrib/tzcode/localtime.c
+++ b/contrib/tzcode/localtime.c
@@ -1987,23 +1987,6 @@ tzset(void)
 }
 #endif
 
-#ifdef __FreeBSD__
-void
-freebsd13_tzsetwall(void)
-{
-  monotime_t now = get_monotonic_time();
-  int err = lock();
-  if (0 < err) {
-    errno = err;
-    return;
-  }
-  tzset_unlocked(!err, true, now);
-  unlock(!err);
-}
-__sym_compat(tzsetwall, freebsd13_tzsetwall, FBSD_1.0);
-__warn_references(tzsetwall,
-    "warning: tzsetwall() is deprecated, use tzset() instead.");
-#endif /* __FreeBSD__ */
 static void
 gmtcheck1(void)
 {
--- a/contrib/tzcode/localtime.c
+++ b/contrib/tzcode/localtime.c
@@ -929,9 +929,6 @@ tzloadbody(char const *name, struct state *sp, char tzloadflags,
 		name = TZDEFAULT;
 		if (! name)
 		  return EINVAL;
-#ifdef __FreeBSD__
-		tzloadflags &= ~TZLOAD_FROMENV;
-#endif /* __FreeBSD__ */
 	}
 
 	if (name[0] == ':')
--- a/lib/libc/stdtime/Makefile.inc
+++ b/lib/libc/stdtime/Makefile.inc
@@ -2,7 +2,7 @@
 
 .PATH:	${LIBC_SRCTOP}/stdtime ${SRCTOP}/contrib/tzcode
 
-TZCODE_SRCS=	asctime.c difftime.c localtime.c
+TZCODE_SRCS=	asctime.c difftime.c localtime-compat.c
 STDTIME_SRCS=	strftime.c strptime.c timelocal.c
 SRCS+=		${TZCODE_SRCS} ${STDTIME_SRCS} time32.c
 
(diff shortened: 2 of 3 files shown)


contrib/tzcode/localtime.c
```diff
@@ -929,9 +929,6 @@ tzloadbody(char const *name, struct state *sp, char tzloadflags,
 		name = TZDEFAULT;
 		if (! name)
 		  return EINVAL;
-#ifdef __FreeBSD__
-		tzloadflags &= ~TZLOAD_FROMENV;
```
dag-erling, line 933: This code is here for a reason, as I've tried unsuccessfully to explain to you in the past.

dag-erling (changes requested):
This appears to be trying to solve a problem that does not exist.

contrib/tzcode/localtime.c
eggert: It formerly was there for a reason, but as far as I can see that reason no longer exists due to other changes in localtime.c. If the reason is still there, could you please re-explain the need for `tzloadflags &= ~TZLOAD_FROMENV;` in its current context, or point me to your old explanation?

dag-erling: I thought I'd explained it in https://bugs.freebsd.org/290520 but I see that I didn't, because I don't have a test that demonstrates the problem. In this part of the code, we set `name` to the default value if it is `NULL` (which it will always be on the first call to `tzset()` if `TZ` is unset, so not at all an obscure edge case). If `tzloadflags` has the `TZLOAD_FROMENV` bit set at this point, `name` will be treated as tainted, which is incorrect. The remaining question is: what are the consequences of treating `name` as tainted when it isn't? Right now the only reference to `TZLOAD_FROMENV` is in the `tzfile_changed()` logic, which doesn't seem right...

eggert: I don't see a problem in the `tzfile_changed` logic dealing with `TZLOAD_FROMENV`. When `tz_change_interval` is nonnegative (as in FreeBSD), a file specified directly or indirectly via the `TZ` environment variable should be checked for updates periodically. This check should be done regardless of whether it's an explicit environment setting like `TZ="America/Los_Angeles"` or an implicit setting like an unset `TZ` that defaults to `TZDEFAULT`, because either way the corresponding TZif file could have been updated.

When `tzset_unlocked` sees that `tz_change_interval` is positive it therefore, regardless of whether `TZ` is set, uses `fresh_tzdata` to decide whether to revalidate the cache . `tzset_unlocked` uses `tzloadbody` (via `zoneinit` and `tzloadbody`) with `TZLOAD_FROMENV` to do the revalidation. Like `tzset_unlocked`, `tzloadbody` should revalidate the cache regardless of whether `TZ` is set. With the FreeBSD-specific patch, though, during this revalidation when `TZ` is unset `tzloadbody` assumes that the TZif file has changed even in the usual case when it has not, and `tzloadbody` therefore goes to extra work to reread and reanalyze the unchanged TZif file. This is a performance issue (not a correctness issue), as re-reading the file will produce the same data as before. Possibly your tests look for correctness not performance, and so don't see the performance issue.

The `TZLOAD_FROMENV` flag is supposed to be absent only for calls derived from `tzalloc`, a NetBSD function that FreeBSD does not support. `tzalloc` does not need the `tz_change_interval` mechanism and by omitting `TZLOAD_FROMENV` it can skip the mechanism for speed.

If, despite this explanation, you still see a need for the FreeBSD-specific patch, it'd be helpful to know why it's needed, as I expect the reason would apply to other platforms, not just to FreeBSD.

dag-erling: So `TZLOAD_FROMENV` no longer means what it used to. It was supposed to indicate that `name` was user-provided and potentially unsafe, and therefore subject to additional checks. I clearly wasn't paying close enough attention when I imported 2026a.

eggert: Thanks for pointing out the lack of clarity. The intent in tzcode localtime.c is that `TZLOAD_FROMENV` means "derived from the environment", not "potentially unsafe". I [patched tzcode's localtime.c](https://github.com/eggert/tz/commit/e3d17ef30e697d62487edf74597f2ce10db92103) to try to make its intent clearer.

Can I do anything more to help move this along? I don't know what the [GitHub pull request page](https://github.com/freebsd/freebsd-src/pull/2395) means when it says "Changes reviewed / No applicable reviews submitted by reviewers with write access / 1 requested change / dag-erling Requested changes with read-only permissions". tzcode 2026d should be released in the next few days, so one possibility is to wait until then before merging.

dag-erling: I struggle to see what problem this PR is intended to solve.

eggert: The PR is intended to solve the following issues:
1. FreeBSD `localtime` and related functions can unnecessarily reread and reanalyze TZif files specified by the absence of the `TZ` environment variable even when the files have not changed, thus hurting performance when this happens.
2. FreeBSD `localtime` contains an unnecessary machine instruction to AND out `TZLOAD_FROMENV`. This is a trivial performance issue; the more important thing is that it unnecessarily makes the code more confusing than it already is.
3. FreeBSD localtime.c can easily be made byte-for-byte identical to upstream localtime.c, thus simplifying future maintenance.

If you don't mind these performance-loss, confusion, and future-maintenance-hassle issues, please feel free to ignore the PR. However, in that case I would suggest adding a comment about (1) and (2) to the FreeBSD copy of localtime.c to help clarify matters there, as it really is puzzling as to why FreeBSD goes to some effort to make performance worse.

PR #2398, opened 2026-09-01
sys/conf/options: Add WITNESS_LOCK_CHILDCOUNT

mrkajetanp:
Make the witness LOCK_CHILDCOUNT a configurable kernel option. On machines with a very high core count the default value is too low, leading to witness exhaustion after boot.

### DIFF
--- a/share/man/man4/witness.4
+++ b/share/man/man4/witness.4
@@ -21,7 +21,7 @@
 .\" OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 .\" SUCH DAMAGE.
 .\"
-.Dd January 26, 2026
+.Dd September 1, 2026
 .Dt WITNESS 4
 .Os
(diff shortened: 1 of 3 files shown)


concussious (approved):
Looks good from docs side, requesting some kernel experts for further review.

kostikbel (approved):

concussious:
Thanks for sending this in!

mrkajetanp:
> Thanks for sending this in!

No problem, thanks!

emaste:
Having this option is better than not, but we really need to make this auto-tune and work well with the compiled-in default MAXCPU (1024).

mrkajetanp:
> Having this option is better than not, but we really need to make this auto-tune and work well with the compiled-in default MAXCPU (1024).

No doubt, the machine I'm using has `nproc == 124` and with the default value the witness exhausts as soon as it boots. Let alone anything higher.

PR #2401, opened 2026-09-02
geom: Change the wait channel message for speedups to something unique

mashtizadeh:
Change the wait channel message for g_io_speedup to something unique so we can tell it apart from flush.

Sponsored by: Netflix


mashtizadeh:
@bsdimp

PR #2402, opened 2026-09-02
cam: Change ccb spare field for use with the I/O scheduler

mashtizadeh:
etime is a spare ccb field that is not in use, this changes the field into a uintptr_t and renames it for use by the I/O scheduler.  There should be no change to the ccb structure size.

Sponsored by: Netflix


mashtizadeh:
@bsdimp

PR #2403, opened 2026-09-02
cam: Queue depth statistics

mashtizadeh:
Collect distribution of queue depths and average latencies per queue depth inside the I/O scheduler.

Sponsored by: Netflix


mashtizadeh:
@bsdimp

PR #2405, opened 2026-09-03
bhyve: fix build with WITHOUT_CAPSICUM

peripherium:
This PR fixes a build failure when `bhyve` is compiled with `WITHOUT_CAPSICUM`.

`net_backend_netgraph.c` uses and several definitions from `fcntl.h` and `sys/param.h`, but the required headers are only indirectly available when Capsicum is enabled.

The patch includes <fcntl.h> and <sys/param.h> when WITHOUT_CAPSICUM is defined.

### DIFF
--- a/usr.sbin/bhyve/net_backend_netgraph.c
+++ b/usr.sbin/bhyve/net_backend_netgraph.c
@@ -28,12 +28,14 @@
 #ifndef WITHOUT_CAPSICUM
 #include <sys/capsicum.h>
 #endif
+#include <sys/param.h>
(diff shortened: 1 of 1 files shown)


emaste:
If we're relying on _fcntl.h_ and _sys/param.h_ (and it works today because they happen to be transitively included via _capsicum.h_) we should just include them unconditionally.

peripherium:
Makes sense. I'll include both headers unconditionally. I had originally added them to the #else branch since the build failure only occurs with WITHOUT_CAPSICUM.

PR #2408, opened 2026-09-04
ixl: fix big-endian hardware interface handling

mchoo7:
Correct the byte order at the driver interfaces that do not operate in
host order. Encode little-endian Admin Queue VSI and MAC/VLAN fields,
and decode firmware-provided queue handles, statistics indices, and
event parameters before using them.

Decode transmit descriptor writeback before examining it, and construct
VLAN descriptor fields in host order for the final descriptor
conversion. Store paged HMC descriptors in little-endian form and retain
the selected field bits when reading HMC contexts.

Keep MMIO values in host order for bus-space accessors and remove the
obsolete le16_to_cpu no-op macro. In particular, retain the atomic
bus_space_read_8() used for 64-bit registers.

On big-endian systems, the unconverted perfect-match filter flag is
presented to firmware as 0x0100 instead of 0x0001. Firmware rejects that
command with EINVAL, leaving receive traffic functional only in
promiscuous mode.

MFC after:      2 weeks
MFC to: stable/14, stable/15
Sponsored by:   FreeBSD Foundation, Reliable Computer Systems Lab

Tested on: FreeBSD 15.1 ppc64

### DIFF
--- a/sys/dev/ixl/i40e_osdep.h
+++ b/sys/dev/ixl/i40e_osdep.h
@@ -133,8 +133,6 @@ struct i40e_spinlock {
         struct mtx mutex;
 };
 
-#define le16_to_cpu 
-
 #if defined(__amd64__) || defined(i386)
 static __inline
 void prefetch(void *x)
--- a/sys/dev/ixl/ixl_pf_iflib.c
+++ b/sys/dev/ixl/ixl_pf_iflib.c
@@ -463,14 +463,15 @@ ixl_initialize_vsi(struct ixl_vsi *vsi)
 	    ctxt.uplink_seid, ctxt.vsi_number,
 	    ctxt.vsis_allocated, ctxt.vsis_unallocated,
 	    ctxt.flags, ctxt.pf_num, ctxt.vf_num,
-	    ctxt.info.stat_counter_idx, ctxt.info.up_enable_bits);
+	    le16toh(ctxt.info.stat_counter_idx), ctxt.info.up_enable_bits);
 	/*
 	** Set the queue and traffic class bits
 	**  - when multiple traffic classes are supported
 	**    this will need to be more robust.
 	*/
-	ctxt.info.valid_sections = I40E_AQ_VSI_PROP_QUEUE_MAP_VALID;
-	ctxt.info.mapping_flags |= I40E_AQ_VSI_QUE_MAP_CONTIG;
+	ctxt.info.valid_sections =
+	    htole16(I40E_AQ_VSI_PROP_QUEUE_MAP_VALID);
+	ctxt.info.mapping_flags = htole16(I40E_AQ_VSI_QUE_MAP_CONTIG);
 	/* In contig mode, que_mapping[0] is first queue index used by this VSI */
 	ctxt.info.queue_mapping[0] = 0;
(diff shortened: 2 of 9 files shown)


sys/dev/ixl/i40e_osdep.h
```diff
@@ -223,10 +221,20 @@ ixl_flush_osdep(struct i40e_osdep *osdep)
...
-   bus_space_read_8( ((struct i40e_osdep *)(a)->back)->mem_bus_space_tag, \
-                     ((struct i40e_osdep *)(a)->back)->mem_bus_space_handle, \
-                     reg))
+static __inline uint64_t
+rd64_osdep(struct i40e_osdep *osdep, uint32_t reg)
+{
+
+#if _BYTE_ORDER == _BIG_ENDIAN
+	return ((uint64_t)rd32_osdep(osdep, reg) |
+	    ((uint64_t)rd32_osdep(osdep, reg + 4) << 32));
```
kev009: This splits a 64bit atomic read into 2 torn 32b reads and needs to be fixed.

Why is the existing bus_space_read_8() insufficient here? PowerPC PCI BARs use bs_le_tag, whose 64-bit accessor performs a single volatile 64-bit read followed by le64toh(). That should provide both the correct register ordering and the atomic access required by the X710 datasheet.  Unless there is evidence that this platform cannot issue the 64-bit MMIO access, I think the original code should remain. If splitting is required, it needs a very careful implementation.

kev009 (changes requested):
The majority of this looks fine but the rd64_osdep change is a hard blocker.

sys/dev/ixl/i40e_osdep.h
mchoo7: I reverted this.

mchoo7:
I'll ping you when I'm done with testing although it should be okay since the only change is reverting the one mentioned above.

sys/dev/ixl/ixl_pf_iflib.c
```diff
@@ -463,14 +463,15 @@ ixl_initialize_vsi(struct ixl_vsi *vsi)
...
 	/*
 	** Set the queue and traffic class bits
 	**  - when multiple traffic classes are supported
 	**    this will need to be more robust.
 	*/
-	ctxt.info.valid_sections = I40E_AQ_VSI_PROP_QUEUE_MAP_VALID;
-	ctxt.info.mapping_flags |= I40E_AQ_VSI_QUE_MAP_CONTIG;
+	ctxt.info.valid_sections =
+	    htole16(I40E_AQ_VSI_PROP_QUEUE_MAP_VALID);
+	ctxt.info.mapping_flags = htole16(I40E_AQ_VSI_QUE_MAP_CONTIG);
```
kev009, line 474: This = is a behavior change but I think it is technically correct and an improvement.

kev009 (approved):
This looks fine now

mchoo7:
Works on ppc64. Added Reliable Computer Systems Lab to `Sponsored by:` for providing ppc64 machine for building and testing.

ping @bsdjhb for mentor approval

kev009:
Tested by: kbowling (X710 on amd64)

PR #2409, opened 2026-09-06
evdev: map ChromeOS "Vivaldi" top-row action keys

kk1987:
Chromebook keyboards driven by the ChromeOS EC emit their top-row action keys as scancode set 1 codes `0xE0 0x11` - `0xE0 0x1E`. Eleven of those are `NONE` in `evdev_scancode2key()`'s e0 table, so on FreeBSD the keys emit no evdev event at all — brightness, keyboard backlight, mic mute, screenshot, fullscreen, overview and the privacy-screen toggle are simply dead, and userspace remappers such as `keyd` have nothing to bind to.

This fills in those eleven entries. The codes and their meanings are taken from the codeset 1 comments on the `SCANCODE_*` enum in chrome-ec `include/keyboard_8042_sharedlib.h`.

Only entries that were previously `NONE` are touched. The EC's remaining top-row codes (`e0 10` prev track, `e0 19` next track, `e0 20` mute, `e0 2e`/`e0 30` volume, `e0 67` refresh, `e0 69` forward, `e0 6a` back) already had mappings, as did `e0 1c` keypad enter and `e0 1d` right control; none of those change. The table still has exactly 256 entries.

Tested on a Lenovo ThinkPad C14 Chromebook (Google `primus` board, ChromeOS EC, MrChromebox UEFI firmware) running FreeBSD 15.1: every top-row key now reports the expected code under `evemu-record`.


wulf7:
Comitted. Thanks!

PR #2414, opened 2026-09-07
lib/msun: use bit operations for round and add C23 roundeven

sriramshastry:
Replace roundf()'s floorf()-based implementation with direct manipulation of the float representation for finite inputs.

Handle 1 <= |x| < 2^23 by adding the rounding bit and clearing the fractional mask.  Handle |x| < 1 by returning signed zero or signed one directly, and return finite values with |x| >= 2^23 unchanged since they are already integral.  NaNs and infinities continue to be propagated with x + x.

This avoids calls to floorf() on the finite-input path while preserving the documented round(3) value semantics.


kargl:
Do you have bit twiddling versions for `round()` and `roundl()`?  One reason the round[fl].c codes uses `floor[fl]()` is that it's the same algorithm  for all precisions.   If a bug/issue occurs in one precision, then it is easy to check and fix the others.

sriramshastry:
> Do you have bit twiddling versions for `round()` and `roundl()`? One reason the round[fl].c codes uses `floor[fl]()` is that it's the same algorithm for all precisions. If a bug/issue occurs in one precision, then it is easy to check and fix the others.

Yes. I have added corresponding bit-twiddling implementations for `round()` and `roundl()` and a shared `fenv` regression test covering all three precision variants.

kargl:
Thanks for following up with `round()` and `roundl()`.  I'll try to carve out some time over next few days to review the patch.  I do not have commit privilege to the source tree, so you will need to find a committer after a review is completed.

Also, note, someone submitted a bugzilla report about C23's `roundeven[fl]()`.  I suspect your patch can be modified to cover those functions as well.

sriramshastry:
> Thanks for following up with `round()` and `roundl()`. I'll try to carve out some time over next few days to review the patch. I do not have commit privilege to the source tree, so you will need to find a committer after a review is completed.
> 
> Also, note, someone submitted a bugzilla report about C23's `roundeven[fl]()`. I suspect your patch can be modified to cover those functions as well.

Thanks for the pointer. I found Bugzilla 298503 and updated the series to implement the `C23 roundeven(), roundevenf()`, and `roundevenl()` functions.

PR #2415, opened 2026-09-08
hwpmc/amd: replace static pmcdesc[] and pc_amdpmcs[] with dynamic allocation

ojanerif:
## Summary

The AMD hwpmc driver sizes its descriptor tables from `AMD_NPMCS_MAX`, a compile-time upper bound that assumes the maximum counter count of every class simultaneously: 342 entries (16 core + 6 L3 + 64 DF + 256 UMC). Real hardware uses a small fraction of this. On a Zen 4 EPYC 9654 with 6 core, 6 L3, 16 DF, and 4 UMC counters, only 32 descriptors are live — the static `amd_pmcdesc[AMD_NPMCS_MAX]` and per-CPU `pc_amdpmcs[AMD_NPMCS_MAX]` arrays over-allocate by roughly 10x, and the per-CPU cost scales with CPU count.

This makes both tables dynamic, sized to the counter count the CPU actually reports via CPUID.

This follows the UMC counter support added in #2368; the dynamic allocation is necessary groundwork for counter multiplexing on UMC-equipped systems.

## Details

- `amd_pmcdesc` and per-CPU `pc_amdpmcs` become `mallocarray()` allocations sized to the registered PMC count (`amd_pmcdesc` in `pmc_amd_initialize()`, `pc_amdpmcs` in `amd_pcpu_init()`, freed in `amd_pcpu_fini()`). All indexed accesses in the class methods and the NMI handler already iterate to `amd_npmcs` and are bounds-checked against it; no loop iterates to `AMD_NPMCS_MAX`.

- `amd_l3_npmcs` and `amd_df_npmcs` are initialized to their defaults (6 and 4) before the `AMDID2_PTSCEL2I` / `AMDID2_PNXC` feature checks. They are now zeroed when those feature bits are absent, so the allocation size equals the number of descriptors actually registered on every hardware configuration. Without this, an INVARIANTS kernel on a CPU lacking L3 or DF counters would panic during hwpmc initialization.

- `amd_umc_npmcs` comes from CPUID Fn8000_0022h ECX[23:16] (NumUMCCounters) and is zero when the leaf is absent; there is no CPUID feature flag for UMC presence, so the count itself encodes absence.

- A file-scope `CTASSERT` enforces that the core default counts are positive. Runtime `KASSERT`s verify `npmcs_total` lies in [`amd_core_npmcs`, `AMD_NPMCS_MAX`] before allocation, that the write cursor is at the expected offset before each registration loop, and that `amd_npmcs == npmcs_total` after all loops.

- The initialization error path and `pmc_amd_finalize()` now free `amd_pmcdesc`; the error path also resets `amd_npmcs`.

## Open question for reviewers

The upper-bound sanity check on the CPUID-reported count is currently a `KASSERT(npmcs_total <= AMD_NPMCS_MAX)`, i.e. it panics on an INVARIANTS kernel and is a no-op otherwise. If a future microcode or emulator were to report an implausible counter count, an alternative would be to clamp `npmcs_total` to `AMD_NPMCS_MAX` and emit a `printf` warning instead of asserting. I kept the `KASSERT` since an out-of-range CPUID result is a "should never happen" condition, but I am happy to switch to clamp-and-warn if that is preferred.

## Test plan

- [x] amd64 `HWPMC_DBG` kernel builds clean
- [x] Booted and exercised on AMD EPYC 9654 (Zen 4, Family 19h Model 11h, 192 threads)
- [x] Full hwpmc ATF suite (IBS / UMC / core / L3 / DF / TSC): 0 failures

Sponsored by: AMD


ojaneri:
This addresses the sizing concern @mhorne raised on #2368: both amd_pmcdesc[AMD_NPMCS_MAX] and the per-CPU pc_amdpmcs[AMD_NPMCS_MAX] are now sized to the actual registered PMC count instead of the worst-case 342-entry bound, so adding UMC counters (256 of the 342 slots) no longer bloats every CPU's allocation on hardware that doesn't populate them.

@emaste also floated a greppable marker (XXXBLOAT) on that thread for this class of "sized by a maximum rather than the real count" allocation — I didn't add one here since this PR removes the bloat rather than just flagging it, but happy to tag the KASSERT(npmcs_total <= AMD_NPMCS_MAX) upper-bound check with it if that convention is still wanted for future greps.

@mashtizadeh this is the initial PR laying the dynamic-allocation groundwork for the counter multiplexing work you mentioned on #2368. Once this gets reviewed, I'll follow up with the next round of code building on top of it.

sys/dev/hwpmc/hwpmc_amd.c
```diff
@@ -1201,7 +1268,17 @@ pmc_amd_initialize(void)
 	return (pmc_mdep);
 
 error:
+	/*
+	 * amd_pcpu is allocated before pmc_mdep_alloc(); pc_amdpmcs arrays
+	 * are allocated in amd_pcpu_init(), not yet called at this point.
+	 * free(NULL) is a no-op for amd_pcpu and amd_pmcdesc.
+	 */
```
mhorne:
```suggestion
```

sys/dev/hwpmc/hwpmc_amd.c
```diff
@@ -1046,6 +1061,35 @@ pmc_amd_initialize(void)
...
+	    ((amd_feature2 & AMDID2_PTSCEL2I) != 0 ? amd_l3_npmcs : 0) +
+	    ((amd_feature2 & AMDID2_PNXC)     != 0 ? amd_df_npmcs : 0) +
+	    amd_umc_npmcs;
+	KASSERT(npmcs_total >= amd_core_npmcs,
+	    ("%s: npmcs_total=%d below amd_core_npmcs=%d; unexpected CPUID result",
+	    __func__, npmcs_total, amd_core_npmcs));
+	KASSERT(npmcs_total <= AMD_NPMCS_MAX,
+	    ("%s: npmcs_total %d exceeds AMD_NPMCS_MAX %d",
+	    __func__, npmcs_total, AMD_NPMCS_MAX));
+	/* M_WAITOK: called at driver init, not from interrupt context. */
```
mhorne:
```suggestion
```

sys/dev/hwpmc/hwpmc_amd.c
```diff
@@ -1118,11 +1168,28 @@ pmc_amd_initialize(void)
...
+	 * Verify that the descriptor table is exactly right-sized.
+	 * npmcs_total was computed from the same per-class counts before
+	 * the registration loops; amd_npmcs is their running sum after.
+	 * If they diverge, a new counter family was added without updating
+	 * npmcs_total, and amd_pmcdesc would be undersized.
+	 */
+	/*
+	 * To add a new counter family: (1) add its count to npmcs_total above,
+	 * (2) add a registration loop here, (3) increment amd_npmcs.
+	 */
```
mhorne:
```suggestion
	/*
	 * Verify that the descriptor table is exactly right-sized.
	 * If the counts diverge, a new counter family was added without updating
	 * npmcs_total, and amd_pmcdesc would be undersized.
	 */
```

sys/dev/hwpmc/hwpmc_amd.c
```diff
@@ -1046,6 +1061,35 @@ pmc_amd_initialize(void)
 		}
 	}
 
+	/*
+	 * Allocate the descriptor table sized to the actual PMC count reported
+	 * by CPUID, replacing the static amd_pmcdesc[AMD_NPMCS_MAX].  We sum the
```
mhorne: I don't think it is useful for the new comment to reference the old version of the code.

sys/dev/hwpmc/hwpmc_amd.c
```diff
@@ -178,7 +185,11 @@ const int amd_event_codes_size = nitems(amd_event_codes);
  * Per-processor information
  */
 struct amd_cpu {
-	struct pmc_hw	pc_amdpmcs[AMD_NPMCS_MAX];
+	/*
+	 * Allocated in amd_pcpu_init() with amd_npmcs entries;
+	 * freed in amd_pcpu_fini().
+	 */
```
mhorne:
```suggestion
```

sys/dev/hwpmc/hwpmc_amd.c
```diff
@@ -65,7 +65,14 @@ struct amd_descr {
 
 static int amd_npmcs;
 static int amd_core_npmcs, amd_l3_npmcs, amd_df_npmcs, amd_umc_npmcs;
-static struct amd_descr amd_pmcdesc[AMD_NPMCS_MAX];
+/*
+ * Allocated in pmc_amd_initialize(); freed on its error path or in
+ * pmc_amd_finalize().  Size == npmcs_total at initialization time.
+ */
```
mhorne:
```suggestion
```
It is well explained within `pmc_amd_initialize()`.

mhorne:
Thank you for handling this!

The implementation looks fine. Please trim the LLM-generated (?) commentary, as it is too verbose in some cases. I made a couple suggestions.

sys/dev/hwpmc/hwpmc_amd.c
```diff
@@ -1091,6 +1135,9 @@ pmc_amd_initialize(void)
 
 	if ((amd_feature2 & AMDID2_PNXC) != 0) {
 		/* Enable the data fabric counters */
+		KASSERT(amd_npmcs == amd_core_npmcs +
```
mashtizadeh: A couple of these asserts aren’t really doing anything probably you can remove.

sys/dev/hwpmc/hwpmc_amd.c
```diff
@@ -1046,6 +1061,35 @@ pmc_amd_initialize(void)
...
+	 * it is zero when the leaf is absent or the field is zero.  No CPUID
+	 * feature flag is defined for UMC presence, so the count itself encodes
+	 * absence (0 means no UMC counters).
+	 * Refer to AMD64 APM Vol.3, CPUID Specification (Fn8000_0001h/Fn8000_0022h).
+	 */
+	npmcs_total = amd_core_npmcs +
+	    ((amd_feature2 & AMDID2_PTSCEL2I) != 0 ? amd_l3_npmcs : 0) +
+	    ((amd_feature2 & AMDID2_PNXC)     != 0 ? amd_df_npmcs : 0) +
+	    amd_umc_npmcs;
+	KASSERT(npmcs_total >= amd_core_npmcs,
```
mashtizadeh: Isn’t this trivially true or are you worried about sign extension

mashtizadeh (approved):
Looks good, you can remove some of those trivially true asserts.

emaste:
> I didn't add [a greppable marker] here since this PR removes the bloat

Yes, that makes sense. My suggestion was really only for the case that we've identified the bloat without a specific plan to address it.

ojanerif:
Thanks for the feedback. I added those comments mostly to document the intent for myself while writing the code, but reading it back now the code really does say all of that already. Trimmed accordingly: removed the verbose blocks, dropped the trivially-true `npmcs_total >= amd_core_npmcs` KASSERT and the DF cursor KASSERT as suggested. Kept only the UMC note since there is no CPUID feature flag for UMC presence and that is genuinely non-obvious, and the final `amd_npmcs == npmcs_total` assert since it catches future divergence when a new counter family is added.

GuestAUser (changes requested):
Hi Osvaldo,

Following up on my review of [ojanerif/freebsd-src#3](https://github.com/ojanerif/freebsd-src/pull/3), I reviewed and tested [ojanerif/freebsd-src#4](https://github.com/ojanerif/freebsd-src/pull/4) at `686d0c7e1642`. The added dynamic allocations have matching normal-path frees, and `checkstyle9.pl --strict` passes. One correction remains in the feature gating.

The [allocation total excludes unavailable L3/DF classes](https://github.com/ojanerif/freebsd-src/blob/686d0c7e16428b0e83c5b0dd58c9612f90f07cea/sys/dev/hwpmc/hwpmc_amd.c#L1038-L1065), but `amd_l3_npmcs` and `amd_df_npmcs` remain unchanged. [Registration skips those classes](https://github.com/ojanerif/freebsd-src/blob/686d0c7e16428b0e83c5b0dd58c9612f90f07cea/sys/dev/hwpmc/hwpmc_amd.c#L1092-L1124), while [`amd_get_msr()` still includes the ungated L3 count in its row offsets](https://github.com/ojanerif/freebsd-src/blob/686d0c7e16428b0e83c5b0dd58c9612f90f07cea/sys/dev/hwpmc/hwpmc_amd.c#L771-L808).

With L3 PMCs unavailable and DF present, the first DF row is classified as L3. A standalone matrix using the exact function on the physical FreeBSD host reproduces the wrong RDPMC selector:

```text
scenario: 6 core, L3 unavailable, 16 DF; first registered DF row = 6
amd_get_msr(6): RDPMC ECX 10
expected: RDPMC ECX 6
```

Normalize the counts once after CPUID parsing, since the extended leaf can overwrite `amd_df_npmcs`, then sum those same counts:

```c
if ((amd_feature2 & AMDID2_PTSCEL2I) == 0)
	amd_l3_npmcs = 0;
if ((amd_feature2 & AMDID2_PNXC) == 0)
	amd_df_npmcs = 0;

npmcs_total = amd_core_npmcs + amd_l3_npmcs + amd_df_npmcs +
	amd_umc_npmcs;
```

The matrix returns the expected ECX 6 after normalizing `amd_l3_npmcs` to zero. This keeps allocation, registration and row mapping consistent. The defect predates dynamic allocation, but the count normalization described in the PR is not present in this revision. The `CTASSERT`, lower-bound `KASSERT`, and pre-L3/pre-DF cursor assertions advertised in the description and commit message are also absent.

The [current decoding masks](https://github.com/ojanerif/freebsd-src/blob/686d0c7e16428b0e83c5b0dd58c9612f90f07cea/sys/dev/hwpmc/hwpmc_amd.h#L35-L64), plus six L3 counters, limit the total to 15 + 6 + 63 + 255 = 339, below [`AMD_NPMCS_MAX = 342`](https://github.com/ojanerif/freebsd-src/blob/686d0c7e16428b0e83c5b0dd58c9612f90f07cea/sys/dev/hwpmc/hwpmc_amd.h#L199-L202). Keep the `KASSERT` as a future-code invariant. Any runtime guard should reject an unsupported total before allocation, not clamp it: registration would still use the larger class counts and could overrun the allocation.

The UMC count field in the description and commit message should be `CPUID Fn8000_0022h EBX[23:16]`, not ECX; ECX holds the active-UMC mask.

Validation complete on an EPYC 9654. I built and booted both `HWPMC_DBG` with built-in hwpmc and `HWPMC_KLD` with the matching loadable `hwpmc.ko`, from `686d0c7e1642`, with `INVARIANTS`, `WITNESS`, and `HWPMC_HOOKS`. Every CPU exposed 32 contiguous K8 rows: 6 core + 6 L3 + 16 DF + 4 UMC. Both L3 and DF are exposed on this server; the absent-L3 case above was exercised in the standalone matrix, not through its hardware configuration.

Runtime results: Process instructions/cycles counting and serialized Core, correctly qualified L3, and UMC system counting returned nonzero values. Process and system sampling logs parsed with 201,157 and 200,487 samples respectively. Rows returned to `FREE` after each test.

Native testing also exposed pre-existing issues that should be tracked separately from this allocation change.

Core and L3 process-counter mappings returned ECX 0–5 and 10–15 correctly. However, all four process-mode UMC allocations reached `amd_get_msr()`, and `pmc_get_msr()` incorrectly reported success with the unchanged `gm.pm_msr` value copied from userspace. [`libpmc` leaves this field uninitialized](https://github.com/ojanerif/freebsd-src/blob/686d0c7e16428b0e83c5b0dd58c9612f90f07cea/lib/libpmc/libpmc.c#L1582-L1590); this is not a kernel-memory disclosure. The callback returns positive `EINVAL`, but [`hwpmc_mod.c` checks for `< 0`](https://github.com/ojanerif/freebsd-src/blob/686d0c7e16428b0e83c5b0dd58c9612f90f07cea/sys/dev/hwpmc/hwpmc_mod.c#L4431-L4435); successful `copyout()` then overwrites that error. Both the callback and adjacent `copyout()` checks should test for nonzero errors. The current UMC tests use `PMC_MODE_SC`, which is rejected before reaching the callback. I'll update this regression test to use a process-virtual UMC allocation attached to its owner, require `-1` with `errno == EINVAL`, and avoid executing RDPMC if the call unexpectedly succeeds.

DF runtime remains skipped, not passed. On the exact kernel, `local_processor_read_data_beats_cs0` fails allocation with `Invalid argument (AMD PMU config has unsupported bits 0x7000000)`, reflecting the Family 19h libpmc/kernel encoding-mask mismatch. The reported `13/13 passed (0 failed)` aggregate includes this DF skip and does not establish DF runtime coverage. Passed, failed, broken and skipped totals need to be reported separately.

Module unload is also not leak-free. A controlled five-cycle comparison retained five allocations and 10,240 bytes with both the `e1f1fc51273f` baseline module and the `686d0c7e1642` module: one 2,048-byte allocation per unload. No additional retention was measured for this PR, but the shared retention needs a separate fix.

Reviewed by: Davi Chaves Azevedo <dchavesa@amd.com>

Sponsored by: AMD

ojanerif:
Thank you for the detailed review and for testing on an EPYC 9654 — the coverage is much appreciated.

**Finding 1 — Feature-gating bug in `amd_get_msr()` row offsets**
Confirmed and fixed. The ternary in `npmcs_total` correctly excluded unavailable L3/DF from the allocation, but `amd_l3_npmcs` and `amd_df_npmcs` retained their defaults, leaving `amd_get_msr()` inconsistent with the registered layout. On a system where L3 is absent, `amd_get_msr()` would misclassify the first DF row as L3 and return the wrong RDPMC ECX selector. The fix normalizes those globals against the feature bits before computing `npmcs_total`, so allocation, registration, and MSR row mapping are all derived from the same values.

**Finding 2 — `AMD_NPMCS_MAX` and KASSERT guidance**
Agreed. The actual maximum reachable by the current decoding masks is 339, below `AMD_NPMCS_MAX = 342`. The upper-bound KASSERT is worth keeping as a forward-looking invariant; we will not clamp the value.

**Finding 3 — Wrong CPUID field cited for UMC count**
Confirmed. The UMC counter field is `CPUID Fn8000_0022h EBX[23:16]` (NumUMCCounters), not ECX. ECX holds the active-UMC mask. Fixed in the commit message.

**Pre-existing issue 1 — `pmc_get_msr()` sign check and uninitialized field**
Valid observation. The `< 0` check in `hwpmc_mod.c` misses positive `EINVAL` returns from the callback, and `libpmc` leaves `gm.pm_msr` uninitialized. Both predate this change. We will track and address them in a separate PR.

**Pre-existing issue 2 — DF runtime skip misreported as pass**
Correct. The `Invalid argument (AMD PMU config has unsupported bits 0x7000000)` failure reflects a Family 19h encoding-mask mismatch between libpmc and the kernel for DF events. The test is being skipped but counted in the pass aggregate. This is a problem in the test reporting, not the driver, and predates this change. We will fix the test accounting separately.

**Pre-existing issue 3 — Module unload leak**
Confirmed. The 2048-byte-per-unload retention appears in both the baseline and our commit, so it was not introduced here. We will investigate and address it in a separate PR.

GuestAUser (approved):
I re-reviewed `c52aa5121e55028eb09b7f1c56c0f231a5335ecc` against `686d0c7e1642`. The requested code correction is resolved, and I approve this dynamic-allocation/count-normalization change.

The L3/DF globals are now normalized after CPUID decoding, before computing the allocation total. Allocation, descriptor registration, and `amd_get_msr()` therefore use consistent counts. The new allocations also have matching frees in the inspected normal and initialization-error paths.

Verification:

- The original absent-L3/present-DF case reproduces ECX **10** on the old revision and returns the expected ECX **6** on the revised code.
- All **16** count/mapping matrix cases passed, covering all four L3/DF feature combinations, UMC absent/present, and the decoded-count maximum of 339.
- All **six** extracted per-CPU allocation/free cases passed with AddressSanitizer and UndefinedBehaviorSanitizer, using table sizes 4, 6, 22, 26, 32, and 339.
- `checkstyle9.pl --strict`: **0 errors, 0 warnings**.

Please finish the text-only cleanup: the PR body still says **ECX[23:16]** instead of **EBX[23:16]**, and advertises removed assertions. The commit message also still claims a `CTASSERT` that is absent. Update the descriptions rather than restoring the redundant assertions.

The pre-existing GETMSR error propagation, DF encoding mismatch/test accounting, and IBS teardown retention remain separate follow-ups, not blockers for this allocation change.

Sponsored by: AMD

ojanerif:
Thank you @mashtizadeh and @GuestAUser for the reviews and testing — the EPYC 9654 validation was particularly valuable in catching the `amd_get_msr()` row offset bug.

The PR is now updated with both fixes addressed (Finding 1 — feature-gating normalization, Finding 3 — CPUID field corrected to EBX). The pre-existing issues noted in the review are tracked separately for follow-up PRs.

@mhorne @emaste — this PR is ready for your review when you have a chance. It is a prerequisite for the UMC clock-gating fix in #2422.

mhorne (approved):
LGTM. Thanks! The final diff is quite clean.

I will merge ASAP.

PR #2418, opened 2026-09-10
imgact_aout: fix i386 header size overflow check

Brine28:
exec_aout_imgact() checks that "virtual_offset + a_text + a_data + bss_size" does not exceed UINT_MAX before trusting those (fully attacker-controlled, straight from the file's uint32_t header fields) sizes to build the process's VM mappings. This check was guarded by #ifdef __amd64__, so it did not exist at all on i386 -- the architecture where the native (non-freebsd32-compat) a.out personality actually runs.

Worse, simply removing the #ifdef would not have been enough: on an ILP32 platform "unsigned long" is itself only 32 bits, so the addition would silently wrap *before* the ">  UINT_MAX" comparison ever ran, making the check a no-op there regardless.

Concretely, a 0-byte file with a_text = a_data = 0x80000000 (each individually page-aligned, so it also passes the earlier alignment check) sums to exactly 0 in 32-bit arithmetic, which then also sails past the "a_data + a_text > imgp->attr->va_size" file-size sanity check a few lines down, since 0 is never greater than any real file's size.

Fix by widening the sum to uint64_t before comparing, which closes the gap on ILP32 platforms and is a no-op behavioural change on LP64 ones (where "unsigned long" was already 64 bits).

### DIFF
--- a/sys/kern/imgact_aout.c
+++ b/sys/kern/imgact_aout.c
@@ -239,13 +239,25 @@ exec_aout_imgact(struct image_params *imgp)
 	    a_out->a_entry >= virtual_offset + a_out->a_text ||
 
 	    /* text and data size must each be page rounded */
-	    a_out->a_text & PAGE_MASK || a_out->a_data & PAGE_MASK
+	    a_out->a_text & PAGE_MASK || a_out->a_data & PAGE_MASK ||
 
-#ifdef __amd64__
-	    ||
-	    /* overflows */
-	    virtual_offset + a_out->a_text + a_out->a_data + bss_size > UINT_MAX
-#endif
+	    /*
+	     * overflows: a_text/a_data/a_bss are attacker-controlled
+	     * uint32_t fields straight from the file header. This sum
+	     * must be computed in a type wider than 32 bits before
+	     * comparing against UINT_MAX: on ILP32 platforms (e.g. i386,
+	     * where this is the native, non-compat a.out personality)
+	     * "unsigned long" is itself only 32 bits, so the addition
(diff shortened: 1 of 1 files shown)


Brine28:
Hi @kib and @emaste Sorry for the direct ping but I wanted to bring this to your attention as it fixes a silent integer overflow on ILP32 platforms (like i386) during attacker controlled a.out header parsing. Since it involves kernel image activation and a potential vulnerability, I'd highly appreciate your review when you have a moment. Thanks

sys/kern/imgact_aout.c
```diff
@@ -141,7 +141,7 @@ struct sysentvec aout_sysvec = {
 	.sv_maxssiz	= &ia32_maxssiz,
 	.sv_flags	= SV_ABI_FREEBSD | SV_AOUT | SV_ILP32 | SV_SIGSYS,
 	.sv_set_syscall_retval = ia32_set_syscall_retval,
-	.sv_fetch_syscall_args = ia32_fetch_syscall_args,
+	.sv_fetch_syscall_args = freebsd32_fetch_syscall_args,
```
markjdb: Why did you change this?

Brine28:
I don't have my computer with me right now I'm trying to handle this on my phone, and things got a bit weird.

sys/kern/imgact_aout.c
Brine28: Okay, I finally got it sorted out. As I mentioned, I don't have my computer with me I had tried doing it on my phone thinking Termux would take too long but ran into some issues. It should be all set now.

PR #2422, opened 2026-09-11
hwpmc/amd: prime UMC perfmon block before programming counter registers

ojanerif:
UMC perfmon counter registers are clock-gated when no UMC_PerfMonCtl
Enable bits are set, causing writes to UMC_PerfMonCntr to be silently
dropped.  This was observed empirically on Zen 4 (Genoa) and Zen 5
(Turin) silicon.

Reserve the last counter slot of each UMC channel to hold the perfmon
block active with a non-counting event (EventSelect=0, Enable=1),
ensuring that writes to any counter register in the same channel
complete as expected.

Each UMC channel has its own independent clock gate, so one slot per
channel is reserved (amd_umc_nchannels, derived from
amd_umc_npmcs / pmcs_per_umc).  Reserved slots are not registered with
MI code and are not visible to userspace.

UMC perfmon MSRs are package-scoped, not per-logical-CPU.  A per-socket
atomic reference count (amd_umc_gate_refcnt[pkg]) ensures the gate MSRs
are armed on the first amd_pcpu_init() call for a given socket and
disarmed only when the last CPU on that socket calls amd_pcpu_fini().
pmc_amd_finalize() issues a synchronous smp_rendezvous() sweep as a
fallback before freeing the gate state.

All gate globals are zeroed in pmc_amd_finalize() and in the error path
of pmc_amd_initialize() so that a kldunload+kldload cycle or a failed
initialization starts with clean state.

Depends on: #2415 (hwpmc/amd: replace static pmcdesc[] and
pc_amdpmcs[] with dynamic allocation) — must be merged first.

### DIFF
--- a/sys/dev/hwpmc/hwpmc_amd.c
+++ b/sys/dev/hwpmc/hwpmc_amd.c
@@ -65,7 +65,7 @@ struct amd_descr {
 
 static int amd_npmcs;
 static int amd_core_npmcs, amd_l3_npmcs, amd_df_npmcs, amd_umc_npmcs;
-static struct amd_descr amd_pmcdesc[AMD_NPMCS_MAX];
+static struct amd_descr *amd_pmcdesc;
 struct amd_event_code_map {
 	enum pmc_event	pe_ev;	 /* enum value */
 	uint16_t	pe_code; /* encoded event mask */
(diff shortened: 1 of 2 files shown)


ojanerif:
Updated the branch to rebase on top of the latest `pr-a/umc-dynamic-alloc-v3` commit, which incorporates the `amd_get_msr()` row offset fix from Finding 1 in the #2415 review. The conflict was in the `npmcs_total` computation: the normalization of `amd_l3_npmcs`/`amd_df_npmcs` (from the base commit) now precedes the UMC gate reservation block (from this commit), keeping both changes intact.

PR #2427, opened 2026-09-12
msun: fix powl() intermediate overflow with extreme arguments

serg-alexv:
## Summary

This patch addresses a failure mode in the `ld80` implementation of
`powl()` for extreme finite arguments.

The table-driven reduction can overflow an intermediate value before the
range checks run.  The resulting infinity can then become a NaN and reach
the floating-to-integer conversion used by the reduction path.

The patch:

- removes the unused `i = w` conversion;
- adds a conservative integer-domain early classification for arguments
  whose result is provably outside the `ld80` range;
- performs the range operation with finite operands so hardware exception
  flags and directed rounding remain authoritative; and
- adds black-box ATF coverage for the bug case, signed results, and all four
  rounding modes.

The guard is intentionally sufficient rather than complete: it only handles
cases where `|y| >= 2^15` and `|log2(|x|)| >= 1`.

## Testing

Tested on FreeBSD 15.1-RELEASE amd64 with a locally built shared `libm`:

```
LD_LIBRARY_PATH=/path/to/lib/msun kyua test -k Kyuafile exponential_test
```

Result: 5/5 test cases passed, including `powl_extreme`.

PR: 292988

cc @dag-erling

### DIFF
--- a/lib/msun/ld80/e_powl.c
+++ b/lib/msun/ld80/e_powl.c
@@ -363,7 +363,6 @@ if( x <= 0.0L )
 
 if( iyflg )
 	{
-	i = w;
 	w = floorl(x);
 	if( (w == x) && (fabsl(y) < 32768.0) )
 		{
@@ -375,6 +374,33 @@ if( iyflg )
 
 if( nflg )
 	x = fabsl(x);
+
+/*
+ * Pure integer bounds classification to avoid fenv pollution.
+ * If |y| >= 2^15 and |log2(x)| >= 1, the product definitively
+ * exceeds ld80 MEXP/MNEXP limits without requiring FP multiplication.
+ * Prevents NaN generation from intermediate calculations and UB on
+ * conversion (bug 292988).
+ */
+{
+	int ex = ilogbl(x);
+	int ey = ilogbl(y);
+
+	if (ey >= 15 && (ex <= -2 || ex >= 1)) {
+		int q_sign = signbit(y) ^ (ex < 0);
+		volatile long double v, s;
+
+		if (q_sign)
+			v = s = 0x1p-10000L; /* Guaranteed underflow */
+		else
+			v = s = 0x1p9000L;  /* Guaranteed overflow */
+
+		if (nflg && yoddint)
+			s = -s;
+		v = v * s;
+		return (v);
+	}
+}
 
 /* separate significand from exponent */
 x = frexpl( x, &i );
--- a/lib/msun/tests/exponential_test.c
+++ b/lib/msun/tests/exponential_test.c
@@ -156,6 +156,64 @@ ATF_TC_BODY(exp2l, tc)
 	}
 }
 
+ATF_TC_WITHOUT_HEAD(powl_extreme);
+ATF_TC_BODY(powl_extreme, tc)
+{
+#if LDBL_MANT_DIG == 64
+	const int rmodes[] = { FE_TONEAREST, FE_UPWARD, FE_DOWNWARD,
+	    FE_TOWARDZERO };
+	long double actual, expected;
+	int i, j;
+
+	struct {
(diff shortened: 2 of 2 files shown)


lib/msun/ld80/e_powl.c
```diff
@@ -376,6 +375,33 @@ if( iyflg )
...
+ * Prevents NaN generation from intermediate calculations and UB on
+ * conversion (bug 292988).
+ */
+{
+	int ex = ilogbl(x);
+	int ey = ilogbl(y);
+
+	if (ey >= 15 && (ex <= -2 || ex >= 1)) {
+		int q_sign = signbit(y) ^ (ex < 0);
+		volatile long double v, s;
```
kostikbel, line 391: What is the purpose of the volatile qualifier there?

serg-alexv: volatile is used here to prevent the compiler from replacing the intended calculation with a precomputed constant.

There are two stages:
1. Compile time: the compiler translates the program and may evaluate expressions whose operands are already known.
2. Runtime: the processor executes the remaining instructions.

Without floating-point environment requirements, the compiler may fold an overflowing multiplication of known constants into infinity. The multiplication would then never happen at runtime. (Clang documentation)

Our code needs the operation’s effects as well as its numerical result: the appropriate floating-point exception flags and the result selected by the current rounding mode.

In this declaration:
[volatile long double v, s;]

Both variables are volatile. Their reads and writes must be preserved as observable accesses. The compiler cannot simply replace those reads with the constants previously assigned, which prevents constant folding of this multiplication. (https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf, §6.7.3)

The patch then multiplies v * s, applying the sign to s beforehand so that rounding operates on the correctly signed product.

One qualification matters: volatile alone does not guarantee correct floating-point environment semantics. It protects the accesses; appropriate compiler settings and tests are still needed to establish correct exception and rounding behavior. (https://clang.llvm.org/docs/UsersManual.html#floating-point-environment)

serg-alexv:
Fresh validation of the current head (`e80b8a3e2b8117932ea094695840aed4d32b110a`) is now attached to the canonical bug:

**[Bugzilla #292988 — evidence archive, attachment 274671](https://bugs.freebsd.org/bugzilla/attachment.cgi?id=274671)**

On a fresh FreeBSD 15.1-RELEASE/amd64 VM (Clang 19.1.7, Kyua 0.13), `exponential_test` passed **5/5, with no skips**, at 2026-09-12 11:21 UTC. The same test executable with the system libm failed `powl_extreme`; inherited ktrace records confirm that the passing test processes loaded the newly built libm.

The archive includes raw Kyua databases, reports, build logs, source/binary hashes, the exact patch, and reproduction instructions. The source files match this PR and were not edited for the run. Building current source on 15.1 required sys headers from the same commit; validation covers the exponential suite, not all msun tests. Details and failed setup attempts are retained in the archive.

For precision: these are **new results**, not a recovered transcript of the earlier 5/5 summary. The original transcript remains unavailable. The new records provide a reproducible check of the current patch.

Archive SHA256: `9c2f2e0fea06312d05ef1967ebdb8ee272ba94d4bfcbae15ce5e323fe2abe8b6`.

lib/msun/ld80/e_powl.c
kostikbel: The texts from you have an awful smell.  But let's suppose that I am talking to you and not to something that put words into your mouth.

Are you saying that compiler is allowed to ignore the FP environment settings that the code establishes?  If yes, we either miss `#pragma STDC FENV_ACCESS` or the compiler has bugs?

Can you demonstrate the difference between generated code with and without the volatile qualifier, please?  Preferably, once for clang and once for gcc.

PR #2428, opened 2026-09-14
Panic fix for failpoint pause

markr111:
Provide PNOLOCK flag to satisfy invariant in _sleep when a failpoint pause action is triggered.

# Testing
## Before Fix:
`sysctl debug.fail_point.sysctl_running=pause` triggers panic in `_sleep()` when INVARIANTS compiled into kernel

## After fix:
Same command no longer triggers panic, but does cause the `sysctl_running` failpoint handler to get stuck indefinitely in `sysctl_root_handler_locked()`, appears to hold up all other kernel threads. Unable to disable the failpoint from other threads. But this is a completely different problem, haven't investigated yet.


markr111:
@rlibby  - FYI

PR #2436, opened 2026-09-17
nuageinit: fix support on OpenStack

etam:
Change added in 90a7728cd8905cd26b90d06f7873df8bad43ae9a contained a lot of bugs:
- If all network interfaces correctly respond to DHCP, then pwait was waiting forever.
- If there was more than 1 interface, then "cat /tmp/ephemeraldhcp.*.pid" joined pid numbers into one long string.
- "for iface in $left; do kill -15 $left; done" is not using $iface.

Tested by:	adam.mizerski@ovhcloud.com
Sponsored by:	OVHcloud

Signed-off-by: adam.mizerski@ovhcloud.com


bapt:
merged thanks!

